PR #729, opened 2021-10-20
RUMM-1661: Gradle 7.2, AGP 7.0.3, Kotlin 1.5.21

0xnm:
### What does this PR do?

This PR updates Gradle to version `7.2`, AGP to version `7.0.3`, Kotlin to `1.5.21`.

Basically it is the same as in https://github.com/DataDog/dd-bridge-android/pull/43 (check details there), but for a wider scope.

Important notes:

* After update you probably will need to clear IDE cache manually by removing files in `~/Library/Caches/JetBrains (or Google)/<product>`, otherwise it may be glitches in the IDE
* I've tested SDK compatibility with Kotlin 1.4 by linking it to the Shopist Android project.- everything works fine. Compatibility with Kotlin 1.3 wasn't tested.
* Some libraries/tools required the update to have Kotlin 1.5 only libs in the classpath. For example, version `1.9.0` of Kotlin Poet now generates classes which are compatible with [explicit API mode](https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md), so there is a plenty of changes in the test classes.
* Licenses and surface check tasks now cover all library modules.

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/config/DetektConfig.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/config/DetektConfig.kt
@@ -14,7 +14,7 @@ import org.gradle.kotlin.dsl.withType
 fun Project.detektConfig(excludes: List<String> = emptyList()) {
 
     extensionConfig<DetektExtension> {
-        input = files("$projectDir/src/main/kotlin")
+        source = files("$projectDir/src/main/kotlin")
         config = files("${project.rootDir}/detekt.yml")
 
         reports {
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -164,6 +164,34 @@ tasks.register("lintCheckAll") {
     )
 }
 
+tasks.register("checkThirdPartyLicensesAll") {
```
xgouchet, line 167: Good call !

mariusc83 (approved):

PR #741, opened 2021-10-29
POC to discuss the WebView tracking strategy

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
@@ -1,0 +1,137 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.webview
+
+import com.datadog.android.core.internal.CoreFeature
+import com.datadog.android.core.internal.persistence.DataWriter
+import com.datadog.android.core.internal.time.TimeProvider
+import com.datadog.android.core.internal.utils.sdkLogger
+import com.datadog.android.rum.GlobalRum
+import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.domain.scope.RumApplicationScope
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.internal.monitor.DatadogRumMonitor
+import com.datadog.android.rum.model.ActionEvent
+import com.datadog.android.rum.model.ErrorEvent
+import com.datadog.android.rum.model.LongTaskEvent
+import com.datadog.android.rum.model.ResourceEvent
+import com.datadog.android.rum.model.ViewEvent
+import com.google.gson.JsonObject
+import com.google.gson.JsonParseException
+
+internal class RumEventConsumer(
+    private val dataWriter: DataWriter<Any>,
+    private val timeProvider: TimeProvider
+) {
+
+    private var currentOffset = timeProvider.getServerOffsetMillis()
+
+    fun consume(event: JsonObject, eventType: String) {
+        if (!isSessionDropped()) {
+            val mappedEvent = map(event, eventType)
+            if (mappedEvent != null) {
+                dataWriter.write(mappedEvent)
+            }
+        }
+    }
+
+    private fun map(event: JsonObject, eventType: String): JsonObject? {
+        val rumContext = GlobalRum.getRumContext()
+
+        return try {
+            when (eventType) {
+                VIEW_EVENT_TYPE -> {
+                    currentOffset = timeProvider.getServerOffsetMillis()
+                    ViewEvent.fromJson(event.toString()).mapToMobile(rumContext, currentOffset)
+                }
+                ACTION_EVENT_TYPE -> {
+                    ActionEvent.fromJson(event.toString()).mapToMobile(rumContext, currentOffset)
+                }
+                RESOURCE_EVENT_TYPE -> {
+                    ResourceEvent.fromJson(event.toString()).mapToMobile(rumContext, currentOffset)
+                }
+                ERROR_EVENT_TYPE -> {
+                    ErrorEvent.fromJson(event.toString()).mapToMobile(rumContext, currentOffset)
+                }
+                LONG_TASK_EVENT_TYPE -> {
+                    LongTaskEvent.fromJson(event.toString()).mapToMobile(rumContext, currentOffset)
+                }
+                else -> {
+                    null
+                }
+            }
+        } catch (e: JsonParseException) {
+            sdkLogger.e(JSON_PARSING_ERROR_MESSAGE, e)
+            null
+        }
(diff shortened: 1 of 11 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
```diff
@@ -0,0 +1,137 @@
...
+        } catch (e: JsonParseException) {
+            sdkLogger.e(JSON_PARSING_ERROR_MESSAGE, e)
+            null
+        }
+    }
+
+    private fun isSessionDropped(): Boolean {
+        val appScope = (GlobalRum.get() as? DatadogRumMonitor)?.rootScope as? RumApplicationScope
+            ?: return false
+        return (appScope.childScope as? RumSessionScope)?.keepSession ?: false
```
ncreated, line 76: This assumes that the native session was already started and that the native code made consideration of sampling. In case when the webview is a very first thing displayed by the app (not preceded by any native "start view") we can't rely on this information, no?

mariusc83: yes...but I don't see how you could have a WebView event without having the native code started already....normally we start a session once the application started. I am not sure we need to add an extra logic here which will sample the events locally in case there is no session yet started. What do you all think ?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
```diff
@@ -0,0 +1,137 @@
...
+import com.datadog.android.rum.model.ActionEvent
+import com.datadog.android.rum.model.ErrorEvent
+import com.datadog.android.rum.model.LongTaskEvent
+import com.datadog.android.rum.model.ResourceEvent
+import com.datadog.android.rum.model.ViewEvent
+import com.google.gson.JsonObject
+import com.google.gson.JsonParseException
+
+internal class RumEventConsumer(
+    private val dataWriter: DataWriter<Any>,
```
ncreated, line 27: Which writer is this? The one just putting data into file or the one making tracking consent consideration? If it's the latter, wouldn't it be cleaner to use specialised type here, for clarity?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private fun map(event: JsonObject, eventType: String): JsonObject? {
+        val rumContext = GlobalRum.getRumContext()
+
+        return try {
+            when (eventType) {
+                VIEW_EVENT_TYPE -> {
+                    currentOffset = timeProvider.getServerOffsetMillis()
+                    ViewEvent.fromJson(event.toString()).mapToMobile(rumContext, currentOffset)
```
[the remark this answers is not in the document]
ncreated: An alternative approach could be to not parse with out specialised types (I agree, it might surface many compatibility issues), but just apply patch to the `JSONObject`, something this:
```kotlin
fun mapViewEvent(browserView: JSONObject, context: RumContext, timeOffset: Long): JsonObject? {
   try {
      browserView.replace(keyPath: "application.id", with: context.application.id)
      browserView.replace(keyPath: "session.id", with: context.session.id)
      browserView.replace(keyPath: "service", with: Core.serviceName)
      browserView.replace(keyPath: "view.date", with: browserView["view"]["date"] + offset)
      return browserView // success
   } catch {
      // notify internal error
      return null // failure
   }
}
```
This way we don't rely on schema validation, but we limit the expected interface to only values we need to modify. This won't break on future schema changes unless it's a change to these core fields.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
```diff
@@ -0,0 +1,137 @@
...
+        )
+        return copy.toJson().asJsonObject
+    }
+
+    private fun ActionEvent.mapToMobile(context: RumContext, timeOffset: Long): JsonObject {
+        val copy = this.copy(
+            application = this.application.copy(id = context.applicationId),
+            session = this.session.copy(id = context.sessionId),
+            service = CoreFeature.serviceName,
+            date = this.date + timeOffset
```
ncreated, line 94: `+` we still need to solve the problem with `timeOffset` being impacted by NTP time sync, right?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
ncreated: IMO all apps that present a webview as their first view fall into this condition - there won't be "start view" called, so [`keepSession` will be always `false`](https://github.com/DataDog/dd-sdk-android/blob/854b11ac8ce0feb1dd758570c683b2139b5d0822/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt#L44) and no browser event will go through.

Even if we ignore this case, there are some remaining considerations regarding session lifecycle. Now, it's driven by events being sent through RUM Scopes. Because we bypass RUM scopes for browser events, IMO we need some other solution for managing RUM session lifecycle for browser SDK events (ideally, we should refactor to have one shared solution for both event pipelines). Otherwise the session won't timeout after `15min` of inactivity, and could last for more than `4h`.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
mariusc83: if the events order is being kept we can go with this simpler way by just reseting the `timeOffset` when a new `ViewEvent` comes in. If we can have multiple events in the same time then indeed we need to rethink this and go with the `Map<ViewId,Offset>` approach

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
mariusc83: Is the one having the tracking consent also

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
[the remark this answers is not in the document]
mariusc83: was thinking exactly to the same approach but this will mean that we will not check if the events format is correct. I would rather go with the simplified approach and only replace the required properties in the JSON object than to have all these transformations.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
mariusc83: Yes, indeed that's a thing that I did not consider, but I think the reason why is that in Android any `WebView` is attached to an activity/fragment so there will be at least one View started before the `WebView` is loaded. Should we exclude this possibility in `iOS` and open an RFC on this part to discuss how to handle the session timeout ?

PR #743, opened 2021-11-03
RUMM-1717: Read custom error source type value

0xnm:
### What does this PR do?

This change introduces the possibility for specify custom error source type, so that we can distinguish errors coming from native Android apps vs another layer calling Android SDK (like Browser or React Native SDK).

This is done by adding `_dd.error.source_type` key to the `attributes` which can be supplied to the `addError` or `addErrorWithStacktrace` calls (`stopResourceWithError` is out of scope, because it is not used currently by non-native SDKs). Provided value will override the default one `Android`. Supported values are `react-native` and `browser` for now, `iOS` is out of scope, because it is not technically possible to have it.


xgouchet (approved):

mariusc83 (approved):
LGTM

PR #747, opened 2021-11-04
RUMM-1708: Allow to specify custom SDK version

0xnm:
### What does this PR do?

This change introduces a possibility to override native SDK version which is need for other SDKs using the native one, like React Native SDK.

This is achieved by having a new property `_dd.sdk_version` in the `additionalConfiguration` dictionary which is passed during the native SDK initialization.

We are going to use custom SDK version only for API v2 uploaders, API v1 uploaders are left untouched.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -278,8 +278,14 @@ object Datadog {
         // so some things may yet not be initialized -> not accessible, some things may already be
         // initialized and be not mutable anymore
         additionalConfiguration[DD_SOURCE_TAG]?.let {
-            if (it.toString().isNotBlank()) {
-                CoreFeature.sourceName = it.toString()
+            if (it is String && it.isNotBlank()) {
+                CoreFeature.sourceName = it
+            }
+        }
+
+        additionalConfiguration[DD_SDK_VERSION_TAG]?.let {
+            if (it is String && it.isNotBlank()) {
+                CoreFeature.sdkVersion = it
             }
         }
     }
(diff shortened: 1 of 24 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -278,8 +278,14 @@ object Datadog {
         // so some things may yet not be initialized -> not accessible, some things may already be
         // initialized and be not mutable anymore
         additionalConfiguration[DD_SOURCE_TAG]?.let {
-            if (it.toString().isNotBlank()) {
-                CoreFeature.sourceName = it.toString()
+            if (it is String && it.isNotBlank()) {
```
xgouchet, line 281: What if it isn't a String ? Shouldn't we still deal with it ?

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
0xnm: What can be the case for that and what type can be instead? I think all external platforms have string as primitive.

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -278,8 +278,14 @@ object Datadog {
...
+                CoreFeature.sourceName = it
+            }
+        }
+
+        additionalConfiguration[DD_SDK_VERSION_TAG]?.let {
+            if (it is String && it.isNotBlank()) {
+                CoreFeature.sdkVersion = it
```
mariusc83, line 288: I am wondering as a follow up on @xgouchet question. Can this parameter be passed a an `Int` here somehow from react-native ?

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
0xnm: We are following semantic versioning, so it will be always as `x.y.z` in the string format imo. Don't see at what point it can be changed.

mariusc83 (approved):

PR #755, opened 2021-11-09
RUMM-1728 Introduce the RumEventConsumer

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/DatadogEventBridge.kt
```diff
@@ -49,4 +49,24 @@ class DatadogEventBridge {
         }
         return origins.toString()
     }
+
+    // endregion
+
+    // region internal
+
+    private fun getWebEventConsumer(): WebEventConsumer {
```
xgouchet, line 57: Is there a reason why you're using this instead of a `by lazy (mode = LazyThreadSafetyMode.SYNCHRONIZED){ … }`?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
```diff
@@ -6,13 +6,157 @@
...
+                }
+            }
+        } catch (e: JsonParseException) {
+            sdkLogger.e(JSON_PARSING_ERROR_MESSAGE, e)
+            null
+        }
+    }
+
+    private fun getOffset(viewId: String): Long {
+        // The JS bridge functions are being called from a different Thread so it is very
```
xgouchet, line 115: Do you mean that multiple JS calls can be dispatched to the native side in different threads ? I'd have thought they would all be processed in the same thread ?

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/DatadogEventBridge.kt
mariusc83: hmmm...didn't know about that trick...I am going to have a look.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
mariusc83: Well I am actually not sure about this and this may depend on the WebKit version, I was just thinking just to make sure we handle this just in case. In case there is a chance that this is handled differently on one specific WebKit version then we will have problems.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/DatadogEventBridge.kt
```diff
@@ -49,4 +49,24 @@ class DatadogEventBridge {
...
+        return webEventConsumer ?: synchronized(this) {
+            val newInstance = webEventConsumer ?: webEventConsumer
```
0xnm, line 59:
```suggestion
            val newInstance = webEventConsumer
```

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
```diff
@@ -6,13 +6,138 @@
...
+    private val webRumEventMapper: WebRumEventMapper = WebRumEventMapper()
+) {
+
+    internal val offsets: LinkedHashMap<String, Long> = LinkedHashMap()
 
     fun consume(event: JsonObject, eventType: String) {
-        // just to make the code compile without warning. This will be removed in the next PR.
-        sdkLogger.i(event.toString() + eventType)
+        val rumContext = getRumContext()
+        if (rumContext != null &&
```
0xnm, line 37: should we emit any warning (or info) message otherwise?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/DatadogEventBridge.kt
mariusc83: Ok I have my answer for this, I actually did try with lazy but apparently you cannot use `lazy` with a `var` and I needed a var there for my unit tests. There is also not possible to have this in the constructor as this can create problems: "the bridge was created before the sdk was initialized" and we do not want that.

dd-sdk-android/src/test/kotlin/com/datadog/android/utils/config/GlobalRumMonitorTestConfiguration.kt
```diff
@@ -13,7 +13,7 @@ import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
 import com.datadog.tools.unit.extensions.config.MockTestConfiguration
 import fr.xgouchet.elmyr.Forge
 
-internal class GlobalRumMonitorTestConfiguration :
+internal open class GlobalRumMonitorTestConfiguration :
```
0xnm, line 16: probably can be reverted

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
mariusc83: not sure honestly here, maybe a dev log ?

0xnm: This is for internal monitoring mostly, because I guess we would like to avoid such situations? And if it happens, we should have enough info to be able to find the root cause (if we consider it is an issue).

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/DatadogEventBridge.kt
0xnm: In which case can we have a situation when bridge is created before the SDK is initialized?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/DatadogEventBridge.kt
```diff
@@ -49,4 +49,23 @@ class DatadogEventBridge {
...
+    private fun getWebEventConsumer(): WebEventConsumer {
+        return webEventConsumer ?: synchronized(this) {
+            val newInstance = webEventConsumer ?: WebEventConsumer(
+                RumEventConsumer(
+                    RumFeature.persistenceStrategy.getWriter(),
+                    CoreFeature.timeProvider
+                ),
+                LogsEventConsumer()
+            )
+            newInstance
```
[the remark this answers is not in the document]
mariusc83: Oh yes, good catch, how could I miss that...hmmm let me check the unit tests 👍

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
xgouchet: Agreed, we should have: 
- A `devlog` to warn the customer that some WebView events will be lost
- Possibly an internal monitoring log if this issue can happen and mean something's wrong internally

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
xgouchet: Could you make a few tests to know exactly if this can **really** happen / refer to official documentation if this is mentionned? I'd rather have a solid basis to decide if we do need this or not.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
xgouchet: On a general basis, if we don't know something about code we use, it means we need to learn it :)

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
mariusc83: yes I did tests on my end, seems that the thread is the same but it looks like a pool thread and I've seen some stackoverflow mentions that WebView is using a thread pool for the bridge. I am going to try to find the source code even though I tried it before and could not find the native implementation in the Webkit where they are doing this. Will keep you posted.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumContextProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import com.datadog.android.core.internal.utils.devLogger
+import com.datadog.android.core.internal.utils.sdkLogger
+import com.datadog.android.rum.GlobalRum
+import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.domain.scope.RumApplicationScope
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.internal.monitor.DatadogRumMonitor
+
+internal class RumContextProvider {
```
xgouchet, line 17: Nice trick, the only downside is that the check is performed outside of this class meaning that multiple calls to getRumContext could still happen and you'd log the issue every time. 

Instead you could have a boolean in here to remember if you already sent an internal log.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumContextProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+        } else {
+            rumContext
+        }
+    }
+
+    companion object {
+        const val RUM_NOT_INITIALIZED_WARNING_MESSAGE = "You are trying to use the WebView " +
+            "tracking API but the RUM feature was not properly initialized."
+        const val RUM_NOT_INITIALIZED_ERROR_MESSAGE = "Trying to consume a bundled rum event" +
+            " but the RUM feature was not yet initialized. Missing the RUM context."
```
xgouchet, line 38: How does this class knows that it is used for webview bundling? Appart from it being in the `webview` package, it could still be generic.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumContextProvider.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.domain.scope.RumApplicationScope
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.internal.monitor.DatadogRumMonitor
+
+internal class RumContextProvider {
+
+    fun getRumContext(): RumContext? {
+        val appScope = (GlobalRum.get() as? DatadogRumMonitor)?.rootScope as? RumApplicationScope
+        val rumContext = (appScope?.childScope as? RumSessionScope)?.getRumContext()
```
xgouchet, line 21: We have a `GlobalRum.getRumContext()` method that you can call that will always have the latest up to date context. This would be even better since you don't need to know that
- the active monitor needs to be a `DatadogRumMonitor`
- the monitor's root scope is a `RumApplicationScope`
- the app's scope has a child scope
- the app's child scope is a RumSessionScope.

Any change to our monitor implementation would ruin this

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
```diff
@@ -127,8 +126,23 @@ internal class RumEventConsumer(
     }
 
     private fun getRumContext(): RumContext? {
```
xgouchet, line 128: Is this still being used ?

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumContextProvider.kt
mariusc83: yeah but this will mix a bit the logic, if you think about it there will not be too many classes as there (one/webview) only ...and this will mean that say you have 3 WebViews in the same time : you will have max 3 logs.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumContextProvider.kt
mariusc83: Yes...in the first place I wanted to use that method, problem is that it is not handling the session update. The only method that does that is the one in the `SessionScope`

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumEventConsumer.kt
mariusc83: Nope, good catch.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumContextProvider.kt
mariusc83: I see your point, I am going to try to change its name.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/RumContextProvider.kt
xgouchet: Hmm… Maybe we should have a mecanism to do this without coupling your work with the implementation details, i.e. add a paramater to "update session if needed" to the `GlobalRum.getRumContext()` ?
Tell you what, use the standard `GlobalRum.getRumContext()` for now, and add a ticket to see if we can have a cleaner way to ensure a new session is updated if needed.

mariusc83: Hmmm...that would make sense...is just that this method is public, I think we will need something more internal for that but I will think about something in the next PR.

mariusc83: My bad, that method is not public so I could actually tweak it to update the session if needed. In the same time I would rather add a new separated method than to actually tweak that one as it is not part of the same context.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebRumEventConsumer.kt
```diff
@@ -0,0 +1,134 @@
...
+        }
+        val rumContext = contextProvider.getRumContext()
+        if (rumContext != null) {
+            val mappedEvent = map(event, eventType, rumContext)
+            if (mappedEvent != null) {
+                dataWriter.write(mappedEvent)
+            }
+        } else {
+            rumFeatureDisabled = true
+        }
```
xgouchet, line 45: Instead of doing this (and losing data), maybe we can write the unmapped event. Because if native rum is disabled, they might still want to have the browser rum events

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebRumEventContextProvider.kt
```diff
@@ -0,0 +1,35 @@
...
+
+internal class WebRumEventContextProvider {
+
+    fun getRumContext(): RumContext? {
+        val rumContext = GlobalRum.getRumContext()
+        return if (rumContext.applicationId == RumContext.NULL_UUID ||
+            rumContext.sessionId == RumContext.NULL_UUID
+        ) {
+            devLogger.w(RUM_NOT_INITIALIZED_WARNING_MESSAGE)
+            sdkLogger.e(RUM_NOT_INITIALIZED_ERROR_MESSAGE)
```
xgouchet, line 22: You can still spam this calling multiple time the `getRumContext()` method.

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebRumEventContextProvider.kt
mariusc83: yes but this will be called only from the `WebRumEventConsumer` and there we have the `rumFeatureDisabled` flag which protects it.

mariusc83: Should we move the flag inside the `WebRumContextProvider` instead ?

xgouchet: I think I made a comment earlier in that sense yes.

mariusc83: Yup, it is done

xgouchet (approved):

0xnm (approved):

PR #757, opened 2021-11-10
RUMM-1611: Track clicks in Jetpack Compose

0xnm:
### What does this PR do?

This change add a small utility function which may simplify click tracking with Jetpack Compose.

Unfortunately, it seems there is no good way to do auto-instrumentation by using public Compose API, so this change is about the manual instrumentation: user has to use it explicitly on the element.

It is worth mentioning that there is very small added value, because it can be just written as:

```
Button(onClick = {
  GloralRum.get().addUserAction(...)
  ...rest of onClick logic...
})
```

But because we proxying original `() -> Unit` type with the help of class with `fun operator invoke()` overload and we have to capture the variables, to avoid new objects creation during the re-composition event, we have to add a bunch of `remember` calls.

~~Another pitfall is `RumActionType` usage. If we use `TAP` and this action had no side-effects, it will be dropped. This is probably fine in the auto-instrumentation approach (because we can indeed have false-positive events which shouldn't be sent and not meaningful), but in case of manual instrumentation I guess user would expect this action to be sent. If we use `CUSTOM` type, then action is sent (we allow `CUSTOM` types to fly even if there is no side-effects), but then in the dashboard it would be just `%action_name%` instead of `tap on %target_name%` as in case of `TAP` usage.~~

~~Maybe we can leverage `attributes` parameter with some internal attribute to allow `TAP` action to be sent even if there is no side-effects in case of manual instrumentation?~~

This PR is using `TAP` action type, question of side-effects will be reviewed in another task.

### DIFF
--- a/dd-sdk-android-compose/README.md
+++ b/dd-sdk-android-compose/README.md
@@ -79,6 +79,19 @@ val configuration = Configuration.Builder(
     .build()
 ```
 
+#### Action tracking
+
+There is no automated instrumentation support for the action tracking in Jetpack Compose. However, report clicks using `trackClicks` method using this example:
+
+```kotlin
+Button(
+    onClick = trackClicks(targetName = "Open View") { ...click logic... }
+    ...
+) {
+    ...layout...
+}
+```
+
 ## Contributing
 
 For details on contributing, read the
--- a/dd-sdk-android-compose/README.md
+++ b/dd-sdk-android-compose/README.md
@@ -81,11 +81,11 @@ val configuration = Configuration.Builder(
 
 #### Action tracking
 
-There is no automated instrumentation support for the action tracking in Jetpack Compose. However, report clicks using `trackClicks` method using this example:
+There is no automated instrumentation support for the action tracking in Jetpack Compose. However, report clicks using `trackClick` method using this example:
 
 ```kotlin
 Button(
-    onClick = trackClicks(targetName = "Open View") { ...click logic... }
+    onClick = trackClick(targetName = "Open View") { ...click logic... }
     ...
 ) {
     ...layout...
--- a/dd-sdk-android-compose/build.gradle.kts
+++ b/dd-sdk-android-compose/build.gradle.kts
@@ -14,6 +14,7 @@ import com.datadog.gradle.config.kotlinConfig
 import com.datadog.gradle.config.ktLintConfig
 import com.datadog.gradle.config.publishingConfig
 import com.datadog.gradle.config.setLibraryVersion
+import com.datadog.gradle.config.taskConfig
 
 plugins {
     // Build
@@ -100,6 +101,11 @@ unMock {
 }
 
 kotlinConfig()
+taskConfig<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
+    kotlinOptions {
+        freeCompilerArgs += listOf("-Xopt-in=kotlin.RequiresOptIn")
+    }
+}
 detektConfig()
 ktLintConfig()
 junitConfig()
--- a/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/ExperimentalTrackingApi.kt
+++ b/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/ExperimentalTrackingApi.kt
@@ -1,0 +1,17 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.compose
+
+/**
+ * Marker for the experimental API.
+ */
+@RequiresOptIn(
+    message = "This is an experimental Jetpack Compose observability API." +
+        " It may change in the future.",
+    level = RequiresOptIn.Level.WARNING
+)
+annotation class ExperimentalTrackingApi
--- a/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
+++ b/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
@@ -25,7 +25,7 @@ import com.datadog.android.rum.RumMonitor
  */
 @ExperimentalTrackingApi
 @Composable
-fun trackClicks(
+fun trackClick(
     targetName: String,
     attributes: Map<String, Any?> = remember { emptyMap() },
     onClick: () -> Unit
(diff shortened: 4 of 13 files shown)


dd-sdk-android-compose/README.md
```diff
@@ -79,6 +79,19 @@ val configuration = Configuration.Builder(
     .build()
 ```
 
+#### Action tracking
+
+There is no automated instrumentation support for the action tracking in Jetpack Compose yet. However, you may report clicks using `trackClicks` method in the following manner:
+
+```kotlin
+Button(
+    onClick = trackClicks(targetName = "Open View") { ...click logic... }
```
ncreated, line 88: I think the added value of this is the abstraction, which may enable us to do internal improvements later, without changing the `RUMMonitor` API 👌.

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
```diff
@@ -0,0 +1,55 @@
...
+ * Creates a proxy around click listener, which will report clicks to Datadog.
+ *
+ * @param targetName Name of the click target.
+ * @param attributes additional custom attributes to attach to the action. Attributes can be
+ * nested up to 9 levels deep. Keys using more than 9 levels will be sanitized by SDK.
+ * @param onClick Click listener.
+ */
+@ExperimentalTrackingApi
+@Composable
+fun trackClicks(
```
ncreated, line 28: Shouldn't it be "taps", not "clicks" 🤔💭?

ncreated (commented):
> Maybe we can leverage attributes parameter with some internal attribute to allow TAP action to be sent even if there is no side-effects in case of manual instrumentation?

I'd rather vote for an explicit element in the `RUMMonitor's` internal API (additional internal parameter or separate action type). It should be easy to have one, as both on Android and iOS we have distinct internal interfaces for `RUMMonitor` ([`AdvanedRumMonitor`](https://github.com/DataDog/dd-sdk-android/blob/f51677801f6966763c24b18e77c78c9764dd2ec5/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/AdvancedRumMonitor.kt#L17) on Android and [`RUMCommandsSubscriber`](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/RUM/AutoInstrumentation/RUMCommandSubscriber.swift) on iOS). The `RUMCommandsSubscriber` on iOS is meant to solve exactly this problem - when particular `RUMCommand` needs something extra, it can be configured outside `RUMMonitor` and passed to it.

dd-sdk-android-compose/README.md
0xnm: true, good point.

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
0xnm: good question! For now it is `click`, because this comes as a wrapper of `onClick` attribute which has specific type. I need to explore other interactions names and types in Compose framework to see if I can generalize this wrapper better.

dd-sdk-android-compose/build.gradle.kts
```diff
@@ -100,6 +101,11 @@ unMock {
 }
 
 kotlinConfig()
+taskConfig<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
+    kotlinOptions {
+        freeCompilerArgs += listOf("-Xopt-in=kotlin.RequiresOptIn")
```
mariusc83, line 106: So in order to use this API you need to specifically `optIn` in the compiler args ?

mariusc83 (approved):

dd-sdk-android-compose/build.gradle.kts
0xnm: Yes https://kotlinlang.org/docs/opt-in-requirements.html. My reasoning behind using that is that once `compose` package is released, its version won't reflect that it is still experimental (because it will be common version, as for other packages), so it is an additional precaution to warn the users.

mariusc83: Hmm...in that case shouldn't you explain it in the docs better ?

0xnm: Not sure, IDE gives you a warning about that and says what you actually need to do.

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/ExperimentalTrackingApi.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.compose
+
+/**
+ * Marker for the experimental API.
+ */
+@RequiresOptIn(
+    message = "This is an experimental animation API for Datadog Tracking." +
```
xgouchet, line 13: Is ** animation** the word we're looking for here? I'd say:
> This is an experimental tracking API…
or 
> This is an experimental Jetpack Compose observability API…

xgouchet (changes requested):

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/ExperimentalTrackingApi.kt
0xnm: oops, copy-paste issue.

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
0xnm: I think I will stick with `click` terminology to be consistent with `onClick` and `onLongClick` arguments in Compose for now, and use `tap` terminology internally to be consistent with `RumActionType.TAP`.

xgouchet (approved):

dd-sdk-android-compose/README.md
```diff
@@ -79,6 +79,19 @@ val configuration = Configuration.Builder(
     .build()
 ```
 
+#### Action tracking
+
+There is no automated instrumentation support for the action tracking in Jetpack Compose yet. However, you may report clicks using `trackClicks` method in the following manner:
```
apigirl, line 84:
```suggestion
There is no automated instrumentation support for the action tracking in Jetpack Compose. However, report clicks using `trackClicks` method using this example:
```

apigirl (commented):
small docs team feedback

sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/ComposeFragment.kt
```diff
@@ -85,7 +91,16 @@ fun SimpleView(
         modifier = Modifier.fillMaxSize()
     ) {
         Text("View $viewId")
-        Button(onClick = onNavigate, modifier = Modifier.padding(top = 32.dp)) {
+        val viewTreeObserver = (LocalContext.current as Activity).window.decorView.viewTreeObserver
+        Button(
+            onClick = trackClicks(targetName = "Open View", onClick = {
```
xgouchet, line 96: Out of consistency with `onClick`, should we rename the method `trackClick` instead of `trackClicks`?

0xnm: sure, can do that

PR #761, opened 2021-11-16
RUMM-1774 Add the GlobalRum#checkForSessionUpdate method

mariusc83:
### What does this PR do?

In this PR we are introducing the `GlobalRum#notifyIngestedWebViewEvent` internal method. The purpose of this method is to trigger an asynchronous `EmptyEvent` at the `SessionScope` level which will check if the session needs to be updated or not. This should be triggered whenever a new RUM event is received through the JS bridge.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/GlobalRum.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/GlobalRum.kt
@@ -21,7 +21,6 @@ import com.datadog.android.tracing.internal.TracingFeature
 import java.util.concurrent.Callable
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicBoolean
-import java.util.concurrent.atomic.AtomicLong
 import java.util.concurrent.atomic.AtomicReference
 
 /**
@@ -37,8 +36,6 @@ import java.util.concurrent.atomic.AtomicReference
 object GlobalRum {
 
     internal val globalAttributes: MutableMap<String, Any?> = ConcurrentHashMap()
-
-    internal val sessionStartNs = AtomicLong(0L)
 
     internal val isRegistered = AtomicBoolean(false)
     internal var monitor: RumMonitor = NoOpRumMonitor()
@@ -142,6 +139,10 @@ object GlobalRum {
     }
 
     // region Internal
+
+    internal fun notifyIngestedWebViewEvent() {
+        (monitor as? AdvancedRumMonitor)?.sendWebViewEvent()
+    }
 
     internal fun getRumContext(): RumContext {
         return activeContext.get()
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/AdvancedRumMonitor.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/AdvancedRumMonitor.kt
@@ -13,10 +13,13 @@ import com.datadog.android.rum.internal.domain.event.ResourceTiming
 import com.datadog.android.rum.model.ViewEvent
 import com.datadog.tools.annotation.NoOpImplementation
 
+@SuppressWarnings("ComplexInterface")
 @NoOpImplementation
 internal interface AdvancedRumMonitor : RumMonitor {
 
     fun resetSession()
+
+    fun sendWebViewEvent()
 
     fun viewTreeChanged(eventTime: Time)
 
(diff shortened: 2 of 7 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -301,6 +305,8 @@ internal class DatadogRumMonitor(
     internal fun handleEvent(event: RumRawEvent) {
         if (event is RumRawEvent.AddError && event.isFatal) {
             rootScope.handleEvent(event, writer)
+        } else if (event is RumRawEvent.SynchronousEmptyEvent) {
```
mariusc83, line 308: so we could do this or we could just make this an `EmptyEvent` and let it asynchronous. I am just afraid that this will turn back to us in some cases. I need your input here to see which way to go.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/GlobalRum.kt
```diff
@@ -143,6 +140,10 @@ object GlobalRum {
 
     // region Internal
 
+    internal fun checkSessionForUpdate() {
```
xgouchet, line 143: This is not what this method does anymore, maybe we should rename this

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/AdvancedRumMonitor.kt
```diff
@@ -13,11 +13,14 @@ import com.datadog.android.rum.internal.domain.event.ResourceTiming
 import com.datadog.android.rum.model.ViewEvent
 import com.datadog.tools.annotation.NoOpImplementation
 
+@SuppressWarnings("ComplexInterface")
 @NoOpImplementation
 internal interface AdvancedRumMonitor : RumMonitor {
 
     fun resetSession()
 
+    fun sendEmptyEvent()
```
xgouchet, line 22: Instead of empty event, maybe we can add some meaning here, e.g.:`fun notifyWebViewRumEventIngested` ? (ok this one is probably a bit too verbose, maybe `fun addWebViewEvent()` ?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebRumEventContextProvider.kt
```diff
@@ -14,6 +14,7 @@ import com.datadog.android.rum.internal.domain.RumContext
 internal class WebRumEventContextProvider {
 
     fun getRumContext(): RumContext? {
```
xgouchet, line 16: Just thinking instead of dropping those events maybe we can still store them locally to be uploaded at a later time, that way events won't be lost

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebRumEventContextProvider.kt
mariusc83: Yes but the way the consumer works now is that if we do not have a valid rum context we are disabling the feature for good. Meaning that we will not allow other events anyway. This is mainly to avoid all the possible internal/dev logs as we discussed.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/GlobalRum.kt
mariusc83: Yeah...I was trying to find a proper name, maybe `notifyIngestedWebViewEvent`

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitorTest.kt
```diff
@@ -1097,6 +1097,20 @@ internal class DatadogRumMonitorTest {
         }
     }
 
+    @Test
+    fun `M delegate event to rootScope W sendWebViewEvent()`() {
```
0xnm, line 1101: should we assert that session is updated after that?

mariusc83: Oh...I think we are going to far, I think this is the `SessionScope` responsibility

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebRumEventContextProvider.kt
mariusc83: Also there is no sense the way I see it to keep the events for later as we are only dropping those in case there is no RumContext. If there is no RumContext that means you forgot to enable the RUM feature or you are not doing this in the `Application#onCreate`. In the first case it will require a process restart anyway and in the latter I would say it is a initialization problem on user's end.

xgouchet (approved):

PR #764, opened 2021-11-18
RUMM-1762 Read is_crash info from cross platform attributes

xgouchet:
### What does this PR do?

Read `error.is_crash` info from cross platform attributes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -259,6 +259,7 @@ internal open class RumViewScope(
         pendingResourceCount++
     }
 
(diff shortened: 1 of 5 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -283,7 +285,7 @@ internal open class RumViewScope(
                 message = message,
                 source = event.source.toSchemaSource(),
                 stack = event.stacktrace ?: event.throwable?.loggableStackTrace(),
-                isCrash = event.isFatal,
+                isCrash = event.isFatal || (isFatal),
```
mariusc83, line 288: why do you use brackets here ?

mariusc83 (approved):

0xnm (approved):

PR #766, opened 2021-11-19
RUMM-1727 Introduce the WebLogsEventConsumer

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebLogEventConsumer.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebLogEventConsumer.kt
@@ -6,13 +6,69 @@
 
 package com.datadog.android.rum.webview
 
+import com.datadog.android.core.internal.CoreFeature
+import com.datadog.android.core.internal.persistence.DataWriter
 import com.datadog.android.core.internal.utils.sdkLogger
+import com.datadog.android.log.LogAttributes
+import com.datadog.android.log.model.LogEvent
 import com.google.gson.JsonObject
+import com.google.gson.JsonParseException
 
-internal class WebLogEventConsumer {
+internal class WebLogEventConsumer(
+    private val userLogsWriter: DataWriter<LogEvent>,
+    private val internalLogsWriter: DataWriter<LogEvent>,
+    private val rumContextProvider: WebRumEventContextProvider
+) {
 
-    fun consume(event: JsonObject) {
-        // just to make the code compile without warning. This will be removed in the next PR.
-        sdkLogger.i(event.toString())
+    private val ddTags: String by lazy {
+        "${LogAttributes.APPLICATION_VERSION}:${CoreFeature.packageVersion}" +
+            ",${LogAttributes.ENV}:${CoreFeature.envName}"
+    }
+
+    fun consume(event: JsonObject, eventType: String) {
+        map(event)?.let {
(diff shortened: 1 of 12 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebLogEventConsumer.kt
```diff
@@ -6,13 +6,51 @@
...
+                userLogsWriter.write(it)
+            }
+        }
+    }
+
+    private fun map(event: JsonObject): LogEvent? {
+        return try {
+            val logEvent = LogEvent.fromJson(event.toString())
+            val rumContext = rumContextProvider.getRumContext()
+            if (rumContext != null) {
```
mariusc83, line 36: We can either do this or we will have to change our Log `DataWriter` to accept pure JsonObjects. I would rather do this than that and see how it goes. In case the deserialization fails for the reason of some missing attributes we could change all the attributes in the log event internal json-schema as optional.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebLogEventConsumer.kt
```diff
@@ -6,13 +6,69 @@
...
+        }
+    }
+
+    companion object {
+        const val DDTAGS_SEPARATOR = ","
+        const val DDTAGS_KEY_NAME = "ddtags"
+        const val USER_LOG_EVENT_TYPE = "log"
+        const val INTERNAL_LOG_EVENT_TYPE = "internal_log"
+        const val JSON_PARSING_ERROR_MESSAGE = "The bundled web log event could not be deserialized"
+        val logEventTypes = setOf(USER_LOG_EVENT_TYPE, INTERNAL_LOG_EVENT_TYPE)
```
0xnm, line 72:
```suggestion
        val LOG_EVENT_TYPES = setOf(USER_LOG_EVENT_TYPE, INTERNAL_LOG_EVENT_TYPE)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/webview/WebEventConsumerTest.kt
```diff
@@ -71,36 +71,79 @@ internal class WebEventConsumerTest {
...
+            fakeUnknownEventType = forge.anAlphabeticalString()
+        }
+        val fakeWebEvent = bundleWebEvent(fakeBundledEvent, fakeUnknownEventType)
 
         // When
         testedWebEventConsumer.consume(fakeWebEvent.toString())
 
         // Then
-        verify(mockLogsEventConsumer).consume(fakeBundledEvent)
+        verify(mockSdkLogHandler).handleLog(
```
0xnm, line 140: maybe it makes sense to merge this test with the test above? they have the same Given and When sections.

0xnm (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/webview/WebEventConsumerTest.kt
mariusc83: Yeah ...but one is testing the log and the other is testing that no message is sent. Following our current guidelines we kind of tend to keep these separated.

xgouchet (approved):

PR #770, opened 2021-11-23
Add a CurlInterceptor

xgouchet:
### What does this PR do?

When the SDK is in debug mode, use a CurlInterceptor to print intake upload requests as `curl` commands in the logcat. This comes in handy to debug requests with the backend.


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/CurlInterceptor.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    // endregion
+
+    companion object {
+
+        private const val FORMAT_HEADER = "-H \"%1\$s:%2\$s\""
+        private const val FORMAT_METHOD = "-X %1\$s"
+        private const val FORMAT_BODY = "-d '%1\$s'"
+        private const val FORMAT_URL = "\"%1\$s\""
```
0xnm, line 100: nit: I guess positional markers `1$` are not really needed here

0xnm (approved):

mariusc83 (approved):

PR #772, opened 2021-11-25
RUMM-1776: Add swipe and scroll manual instrumentation to jetpack compose

0xnm:
### What does this PR do?

This change brings support of manual instrumentation of Swipe and Scroll actions in Jetpack Compose.

It works by listening for the interactions happening with a help of [InteractionSource](https://developer.android.com/reference/kotlin/androidx/compose/foundation/interaction/InteractionSource) and then reading the current state of Swipe/Scroll when certain interaction happens.

Under the hood of Jetpack Compose both scroll and swipe rely on the generic `drag` operation, so code for their tracking have some similarities. However, there are also differences: 

* Direction: swiping down will **increase** offset from the base point, however scrolling down will **decrease** offset from the base point.
* Anchors: scroll doesn't have anchors, but swipe should have at least one anchor is `Modifier.swipeable` is used. Anchors are used to know the current state of the swipe (say in case of the `Switch`), and serve as sticky points  to come to, when finger is up.

Also 2 attributes are added `from_state` and `to_state` for the swipe interaction: they can be useful later in the scrubbing API is user wants to filter out swipe events which didn't result in the state change. They are not in the `action` namespace yet, but I think we should add them there.

Also the UI is added in the sample app to play with Swipe/Scroll actions reporting.

### DIFF
--- a/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
+++ b/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
@@ -275,9 +275,6 @@ internal suspend fun <T> trackDragInteraction(
     }
 }
 
-internal const val FROM_SWIPE_STATE_ATTRIBUTE = "from_state"
-internal const val TO_SWIPE_STATE_ATTRIBUTE = "to_state"
-
 // endregion
 
 // region private
@@ -323,12 +320,12 @@ private fun resolveSwipeChangeAttributes(
     )
 
     return mapOf(
-        FROM_SWIPE_STATE_ATTRIBUTE to swipeStartProps.anchorState,
+        RumAttributes.ACTION_GESTURE_FROM_STATE to swipeStartProps.anchorState,
         // https://issuetracker.google.com/issues/149549482
         // There is a Compose bug: if drag stopped (pointer up) and threshold for the next value is
         // not yet reached, but there is enough velocity to continue the fling, this will
         // still report current value, this affects reporting direction as well
-        TO_SWIPE_STATE_ATTRIBUTE to swipeableState.targetValue,
+        RumAttributes.ACTION_GESTURE_TO_STATE to swipeableState.targetValue,
         RumAttributes.ACTION_GESTURE_DIRECTION to direction
     )
 }
--- a/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
+++ b/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
@@ -29,6 +29,7 @@ import com.datadog.android.rum.RumActionType
 import com.datadog.android.rum.RumAttributes
 import com.datadog.android.rum.RumMonitor
 import java.lang.Exception
+import kotlin.coroutines.cancellation.CancellationException
 import kotlin.math.roundToInt
 import kotlinx.coroutines.flow.collect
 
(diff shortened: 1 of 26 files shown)


daniel-m-dd (approved):

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
```diff
@@ -55,3 +158,273 @@ internal class TapActionTracker(
...
+        return when (this) {
+            is LazyListState -> {
+                this.firstVisibleItemIndex
+            }
+            is ScrollState -> {
+                this.value
+            }
+            else -> {
+                null
+            }
```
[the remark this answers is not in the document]
0xnm: they are different indeed (index vs offset in pixels), but since they follow the same logic - when scrolling down both index and offset decrease, up - increase, they can be unified behind the common property `position`. `ScrollableState` interface itself doesn't expose any field which can tell about position unfortunately.

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
```diff
@@ -55,3 +158,273 @@ internal class TapActionTracker(
...
+    targetName: String,
+    interactionSource: InteractionSource,
+    interactionType: InteractionType.Swipe<*>,
+    isRtl: Boolean,
+    attributes: Map<String, Any?>
+) {
+    trackDragInteraction<SwipeStartProps>(
+        interactionSource,
+        onStart = { interactions, start ->
+            interactions[start] = SwipeStartProps(
```
mariusc83, line 178: Hmm I would rather put this in the `trackDragInteraction` function where you call this lambda than here ? And just keep that RUM call here, that's makes it more visible don't you think ?

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
```diff
@@ -55,3 +158,273 @@ internal class TapActionTracker(
...
+                startProps,
+                interactionType,
+                isRtl,
+                attributes
+            )
+        }
+    )
+}
+
+internal const val FROM_SWIPE_STATE_ATTRIBUTE = "from_state"
```
mariusc83, line 230: Do we really need to report these values ?

mariusc83 (commented):
Some comments but for me so far looks like a very good job

xgouchet (approved):

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
0xnm: Start properties objects are different for swipe and scroll interaction, so we have to keep it here.

At the beginning I was trying to do indeed everything in the single `trackInteraction` function, but that led to the many `is` checks and it wasn't easy to follow logically. So I decided to split it and extract common type-independent logic.

dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
0xnm: Good question. I'm still thinking about it. Initially my motivation was to give an opportunity to filter out swipe interactions which didn't change the state (if user wants) in the scrubbing API, especially considering we removed side-effect requirement.

On another hand https://issuetracker.google.com/issues/207695812 is resolved as `Intended behavior`, so we won't know correct target state at the end of interaction (although it may be an edge case - very fast and very strong swipe).

mariusc83: I would say keep them for now then.

mariusc83 (approved):

xgouchet (approved):

apigirl (approved):
👍 for docs

xgouchet (approved):

Issue #760, opened 2021-11-16
Make it possible to use `NavigationViewTrackingStrategy` for dynamic `NavHostFragment`


tomaz:
**Are you requesting automatic instrumentation for a framework or library? Please describe.**
/

**Is your feature request related to a problem? Please describe.**
`NavigationViewTrackingStrategy` only attempts to find nav controller in `onActivityStarted`, which is called before dynamically created and inserted `NavHostFragment.onCreate` (in our app, `onCreate` is called from within parent activity `onStart`, after activity lifecycle `onActivityStarted`). Since `NavController` is created in `Fragment.onCreate`, it's still null at this point. Because `NavigationViewTrackingStrategy` doesn't allow any other means of injecting `NavController`, we can't use it.

**Describe the solution you'd like**
Ideally, `NavigationViewTrackingStrategy` should have alternate method of injecting `NavController`. For example as optional public property or a setup function:

```
class NavigationViewTrackingStrategy(
    ...
) ... {

    private var navController: NavController? = null
    ...

    // region ActivityLifecycleTrackingStrategy

    override fun onActivityStarted(activity: Activity) {
        super.onActivityStarted(activity)
        setup(activity.findNavControllerOrNull(navigationViewId))
    }

    ...

    fun setup(controller: NavController?) {
        navController?.removeOnDestinationChangedListener(this)

        navController = controller

        navController?.let {
            if (FragmentActivity::class.java.isAssignableFrom(activity::class.java)) {
                val navControllerFragmentCallbacks = NavControllerFragmentLifecycleCallbacks(
                    it,
                    argumentsProvider = { emptyMap() },
                    componentPredicate = predicate
                )
                navControllerFragmentCallbacks.register(activity as FragmentActivity)
                lifecycleCallbackRefs[activity] = navControllerFragmentCallbacks
            }
            it.addOnDestinationChangedListener(this)
        }
    }
```

**Describe alternatives you've considered**
We're implementing our own tracking strategy, which is a shame since `NavigationViewTrackingStrategy` provides so much out of the box...

**Additional context**
/

xgouchet:
Hi @tomaz, that's indeed a use case we haven't considered yet. Since you're going to write your own Strategy, feel free to submit it as an `LazyNavigationViewTrackingStrategy` in a PR. That we we can enhance it whenever we add new features to the SDK, and it could be used by other customers as well.

tomaz:
[...]

PR #773, opened 2021-11-26
Improving `NavigationViewTrackingStrategy` to allow dynamic `NavHostFrament` setup [#760]

tomaz:
### What does this PR do?

As discussed in #760, but without separate `LazyNavigationViewTrackingStrategy` - there's no need for a separate class, it will just confuse, especially since it would either need to re-implement all of the functionality of the base class, or need to be derived from it. Original implementation still works exactly the same as before, but exposes tracking functionality through public functions `startTracking` and `stopTracking`. This way, it's possible to use the strategy with programmatically created `NavHostFragment`.

In most cases, only a call to `startTracking` is needed, `stopTracking` is called automatically when activity stops. But we do call `stopTracking` in our custom implementation as well because we replace `NavHostFragment` multiple times during activity lifetime. Having it ensures there's no tracking left in case new nav host is not needed. Since `startedActivity` property is managed inside activity lifecycle callbacks, the two tracking functions can be called independently multiple times without an issue.

### Motivation

`NavigationViewTrackingStrategy` is only usable for static nav host controller without these changes so doesn't work for our project.

### Additional Notes

I implemented the changes based on the custom implementation suited for our project. I can't easily test local implementation within our project, too many changes required, but I tested it with sample project and works the same. Also ran `intrumented.integration` which opened an app but not much was happening other than that. And `instrumented.nightly-tests` which again built fine, but refused to run due to error `Could not identify launch activity: Default Activity not found`. I get the same results without my changes too though.

I only committed `NavigationViewTrackingStrategy`, but this change resulted in showing multiple other files as changed. `apiSurface` is the only one that actually shows changes (2 new functions added). I assume these are auto-generated, so don't include them here.

### DIFF
--- a/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategyTest.kt
+++ b/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategyTest.kt
@@ -155,6 +155,17 @@ internal class NavigationViewTrackingStrategyTest {
     }
 
     @Test
+    fun `registers navigation listener when nav controller added after activity started`() {
+        whenever(mockNavView.getTag(R.id.nav_controller_view_tag)) doReturn null
+        testedStrategy.onActivityStarted(mockActivity)
+        verifyZeroInteractions(mockNavController)
+
+        whenever(mockNavView.getTag(R.id.nav_controller_view_tag)) doReturn mockNavController
+        testedStrategy.startTracking()
+        verify(mockNavController).addOnDestinationChangedListener(testedStrategy)
+    }
+
+    @Test
     fun `registers fragment lifecycle callback onActivityStarted`() {
         testedStrategy.onActivityStarted(mockActivity)
 
@@ -165,10 +176,18 @@ internal class NavigationViewTrackingStrategyTest {
     }
 
     @Test
-    fun `unregisters navigation listener onActivityStopped`() {
+    fun `unregisters navigation listener onActivityStopped if activity was started`() {
+        testedStrategy.onActivityStarted(mockActivity)
         testedStrategy.onActivityStopped(mockActivity)
 
         verify(mockNavController).removeOnDestinationChangedListener(testedStrategy)
+    }
+
+    @Test
+    fun `doesn't unregister navigation listener onActivityStopped if activity not started`() {
+        testedStrategy.onActivityStopped(mockActivity)
+
+        verifyZeroInteractions(mockNavController)
     }
 
     @Test
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategy.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategy.kt
@@ -42,6 +42,8 @@ class NavigationViewTrackingStrategy(
     ViewTrackingStrategy,
     NavController.OnDestinationChangedListener {
 
+    private var startedActivity: Activity? = null
+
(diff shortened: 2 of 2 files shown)


xgouchet (changes requested):
Thanks a lot for this @tomaz, this looks quite good overall.
Can you add a test to validate that the target usecase works: 

```kotlin
    @Test
    fun `registers navigation listener when nav controller added after activity started`() {
        // TODO reset mockActivity
        testedStrategy.onActivityStarted(mockActivity)
        verifyZeroInteraction(mockNavController

        // TODO setup mockActivity to have a mockNavController
        testedStrategy. startTracking()
        verify(mockNavController).addOnDestinationChangedListener(testedStrategy)
    }
```

tomaz:
Added the test - copied it from above and fixed 2 typos, so it compiles.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategyTest.kt
```diff
@@ -154,6 +154,17 @@ internal class NavigationViewTrackingStrategyTest {
         verify(mockNavController).addOnDestinationChangedListener(testedStrategy)
     }
 
+    @Test
+    fun `registers navigation listener when nav controller added after activity started`() {
+        // TODO reset mockActivity
```
xgouchet, line 159: Those were to be completed, I'm sorry I might have not been clear enough. As is I'm not sure the unit test passes. 

Reset would be something like: 
```
whenever(mockNavView.getTag(R.id.nav_controller_view_tag)) doReturn null
```

and setup below would be 
```
whenever(mockNavView.getTag(R.id.nav_controller_view_tag)) doReturn mockNavController
```

Also make sure that all other tests in this class pass as well.

xgouchet (changes requested):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategyTest.kt
tomaz: Ok, I'll see what I can do - as said, unit tests refused to work on my machine (both `instrumented.integration` and `instrumented.nightly-tests`)

xgouchet: Those are just unit tests, you should be able to run them with `gradlew uTD` or `gradlew :dd-sdk-android:testDebugUnitTest`

tomaz: Allright, updated unit tests as well (sorry, took me a while, can't take my full time on this).

With new code `unregisters navigation listener onActivityStopped` started failing because now `startedActivity` needs to be assigned, so:

- added unit test for dynamic setup (as discussed above)
- split `unregisters navigation listener onActivityStopped` into 2 unit tests:
- `unregisters navigation listener onActivityStopped if activity was started` verifies the same thing as original "unregister" test, except that it calls `onActivityStarted` as well
- `doesn't unregister navigation listener onActivityStopped if activity not started` takes care of use case where only `onActivityStopped` is called and hence no unregistration occurs

Technically, in real world usage, last case should not happen - `onActivityStopped` is only called after prior `onActivityStarted`. But the code is ready for this so I thought unit test should reflect it too.

Note: there are couple flaky unit tests (3-5) that fail on any test run for me. But they're the same as before my changes.

Let me know your further thoughts...

xgouchet (approved):

xgouchet:
Thanks a lot for your help @tomaz

PR #781, opened 2021-12-03
Add a flag that allows increased info when debugging

fuzzybinary:
The `verboseDebugInfo` flag will override any other user settings to increase upload frequency, disable sampling, and set logging to verbose, but only when the application is marked as `debuggable` in the manifest.

Implements RUMM-1773 

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -9,8 +9,11 @@ package com.datadog.android
 import android.app.Application
 import android.content.Context
 import android.content.pm.ApplicationInfo
+import android.util.Log
+import com.datadog.android.core.configuration.BatchSize
 import com.datadog.android.core.configuration.Configuration
 import com.datadog.android.core.configuration.Credentials
+import com.datadog.android.core.configuration.UploadFrequency
 import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleCallback
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleMonitor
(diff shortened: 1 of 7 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -151,6 +152,21 @@ internal constructor(
...
+         * Sets the DataDog SDK to be more verbose when an application is set to `debuggable`.
+         * This is equivalent to setting:
+         *   sampleRumSessions(100)
+         *   setBatchSize(BatchSize.SMALL)
+         *   setUploadFrequency(UploadFrequency.FREQUENT)
+         *   Datadog.setVerbosity(Log.VERBOSE)
+         * These settings will override your configuration, but only when the application is `debuggable`
+         * @param verboseDebugInfo Enable or disable verbose info when debugging
+         */
+        fun setVerboseDebugInfo(verboseDebugInfo: Boolean): Builder {
```
xgouchet, line 165: I think we should come up with another name. `setVerboseDebugInfo` only hints that it will do `Datadog.setVerbosity(Log.VERBOSE)`, but we're making more changes. We could make this a bit more generic with a name like `enableDeveloperModeWhenDebuggable` or something in the same sense.

dd-sdk-android/src/test/kotlin/com/datadog/android/DatadogTest.kt
```diff
@@ -106,6 +108,8 @@ internal class DatadogTest {
     @AfterEach
     fun `tear down`() {
         Datadog.isDebug = false
+        Datadog.setVerbosity(Int.MAX_VALUE)
```
xgouchet, line 111: Good catch

dd-sdk-android/src/test/kotlin/com/datadog/android/DatadogTest.kt
```diff
@@ -227,6 +231,51 @@ internal class DatadogTest {
         verify(CoreFeature.trackingConsentProvider).setConsent(fakeConsent)
     }
 
+    @Test
+    fun `M no changes W initialize() { verboseDebug } in non debug mode`() {
+        val credentials = Credentials(fakeToken, fakeEnvName, fakeVariant, fakeApplicationId, null)
```
xgouchet, line 236: Out of consistency, can you also make sure the `ApplicationInfo.FLAG_DEBUGGABLE` is not present in `appContext.fakeAppInfo.flags`. Even though it's initialized by default to `0`, it'd be safer to ensure it's in the state you expect.

dd-sdk-android/src/test/kotlin/com/datadog/android/DatadogTest.kt
```diff
@@ -227,6 +231,51 @@ internal class DatadogTest {
         verify(CoreFeature.trackingConsentProvider).setConsent(fakeConsent)
     }
 
+    @Test
+    fun `M no changes W initialize() { verboseDebug } in non debug mode`() {
```
xgouchet, line 235: Can you add `// Given`, `// When` and `// Then` comments to split your tests in 3 parts?

dd-sdk-android/src/test/kotlin/com/datadog/android/core/configuration/ConfigurationBuilderTest.kt
```diff
@@ -102,6 +102,7 @@ internal class ConfigurationBuilderTest {
         assertThat(config.coreConfig).isEqualTo(
             Configuration.Core(
                 needsClearTextHttp = false,
+                verboseDebugInfo = false,
```
xgouchet, line 105: Can you add a test ensuring the effects of the new `setVerboseDebugInfo()` method here

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -76,16 +79,30 @@ object Datadog {
             return
         }
 
+        var mutableConfig = configuration
+        if (isDebug and configuration.coreConfig.verboseDebugInfo) {
+            mutableConfig = configuration.copy(
+                coreConfig = configuration.coreConfig.copy(
+                    batchSize = BatchSize.SMALL,
+                    uploadFrequency = UploadFrequency.FREQUENT
```
mariusc83, line 87: Can you extract this in a separated method please ?

dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
fuzzybinary: Yeah, that's a much better name.

mariusc83 (approved):

0xnm (commented):
This PR lowers down API surface test coverage. As a quick solution we can maybe change the threshold from `0.94` to `0.93` for now, because we don't have an API to ignore certain methods, only classes for now.

xgouchet (approved):

fuzzybinary:
@0xnm Any ideas why this might have lowered coverage? I don't see anything new that's not tested but maybe I'm just missing something.

PR #790, opened 2021-12-17
RUMM-1833: Stop current action if another view starts

0xnm:
### What does this PR do?

This change brings the same logic in `ActionScope` when `StartView` event is received as for the `StopView`, because action lives in the boundaries of the parent view scope, and if another view starts without calling `stopView` for the previous view, we need to release current action, to be able to mark current view as complete.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScope.kt
@@ -68,6 +68,7 @@ internal class RumActionScope(
             shouldStop -> sendAction(lastInteractionNanos, writer)
             isLongDuration -> sendAction(now, writer)
             event is RumRawEvent.SendCustomActionNow -> sendAction(lastInteractionNanos, writer)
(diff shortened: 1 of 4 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScope.kt
```diff
@@ -88,6 +89,15 @@ internal class RumActionScope(
 
     // region Internal
 
+    private fun onStartView(
+        now: Long,
+        writer: DataWriter<Any>
+    ) {
+        // another view starts, complete this action
+        ongoingResourceKeys.clear()
```
mariusc83, line 97: this means that an ongoing resource which was start by this action will be associated with the new action right ? Are we ok with this ?

0xnm: Yes, it will be. Not sure if it is the best logic, but it is basically exactly the same as in the `RumActionScope#onStopView`.

xgouchet (approved):

mariusc83 (approved):

PR #791, opened 2021-12-20
RUMM-1807 Handle App launch events

xgouchet:
### What does this PR do?

Distinguish view-less event's happening early during the app's lifetime as `AppLaunch` instead of just `Background`

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
@@ -6,10 +6,12 @@
 
 package com.datadog.android.rum.internal.domain.scope
 
+import android.app.ActivityManager.RunningAppProcessInfo
 import android.os.Build
 import android.os.Process
 import android.os.SystemClock
 import com.datadog.android.Datadog
+import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.net.FirstPartyHostDetector
 import com.datadog.android.core.internal.persistence.DataWriter
 import com.datadog.android.core.internal.persistence.NoOpDataWriter
@@ -117,7 +119,9 @@ internal class RumSessionScope(
         event: RumRawEvent,
         writer: DataWriter<Any>
     ) {
-        if (applicationDisplayed) {
+        val isForegroundProcess =
+            CoreFeature.processImportance == RunningAppProcessInfo.IMPORTANCE_FOREGROUND
+        if (applicationDisplayed || !isForegroundProcess) {
             handleBackgroundEvent(event, writer)
         } else {
             handleAppLaunchEvent(event, writer)
@@ -194,11 +198,15 @@ internal class RumSessionScope(
     ) {
         if (!applicationDisplayed) {
             applicationDisplayed = true
-            val applicationStartTime = resolveStartupTimeNs()
-            viewScope.handleEvent(
-                RumRawEvent.ApplicationStarted(event.eventTime, applicationStartTime),
-                writer
-            )
+            val isForegroundProcess =
+                CoreFeature.processImportance == RunningAppProcessInfo.IMPORTANCE_FOREGROUND
+            if (isForegroundProcess) {
+                val applicationStartTime = resolveStartupTimeNs()
+                viewScope.handleEvent(
+                    RumRawEvent.ApplicationStarted(event.eventTime, applicationStartTime),
+                    writer
+                )
+            }
         }
     }
 
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
@@ -48,7 +48,7 @@ internal class RumSessionScope(
 
     private var resetSessionTime: Long? = null
 
-    private var applicationDisplayed: Boolean = false
+    internal var applicationDisplayed: Boolean = false
 
     private val random = SecureRandom()
     private val noOpWriter = NoOpDataWriter<Any>()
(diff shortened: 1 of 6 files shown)


dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -741,8 +823,60 @@ internal class RumSessionScopeTest {
...
+                RumRawEvent.AddError(
+                    fakeName,
+                    this.aValueFrom(RumErrorSource::class.java),
+                    stacktrace = null,
+                    throwable = null,
+                    isFatal = this.aBool(),
+                    attributes = emptyMap(),
+                    eventTime = fakeEventTime
+                ),
+                RumRawEvent.StartAction(
```
0xnm, line 865: it should be `StartResource` here

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -741,8 +823,60 @@ internal class RumSessionScopeTest {
...
+        )
+    }
+
+    private fun Forge.forgeValidAppLaunchEvent(
+        eventTime: Time? = null
+    ): RumRawEvent {
+        val fakeEventTime = eventTime ?: Time()
         val fakeName = this.anAlphabeticalString()
 
         return this.anElementFrom(
```
0xnm, line 882: I cannot put a comment on the valid line (github doesn't let me put it outside of the diff context), but this list contains `StartAction` twice, it should be a `StartResource` in the latter case.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/widget/WidgetIntentService.kt
```diff
@@ -31,19 +31,17 @@ class WidgetIntentService : IntentService("WidgetIntentService") {
                 val hasRumContext = widgetId != 0 && widgetName != null
 
                 if (hasRumContext) {
-                    GlobalRum.get().startView(widgetId, widgetName ?: "DatadogWidget", emptyMap())
                     val clickedTargetName = intent.getStringExtra(WIDGET_CLICKED_TARGET_NAME)
                     if (clickedTargetName != null) {
-                        GlobalRum.get()
-                            .addUserAction(RumActionType.CLICK, clickedTargetName, emptyMap())
+                        GlobalRum.get().addUserAction(
```
0xnm, line 36: So this will be reported with a background scope?

0xnm (changes requested):
Nice work! I just added few questions to better understand the change and there is a copy-paste issue.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -198,9 +243,30 @@ internal class RumSessionScope(
...
+            RumRawEvent.AddError::class.java,
+            RumRawEvent.AddLongTask::class.java,
+            RumRawEvent.StartAction::class.java,
+            RumRawEvent.StartResource::class.java
+        )
+
         internal val DEFAULT_SESSION_INACTIVITY_NS = TimeUnit.MINUTES.toNanos(15)
         internal val DEFAULT_SESSION_MAX_DURATION_NS = TimeUnit.HOURS.toNanos(4)
 
+        internal val APPLICATION_START_MAX_DURATION_NS = TimeUnit.SECONDS.toNanos(15)
```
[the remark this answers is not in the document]
xgouchet: Based on some research on our customers (aggregating ~50M app start events over 1 months), it seems that the "normal" distribution is around this mark yes. Just to be safe, I'll increase this to 30s for now, and we can lower it if it makes more sense or when we change the `Application start` definition

sample/kotlin/src/main/kotlin/com/datadog/android/sample/widget/WidgetIntentService.kt
xgouchet: This was a test I made, I'll revert that

0xnm (changes requested):
There is still an issue with duplicating `StartAction` in `forgeValidBackgroundEvent` and `forgeValidAppLaunchEvent` methods of `RumSessionScopeTest`. It should be `StartResource` for the 2nd occurrence.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -112,33 +112,57 @@ internal class RumSessionScope(
...
     private fun handleOrphanEvent(
+        event: RumRawEvent,
+        writer: DataWriter<Any>
+    ) {
+        val durationNs = durationSinceLaunchNs(event)
+        if (applicationDisplayed || durationNs > APPLICATION_START_MAX_DURATION_NS) {
+            handleBackgroundEvent(event, writer)
+        } else {
+            handleAppLaunchEvent(event, writer)
+        }
```
ncreated, line 125: With no other condition, doesn't it mean we will be starting "ApplicationLaunch" view for eventual app background wake-ups? The reason why we introduced [`trackBackgroundRumEvents()`](https://github.com/DataDog/dd-sdk-android/blob/5ad87041b5e094e5d3d8d17cbcf41f67399007ad/dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt#L453-L454) API was to avoid creating separate RUM sessions for each background wake-up and impacting user billing. Won't be regression?

xgouchet: Indeed it might be a slight regression. I'll try using the `RunningAppProcessInfo` trick (it seems to be working fine)

0xnm (approved):

mariusc83 (approved):

PR #795, opened 2021-12-28
RUMM-1158 improve test performances

xgouchet:
### What does this PR do?

Improve time spent during tests in CI (and other improvements)


build.gradle.kts
```diff
@@ -267,3 +247,19 @@ tasks.register("buildNdkIntegrationTestsArtifacts") {
...
+        "integration",
+        "nightly-tests",
+        "kotlin",
+        "detekt",
+        "javabackport",
+        "noopfactory",
+        "unit"
+    )
+    instrumentAndroidPackage = false
+}
```
0xnm, line 265: minor: missing newline

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/FileExt.kt
```diff
@@ -38,7 +37,7 @@ private fun <T> File.safeCall(
         sdkLogger.e("Security exception was thrown for file ${this.path}", e)
         default
     } catch (e: Exception) {
-        sdkLogger.e("Security exception was thrown for file ${this.path}", e)
+        sdkLogger.e("Unexpected found exception was thrown for file ${this.path}", e)
```
0xnm, line 40:
```suggestion
        sdkLogger.e("Unexpected exception was thrown for file ${this.path}", e)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/persistence/file/FileExtTest.kt
```diff
@@ -0,0 +1,525 @@
...
+        // When
+        val readLines = mockFile.readLinesSafe()
+
+        // Then
+        assertThat(readLines).isNull()
+    }
+
+    @Test
+    fun `M catch exception W readLinesSafe() {FileNotFoundException}`() {
+        // When
```
0xnm, line 519: should we mock behavior with `FileNotFoundException` here?

mariusc83 (approved):
LGTM

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/anr/ANRDetectorRunnableTest.kt
```diff
@@ -48,7 +48,7 @@ import org.mockito.quality.Strictness
     ExtendWith(TestConfigurationExtension::class)
 )
 @MockitoSettings(strictness = Strictness.LENIENT)
-@ForgeConfiguration(Configurator::class)
+@ForgeConfiguration(value = Configurator::class, seed = 0x8d65a9689a4c1L)
```
0xnm, line 51: should be reverted

.gitlab-ci.yml
```diff
@@ -150,6 +148,22 @@ test:tools:
...
+  stage: test
+  timeout: 1h
+  script:
+    - pip3 install datadog
+    - git fetch --depth=1 origin master
+    - rm -rf ~/.gradle/daemon/
+    - export DD_API_KEY=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.api_key --with-decryption --query "Parameter.Value" --out text)
+    - export DD_APP_KEY=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.app_key --with-decryption --query "Parameter.Value" --out text)
+    - CODECOV_TOKEN=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.codecov-token  --with-decryption --query "Parameter.Value" --out text)
+    - GRADLE_OPTS="-Xmx2560m" ./gradlew :koverReportAll --stacktrace --no-daemon
```
0xnm, line 163: Does `kover` task trigger a compilation?

Maybe to speed-up overall build pipeline we can use Gitlab build cache https://docs.gitlab.com/ee/ci/caching/ (we can use it between `analysis` and `test` groups)? wdyt?

build.gradle.kts
```diff
@@ -267,3 +247,19 @@ tasks.register("buildNdkIntegrationTestsArtifacts") {
 }
 
 nightlyTestsCoverageConfig(threshold = 0.9f)
+kover {
+    isDisabled = false
+    disabledProjects = setOf(
+        "instrumented",
+        "sample",
+        "tools",
```
0xnm, line 255: Will it work for `tools:unit` and `tools:detekt`? These projects are disabled here, but mentioned in `koverReportAll `task`.

build.gradle.kts
```diff
@@ -267,3 +247,19 @@ tasks.register("buildNdkIntegrationTestsArtifacts") {
...
+        "integration",
+        "nightly-tests",
+        "kotlin",
+        "detekt",
```
0xnm, line 259: Kover `disabledProjects` doesn't support full path to the project? Like `tools:detekt`.

0xnm (changes requested):
Looks good overall! I just left some comment/thoughts.

Also one thing not clear to me is why coverage dropped so much, like many files which are not touched in the PR now report much lower coverage. Is it linked to how `Kover` works?

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/persistence/file/FileExtTest.kt
xgouchet: The fakeFile by definition doesn't exist, so we're guaranteed to trigger a FNFE

.gitlab-ci.yml
xgouchet: We could use the build cache, indeed I'll create a task to set this up

xgouchet: Yes Kover triggers a full Debug+Release compilation

build.gradle.kts
xgouchet: Indeed they should be removed from the `koverReportAll` tasks

build.gradle.kts
xgouchet: No unfortunately not.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/persistence/file/FileExtTest.kt
0xnm: It seems it won't be triggered, because in the `readLinesSafe` call `existsSafe` will return `false`, so it won't come to the point when `readLines` is executed (and I guess we want to test FNFE exception thrown by this call).

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventSerializerTest.kt
```diff
@@ -80,13 +79,28 @@ internal class RumEventSerializerTest {
...
-                containsAttributes(it.additionalProperties)
+                hasNullableField("id", usr.id)
+                hasNullableField("name", usr.name)
+                hasNullableField("email", usr.email)
+                containsAttributes(usr.additionalProperties)
+            }
+        }
+        event.connectivity?.let { connectivity ->
+            assertThat(jsonObject).hasField("connectivity") {
+                hasNullableField("status", connectivity.status.name.lowercase())
```
0xnm, line 96: `lowercase` without a `Locale` doesn't raise a warning?

tools/unit/src/main/kotlin/com/datadog/tools/unit/assertj/JsonElementAssert.kt
```diff
@@ -0,0 +1,57 @@
...
+            is Boolean -> assertThat(actual.asBoolean).isEqualTo(expected)
+            is Int -> assertThat(actual.asInt).isEqualTo(expected)
+            is Long -> assertThat(actual.asLong).isEqualTo(expected)
+            is Float -> assertThat(actual.asFloat).isEqualTo(expected)
+            is Double -> assertThat(actual.asDouble).isEqualTo(expected)
+            is String -> assertThat(actual.asString).isEqualTo(expected)
+            is Date -> assertThat(actual.asString).isEqualTo(expected)
+            is BigInteger -> assertThat(actual.asBigInteger).isEqualTo(expected)
+            is BigDecimal -> assertThat(actual.asBigDecimal).isEqualTo(expected)
+            is List<*> -> JsonArrayAssert(actual.asJsonArray).isJsonArrayOf(expected)
```
0xnm, line 38: should we support `Map` here as well?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventSerializerTest.kt
xgouchet: no it doesn't but I'll add that indeed

tools/unit/src/main/kotlin/com/datadog/tools/unit/assertj/JsonElementAssert.kt
xgouchet: For now I don't think we need to as it's unused, we can still add it later if the use case presents itself

0xnm (approved):

PR #796, opened 2021-12-29
RUMM-1228: Reduce flakiness in tests coverage

0xnm:
### What does this PR do?

This change reduces flakiness in test coverage by making some tests more predictable in terms of the branches they cover.

It is still possible to have some flakiness though, but it will be less often and will have less impact on the overall test coverage for PRs.

### DIFF
--- a/dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProviderTest.kt
+++ b/dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProviderTest.kt
@@ -30,6 +30,7 @@ import fr.xgouchet.elmyr.junit5.ForgeExtension
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.RepeatedTest
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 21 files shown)


xgouchet (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProviderTest.kt
```diff
@@ -247,7 +248,7 @@ internal class BroadcastReceiverNetworkInfoProviderTest {
             .hasCellularTechnology(mobileSubtypeNames[subtype])
     }
 
-    @Test
+    @RepeatedTest(4)
```
mariusc83, line 251: why repeat it 4 times ? Is this number following some sort of best practices ?

mariusc83 (changes requested):

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProviderTest.kt
0xnm: no, just some reasonable number to cover more branches for the possible test inputs (`forge.anElementFrom`), so that we have more consistency between test suite runs in terms of the branch checked

mariusc83 (approved):

PR #802, opened 2022-01-03
RUMM-1877 Send dummy RUM WebViewEvent only for RUM JS ingested events

mariusc83:
### What does this PR do?

We were currently sending the dummy `WebViewEvent` to refresh the session in the `RumSessionScope` for all ingested JS events resulting from WebView tracking including the `log` events. This PR corrects this behaviour by sending the event only for RUM types ingested events.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/test/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumerTest.kt
+++ b/dd-sdk-android/src/test/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumerTest.kt
@@ -10,6 +10,8 @@ import android.util.Log
 import com.datadog.android.core.internal.persistence.DataWriter
 import com.datadog.android.core.internal.time.TimeProvider
 import com.datadog.android.log.internal.logger.LogHandler
+import com.datadog.android.rum.GlobalRum
+import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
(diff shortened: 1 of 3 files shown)


xgouchet (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumerTest.kt
```diff
@@ -131,6 +133,29 @@ internal class WebViewRumEventConsumerTest {
 
     // region View
 
+    @Test
+    fun `M send a noop WebViewEvent W consume { View event }`(forge: Forge) {
+        // When
```
0xnm, line 138: minor: this should be probably `Given` and call itself only behind `When`, same for new tests below

0xnm (approved):
All good! I left a minor comment regarding test sections naming, but it is not blocking.

dd-sdk-android/src/test/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumerTest.kt
mariusc83: Yes...I forgot to change those as in the first place I wanted to go with parameterized tests there but realized in the end that will not work :)

PR #803, opened 2022-01-04
RUMM-1873: Improve build pipeline speed

0xnm:
### What does this PR do?

This change improves build pipeline speed by doing the following:

* Enables [Gitlab cache](https://docs.gitlab.com/ee/ci/caching/) between `analysis` and `test` stages.
* Makes `test:tools` to build only `US1` variant, because we don't have different source code sets for the flavors, so this will be enough.

#### Notes on the caching strategy

Cache is **only produced** by `analysis:detekt` job (as it covers most of task graph among executions in the `analysis` stage), and **only consumed** by `test` group.

Cache is not consumed by the `publish` stage, because this stage is executed very rarely.

Cache is valid only inside the certain pipeline (because it has pipeline ID as a key, check [Gitlab predefined variables](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html)), so we will monitor `test` stage execution trend to see if there is any gain (few executions is not enough to make a decision), so after some time we may:

* keep everything as it is
* expand cache to be valid, say, for the branch, not only for pipeline. This theoretically may make building for subsequent pushes faster (if we allow `analysis` stage also to consume cache)
* remove cache if there is no performance gain. This is also possible, because to use the cache it should be uploaded for S3 and then downloaded in another job and unpacked, it takes time.

What is cached: simply [Gradle build cache](https://docs.gradle.org/current/userguide/build_cache.html) is cached for now, which includes dependencies and compilation units.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -77,11 +77,10 @@ analysis:detekt:
   stage: analysis
   timeout: 30m
   cache:
-    key: $CI_PIPELINE_ID
+    key: $CI_COMMIT_REF_SLUG
     paths:
       - cache/caches/
       - cache/notifications/
-    policy: push
   script:
     - git fetch --depth=1 origin master
     - GRADLE_OPTS="-Xmx2560m" ./gradlew :detektAll --stacktrace --no-daemon --build-cache --gradle-user-home cache/
@@ -130,7 +129,7 @@ test:debug:
   stage: test
   timeout: 1h
   cache:
-    key: $CI_PIPELINE_ID
+    key: $CI_COMMIT_REF_SLUG
     paths:
       - cache/caches/
       - cache/notifications/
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -123,18 +129,30 @@ test:debug:
...
-    - GRADLE_OPTS="-Xmx2560m" ./gradlew :unitTestDebug --stacktrace --no-daemon
+    - GRADLE_OPTS="-Xmx2560m" ./gradlew :unitTestDebug --stacktrace --no-daemon --build-cache --gradle-user-home cache/
     - bash <(cat ./codecov.sh) -t $CODECOV_TOKEN
 
 test:release:
   tags: [ "runner:main" ]
   image: $CI_IMAGE_DOCKER
   stage: test
   timeout: 1h
+  cache:
```
mariusc83, line 150: shouldn't you specify to use the cache here : `./gradlew :unitTestRelease --build-cache` or I did not get how the caching works ?

0xnm: good catch! indeed I forgot to do it.

mariusc83 (approved):

xgouchet (approved):

PR #807, opened 2022-01-11
RUMM-1889 Strip query params from traced resource Span

mariusc83:
### What does this PR do?

APM limits is based on unique resource (i.e. url) . When url are reported with query params, the number of unique resource grows exponentially and make our customers reach the limit early.
As a fix we are removing the `queryParams` block from the resource URL when resolve the `Span#resourceName` value.

### Motivation

### Additional Notes


0xnm (approved):
Looks good!

I would also propose to mention about APM limits here https://docs.datadoghq.com/real_user_monitoring/connect_rum_and_traces?tab=androidrum, because customer may hit the same issue in case of some REST API, where URL may include ID or any other identifier, say `GET https://foo.bar/users/{userId}`, in this case it would be nice to document the necessary use of Span event mapper.

dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/TracingInterceptor.kt
```diff
@@ -246,7 +246,7 @@ internal constructor(
             .asChildOf(parentContext)
             .start()
 
-        (span as? MutableSpan)?.resourceName = url
+        (span as? MutableSpan)?.resourceName = url.substringBefore(URL_QUERY_PARAMS_BLOCK_SEPARATOR)
```
mariusc83, line 249: @xgouchet should we also do something about automatically created spans ? Those are automatically created on the backend so I am thinking that they could apply this fix there right ?

mariusc83:
> Looks good!
> 
> I would also propose to mention about APM limits here https://docs.datadoghq.com/real_user_monitoring/connect_rum_and_traces?tab=androidrum, because customer may hit the same issue in case of some REST API, where URL may include ID or any other identifier, say `GET https://foo.bar/users/{userId}`, in this case it would be nice to document the necessary use of Span event mapper.

@0xnm Yeah, that is a good suggestion actually. I will open a task for it.

dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/TracingInterceptor.kt
xgouchet: The ones created in the backend already follow the rule

dd-sdk-android/src/test/kotlin/com/datadog/android/tracing/TracingInterceptorTest.kt
```diff
@@ -362,13 +362,36 @@ internal open class TracingInterceptorTest {
         val response = testedInterceptor.intercept(mockChain)
 
         verify(mockSpanBuilder).withOrigin(getExpectedOrigin())
+        verify(mockSpan as MutableSpan).resourceName = fakeUrl.substringBefore("?")
```
xgouchet, line 365: Here you're duplicating the logic, making your test useless. Because the algorithm written is the same, it's bound to be green, and yet it doesn't test what you want it to test. Instead you should do something like : 

```
    val baseUrl: String = … // a random url with protocol, host and path, e.g.: https://www.example.com/foo/bar/spam
    val queryParams: Map<String,String> = … // a random map of qp
    val queryParamString = queryParams.joinToString("&") { "${it.key}=${it.value}" }
    val url = "$baseUrl?$queryParamString"
```

xgouchet (changes requested):

dd-sdk-android/src/test/kotlin/com/datadog/android/tracing/TracingInterceptorTest.kt
```diff
@@ -704,18 +734,27 @@ internal open class TracingInterceptorTest {
...
 
-    private fun forgeUrl(forge: Forge, knownHost: String? = null): String {
+    private fun forgeUrlWithoutQueryParams(forge: Forge, knownHost: String? = null): String {
         val protocol = forge.anElementFrom("http", "https")
         val host = knownHost ?: forge.aStringMatching(HOSTNAME_PATTERN)
         val path = forge.anAlphaNumericalString()
         return "$protocol://$host/$path"
     }
 
+    private fun forgeUrl(forge: Forge, knownHost: String? = null): String {
```
xgouchet, line 744: Out of consistency, this one should be named `forgeUrlWithQueryParams`

xgouchet (changes requested):

xgouchet (approved):

PR #808, opened 2022-01-11
Update RUM Schema

fuzzybinary:
Completes RUMM-1899 To adds Flutter as an error source type

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumEventExt.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumEventExt.kt
@@ -55,14 +55,14 @@ internal fun RumResourceKind.toSchemaType(): ResourceEvent.ResourceType {
     }
 }
 
-internal fun RumErrorSource.toSchemaSource(): ErrorEvent.Source {
+internal fun RumErrorSource.toSchemaSource(): ErrorEvent.ErrorSource {
     return when (this) {
-        RumErrorSource.NETWORK -> ErrorEvent.Source.NETWORK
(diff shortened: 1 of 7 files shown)


fuzzybinary:
@buranmert Same for this as with the iOS PR. I've created this as a draft and once DataDog/rum-events-format#53 goes in we can do both at the same time.

mariusc83 (approved):

xgouchet (approved):

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumEventExt.kt
```diff
@@ -55,14 +55,14 @@ internal fun RumResourceKind.toSchemaType(): ResourceEvent.ResourceType {
     }
 }
 
-internal fun RumErrorSource.toSchemaSource(): ErrorEvent.Source {
+internal fun RumErrorSource.toSchemaSource(): ErrorEvent.ErrorSource {
```
0xnm, line 58: where this change (`ErrorEvent.Source` -> `ErrorEvent.ErrorSource` renaming) comes from? In the schema this property still called `source`.

fuzzybinary: That is a very good question. I had assumed either the schema changed for that or that the generator was attempting to avoid a name collision.

mariusc83: Yup that's it

0xnm (approved):

PR #815, opened 2022-01-14
RUMM-1920 Do not override 'service' property for WebView tracked events

mariusc83:
### What does this PR do?

We do not override anymore the `service` property in the RUM events received through the `JSBridge` with out own internal value. 

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/apiSurface
+++ b/dd-sdk-android/apiSurface
@@ -450,7 +450,7 @@ data class com.datadog.android.rum.model.ActionEvent
   data class CiTest
     constructor(kotlin.String)
     fun toJson(): com.google.gson.JsonElement
-    companion object
+    companion object 
       fun fromJson(kotlin.String): CiTest
   data class Dd
     constructor(DdSession? = null, kotlin.String? = null)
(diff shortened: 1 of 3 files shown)


ncreated (commented):
LGTM 👌, however I wonder if we won't be still overriding `service` from here:
https://github.com/DataDog/dd-sdk-android/blob/c73d65ea3fd0f1f342683e5f814c9e7674589537/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/net/RumOkHttpUploaderV2.kt#L33

We have the same logic in `dd-sdk-ios`, where `service` is sent as query parameter (in contrast - we don't write it in events). What will be the result of processing it on BE if service `A` is given by event, but service `B` is given in query?

mariusc83:
> > LGTM 👌, however I wonder if we won't be still overriding `service` from here:
> > https://github.com/DataDog/dd-sdk-android/blob/c73d65ea3fd0f1f342683e5f814c9e7674589537/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/net/RumOkHttpUploaderV2.kt#L33
> > 
> > We have the same logic in `dd-sdk-ios`, where `service` is sent as query parameter (in contrast - we don't write it in events). What will be the result of processing it on BE if service `A` is given by event, but service `B` is given in query?
> 
> @ncreated that is a very good observation and I actually do not know the answer. I will take this further with Louis next week to see if there are any implications.

I am actually not even sure if the `service` at the event level is used, probably is the same situation as with the `source` and the tag is created directly from the `service` in the query parameters. If that's the case I think to support the sub - apps we will have to drop this and only use the `service` at the event level.

0xnm (approved):

xgouchet (approved):

PR #817, opened 2022-01-18
RUMM-1868: Explicitly define supported TLS versions and cipher suites

0xnm:
### What does this PR do?

This change explicitly define supported TLS versions and cipher suites. The precise change list is the following:

* Dropped completely support of TLS 1.0 and 1.1 (which was enabled for API 19), because they are generally not supported (they are supported in EU1 only, but it has support of TLS 1.2 as well). It is ok to use TLS 1.2 on API 19, because it exists there, although it should be enabled explicitly, but OkHttp does that (see https://github.com/square/okhttp/blob/okhttp_3.12.x/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java#L452).
* TLS 1.3 cipher suites are kept (even though `TLS_CHACHA20_POLY1305_SHA256`, for example, is not yet confirmed with FIPS 140-2), because US5 supports it (it is only DC which supports it so far). We will state exact requirements for the FIPS 140-2 compliance in the docs and provide interface to configure network layer per customer needs.
* Final list of suites matched `RESTRICTED_CIPHER_SUITES` from OkHttp, with only difference that ChaCha20 based suites in TLS 1.2 are removed (they are not supported by any of intakes) and added some ciphers for KitKat compatibility (which may be removed once we bump min SDK).

Note on API 19: by looking on [this](https://developer.android.com/reference/javax/net/ssl/SSLEngine#getSupportedCipherSuites()) table it turns out that vanilla API 19 is not compatible with any of our intakes (except EU1), because it has no necessary cipher suites. However, `4.4.2` probably brings the necessary suites, because [ssllabs.com](https://www.ssllabs.com) can successfully negotiate using `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` (device details https://www.ssllabs.com/ssltest/viewClient.html?name=Android&version=4.4.2&key=62), but I cannot test it personally, because there is no AVD image with 4.4.2.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
@@ -62,9 +62,11 @@ import java.util.concurrent.ScheduledThreadPoolExecutor
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
+import okhttp3.CipherSuite
 import okhttp3.ConnectionSpec
 import okhttp3.OkHttpClient
 import okhttp3.Protocol
+import okhttp3.TlsVersion
 
 internal object CoreFeature {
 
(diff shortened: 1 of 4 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -93,7 +124,7 @@ internal object CoreFeature {
     internal var trackingConsentProvider: ConsentProvider = NoOpConsentProvider()
     internal var userInfoProvider: MutableUserInfoProvider = NoOpMutableUserInfoProvider()
 
-    internal var okHttpClient: OkHttpClient = OkHttpClient.Builder().build()
```
mariusc83, line 96: good catch

mariusc83 (approved):

xgouchet (approved):

PR #822, opened 2022-01-20
RUMM-1326 track Activity Intent action and data URI

xgouchet:
### What does this PR do?

Track the Intent's action and data URI, only if the "track extra" option is enabled.

### DIFF
--- a/dd-sdk-android/apiSurface
+++ b/dd-sdk-android/apiSurface
@@ -1253,6 +1253,7 @@ abstract class com.datadog.android.rum.tracking.ActivityLifecycleTrackingStrateg
   override fun onActivityStopped(android.app.Activity)
   override fun onActivityCreated(android.app.Activity, android.os.Bundle?)
   override fun onActivityResumed(android.app.Activity)
+  protected fun convertToRumAttributes(android.content.Intent?): Map<String, Any?>
   protected fun convertToRumAttributes(android.os.Bundle?): Map<String, Any?>
   companion object 
 class com.datadog.android.rum.tracking.ActivityViewTrackingStrategy : ActivityLifecycleTrackingStrategy, ViewTrackingStrategy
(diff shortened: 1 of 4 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/tracking/ActivityViewTrackingStrategy.kt
```diff
@@ -19,7 +19,8 @@ import com.datadog.android.rum.model.ViewEvent
  * A [ViewTrackingStrategy] that will track [Activity] as RUM Views.
  *
  * Each activity's lifecycle will be monitored to start and stop RUM Views when relevant.
- * @param trackExtras whether to track Activity Intent extras
+ * @param trackExtras whether to track the Activity's Intent information (extra attributes,
+ * action, data URI)
```
0xnm, line 23:
```suggestion
 * action, data URI).
```

0xnm (approved):
lgtm! Added minor spelling suggestion.

dd-sdk-android/apiSurface
```diff
@@ -1253,6 +1253,7 @@ abstract class com.datadog.android.rum.tracking.ActivityLifecycleTrackingStrateg
   override fun onActivityStopped(android.app.Activity)
   override fun onActivityCreated(android.app.Activity, android.os.Bundle?)
   override fun onActivityResumed(android.app.Activity)
+  protected fun convertToRumAttributes(android.content.Intent?): Map<String, Any?>
```
mariusc83, line 1256: Shouldn't we also add a monitor and a nightly test for this use case ?

mariusc83 (changes requested):

dd-sdk-android/apiSurface
xgouchet: No, it'll already be taken care of by the callers (tracking activities etc…). I'm not even sure why it is public.

mariusc83 (approved):

PR #823, opened 2022-01-20
RUMM-1898 RumMonitor#stopResourceWithError with stacktrace and errorType properties

mariusc83:
### What does this PR do?

This PR introduces a new overloaded method in the `RumMonitor` interface:
```
fun stopResourceWithError(
        key: String,
        statusCode: Int?,
        message: String,
        source: RumErrorSource,
        stackTrace: String,
        errorType: String?,
        attributes: Map<String, Any?> = emptyMap()
    )
```
This method is mainly added for being used from hybrid apps/sdks.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
@@ -208,7 +208,7 @@ interface RumMonitor {
         message: String,
         source: RumErrorSource,
         stackTrace: String,
-        errorType: String,
+        errorType: String?,
         attributes: Map<String, Any?> = emptyMap()
     )
 
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
@@ -185,6 +185,34 @@ interface RumMonitor {
     )
 
     /**
+     * Stops a previously started Resource that failed loading, linked with the [key] instance by
+     * providing the intercepted stacktrace.
+     * Note: This method should only be used from hybrid application.
+     * @param key the instance that represents the active view (usually your
+     * request or network call instance).
+     * @param statusCode the status code of the resource (if any)
+     * @param message a message explaining the error
+     * @param source the source of the error
+     * @param stackTrace the error stacktrace
+     * @param errorType the type of the error. Usually it should be the canonical name of the
+     * of the Exception class.
+     * @param attributes additional custom attributes to attach to the error. Attributes can be
+     * nested up to 9 levels deep. Keys using more than 9 levels will be sanitized by SDK.
+     * @see [startResource]
+     * @see [stopResource]
+     */
+    @SuppressWarnings("LongParameterList")
+    fun stopResourceWithError(
+        key: String,
+        statusCode: Int?,
+        message: String,
+        source: RumErrorSource,
+        stackTrace: String,
+        errorType: String,
+        attributes: Map<String, Any?> = emptyMap()
+    )
+
+    /**
      * Notifies that an error occurred in the active View.
      * @param message a message explaining the error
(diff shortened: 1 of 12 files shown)


fuzzybinary (approved):

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumMonitorE2ETests.kt
```diff
@@ -624,6 +625,42 @@ class RumMonitorE2ETests {
...
+    fun rum_rummonitor_stop_resource_with_error_stacktrace() {
+        val testMethodName = "rum_rummonitor_stop_resource_with_error_stacktrace"
+        val viewKey = forge.aViewKey()
+        val viewName = forge.aViewName()
+        val resourceKey = forge.aResourceKey()
+        val statusCode = forge.anInt(min = 400, max = 511)
+        val message = forge.aResourceErrorMessage()
+        val source = forge.aValueFrom(RumErrorSource::class.java)
+        val throwable = forge.aThrowable()
+        val stackTrace = throwable.loggableStackTrace()
```
xgouchet, line 641: Why do you use a "real" stack trace here ? It'd be better to use `forge.aString()` to ensure that _any_ String will still work and not jus strings formatted like a java stack trace

xgouchet (changes requested):

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumMonitorE2ETests.kt
mariusc83: Was thinking to use the real case but you are right, as the stackTrace can look different when coming from other platforms better to use a random string here.

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumMonitorE2ETests.kt
```diff
@@ -624,6 +624,41 @@ class RumMonitorE2ETests {
...
+    @Test
+    fun rum_rummonitor_stop_resource_with_error_stacktrace() {
+        val testMethodName = "rum_rummonitor_stop_resource_with_error_stacktrace"
+        val viewKey = forge.aViewKey()
+        val viewName = forge.aViewName()
+        val resourceKey = forge.aResourceKey()
+        val statusCode = forge.anInt(min = 400, max = 511)
+        val message = forge.aResourceErrorMessage()
+        val source = forge.aValueFrom(RumErrorSource::class.java)
+        val stackTrace = forge.anAlphaNumericalString()
```
xgouchet, line 639: you should use `aString()`. There's no guarantee at all that a stack trace will only contain alphanumerical chars. `aString()` will forge a string with any UTF8 character (so smileys, foreign language characters, weird punctuation symbols, …)

xgouchet (changes requested):

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -184,6 +184,34 @@ interface RumMonitor {
...
+     * @param stackTrace the error stacktrace
+     * @param errorType the type of the error. Usually it should be the canonical name of the
+     * of the Exception class.
+     * @param attributes additional custom attributes to attach to the error. Attributes can be
+     * nested up to 9 levels deep. Keys using more than 9 levels will be sanitized by SDK.
+     * @see [startResource]
+     * @see [stopResource]
+     */
+    @SuppressWarnings("LongParameterList")
+    fun stopResourceWithError(
```
0xnm, line 205: minor: shouldn't this be called `stopResourceWithStacktrace` (to be consistent and simplify the search)? Because new raw object is called `StopResourceWithStackTrace`.

PR #832, opened 2022-01-26
RUMM-1950 DatadogEventBridge - consume serialized rum web events

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
```diff
@@ -38,21 +38,30 @@ internal class WebViewLogEventConsumer(
         }
     }
 
-    private fun map(event: JsonObject): WebViewLogEvent? {
+    private fun map(event: JsonObject): JsonObject? {
```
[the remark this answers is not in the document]
buranmert: [we try to mutate some keys but if we can't, we still go on with the original JSON](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/FeaturesIntegration/WebView/WebLogEventConsumer.swift#L54)

our method can still throw before writing JSON to the disk but we don't expect that to happen (and we can't recover from it if it happens)

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
```diff
@@ -38,21 +38,30 @@ internal class WebViewLogEventConsumer(
...
-                logEvent.copy(additionalProperties = resolvedProperties, date = correctedDate)
+                event.addProperty(LogAttributes.RUM_APPLICATION_ID, rumContext.applicationId)
+                event.addProperty(LogAttributes.RUM_SESSION_ID, rumContext.sessionId)
+                event
             } else {
-                logEvent.copy(date = correctedDate)
+                event
             }
-        } catch (e: JsonParseException) {
+        } catch (e: ClassCastException) {
```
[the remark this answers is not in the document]
mariusc83: it's this call: ` event.get(DATE_KEY_NAME)?.asLong`. Basically `asLong` function can throw that if that property is not of a type that can be converted to Long (e.g. JsonArray)

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
```diff
@@ -38,21 +38,30 @@ internal class WebViewLogEventConsumer(
...
+            sdkLogger.e(JSON_PARSING_ERROR_MESSAGE, e)
+            return null
+        } catch (e: IllegalStateException) {
```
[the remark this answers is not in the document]
mariusc83: same call: https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5/com/google/gson/JsonElement.html#getAsLong--

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
```diff
@@ -29,75 +25,34 @@ internal class WebViewRumEventConsumer(
...
-                else -> {
-                    sdkLogger.e(WRONG_EVENT_TYPE_ERROR_MESSAGE.format(US, eventType))
-                    null
-                }
-            }
-        } catch (e: JsonParseException) {
+            val timeOffset = event.get(VIEW_KEY_NAME)?.asJsonObject?.get(VIEW_ID_KEY_NAME)
+                ?.asString?.let { getOffset(it) } ?: 0L
+            webViewRumEventMapper.mapEvent(event, rumContext, timeOffset)
+        } catch (e: ClassCastException) {
```
mariusc83, line 44: the `JsonPrimitve#asLong` methods states that can throw this exception. For now I could not see where and how can we reproduce this. I still think we need to keep it here just in case.

0xnm (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
```diff
@@ -38,8 +38,8 @@ internal class WebViewLogEventConsumer(
         }
     }
 
-    private fun map(event: JsonObject): JsonObject? {
-        try {
+    private fun map(event: JsonObject): JsonObject {
+        return try {
```
xgouchet, line 42: I'm not a fan of using this syntax. The `return` encompass a large try/catch block where the return value is always the same, so it doesn't make sense to do that. 
Also most of the caught exceptions are for a single line, so maybe only the unsafe line should be within the try-catch block instead of the whole method.

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
```diff
@@ -29,28 +30,29 @@ internal class WebViewRumEventConsumer(
...
-        }
+        dataWriter.write(mappedEvent)
     }
 
     private fun map(
         event: JsonObject,
         rumContext: RumContext?
-    ): JsonObject? {
+    ): JsonObject {
         return try {
```
xgouchet, line 40: Same thing, the return value is always the same, so this syntax makes it harder to read.

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
mariusc83: Regarding last remark, not necessarily as the `addTags` function can also throw.

xgouchet: It can throw but not the same exceptions

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
```diff
@@ -38,28 +38,47 @@ internal class WebViewLogEventConsumer(
         }
     }
 
-    private fun map(event: JsonObject): WebViewLogEvent? {
+    private fun map(event: JsonObject): JsonObject {
```
xgouchet, line 41: Yes much better here 👍

xgouchet (approved):

PR #845, opened 2022-02-10
Fix the signal handler to use a statically allocated backtrace heap

fuzzybinary:
### What does this PR do?

This changes the NDK signal handler to use a statically allocated buffer for recording the backtrace string instead of using the Stack. The way it was configured before, we used exactly the amount of allocated stack just for our backtrace string, which would overflow when any other variables were put on the stack, causing a SIGSEGV loop.

### Motivation

Fix bugs that were popping up when using the crash handler in Flutter.

### DIFF
--- a/dd-sdk-android-ndk/src/main/cpp/utils/signal-monitor.c
+++ b/dd-sdk-android-ndk/src/main/cpp/utils/signal-monitor.c
@@ -24,11 +24,11 @@
 static const char *LOG_TAG = "DatadogNdkCrashReporter";
 static stack_t signal_stack;
 
-/* the datadog sigaction which overrides the original signal handler */
-static volatile struct sigaction *volatile datadog_sigaction;
-
 /* the original sigaction array which will hold the original handlers for each overridden signal */
-volatile struct sigaction *volatile original_sigaction;
+volatile struct sigaction *volatile original_sigactions = NULL;
+
+/* pre-allocated memory for working with the backtrace */
+char* backtrace_scratch = NULL;
 
 volatile bool handlers_installed = false;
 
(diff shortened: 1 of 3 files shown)


dd-sdk-android-ndk/src/main/cpp/utils/signal-monitor.c
```diff
@@ -142,24 +143,26 @@ void uninstall_handlers() {
 }
 
 void handle_signal(int signum, siginfo_t *info, void *user_context) {
+    __android_log_write(ANDROID_LOG_INFO, LOG_TAG, "Handling signal in Datadog Crash Handler");
```
xgouchet, line 146: Not sure if we want to keep this here in prod

dd-sdk-android-ndk/src/main/cpp/utils/signal-monitor.c
```diff
@@ -249,6 +246,13 @@ bool try_to_install_handlers() {
 bool start_monitoring() {
     pthread_mutex_lock(&mutex);
     bool installed = try_to_install_handlers();
+    if (installed) {
+        __android_log_write(ANDROID_LOG_ERROR, LOG_TAG,
```
xgouchet, line 250: This one probably should be an INFO log and not an ERROR

dd-sdk-android-ndk/src/main/cpp/utils/signal-monitor.c
```diff
@@ -249,6 +246,13 @@ bool try_to_install_handlers() {
...
+                            "Successfully installed Datadog signal handlers");
+    } else {
+        __android_log_write(ANDROID_LOG_ERROR, LOG_TAG,
+                            "Unable to install signal handlers");
```
xgouchet, line 254:
```suggestion
                            "Unable to install Datadog NDK signal handlers");
```

xgouchet (changes requested):

xgouchet (approved):

PR #853, opened 2022-02-17
RUMM-1968 don't warn for silent orphan events

xgouchet:
### What does this PR do?

Prevent printing warning about dropped event for internal events


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -140,12 +140,13 @@ internal class RumSessionScope(
         actualWriter: DataWriter<Any>
     ) {
         val isValidAppLaunchEvent = event.javaClass in validAppLaunchEventTypes
+        val isSilentOrphanEvent = event.javaClass in silentOrphanEventTypes
```
mariusc83, line 143: as we will perform this operations every time we get an event I was thinking that maybe using sets here will be faster.

xgouchet: Given the size of the array, no it won't really change much

PR #858, opened 2022-02-28
RUMM-2039 Always send data when device has external power source

mariusc83:
### What does this PR do?

We are currently conditioning the data upload on the device battery status but this creates a problem when the device has a different power supply than a battery. In that case the `BATTERY_STATUS` is usually unknown and the `BATTERY_LEVEL` usually 0 which means no data will ever be sent from that device. Such devices can be the Android TV or any other device using the Android OS and running on external power supply.

In this PR we are fixing this issue by taking into account also the `PLUGGED_STATUS` in the `BatteryManager` information Intent. This will usually return one of the values `BatteryManager.BATTERY_PLUGGED_AC`, `BatteryManager.BATTERY_PLUGGED_WIRELESS`, `BatteryManager.BATTERY_PLUGGED_USB` in case the device is running on external power source.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
@@ -8,11 +8,9 @@ package com.datadog.android.core.internal.data.upload
 
 import com.datadog.android.core.configuration.UploadFrequency
 import com.datadog.android.core.internal.net.DataUploader
-import com.datadog.android.core.internal.net.UploadStatus
 import com.datadog.android.core.internal.net.info.NetworkInfoProvider
 import com.datadog.android.core.internal.persistence.Batch
 import com.datadog.android.core.internal.persistence.DataReader
-import com.datadog.android.core.internal.system.SystemInfo
 import com.datadog.android.core.internal.system.SystemInfoProvider
 import com.datadog.android.core.internal.utils.scheduleSafe
 import com.datadog.android.core.model.NetworkInfo
@@ -61,10 +59,10 @@ internal class DataUploadRunnable(
 
     private fun isSystemReady(): Boolean {
         val systemInfo = systemInfoProvider.getLatestSystemInfo()
-        val batteryFullOrCharging = systemInfo.batteryStatus in batteryFullOrChargingStatus
-        val batteryLevel = systemInfo.batteryLevel
-        val powerSaveMode = systemInfo.powerSaveMode
-        return (batteryFullOrCharging || batteryLevel > LOW_BATTERY_THRESHOLD) && !powerSaveMode
+        val hasEnoughPower = systemInfo.batteryFullOrCharging ||
+            systemInfo.onExternalPowerSource ||
+            systemInfo.batteryLevel > LOW_BATTERY_THRESHOLD
+        return hasEnoughPower && !systemInfo.powerSaveMode
     }
 
     private fun scheduleNextUpload() {
@@ -100,21 +98,7 @@ internal class DataUploadRunnable(
     // endregion
 
     companion object {
-
-        private val droppableBatchStatus = setOf(
-            UploadStatus.SUCCESS,
-            UploadStatus.HTTP_REDIRECTION,
-            UploadStatus.HTTP_CLIENT_ERROR,
-            UploadStatus.UNKNOWN_ERROR,
-            UploadStatus.INVALID_TOKEN_ERROR
-        )
-
-        private val batteryFullOrChargingStatus = setOf(
-            SystemInfo.BatteryStatus.CHARGING,
-            SystemInfo.BatteryStatus.FULL
-        )
-
-        private const val LOW_BATTERY_THRESHOLD = 10
+        internal const val LOW_BATTERY_THRESHOLD = 10
 
         internal const val MIN_DELAY_FACTOR = 1
         internal const val DEFAULT_DELAY_FACTOR = 5
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/system/BroadcastReceiverSystemInfoProvider.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/system/BroadcastReceiverSystemInfoProvider.kt
@@ -76,12 +76,15 @@ internal class BroadcastReceiverSystemInfoProvider(
         )
         val level = intent.getIntExtra(BatteryManager.EXTRA_LEVEL, -1)
         val scale = intent.getIntExtra(BatteryManager.EXTRA_SCALE, 100)
-
-        val resolvedBatteryStatus = SystemInfo.BatteryStatus.fromAndroidStatus(status)
-        val resolvedBatteryLevel = (level * 100) / scale
+        val pluggedStatus = intent.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1)
+        val batteryStatus = SystemInfo.BatteryStatus.fromAndroidStatus(status)
+        val batteryLevel = (level * 100) / scale
+        val onExternalPowerSource = pluggedStatus in PLUGGED_IN_STATUS_VALUES
+        val batteryFullOrCharging = batteryStatus in batteryFullOrChargingStatus
         systemInfo = systemInfo.copy(
-            batteryStatus = resolvedBatteryStatus,
-            batteryLevel = resolvedBatteryLevel
+            batteryFullOrCharging = batteryFullOrCharging,
+            batteryLevel = batteryLevel,
+            onExternalPowerSource = onExternalPowerSource
         )
     }
 
@@ -97,4 +100,18 @@ internal class BroadcastReceiverSystemInfoProvider(
     }
 
     // endregion
+
+    companion object {
+
+        private val batteryFullOrChargingStatus = setOf(
+            SystemInfo.BatteryStatus.CHARGING,
+            SystemInfo.BatteryStatus.FULL
+        )
+
+        private val PLUGGED_IN_STATUS_VALUES = setOf(
+            BatteryManager.BATTERY_PLUGGED_AC,
+            BatteryManager.BATTERY_PLUGGED_WIRELESS,
+            BatteryManager.BATTERY_PLUGGED_USB
+        )
+    }
 }
--- a/dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
+++ b/dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
@@ -86,8 +86,10 @@ internal class DataUploadRunnableTest {
             )
         whenever(mockNetworkInfoProvider.getLatestNetworkInfo()) doReturn fakeNetworkInfo
         val fakeSystemInfo = SystemInfo(
-            batteryStatus = forge.aValueFrom(SystemInfo.BatteryStatus::class.java),
-            batteryLevel = forge.anInt(20, 100)
+            batteryFullOrCharging = true,
+            batteryLevel = forge.anInt(min = 20, max = 100),
+            powerSaveMode = false,
+            onExternalPowerSource = true,
         )
         whenever(mockSystemInfoProvider.getLatestSystemInfo()) doReturn fakeSystemInfo
 
@@ -108,6 +110,8 @@ internal class DataUploadRunnableTest {
                 NetworkInfo.Connectivity.NETWORK_NOT_CONNECTED
             )
         whenever(mockNetworkInfoProvider.getLatestNetworkInfo()) doReturn networkInfo
+        whenever(mockReader.lockAndReadNext()) doReturn batch
+        whenever(mockDataUploader.upload(batch.data)) doReturn UploadStatus.SUCCESS
 
         testedRunnable.run()
 
@@ -121,60 +125,26 @@ internal class DataUploadRunnableTest {
         )
     }
 
-    @ParameterizedTest
-    @EnumSource(
-        SystemInfo.BatteryStatus::class,
-        names = ["DISCHARGING", "NOT_CHARGING"],
-        mode = EnumSource.Mode.INCLUDE
-    )
-    fun `doesn't send batch when battery is low and unplugged`(
-        batteryStatus: SystemInfo.BatteryStatus,
+    @Test
+    fun `M send batch W run() { batteryFullOrCharging }`(
         @Forgery batch: Batch,
-        forge: Forge
+        @IntForgery(min = 0, max = DataUploadRunnable.LOW_BATTERY_THRESHOLD) batteryLevel: Int
     ) {
-        val systemInfo = SystemInfo(
-            batteryStatus,
-            forge.anInt(1, 10)
+        val fakeSystemInfo = SystemInfo(
+            batteryFullOrCharging = true,
+            batteryLevel = batteryLevel,
+            onExternalPowerSource = false,
+            powerSaveMode = false
         )
-        whenever(mockSystemInfoProvider.getLatestSystemInfo()) doReturn systemInfo
         whenever(mockReader.lockAndReadNext()) doReturn batch
+        whenever(mockSystemInfoProvider.getLatestSystemInfo()) doReturn fakeSystemInfo
+        whenever(mockDataUploader.upload(batch.data)) doReturn UploadStatus.SUCCESS
 
         testedRunnable.run()
 
-        verify(mockReader, never()).drop(anyOrNull())
+        verify(mockReader).drop(batch)
         verify(mockReader, never()).release(batch)
-        verify(mockDataUploader, never()).upload(anyOrNull())
-        verify(mockThreadPoolExecutor).schedule(
-            same(testedRunnable),
-            any(),
-            eq(TimeUnit.MILLISECONDS)
-        )
-    }
-
-    @ParameterizedTest
-    @EnumSource(
-        SystemInfo.BatteryStatus::class,
-        names = ["DISCHARGING", "NOT_CHARGING"],
-        mode = EnumSource.Mode.INCLUDE
-    )
-    fun `doesn't send batch when power save mode is enabled`(
-        batteryStatus: SystemInfo.BatteryStatus,
-        @Forgery batch: Batch,
-        forge: Forge
-    ) {
-        val systemInfo = SystemInfo(
-            batteryStatus = batteryStatus,
-            batteryLevel = forge.anInt(50, 100),
-            powerSaveMode = true
-        )
-        whenever(mockSystemInfoProvider.getLatestSystemInfo()) doReturn systemInfo
-        whenever(mockReader.lockAndReadNext()) doReturn batch
-
-        testedRunnable.run()
-
-        verify(mockReader, never()).drop(anyOrNull())
-        verify(mockReader, never()).release(batch)
-        verify(mockDataUploader, never()).upload(anyOrNull())
+        verify(mockDataUploader).upload(batch.data)
         verify(mockThreadPoolExecutor).schedule(
             same(testedRunnable),
             any(),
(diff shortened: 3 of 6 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
```diff
@@ -64,7 +64,8 @@ internal class DataUploadRunnable(
         val batteryFullOrCharging = systemInfo.batteryStatus in batteryFullOrChargingStatus
         val batteryLevel = systemInfo.batteryLevel
         val powerSaveMode = systemInfo.powerSaveMode
-        return (batteryFullOrCharging || batteryLevel > LOW_BATTERY_THRESHOLD) && !powerSaveMode
+        val hasEnoughBattery = (batteryFullOrCharging || batteryLevel > LOW_BATTERY_THRESHOLD)
```
xgouchet, line 67: That's just nitpicking, but why don't  you do : 
```
val hasEnoughBattery = (systemInfo.externalPowerSource || batteryFullOrCharging || batteryLevel > LOW_BATTERY_THRESHOLD) 
return hasEnoughBattery && !powerSaveMode
```

That way we don't have complex boolean expressions, and IMO being plugged in means you do have enough battery.

xgouchet: Also it might be worth a refactor to be more consistent : 
- the `systemInfo.batteryStatus` is of type SystemInfo.BatteryStatus, and we have the list of valid values in the current class
- the `systemInfo.externalPowerSource` is of type boolean and the list of valid values is in the `BroadcastReceiverSystemInfoProvider` class

We should have both list used in the same way in only this or the other class.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -127,7 +127,7 @@ internal class DataUploadRunnableTest {
         names = ["DISCHARGING", "NOT_CHARGING"],
         mode = EnumSource.Mode.INCLUDE
     )
-    fun `doesn't send batch when battery is low and unplugged`(
+    fun `doesn't send batch when battery is low and charging`(
```
xgouchet, line 130: If the battery is low and charging, it should send the batch. I don't think the renaming was expected here.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -207,6 +207,101 @@ internal class DataUploadRunnableTest {
...
+        mode = EnumSource.Mode.INCLUDE
+    )
+    fun `M send batch W device has external power source`(
+        batteryStatus: SystemInfo.BatteryStatus,
+        @Forgery batch: Batch,
+        forge: Forge
+    ) {
+        val systemInfo = SystemInfo(
+            batteryStatus = batteryStatus,
+            batteryLevel = forge.anInt(0, 100),
```
xgouchet, line 222: Here we don't test that a battery level of 0 actually works (I mean we do once every 100 test). Maybe we should ensure that even with a very low battery, being plugged in always works. 
```suggestion
            batteryLevel = forge.anInt(0, 5),
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -207,6 +207,101 @@ internal class DataUploadRunnableTest {
...
+        )
+    }
+
+    @ParameterizedTest
+    @EnumSource(
+        SystemInfo.BatteryStatus::class,
+        names = ["DISCHARGING", "NOT_CHARGING"],
+        mode = EnumSource.Mode.INCLUDE
+    )
+    fun `M send batch W device has external power source { battery low }`(
```
xgouchet, line 247: Your test use a random battery level between 0 and 100, so it could be 90, which doesn't look like a low battery. Based on the enum source, it seems you're checking for unplugged state ?

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/system/BroadcastReceiverSystemInfoProviderTest.kt
```diff
@@ -62,6 +65,9 @@ internal class BroadcastReceiverSystemInfoProviderTest {
     @Mock
     lateinit var mockBuildSdkVersionProvider: BuildSdkVersionProvider
 
+    @IntForgery
+    var fakePluginStatus: Int = 0
```
xgouchet, line 69:
```suggestion
    var fakePluggedStatus: Int = 0
```

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
mariusc83: Because for me `externalPowerSource` is a bit out of the battery level context. It is more saying that no matter if this device has a battery or not right now it is using an external power source.
Regarding suggestion number 2 I am not sure if I understand what you mean. If you meant to make the `externalPowerSource` status as a value of the `SystemInfo.BatteryStatus` I am not really sure we want that as for me `externalPowerSource` is a different information. Or you want instead of the `externalPowerSource` as boolean to have `SystemInfo.PowerSource` enum and to have `WIFI`, `BLUETOOTH`, `AC` and `NOT_PLUGGED` ? I also thought about this but I found it a bit over doing it.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
mariusc83: Yeah I am still correcting those tests that's why I marked it as a draft ;)

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
mariusc83: Same thing as above

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
xgouchet: For me they both relate to the same thing, if you rename `hasEnoughBattery` to `hasEnoughPower`. 

Regarding solution to, basically we could simply move the computation of `batteryFullOrCharging` to the `BroadcastReceiverSystemInfoProvider` and store that as a boolean in the `SystemInfo` data class

mariusc83: Ok for the first point, I guess we could do the suggested for the second point but I would rather name that `hasEnoughPower` because in the case the device is plugged in there is no battery involved.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/system/BroadcastReceiverSystemInfoProviderTest.kt
```diff
@@ -322,5 +516,61 @@ internal class BroadcastReceiverSystemInfoProviderTest {
...
+            BatteryManager.BATTERY_STATUS_NOT_CHARGING,
+            BatteryManager.BATTERY_STATUS_FULL,
+            BatteryManager.BATTERY_STATUS_UNKNOWN
+        )
+
+        private val PLUGGED_STATUS_VALUES = intArrayOf(
+            BatteryManager.BATTERY_PLUGGED_AC,
+            BatteryManager.BATTERY_PLUGGED_WIRELESS,
+            BatteryManager.BATTERY_PLUGGED_USB,
+            0
```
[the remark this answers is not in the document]
mariusc83: yes, in case the device is on battery that value is 0 so we will test for this case also

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/system/BroadcastReceiverSystemInfoProvider.kt
```diff
@@ -76,12 +76,19 @@ internal class BroadcastReceiverSystemInfoProvider(
...
-
+        val pluggedStatus = intent.getIntExtra(BatteryManager.EXTRA_PLUGGED, -1)
         val resolvedBatteryStatus = SystemInfo.BatteryStatus.fromAndroidStatus(status)
         val resolvedBatteryLevel = (level * 100) / scale
+        val resolvedExternalPowerSource = pluggedStatus in PLUGGED_IN_STATUS_VALUES
+        val batteryIsFullOrCharging = resolvedBatteryStatus in batteryFullOrChargingStatus
+        val resolvedHasEnoughPower = (
+            resolvedExternalPowerSource || batteryIsFullOrCharging ||
+                resolvedBatteryLevel > LOW_BATTERY_THRESHOLD
+            ) && !systemInfo.powerSaveMode
```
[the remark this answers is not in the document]
xgouchet: > this is the way I had it in the first
Nope, what I asked in my PR in your draft was to have the comparison between the BatteryStatus / PluggedState in the intent and the constant in the Android Framework both be done in the same place, to have a consistent SystemInfo.
And again, the role of the receiver is to give information about the system, without knowing what this info is used for. The role of the runnable is to use this information without knowing how it's gathered.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/system/BroadcastReceiverSystemInfoProvider.kt
```diff
@@ -91,10 +98,27 @@ internal class BroadcastReceiverSystemInfoProvider(
             val powerManager = context.getSystemService(Context.POWER_SERVICE) as? PowerManager
             val powerSaveMode = powerManager?.isPowerSaveMode ?: false
             systemInfo = systemInfo.copy(
-                powerSaveMode = powerSaveMode
+                powerSaveMode = powerSaveMode,
+                hasEnoughPower = !powerSaveMode
```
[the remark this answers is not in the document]
mariusc83: Yes my mistake there, should only be setting `hasEnoughPower=false` in case `powerSaveMode=true` otherwise it should keep the old value. In any case I am going to change it again and follow the last suggestion.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -86,8 +86,8 @@ internal class DataUploadRunnableTest {
             )
         whenever(mockNetworkInfoProvider.getLatestNetworkInfo()) doReturn fakeNetworkInfo
         val fakeSystemInfo = SystemInfo(
-            batteryStatus = forge.aValueFrom(SystemInfo.BatteryStatus::class.java),
-            batteryLevel = forge.anInt(20, 100)
+            batteryFullOrCharging = true,
+            batteryLevel = forge.anInt(min = 90, max = 100)
```
xgouchet, line 90: why are you changing the min value here ?

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -121,60 +123,97 @@ internal class DataUploadRunnableTest {
...
+    @Test
+    fun `M send batch W run() { battery level high }`(
         @Forgery batch: Batch,
-        forge: Forge
+        @IntForgery(min = DataUploadRunnable.LOW_BATTERY_THRESHOLD + 1) batteryLevel: Int
     ) {
-        val systemInfo = SystemInfo(
-            batteryStatus,
-            forge.anInt(1, 10)
+        val fakeSystemInfo = SystemInfo(batteryLevel = batteryLevel)
```
xgouchet, line 150: Can we explicit the other values in system Info:
```
SystemInfo(
    batteryLevel = batteryLevel,
    batteryFullOrCharging = randomBool(), 
    onExternalPowerSource = randomBool(),
    powerSaveMode = false
)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -121,60 +123,97 @@ internal class DataUploadRunnableTest {
...
-    @EnumSource(
-        SystemInfo.BatteryStatus::class,
-        names = ["DISCHARGING", "NOT_CHARGING"],
-        mode = EnumSource.Mode.INCLUDE
-    )
-    fun `doesn't send batch when battery is low and unplugged`(
-        batteryStatus: SystemInfo.BatteryStatus,
+    @Test
+    fun `M send batch W run() { batteryFullOrCharging }`(@Forgery batch: Batch) {
+        val fakeSystemInfo = SystemInfo(batteryFullOrCharging = true)
```
xgouchet, line 128: Can we explicit the other values in system Info:
```
SystemInfo(
    batteryLevel = randomInt(0,100),
    batteryFullOrCharging = true, 
    onExternalPowerSource = randomBool(),
    powerSaveMode = false
)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -121,60 +123,97 @@ internal class DataUploadRunnableTest {
...
+            same(testedRunnable),
+            any(),
+            eq(TimeUnit.MILLISECONDS)
         )
-        whenever(mockSystemInfoProvider.getLatestSystemInfo()) doReturn systemInfo
+    }
+
+    @Test
+    fun `M send batch W run() { onExternalPower }`(@Forgery batch: Batch) {
+        val fakeSystemInfo = SystemInfo(onExternalPowerSource = true)
```
xgouchet, line 169: Can we explicit the other values in system Info:
```
SystemInfo(
    batteryLevel = randomInt(0,100),
    batteryFullOrCharging = randomBool(), 
    onExternalPowerSource = true,
    powerSaveMode = false
)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -121,60 +123,97 @@ internal class DataUploadRunnableTest {
...
-        @Forgery batch: Batch,
-        forge: Forge
-    ) {
-        val systemInfo = SystemInfo(
-            batteryStatus = batteryStatus,
-            batteryLevel = forge.anInt(50, 100),
-            powerSaveMode = true
+    @Test
+    fun `M not send batch W run() { not enough battery }`(@Forgery batch: Batch) {
+        val fakeSystemInfo = SystemInfo()
```
xgouchet, line 188: Can we explicit the other values in system Info:
```
SystemInfo(
    batteryLevel = randomInt(0, threshold),
    batteryFullOrCharging = false, 
    onExternalPowerSource = false,
    powerSaveMode = false
)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -121,60 +123,97 @@ internal class DataUploadRunnableTest {
...
+            same(testedRunnable),
+            any(),
+            eq(TimeUnit.MILLISECONDS)
         )
-        whenever(mockSystemInfoProvider.getLatestSystemInfo()) doReturn systemInfo
+    }
+
+    @Test
+    fun `M not send batch W run() { batteryFullOrCharging, powerSaveMode}`(@Forgery batch: Batch) {
+        val fakeSystemInfo = SystemInfo(batteryFullOrCharging = true, powerSaveMode = true)
```
xgouchet, line 207: Can we explicit the other values in system Info:
```
SystemInfo(
    batteryLevel = randomInt(0, 100),
    batteryFullOrCharging = true, 
    onExternalPowerSource = false,
    powerSaveMode = true
)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -183,23 +222,41 @@ internal class DataUploadRunnableTest {
...
-    fun `batch sent when battery is low and charging`(
+    fun `M not send batch W run() { batteryLeveHigh, powerSaveMode}`(
         @Forgery batch: Batch,
-        forge: Forge
+        @IntForgery(min = DataUploadRunnable.LOW_BATTERY_THRESHOLD + 1) batteryLevel: Int
     ) {
-        val systemInfo = SystemInfo(
-            SystemInfo.BatteryStatus.CHARGING,
-            forge.anInt(1, 10)
+        val fakeSystemInfo = SystemInfo(batteryLevel = batteryLevel, powerSaveMode = true)
```
xgouchet, line 229: Can we explicit the other values in system Info:
```
SystemInfo(
    batteryLevel = batteryLevel,
    batteryFullOrCharging = false, 
    onExternalPowerSource = false,
    powerSaveMode = true
)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -183,23 +222,41 @@ internal class DataUploadRunnableTest {
...
+            eq(TimeUnit.MILLISECONDS)
         )
-        whenever(mockSystemInfoProvider.getLatestSystemInfo()) doReturn systemInfo
+    }
+
+    @Test
+    fun `M not send batch W run() { onExternalPower, powerSaveMode}`(
+        @Forgery batch: Batch
+    ) {
+        val fakeSystemInfo = SystemInfo(onExternalPowerSource = true, powerSaveMode = true)
```
xgouchet, line 250: Can we explicit the other values in system Info:
```
SystemInfo(
    batteryLevel = randomInt(0,100),
    batteryFullOrCharging = false, 
    onExternalPowerSource = true,
    powerSaveMode = true
)
```

xgouchet (changes requested):

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
mariusc83: not sure why, probably was a left over from pervious iterations, will put it back

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
mariusc83: if you do that then you will actually add a false positive in the test, having the `onExternalPowerSource=true` will actually make the test positive event if the battery level condition fails.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
mariusc83: same remark as above, you will introduce false positives

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
mariusc83: I will make them explicit though but without adding the randomness

xgouchet (approved):

0xnm (approved):

PR #869, opened 2022-03-14
RUMM-2021: Support telemetry events in the events processing pipeline

0xnm:
### What does this PR do?

This PR adds support of sending telemetry events as a part of general RUM events processing pipeline (because these events go to the RUM intake).

The necessary private methods are added to the `AdvancedRumMonitor` (they are quite basic ones for now, without support of stack as a string, for example, for now).

Processing happens outside of any RUM scope (to not depend on the scope being created) and also outside of application and session scopes (to avoid the possible confusion with taking RUM context of those, which will miss info about current view and action).

Also this change updates the version of the Kotlin AST parsing lib, because I've hit a bug with existing version (also update solves the issue with wrong entries in the `apiSurface`).

The rest of work like replacing the existing internal monitoring feature will be in other PRs.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
@@ -1,0 +1,101 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.telemetry.internal
+
+import com.datadog.android.core.internal.persistence.DataWriter
+import com.datadog.android.core.internal.time.TimeProvider
+import com.datadog.android.core.internal.utils.loggableStackTrace
+import com.datadog.android.rum.GlobalRum
+import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.domain.event.RumEventSourceProvider
+import com.datadog.android.rum.internal.domain.scope.RumRawEvent
+import com.datadog.android.telemetry.model.TelemetryDebugEvent
+import com.datadog.android.telemetry.model.TelemetryErrorEvent
+
+internal class TelemetryEventHandler(
+    private val serviceName: String,
+    private val sdkVersion: String,
+    private val sourceProvider: RumEventSourceProvider,
+    private val timeProvider: TimeProvider
+) {
+
+    fun handleEvent(event: RumRawEvent.SendTelemetry, writer: DataWriter<Any>) {
+
+        val timestamp = event.eventTime.timestamp + timeProvider.getServerOffsetMillis()
+
+        val rumContext = GlobalRum.getRumContext()
+
+        val telemetryEvent: Any = when (event.type) {
+            TelemetryType.DEBUG -> {
+                createDebugEvent(
+                    timestamp, rumContext, event.message
+                )
+            }
+            TelemetryType.ERROR -> {
+                createErrorEvent(
+                    timestamp,
+                    rumContext,
+                    event.message,
+                    event.throwable
+                )
+            }
+        }
+
+        writer.write(telemetryEvent)
+    }
+
+    private fun createDebugEvent(
+        timestamp: Long,
+        rumContext: RumContext,
+        message: String
+    ): TelemetryDebugEvent {
+        return TelemetryDebugEvent(
+            dd = TelemetryDebugEvent.Dd(),
+            date = timestamp,
+            source = sourceProvider.telemetryDebugEventSource
+                ?: TelemetryDebugEvent.Source.ANDROID,
+            service = serviceName,
+            version = sdkVersion,
+            application = TelemetryDebugEvent.Application(rumContext.applicationId),
+            session = TelemetryDebugEvent.Session(rumContext.sessionId),
+            view = rumContext.viewId?.let { TelemetryDebugEvent.View(it) },
+            action = rumContext.actionId?.let { TelemetryDebugEvent.Action(it) },
+            telemetry = TelemetryDebugEvent.Telemetry(
+                message = message
+            )
+        )
+    }
+
+    private fun createErrorEvent(
+        timestamp: Long,
+        rumContext: RumContext,
+        message: String,
+        throwable: Throwable?
+    ): TelemetryErrorEvent {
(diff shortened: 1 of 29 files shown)


mariusc83 (approved):
Great job. LGTM !!

dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -0,0 +1,80 @@
...
+                    dd = TelemetryDebugEvent.Dd(),
+                    date = timestamp,
+                    source = sourceProvider.telemetryDebugEventSource
+                        ?: TelemetryDebugEvent.Source.ANDROID,
+                    service = serviceName,
+                    version = sdkVersion,
+                    application = TelemetryDebugEvent.Application(applicationId),
+                    session = TelemetryDebugEvent.Session(sessionId),
+                    view = TelemetryDebugEvent.View(activeViewId),
+                    action = TelemetryDebugEvent.Action(activeActionId),
```
xgouchet, line 47: view and actions are optional in the schema, so we should make them null when no viewid/actionid is available instead of sending an empty string

dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.datadog.android.telemetry.model.TelemetryErrorEvent
+
+internal class TelemetryEventHandler(
+    private val serviceName: String,
+    private val sdkVersion: String,
+    private val sourceProvider: RumEventSourceProvider,
+    private val timeProvider: TimeProvider
+) {
+
+    fun handleEvent(event: RumRawEvent.SendTelemetry, writer: DataWriter<Any>) {
```
xgouchet, line 25: Nitpicking, this method is quite large, can we split it a bit?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventSerializerTest.kt
```diff
@@ -389,6 +391,82 @@ internal class RumEventSerializerTest {
...
+        }
+        event.session?.let {
+            assertThat(jsonObject).hasField("session") { hasField("id", it.id) }
+        }
+        event.view?.let {
+            assertThat(jsonObject).hasField("view") { hasField("id", it.id) }
+        }
+        event.action?.let {
+            assertThat(jsonObject).hasField("action") { hasField("id", it.id) }
+        }
```
xgouchet, line 425: theorically, if `event.x` is null, we should test that the jsonObject *doesn't* have a field named `x`

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventSerializerTest.kt
```diff
@@ -389,6 +391,82 @@ internal class RumEventSerializerTest {
...
+        event.telemetry.error?.let {
+            assertThat(jsonObject).hasField("telemetry") {
+                hasField("error") {
+                    hasNullableField("stack", it.stack)
+                    hasNullableField("kind", it.kind)
+                }
+            }
+        }
```
xgouchet, line 467: Same here

dd-sdk-android/src/test/kotlin/com/datadog/android/utils/assertj/DeserializedTelemetryDebugEventAssert.kt
```diff
@@ -0,0 +1,30 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.utils.assertj
+
+import com.datadog.android.telemetry.model.TelemetryDebugEvent
+import org.assertj.core.api.AbstractAssert
+
+internal class DeserializedTelemetryDebugEventAssert(actual: TelemetryDebugEvent) :
```
xgouchet, line 12: Why do you need this one, when there's already a `TelemetryDebugEventAssert`?

dd-sdk-android/src/test/kotlin/com/datadog/android/utils/assertj/DeserializedTelemetryErrorEventAssert.kt
```diff
@@ -0,0 +1,30 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.utils.assertj
+
+import com.datadog.android.telemetry.model.TelemetryErrorEvent
+import org.assertj.core.api.AbstractAssert
+
+internal class DeserializedTelemetryErrorEventAssert(actual: TelemetryErrorEvent) :
```
xgouchet, line 12: Same, there's already a `TelemetryErrorEventAssert`

xgouchet (changes requested):

dd-sdk-android/src/test/kotlin/com/datadog/android/utils/assertj/DeserializedTelemetryDebugEventAssert.kt
0xnm: To be consistent with other `DeserializedXXXAssert` classes which exist for other RUM events, even though these RUM events have `XXXAssert` classes as well. Do they (`DeserializedXXXAssert`) exist only to skip `context`/`usr`/`additionalProperties` fields from comparison?

xgouchet: I don't know why those exist… I probably missed them in previous code reviews. IMHO we should only have one class, and if need be create distinct `equalsWithX` or `equalsWithoutY` methods to make it understandable.

0xnm: I've extracted the assertion code in https://github.com/DataDog/dd-sdk-android/pull/869/commits/22097e1fddfc518593360741c2663bc497bb477c. Using recursive comparison has one advantage over `TelemetryXXXAssert`: no need to do anything if new field is added (but readability in case of the failure is worse). And in this test actual and expected share the same type, so recursive comparison also makes sense.

xgouchet (approved):

PR #871, opened 2022-03-16
Handle missed orphans

### LINKED ISSUE #849: Crash Reporting Dependent on View Tracking Strategy
**Describe what happened**

My team is running into some issues with DataDog's RUM crash reporting and how it seems to be integrated with the ViewTrackingStrategy configuration. We are interested in the insight that comes from view tracking, but we also want reporting on crashes that can occur outside of the `onPause` and `onResume` fragment and activity lifecycles. Is this possible with the DataDog SDK?

In the following scenarios, the RUM session does not show that the session ends with an error. No crash seems to be reported.

Crash in Fragment
1. Configure the DataDog SDK with `useViewTrackingStrategy(MixedViewTrackingStrategy(false)`) or `FragmentViewTrackingStrategy(false)`.
2. Force a crash in a Fragment within the `onViewCreated` lifecycle method.
3. Observe that no crash report is sent to DataDog.

Crash in JobIntentService.
1. Configure the DataDog SDK to use `ActivityViewTrackingStrategy` or use no view tracking strategy
2. Schedule a job in a `JobIntentService` with a crash in the `onHandleWork` lifecycle method. 
3. Observe that no crash report is sent to DataDog. 

**Note**: We noticed here that adding `GlobalRum.get().addView()` before triggering the crash will allow DataDog to report the crash.

Crashes are always reported regardless of any UI lifecycle.

**Additional context**
 - Android OS version: Android 11
 - Device Model: Pixel 5
 - Datadog SDK version: 1.11.1

Thank you!

xgouchet:
### What does this PR do?

Fixes #849 

Some orphan events were not counted as orphans

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
@@ -47,7 +47,7 @@ internal class RumSessionScope(
     private val sessionMaxDurationNanos: Long = DEFAULT_SESSION_MAX_DURATION_NS
 ) : RumScope {
 
-    internal val activeChildrenScopes = mutableListOf<RumScope>()
+    internal val childrenScopes = mutableListOf<RumScope>()
 
     internal var keepSession: Boolean = false
     internal var sessionId = RumContext.NULL_UUID
@@ -80,7 +80,7 @@ internal class RumSessionScope(
 
         val actualWriter = if (keepSession) writer else noOpWriter
 
-        val iterator = activeChildrenScopes.iterator()
+        val iterator = childrenScopes.iterator()
         @Suppress("UnsafeThirdPartyFunctionCall") // next/remove can't fail: we checked hasNext
         while (iterator.hasNext()) {
             val scope = iterator.next().handleEvent(event, actualWriter)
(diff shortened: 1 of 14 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -102,7 +102,7 @@ internal class RumSessionScope(
             )
             onViewDisplayed(event, viewScope, actualWriter)
             activeChildrenScopes.add(viewScope)
-        } else if (activeChildrenScopes.size == 0) {
+        } else if (activeChildrenScopes.count { it.isActive() } == 0) {
```
0xnm, line 105: Should we rename `activeChildrenScopes` field somehow? Or use other naming for `isActive`? Because with that change it introduces some ambiguity: basically this line says that `activeChildrenScopes` can contain some scope which is not actually `active` (and `activeChildrenScopes` was filtered just few lines above).

Or we can introduce set of states for `RumScope`: `active`, `stopping`, `stopped` or something like this. WDYT?

xgouchet: Indeed I'll rename `activeChildrenScopes` -> `childrenScopes`

PR #877, opened 2022-03-22
RUMM-2028 Reduce number of views updates in payloads

mariusc83:
### What does this PR do?

During a `ViewScope` lifecycle we are sending too many `ViewEvent` updates which are significantly increasing the size of our payloads and also increasing the data traffic. Above all these extra `ViewEvents` are affecting our reducers performance in the backend. We are currently introducing a special `predicate` that will limit the `ViewEvent` updates to 1/30 seconds with a wildcard in case the scope `ViewScope` was finalized.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/DefaultViewUpdatePredicate.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/DefaultViewUpdatePredicate.kt
@@ -13,8 +13,13 @@ internal class DefaultViewUpdatePredicate(
 ) : ViewUpdatePredicate {
     private var lastViewUpdateTimestamp = System.nanoTime() - VIEW_UPDATE_THRESHOLD_IN_NS
 
-    override fun canUpdateView(isViewComplete: Boolean): Boolean {
-        if (isViewComplete || System.nanoTime() - lastViewUpdateTimestamp > viewUpdateThreshold) {
+    override fun canUpdateView(isViewComplete: Boolean, event: RumRawEvent): Boolean {
+        val isFatalError = event is RumRawEvent.AddError && event.isFatal
+        val isThresholdReached = System.nanoTime() - lastViewUpdateTimestamp > viewUpdateThreshold
+        if (isViewComplete ||
+            isFatalError ||
+            isThresholdReached
+        ) {
             lastViewUpdateTimestamp = System.nanoTime()
             return true
         }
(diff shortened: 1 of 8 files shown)


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/DefaultViewUpdatePredicate.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val viewUpdateThreshold: Long = VIEW_UPDATE_THRESHOLD_IN_NS
+) : ViewUpdatePredicate {
+    private var lastViewUpdateTimestamp = System.nanoTime() - VIEW_UPDATE_THRESHOLD_IN_NS
+
+    override fun canUpdateView(isViewComplete: Boolean, event: RumRawEvent): Boolean {
+        val isFatalError = event is RumRawEvent.AddError && event.isFatal
+        if (isViewComplete ||
+            isFatalError ||
+            (System.nanoTime() - lastViewUpdateTimestamp > viewUpdateThreshold)
+        ) {
```
xgouchet, line 21: Just a nitpick for readability

```suggestion
        val isFatalError = event is RumRawEvent.AddError && event.isFatal
        val isRequired = System.nanoTime() - lastViewUpdateTimestamp > viewUpdateThreshold
        if (isViewComplete || isFatalError || isRequired) {
```

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/DefaultViewUpdatePredicate.kt
mariusc83: Yes, I wanted to add that too but then I thought that we should not compute that all the time, but you are right, it is not that much and is better for visibility

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/DefaultViewPredicateTest.kt
```diff
@@ -0,0 +1,159 @@
...
+        // When
+        val canUpdateViewSecond = testedViewPredicate.canUpdateView(false, mockEvent)
+
+        // Then
+        assertThat(canUpdateViewFirst).isTrue
+        assertThat(canUpdateViewSecond).isFalse
+    }
+
+    @Test
+    fun `M return true W canUpdateView { second call before threshold, fatal error }`() {
```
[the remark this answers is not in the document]
mariusc83: Yes, I was aware of those but was thinking that they are not relevant.

PR #879, opened 2022-03-25
RUMM-2061 Disable RUM vitals monitor if no active foreground view

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -54,7 +54,8 @@ internal open class RumViewScope(
     internal val timeProvider: TimeProvider,
     private val rumEventSourceProvider: RumEventSourceProvider,
     private val buildSdkVersionProvider: BuildSdkVersionProvider = DefaultBuildSdkVersionProvider(),
-    private val viewUpdatePredicate: ViewUpdatePredicate = DefaultViewUpdatePredicate()
+    private val viewUpdatePredicate: ViewUpdatePredicate = DefaultViewUpdatePredicate(),
+    internal val type: RumViewType = RumViewType.FOREGROUND
 ) : RumScope {
 
     internal val url = key.resolveViewUrl().replace('.', '/')
@@ -156,7 +157,6 @@ internal open class RumViewScope(
 
             is RumRawEvent.StartView -> onStartView(event, writer)
             is RumRawEvent.StopView -> onStopView(event, writer)
-
             is RumRawEvent.StartAction -> onStartAction(event, writer)
             is RumRawEvent.StartResource -> onStartResource(event, writer)
             is RumRawEvent.AddError -> onAddError(event, writer)
@@ -189,7 +189,8 @@ internal open class RumViewScope(
                 viewId = viewId,
                 viewName = name,
                 viewUrl = url,
-                actionId = (activeActionScope as? RumActionScope)?.actionId
+                actionId = (activeActionScope as? RumActionScope)?.actionId,
+                viewType = type
             )
     }
 
@@ -216,6 +217,7 @@ internal open class RumViewScope(
         val startedKey = keyRef.get()
         val shouldStop = (event.key == startedKey) || (startedKey == null)
         if (shouldStop && !stopped) {
+            GlobalRum.updateRumContext(getRumContext().copy(viewType = RumViewType.NONE))
             attributes.putAll(event.attributes)
             stopped = true
             sendViewUpdate(event, writer)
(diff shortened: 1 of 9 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -170,10 +171,11 @@ internal open class RumViewScope(
             else -> delegateEventToChildren(event, writer)
         }
 
-        return if (isViewComplete()) {
-            null
+        if (isViewComplete()) {
+            GlobalRum.updateRumContext(getRumContext().copy(viewType = RumViewType.NONE))
```
xgouchet, line 175: You should do that when stopping the current view only. The `isViewComplete()` can return true a long time after the view is stopped (e.g.: an activity goes in background, but there are ongoing network requests, so the view is still not complete). 
You could also introduce some issues if you navigate to another view, and 5 secondes later you update the rum context with your own...

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -647,6 +647,7 @@ internal class RumSessionScopeTest {
                 assertThat(it.cpuVitalMonitor).isInstanceOf(NoOpVitalMonitor::class.java)
                 assertThat(it.memoryVitalMonitor).isInstanceOf(NoOpVitalMonitor::class.java)
                 assertThat(it.frameRateVitalMonitor).isInstanceOf(NoOpVitalMonitor::class.java)
+                assertThat(it.type).isEqualTo(RumViewScope.RumViewType.BACKGROUND)
```
xgouchet, line 650: You should also check that when a view is started by a `startView(…)` event, the type is `FOREGROUND`

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -295,6 +295,53 @@ internal class RumViewScopeTest {
...
+            mockViewUpdatePredicate,
+            type = fakeViewEventType
+        )
+
+        // Then
+        assertThat(GlobalRum.getRumContext().viewType).isEqualTo(fakeViewEventType)
+    }
+
+    @Test
+    fun `𝕄 update the context with viewType NONE W handleEvent(StopView)`(
```
xgouchet, line 326: Can you add a test 
`"𝕄 not update the context with viewType NONE W handleEvent(StartView) {another view is}"` to ensure we don't mess up the context from the next view

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
mariusc83: You're right, I am going to do the change here.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/vitals/VitalReaderRunnableTest.kt
```diff
@@ -83,6 +94,28 @@ internal class VitalReaderRunnableTest {
...
+    @EnumSource(
+        value = RumViewScope.RumViewType::class,
+        names = ["FOREGROUND"],
+        mode = EnumSource.Mode.EXCLUDE
+    )
+    fun `𝕄 not read data, not notify observer but schedule 𝕎 run { viewType != FOREGROUND }()`(
+        viewType: RumViewScope.RumViewType
+    ) {
+        // Given
+        whenever(mockReader.readVitalData()) doReturn fakeValue
```
0xnm, line 107: minor: this line is not needed

xgouchet (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -295,10 +295,152 @@ internal class RumViewScopeTest {
...
+            fakeEventTime,
+            fakeAttributes,
+            mockDetector,
+            mockCpuVitalMonitor,
+            mockMemoryVitalMonitor,
+            mockFrameRateVitalMonitor,
+            mockTimeProvider,
+            mockRumEventSourceProvider,
+            mockBuildSdkVersionProvider,
+            mockViewUpdatePredicate
```
0xnm, line 393: minor: we maybe need to write parameterized test here, by checking both `FOREGROUND` and `BACKGROUND` values, same for the test below.

0xnm (approved):
lgtm! I added a comment about to improve the test, but it is non-blocking for me.

PR #880, opened 2022-03-25
RUMM-2023: Use telemetry for the SDK internal logs

0xnm:
### What does this PR do?

This change removes `Internal Logs` feature, which was forwarding SDK internal logs to the Logs endpoint, and replaces it with `Telemetry` feature - SDK internal logs will be forwarded as telemetry events as a part of RUM pipeline.

This is achieved by changing `sdkLogger` - in debug builds it will output to both logcat and telemetry, in release builds only to telemetry.

Important thing to mention is the during the local development (debug builds) we want as much as possible information in the logcat, but we don't want all this information in the telemetry. To achieve this conditional dispatch new log levels (for internal use only) are introduced: error with telemetry, warn with telemetry, debug with telemetry. They are the same as normal `error`, `warn` or `debug`, but also indicate that such log entries should be sent to the telemetry as well, with the help of the `telemetry` bit flag. If flag is not there, no telemetry will be sent.

Levels above are applied to __some__ of `sdkLogger` calls following the logic:

* If there is a `devLogger` call next to `sdkLogger` call - don't use telemetry (but not in all cases)
* Reported entry should be actionable.

Note that the list of such calls (with telemetry) is not yet final, subject to change.

Also this PR tweaks the usage of `CoreFeature` in the tests to prevent `Kronos` spinning sync threads (which are not necessary in the unit testing and also avoid calls to the `LoggingSyncListener`).

### Motivation

### Additional Notes


dd-sdk-android/src/main/kotlin/com/datadog/android/log/internal/logger/InternalLogHandler.kt
```diff
@@ -0,0 +1,46 @@
...
+) : LogHandler {
+    override fun handleLog(
+        level: Int,
+        message: String,
+        throwable: Throwable?,
+        attributes: Map<String, Any?>,
+        tags: Set<String>,
+        timestamp: Long?
+    ) {
+        val reportLevel = level and TELEMETRY_LOG_FLAG.inv()
```
mariusc83, line 23: Nice trick

dd-sdk-android/src/main/kotlin/com/datadog/android/log/internal/utils/LogUtils.kt
```diff
@@ -18,3 +20,55 @@ internal fun buildLogDateFormat(): SimpleDateFormat =
     SimpleDateFormat(ISO_8601, Locale.US).apply {
         timeZone = TimeZone.getTimeZone("UTC")
     }
+
+// region Telemetry logging
+
+internal const val TELEMETRY_LOG_FLAG = 32
```
mariusc83, line 26: This is more like a mask for me

dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/Telemetry.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.telemetry.internal
+
+import com.datadog.android.rum.GlobalRum
+import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
+import com.datadog.android.rum.internal.monitor.NoOpAdvancedRumMonitor
+
+internal class Telemetry {
```
mariusc83, line 13: Why do you need this extra class and not call `RumMonitor` directly in the `TelemetryLogHandler` ?

mariusc83 (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/Telemetry.kt
0xnm: because per RFC it is supposed also to use `telemetry` for reporting configuration and, maybe, performance metrics, so generally telemetry != logs.

dd-sdk-android/src/testDebug/kotlin/com/datadog/android/core/internal/utils/RuntimeUtilsDebugTest.kt
```diff
@@ -38,67 +26,18 @@ import org.mockito.quality.Strictness
...
-            .hasSize(1)
-            .allMatch { it.serviceName == SDK_LOG_PREFIX }
-        assertThat(handlers.filterIsInstance<DatadogLogHandler>())
-            .hasSize(1)
-            .allMatch { it.logGenerator.serviceName == InternalLogsFeature.SERVICE_NAME }
-            .allMatch { it.logGenerator.envTag == "env:prod" }
-            .allMatch { it.logGenerator.loggerName == SDK_LOGGER_NAME }
-            .allMatch { it.logGenerator.userInfoProvider is NoOpUserInfoProvider }
+        assertThat(handler).isInstanceOf(LogcatLogHandler::class.java)
+        assertThat((handler as LogcatLogHandler).serviceName)
```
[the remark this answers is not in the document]
0xnm: Yes, we need to, because `serviceName` is passed in the constructor which is called in the `RuntimeUtils` with the value defined in the `RuntimeUtils` as well.

mariusc83 (approved):

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/log/internal/utils/LogUtils.kt
```diff
@@ -18,3 +20,58 @@ internal fun buildLogDateFormat(): SimpleDateFormat =
...
+
+internal const val ERROR_WITH_TELEMETRY_LEVEL =
+    AndroidLog.ERROR or TELEMETRY_LOG_FLAG
+internal const val WARN_WITH_TELEMETRY_LEVEL =
+    AndroidLog.WARN or TELEMETRY_LOG_FLAG
+internal const val DEBUG_WITH_TELEMETRY_LEVEL =
+    AndroidLog.DEBUG or TELEMETRY_LOG_FLAG
+
+internal fun Logger.errorWithTelemetry(
```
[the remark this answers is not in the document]
0xnm: I've opened a ticket for a better approach and added a TODO here.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/UploadStatus.kt
```diff
@@ -45,13 +46,23 @@ internal enum class UploadStatus(val shouldRetry: Boolean) {
...
                     "the batch was dropped."
-            )
-            HTTP_CLIENT_RATE_LIMITING -> logger.e(
-                "$batchInfo failed because of a request error; we will retry later."
-            )
+                if (sendToTelemetry) {
+                    logger.errorWithTelemetry(message)
+                } else {
+                    logger.e(message)
+                }
```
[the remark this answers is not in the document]
0xnm: In this case if we really send a telemetry, then entry will be logged twice in the logcat. Anyway, new ticket created to change the approach, which shouldn't have this issue.

0xnm:
@xgouchet @mariusc83 I've added 2 important commits: https://github.com/DataDog/dd-sdk-android/pull/880/commits/df323371db2c89b988571aa42f2e61801789e740 and https://github.com/DataDog/dd-sdk-android/pull/880/commits/8c32d9ae10495c370ed2a6840ef21bdb82dfdeab, can you please review them?

xgouchet (approved):

PR #887, opened 2022-03-30
RUMM-2078 Prevent sending negative View durations

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -487,7 +487,7 @@ internal open class RumViewScope(
         }
         attributes.putAll(GlobalRum.globalAttributes)
         version++
-        val updatedDurationNs = event.eventTime.nanoTime - startedNanos
+        val updatedDurationNs = resolveViewDuration(event)
         val context = getRumContext()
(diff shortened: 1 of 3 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -541,6 +541,14 @@ internal open class RumViewScope(
         writer.write(viewEvent)
     }
 
+    private fun resolveViewDuration(event: RumRawEvent): Long {
+        val duration = event.eventTime.nanoTime - startedNanos
+        if (duration <= 0) {
+            devLogger.w(NEGATIVE_DURATION_WARNING_MESSAGE.format(Locale.US, name))
```
mariusc83, line 547: @xgouchet , @0xnm do you think we should add a console Log also in RN ? It will be more much more intrusive there as we will have to persist the start time for each view in a `Map<Key, Long>`.

0xnm: The question is if it is actionable on the RN side. Let's maybe try to find the root cause and if we fail to do so, then we may think about adding it there as well considering it will be more visible for the customer (but also more complex to add at the same time).

0xnm (approved):

xgouchet (approved):

PR #901, opened 2022-04-08
RUMM-2055: Remove view from Global RUM context when it is stopped

0xnm:
### What does this PR do?

This PR makes sure that the view is removed from the global RUM context once stopped.

Additionally, the following was added:

* Action context is also removed. This makes sense as it is `parent-child` relationship, even if action can outlive the view (view may be stopped, but may wait for the action to complete). This may have an impact on the events, which may start when view is active as a consequence of some action and only complete when action completes after view is stopped (so it won't be reference to the action in the global context), but overall this change makes relationship between view and action more strict and will prevent the situations when action has no parent reference.

* Add a check for updating global RUM context - it should contain the reference to the current view. This will prevent the situations when global RUM context is already taken by another view (shouldn't happen with auto-instrumentation we have, but may happen with the manual instrumentation), and also updating the context when actions arrive after view is already stopped.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -21,6 +21,8 @@ import com.datadog.android.core.internal.time.TimeProvider
 import com.datadog.android.core.internal.utils.devLogger
 import com.datadog.android.core.internal.utils.loggableStackTrace
 import com.datadog.android.core.internal.utils.resolveViewUrl
+import com.datadog.android.core.internal.utils.sdkLogger
+import com.datadog.android.log.internal.utils.debugWithTelemetry
 import com.datadog.android.rum.GlobalRum
 import com.datadog.android.rum.RumActionType
 import com.datadog.android.rum.RumAttributes
@@ -203,6 +205,9 @@ internal open class RumViewScope(
         writer: DataWriter<Any>
     ) {
         if (!stopped) {
+            // no need to update RUM Context here erasing current view, because this is called
+            // only with event starting a new view, which itself will update a context
+            // at the construction time
             stopped = true
             sendViewUpdate(event, writer)
             delegateEventToChildren(event, writer)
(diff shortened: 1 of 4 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -217,7 +222,25 @@ internal open class RumViewScope(
...
+            GlobalRum.updateRumContext(
+                getRumContext().copy(
+                    viewType = RumViewType.NONE,
+                    viewId = null,
+                    viewName = null,
+                    viewUrl = null,
+                    actionId = null
+                ),
+                applyOnlyIf = { currentContext ->
+                    if (currentContext.viewId == this.viewId) {
```
mariusc83, line 234: When is this possible ?

0xnm: In the current code it is not possible, because if another view starts before current one is stopped, it will be stopped receiving `StartView` for another view (so next `StopView` for this view won't have an effect), but it is just a safeguard to prevent possible bugs if code is changed and allows to have such situation, because accessing `GlobalRum` is dangerous (and so that we are aware of such situation).

mariusc83: Yup...this is why I asked because I could not find this possible with the current logic. Anyway, is good to have the safeguard.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -719,6 +751,13 @@ internal open class RumViewScope(
         internal const val ACTION_DROPPED_WARNING = "RUM Action (%s on %s) was dropped, because" +
             " another action is still active for the same view"
 
+        internal const val RUM_CONTEXT_UPDATE_IGNORED_AT_STOP_VIEW_MESSAGE =
+            "Trying to update global RUM context when StopView event arrived, but context" +
+                " doesn't reference this view already."
```
mariusc83, line 756: I think instead of `doesn't reference this view already` which is a bit confusing maybe you can say ` that the context doesn't reference this view anymore` or simply `doesn't reference this view`

mariusc83 (changes requested):

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
0xnm: Indeed. I've applied this suggestion.

mariusc83 (approved):

PR #913, opened 2022-04-26
RUMM-2142: Allow global context update if view is stopped in a new session

0xnm:
### What does this PR do?

This change should fix having a telemetry entry saying that current view ID doesn't match view ID in the global RUM context. This may happen if we update the context when new session started - this means we also change view ID for the current view.

This PR fixes that allowing update of the global context in case of the session ID mismatch.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -43,6 +43,7 @@ import java.util.UUID
 import java.util.concurrent.TimeUnit
 import kotlin.math.max
 
+@Suppress("LargeClass")
 internal open class RumViewScope(
     private val parentScope: RumScope,
     key: Any,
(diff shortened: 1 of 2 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -235,13 +235,21 @@ internal open class RumViewScope(
...
                 applyOnlyIf = { currentContext ->
-                    if (currentContext.viewId == this.viewId) {
-                        true
-                    } else {
-                        sdkLogger.debugWithTelemetry(
-                            RUM_CONTEXT_UPDATE_IGNORED_AT_STOP_VIEW_MESSAGE
-                        )
-                        false
+                    when {
+                        currentContext.sessionId != this.sessionId -> {
```
mariusc83, line 239: good catch

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
xgouchet: If the session id is different then it might mean that another view is active, which you could be erasing here. What's your logic for doing this check?

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
0xnm: Normally it shouldn't be another view at this point, because the current one should be stopped first (if we are saying that `stopView` is called). If `stopView` is not called for current view and another view starts (via `startView`), then `stopped` flag will be `true` after `onStartView` with event having key of another view, so we [won't even go in this logical branch](https://github.com/DataDog/dd-sdk-android/blob/75882597ab6c730cb1e98e8d71e1367fb3c6acf4/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L228).

Anyway, such override was a case even before, when `GlobalRum.updateRumContext` was called in `onStopView` without such verification, so at least we are not making things worse.

PR #920, opened 2022-04-29
RUMM-2042: Filter Datadog logs per level

### LINKED ISSUE #847: Add possibility to set minimal log level for DatadogTree
**Is your feature request related to a problem? Please describe.**

We have many Timber Trees in the app, including the standard LogCat one. We use different Log levels, including Verbose which is... verbose by its nature.

Our Non-release builds have LogCat Tree enabled, so we have those Verbose logs. Release trees don't handle Verbose level at all.
It looks like the DatadogTree (or its Logger) is not allowing to set the Log Level

As a result, we have a high-volume Verbose logs sent to Datadog while we are not interested in this info.

**Describe the solution you'd like**
1) Add configuration to set min log level

**Describe alternatives you've considered**
Copypaste DatadogTree and configure it internally - this is my current solution.
Checked out both the Tree and Logger classes, haven't found a better workaround

0xnm:
### What does this PR do?

Closes #847. This PR includes the following:

* Adds a possibility to set up default `logger` with min priority log level set in the `DatadogTree` (Timber integration module).
* Adds a possibility to define minimum log level sent to the Datadog servers in the `Logger.Builder` configuration (main SDK module). Note: `logcat` logs are not affected by this parameter.

### Motivation

This feature is useful when Datadog logger is used in the collaboration with other loggers and there is need for more granular control for Datadog logger to avoid sending unwanted log levels to avoid exploding log volume usage (can be the case of verbose or debug logs).

### DIFF
--- a/dd-sdk-android-timber/src/main/kotlin/com/datadog/android/timber/DatadogTree.kt
+++ b/dd-sdk-android-timber/src/main/kotlin/com/datadog/android/timber/DatadogTree.kt
@@ -19,6 +19,21 @@ class DatadogTree(
 ) :
     Timber.Tree() {
 
+    /**
+     * Creates a [Timber.Tree] with a default [Logger] having a minimum log priority
+     * for Datadog logs set to specified value.
+     *
+     * See [Logger.Builder.setDatadogLogsMinPriority] for details.
+     *
+     * @param minLogPriority Minimum log priority to be sent to the Datadog servers.
+     */
+    constructor(minLogPriority: Int) :
+        this(
+            Logger.Builder()
+                .setDatadogLogsMinPriority(minLogPriority)
+                .build()
+        )
+
     init {
         logger.addTag("android:timber")
     }
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
@@ -33,8 +33,8 @@ import org.json.JSONObject
 /**
  * A class enabling Datadog logging features.
  *
- * It allows you to create a specific context (automatic information, custom attributes, tags) that will be embedded in all
- * logs sent through this logger.
+ * It allows you to create a specific context (automatic information, custom attributes, tags) that
+ * will be embedded in all logs sent through this logger.
  *
  * You can have multiple loggers configured in your application, each with their own settings.
  */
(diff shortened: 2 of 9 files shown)


dd-sdk-android-timber/src/main/kotlin/com/datadog/android/timber/DatadogTree.kt
```diff
@@ -13,9 +13,12 @@ import timber.log.Timber
  * An implementation of a [Timber.Tree], forwarding all logs to the provided [Logger].
  *
  * @param logger the logger to use with Timber.
+ * @param minLogPriority the minimum log priority to proceed with. If priority is below this one,
+ * then log entry will be rejected. Default is -1 (allow all).
  */
 class DatadogTree(
-    private val logger: Logger
+    private val logger: Logger,
+    private val minLogPriority: Int = -1
```
xgouchet, line 21: Is this one really necessary? After all people can set the min priority when building the logger passed in parameter.

dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
```diff
@@ -217,14 +218,25 @@ internal constructor(internal var handler: LogHandler) {
...
             datadogLogsEnabled = enabled
             return this
         }
 
+        /**
+         * Sets a minimum priority for the log to be sent to the Datadog servers. If log priority
+         * is below this one, then it won't be sent. Default value is -1 (allow all).
+         * @param minLogPriority Minimum log priority to be sent to the Datadog servers.
+         */
+        fun setDatadogLogsMinPriority(minLogPriority: Int): Builder {
```
mariusc83, line 235: the way I see it this is something similar to `setVerbosity` feature for our `devLogger` right ? I think it will be really useful for our users.

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
0xnm: yes, indeed. Users may want to see verbose (all) information in the logcat, but send only certain levels to Datadog.

dd-sdk-android-timber/src/main/kotlin/com/datadog/android/timber/DatadogTree.kt
0xnm: Updated my PR, replaced it by a secondary constructor which will create a default logger with min priority set.

xgouchet (approved):

PR #925, opened 2022-05-04
Fix the logic of joining byte arrays when there is no items

0xnm:
### What does this PR do?

Fixes a silly bug: size taken by separators could be negative in case of empty collection (doesn't impact any data upload though).

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/utils/ByteArrayExt.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/utils/ByteArrayExt.kt
@@ -40,10 +40,11 @@ internal fun Collection<ByteArray>.join(
     prefix: ByteArray = ByteArray(0),
     suffix: ByteArray = ByteArray(0)
 ): ByteArray {
-    val result = ByteArray(
-        this.sumOf { it.size } +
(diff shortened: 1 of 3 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/utils/ByteArrayExt.kt
```diff
@@ -42,7 +42,8 @@ internal fun Collection<ByteArray>.join(
 ): ByteArray {
     val result = ByteArray(
         this.sumOf { it.size } +
-            prefix.size + suffix.size + separator.size * (this.size - 1)
+            prefix.size + suffix.size +
+            if (this.isNotEmpty()) separator.size * (this.size - 1) else 0
```
xgouchet, line 46: This begins to be hard to read, it'd be nice to split that into variables, e.g.: 

```
val dataSize = this.sumOf { it.size }
val separatorsSize = if (this.isNotEmpty()) separator.size * (this.size - 1) else 0
val size = dataSize + separatorsSize + prefix.size + suffix.size
val result = ByteArray(size)
```

xgouchet (changes requested):

xgouchet (approved):

mariusc83 (approved):

PR #926, opened 2022-05-05
RUMM-2076 Add the Configuration#setVitalsMonitorUpdateFrequency API

mariusc83:
### What does this PR do?

In this PR we are introduce a new API : `ConfigurationBuilder.setVitalsUpdateFrequency(frequency)` where the frequency is an enum :

```
/** Every 100 milliseconds. */
FREQUENT(100L),

/** Every 500 milliseconds. This is the default frequency. */
AVERAGE(500L),

/** Every 1000 milliseconds. */
RARE(1000L),

/** No data will be sent for mobile vitals. */
NEVER(0)
```
This will allow the client to customise the update frequency for mobile vitals reader or disable this feature.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
@@ -103,7 +103,8 @@ internal constructor(
             val viewTrackingStrategy: ViewTrackingStrategy?,
             val longTaskTrackingStrategy: TrackingStrategy?,
             val rumEventMapper: EventMapper<Any>,
-            val backgroundEventTracking: Boolean
+            val backgroundEventTracking: Boolean,
+            val vitalsMonitorUpdateFrequency: VitalsUpdateFrequency
         ) : Feature()
     }
 
@@ -691,6 +692,17 @@ internal constructor(
                 devLogger.e(ERROR_FEATURE_DISABLED.format(Locale.US, feature.featureName, method))
             }
         }
+
+        /**
+         * Allows to specify the frequency at which to update the mobile vitals
+         * data provided in the RUM [ViewEvent].
+         * @param frequency as [VitalsUpdateFrequency]
+         * @see [VitalsUpdateFrequency]
+         */
+        fun setVitalsUpdateFrequency(frequency: VitalsUpdateFrequency): Builder {
+            rumConfig = rumConfig.copy(vitalsMonitorUpdateFrequency = frequency)
+            return this
+        }
     }
 
     // endregion
@@ -737,7 +749,8 @@ internal constructor(
             viewTrackingStrategy = ActivityViewTrackingStrategy(false),
             longTaskTrackingStrategy = MainLooperLongTaskStrategy(DEFAULT_LONG_TASK_THRESHOLD_MS),
             rumEventMapper = NoOpEventMapper(),
-            backgroundEventTracking = false
+            backgroundEventTracking = false,
+            vitalsMonitorUpdateFrequency = VitalsUpdateFrequency.AVERAGE
         )
 
         internal const val ERROR_FEATURE_DISABLED = "The %s feature has been disabled in your " +
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/VitalsUpdateFrequency.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/VitalsUpdateFrequency.kt
@@ -1,0 +1,27 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core.configuration
+
+/**
+ * Defines the frequency at which mobile vitals monitor updates the data.
+ */
+enum class VitalsUpdateFrequency(
+    internal val periodInMs: Long
+) {
+
+    /** Every 100 milliseconds. */
+    FREQUENT(100L),
+
+    /** Every 500 milliseconds. This is the default frequency. */
+    AVERAGE(500L),
+
+    /** Every 1000 milliseconds. */
+    RARE(1000L),
+
+    /** No data will be sent for mobile vitals. */
+    NEVER(0)
+}
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
@@ -12,12 +12,14 @@ import android.os.Handler
 import android.os.Looper
 import android.view.Choreographer
 import com.datadog.android.core.configuration.Configuration
+import com.datadog.android.core.configuration.VitalsUpdateFrequency
 import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.SdkFeature
 import com.datadog.android.core.internal.event.NoOpEventMapper
 import com.datadog.android.core.internal.net.DataUploader
 import com.datadog.android.core.internal.persistence.PersistenceStrategy
 import com.datadog.android.core.internal.system.StaticAndroidInfoProvider
+import com.datadog.android.core.internal.thread.NoOpScheduledExecutorService
 import com.datadog.android.core.internal.utils.devLogger
 import com.datadog.android.core.internal.utils.executeSafe
 import com.datadog.android.core.internal.utils.scheduleSafe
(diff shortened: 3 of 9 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/VitalsUpdateFrequency.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core.configuration
+
+/**
+ * Defines the frequency at which mobile vitals monitor updates the data.
+ */
+enum class VitalsUpdateFrequency(
+    internal val frequencyInMs: Long
```
xgouchet, line 13: This should be renamed `periodInMs`.

_Frequency is measured in Hertz. The value in ms is the period (the inverse of frequency). E.g. for a period of 100ms, the frequency is 10 Hz._

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -86,7 +89,10 @@ internal object RumFeature : SdkFeature<Any, Configuration.Feature.RUM>() {
         configuration.userActionTrackingStrategy?.let { actionTrackingStrategy = it }
         configuration.longTaskTrackingStrategy?.let { longTaskTrackingStrategy = it }
 
-        initializeVitalMonitors()
+        if (configuration.vitalsMonitorUpdateFrequency != VitalsUpdateFrequency.NEVER) {
+            initializeVitalMonitors(configuration.vitalsMonitorUpdateFrequency.frequencyInMs)
+        }
```
xgouchet, line 94: Two quick things, first of all, it'd make the `onInitialize` method easier to read if you'd call just 

`initializeVitalMonitors(configuration.vitalsMonitorUpdateFrequency)` 

and did the check just there. Also let's only extract the `periodInMs` when it's actually needed

dd-sdk-android/src/test/kotlin/com/datadog/android/core/configuration/ConfigurationBuilderTest.kt
```diff
@@ -1643,6 +1644,22 @@ internal class ConfigurationBuilderTest {
         assertThat(config.logsConfig).isEqualTo(Configuration.DEFAULT_LOGS_CONFIG)
     }
 
+    @Test
+    fun `M use the given frequency W setVitalsMonitorUpdateFrequency`(forge: Forge) {
+        // Given
+        val fakeFrequency = forge.aValueFrom(VitalsUpdateFrequency::class.java)
+
```
xgouchet, line 1651:
```suggestion
    fun `M use the given frequency W setVitalsMonitorUpdateFrequency`(
        @Forgery fakeFrequency: VitalsUpdateFrequency
    ) {
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/RumFeatureTest.kt
```diff
@@ -217,9 +220,17 @@ internal class RumFeatureTest : SdkFeatureTest<Any, Configuration.Feature.RUM, R
     }
 
     @Test
-    fun `𝕄 setup vital monitors 𝕎 initialize()`() {
+    fun `𝕄 setup vital monitors 𝕎 initialize { frequency != NEVER }`(forge: Forge) {
```
xgouchet, line 223:
```
    @ParameterizedTest
    @EnumSource(VitalsUpdateFrequency::class, names = ["NEVER"], mode = EnumSource.Mode.EXCLUDE)
    fun `𝕄 setup vital monitors 𝕎 initialize { frequency != NEVER }`(fakeFrequency: VitalsUpdateFrequency) {
        …
    )
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/RumFeatureTest.kt
```diff
@@ -311,15 +371,38 @@ internal class RumFeatureTest : SdkFeatureTest<Any, Configuration.Feature.RUM, R
     }
 
     @Test
-    fun `𝕄 initialize vital executor 𝕎 initialize()`() {
+    fun `𝕄 initialize vital executor 𝕎 initialize { frequency != NEVER }()`(forge: Forge) {
```
xgouchet, line 374:
```
    @ParameterizedTest
    @EnumSource(VitalsUpdateFrequency::class, names = ["NEVER"], mode = EnumSource.Mode.EXCLUDE)
    fun `𝕄 initialize vital executor 𝕎 initialize { frequency != NEVER }`(fakeFrequency: VitalsUpdateFrequency) {
        …
    )
```

xgouchet (changes requested):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/RumFeatureTest.kt
mariusc83: Yeah...make sense to have a parameterized test here instead.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
mariusc83: I had it like that in the first place, honestly was not sure which way would be better as doing the check in the method means a premature return, sometimes I like to avoid this. But yeah, I think it looks cleaner like that.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -737,7 +749,8 @@ internal constructor(
             viewTrackingStrategy = ActivityViewTrackingStrategy(false),
             longTaskTrackingStrategy = MainLooperLongTaskStrategy(DEFAULT_LONG_TASK_THRESHOLD_MS),
             rumEventMapper = NoOpEventMapper(),
-            backgroundEventTracking = false
+            backgroundEventTracking = false,
+            vitalsMonitorUpdateFrequency = VitalsUpdateFrequency.AVERAGE
```
0xnm, line 753: so we are moving from 100 ms to 500 ms by default?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -68,7 +71,7 @@ internal object RumFeature : SdkFeature<Any, Configuration.Feature.RUM>() {
 
     internal var debugActivityLifecycleListener: Application.ActivityLifecycleCallbacks? = null
 
-    internal lateinit var vitalExecutorService: ScheduledThreadPoolExecutor
+    internal var vitalExecutorService: ScheduledExecutorService = NoOpScheduledExecutorService()
```
0xnm, line 74: do we introduce `NoOpScheduledExecutorService` because of `ProhibitLeavingStaticMocksIn`? If yes, then normally `lateinit` should be skipped by this check and no change is needed.

mariusc83: Not because of that, it is just because is more readable this way otherwise I would have to make it an optional and will have then to check it everywhere. Forgot to remove the late init 👍

dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
mariusc83: Yeah, that was the agreement :)

0xnm (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -691,6 +692,17 @@ internal constructor(
...
             }
         }
+
+        /**
+         * Allows to specify the frequency at which to update the mobile vitals
+         * data provided in the RUM [ViewEvent].
+         * @param frequency as [VitalsUpdateFrequency]
+         * @see [VitalsUpdateFrequency]
+         */
+        fun setVitalsUpdateFrequency(frequency: VitalsUpdateFrequency): Builder {
```
0xnm, line 702: We need to add the E2E test for this one.

0xnm (commented):
Friday: I approved, but them I realized that we have no E2E test added for this new API.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
mariusc83: Good point, I will add a task for this.

xgouchet (changes requested):

xgouchet (approved):

PR #934, opened 2022-05-19
RUMM-2165: Add trace sampling for instrumented network requests

0xnm:
### What does this PR do?

This change adds the trace sampling for auto-instrumented network requests (instrumentation via `TracingInterceptor`, `RumInterceptor` and `DatadogInterceptor`) by adding `x-datadog-sampling-priority` header to the outgoing requests which may have a value of `1` (keep) or `0` (drop). This header will later be parsed on the backend side and sampling decision will be respected for the services downstream.

Sampler will make a decision if request should be traced (by injecting tracing headers) or not. This will allow customer to control amount of traces created directly (via `TracingInterceptor`) or indirectly (via RUM2APM integration in case of `DatadogInterceptor` or `RumInterceptor`), which is necessary to align with a new billing model. Default trace sampling is set to 20%.

The sampling is applied **only** to the tracing part of the tracking, meaning that if, for example, trace sampling rate is 75% and 100 requests were made, in case of `DatadogInterceptor` or `RumInterceptor` it will be 100 RUM resources created, but only 75 spans roughly.

Also this change removes `x-datadog-sampled` header, which is not really used.

Generally, we can add the sampling priority information to the span in the following way:

* If span is `MutableSpan`, we can use `MutableSpan#setSamplingPriority`
* We can also use `Span#setTag` with values `Tags.MANUAL_KEEP` (or `MANUAL_DROP`)

The problem is that in the former case `MutableSpan` exists only for spans created by Datadog tracer, but we need also support non-Datadog tracers, so we cannot rely on this interface. In the latter case there is no way to read a tag from the span and it is also not available in the `SpanContext` interface.

So in this implementation we have to access `x-datadog-sampling-priority` header directly for reading/writing sampling decision instead of relying on the `Span` (and we need to add it explicitly anyway even if non-Datadog tracer is used, because in the end data is sent to the Datadog servers anyway).

Also this PR includes commit https://github.com/DataDog/dd-sdk-android/commit/2a08eaebd70e2b0b7159db0dd4c00ab4a4286c15 which is optional and can be discussed - it just makes chained interceptors respect sampling decision for the request which was made in the upstream interceptor.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/TracingInterceptor.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/TracingInterceptor.kt
@@ -6,9 +6,13 @@
 
 package com.datadog.android.tracing
 
+import androidx.annotation.FloatRange
 import com.datadog.android.DatadogInterceptor
+import com.datadog.android.core.configuration.Configuration
 import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.net.FirstPartyHostDetector
+import com.datadog.android.core.internal.sampling.RateBasedSampler
+import com.datadog.android.core.internal.sampling.Sampler
 import com.datadog.android.core.internal.utils.devLogger
 import com.datadog.android.core.internal.utils.loggableStackTrace
 import com.datadog.android.core.internal.utils.sdkLogger
@@ -56,8 +60,9 @@ import okhttp3.Response
  * ```
  *
  * @param tracedHosts a list of all the hosts that you want to be automatically tracked
- * by this interceptor. If no host is provided the interceptor won't trace any OkHttpRequest,
- * nor propagate tracing information to the backend.
+ * by this interceptor. If no host is provided (via this argument or global
+ * configuration [Configuration.Builder.setFirstPartyHosts]) the interceptor won't trace
+ * any [okhttp3.Request], nor propagate tracing information to the backend.
  * @param tracedRequestListener a listener for automatically created [Span]s
  *
  */
@@ -68,6 +73,7 @@ internal constructor(
     internal val tracedRequestListener: TracedRequestListener,
     internal val firstPartyHostDetector: FirstPartyHostDetector,
     internal val traceOrigin: String?,
+    internal val traceSampler: Sampler,
     internal val localTracerFactory: () -> Tracer
 ) : Interceptor {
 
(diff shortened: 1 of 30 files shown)


alai97 (approved):
Looks good for docs!

dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/TracingInterceptor.kt
```diff
@@ -271,59 +293,78 @@ internal constructor(
...
+                SAMPLING_PRIORITY_HEADER,
+                TRACE_ID_HEADER,
+                SPAN_ID_HEADER
+            ).forEach {
+                tracedRequestBuilder.removeHeader(it)
             }
-        )
+            tracedRequestBuilder.addHeader(SAMPLING_PRIORITY_HEADER, DROP_SAMPLING_DECISION)
+        } else {
+            tracer.inject(
```
mariusc83, line 320: shouldn't you add the `SAMPLING_PRIORITY_HEADER` here with `1` ? Otherwise when you check for it in the `TracingInterceptor` the way I've seen you are going to apply a sampling there which is not wanted if the request was not sampled out here.

0xnm: Good question :) I was thinking about it, but let it as it is now. So normally the expectation is that `Injector` will add it, because this is what `DatadogHttpCodec` is doing here https://github.com/DataDog/dd-sdk-android/blob/c05f878ec4d84c65d70aedacb20c6bce6cc9c9ba/dd-sdk-android/src/main/java/com/datadog/opentracing/propagation/DatadogHttpCodec.java#L54.

We could check if header is there after the injection and if it is not, we could add it, but the thing is that `Request.Builder` has no method to check if header is already there, and calling `removeHeader` + `addHeader` (because simple `addHeader` just adds it and not replacing if there is an existing one) can set a header with a value different from injector. But anyway, with `DatadogHttpCodec` everything should be fine.

Now the question is: what if customer is using non-Datadog tracer (generally speaking it is possible, because we accept high-level `io.opentracing.Tracer` interface, not necessary Datadog tracer. So it won't be `DatadogHttpCodec` there and so maybe `x-datadog-sampling-priority` won't be set in this branch. If it is not set, then simply on the backend side service will make its own decision about the sampling, so not a big deal.

And since we are accepting `io.opentracing.Tracer` and `io.opentracing.Span` interface which generally **have** a way to set priority sampling, but there is no way to read it, we have to rely on the headers with such kind of hacky decision.

We could assume that Datadog tracer is always used (and then use `DDSpan`, which have an opportunity to read back the sampling priority) and not rely on the headers, but then `traceSamplingRate` parameter would be useless if customer is using non-Datadog tracer and so they will face the new billing model in a not very nice way.

xgouchet: This is done automatically by the inject method

xgouchet (approved):

mariusc83 (approved):

PR #936, opened 2022-05-23
RUMM-2202 Ensure thread safety

xgouchet:
### What does this PR do?

- adds a Detekt rule to verify the call compatibility for methods annotated with `@WorkerThread`/ `@UiThread`…
- Start flagging internal methods with the relevant annotations


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/BatchConfirmation.kt
```diff
@@ -6,6 +6,9 @@
 
 package com.datadog.android.v2.core.internal.storage
 
-internal fun interface BatchConfirmation {
```
0xnm, line 9: are we removing functional interface property here (and in the other interfaces), because it is complicated to get annotation in case of lamba implementation of the functional interface?

0xnm (approved):
LGTM! I've left a suggestion in the test and a question regarding the TODO item, but overall it is non-blocking.

mariusc83 (approved):
LGTM

PR #943, opened 2022-05-31
RUMM-2186: Write batch metadata

0xnm:
### What does this PR do?

This change adds support for reading/writing optional batch metadata (in a non-typed format, currently as `ByteArray`).

Batch metadata file will be written to the batch data file and will be associated with it by the name (metadata file will have the same name as batch file, just with `_metadata` suffix).

This change also does the following:

* Removes the dependency on the event size in the `FileOrchestrator` to prevent the possibility of the switching to the next batch file if current batch size + event size exceeds the limit (we can do this, because the limit set in the SDK is still less than the actual limit of the intake(s) and the margin we have is still bigger than maximum event size). This makes us to keep batch metadata file and batch file in sync for the particular `BatchWriter` (otherwise we could have a metadata file from another batch). Responsibility of checking event size is now shifted to the `Writer` layer.
* Temporarily separates classes/interfaces for writing/reading batch files and simple files (with only 1 item, or without the need of metadata included), because return type of the read and file format are different now. This is not a final state (set of classes, content, names) and better rework will be done in RUMM-2235.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileDataWriter.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileDataWriter.kt
@@ -10,10 +10,12 @@ import androidx.annotation.WorkerThread
 import com.datadog.android.core.internal.persistence.DataWriter
 import com.datadog.android.core.internal.persistence.PayloadDecoration
 import com.datadog.android.core.internal.persistence.Serializer
-import com.datadog.android.core.internal.persistence.file.FileHandler
+import com.datadog.android.core.internal.persistence.file.ChunkedFileHandler
 import com.datadog.android.core.internal.persistence.file.FileOrchestrator
+import com.datadog.android.core.internal.persistence.file.FilePersistenceConfig
 import com.datadog.android.core.internal.persistence.serializeToByteArray
 import com.datadog.android.log.Logger
+import java.util.Locale
 
 /**
  * A [DataWriter] storing data in batch files.
@@ -22,8 +24,10 @@ internal open class BatchFileDataWriter<T : Any>(
     internal val fileOrchestrator: FileOrchestrator,
     internal val serializer: Serializer<T>,
     internal val decoration: PayloadDecoration,
-    internal val handler: FileHandler,
-    internal val internalLogger: Logger
+    internal val handler: ChunkedFileHandler,
+    internal val internalLogger: Logger,
+    // TODO RUMM-0000 don't use default value
+    internal val filePersistenceConfig: FilePersistenceConfig = FilePersistenceConfig()
 ) : DataWriter<T> {
 
     // region DataWriter
@@ -76,9 +80,28 @@ internal open class BatchFileDataWriter<T : Any>(
 
     @WorkerThread
     private fun writeData(byteArray: ByteArray): Boolean {
-        val file = fileOrchestrator.getWritableFile(byteArray.size) ?: return false
+        if (!checkEventSize(byteArray.size)) return false
+
+        val file = fileOrchestrator.getWritableFile() ?: return false
         return handler.writeData(file, byteArray, true)
+    }
+    private fun checkEventSize(eventSize: Int): Boolean {
+        if (eventSize > filePersistenceConfig.maxItemSize) {
+            internalLogger.e(
+                ERROR_LARGE_DATA.format(
+                    Locale.US,
+                    eventSize,
+                    filePersistenceConfig.maxItemSize
+                )
+            )
+            return false
+        }
+        return true
     }
 
     // endregion
+
+    companion object {
+        internal const val ERROR_LARGE_DATA = "Can't write data with size %d (max item size is %d)"
+    }
 }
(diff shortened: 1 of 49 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileDataWriter.kt
```diff
@@ -76,9 +80,28 @@ internal open class BatchFileDataWriter<T : Any>(
 
     @WorkerThread
     private fun writeData(byteArray: ByteArray): Boolean {
-        val file = fileOrchestrator.getWritableFile(byteArray.size) ?: return false
+        if (!checkEventSize(byteArray.size)) return false
```
mariusc83, line 83: Means that this event will be lost ? I wonder if this will create problems for SessionReplay records 🤔 , I guess they are not that big though.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/single/SingleItemDataWriter.kt
```diff
@@ -9,16 +9,21 @@ package com.datadog.android.core.internal.persistence.file.single
...
 import com.datadog.android.log.Logger
+import com.datadog.android.log.internal.utils.errorWithTelemetry
+import java.util.Locale
 
 internal open class SingleItemDataWriter<T : Any>(
     internal val fileOrchestrator: FileOrchestrator,
     internal val serializer: Serializer<T>,
-    internal val handler: FileHandler,
-    internal val internalLogger: Logger
+    internal val handler: ChunkedFileHandler,
```
mariusc83, line 23: shouldn't be a `SingleItemFileHandler` here ?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/ConsentAwareStorage.kt
```diff
@@ -41,24 +49,34 @@ internal class ConsentAwareStorage(
...
-            override fun write(event: ByteArray, eventId: String, newMetadata: ByteArray) {
+            override fun write(event: ByteArray, eventId: String, newMetadata: ByteArray?) {
                 // prevent useless operation for empty event / null orchestrator
                 if (event.isEmpty() || orchestrator == null) {
                     listener.onDataWritten(eventId)
                     return
                 }
 
-                val file = orchestrator.getWritableFile(event.size)
+                if (!checkEventSize(event.size)) {
```
mariusc83, line 71: Isn't this already checked in the `FileHandler` ?

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileDataWriter.kt
0xnm: Yes, it will be lost. But this behavior is not different from the current one. Currently if item is too big we won't provide any file for writing it https://github.com/DataDog/dd-sdk-android/blob/e5ddff44deff481cebafcb6b735938fb9fe04998/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt#L48-L57

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/ConsentAwareStorage.kt
0xnm: It was moved from `FileOrchestrator` to the `DataWriter`, but `DataWriter` is not used by the `ConsentAwareStorage` right now (and probably won't, in fact in the writer callback we need only `Serializer`, so logically `Storage` replaces `DataWriter`), so adding this check here as well.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/single/PlainFileHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+import java.io.FileNotFoundException
+import java.io.FileOutputStream
+import java.io.IOException
+import java.util.Locale
+
+// TODO RUMM-2235 Rework file persistence classes/interfaces
+// This file may go away after rework, not adding tests for now (it is a simplified copy of
+// BatchFileHandler)
+@Suppress("unused")
+internal class PlainFileHandler(
```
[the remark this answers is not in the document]
0xnm: Probably yes, but it will be a subject of RUMM-2235 anyway.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/single/SingleItemDataWriter.kt
0xnm: Ideally, but it will be a subject of RUMM-2235 anyway. Currently the only implementation of `ChunkedFileHandler` is `BatchFileHandler` which will write data in TLV format adding meta as well, which are (both meta and TLV) not really needed for the use-cases of `SingleItemDataWriter` which are `NdkUserInfoDataWriter` and `NdkNetworkInfoDataWriter` (since we just want to store plain json), but these cases will work with `BatchFileHandler` as well.

xgouchet (approved):

mariusc83 (approved):

PR #947, opened 2022-06-13
RUMM-2235: Rework file persistence layer

0xnm:
### What does this PR do?

This change makes some re-work of the file persistence layer to answer the new requirements:

* We have now 2 file formats to maintain: TLV format with meta+data types used for events storage and plain file format used for metadata storage. Reading is also different for these 2: either we return single data chunk, or multiple data chunks.
* File move/delete operations shouldn't depend on the file content/format.
* Encryption should be supported by both file formats as well.

To support these requirements the following was done in this PR:

* Moved out file move/delete logic into a dedicated class `FileMover` (not the best name, I know, but `FileOrchestrator` was already taken).
* Introduced different interfaces for file writing (the same interface for both file formats) and file reading (2 different interfaces, because Kotlin/Java doesn't support method overload for different return types).
* Moved metadata file handling responsibility into `FileOrchestrator` - this gives an advantage for the proper file cleanup once batch is dropped, expired or simply disk space should be freed.
* Introduced a dedicated class for writing data in the plain format (for metadata cases).


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -93,6 +94,28 @@ internal class BatchFileOrchestrator(
...
     }
 
+    @WorkerThread
+    override fun getMetadataFile(file: File): File? {
+        if (file.parent != rootDir.path) {
+            // may happen if batch file was requested with pending orchestrator, but meta file
+            // is requested with granted orchestrator (due to consent change). Not an issue, because
+            // batch file should be migrated to the same folder, but leaving this debug point
+            // just in case.
+            internalLogger.debugWithTelemetry(
```
xgouchet, line 104: Nice 👍

xgouchet (approved):
LGTM

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataFlusher.kt
```diff
@@ -9,28 +9,35 @@ package com.datadog.android.core.internal.data.upload
...
+            val batch = fileReader.readData(it)
                 .join(
                     separator = decoration.separatorBytes,
                     prefix = decoration.prefixBytes,
                     suffix = decoration.suffixBytes
                 )
             uploader.upload(batch)
-            handler.delete(it)
+            fileMover.delete(it)
+            val metaFile = fileOrchestrator.getMetadataFile(it)
```
mariusc83, line 37: where is this data uploaded if you delete it here ?

0xnm: The data is uploaded few lines above (`uploader.upload(batch)`), basically the logic is the same, I just added metadata deletion.

But I see your concern: metadata is requested for the file which is deleted and while it is ok to have (`File` object stays there still and we just use its path to get associated metadata file), I swapped the lines - now metadata file is requested before batch file deletion, which makes the operation more clear.

mariusc83 (approved):

PR #956, opened 2022-06-22
RUMM-2208: SDK v2 upload pipeline

0xnm:
### What does this PR do?

This change brings the foundation of the upload pipeline for SDK v2 which is using a single `Storage` concept introduced before (meaning pipeline also leverages batch metadata).

In the nutshell upload job will read the batch available for the upload via `Storage#readNextBatch` call and will pass the batch and its metadata (if exists) to the uploader, which knows a) how to create a request (responsibility of the feature); b) how to execute it (responsibility of the SDK).

Most of the changes are done in the `UploadWorker`, `DataOkHttpUploader`, `DataUploadRunnable` classes (the last 2 are the copies of the v1 classes residing in the v2 package, this is done in order to avoid big changes in the `SdkFeature`, which will be done later).

Note, that this change doesn't yet support `1:many` relationship between batch and requests created, this will be done in another task(s).

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadWorker.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadWorker.kt
@@ -11,13 +11,13 @@ import androidx.annotation.WorkerThread
 import androidx.work.Worker
 import androidx.work.WorkerParameters
 import com.datadog.android.Datadog
-import com.datadog.android.core.internal.net.DataUploader
 import com.datadog.android.core.internal.net.UploadStatus
-import com.datadog.android.core.internal.persistence.Batch
-import com.datadog.android.core.internal.persistence.DataReader
 import com.datadog.android.core.internal.utils.devLogger
-import com.datadog.android.core.internal.utils.sdkLogger
+import com.datadog.android.v2.api.context.DatadogContext
 import com.datadog.android.v2.core.DatadogCore
+import com.datadog.android.v2.core.DatadogFeature
+import com.datadog.android.v2.core.internal.net.DataUploader
+import com.datadog.android.v2.core.internal.storage.Storage
 
 internal class UploadWorker(
     appContext: Context,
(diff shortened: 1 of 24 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
```diff
@@ -61,6 +62,9 @@ class DatadogCore(
     internal var webViewLogsFeature: SdkFeature<JsonObject, Configuration.Feature.Logs>? = null
     internal var webViewRumFeature: SdkFeature<Any, Configuration.Feature.RUM>? = null
 
+    // TODO RUMM-0000 handle context
```
mariusc83, line 65: RUM-0000 ?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
```diff
@@ -143,6 +147,15 @@ class DatadogCore(
         webViewRumFeature?.flushStoredData()
     }
 
+    /**
+     * Returns all registered features.
+     */
+    fun getAllFeatures(): List<FeatureScope> {
+        // TODO-2138
+        // should it be a part of SDKCore?
+        return emptyList()
```
mariusc83, line 156: IMO it can be

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadWorker.kt
```diff
@@ -33,46 +33,49 @@ internal class UploadWorker(
...
+        // context is unique for each batch query instead of using the same one for all the batches
+        // which will be uploaded, because it can change by the time the upload of the next batch
+        // is requested.
+        val context = datadogCore.context ?: return
+
+        storage.readNextBatch(context) { batchId, reader ->
+            val batch = reader.read()
+            val batchMeta = reader.currentMetadata()
+
+            val success = consumeBatch(context, batch, batchMeta, uploader)
```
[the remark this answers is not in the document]
0xnm: indeed, it can be inlined now

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
0xnm: yes, it is used like that in some place in the codebase, for the things which have task missing. Not sure where context will be handled, maybe in some existing task.

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
0xnm: I didn't pull it up to the `SDKCore` interface definition, because the question is why would some feature request all registered features, shouldn't we limit interface API only to getting explicit features by names? Otherwise too much power is exposed.

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/FeatureUploadConfiguration.kt
```diff
@@ -9,10 +9,8 @@ package com.datadog.android.v2.api
...
  * Contains the upload configuration for an [FeatureScope] instance.
  *
- * @property endpointUrl the url endpoint data should be uploaded to
- * @property payloadFormat the expected format of the payload
+ * @property requestFactory creates a request from a given batch and its metadata
  */
 data class FeatureUploadConfiguration(
-    val endpointUrl: String,
-    val payloadFormat: PayloadFormat
+    val requestFactory: RequestFactory
```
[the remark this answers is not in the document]
ncreated: Yeah, even though it will be empty for now (I don't foresee any instant need) I think it makes sense as a convention. It coherently separates "storage" and "upload" configurations which is one of the goals in V2. While the `requestFactory` answers "what" to upload, we will need `FeatureUploadConfiguration` as soon as we run into "how" (e.g. how frequent or what timeout, like @0xnm said).

I think we will keep it on iOS as it adds good verbosity in code and tests (e.g. mocking `.noOp()` upload configuration will make it clear that the uploader is muted).

xgouchet (approved):

PR #960, opened 2022-06-27
Custom battery intent handler

BobPfingsten:
If a custom build of AOSP does not support or report the battery state for the device through the BatteryIntent broadcast logging is delayed or skipped.  When this scenario happens there is no work around.

For example when a battery status is dumped from the ADB shell that reports something similar to the following:

dumpsys battery >>>
Current Battery Service state:
  AC powered: false
  USB powered: false
  Wireless powered: false
  Max charging current: 0
  Max charging voltage: 0
  Charge counter: 0
  status: 1
  health: 1
  present: false
  level: 0
  scale: 100
  voltage: 0
  temperature: 0
  technology:

This will prevent the SDK from syncing the logs back to the server.  The only work around is to update the AOSP build which in some instances cannot be done in a timely manner.

To support custom builds that may not report battery stats there needs to be a way to provide a custom handler in BroadcastReceiverSystemInfoProvider for the BatteryIntent instead of only allowing the internal logic.  

To facilitate user logic I propose the following:

1. Create an interface BatteryIntentHandler with a method for processing and returning the results of the BatteryIntent.
2. Represent the results as BatterIntentHandlerResults that has values for full or charging (bool), charge level (int), and external power applied(bool).
3. Move the existing logic into a class called DefaultBatteryIntentHandler
4. Modify the BroadcastReceiverSystemInfoProvider constructor to accept the BatteryIntentHandler with a default value set to the DefaultBatteryIntentHandler.
5. In Configuration.Builder add new setter to pass in the BatteryIntentHandler: setBatteryIntentHandler(handler: BatteryIntentHandler)
6. In CoreFeature.setupInfoProviders use the BatteryIntentHandler if specified in the Configuration.Builder otherwise use the default constructor for BroadcastReceiverSystemInfoProvider.

The system can then be initialized like this:

``` 
var configuration = Configuration.Builder(
            logsEnabled = true,
            tracesEnabled = true,
            crashReportsEnabled = true,
            rumEnabled = true
        )
        if (NeedsCustomLogic) {
            configuration.setBatteryIntentHandler(object : BatteryIntentHandler {
                override fun handleBatteryIntent(intent: Intent): BatteryIntentResults {
                    return BatteryIntentResults(true, 100, true)
                }
            })
        }
        Datadog.initialize(
            context = this,
            credentials = Credentials(
                clientToken = "DD_CLIENT_TOKEN",
                envName = "environment",
                variant = "variant",
                rumApplicationId = "123",
                serviceName = "sample"
            ),
            configuration = configuration.build(),
            trackingConsent = TrackingConsent.GRANTED
        )
```

### DIFF
--- a/dd-sdk-android/apiSurface
+++ b/dd-sdk-android/apiSurface
@@ -78,6 +78,7 @@ data class com.datadog.android.core.configuration.Configuration
     constructor(Boolean, Boolean, Boolean, Boolean)
     fun build(): Configuration
     fun setUseDeveloperModeWhenDebuggable(Boolean): Builder
+    fun setBatteryIntentHandler(com.datadog.android.plugin.BatteryIntentHandler)
     fun setFirstPartyHosts(List<String>): Builder
     fun setWebViewTrackingHosts(List<String>): Builder
     fun useSite(com.datadog.android.DatadogSite): Builder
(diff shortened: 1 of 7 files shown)


xgouchet:
Hi @BobPfingsten and thanks for taking the time to submit this review. I'm going to expand more on this in the issue itself, but this approach wouldn't fit our SDK, as it's basically just a way to disable the battery check altogether, and this could be misused to ignore the battery level / powersave mode for end users.

PR #976, opened 2022-07-08
RUMM-2315 Add support to Json Schema's oneOf syntax

xgouchet:
### What does this PR do?

This adds support to the `oneOf` option of JsonSchema which maps to a sealed class.

### Additional Notes

Did a bit of refactoring and ensured the tests ran on all types.

- [X] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](CONTRIBUTING.md) doc)


mariusc83 (approved):
LGTM

buildSrc/src/test/kotlin/com/datadog/android/core/internal/utils/MiscUtils.kt
```diff
@@ -4,7 +4,7 @@
  * Copyright 2016-Present Datadog, Inc.
  */
 
-package com.example.model
+package  com.datadog.android.core.`internal`.utils
```
0xnm, line 7:
```suggestion
package com.datadog.android.core.internal.utils
```

0xnm (commented):
nice work! I've left few questions, but overall looks good.

buildSrc/src/test/kotlin/com/datadog/android/core/internal/utils/MiscUtils.kt
mariusc83: Sorry I need to get back to this as I tested it based on my schemas and we are missing some things:

Given this example:

```
{
    "$schema": "http://json-schema.org/draft-07/schema",
    "$id": "record-schema.json",
    "title": "Record",
    "type": "object",
    "description": "Schema of a Session Replay Record.",
    "oneOf": [
        {
            "$ref": "metadata-record-schema.json"
        },
        {
            "$ref": "focus-record-schema.json"
        },
        {
            "$ref": "view-end-record-schema.json"
        }
    ]
}
```

It will generate the sealed Classes but it will create duplicates for each of the referenced schema in separated classes.

I also tried to reference this `record-schema.json` from an array inside another schema:

```
 "records": {
                    "type": "array",
                    "items": {
                        "$ref": "record-schema.json"
                    },
                    "description": "The records contained by this Segment.",
                    "readOnly": true
                }
```
It crashes in this moment, in this case this array should reference the sealed Class.

mariusc83 (changes requested):
There are some use - cases that are failing now.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/KotlinSpecGenerator.kt
```diff
@@ -0,0 +1,114 @@
...
+                } else {
+                    LIST.parameterizedBy(items.asKotlinTypeName(rootTypeName))
+                }
+            }
+            is TypeDefinition.Enum -> {
+                val def = withUniqueTypeName()
+                nestedTypes.add(def)
+                ClassName(packageName, rootTypeName, def.name)
+            }
+            is TypeDefinition.MultiClass -> {
```
mariusc83, line 70: we are missing the case where the `ref` is defined outside of the current `schema` in a separated one. In that case it needs to be `ClassName(packageName, name)`.
See my example in the comments section of this PR.

mariusc83 (changes requested):

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/MultiClassGenerator.kt
```diff
@@ -0,0 +1,135 @@
...
+                "errors.add(%L)",
+                Identifier.CAUGHT_EXCEPTION
+            )
+            funBuilder.addStatement("null")
+            funBuilder.endControlFlow()
+            options.add(variableName)
+        }
+
+        funBuilder.addStatement(
+            "val result = arrayOf(${options.joinToString(", ")}).firstOrNull { it != null }"
```
mariusc83, line 119: options.joinToString(",\n"), otherwise when you use have a long array it will look ugly.

mariusc83: On my end if I have more than 3 sealed classes there it goes to another line and it breaks the statement when it generates the code. Either we do what I suggested or we find a way to increase Kotlin Poet line length ?

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/JsonSchemaReader.kt
```diff
@@ -193,7 +196,7 @@ class JsonSchemaReader(
     private fun transformEnum(
         typeName: String,
         type: JsonType?,
-        values: List<String>,
+        values: List<String?>,
```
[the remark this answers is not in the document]
xgouchet: Following the [Json Schema RFC](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-01#section-6.1.2) : 
> **6.1.2.**  enum
> 
> The value of this keyword MUST be an array.  This array SHOULD have at least one element.  Elements in the array SHOULD be unique.
>  An instance validates successfully against this keyword if its value is equal to one of the elements in this keyword's array value.
> Elements in the array might be of any type, including null.

Just wanted to be thorough here

buildSrc/src/test/kotlin/com/example/model/Animal.kt
```diff
@@ -0,0 +1,151 @@
...
+            val asBird = try {
+                Bird.fromJson(jsonString)
+            } catch (e: JsonParseException) {
+                errors.add(e)
+                null
+            }
+            val result = arrayOf(asFish, asBird).firstOrNull { it != null }
+            if (result == null) {
+                val message = "Unable to parse json into one of type Animal"
+                throw JsonParseException(message, errors[0])
```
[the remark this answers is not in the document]
xgouchet: Because we can only use one throwable as a cause. But indeed I might write an error message including all errors.

mariusc83 (approved):
Well done !! LGTM now !!

0xnm (approved):
very good job!

PR #980, opened 2022-07-12
RUMM-2322: Report back interaction as back type, prevent reporting empty target name

### LINKED ISSUE #964: Seeing unexpected custom actions "back" and "" reported in the DataDog console
**Describe what happened**

I'm setting up DataDog in an Android project and seeing some unexpected custom events being reported. We are seeing two unknown custom events/actions being reported in our DataDog console. We are not calling `GlobalRum.get().addUserAction(RumActionType.CUSTOM, event, attributes)` with either "back" or "" as parameters:

1. "back": there is a custom event being reported with a name/description of "back" that is reported when the Android system back button is pressed.

2. "": we have also seen a custom event reported with no name/description.

Screenshot included below shows both in the DataDog console, filtered by actions with type `custom`:

Screen Shot 2022-06-29 at 2 47 44 PM

1. System back button is pressed, a custom action/event is reported with name "back"

2. Have not reproduced getting the empty name event, but less important to us as it's only ever occurred 2 times during testing.

Expecting to only see custom events sent when `GlobalRum.get().addUserAction(RumActionType.CUSTOM, event, attributes)` is called. (Or expecting to see back button reported as a TAP action, maybe)

**Additional context**
 - Android OS version: 10
 - Device Model: Samsung SM-G975U1
 - Datadog SDK version: 1.12.0
 - Versions of any other relevant dependencies (OkHttp, …): 
 - Proguard configuration: 
 - Gradle Plugins:

0xnm:
### What does this PR do?

* `back` interaction are reported as `BACK` now instead `CUSTOM`.
* Prevents sending empty target name. It _seems_ it happens because we may start scroll/swipe action with empty target name and without attributes, but `onUp` is not called, because screen is closed while finger is still touching the screen. To avoid that we will report `targetId` and `targetName` for the `startUserAction` as well, so that once it completes due to timeout, it has the necessary data.

Fixes #964.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/WindowCallbackWrapper.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/WindowCallbackWrapper.kt
@@ -130,7 +130,7 @@ internal class WindowCallbackWrapper(
         } else {
             customTargetName
         }
-        GlobalRum.get().addUserAction(RumActionType.CUSTOM, targetName, emptyMap())
+        GlobalRum.get().addUserAction(RumActionType.BACK, targetName, emptyMap())
     }
 
     // endregion
--- a/dd-sdk-android/apiSurface
+++ b/dd-sdk-android/apiSurface
@@ -308,6 +308,7 @@ enum com.datadog.android.rum.RumActionType
   - SCROLL
   - SWIPE
   - CLICK
+  - BACK
   - CUSTOM
 object com.datadog.android.rum.RumAttributes
   const val APPLICATION_VERSION: String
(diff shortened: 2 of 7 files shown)


mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/WindowCallbackWrapper.kt
```diff
@@ -130,7 +130,7 @@ internal class WindowCallbackWrapper(
         } else {
             customTargetName
         }
-        GlobalRum.get().addUserAction(RumActionType.CUSTOM, targetName, emptyMap())
+        GlobalRum.get().addUserAction(RumActionType.TAP, targetName, emptyMap())
```
xgouchet, line 133: This interaction is `handleBackEvent`, and there's a BACK type for actions, maybe we should use that instead as it more precise

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/WindowCallbackWrapper.kt
0xnm: thanks! I didn't know `back` type exists in the RUM schema. I've updated my PR.

xgouchet (approved):

PR #981, opened 2022-07-12
RUMM-2324: Let user to disable RUM view tracking strategy

0xnm:
### What does this PR do?

This change allows user to disable the default view tracking strategy if there is a need/wish to do view tracking manually by using `RumMonitor` on the user side.

This is achieved by letting call `Configuration#useViewTrackingStrategy` with `null` argument, which will make SDK use `NoOpViewTrackingStrategy`.

### DIFF
--- a/dd-sdk-android/apiSurface
+++ b/dd-sdk-android/apiSurface
@@ -78,7 +78,7 @@ data class com.datadog.android.core.configuration.Configuration
     fun useCustomRumEndpoint(String): Builder
     fun trackInteractions(Array<com.datadog.android.rum.tracking.ViewAttributesProvider> = emptyArray(), com.datadog.android.rum.tracking.InteractionPredicate = NoOpInteractionPredicate()): Builder
     fun trackLongTasks(Long = DEFAULT_LONG_TASK_THRESHOLD_MS): Builder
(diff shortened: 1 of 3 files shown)


dd-sdk-android/apiSurface
```diff
@@ -78,7 +78,7 @@ data class com.datadog.android.core.configuration.Configuration
     fun useCustomRumEndpoint(String): Builder
     fun trackInteractions(Array<com.datadog.android.rum.tracking.ViewAttributesProvider> = emptyArray(), com.datadog.android.rum.tracking.InteractionPredicate = NoOpInteractionPredicate()): Builder
     fun trackLongTasks(Long = DEFAULT_LONG_TASK_THRESHOLD_MS): Builder
-    fun useViewTrackingStrategy(com.datadog.android.rum.tracking.ViewTrackingStrategy): Builder
+    fun useViewTrackingStrategy(com.datadog.android.rum.tracking.ViewTrackingStrategy?): Builder
```
mariusc83, line 81: should we do this or should we have it null by default in the configuration and have this method as before ?

0xnm: I think we should have it `ActivityViewTrackingStrategy` in the configuration by default, to ease onboarding.

xgouchet (approved):
👍

PR #983, opened 2022-07-12
Expose Telemetry and addLongTask for internal use

fuzzybinary:
### What does this PR do?

This exposes some internal Datadog classes and methods for use by cross platform frameworks.

We expose internal methods through the _InternalProxy class, which is documented for internal use, and the `_internal` static member on the Datadog object. We keep Telemetry and RUM as separate proxies, in case we have to split them as v2 takes shape.

As part of exposing telemetry I also needed a way to report errors that are not Throwable, so I modified the event to replace the `throwable` member with `stack` and `kind`, and moved the creation of those strings farther up the chain (they were created on event write prior).

### Additional Notes

Please double check all my logic on the switch from `throwable` to `stack` and `kind`

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -19,13 +19,13 @@ import com.datadog.android.core.internal.lifecycle.ProcessLifecycleCallback
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleMonitor
 import com.datadog.android.core.internal.utils.devLogger
 import com.datadog.android.core.internal.utils.sdkLogger
+import com.datadog.android.core.internal.utils.telemetry
 import com.datadog.android.core.model.UserInfo
 import com.datadog.android.error.internal.CrashReportsFeature
 import com.datadog.android.log.internal.LogsFeature
 import com.datadog.android.privacy.TrackingConsent
 import com.datadog.android.rum.GlobalRum
 import com.datadog.android.rum.internal.RumFeature
-import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
 import com.datadog.android.rum.internal.monitor.DatadogRumMonitor
 import com.datadog.android.tracing.internal.TracingFeature
 import com.datadog.android.webview.internal.log.WebViewLogsFeature
@@ -289,13 +289,9 @@ object Datadog {
      *
      * @see _InternalProxy
      */
-    @JvmStatic
     @Suppress("ObjectPropertyNaming")
-    val _internal: _InternalProxy by lazy {
-        _InternalProxy(
-            GlobalRum.get() as? AdvancedRumMonitor
-        )
-    }
+    var _internal: _InternalProxy = _InternalProxy(telemetry)
+        private set
 
     // endregion
 
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -290,6 +290,7 @@ object Datadog {
      * @see _InternalProxy
      */
     @JvmStatic
+    @Suppress("ObjectPropertyNaming")
     val _internal: _InternalProxy by lazy {
         _InternalProxy(
             GlobalRum.get() as? AdvancedRumMonitor
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -25,6 +25,7 @@ import com.datadog.android.log.internal.LogsFeature
 import com.datadog.android.privacy.TrackingConsent
 import com.datadog.android.rum.GlobalRum
 import com.datadog.android.rum.internal.RumFeature
+import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
 import com.datadog.android.rum.internal.monitor.DatadogRumMonitor
 import com.datadog.android.tracing.internal.TracingFeature
 import com.datadog.android.webview.internal.log.WebViewLogsFeature
@@ -281,6 +282,18 @@ object Datadog {
         } else {
             RumFeature.disableDebugging()
         }
+    }
+
+    /**
+     * For Datadog internal use only.
+     *
+     * @see _InternalProxy
+     */
+    @JvmStatic
+    val _internal: _InternalProxy by lazy {
+        _InternalProxy(
+            GlobalRum.get() as? AdvancedRumMonitor
+        )
     }
 
     // endregion
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/_InternalProxy.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/_InternalProxy.kt
@@ -19,6 +19,13 @@ import com.datadog.android.telemetry.internal.Telemetry
  * Methods, members, and functionality of this class  are subject to change without notice, as they
  * are not considered part of the public interface of the Datadog SDK.
  */
+@Suppress(
+    "UndocumentedPublicClass",
+    "UndocumentedPublicFunction",
+    "UndocumentedPublicProperty",
+    "ClassNaming",
+    "VariableNaming"
+)
 class _InternalProxy {
     class _TelemetryProxy {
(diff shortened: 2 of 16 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -325,11 +326,21 @@ internal class DatadogRumMonitor(
...
 
     override fun sendErrorTelemetryEvent(message: String, throwable: Throwable?) {
-        handleEvent(RumRawEvent.SendTelemetry(TelemetryType.ERROR, message, throwable))
+        var stack: String? = null
+        var kind: String? = null
+        throwable?.let {
+            stack = it.loggableStackTrace()
+            kind = it.javaClass.canonicalName ?: it.javaClass.simpleName
+        }
+        handleEvent(RumRawEvent.SendTelemetry(TelemetryType.ERROR, message, stack, kind))
```
xgouchet, line 339:
```suggestion
        val stack: String? = throwable?.loggableStackTrace()
        var kind: String? = throwable?.javaClass?.canonicalName ?: throwable?.javaClass?.simpleName
        handleEvent(RumRawEvent.SendTelemetry(TelemetryType.ERROR, message, stack, kind))
```

dd-sdk-android/src/test/kotlin/com/datadog/android/InternalProxyTest.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    fun `M proxy telemetry to RumMonitor W error()`(
+        @StringForgery message: String,
+        @StringForgery stack: String,
+        @StringForgery kind: String
+    ) {
+        // Given
+        val proxy = _InternalProxy(GlobalRum.get() as? AdvancedRumMonitor)
```
xgouchet, line 73:
```suggestion
        val proxy = _InternalProxy(rumMonitor.mockInstance)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/InternalProxyTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        // Then
+        verify(rumMonitor.mockInstance).addLongTask(time, target)
+    }
+
+    @Test
+    fun `M proxy telemetry to RumMonitor W debug()`(
+        @StringForgery message: String
+    ) {
+        // Given
+        val proxy = _InternalProxy(GlobalRum.get() as? AdvancedRumMonitor)
```
xgouchet, line 57:
```suggestion
        val proxy = _InternalProxy(rumMonitor.mockInstance)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/InternalProxyTest.kt
```diff
@@ -0,0 +1,91 @@
...
+@ForgeConfiguration(Configurator::class)
+internal class InternalProxyTest {
+
+    @Test
+    fun `M proxy addLongTask to RumMonitor W addLongTask()`(
+        @LongForgery time: Long,
+        @StringForgery target: String
+    ) {
+        // Given
+        val proxy = _InternalProxy(GlobalRum.get() as? AdvancedRumMonitor)
```
xgouchet, line 43: You shouldn't rely on static/global states in your tests. 

```suggestion
        val proxy = _InternalProxy(rumMonitor.mockInstance)
```

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -283,6 +284,19 @@ object Datadog {
         }
     }
 
+    /**
+     * For Datadog internal use only.
+     *
+     * @see _InternalProxy
+     */
+    @JvmStatic
```
0xnm, line 292: I don't think we need this annotation here, because this property is supposed to be called only by us, so no interop with Java is needed (both Flutter and RN use Kotlin for the native modules).

0xnm (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/_InternalProxy.kt
```diff
@@ -0,0 +1,60 @@
...
+@Suppress(
+    "UndocumentedPublicClass",
+    "UndocumentedPublicFunction",
+    "UndocumentedPublicProperty",
+    "ClassNaming",
+    "VariableNaming"
+)
+class _InternalProxy {
+    class _TelemetryProxy {
+        private val telemetry: Telemetry = Telemetry()
```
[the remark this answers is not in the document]
fuzzybinary: Do you think we should inject Telemetry instead?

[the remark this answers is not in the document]
0xnm: you can use default arguments in the constructor and provide them explicitly in the test.

fuzzybinary:
Given some things we found in the iOS PR, I've restructured this to separate the "core" internals (only telemetry at the moment) and RUM Internals. This simplifies testing as well as avoids some issues with lazy initializers having unwelcome side affects.

xgouchet (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/InternalProxyTest.kt
```diff
@@ -0,0 +1,62 @@
...
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class InternalProxyTest {
+
+    @Test
+    fun `M proxy telemetry to RumMonitor W debug()`(
+        @StringForgery message: String
+    ) {
+        // Given
+        val mockTelemetry = Mockito.mock(Telemetry::class.java)
```
0xnm, line 36: nit: we can do `import org.mockito.Mockito.mock` and just use `mock` call directly without `Mockito.` prefix as it is done in the codebase.

dd-sdk-android/src/test/kotlin/com/datadog/android/InternalProxyTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        // When
+        proxy._telemetry.debug(message)
+
+        // Then
+        verify(mockTelemetry).debug(message)
+    }
+
+    @Test
+    fun `M proxy telemetry to RumMonitor W error()`(
```
0xnm, line 47: test with `throwable` is missing.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/RumInternalProxyTest.kt
```diff
@@ -0,0 +1,59 @@
...
+@ForgeConfiguration(Configurator::class)
+internal class RumInternalProxyTest {
+
+    @Test
+    fun `M proxy addLongTask to RumMonitor W addLongTask()`(
+        @LongForgery time: Long,
+        @StringForgery target: String
+    ) {
+        // Given
+        val proxy = _RumInternalProxy(rumMonitor.mockInstance)
```
0xnm, line 41: we can simply `mock<AdvancedRumMonitor>()` here, don't think we need to use `GlobalRumMonitorTestConfiguration`.

0xnm (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/_InternalProxy.kt
```diff
@@ -0,0 +1,54 @@
...
+ * are not considered part of the public interface of the Datadog SDK.
+ */
+@Suppress(
+    "UndocumentedPublicClass",
+    "UndocumentedPublicFunction",
+    "UndocumentedPublicProperty",
+    "ClassNaming",
+    "VariableNaming"
+)
+class _InternalProxy {
```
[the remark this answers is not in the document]
fuzzybinary: Yeah, it's a good point. My thinking was that we might want Telemetry without RUM -- or I guess more importantly we might initialize and send telemetry even if the end user doesn't initialize RUM, so I wanted to keep it in the Core internals rather than on the RUM one.

fuzzybinary:
@0xnm updated and rebased on develop for a final review.

0xnm (approved):
lgtm!

PR #988, opened 2022-07-20
RUMM-2347: Provide core SDK context

0xnm:
### What does this PR do?

This change brings a complete `DatadogContext` which includes some core SDK information, like service name, env, time information, device information, etc.

Having `DatadogContext` allows to replace `userInfoProvider`, `timeProvider`, `networkInfoProvider`, `deviceInfoProvider` with it (currently replacement is done mostly in RUM, not everywhere).

The data inside `DatadogContext` is from the time the context was requested, there are no computed properties.

**NB**: this change doesn't bring any thread-safety invariants or non-blocking API, those will be addresses later.

Reference in iOS repository https://github.com/DataDog/dd-sdk-ios/pull/916.


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/ContextProvider.kt
```diff
@@ -9,5 +9,9 @@ package com.datadog.android.v2.core.internal
 import com.datadog.android.v2.api.context.DatadogContext
 
 internal interface ContextProvider {
-    val context: DatadogContext?
+    // TODO RUMM-0000 getting context may be quite heavy, should it be something non-blocking here?
```
mariusc83, line 12: You mean like an async call ? Maybe we can provide both ?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/DatadogContextProvider.kt
```diff
@@ -0,0 +1,101 @@
...
+                        email = email,
+                        additionalProperties = additionalProperties
+                    )
+                },
+                trackingConsent = coreFeature.trackingConsentProvider.getConsent(),
+                featuresContext = emptyMap()
+            )
+        }
+
+    private fun NetworkInfoV1.Connectivity.asV2(): NetworkInfo.Connectivity {
```
mariusc83, line 80: why doing the conversion and not have one single `NetworkInfo` for both v1 and v2 ?

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/context/DeviceType.kt
```diff
@@ -4,9 +4,12 @@
...
  */
 
-package com.datadog.android.core.internal.system
+package com.datadog.android.v2.api.context
 
-internal enum class DeviceType {
+/**
+ * Device type.
+ */
+public enum class DeviceType {
```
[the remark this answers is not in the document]
0xnm: indeed, `public` can be removed here

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/ContextProvider.kt
0xnm: We are currently not sure if we will expose core context provider as a part of public API. For internal needs we are ok with blocking call for now, considering the way we get context fields and what fields are added to the context, so keeping things simple for now.

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/DatadogContextProvider.kt
0xnm: good point. The thing is that we indeed have some duplication for models between V1 and V2: `UserInfo` as well, for example. Eventually it will be only one, duplication is only for the V1 to V2 transition period, to have smaller changes footprint in PRs.

PR #991, opened 2022-07-21
RUMM-2327 add internal DNS resolver

xgouchet:
### What does this PR do?

Ensures a different IP address is used whenever we make a call to intake, and regularly refresh the DNS lookup.

### Motivation

Prevent retrying on the same IP in case of a failing server node.


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RotatingDnsResolver.kt
```diff
@@ -0,0 +1,69 @@
...
+    private val knownHosts = mutableMapOf<String, ResolvedHost>()
+
+    // region Dns
+
+    override fun lookup(hostname: String): MutableList<InetAddress> {
+        val knownHost = knownHosts[hostname]
+
+        return if (knownHost != null && isValid(knownHost)) {
+            knownHost.rotate()
+            knownHost.addresses.toMutableList()
```
0xnm, line 47: `ResolvedHost#addresses` already has `MutableList<InetAddress>`, so I think `toMutableList()` call is not needed here

```suggestion
            knownHost.addresses
```

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RotatingDnsResolver.kt
```diff
@@ -0,0 +1,69 @@
...
+        } else {
+            @Suppress("UnsafeThirdPartyFunctionCall") // handled by caller
+            val result = delegate.lookup(hostname)
+            knownHosts[hostname] = ResolvedHost(hostname, result.toMutableList())
```
0xnm, line 51: here as well no need for `toMutableList()`, we have `MutableList` as a result of `lookup` call.

0xnm (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RotatingDnsResolver.kt
xgouchet: This is done in order to not have the list we keep mutated from outside this class

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RotatingDnsResolver.kt
xgouchet: Again this is to prevent external code making changes to our internal list

PR #1000, opened 2022-07-25
Support sending CPU architecture as part of device info

fuzzybinary:
### What does this PR do?
Supports sending CPU architecture as part of the Device Info on events.

This utilizes a partial update of the RUM event schema to include 'architecture' but not some other changes in the most up to date schema, essentially putting it in a 'detached HEAD' state.

Architecture is taken from String.getProperty("os.arch") over Build.CPU_ABI both because CPU_ABI is deprecated and because os.arch appears to give us a result closer to what we want, without including the name of the CPU.

Fixes RUMM-2368

### Motivation

Dart / Flutter (and, eventually, NDK crashes) will require the CPU architecture to determine which symbol files to utilize to desymbolicate crash reports.


mariusc83 (approved):
LGTM

0xnm (approved):

PR #1007, opened 2022-08-03
RUMM-2356 Add session replay recorder basic logic

mariusc83:
### What does this PR do?

This PR introduces the basic logic in our `SnapshotProducer` as a part of the `ScreenRecorder` component where we are actually going to capture the application screens and resolve them as `Node` tree equivalents. 

Please note that for now the `SnapshotProducer` is quite dummy and can only handle basic application screens mostly from our Android Sample App. The way we will work on this component is through multiple iterations by running it with different sample applications and improving the output.
 
### Motivation

### Additional Notes

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
@@ -6,13 +6,95 @@
 
 package com.datadog.android.sessionreplay.recorder
 
+import android.os.Build
 import android.view.View
+import android.view.ViewGroup
+import android.view.ViewStub
+import androidx.appcompat.widget.ActionBarContextView
+import androidx.appcompat.widget.Toolbar
+import com.datadog.android.sessionreplay.recorder.mapper.GenericWireframeMapper
+import com.datadog.android.sessionreplay.recorder.mapper.ViewScreenshotWireframeMapper
+import java.util.LinkedList
 
-@Suppress("UNUSED_PARAMETER")
-internal class SnapshotProducer {
+internal class SnapshotProducer(
+    private val viewScreenshotWireframeMapper: ViewScreenshotWireframeMapper =
+        ViewScreenshotWireframeMapper(),
+    private val genericWireframeMapper: GenericWireframeMapper =
+        GenericWireframeMapper(imageMapper = viewScreenshotWireframeMapper)
+) {
 
-    fun produce(rootView: View, scale: Float): Node? {
-        // TODO: RUMM-2356
-        return null
+    private val systemViewIds by lazy {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            setOf(android.R.id.navigationBarBackground, android.R.id.statusBarBackground)
+        } else {
+            emptySet()
+        }
+    }
+
+    fun produce(rootView: View, pixelsDensity: Float): Node? {
+        return convertViewToNode(rootView, pixelsDensity)
+    }
+
+    @Suppress("ComplexMethod")
+    private fun convertViewToNode(view: View, pixelsDensity: Float): Node? {
+        if (view.isNotVisible()) {
+            return null
+        }
+
+        if (view.isSystemNoise()) {
+            return null
+        }
+
+        if (view.isToolbar()) {
+            // skip adding the children and just take a screenshot of the toolbar.
+            // It is too complex to de - structure this in multiple wireframes
+            // and we cannot actually get all the details here.
+            return Node(
+                wireframes = listOf(viewScreenshotWireframeMapper.map(view, pixelsDensity))
+            )
+        }
+
+        val childNodes = LinkedList<Node>()
+        if (view is ViewGroup && view.childCount > 0) {
+            for (i in 0 until view.childCount) {
+                val viewChild = view.getChildAt(i) ?: continue
+                convertViewToNode(viewChild, pixelsDensity)?.let {
+                    childNodes.add(it)
+                }
+            }
+        }
+
+        return convertViewToNode(view, childNodes, pixelsDensity)
+    }
+
+    private fun convertViewToNode(
+        view: View,
+        childNodes: LinkedList<Node>,
+        pixelsDensity: Float
+    ) = Node(
+        children = childNodes,
+        wireframes = listOf(genericWireframeMapper.map(view, pixelsDensity))
+    )
+
+    private fun View.isToolbar(): Boolean {
+        return (
+            Toolbar::class.java.isAssignableFrom(this::class.java) &&
+                this.id == androidx.appcompat.R.id.action_bar
+            ) ||
+            (
+                Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP &&
+                    android.widget.Toolbar::class.java
+                        .isAssignableFrom(this::class.java)
+                )
+    }
+
+    private fun View.isNotVisible(): Boolean {
+        return !this.isShown || this.width <= 0 || this.height <= 0
+    }
+
+    private fun View.isSystemNoise(): Boolean {
+        return id in systemViewIds ||
+            this is ViewStub ||
+            this is ActionBarContextView
     }
 }
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/BaseWireframeMapper.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/BaseWireframeMapper.kt
@@ -1,0 +1,29 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.recorder.mapper
+
+import android.view.View
+import com.datadog.android.sessionreplay.model.MobileSegment
+import java.util.Locale
+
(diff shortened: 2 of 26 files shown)


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/Node.kt
```diff
@@ -6,4 +6,10 @@
 
 package com.datadog.android.sessionreplay.recorder
 
-internal data class Node(val id: Int, val parent: Node? = null)
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal data class Node(
+    val id: Long,
+    val wireframes: List<MobileSegment.Wireframe>,
```
ncreated, line 13: suggestion: I think it might be healthier to not depend directly on (HTTP) transport details in the core SR logic (the `MobileSegment.Wireframe` is what we serialize and send in payload). Introducing intermediate (DTO) and doing `Node` → indermediate → `Wireframe` transformation could IMO help on a long run by clearly separating concerns. Otherwise, the Recorder code will have to be changed if transport details change, which seems to be against SoC as both layers clearly overlap here. WDYT?

ncreated (commented):
Looks really good, I can easily understand what's happening in there 👍. Left one arch suggestion, but non-blocking.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/Node.kt
mariusc83: Yes I thought about it but then I think I prefer it this way even though maybe at some point I need to change the wireframes a bit. Main reason why is:
- I don't want to introduce another layer which will require a transformer. If there's any change on the Wireframe I will still have to apply it at the transformer level from the intermediate DTO to Wireframe
- Most important I want to avoid creating too many objects. With this approach the wireframes instances created here will be re - used at the Processor level.

urseberry (commented):
No docs impact.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
```diff
@@ -6,13 +6,305 @@
...
+        }
+    }
+
+    fun produce(rootView: View, pixelsDensity: Float): Node? {
+        return rootView.toNode(pixelsDensity)
+    }
+
+    @Suppress("ComplexMethod")
+    private fun View.toNode(pixelsDensity: Float): Node? {
+        if (isNotValid()) {
```
xgouchet, line 46: Same here, using extension makes things ambiguous here. Does `isNotValid` refer to the implicit `this: View`, or to the parent `this: SnapshotProducer`?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
```diff
@@ -6,13 +6,305 @@
...
+            return null
+        }
+
+        if (isSystemNoise()) {
```
xgouchet, line 50: Same here

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
```diff
@@ -6,13 +6,305 @@
...
+            return null
+        }
+
+        if (isToolbar()) {
```
xgouchet, line 54: Same here

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
```diff
@@ -6,13 +6,305 @@
...
+                view.toNode(pixelsDensity)?.let {
+                    childNodes.add(it)
+                }
+            }
+        }
+
+        return toNode(childNodes, pixelsDensity)
+    }
+
+    private fun View.toNode(childNodes: LinkedList<Node>, pixelsDensity: Float) =
```
xgouchet, line 77: Please avoid using block function when the implementation is this long, as it hides the return type and make things harder to read

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
```diff
@@ -6,13 +6,305 @@
...
+            y,
+            scaledWidth,
+            scaledHeight,
+            shapeStyle = styleBorderPair?.first,
+            border = styleBorderPair?.second
+        )
+    }
+
+    private fun Drawable.resolveShapeStyleAndBorder(): Pair<MobileSegment
+        .ShapeStyle?, MobileSegment.ShapeBorder?>? {
```
xgouchet, line 161: Can we pick a consistent name for all those conversion functions? You have at least three in this class: `toSomething`, `asSomething` and `resolveSomething`.

xgouchet (changes requested):
Overall the `SnapshotProducer` class seems quite complex as you have many distinct usecases (TextView, ImageView, ButtonView, …). 
I think it might be better to have a `WireframeAdapter` interface (or converter, or whatever name suits best), looking like

```
interface WireframeAdapter <T: View> {
    fun adapt(input: T): Node
}
```

and then have independent implementations for each type we want to support.

And then, _if_ you really need shared extension functions, then place them in a separate file with documentation

mariusc83:
> Overall the `SnapshotProducer` class seems quite complex as you have many distinct usecases (TextView, ImageView, ButtonView, …). I think it might be better to have a `WireframeAdapter` interface (or converter, or whatever name suits best), looking like
> 
> ```
> interface WireframeAdapter <T: View> {
>     fun adapt(input: T): Node
> }
> ```
> 
> and then have independent implementations for each type we want to support.
> 
> And then, _if_ you really need shared extension functions, then place them in a separate file with documentation

I do agree that I might not need all these extension functions, will try to get rid of at least part of them.
Regarding the extraction of the logic into a `WireframeAdapter` I do have this in mind for later. For now I just wanted to keep it simple and have this basic functionality here but I do have in mind that this will require some logic extraction and more components as we will add more things in here. For now I really can't anticipate what we will add and where to see exactly how my Adapter will look like so I just want to wait a bit more.

xgouchet (changes requested):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
```diff
@@ -6,13 +6,101 @@
...
+                this.id == androidx.appcompat.R.id.action_bar
+            ) ||
+            (
+                Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP &&
+                    android.widget.Toolbar::class.java
+                        .isAssignableFrom(this::class.java)
+                )
+    }
+
+    private fun View.isNotValid(): Boolean {
```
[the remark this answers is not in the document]
mariusc83: I will change it to `trackable`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/SnapshotProducer.kt
```diff
@@ -6,13 +6,101 @@
...
+        }
+
+        if (view.isSystemNoise()) {
+            return null
+        }
+
+        if (view.isToolbar()) {
+            // skip adding the children and just take a screenshot of the toolbar.
+            // It is too complex to de - structure this in multiple wireframes
+            // and we cannot actually get all the details here.
```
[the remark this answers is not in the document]
mariusc83: It will already be handled in v1 once we introduce the logic in the `screenshotWireframeMapper`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/BaseWireframeMapper.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import android.view.View
+import com.datadog.android.sessionreplay.model.MobileSegment
+import java.util.Locale
+
+internal abstract class BaseWireframeMapper<T : View, S : MobileSegment.Wireframe> :
+    WireframeMapper<T, S> {
+
+    protected fun resolveViewId(view: View): Long {
+        return if (view.id != View.NO_ID) view.id.toLong() else view.hashCode().toLong()
```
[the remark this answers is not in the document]
mariusc83: Hmmm you are right...it might be a problem due the the `view.hashcode` maybe I should use the reference address there instead ? What do you think ? It can create problems if it happens as it might override other view properties in the mutation computation.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/RecorderWireframeMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+
+package com.datadog.android.sessionreplay.recorder.mapper
+
+import android.view.View
+import android.widget.Button
+import android.widget.ImageView
+import android.widget.TextView
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal class RecorderWireframeMapper(
```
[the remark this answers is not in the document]
mariusc83: Yes...I think I need to find better names for these 🤔

xgouchet (approved):

PR #1023, opened 2022-08-29
Add configuration method to disable action tracking.

fuzzybinary:
### What does this PR do?

This allows users to replace the current UserAcitonTrackingStategy with a NoOpUserActionTrackingStrategy.

### Motivation

This is mostly for Flutter, since the initialization of the UserActionTrackingStrategy happens after the MainActivity is live, so the initial creation misses hooking into the Activity anyway, and subsequent attempts to listen to the Activity result in logcat spam, as there are no views that the tracking strategy can see.

### DIFF
--- a/dd-sdk-android/apiSurface
+++ b/dd-sdk-android/apiSurface
@@ -85,6 +85,7 @@ data class com.datadog.android.core.configuration.Configuration
     fun useCustomCrashReportsEndpoint(String): Builder
     fun useCustomRumEndpoint(String): Builder
     fun trackInteractions(Array<com.datadog.android.rum.tracking.ViewAttributesProvider> = emptyArray(), com.datadog.android.rum.tracking.InteractionPredicate = NoOpInteractionPredicate()): Builder
+    fun disableInteractionTracking(): Builder
     fun trackLongTasks(Long = DEFAULT_LONG_TASK_THRESHOLD_MS): Builder
     fun useViewTrackingStrategy(com.datadog.android.rum.tracking.ViewTrackingStrategy?): Builder
     fun addPlugin(com.datadog.android.plugin.DatadogPlugin, com.datadog.android.plugin.Feature): Builder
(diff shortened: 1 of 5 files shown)


0xnm (approved):

xgouchet (approved):

mariusc83 (approved):

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumConfigE2ETests.kt
```diff
@@ -375,6 +375,28 @@ class RumConfigE2ETests {
...
+                defaultConfigurationBuilder(
+                    logsEnabled = true,
+                    tracesEnabled = true,
+                    rumEnabled = true,
+                    crashReportsEnabled = true
+                ).disableInteractionTracking().build()
+            )
+        }
+
+        sendRandomActionEvent(testMethodName)
```
0xnm, line 397: It seems we don't have a monitor for this test yet, right? Also I'm curious if we can test `disableInteractionTracking ` at all - `sendRandomActionEvent` has no connection to this method it seems, because it will send a manual action, which is not guarded by the `disableInteractionTracking`.

fuzzybinary: I was going to add the monitor after this is in, since creating it early would just cause some trigger noise.

Good call on the sendRandomAction. This is why I wanted you to re-review.

Do we have a method that does a non-manual action?

0xnm: good question. We can `launch` some activity and then use [Espresso](https://developer.android.com/training/testing/espresso/basics) to make a tap on some element. And then if interactions tracking is disabled we shouldn't see it in RUM.

But maybe it is too much for such small method. I'm ok with removing E2E test as well. WDYT?

fuzzybinary:
One more time for anyone who's interested. I've pulled out the e2e and marked the API as not testable. If I can get someone to just give a 👍 we'll merge this in.

0xnm (approved):

xgouchet (approved):

PR #1042, opened 2022-09-14
RUMM-2492: Check if WorkManager is initialized before using it

0xnm:
### What does this PR do?

This PR adds a check to see if `WorkManager` is initialized before using it, because `WorkManager.getInstance` will throw `IllegalStateException` if `WorkManager` is not initialized and this add the noise to our telemetry.

Unfortunately, there is no `WorkManager#isInitialized` method yet (it will be added only in WorkManager 2.8, which is still in alpha), so had to write a hacky workaround.

Another option would be to check if we are in the main process before accessing `WorkManager`, but this won't work because:

* with `work-multiprocess` artifact execution/initialization of WorkManager may be shifted to any process
* User may do a custom initialization which happens after we are calling `WorkManager`.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/utils/WorkManagerUtils.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/utils/WorkManagerUtils.kt
@@ -56,3 +56,15 @@ internal fun triggerUploadWorker(context: Context) {
         sdkLogger.errorWithTelemetry(SETUP_ERROR_MESSAGE, e)
     }
 }
+
+@Suppress("TooGenericExceptionCaught", "SwallowedException")
+internal fun isWorkManagerInitialized(context: Context): Boolean {
+    return try {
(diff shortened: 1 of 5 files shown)


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/utils/WorkManagerUtils.kt
```diff
@@ -56,3 +56,15 @@ internal fun triggerUploadWorker(context: Context) {
         sdkLogger.errorWithTelemetry(SETUP_ERROR_MESSAGE, e)
     }
 }
+
+@Suppress("TooGenericExceptionCaught", "SwallowedException")
+internal fun isWorkManagerInitialized(context: Context): Boolean {
```
mariusc83, line 61: I think I had this before, not sure why I have this impression 🤔

mariusc83 (approved):

PR #1051, opened 2022-09-21
RUMM-2504: Switch to the SDK v2 storage component

0xnm:
### What does this PR do?

This change does a switch to the `Storage` implementation written earlier. While it is not used as it should be in the feature configuration per SDK v2 plan, this change helps to minimize the diff of the transition and makes a proof that new component works.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileDataWriter.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileDataWriter.kt
@@ -9,25 +9,41 @@ package com.datadog.android.core.internal.persistence.file.batch
 import androidx.annotation.WorkerThread
 import com.datadog.android.core.internal.persistence.DataWriter
 import com.datadog.android.core.internal.persistence.Serializer
-import com.datadog.android.core.internal.persistence.file.FileOrchestrator
-import com.datadog.android.core.internal.persistence.file.FilePersistenceConfig
-import com.datadog.android.core.internal.persistence.file.FileWriter
 import com.datadog.android.core.internal.persistence.serializeToByteArray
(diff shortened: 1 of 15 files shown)


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileDataWriter.kt
```diff
@@ -68,39 +83,14 @@ internal open class BatchFileDataWriter<T : Any>(
...
         synchronized(this) {
-            val success = writeData(byteArray)
-            if (success) {
-                onDataWritten(data, byteArray)
-            } else {
-                onDataWriteFailed(data)
+            val context = contextProvider.context
+            val eventId = System.identityHashCode(data).toString()
+            storage.writeCurrentBatch(context) {
+                pendingWriteEvents[eventId] = data to byteArray
```
mariusc83, line 89: aren't we concerned that we will keep too much data in HEAP ?

0xnm: Good question. For now we need to store both event and serialized event mostly because of the RUM, but maybe later on with once RUM will start using write scope to perform writes we can change it.

What is your expectation for the memory pressure during SR? How often and how big are going to be events received? I guess concern is mostly because of the bitmaps?

I would propose to add TODO for now. WDYT?

mariusc83: Yes I am mostly thinking for Bitmaps here, let's benchmark this when we get there and see how it goes.

mariusc83 (approved):

PR #1066, opened 2022-09-30
Minor performance optimization during serialization into JSON format

0xnm:
### What does this PR do?

I noticed a small issue in the related code: since `JsonArray` actually implements `Iterable`, `is JsonArray` branch will never be executed, because it goes after `is Iterable` branch.

By moving `is JsonArray` up we skip useless `toJsonArray` call which will create new object and copy elements there from the old object.


mariusc83 (approved):
Good catch !!

PR #1070, opened 2022-10-05
RUMM-2563: Create storage and uploader outside of the feature

0xnm:
### What does this PR do?

`Storage` and `DataUploader` are pulled outside of the feature-specific classes, because `core` module will be responsible for those two.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
@@ -20,6 +20,10 @@ import com.datadog.android.core.internal.SdkFeature
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleCallback
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleMonitor
 import com.datadog.android.core.internal.persistence.NoOpDataWriter
+import com.datadog.android.core.internal.persistence.file.FileMover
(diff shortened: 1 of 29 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
```diff
@@ -92,31 +104,38 @@ internal class DatadogCore(
...
+        val uploader = DataOkHttpUploader(
+            requestFactory = uploadConfiguration.requestFactory,
+            internalLogger = sdkLogger,
+            callFactory = coreFeature.okHttpClient,
+            sdkVersion = coreFeature.sdkVersion,
+            androidInfoProvider = coreFeature.androidInfoProvider
+        )
+        // TODO RUMM-0000 those still need to be registered explicitly,
+        //  because of cross-functionality. We will get rid of them eventually and use getFeature
+        when (featureName) {
```
mariusc83, line 117: are we sure they are not stepping each other foot ? Meaning that now you have 2 features for same feature (e.g. for RUM you have `rumFeature` and `DatadogFeature`)

0xnm: currently no, because `DatadogFeature` is just a holder for now, it doesn't have a lifecycle. In the future I'm thinking about the possibility of getting rid of `DatadogFeature` and put things into `SdkFeature` instead.

mariusc83 (approved):

PR #1073, opened 2022-10-06
RUMM-2562 Correct the way we handle orientation change events in SR

mariusc83:
### What does this PR do?

This PR fixes the way we are handling the orientation change for Session Replay. Previously we were handling this in a separated flow which introduced a new state into the `SnapshotProcessor` causing problems regarding the `new view` detection and the order of the records. With this new approach we are handling the `OrientationChanged` event in the same stack used for handling the screen snapshot and by doing this we eliminate the potential issues resulting from introducing a new state for the `SnapshotProcessor`.

### Motivation

### Additional Notes

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/SnapshotProcessor.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/SnapshotProcessor.kt
@@ -35,7 +35,7 @@ internal class SnapshotProcessor(
     private var lastSnapshotTimestamp = 0L
 
     @MainThread
-    override fun process(node: Node) {
+    override fun process(node: Node, orientationChanged: OrientationChanged?) {
         if (node.wireframes.isEmpty()) {
             // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
             return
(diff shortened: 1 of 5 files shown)


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/SnapshotProcessor.kt
```diff
@@ -159,7 +147,7 @@ internal class SnapshotProcessor(
         }
     }
 
-    private fun isLastFullSnapshotTime(): Boolean {
+    private fun resolveIsFullSnapshotTime(): Boolean {
```
0xnm, line 150: nit: maybe `isTimeForFullSnapshort` (but yeah , this may not sound better)?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/SnapshotProcessor.kt
```diff
@@ -226,7 +214,7 @@ internal class SnapshotProcessor(
         )
     }
 
-    private fun isNewView(
+    private fun resolveIsNewView(
```
0xnm, line 217: why do we rename this? imo without `resolve` it also looks reasonable.

0xnm (approved):

PR #1078, opened 2022-10-11
RUMM-2651 SR - skip new lines and spaces when obfuscating texts

mariusc83:
### What does this PR do?

We need to skip the new lines and spaces characters when applying the obfuscation on the text values to keep the aspect ratio of the replicated text view element in the Web Player.

### Motivation

### Additional Notes

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextWireframeMapper.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextWireframeMapper.kt
@@ -9,14 +9,11 @@ package com.datadog.android.sessionreplay.recorder.mapper
 import android.widget.TextView
 
 internal class MaskAllTextWireframeMapper(
-    viewWireframeMapper: ViewWireframeMapper = ViewWireframeMapper()
+    viewWireframeMapper: ViewWireframeMapper = ViewWireframeMapper(),
+    private val stringObfuscator: StringObfuscator = StringObfuscator()
 ) : TextWireframeMapper(viewWireframeMapper) {
 
     override fun resolveTextValue(textView: TextView): String {
-        return String(CharArray(textView.text.length) { CHARACTER_MASK })
-    }
-
-    companion object {
-        private const val CHARACTER_MASK = 'x'
+        return stringObfuscator.obfuscate(textView.text.toString())
     }
 }
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscator.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscator.kt
@@ -1,0 +1,65 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.recorder.mapper
+
+import android.os.Build
+import androidx.annotation.RequiresApi
+
+internal class StringObfuscator {
+
+    fun obfuscate(stringValue: String): String {
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            obfuscateUsingCodeStream(stringValue)
+        } else {
+            obfuscateUsingCharacterCode(stringValue)
+        }
+    }
+
+    private fun obfuscateUsingCharacterCode(stringValue: String): String {
+        return String(
+            CharArray(stringValue.length) {
+                val character = stringValue[it]
+                // Given that we replace each printable character with `x` for 2 chars expressions
+                // as emojis we will have 2 `x` instead of 1 but this will not be a problem as the
+                // obfuscation will still be applied.
+                if (Character.isWhitespace(character.code)) {
+                    character
+                } else {
+                    CHARACTER_MASK
+                }
+            }
+        )
+    }
+
+    @RequiresApi(Build.VERSION_CODES.N)
+    private fun obfuscateUsingCodeStream(stringValue: String): String {
+        // Because we are using the CharSequence.codePoints() stream we are going to correctly
+        // handle the cases where the text contains 2 chars expression. In this case one single
+        // codePoint will be returned for those 2 chars and the obfuscation char will be `x`.
+        val stringBuilder = StringBuilder()
+        stringValue.codePoints().forEach {
+            if (Character.isWhitespace(it)) {
+                // I don't think we should log this case here. I could not even reproduce it
+                // in my tests. As long as there is a valid string there there should not be
+                // any problem.
+                @Suppress("SwallowedException")
+                try {
+                    stringBuilder.append(Character.toChars(it))
+                } catch (e: IllegalArgumentException) {
+                    stringBuilder.append(CHARACTER_MASK)
+                }
+            } else {
+                stringBuilder.append(CHARACTER_MASK)
+            }
+        }
+        return stringBuilder.toString()
+    }
+
+    companion object {
+        private const val CHARACTER_MASK = 'x'
+    }
+}
--- a/library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextViewWireframeMapperTest.kt
+++ b/library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextViewWireframeMapperTest.kt
@@ -6,7 +6,6 @@
 
 package com.datadog.android.sessionreplay.recorder.mapper
 
-import android.widget.Button
 import android.widget.TextView
 import com.datadog.android.sessionreplay.recorder.aMockView
 import com.datadog.android.sessionreplay.utils.ForgeConfigurator
@@ -14,12 +13,14 @@ import com.datadog.tools.unit.extensions.ApiLevelExtension
 import com.nhaarman.mockitokotlin2.mock
 import com.nhaarman.mockitokotlin2.whenever
 import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.annotation.StringForgery
 import fr.xgouchet.elmyr.junit5.ForgeConfiguration
 import fr.xgouchet.elmyr.junit5.ForgeExtension
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
+import org.mockito.Mock
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
 import org.mockito.quality.Strictness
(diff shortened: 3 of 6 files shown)


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextWireframeMapper.kt
```diff
@@ -13,10 +13,22 @@ internal class MaskAllTextWireframeMapper(
 ) : TextWireframeMapper(viewWireframeMapper) {
 
     override fun resolveTextValue(textView: TextView): String {
-        return String(CharArray(textView.text.length) { CHARACTER_MASK })
+        val text = textView.text
+        return String(
+            CharArray(text.length) {
+                val character = text[it]
+                if (character == CHARACTER_SPACE || character == CHARACTER_NEW_LINE) {
```
xgouchet, line 20: you can use `character.isWhitespace()` to check this rather than comparing to our own hardcoded values

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextViewWireframeMapperTest.kt
```diff
@@ -40,27 +39,108 @@ internal class MaskAllTextViewWireframeMapperTest : BaseTextViewWireframeMapperT
...
             whenever(this.typeface).thenReturn(mock())
         }
 
         // When
-        val textWireframe = testedTextWireframeMapper.map(mockButton, fakePixelDensity)
+        val textWireframe = testedTextWireframeMapper.map(mockTextView, fakePixelDensity)
 
         // Then
-        val expectedWireframe = mockButton.toTextWireframe().copy(text = fakeText)
+        val expectedWireframe = mockTextView.toTextWireframe().copy(text = fakeExpectedText)
```
xgouchet, line 75: You should also have a test in the form 

```kotlin
val chunk1 = randomText
val chunk2 = randomText
val spacer = randomWhitespace // forge.forge.aWhitespaceString()
val input = chunk1 + spacer + chunk2
val expectedOutput = ("x" * chunk1.length) + spacer + ("x" * chunk2.length)
```

with this (I don't know if it exist in our repo

```kotlin
operator fun String.times(i: Int) {
    val builder = StringBuilder(i * length)
    repeat(i) {
        builder.append(this)
    }
}
```

xgouchet (changes requested):

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextViewWireframeMapperTest.kt
```diff
@@ -40,27 +39,108 @@ internal class MaskAllTextViewWireframeMapperTest : BaseTextViewWireframeMapperT
...
+        val text = textView.text
+        return String(
+            CharArray(text.length) {
+                val character = text[it]
+                if (character == ' ' || character == '\n') {
+                    character
+                } else {
+                    'x'
+                }
+            }
```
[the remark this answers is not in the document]
mariusc83: yeah I know, this is for all the other tests that are not exactly testing the obfuscation, is just that we are asserting everything in those. For all the obfuscation cases I added the 3 tests below.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/MaskAllTextViewWireframeMapperTest.kt
mariusc83: Good point

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscator.kt
```diff
@@ -0,0 +1,26 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.recorder.mapper
+
+internal class StringObfuscator {
+    fun obfuscate(stringValue: String): String {
+        return String(
```
0xnm, line 11: let's add TODO maybe here, saying it works only for ASCII. Also character can be non-printable Control character, but I guess we can ignore it.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscatorTest.kt
```diff
@@ -0,0 +1,145 @@
...
+            ).times(n)
+
+        // When
+        val obfuscatedText = testedObfuscator.obfuscate(fakeText)
+
+        // Then
+        assertThat(obfuscatedText).isEqualTo(fakeExpectedText)
+    }
+
+    private fun String.times(n: Int): String {
```
0xnm, line 138: it seems the default [repeat](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/repeat.html) should do the job?

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscatorTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        val fakeExpectedChunk4 = forge.aString { 'x' }
+        val fakeExpectedChunk5 = forge.aString(size = forge.anInt(1, max = 10)) { '\n' }
+        val n = forge.anInt(min = 1, max = 10)
+        val fakeExpectedText = (
+            fakeExpectedChunk1 +
+                fakeExpectedChunk2 +
+                fakeExpectedChunk3 +
+                fakeExpectedChunk4 +
+                fakeExpectedChunk5
+            ).times(n)
```
0xnm, line 56: why do we need `times` call if it just copies the same?

mariusc83: to have a bigger text. I was following @xgouchet suggestion, not sure that this is what he wanted but I think is a bit overkill also. @xgouchet do you think this is still needed ?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscator.kt
mariusc83: Apparently it handles almost every non - printable characters in UNICODE as they are stating below:

```
    /**
     * Determines if the specified character is white space according to Java.
     * A character is a Java whitespace character if and only if it satisfies
     * one of the following criteria:
     * <ul>
     * <li> It is a Unicode space character ({@code SPACE_SEPARATOR},
     *      {@code LINE_SEPARATOR}, or {@code PARAGRAPH_SEPARATOR})
     *      but is not also a non-breaking space ({@code '\u005Cu00A0'},
     *      {@code '\u005Cu2007'}, {@code '\u005Cu202F'}).
     * <li> It is {@code '\u005Ct'}, U+0009 HORIZONTAL TABULATION.
     * <li> It is {@code '\u005Cn'}, U+000A LINE FEED.
     * <li> It is {@code '\u005Cu000B'}, U+000B VERTICAL TABULATION.
     * <li> It is {@code '\u005Cf'}, U+000C FORM FEED.
     * <li> It is {@code '\u005Cr'}, U+000D CARRIAGE RETURN.
     * <li> It is {@code '\u005Cu001C'}, U+001C FILE SEPARATOR.
     * <li> It is {@code '\u005Cu001D'}, U+001D GROUP SEPARATOR.
     * <li> It is {@code '\u005Cu001E'}, U+001E RECORD SEPARATOR.
     * <li> It is {@code '\u005Cu001F'}, U+001F UNIT SEPARATOR.
     * </ul>
     *
     * <p><b>Note:</b> This method cannot handle <a
     * href="#supplementary"> supplementary characters</a>. To support
     * all Unicode characters, including supplementary characters, use
     * the {@link #isWhitespace(int)} method.
     *
     * @param   ch the character to be tested.
     * @return  {@code true} if the character is a Java whitespace
     *          character; {@code false} otherwise.
     * @see     Character#isSpaceChar(char)
     * @since   1.1
     */
```

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscatorTest.kt
xgouchet: the `times` was more to generate a string with just 'x' of a given size, here it doesn't make sense

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscator.kt
mariusc83: I just realized that I need to use another method to support all unicode characters the way they are specifying in their comments.

mariusc83: Yes I imagine this is the way it should work but I think this will not be a problem. The only issue would be if one of this emojis will contain one of the `non printable` characters in their composition. In that case we will brake the text but I am pretty sure `Character.isWhitespace()` handles that somehow or maybe these `whitespace` characters are not used in emojis.

0xnm: My point is that in case of emoji we will have `xx` instead of single `x`, not sure if this is what we want to achieve (I guess `Character.isWhitespace()` will return `false` for any surrogate).

mariusc83: Yeah...I don't think this should be an issue as long as we obfuscate the texts. It will be a corner case anyway.

0xnm: I won't say it is a corner case, because there are plenty of symbols with above `\uffff`, and if we think of emojis we may find a lot of them in messenger apps, for example, and obfuscation may break formatting. Not a blocker for me, but let's acknowledge this.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscator.kt
```diff
@@ -0,0 +1,33 @@
...
+            CharArray(stringValue.length) {
+                val character = stringValue[it]
+                // Using the Character.isWhitespace() function we will support also the UNICODE
+                // characters according with the docs. Also have in mind that in JAVA the
+                // `char` primitive is a 16bits long number which covers all the UNICODE characters:
```
0xnm, line 16: `char` doesn't cover all the Unicode characters. It covers [BMP plain](https://en.wikipedia.org/wiki/Plane_(Unicode)) Unicode characters, but it doesn't cover Supplementary plain characters (above `0xFFFF`). When we call `character.code` we just cast from `char` to `int`, but to get `code` of the symbol from supplementary plain we need to merge 2 chars into single int.

0xnm (approved):

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/mapper/StringObfuscatorTest.kt
```diff
@@ -0,0 +1,310 @@
...
+    @BeforeEach
+    fun `set up`() {
+        testedObfuscator = StringObfuscator()
+    }
+
+    // region Android N and above
+
+    @Test
+    @TestTargetApi(Build.VERSION_CODES.N)
+    fun `M mask String W maskString(){string with newline, Android N}`(
```
0xnm, line 46: It is not `maskString`, but `obfuscate`. Same for other tests.

```suggestion
    fun `M mask String W obfuscate(){string with newline, Android N}`(
```

0xnm (approved):
lgtm!

xgouchet (approved):

PR #1081, opened 2022-10-12
D.marin/ktlint

daniel-m-dd:
### What does this PR do?

Update Ktlint to 0.45.1 and use a dedicated runner image

### Motivation

### Additional Notes


xgouchet (commented):
Nice job. One quick thought, we might want to remove the ktLint Gradle plugin which mentions a different version of ktLint and might return different reports when ran locally

xgouchet (approved):

daniel-m-dd:
> Nice job. One quick thought, we might want to remove the ktLint Gradle plugin which mentions a different version of ktLint and might return different reports when ran locally

Indeed, updated

.gitlab-ci.yml
```diff
@@ -56,12 +56,13 @@ create_key:
...
 analysis:ktlint:
-  tags: [ "runner:main" ]
-  image: $CI_IMAGE_DOCKER
+  tags: [ "runner:docker" ]
   stage: analysis
   timeout: 30m
+  image: "registry.ddbuild.io/mobile/images/ktlint:v0.1.0"
   script:
-    - GRADLE_OPTS="-Xmx2560m" ./gradlew :ktlintCheckAll --stacktrace --no-daemon
+    - echo "Using Ktlint version $(ktlint --version)"
```
0xnm, line 64: it also means we have to keep local `ktlint` version in sync with the version used in the image (because the rules may be different)?

daniel-m-dd: Yes, exactly.
In a second step, we might find a way to ensure this condition or make it easier to upgrade/be in sync. Any ideas?

0xnm: I guess the only way is to just build a new docker image with a different tag (or have a fat docker image with multiple ktlint versions embedded).

build.gradle.kts
```diff
@@ -23,7 +23,6 @@ buildscript {
         classpath(libs.androidToolsGradlePlugin)
         classpath(libs.kotlinGradlePlugin)
         classpath(libs.kotlinSPGradlePlugin)
-        classpath(libs.ktLintGradlePlugin)
```
0xnm, line 26: I'm wondering what the flow will be. Before I was running `ktlintFormat` before pushing to git - at which step formatting will be done now? Also sometimes `ktlintFormat` fails to do the job, like in case of too long lines - in that case some manual changes were needed.

daniel-m-dd: You ca still run `ktlint -F "**/*.kt" "**/*.kts" '!**/build/generated/**' '!**/build/kspCaches/**'` or make it a pre-commit hook. Will behave the same.
However I guess I should update the doc with the current required ktlint version (0.45.1)

0xnm: but in this case I need to install ktlint CLI on my machine manually, do I get it right? Or you are using the Dockerimage for that which has a project folder mounted to it?

daniel-m-dd: I personally use the local ktlint CLI on my machine (the gradle plugin was just already a wrapper around the CLI on your machine no?)

0xnm: No, Gradle plugin is not a wrapper for the existing `ktlint`, I don't have `ktlint` installed on my machine. But I see the proposed approach now :)

CONTRIBUTING.md
```diff
@@ -48,9 +48,6 @@ The whole project is covered by a set of static analysis tools, linters and test
 # launches the detekt static analysis for all modules
 ./gradlew detektAll
 
-# launches the ktlint format check for all modules
```
xgouchet, line 51: Can we mention the ktlint cli here instead to make sure contributors know that this is a check they should use: 

```console
ktlint -F "**/*.kt" "**/*.kts" '!**/build/generated/**' '!**/build/kspCaches/**'
```

CONTRIBUTING.md
```diff
@@ -180,14 +178,6 @@ class Foo :Observable(), Runnable {
...
 }
 
-```
-There is also a gradle task that you can use to automatically format the code following the
-required styling rules:
-
-```console
-
-./gradlew :dd-sdk-android:ktlintFormat
-
```
xgouchet, line 190: Same here

mariusc83 (approved):

xgouchet (approved):

0xnm (approved):

PR #1083, opened 2022-10-12
Create internal API for sending technical performance metrics

fuzzybinary:
### What does this PR do?

This creates an internal API for adding cross platform technical performance metrics (currently only Flutter Build Time, Flutter Raster Time and JS Refresh Rate). Each update is a single sample, which is aggregated via the same method as the Vitals aggregation, and posted to View events.

### Motivation

The Flutter and React Native SDKs will use this method to send in their performance metrics. The API was left open to potentially support sending different metrics in the future just by adding to the enum and serializing to the proper property on the View event.

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -27,6 +27,7 @@ import com.datadog.android.log.internal.utils.debugWithTelemetry
 import com.datadog.android.rum.GlobalRum
 import com.datadog.android.rum.RumActionType
 import com.datadog.android.rum.RumAttributes
+import com.datadog.android.rum.RumPerformanceMetric
 import com.datadog.android.rum.internal.domain.RumContext
 import com.datadog.android.rum.internal.domain.Time
 import com.datadog.android.rum.internal.domain.event.RumEventSourceProvider
@@ -43,6 +44,7 @@ import java.util.Locale
 import java.util.UUID
 import java.util.concurrent.TimeUnit
 import kotlin.math.max
+import kotlin.math.min
 
 @Suppress("LargeClass", "LongParameterList")
 internal open class RumViewScope(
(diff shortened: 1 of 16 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -869,3 +905,11 @@ internal open class RumViewScope(
         }
     }
 }
+
+private fun VitalInfo.toPerformanceMetric(): ViewEvent.FlutterBuildTime {
```
louiszawadzki, line 909: Having one specific type that is `FlutterBuildTime` might be harder to change later. Maybe we can create a dedicated type for this to be more generic?

fuzzybinary: This came from the generator, I think it just chose the first property that used the type? @0xnm do we maybe want to adjust the generator somehow?

0xnm: I'm not quite sure what is the concern here. Can you please elaborate?

fuzzybinary: Basically just that "FlutterBuildTime" is used as the type for things that are not FlutterBuildTime. But that type came from the schema generator so I just rolled with it.

0xnm: Well, this is how it is called in the schema. If it needs to have more generic name, the property should have generic name as well. But I don't see the issue here, because once we add a property for JS, it will get its own type. Types are basically not reused.

xgouchet: In the generator we have a way to reuse types when they have the same schema. I'll create a task to try and find a better way to name types

xgouchet (approved):

mariusc83 (approved):

PR #1085, opened 2022-10-13
RUMM-2463 Add API to enable/disable tracking of frustration signals

plousada:
### What does this PR do?

Adds a public API to disable/enable tracking of frustration signals. 

### Motivation

### Additional Notes


dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -420,6 +421,22 @@ internal constructor(
             return this
         }
 
+
+        /**
+         * Enables/Disables tracking of frustration signals.
+         *
+         * By default frustration signals are tracked. Currently the SDK supports detecting
+         * error_taps which occur when an error follows a user action tap.
```
0xnm, line 429:
```suggestion
         * error taps which occur when an error follows a user action tap.
```

Need also not to forget to update this statement when more signals are added.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScope.kt
```diff
@@ -31,8 +31,9 @@ internal class RumActionScope(
     inactivityThresholdMs: Long = ACTION_INACTIVITY_MS,
     maxDurationMs: Long = ACTION_MAX_DURATION_MS,
     private val rumEventSourceProvider: RumEventSourceProvider,
-    private val androidInfoProvider: AndroidInfoProvider
-) : RumScope {
+    private val androidInfoProvider: AndroidInfoProvider,
+    val trackFrustrations: Boolean
```
0xnm, line 35: can be made `private`, it seems it is not accessed from outside

```suggestion
    private val trackFrustrations: Boolean
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScopeTest.kt
```diff
@@ -99,6 +100,9 @@ internal class RumActionScopeTest {
     @Mock
     lateinit var mockRumEventSourceProvider: RumEventSourceProvider
 
+    @BoolForgery
+    var fakeTrackFrustrations: Boolean = true
```
0xnm, line 104:
```suggestion
    lateinit var fakeTrackFrustrations
```

because value will be assigned by injector

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScopeTest.kt
```diff
@@ -131,7 +135,8 @@ internal class RumActionScopeTest {
             TEST_INACTIVITY_MS,
             TEST_MAX_DURATION_MS,
             mockRumEventSourceProvider,
-            fakeAndroidInfoProvider
+            fakeAndroidInfoProvider,
+            true
```
0xnm, line 139: shouldn't we take the forged value?

```suggestion
            fakeTrackFrustrations
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScopeTest.kt
```diff
@@ -80,12 +80,17 @@ internal class RumApplicationScopeTest {
     @BoolForgery
     var fakeBackgroundTrackingEnabled: Boolean = false
 
+    @BoolForgery
+    var fakeTrackFrustrations: Boolean = true
```
0xnm, line 84: same
```suggestion
    lateinit var fakeTrackFrustrations
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumContinuousActionScopeTest.kt
```diff
@@ -129,7 +129,8 @@ internal class RumContinuousActionScopeTest {
             TEST_INACTIVITY_MS,
             TEST_MAX_DURATION_MS,
             mockRumEventSourceProvider,
-            fakeAndroidInfoProvider
+            fakeAndroidInfoProvider,
+            true
```
0xnm, line 133: we probably can inject forged value? and below as well. Or is there any reason we want only `true`?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -106,6 +106,9 @@ internal class RumSessionScopeTest {
     @BoolForgery
     var fakeBackgroundTrackingEnabled: Boolean = false
 
+    @BoolForgery
+    var fakeTrackFrustrations: Boolean = true
```
0xnm, line 110:
```suggestion
    lateinit var fakeTrackFrustrations
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -362,6 +363,7 @@ internal class RumViewManagerScopeTest {
         testedScope = RumViewManagerScope(
             parentScope = mockParentScope,
             backgroundTrackingEnabled = false,
+            trackFrustrations = true,
```
0xnm, line 366: here and below we need to inject forged value instead. `backgroundTrackingEnabled` is set to the hardcoded value just because the test requires that value.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -175,6 +175,9 @@ internal class RumViewScopeTest {
     @Mock
     lateinit var mockViewUpdatePredicate: ViewUpdatePredicate
 
+    @BoolForgery
+    var fakeTrackFrustrations = true
```
0xnm, line 179:
```suggestion
    lateinit var fakeTrackFrustrations: Boolean
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitorTest.kt
```diff
@@ -133,13 +133,17 @@ internal class DatadogRumMonitorTest {
     @BoolForgery
     var fakeBackgroundTrackingEnabled: Boolean = false
 
+    @BoolForgery
+    var fakeFrustrationTracking: Boolean = true
```
0xnm, line 137:
```suggestion
    lateinit var fakeFrustrationTracking: Boolean
```

0xnm (changes requested):
Overall lgtm! I just added few minor suggestions.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScopeTest.kt
plousada: We can. The issue is that we have several tests with code blocks similar to this: 
```
if (fakeType == RumActionType.TAP) {
        hasFrustration(ActionEvent.Type.ERROR_TAP)
    } else {
        hasNoFrustration()
    }
}
```
We can either keep this value as true and not alter them, or use a forged value and change the if condition to something like `fakeTrackFrustraions && fakeType == RumActionType.TAP`.

What do you think is best?

0xnm: ok, makes sense. Let's keep it as `true` by default then.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -27,6 +27,7 @@ internal class RumSessionScope(
     private val parentScope: RumScope,
     internal val samplingRate: Float,
     internal val backgroundTrackingEnabled: Boolean,
+    internal val trackFrustrations: Boolean,
```
mariusc83, line 30: Lol, another parameter in the Constructor :D. We really need to find a scalable solution for this @xgouchet @0xnm.

mariusc83 (approved):
LGTM

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
mariusc83: Yes but then you will move the problem in the container (context). I guess it will be much more cleaner though as we only change it there and not in all the scopes constructor.

xgouchet (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScopeTest.kt
0xnm, line 104: seems this is the only thing remaining to address in the test, it should be:

```suggestion
    lateinit var fakeTrackFrustrations: Boolean
```

same for other similar occurrences in other files.

0xnm (changes requested):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScopeTest.kt
plousada: The change makes sense, but `lateinit` isn't allowed on primitive type as far as I could see. 
(I'm still not the most experienced kotlin dev)

0xnm: yes, you are right :) I missed that

0xnm (approved):

PR #1087, opened 2022-10-14
RUMM-2331: Simple message bus for cross-feature communication

0xnm:
### What does this PR do?

This change introduces the way to send messages to the particular feature which is useful in case if this feature is an optional dependency.

This is achieved by using `FeatureScope#sendEvent` method which just dispatches the event to the registered receiver for the given feature.

I didn't find the need for the broadcast message and also since events are always targeting the particular feature the host of the `sendEvent` method is not `SdkCore`, but `FeatureScope`. This removes the need for the `target` argument of the `sendEvent` method, because we always target the feature by doing `SdkCore#getFeature` which returns `FeatureScope`.

Message bus in the current implementation doesn't add any thread model: event will be received on the same thread as it was sent.


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
```diff
@@ -465,5 +483,10 @@ internal class DatadogCore(
                 "Credentials" +
                 "(\"<CLIENT_TOKEN>\", \"<ENVIRONMENT>\", \"<VARIANT>\", \"<APPLICATION_ID>\")\n" +
                 "Datadog.initialize(context, credentials, configuration, trackingConsent);"
+
+        internal const val MISSING_FEATURE_FOR_EVENT_LISTENER =
+            "Cannot add event receiver for feature \"%s\", it is not registered."
+        internal const val EVENT_LISTENER_ALREADY_EXISTS =
```
xgouchet, line 489: NitPick, should be named `EVENT_RECEIVER_ALREADY_EXISTS`

xgouchet (approved):

plousada (approved):
LGTM

mariusc83 (approved):

PR #1092, opened 2022-10-18
RUMM-2289 Add tracing sampling attribute

xgouchet:
### What does this PR do?

Add the `_dd.rule_psr` attribute for RUM Resources to let customer know the sampling rate in APM

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassDeserializerGenerator.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassDeserializerGenerator.kt
@@ -167,7 +167,7 @@ class ClassDeserializerGenerator(
     ) {
         val opt = if (nullable) "?" else ""
         beginControlFlow(
-            "$assignee = $getter$opt.asJsonArray$opt.let { %L ->",
+            "$assignee = $getter$opt.asJsonArray$opt.let·{·%L·->",
             Identifier.PARAM_JSON_ARRAY
         )
         val collectionClassName: ClassName = if (arrayType.uniqueItems) {
--- a/dd-sdk-android/src/main/json/rum/_action-child-schema.json
+++ b/dd-sdk-android/src/main/json/rum/_action-child-schema.json
@@ -1,0 +1,34 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/_action-child-schema.json",
+  "title": "ActionChildProperties",
+  "type": "object",
+  "description": "Schema of all properties of events that can have parent actions",
+  "properties": {
+    "action": {
+      "type": "object",
+      "description": "Action properties",
+      "required": ["id"],
+      "properties": {
+        "id": {
+          "description": "UUID of the action",
+          "readOnly": true,
+          "oneOf": [
+            {
+              "type": "string",
+              "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
+            },
+            {
+              "type": "array",
+              "items": {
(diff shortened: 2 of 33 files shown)


ncreated (approved):
Looks good for the sampling logic 👍.

dd-sdk-android/src/main/json/rum/_action-child-schema.json
```diff
@@ -0,0 +1,34 @@
...
+        "id": {
+          "description": "UUID of the action",
+          "readOnly": true,
+          "oneOf": [
+            {
+              "type": "string",
+              "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
+            },
+            {
+              "type": "array",
```
mariusc83, line 22: are we using this event ? I don't remember to have something like this yet in the SDK.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassDeserializerGenerator.kt
```diff
@@ -167,7 +167,7 @@ class ClassDeserializerGenerator(
     ) {
         val opt = if (nullable) "?" else ""
         beginControlFlow(
-            "$assignee = $getter$opt.asJsonArray$opt.let { %L ->",
+            "$assignee = $getter$opt.asJsonArray$opt.let·{·%L·->",
```
0xnm, line 170: why there is a change with middle dot instead of space here and in `ClassGenerator`?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -362,7 +362,7 @@ internal open class RumViewScope(
                 type = errorType,
                 sourceType = event.sourceType.toSchemaSourceType()
             ),
-            action = context.actionId?.let { ErrorEvent.Action(it) },
+            action = context.actionId?.let { ErrorEvent.Action(listOf(it)) },
```
mariusc83, line 365: why are we doing this ? Is there any situation where a `LongTask` will be caused by multiple actions ? Same question for all the other scopes.

0xnm (approved):
lgtm! I added one question regarding unexpected symbol.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassDeserializerGenerator.kt
xgouchet: This is part of KotlinPoet internals to mark non-breaking space. This ensures that the KotlinPoet file writer won't break a line on any of those space. 
This is because some classNames get very large and the line breaking happens at an arbitrary size without knowledge of the syntax, leading to code that don't compile

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
xgouchet: This can happen with frustration signals. The schema makes it so that an `action.id` can be either a string, or a list of string. Because we can't have such a thing in Kotlin, we go for the list of string which allows us to report both a single or multiple ids

dd-sdk-android/src/main/json/rum/_action-child-schema.json
xgouchet: This schema is shared for all action children (i.e. Resources, Errors and Long Tasks). we used to only have an `action.id: String`, but now it can be either a String, or an array of Strings (for Frustration Signals)

mariusc83 (approved):

PR #1093, opened 2022-10-18
Publish SNAPSHOT builds to sonatype on pushes to develop

fuzzybinary:
### What does this PR do?

Add 'SNAPSHOT' as a version tag, and default new versions to it. Change the build so that `develop` triggers all `publish` stages.

### Motivation

The cross platform frameworks would like to work from SNAPSHOTS in order to speed up development.

### Additional Notes

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -181,6 +181,7 @@ publish:release:
   tags: [ "runner:main" ]
   only:
     - tags
+    - develop
   image: $CI_IMAGE_DOCKER
   stage: publish
   timeout: 30m
@@ -196,6 +197,7 @@ publish:release-coil:
   tags: [ "runner:main" ]
   only:
     - tags
+    - develop
   image: $CI_IMAGE_DOCKER
   stage: publish
   timeout: 30m
@@ -211,6 +213,7 @@ publish:release-compose:
   tags: [ "runner:main" ]
   only:
     - tags
+    - develop
   image: $CI_IMAGE_DOCKER
   stage: publish
   timeout: 30m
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -317,6 +326,7 @@ publish:release-timber:
   tags: [ "runner:main" ]
   only:
     - tags
+    - develop
```
mariusc83, line 329: if you add this here, doesn't this mean that this task will always be performed when develop builds ? That means we are going to push snapshots every time we merge something in develop, is that correct ?

xgouchet: Yes that's the goal

xgouchet (approved):

buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
```diff
@@ -17,7 +17,7 @@ object AndroidConfig {
     const val MIN_SDK_FOR_COMPOSE = 21
     const val BUILD_TOOLS_VERSION = "31.0.0"
 
-    val VERSION = Version(1, 15, 0, Version.Type.Dev)
+    val VERSION = Version(1, 15, 0, Version.Type.Snapshot)
```
0xnm, line 20: is it enough to publish the snapshot on Sonatype? will it skip staging/approval step if version has `-SNAPSHOT` suffix? Or will it be some other PR?

mariusc83 (approved):

buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
xgouchet: Sonatype will automatically skip staging based on the `-SNAPSHOT` suffix

xgouchet (approved):

PR #1094, opened 2022-10-19
RUMM-2672: Make Storage#writeCurrentBatch async

0xnm:
### What does this PR do?

This PR may look big, but essentially it does the following:

* `Storage#writeCurrentBatch` implementation is now async (`ConsentAwareStorage` takes an executor), meaning I/O access inside this method is happening on the worker thread and callback is invoked on the worker thread as well.
* Extracted anonymous implementation of `EventBatchWriter` into a dedicated class (and moved tests as well), because `ConsentAwareStorage` and its tests were too big already. Introduced also a no-op implementation of writer if there is no batch file or consent is not granted - this is where the most of the diff comes from.

This change is needed to make `FeatureScope#withWriteContext` truly async to be able to implement event write context further.


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -123,11 +123,11 @@ internal class SdkFeature(
 
     // region FeatureScope
 
-    // TODO RUMM-0000 there is no thread switch here, it stays the same.
-    // Need to clarify the threading. We either switch thread here or in Storage.
-    // Or give the ability to specify the executor to the caller.
-    @Suppress("ThreadSafety")
     override fun withWriteContext(callback: (DatadogContext, EventBatchWriter) -> Unit) {
+        // TODO RUMM-0000 thread safety. Thread switch happens in Storage right now. Open questions:
+        // * what if called wants to have a sync operation, without thread switch
```
mariusc83, line 128: I am thinking when will be the case/need for that. Do you have some situations in mind ?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/ConsentAwareStorage.kt
```diff
@@ -39,56 +42,46 @@ internal class ConsentAwareStorage(
...
-                ) {
-                    if (newMetadata?.isNotEmpty() == true && metadataFile != null) {
-                        writeBatchMetadata(metadataFile, newMetadata)
-                    }
-                    listener.onDataWritten(eventId)
+                val writer = if (orchestrator == null || batchFile == null) {
+                    NoOpEventBatchWriter()
                 } else {
-                    listener.onDataWriteFailed(eventId)
+                    FileEventBatchWriter(
```
mariusc83, line 65: are we creating a writer every time we are writing something ?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/Storage.kt
```diff
@@ -16,13 +18,17 @@ import com.datadog.tools.annotation.NoOpImplementation
 @NoOpImplementation
 internal interface Storage {
 
+    // TODO RUMM-0000 It seems only write part is useful to be async. Having upload part as async
```
mariusc83, line 21: I do agree

mariusc83 (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/ConsentAwareStorage.kt
0xnm: yes, because writer is associated with a particular batch file

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
0xnm: Crash reporting - we need write it as log or RUM event in a blocking manner.

mariusc83: true, forgot about that.

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/ConsentAwareStorage.kt
mariusc83: Damn...I am thinking that for SR I will have to write tons of events....meaning that we will be creating those writers all the time. Aren't we afraid of too many objects there ? I mean I know they are not so heavy but was thinking that maybe if we could do it differently 🤔

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/ConsentAwareStorage.kt
0xnm: We can profile it. Anyway, the number of these objects will be much less that number of snapshot/wireframes objects to write :)

PR #1100, opened 2022-10-24
Deprecate DatadogPlugin class and its usage

0xnm:
### What does this PR do?

This was done already in SDK v2 branch, just doing the same in the `develop`, so that it can be in the 1.x releases.


mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -341,7 +343,14 @@ internal constructor(
          * @see [Feature.Tracing]
          * @see [Feature.RUM]
          */
-        fun addPlugin(plugin: DatadogPlugin, feature: PluginFeature): Builder {
+        @Deprecated(
+            "Datadog Plugins won't work in SDK v2, you'll need to write" +
+                " your own Feature"
```
xgouchet, line 348:
```suggestion
            "Datadog Plugins will be removed in SDK v2.0.0." +
                "You will then need to write your own Feature (check our own code for guidance)."
```

dd-sdk-android/src/main/kotlin/com/datadog/android/plugin/DatadogPlugin.kt
```diff
@@ -16,6 +16,7 @@ import com.datadog.android.privacy.TrackingConsentProviderCallback
  * @see [Feature.TRACE]
  * @see [Feature.RUM]
  */
+@Deprecated("Datadog Plugins won't work in SDK v2, you'll need to write your own Feature")
```
xgouchet, line 19:
```suggestion
            "Datadog Plugins will be removed in SDK v2.0.0." +
                "You will then need to write your own Feature (check our own code for guidance)."
```

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -89,6 +91,7 @@ internal abstract class SdkFeature<T : Any, C : Configuration.Feature> {
         }
     }
 
+    @Deprecated("Plugins won't work that way in SDK v2")
```
xgouchet, line 94:
```suggestion
            "Datadog Plugins will be removed in SDK v2.0.0." +
                "You will then need to write your own Feature (check our own code for guidance)."
```

xgouchet (commented):
It'd be nice to also call the `warnDeprecated(…)` method. The deprecated annotation won't necessarily be seen and this will also write a warning in Logcat.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
0xnm: The problem is that there is no code to check with the guidance, it will be available (with all the public API changes) in version 2.0.0, so that it means we cannot remove things related to the `DatadogPlugin` in 2.0.0 and should do it later at some point. WDYT? In theory it is not a problem to keep `DatadogPlugin` a bit longer. Or we don't add the statement about checking the code.

xgouchet (approved):

PR #1105, opened 2022-10-25
RUMM-2687 Invert frame time to get js refresh rate

louiszawadzki:
### What does this PR do?

Let the RN SDK report the frame time, then invert it when reporting to backend to report frame rate.

### Motivation

Get a more sensible for the average js refresh rate by getting the harmonic mean instead of the arithmetic mean.
See https://gamedev.net/forums/topic/628085-math-fail-adding-frames-per-second/4960225/ and https://en.wikipedia.org/wiki/Harmonic_mean#Average_speed to understand why this makes more sense.

### Additional Notes


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -918,3 +918,15 @@ private fun VitalInfo.toPerformanceMetric(): ViewEvent.FlutterBuildTime {
         average = meanValue
     )
 }
+
+private fun VitalInfo.toInversePerformanceMetric(): ViewEvent.FlutterBuildTime {
+    return ViewEvent.FlutterBuildTime(
+        min = invertValue(maxValue),
```
mariusc83, line 924: why are you switching `min` with `max` here ?

mariusc83: Maybe a comment to explain that would be nice here. It will be very hard to remember why we are doing this.

mariusc83 (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
mariusc83: I got that...what is hard for me to understand is that right now we are inverting the `max` value to create the `min`. But is this relevant ?

mariusc83 (approved):

PR #1106, opened 2022-10-26
RUMM-2608: Use event write context for Traces

0xnm:
### What does this PR do?

This PR brings the usage of event write context (which can be retrieved by calling `withWriteContext`) for traces.

It is done by changing the `TraceWriter` so that it requests event write context for the collection of `DDSpan` to write.

EWC will be requests only once, so that DD context is the same for all the spans to be written.

Also introduced `ContextAwareMapper` and `ContextAwareSerializer`, because both mapping (from legacy `DDSpan` to `SpanEvent` and serialization) require `DatadogContext` access, and options like to create them each time inside EWC or use `Pair<DatadogContext, T>` as argument seemed less reasonable to me.


mariusc83 (approved):
Nicely done. Lgtm !!

xgouchet (approved):

PR #1111, opened 2022-10-28
Fix caching of models generation task

0xnm:
### What does this PR do?

This change fixes the caching of `GenerateJsonSchemaTask` task. Before it was reading input files by calling `listFiles` for the given input dir, which will return only top-level files.

But it can be the case that top-level files are actually only the reference to the files deeper, like this:

```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "session-replay-mobile-schema.json",
  "$ref": "session-replay/mobile/segment-schema.json"
}
```

It means that top-level file will receive no change when referenced file `session-replay/mobile/segment-schema.json` actually changes, making Gradle think that inputs didn't change and skipping models generation.

With this change we track all files recursively with Gradle, but take only top-level files for the model generation.

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/GenerateJsonSchemaTask.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/GenerateJsonSchemaTask.kt
@@ -36,13 +36,21 @@ open class GenerateJsonSchemaTask : DefaultTask() {
     // region Input/Output
 
     /**
-     * The [InputFiles] (E.g.: all the json files in `resources/json`).
+     * The [InputFiles] (E.g.: all the json files in `resources/json`). Note, that it will get all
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/GenerateJsonSchemaTask.kt
```diff
@@ -95,7 +103,10 @@ open class GenerateJsonSchemaTask : DefaultTask() {
     fun performTask() {
         val inputDir = getInputDir()
         val outputDir = getOutputDir()
-        val files = getInputFiles().filter { it.name !in ignoredFiles }
+        val files = getInputFiles()
```
mariusc83, line 106: I guess it can work...we are forcing it to only the root files

mariusc83 (approved):

PR #1117, opened 2022-10-31
RUMM-2602: Use event write context in RUM

0xnm:
### What does this PR do?

This change brings the usage of event write context (EWC) to RUM (and to Web View RUM) and is similar to the move done in #1106, #1103.

Basically to write RUM events we now first request EWC via `FeatureScope#withWriteContext` call. This call is done at the very last point, so that the switch to the worker thread (where the closure is executed) doesn't affect the sequential flow and no changes to the shared state is done from the closure context.

There are some things which still keep RUM bound to the crash reporting or to the core feature (like the need to get server time offset outside of EWC, etc.), they will be addressed in the other PRs.


mariusc83 (approved):
LGTM

xgouchet (approved):

PR #1118, opened 2022-11-02
RUMM-2630 RUM Configuration Telemetry

xgouchet:
### What does this PR do?

Implement a RUM Configuration Telemetry Event.


dd-sdk-android/src/test/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandlerTest.kt
```diff
@@ -141,22 +163,270 @@ internal class TelemetryEventHandlerTest {
...
+        val rumContext = GlobalRum.getRumContext()
+
+        // When
+        testedTelemetryHandler.handleEvent(configRawEvent, mockWriter)
+
+        // Then
+        argumentCaptor<TelemetryConfigurationEvent> {
+            verify(mockWriter).write(capture())
+            assertConfigEventMatchesRawEvent(firstValue, configRawEvent, rumContext)
+            assertThat(firstValue)
```
0xnm, line 187: should it have the continuation or be removed?

0xnm (approved):
looks nice! I've added some comments, but they are not blocking.

fuzzybinary:
What do we think the process will be for the Cross Platform frameworks to set their configuration properties? It looks like the current event just takes the Configuration which won't have some of the cross platform information available.

mariusc83 (approved):
Good job. LGTM !!

fuzzybinary:
@xgouchet 

> The idea will be to embed some configuration attributes in the `additionalConfig` property, and read them in the TelemetryHandler (e.g.: `trackFlutterPerformance = configuration.additionalConfig["_dd.configuration.trackFlutterPerformance"]`)

I took a different approach on iOS, using a mapper style API instead. The problem is, we don't know the value of a lot of theses during the first call for initialization, so we can't send them in the additional config.

See https://github.com/DataDog/dd-sdk-ios/pull/1039/files#diff-dbaa857f0082fd2cf6795c86b222655eb9d9aafe9e5efd8a5d472891bb5c0e8fR177

xgouchet:
@fuzzybinary 
Indeed that would be a possible approach too. I think it'll be easy to do one or the other in another PR when needed.

plousada (approved):

PR #1120, opened 2022-11-02
Remove Flutter from Dogfooding scripts

fuzzybinary:
### What does this PR do?

Removes all dogfooding steps that push new versions to the Flutter repo.

### Motivation

Flutter is now developing off source and will "pull" new versions when it deploys. This removes the Android step of "pushing" the new releases as that is no longer necessary

### DIFF
--- a/dogfood.py
+++ b/dogfood.py
@@ -21,18 +21,13 @@ TARGET_APP = "app"
 TARGET_DEMO = "demo"
 TARGET_BRIDGE = "bridge"
 TARGET_GRADLE_PLUGIN = "gradle-plugin"
-TARGET_FLUTTER = "flutter"
 
 REPOSITORIES = {
     TARGET_APP: "datadog-android",
     TARGET_DEMO: "shopist-android",
     TARGET_BRIDGE: "dd-bridge-android",
     TARGET_GRADLE_PLUGIN: "dd-sdk-android-gradle-plugin",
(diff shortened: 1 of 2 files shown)


dogfood.py
```diff
@@ -21,34 +21,26 @@
...
 TARGET_BRIDGE = "bridge"
 TARGET_GRADLE_PLUGIN = "gradle-plugin"
-TARGET_FLUTTER = "flutter"
 
 REPOSITORIES = {
     TARGET_APP: "datadog-android",
     TARGET_DEMO: "shopist-android",
     TARGET_BRIDGE: "dd-bridge-android",
     TARGET_GRADLE_PLUGIN: "dd-sdk-android-gradle-plugin",
-    TARGET_FLUTTER: "dd-sdk-flutter"
```
0xnm, line 31: let's add the comment why flutter doesn't need to be presented in this script

fuzzybinary: I can do that -- should we add a comment about React Native too?

Actually, I don't think @louiszawadzki is using the bridge anymore, should we remove the bridge from this update as well?

0xnm: I think comment will be useful. At least me personally - I will forget quickly why Flutter is not here :)

For the React Native we have a dedicated task to update this script, so it will be done later.

xgouchet (approved):

0xnm (approved):

mariusc83 (approved):

PR #1121, opened 2022-11-02
RUMM-2700: Use event write context in WebView Logs

0xnm:
### What does this PR do?

This change brings usage of event write context (EWC) to the WebView Logs feature. The approach is similar to what was done before in https://github.com/DataDog/dd-sdk-android/pull/1117, https://github.com/DataDog/dd-sdk-android/pull/1106, https://github.com/DataDog/dd-sdk-android/pull/1103, etc.

There is one non-related change though: in `WebViewLogEventConsumer` the order of event processing is changed: now it is first checked for the proper type, then mapped instead of the opposite (otherwise we may map events which are not going to be written).


dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
```diff
@@ -63,11 +62,11 @@ internal class WebViewRumEventConsumer(
...
     }
 
-    private fun getOffset(viewId: String): Long {
+    private fun getOffset(viewId: String, datadogContext: DatadogContext): Long {
         var offset = offsets[viewId]
         if (offset == null) {
-            offset = timeProvider.getServerOffsetMillis()
-            offsets[viewId] = offset
+            offset = datadogContext.time.serverTimeOffsetMs
+            synchronized(offsets) { offsets[viewId] = offset }
```
mariusc83, line 69: why do you use `synchronized` block here ?

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
```diff
@@ -76,8 +75,10 @@ internal class WebViewRumEventConsumer(
     private fun purgeOffsets() {
         while (offsets.entries.size > MAX_VIEW_TIME_OFFSETS_RETAIN) {
             try {
-                val viewId = offsets.entries.first()
-                offsets.remove(viewId.key)
+                synchronized(offsets) {
```
mariusc83, line 78: same here

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
0xnm: because `offsets` field can be accessed from the different threads (it gets called from EWC, which can be invoked on the different worker threads), so we need to make the work with this field thread-safe.

The work happens on the worker thread anyway, so synchronisation penalty is not important.

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
mariusc83: Yeah, I had that one before but then looked under the hood and noticed that Android uses a SingleThreadExecutor for the JS bridge, I tested it also and made sure we only have one thread there.

mariusc83: I see the reason now actually with the new architecture, good point.

mariusc83 (approved):

PR #1125, opened 2022-11-04
RUMM-2684: Observe uncaught exception in executors

0xnm:
### What does this PR do?

Turns out we were swallowing uncaught exceptions raised during the execution of the tasks submitted via `Executor#execute`, `ExecutorService#submit` or `ScheduledExecutorService#schedule` calls, because the behavior is like the following:

* If uncaught exception is raised for task submitted via `Executor#execute`, then worker thread is killed and exception is printed to the console.
* If uncaught exception is raised for task submitted via `ExecutorService#submit` or `ScheduledExecutorService#schedule`, then worker thread is kept running and exception is thrown on the thread is which doing `Future#get` call (normally the thread getting the result). We were always ignoring the `Future` returned after the task is submitted, so we were swallowing such exceptions.

This change closes this gap by adding the hook to the `ThreadPoolExecutor#afterExecute` method, which allows to observe such uncaught exceptions. `ThreadPoolExecutor#afterExecute` is called on the thread which did the actual task execution.

The logic of this hook is taken from the [afterExecute documentation](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html#afterExecute(java.lang.Runnable,%20java.lang.Throwable)).

Exceptions will be sent to the `devLogger` and to the internal telemetry as well.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
@@ -44,6 +44,8 @@ import com.datadog.android.core.internal.system.NoOpAndroidInfoProvider
 import com.datadog.android.core.internal.system.NoOpAppVersionProvider
 import com.datadog.android.core.internal.system.NoOpSystemInfoProvider
 import com.datadog.android.core.internal.system.SystemInfoProvider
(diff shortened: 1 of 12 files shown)


mariusc83 (approved):
LGTM

xgouchet (approved):

PR #1131, opened 2022-11-08
Filter out telemetry in the assertions of instrumented RUM tests

0xnm:
### What does this PR do?

Follow-up for #1118: now we have configuration telemetry event [sent with the delay](https://github.com/DataDog/dd-sdk-android/blob/b260f33d8355290d3606f8bab085786a4791d940/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt#L418), which makes RUM instrumented tests flaky, because configuration telemetry event may or may not be in the intercepted RUM requests.

This changes filters out telemetry events from the list of RUM events against which the assertion will be done.


plousada (approved):
I'm assuming it's ok to filter out all telemetry events (`debug` and `error` inc) even when we weren't doing it before.

xgouchet (approved):

PR #1143, opened 2022-11-11
Add a logger method to log error information from strings

fuzzybinary:
### What does this PR do?

Add a logger method to log error information from strings.  This is needed for Flutter and React Native to pass their stack traces to the SDK for serialization.

I added a method to Logger (without adding any additional convenience methods, as Flutter and ReactNative likely won't use them anyway) as well as a new method to the LogHandler interface, which I propagated to all of the implementations.

### Motivation

Cross Platform frameworks needed a way to post errors / stack traces to logs, which we do using Strings.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
@@ -157,7 +157,6 @@ internal constructor(internal var handler: LogHandler) {
      * @param throwable a (nullable) throwable to be logged with the message
      * @param attributes a map of attributes to include only for this message. If an attribute with
      * the same key already exist in this logger, it will be overridden (just for this message)
-     *
      */
     @JvmOverloads
     fun log(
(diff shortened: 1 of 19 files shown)


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
```diff
@@ -169,6 +168,30 @@ internal constructor(internal var handler: LogHandler) {
...
+     * the same key already exist in this logger, it will be overridden (just for this message)
+     */
+    @JvmOverloads
+    @Suppress("LongParameterList")
+    fun log(
+        priority: Int,
+        message: String,
+        errorKind: String?,
+        errorMessage: String?,
+        errorStack: String?,
```
0xnm, line 189: minor suggestion to have the consistency with the similar method in `RumMonitor`

```suggestion
        errorStacktrace: String?,
```

dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
```diff
@@ -169,6 +168,30 @@ internal constructor(internal var handler: LogHandler) {
         internalLog(priority, message, throwable, attributes)
     }
 
+    /**
+     * Sends a log message with strings for error information.
```
0xnm, line 172: I'm wondering if we should add the note here that this method is not supposed to be used in native apps directly. My motivation is that is this method is misused in the native apps and stack trace is not a valid JVM stacktrace, then we will have a growth of the de-obfuscation call errors, which may be confusing for the infra monitoring. WDYT?

But anyway we don't have this for the `RumMonitor` counterpart.

0xnm (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
fuzzybinary: Yes, I'll add a note to prefer the other methods.

We do / should have a counterpart in RumMonitor? We have a way to send a string stackTrace anyway....

0xnm: Yes, I mean we don't have the sentence saying that it is mostly for the internal usage in case of `RumMonitor`'s counterpart.

fuzzybinary: Ah, I'll add it in both places.

0xnm (approved):
nice!

PR #1146, opened 2022-11-16
RUMM-2713: Move global RUM context into generic feature context storage

0xnm:
### What does this PR do?

This change removes global RUM context from `GlobalRum` singleton by moving it to the feature's context (`DatadogContext#featuresContext`).

Now RUM context will be stored in the generic dictionary `Map<String, Any?>` and can be accessed either by calling `DatadogContext#featuresContext` or by calling `SdkCore#getFeatureContext`.

Also `SdkCore#setFeatureContext` is removed in the favour of `SdkCore#updateFeatureContext` method which gives a mutable dictionary for the update, final update is thread-safe.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
@@ -25,6 +25,8 @@ import com.datadog.android.core.internal.utils.sdkLogger
 import com.datadog.android.core.model.UserInfo
 import com.datadog.android.error.internal.CrashReportsFeature
 import com.datadog.android.log.internal.LogsFeature
+import com.datadog.android.plugin.DatadogContext
+import com.datadog.android.plugin.DatadogRumContext
 import com.datadog.android.privacy.TrackingConsent
 import com.datadog.android.rum.GlobalRum
 import com.datadog.android.rum.internal.RumFeature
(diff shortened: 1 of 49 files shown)


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
```diff
@@ -183,14 +185,33 @@ internal class DatadogCore(
...
+            // workaround for the backward compatibility with DatadogPlugin, we don't want to have
+            // context update in plugins in the synchronized block
+            val updatedContext = synchronized(feature) {
+                val featureContext = it.getFeatureContext(featureName)
+                val mutableContext = featureContext.toMutableMap()
+                updateCallback(mutableContext)
+                it.setFeatureContext(featureName, mutableContext)
+                mutableContext
+            }
+            if (featureName == RumFeature.RUM_FEATURE_NAME) {
```
mariusc83, line 203: why are we doing this only for the `RUM` feature name ?

0xnm: Because context which is received by `DatadogPlugin` contains [only RUM-related info](https://github.com/DataDog/dd-sdk-android/blob/feature/sdkv2/dd-sdk-android/src/main/kotlin/com/datadog/android/plugin/DatadogContext.kt), so it makes sense to update it only when RUM context is updated.

Anyway, `DatadogPlugin` will be removed in v2. This code won't live long.

mariusc83 (approved):
LGTM

PR #1148, opened 2022-11-17
Add fromJsonElement method to generated models

fuzzybinary:
### What does this PR do?

Add a method to the model generator to change the current `fromJson` method to parse the JSON, then pass the result to `fromJsonElement`. Additionally, change all of the methods that used `fromJson` for nested objects to instead use `fromJsonElement` for efficiency.

### Motivation

This should be more for nested models, and makes it easier for Flutter to build the models from the Maps we'll get back from the method channel. Otherwise, Flutter will need to convert the object from a map to a JSON string, then ask the model to parse it back out.


buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassJsonElementDeserializerGenerator.kt
```diff
@@ -27,17 +27,17 @@ class ClassDeserializerGenerator(
         val isConstantClass = definition.isConstantClass()
         val returnType = ClassName.bestGuess(definition.name)
 
-        val funBuilder = FunSpec.builder(Identifier.FUN_FROM_JSON)
+        val funBuilder = FunSpec.builder(Identifier.FUN_FROM_JSON_ELT)
```
0xnm, line 30: Should we call the function as `fromJsonObject` instead given that it takes `JsonObject` and not `JsonElement`?

`JsonElement` is an abstract class for `JsonArray`, `JsonNull`, `JsonPrimitive` and `JsonObject`.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassJsonElementDeserializerGenerator.kt
```diff
@@ -202,10 +194,10 @@ class ClassDeserializerGenerator(
             )
             is TypeDefinition.OneOfClass,
             is TypeDefinition.Class -> addStatement(
-                "%L.add(%T.%L(it.toString()))",
+                "%L.add(%T.%L(it as JsonObject))",
```
0xnm, line 197: If `it` is not `JsonObject` here then `ClassCastException` will be thrown which is not caught by the `catch` blocks. Alternatively we can use `.asJsonObject` call which will throw `IllegalStateException` which is caught.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassJsonElementDeserializerGenerator.kt
```diff
@@ -251,7 +243,7 @@ class ClassDeserializerGenerator(
         rootTypeName: String
     ) {
         val opt = if (nullable) "?" else ""
-        beginControlFlow("$assignee = $getter$opt.toString()$opt.let")
+        beginControlFlow("$assignee = ($getter as$opt JsonObject)$opt.let")
```
0xnm, line 246: same: I think it is better to use `asJsonObject` here, different exception will be thrown. And below on the L268 as well.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ClassStringDeserializerGenerator.kt
```diff
@@ -0,0 +1,46 @@
...
+            .addAnnotation(AnnotationSpec.builder(JvmStatic::class).build())
+            .returns(returnType)
+
+        if (!isConstantClass) {
+            funBuilder.throws(ClassNameRef.JsonParseException)
+            funBuilder.addParameter(Identifier.PARAM_JSON_STR, STRING)
+        }
+
+        funBuilder.addStatement(
+            "val %L = %T.parseString(%L).asJsonObject",
```
0xnm, line 37: I think we should wrap this call in the `try-catch` block which will throw `JsonParseException`, because otherwise `IllegalStateException` from `asJsonObject` call will propagate much higher in the stack.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/MultiClassGenerator.kt
```diff
@@ -77,19 +80,42 @@ class MultiClassGenerator(
...
+        val returnType = ClassName.bestGuess(definition.name)
+
+        val funBuilder = FunSpec.builder(Identifier.FUN_FROM_JSON)
+            .addAnnotation(AnnotationSpec.builder(JvmStatic::class).build())
+            .throws(ClassNameRef.JsonParseException)
+            .addParameter(Identifier.PARAM_JSON_STR, STRING)
+            .returns(returnType)
+
+        funBuilder.addStatement(
+            "val %L = %T.parseString(%L).asJsonObject",
```
0xnm, line 100: here as well: try-catch is needed to guard `asJsonObject` call.

xgouchet (approved):

0xnm (approved):

0xnm (approved):

0xnm (approved):

PR #1153, opened 2022-11-22
RUMM-2753: Improve SDK performance a bit

0xnm:
### What does this PR do?

* Wraps OS/device info in `lazy`, because it shouldn't be changed during the app lifetime anyway. `PUBLICATION` mode is used for `lazy` which is using `compareAndSet` vs synchronisation by default (see docs for the difference), which should be a bit faster.
* `SdkCore` gets `time` property in order to avoid pulling the whole `DatadogContext` when only time piece is needed. I had to implement `NoOpSdkCore` by hand because of that, because no-op generation doesn't support properties and anyway if it would it cannot know what default value of `TimeInfo` to create.
* Usage of `CoreFeature` is completely removed from the `SessionReplay` feature.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
@@ -9,7 +9,6 @@ package com.datadog.android.sessionreplay.internal
 import android.app.Application
 import android.content.Context
 import com.datadog.android.core.configuration.Configuration
-import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.utils.devLogger
 import com.datadog.android.sessionreplay.LifecycleCallback
 import com.datadog.android.sessionreplay.RecordWriter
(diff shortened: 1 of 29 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -32,7 +30,7 @@ internal class SessionReplayFeature(
                 SessionReplayRumContextProvider(sdkCore),
                 configuration.privacy,
                 recordWriter,
-                SessionReplayTimeProvider(coreFeature.contextProvider),
+                SessionReplayTimeProvider(sdkCore),
```
mariusc83, line 33: I did use the `sdkCore` before here but I had problems as it was using the `NoOpSdkCore`. For some reason when using the lazy instantiation it was keeping the initial instance which was the `NoOpSdkCore`. Can you please verify that is not the case now.

mariusc83 (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
0xnm: It is all good, `SessionReplayFeature` is created from `DatadogCore` only:

image

No different from the `SessionReplayRumContextProvider` anyway which also takes `sdkCore`.

mariusc83 (approved):

xgouchet (approved):

PR #1167, opened 2022-11-29
Fix NPE on Flutter Unit Tests

fuzzybinary:
### What does this PR do?

Because Flutter flushes during unit tests, the Build properties used in the DefaultAndroidInfoProvider are accessed and they return null, causing an NPE. This checks if the properties are null and returns empty strings instead.


mariusc83 (approved):
Damn...we were trying to fix those yesterday while applying the fix for the dependency.

xgouchet (approved):

fuzzybinary:
After discussion we're not going to modify this. Flutter will use `stop` over attempting to flush and shut down.

PR #1171, opened 2022-12-05
Make a local copy of tags before creating LogEvent

0xnm:
### What does this PR do?

We need to make a copy of `tags`, because by the time we create `LogEvent` (on the worker thread) it may have been changed by another thread.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
@@ -43,7 +43,7 @@ class Logger
 internal constructor(internal var handler: LogHandler) {
 
     private val attributes = ConcurrentHashMap<String, Any?>()
-    private val tags = CopyOnWriteArraySet<String>()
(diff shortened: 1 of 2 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt
```diff
@@ -590,7 +590,10 @@ internal constructor(internal var handler: LogHandler) {
         val combinedAttributes = mutableMapOf<String, Any?>()
         combinedAttributes.putAll(attributes)
         combinedAttributes.putAll(localAttributes)
-        handler.handleLog(level, message, throwable, combinedAttributes, tags, timestamp)
+        // need to make a copy, because the content will be access on another thread and it
+        // can change by then
+        val tagsSnapshot = HashSet(tags)
```
fuzzybinary, line 595: This needs to be added to the other internalLog method as well.

fuzzybinary (changes requested):

fuzzybinary (approved):

PR #1174, opened 2022-12-06
RUMM-2767 Add the 'forceNewBatch' option into the FeatureScope

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -56,6 +56,17 @@ internal class BatchFileOrchestrator(
         return reusableFile ?: createNewFile()
     }
 
+    override fun getNewWritableFile(): File? {
```
xgouchet, line 59: instead of making a new method that is almost identical to `getWritableFile`  you could instead do 

```
    @WorkerThread
    override fun getWritableFile(forceNewFile: Boolean = false): File? {
        if (!isRootDirValid()) {
            return null
        }
        deleteObsoleteFiles()
        freeSpaceIfNeeded()
        return if (!forceNewFile){
            getReusableWritableFile() ?: createNewFile()
        } else {
            createNewFile()
        }
    }
```

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
mariusc83: Yes, was not very sure which way to go here. Thanks for sharing your opinion, will change that :)

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -126,17 +126,18 @@ internal class SdkFeature(
 
     // region FeatureScope
 
-    override fun withWriteContext(callback: (DatadogContext, EventBatchWriter) -> Unit) {
+    override fun withWriteContext(
+        forceNewBoolean: Boolean,
```
0xnm, line 130: `forceNewBatch`, maybe? :)

```suggestion
        forceNewBatch: Boolean,
```

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/FileOrchestrator.kt
```diff
@@ -21,11 +21,13 @@ import java.io.File
 internal interface FileOrchestrator {
 
     /**
+     * @param forceNewBatch if `true` will force the orchestrator to start a new file.
```
0xnm, line 24: Mismatch between name in the doc and name in the function signature. I suppose it should be `forceNewFile` here?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/FeatureScope.kt
```diff
@@ -14,10 +14,15 @@ import com.datadog.android.v2.api.context.DatadogContext
 interface FeatureScope {
     /**
      * Utility to write an event, asynchronously.
+     * @param forceNewBoolean if `true` forces the [EventBatchWriter] to write in a new file and
```
0xnm, line 17:
```suggestion
     * @param forceNewBatch if `true` forces the [EventBatchWriter] to write in a new file and
```

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/FeatureScope.kt
```diff
@@ -14,10 +14,15 @@ import com.datadog.android.v2.api.context.DatadogContext
...
+     * not reuse the already existing pending data persistence file. By default it is `false`.
      * @param callback an operation called with an up-to-date [DatadogContext]
      * and an [EventBatchWriter]. Callback will be executed on a worker thread from I/O pool
      */
-    fun withWriteContext(callback: (DatadogContext, EventBatchWriter) -> Unit)
+    fun withWriteContext(
+        forceNewBoolean: Boolean = false,
```
0xnm, line 23:
```suggestion
        forceNewBatch: Boolean = false,
```

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/storage/Storage.kt
```diff
@@ -24,11 +24,17 @@ internal interface Storage {
...
+     * @param forceNewBatch if `true` will force the writer to start a new batch before storing the
+     * data. By default this flag is `false`.
      * @param callback an operation to perform with a [EventBatchWriter] that will target the current
      * writeable Batch
      */
     @AnyThread
-    fun writeCurrentBatch(datadogContext: DatadogContext, callback: (EventBatchWriter) -> Unit)
+    fun writeCurrentBatch(
+        datadogContext: DatadogContext,
+        forceNewBatch: Boolean = false,
```
0xnm, line 35: do we need to have the default value here if in all the calls (apart of the tests) we will get the value explicitly?

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -266,8 +268,10 @@ internal class SdkFeatureTest {
 
     // region FeatureScope
 
-    @Test
+    @ParameterizedTest
```
0xnm, line 271: not sure if we need a parametrized test here, given that we don't assert anything related to the `forceNewBatch` value.

We can simply use `BoolForgery` on the `forceNewBatch` test parameter. And coverage will be the same, because the only place where code path is different depending on the `forceNewBatch` parameter is `BatchFileOrchestrator`.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -291,16 +301,19 @@ internal class SdkFeatureTest {
         )
     }
 
-    @Test
-    fun `𝕄 do nothing 𝕎 withWriteContext(callback) { no Datadog context }`() {
+    @ParameterizedTest
```
0xnm, line 304: Same as above, `BoolForgery` is enough, no need for the parametrized test here and in other test files where `ValueSource` is just true or false.

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestratorTest.kt
```diff
@@ -415,6 +438,29 @@ internal class BatchFileOrchestratorTest {
...
+        // Given
+        assumeTrue(fakeRootDir.listFiles().isNullOrEmpty())
+
+        // When
+        val start = System.currentTimeMillis()
+        val result = testedOrchestrator.getWritableFile(forceNewFile = true)
+        val end = System.currentTimeMillis()
+
+        // Then
+        checkNotNull(result)
```
0xnm, line 454: I think it is better to use assertion method from AssertJ here instead of the Kotlin stdlib

```suggestion
        assertThat(result).isNotNull()
```

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestratorTest.kt
```diff
@@ -415,6 +438,29 @@ internal class BatchFileOrchestratorTest {
...
+    // region getNewWritableFile
+
+    @Test
+    fun `𝕄 return new File 𝕎 getWritableFile() {forceNewFile=true}`() {
+        // Given
+        assumeTrue(fakeRootDir.listFiles().isNullOrEmpty())
+
+        // When
+        val start = System.currentTimeMillis()
+        val result = testedOrchestrator.getWritableFile(forceNewFile = true)
```
0xnm, line 450: I think result with `forceNewFile = false` will be the same, because it is the fresh state of the orchestrator and it will always return the new file.

What we need to do is to call `getWritableFile` twice (one is default call, another with `forceNewFile = true`) and check that second result is not the same as first and was created after the first call.

0xnm (changes requested):
Nice change! Finally we will get rid of the hack for the SR.

I've added some comments, mostly regarding the tests or typos.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
mariusc83: Damn...yes, miss typing

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
mariusc83: Yeah but I rather make sure we test with both true/false value.

0xnm: IMO this would make sense if some assertion depends on that, which is not the case in most of the test. Anyway we already have kinda check of the proper flow by doing `whenever(object.methodCall(parameter))`, because if the parameter is not the one we provided in `whenever`, mock won't work and test will break anyway.

But I'm not pushing for it :)

dd-sdk-android/src/test/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestratorTest.kt
mariusc83: hmm yes, I think I just duplicated that function so I need to change it also in the original.

mariusc83: Yeah I think this is everywhere in this test class and removing it will imply other changes :(. I would rather keep it like this and maybe open a new PR.

0xnm (approved):
lgtm!

PR #1176, opened 2022-12-07
RUMM-2767 Handle SR requests through SessionReplayRequestsFactory

mariusc83:
### What does this PR do?

As a follow up task on the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2714764346/RFC+-+Handle+SR+data+storage+upload) in this PR we are handling the SR data upload requests through the `SessionReplayRequestFactory`. Also we are making sure that the new events are stored in a new batch whenever the RUM `context` changed.

### Motivation

### Additional Notes


dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/RequestBodyFactory.kt
```diff
@@ -0,0 +1,111 @@
...
+    fun create(
+        segment: MobileSegment,
+        serializedSegment: JsonObject
+    ): RequestBody? {
+        return try {
+            // we need to add a new line at the end of each segment for being able to format it
+            // as an Array when read by the player
+            val segmentAsBinary = (serializedSegment.toString() + "\n").toByteArray()
+            buildRequestBody(segment, segmentAsBinary)
+        } catch (e: Throwable) {
```
0xnm, line 33: minor: we can limit it to only `e: Exception` (`Error` is unlikely can be recoverable or even be thrown by the code below) and also move `@Suppress("TooGenericExceptionCaught")` annotation directly to the `catch` block (`catch (@Suppress("TooGenericExceptionCaught" e: Exception)`)

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/RequestBodyFactory.kt
```diff
@@ -0,0 +1,111 @@
...
+            .build()
+    }
+
+    companion object {
+
+        internal const val HEADER_API_KEY = "DD-API-KEY"
+        internal const val HEADER_EVP_ORIGIN = "DD-EVP-ORIGIN"
+        internal const val HEADER_EVP_ORIGIN_VERSION = "DD-EVP-ORIGIN-VERSION"
+        internal const val HEADER_REQUEST_ID = "DD-REQUEST-ID"
+        internal const val QUERY_PARAM_SOURCE = "ddsource"
```
0xnm, line 97: looks like a leftover which can be removed?

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
```diff
@@ -6,41 +6,178 @@
...
+        internal const val ERROR_WRITING_BODY_TO_BUFFER_MESSAGE = "Unable to extract the ByteArray " +
+            "from session replay request body."
+        internal const val ERROR_READING_BODY_TO_BUFFER_MESSAGE = "Unable to read the ByteArray " +
+            "from session replay request body."
+        internal const val HEADER_API_KEY = "DD-API-KEY"
+        internal const val HEADER_EVP_ORIGIN = "DD-EVP-ORIGIN"
+        internal const val HEADER_EVP_ORIGIN_VERSION = "DD-EVP-ORIGIN-VERSION"
+        internal const val HEADER_REQUEST_ID = "DD-REQUEST-ID"
+        internal const val QUERY_PARAM_SOURCE = "ddsource"
+        internal const val QUERY_PARAM_TAGS = "ddtags"
```
0xnm, line 179: those are defined in the `RequestFactory` interface already if needed.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
```diff
@@ -6,41 +6,178 @@
...
-                context,
-                it.first,
-                (it.second.toString() + "\n").toByteArray()
-            )
+        val serializedSegmentPair = batchToSegmentsMapper.map(batchData)
+        if (serializedSegmentPair == null) {
+            // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
+            // the data in the batch is corrupted. We return an empty request just to make
+            // sure the batch will be deleted by the core.
+            return getEmptyRequest()
```
0xnm, line 46: should we maybe change `RequestFactory` signature to allow null return value? Otherwise this case looks like a hack, which will also trigger some telemetry.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
```diff
@@ -6,41 +6,178 @@
...
-                DataOkHttpUploaderV2.HEADER_API_KEY to context.clientToken
-            ),
-            ByteArray(0)
+            "",
+            emptyMap(),
+            body = ByteArray(0)
         )
     }
+
+    private val intakeUrl by lazy {
```
0xnm, line 66: minor: I would move property up, before any functions.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
```diff
@@ -6,41 +6,178 @@
...
+            "${RumAttributes.SERVICE_NAME}:${datadogContext.service}",
+            "${RumAttributes.APPLICATION_VERSION}:" +
+                datadogContext.version,
+            "${RumAttributes.SDK_VERSION}:${datadogContext.sdkVersion}",
+            "${RumAttributes.ENV}:${datadogContext.env}"
+        )
+        if (datadogContext.variant.isNotEmpty()) {
+            elements.add("${RumAttributes.VARIANT}:${datadogContext.variant}")
+        }
+        return elements.joinToString(",")
```
0xnm, line 86: are these tags are still applicable for the SR? `RumAttributes` will be in the RUM module eventually, but we can make them a part of public API.

0xnm (changes requested):

detekt.yml
```diff
@@ -726,6 +726,7 @@ datadog:
       - "okhttp3.OkHttpClient.Builder.eventListenerFactory(okhttp3.EventListener.Factory):java.lang.NullPointerException"
       - "okio.BufferedSink.close():java.io.IOException"
       - "okio.Okio.buffer(okio.Sink):java.lang.NullPointerException"
+      - "okio.Buffer.readByteArray():java.lang.AssertionError"
```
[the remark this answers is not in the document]
mariusc83: Good catch 👍

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
mariusc83: I think this will be nice if we can do so.

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactoryTest.kt
```diff
@@ -0,0 +1,280 @@
...
+    lateinit var fakeMediaType: MediaType
+
+    var fakeBatchMetadata: ByteArray? = null
+
+    @Mock
+    lateinit var mockInternalLogger: Logger
+
+    @BeforeEach
+    fun `set up`(forge: Forge) {
+        fakeMediaType = MediaType.get(SessionReplayRequestFactory.CONTENT_TYPE_MULTIPART_FORM)
```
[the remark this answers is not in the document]
mariusc83: I think here I need to rethink a bit this test maybe I will need more values there.

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriterTest.kt
```diff
@@ -0,0 +1,163 @@
...
+        val fakeRecord1 = forge.forgeEnrichedRecord()
+        val fakeMetaRecord = MobileSegment.MobileRecord.MetaRecord(
+            timestamp = System.currentTimeMillis(),
+            data = MobileSegment.Data1(
+                width = forge.aPositiveLong(),
+                height = forge.aPositiveLong()
+            )
+        )
+        val fakeRecord2 = fakeRecord1.copy(records = listOf(fakeMetaRecord))
+        whenever(mockSessionReplayFeature.withWriteContext(eq(true), any())) doAnswer {
```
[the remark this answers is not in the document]
mariusc83: hmmm I like this approach this way I can get rid of this `fakeMetaRecord` workaround

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriterTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Test
+    fun `M write the record in a new batch W write { first view in session }`(forge: Forge) {
+        // Given
+        val fakeRecord = forge.forgeEnrichedRecord()
+        whenever(mockSessionReplayFeature.withWriteContext(eq(true), any())) doAnswer {
+            val callback = it.getArgument<(DatadogContext, EventBatchWriter) -> Unit>(1)
+            callback.invoke(fakeDatadogContext, mockEventBatchWriter)
+        }
```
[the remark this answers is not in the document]
mariusc83: I'd rather keep this in the test for better visibility

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
```diff
@@ -6,41 +6,145 @@
...
+        }
+    }
+
+    // endregion
+
+    companion object {
+        internal const val EXTRACT_BYTE_ARRAY_FROM_BODY_ERROR_MESSAGE =
+            "Unable to extract the ByteArray from session replay request body."
+        internal const val QUERY_PARAM_SOURCE = "ddsource"
+        internal const val QUERY_PARAM_TAGS = "ddtags"
```
0xnm, line 147: minor: those exist in `RequestFactory` as well if needed.

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriterTest.kt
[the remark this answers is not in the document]
0xnm: Actually maybe this test can be simplified: we don't care about the write, we just want to check that `withWriteContext` is never called with `forceNewBatch = true`. Probably we don't care what happens inside the lambda.

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriterTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    @Test
+    fun `M do nothing W write { feature not properly initialized }`(forge: Forge) {
+        // Given
+        val fakeRecord = forge.forgeEnrichedRecord()
+        whenever(mockSdkCore.getFeature(SessionReplayFeature.SESSION_REPLAY_FEATURE_NAME))
+            .thenReturn(null)
+        whenever(mockSessionReplayFeature.withWriteContext(any(), any())) doAnswer {
+            val callback = it.getArgument<(DatadogContext, EventBatchWriter) -> Unit>(1)
+            callback.invoke(fakeDatadogContext, mockEventBatchWriter)
+        }
```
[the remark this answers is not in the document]
0xnm: also can be simplified I think, we can simply assert that `mockSessionReplayFeature.withWriteContext` was never called.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
0xnm: Yes, empty request looks like a hack and we shouldn't even make an attempt to send it.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/net/BatchesToSegmentsMapper.kt
```diff
@@ -42,74 +48,96 @@ class BatchesToSegmentsMapper {
...
 
-        @Suppress("UnsafeThirdPartyFunctionCall")
-        val stopTimestamp = records.last().getAsJsonPrimitive(TIMESTAMP_KEY).asLong
-        val hasFullSnapshotRecord = hasFullSnapshotRecord(records)
+        val startTimestamp = orderedRecords.firstOrNull()?.asJsonObject?.timestamp()
+        val stopTimestamp = orderedRecords.lastOrNull()?.asJsonObject?.timestamp()
+
+        if (startTimestamp == null || stopTimestamp == null) {
+            // this is just to avoid having kotlin warnings but the elements
+            // without timestamp property were already removed int the logic above
```
0xnm, line 89:
```suggestion
            // without timestamp property were already removed in the logic above
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/net/BatchesToSegmentsMapper.kt
```diff
@@ -42,74 +48,96 @@ class BatchesToSegmentsMapper {
...
+                val records = accumulator.second
+                val newRecords = pair.second
+                records.addAll(newRecords)
+                Pair(accumulator.first, records)
+            } ?: return null
+
+        return mapToSegment(reducedEnrichedRecord.first, reducedEnrichedRecord.second)
+    }
+
+    private fun mapToSegment(rumContext: SessionReplayRumContext, records: JsonArray):
```
0xnm, line 61: shouldn't operations with json in this method be in the `try-catch` as well? like in `groupBatchDataIntoSegments` method. Or maybe we can extract everything in the typed object in one place and then continue working with it.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
mariusc83: oh I didn't know that, good :)

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriterTest.kt
```diff
@@ -0,0 +1,159 @@
...
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.Mock
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+import java.util.*
```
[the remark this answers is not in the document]
mariusc83: oh yes...not sure why it didn't :(

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/net/BatchesToSegmentsMapper.kt
mariusc83: Good catch, not sure extracting it will work as I still need to modify the serialized segment by adding the ordered records.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
mariusc83: Do we add them by default there? Then I don't see the point for me to add them also here.

0xnm: no, there is only declaration there, they are not added automatically to any request.

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriterTest.kt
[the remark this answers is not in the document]
mariusc83: it is AS that does me the trick every time when I use auto imports, need to check where i can disable that.

detekt.yml
```diff
@@ -1177,9 +1179,15 @@ datadog:
       - "kotlin.collections.MutableSet.map(kotlin.Function1)"
       - "kotlin.collections.MutableSet.remove(java.io.File)"
       - "kotlin.collections.MutableSet.remove(com.datadog.android.v2.core.internal.storage.ConsentAwareStorage.Batch)"
+      - "kotlin.collections.Iterable.toMap(kotlin.collections.MutableMap)"
```
0xnm, line 1182: minor: should go up, to L1058 (where `Iterable` is declared)

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/net/BatchesToSegmentsMapper.kt
```diff
@@ -42,74 +43,108 @@ class BatchesToSegmentsMapper {
...
+    }
+
+    private fun JsonObject.timestamp(): Long? {
+        return getAsJsonPrimitive(TIMESTAMP_KEY)?.asLong
+    }
+
+    private fun JsonObject.rumContext(): SessionReplayRumContext? {
+        val applicationId = get(EnrichedRecord.APPLICATION_ID_KEY)?.asString
+        val sessionId = get(EnrichedRecord.SESSION_ID_KEY)?.asString
+        val viewId = get(EnrichedRecord.VIEW_ID_KEY)?.asString
```
0xnm, line 136: these methods can still throw json related exceptions and they are not covered by try-catch

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/gson/GsonExtTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
```
0xnm, line 25: minor: we don't really need this one (and `MockitoSettings` annotation as well), because we don't mock anything in this test.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/utils/GsonJsonObjectForgeryFactory.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.google.gson.JsonPrimitive
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+class GsonJsonObjectForgeryFactory : ForgeryFactory<JsonObject> {
+
+    override fun getForgery(forge: Forge): JsonObject {
+        return JsonObject().apply {
+            val fieldCount = forge.aTinyInt()
+            for (i in 0..fieldCount) {
```
0xnm, line 20:
```suggestion
            repeat(fieldCount + 1) {
```

0xnm (approved):
LGTM. I've added few suggestions, but they are not blocking for me.

The only concern is "empty request" thing, but let's create another task to do something with it eventually.

detekt.yml
mariusc83: Actually that's a duplicate from a rebase :)

PR #1177, opened 2022-12-07
RUMM-2789: Use message bus to report NDK crashes to RUM

0xnm:
### What does this PR do?

This PR changes the way NDK crashes are reported to RUM: now they will be sent through the message bus, because RUM feature may be optional and will reside in another module anyway.

Also there are now 2 different message types for the crashes coming to Logs and RUM: `jvm_crash` and `ndk_crash`. This way the code is simpler and is easier to test.

`DatadogNdkCrashHandler` now reads last view event as JSON (because RUM-specific deserializers won't be available), takes view ID, session ID and application ID from it and passes it to the RUM feature where further deserialization from `JsonObject` to the RUM-specific type is done.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/JsonObjectDeserializer.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/JsonObjectDeserializer.kt
@@ -4,31 +4,29 @@
  * Copyright 2016-Present Datadog, Inc.
  */
 
-package com.datadog.android.rum.internal.ndk
+package com.datadog.android.core.internal.persistence
 
-import com.datadog.android.core.internal.persistence.Deserializer
-import com.datadog.android.log.Logger
+import com.datadog.android.core.internal.utils.sdkLogger
 import com.datadog.android.log.internal.utils.errorWithTelemetry
+import com.google.gson.JsonObject
 import com.google.gson.JsonParseException
+import com.google.gson.JsonParser
 import java.util.Locale
 
-internal class NdkCrashLogDeserializer(
-    private val internalLogger: Logger
-) : Deserializer<String, NdkCrashLog> {
-
-    override fun deserialize(model: String): NdkCrashLog? {
+internal class JsonObjectDeserializer : Deserializer<String, JsonObject> {
+    override fun deserialize(model: String): JsonObject? {
         return try {
-            NdkCrashLog.fromJson(model)
-        } catch (e: JsonParseException) {
-            internalLogger.errorWithTelemetry(
+            JsonParser.parseString(model).asJsonObject
+        } catch (jpe: JsonParseException) {
+            sdkLogger.errorWithTelemetry(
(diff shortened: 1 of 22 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/JsonObjectDeserializer.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.google.gson.JsonParseException
+import com.google.gson.JsonParser
+import java.util.Locale
+
+internal class JsonObjectDeserializer : Deserializer<String, JsonObject> {
+    override fun deserialize(model: String): JsonObject? {
+        return try {
+            JsonParser.parseString(model).asJsonObject
+        } catch (jpe: JsonParseException) {
+            sdkLogger.errorWithTelemetry(
```
mariusc83, line 21: I am afraid that if we have a problem with one of our formats in one big customer we will overload the telemetry here. I know you have your back pressure mechanism in the telemetry handler just wanted to make sure you considered this.

dd-sdk-android/src/test/kotlin/com/datadog/android/log/internal/LogsFeatureTest.kt
```diff
@@ -419,133 +421,95 @@ internal class LogsFeatureTest {
...
             "throwable" to fakeThrowable
         )
 
         // When
         testedFeature.onReceive(event)
 
         // Then
-        argumentCaptor<LogEvent> {
-            verify(mockDataWriter).write(eq(mockEventBatchWriter), capture())
+        verify(mockLogsFeatureScope).withWriteContext(any(), any())
```
mariusc83, line 455: I don't see where are you verifying that you are writing the log here.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/JsonObjectDeserializer.kt
0xnm: It is not different from what we have in any `Deserializer` right now, all of them also submit such telemetry. So far no issues spotted :)

dd-sdk-android/src/test/kotlin/com/datadog/android/log/internal/LogsFeatureTest.kt
0xnm: Nothing will be written because it is the test for the timeout - we check that we can complete the test (caller thread is not blocked forever) and by the time we complete nothing is written yet (`verifyZeroInteractions(mockDataWriter)` checks that), because write operation is still hanging.

mariusc83: I think in this case maybe you want to update the test function name ?

mariusc83 (approved):

xgouchet (approved):

PR #1198, opened 2022-12-15
RUMM-2806 Drop the covered wireframes only if top ones have a background

mariusc83:
### What does this PR do?

Before we were filtering out all the overlayed wireframes no matter if the top ones were having a solid background or not. This creates a problem because if the removed wireframe is actual the one that gives the main background in a View we will be loosing the contrast in the replay and will make some elements invisible (e.g. texts).

In order to fix this we are now only considering as valid overlays the wireframes that also have a solid backround.

### Motivation

### Additional Notes

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
@@ -51,7 +51,7 @@ internal class WireframeUtils {
             return false
         }
         topWireframes.forEach {
-            if (it.bounds().isCovering(wireframeBounds)) {
+            if (it.bounds().isCovering(wireframeBounds) && it.isOpaque()) {
                 return false
             }
         }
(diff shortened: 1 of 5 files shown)


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
```diff
@@ -72,6 +72,13 @@ internal class WireframeUtils {
         }
     }
 
+    private fun MobileSegment.Wireframe.hasBackground(): Boolean {
```
xgouchet, line 75: what if the background is transparent?

mariusc83: Good catch, I am going to filter those out also.

ncreated: 👍 or semi-transparent. On iOS we recently added [`isTranslucent`](https://github.com/DataDog/dd-sdk-ios/blob/15b5accc7b9bfa65228b5c69205a055dd46c2ec7/session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeSnapshot.swift#L96-L99) for similar use case, which also takes view's `alpha` (opacity) into account. It doesn't yet check the background color's alpha, but remains extensible in that direction (e.g. a background of `#FF000099` should be considered translucent).

mariusc83: yup I modified it in order to just consider alpha `FF` in the color and `1f` in the `ShapeStyle` opacity. Any other wireframe with any alpha value will be not considered as overlay.

0xnm (approved):

xgouchet (approved):

PR #1201, opened 2022-12-20
MTG-8 use latest version of shared pipeline

xgouchet:
### What does this PR do?

Bump the pipeline dependency (cf [dd-source#21171](https://github.com/DataDog/dd-source/pull/21171) )

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -73,10 +73,10 @@ analysis:licenses:
 
 static-analysis:
   stage: analysis
-  variables: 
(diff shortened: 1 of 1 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/sampling/Sampler.kt
```diff
@@ -8,6 +8,7 @@ package com.datadog.android.core.internal.sampling
 
 internal interface Sampler {
 
+
```
0xnm, line 11: seems this one should be reverted

0xnm (approved):

PR #1212, opened 2023-01-03
RUMM-2796 RequestFactory can throw exceptions

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/RequestFactory.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/RequestFactory.kt
@@ -19,7 +19,9 @@ fun interface RequestFactory {
      * @param context Datadog SDK context.
      * @param batchData Raw data of the batch.
      * @param batchMetadata Raw metadata of the batch.
+     * @throws [Exception] in case the request could not be created.
      */
+    @kotlin.jvm.Throws
     fun create(
         context: DatadogContext,
         batchData: List<ByteArray>,
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
@@ -10,6 +10,7 @@ import com.datadog.android.core.internal.net.UploadStatus
 import com.datadog.android.core.internal.system.AndroidInfoProvider
 import com.datadog.android.core.internal.utils.devLogger
 import com.datadog.android.log.Logger
+import com.datadog.android.log.internal.utils.errorWithTelemetry
 import com.datadog.android.v2.api.RequestFactory
 import com.datadog.android.v2.api.context.DatadogContext
 import okhttp3.Call
@@ -35,7 +36,12 @@ internal class DataOkHttpUploader(
         batch: List<ByteArray>,
         batchMeta: ByteArray?
     ): UploadStatus {
-        val request = requestFactory.create(context, batch, batchMeta)
+        val request = try {
+            requestFactory.create(context, batch, batchMeta)
+        } catch (e: Exception) {
+            internalLogger.errorWithTelemetry("Unable to create the upload request.", e)
+            return UploadStatus.UNKNOWN_ERROR
+        }
 
         val uploadStatus = try {
             executeUploadRequest(request)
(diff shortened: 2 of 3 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
```diff
@@ -35,7 +36,12 @@ internal class DataOkHttpUploader(
         batch: List<ByteArray>,
         batchMeta: ByteArray?
     ): UploadStatus {
-        val request = requestFactory.create(context, batch, batchMeta)
+        val request = try {
+            requestFactory.create(context, batch, batchMeta)
+        } catch (e: Exception) {
+            internalLogger.errorWithTelemetry("Unable to create the upload request.", e)
+            return UploadStatus.UNKNOWN_ERROR
```
mariusc83, line 43: should we create a dedicated error for this case ?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/RequestFactory.kt
```diff
@@ -19,7 +19,9 @@ fun interface RequestFactory {
      * @param context Datadog SDK context.
      * @param batchData Raw data of the batch.
      * @param batchMetadata Raw metadata of the batch.
+     * @throws [Exception] in case the request could not be created.
      */
+    @kotlin.jvm.Throws
```
0xnm, line 24: I think this one is not needed. According to the [description](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/) this is a compiler-related annotation which says to add `throws` keyword to the method signature in Java. This keyword will force the caller to handle this exception.

While in our case caller is not in Java, but in Kotlin, this obligation has no effect. Also not sure if `Throws` annotation without the list of exception will generate anything at all for JVM, will it?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
```diff
@@ -35,7 +36,12 @@ internal class DataOkHttpUploader(
         batch: List<ByteArray>,
         batchMeta: ByteArray?
     ): UploadStatus {
-        val request = requestFactory.create(context, batch, batchMeta)
+        val request = try {
+            requestFactory.create(context, batch, batchMeta)
+        } catch (e: Exception) {
+            internalLogger.errorWithTelemetry("Unable to create the upload request.", e)
```
0xnm, line 42: you need to rebase, there is no such method anymore :)

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
0xnm: Yes, I think we should. Also we should show the exception happened to the user as well, so let's log it with the target=user, not contributor.

0xnm (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
```diff
@@ -35,7 +36,12 @@ internal class DataOkHttpUploader(
         batch: List<ByteArray>,
         batchMeta: ByteArray?
     ): UploadStatus {
-        val request = requestFactory.create(context, batch, batchMeta)
+        val request = try {
+            requestFactory.create(context, batch, batchMeta)
```
xgouchet, line 40: Can we be more explicit in which exception can happen here? It is a code smell to catch any Exception, so we need to have a better understanding of what happens in the request factory.
Also shouldn't this be handled in the request factory directly instead?

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/api/RequestFactory.kt
0xnm: Probably mentioning it in the documentation will be enough. Not sure if `Throws` is copied to the implementation of the interface automatically by the IDE (so need to check interface description manually anyway in this case).

xgouchet (changes requested):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
mariusc83: the idea is that the exception can be `Any` as this can be thrown by an implementation of the `RequestFactory`. You can't know what this will be. We are actually following `iOS` here as they have the same approach on their end. 
You can't handle this in the `RequestFactory` as you need to tell the core what to do with the batch in case of an exception when creating the `Request`

PR #1215, opened 2023-01-04
RUMM-2516: Remove cross platform duplicates crashes

louiszawadzki:
### What does this PR do?

- Makes sure a view update is sent when a cross-platform crash is detected
- Do not send the native crash if there already was a crash

### Motivation

Remove duplicated errors for JS crashes in React Native.
I'm not sure why we had 2 different behaviours in the first place depending on whether the crash was cross-platform or native: https://github.com/DataDog/dd-sdk-android/pull/764

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -359,7 +359,7 @@ internal open class RumViewScope(
         val rumContext = getRumContext()
 
         val updatedAttributes = addExtraAttributes(event.attributes)
-        val isFatal = updatedAttributes.remove(RumAttributes.INTERNAL_ERROR_IS_CRASH) as? Boolean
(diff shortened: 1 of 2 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -359,7 +359,9 @@ internal open class RumViewScope(
         val rumContext = getRumContext()
 
         val updatedAttributes = addExtraAttributes(event.attributes)
-        val isFatal = updatedAttributes.remove(RumAttributes.INTERNAL_ERROR_IS_CRASH) as? Boolean
+        val isFatal = updatedAttributes.remove(RumAttributes.INTERNAL_ERROR_IS_CRASH) as? Boolean == true || event.isFatal
+        if (crashCount > 0 && isFatal) return // if a cross-platform crash was already reported, do not send its native version
```
mariusc83, line 363: This could work only if we are sure there's not going to be another fatal crash reported after this.

xgouchet: If the crash is fatal, none could be reported after.

xgouchet (approved):

PR #1216, opened 2023-01-04
RUMM-2834: Tracing feature stores context in the common context storage

0xnm:
### What does this PR do?

This change makes tracing feature to store its context (active span) in the common context storage.

The tricky part is that active span (span backed by the current active scope) is thread-local, so each thread may have its own active span. To support such thread-local storage I'm using `addScopeListener` API to associate the active context with the particular thread and then later to read it from `DatadogContext` for the particular thread as well (reminder: `DatadogContext` is created on the caller thread). 


mariusc83 (approved):
Nice work !!

xgouchet (approved):

PR #1217, opened 2023-01-05
RUMM-2797 SR RequestFactory allow to throw exceptions

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
@@ -6,10 +6,9 @@
 
 package com.datadog.android.sessionreplay.internal.domain
 
-import com.datadog.android.core.internal.utils.internalLogger
 import com.datadog.android.rum.RumAttributes
+import com.datadog.android.sessionreplay.internal.exception.InvalidPayloadFormatException
 import com.datadog.android.sessionreplay.net.BatchesToSegmentsMapper
-import com.datadog.android.v2.api.InternalLogger
 import com.datadog.android.v2.api.Request
 import com.datadog.android.v2.api.RequestFactory
 import com.datadog.android.v2.api.context.DatadogContext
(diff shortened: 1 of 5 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
```diff
@@ -109,45 +89,33 @@ internal class SessionReplayRequestFactory(
...
-                e
-            )
-            null
-        }
+    private fun extractByteArrayFromBody(body: RequestBody): ByteArray {
+        val buffer = Buffer()
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        body.writeTo(buffer)
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        return buffer.readByteArray()
```
0xnm, line 113: this `readByteArray` method is strange - it says that it can throw `IOException`, but the actual implementation actually can also throw `Error`:

```
    try {
      return readByteArray(size);
    } catch (EOFException e) {
      throw new AssertionError(e);
    }
```

Do you think we should catch it? Because the outer code only handles exceptions, but not errors. Normally errors are not recoverable though, so maybe we don't need to.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/exception/InvalidPayloadFormatException.kt
```diff
@@ -0,0 +1,9 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.exception
+
+internal class InvalidPayloadFormatException(message: String) : Exception(message)
```
0xnm, line 9: Let's make it extend `RuntimeException`, so that it is not checked one.

```suggestion
internal class InvalidPayloadFormatException(message: String) : RuntimeException(message)
```

0xnm (approved):
I left some comments, but overall ltgm.

dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/SessionReplayRequestFactory.kt
mariusc83: yeah...that's the thing...I would leave errors not caught ...it will be another thread anyway not the main thread.

PR #1221, opened 2023-01-09
RUMM-2845: Speed up build of sample app for CodeQL scan

0xnm:
### What does this PR do?

Gradle build for the CodeQL scan take 35-40 minutes it seems, because all the variants ((`eu1`, `us1`, `us3`, `us5`, `us1_fed`, `staging`) x (`debug`,`release`) = 12) are built (it takes majority of the build).

~~To speed it up and still have a complete coverage of all the binaries which can be created in the customer environment (not only release artifacts, but also binaries which can be created by cloning this repo), let's build only `us1` flavor for sample app, because all DC flavors are the same code-wise.~~

To speed it up we are going to build only releasable modules.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -47,7 +47,7 @@ jobs:
 
     # Manually build the java bytecode
     - name: Execute Gradle build
-      run: ./gradlew assemble
+      run: ./gradlew assembleAll
 
     # Perform the analysis
     - name: Perform CodeQL Analysis


.github/workflows/codeql-analysis.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
     # Manually build the java bytecode
     - name: Execute Gradle build
-      run: ./gradlew assemble
+      run: ./gradlew assemble -Pcodeql-build
```
xgouchet, line 50: Instead of making this conditonal in the sample project, we could limit this script to run the assemble tasks on the library content only. The codeql check is only mandatory for code distributed to our customers' apps

0xnm: I was thinking we can also cover the binaries which can be built in the customer environment by cloning this repository, not only releasable artifacts. With the current change it seems overall CodeQL analysis takes the same time as other stages (the slowest is Bitrise), so at least we are not making overall checks timeline slower.

But yes, we can push it more by ignoring non-releasable content. WDYT?

xgouchet: Covering the samples and other non releasable contents don't really make sense, as no customer would publish our sample app on the store.

0xnm: Yes, I was more thinking about running "bad" binary (even for the sample app) in the customer internal environment, which still have a risk to be exploited.

I will update this PR to build only releasable artifacts.

xgouchet (approved):

PR #1223, opened 2023-01-09
SR use case insensitive when checking wireframe background color opacity

mariusc83:
### What does this PR do?

We were using assuming that the Wireframe `backgroundColor` will always be reported in capital letters when checking for its opacity. This was buggy and in this PR we are fixing this by using a case insensitive `RegEx` instead.

### Motivation

### Additional Notes

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
@@ -7,6 +7,7 @@
 package com.datadog.android.sessionreplay.processor
 
 import com.datadog.android.sessionreplay.model.MobileSegment
+import java.util.Locale
 import kotlin.math.max
 
 internal class WireframeUtils {
@@ -86,7 +87,7 @@ internal class WireframeUtils {
     private fun MobileSegment.ShapeStyle.isOpaque(): Boolean {
         return this.opacity == FULL_OPACITY_ALPHA &&
             this.backgroundColor != null &&
-            this.backgroundColor.takeLast(2) == FULL_OPACITY_ALPHA_AS_HEXA_STRING
+            this.backgroundColor.takeLast(2).lowercase(Locale.US) == FULL_OPACITY_STRING_HEXA
     }
 
     private fun MobileSegment.Wireframe.ShapeWireframe.bounds(): Bounds {
@@ -121,7 +122,7 @@ internal class WireframeUtils {
     )
 
     companion object {
-        const val FULL_OPACITY_ALPHA_AS_HEXA_STRING = "FF"
-        const val FULL_OPACITY_ALPHA = 1f
+        private const val FULL_OPACITY_STRING_HEXA = "ff"
+        private const val FULL_OPACITY_ALPHA = 1f
     }
 }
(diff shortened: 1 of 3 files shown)


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
```diff
@@ -86,7 +86,7 @@ internal class WireframeUtils {
     private fun MobileSegment.ShapeStyle.isOpaque(): Boolean {
         return this.opacity == FULL_OPACITY_ALPHA &&
             this.backgroundColor != null &&
-            this.backgroundColor.takeLast(2) == FULL_OPACITY_ALPHA_AS_HEXA_STRING
+            this.backgroundColor.takeLast(2).matches(FULL_OPACITY_REG_EX)
```
0xnm, line 89: probably we don't need regex here, we can just convert to upper/lowecase and compare?

mariusc83: yes was not sure which one is better but I think conversion is faster. I will do that.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
```diff
@@ -86,7 +86,7 @@ internal class WireframeUtils {
     private fun MobileSegment.ShapeStyle.isOpaque(): Boolean {
         return this.opacity == FULL_OPACITY_ALPHA &&
             this.backgroundColor != null &&
-            this.backgroundColor.takeLast(2) == FULL_OPACITY_ALPHA_AS_HEXA_STRING
+            this.backgroundColor.takeLast(2).lowercase() == FULL_OPACITY_STRING_HEXA
```
0xnm, line 89:
```suggestion
            this.backgroundColor.takeLast(2).lowercase(Locale.US) == FULL_OPACITY_STRING_HEXA
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
```diff
@@ -121,7 +121,7 @@ internal class WireframeUtils {
     )
 
     companion object {
-        const val FULL_OPACITY_ALPHA_AS_HEXA_STRING = "FF"
+        val FULL_OPACITY_STRING_HEXA = "ff"
```
0xnm, line 124:
```suggestion
        const val FULL_OPACITY_STRING_HEXA = "ff"
```

0xnm (approved):
lgtm! I added some suggestions.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/WireframeUtils.kt
mariusc83: Isn't it by default ?

mariusc83: Oh the default is the `ROOT` meaning the project locale....damn

PR #1234, opened 2023-01-17
RUMM-2898 Add hasReplay=true only when the view produced SR records

mariusc83:
### What does this PR do?

We were adding the has_replay:true property on the RUM events whenever the SR is recording no matter if the view in progress produced any records or not. Following the latest discussions we needed to fix this and set this property to true only if there was at least one SR record sent for this event.

### Motivation

### Additional Notes

There is still a corner case for which this flag can have wrong values due to the async way of intercommunication between our features but this was explained in the PR comments and for now we are accepting this as a non blocker issue.
 
### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
@@ -158,6 +158,5 @@ internal class SessionReplayFeature(
         const val SESSION_REPLAY_BUS_MESSAGE_TYPE_KEY = "type"
         const val RUM_SESSION_RENEWED_BUS_MESSAGE = "rum_session_renewed"
         const val RUM_KEEP_SESSION_BUS_MESSAGE_KEY = "keepSession"
-        const val IS_RECORDING_CONTEXT_KEY = "is_recording"
     }
 }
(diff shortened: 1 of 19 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -159,5 +159,6 @@ internal class SessionReplayFeature(
         const val RUM_SESSION_RENEWED_BUS_MESSAGE = "rum_session_renewed"
         const val RUM_KEEP_SESSION_BUS_MESSAGE_KEY = "keepSession"
         const val IS_RECORDING_CONTEXT_KEY = "is_recording"
+        const val HAS_REPLAY_VIEWS_MAP = "has_replay_views_map"
```
0xnm, line 162: seems like this one is not used

dd-sdk-android/src/test/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayRecordCallbackTest.kt
```diff
@@ -36,30 +44,12 @@ internal class SessionReplayRecordCallbackTest {
...
-        // Then
-        argumentCaptor<(MutableMap<String, Any?>) -> Unit> {
-            verify(mockDatadogCore).updateFeatureContext(
-                eq(SessionReplayFeature.SESSION_REPLAY_FEATURE_NAME),
-                capture()
-            )
-
-            val featureContext = mutableMapOf<String, Any?>()
-            lastValue.invoke(featureContext)
+    fun `M update session replay context W onRecordForViewSent`(@Forgery forge: Forge) {
```
0xnm, line 47: `Forgery` annotation is not needed here

```suggestion
    fun `M update session replay context W onRecordForViewSent`(forge: Forge) {
```

0xnm (approved):
lgtm! I've added some minor comments not related to the functional part

xgouchet (approved):

PR #1239, opened 2023-01-18
RUMM-2931 prevent reporting ANR when app is in background

### LINKED ISSUE #1067: ANR reports number is significantly different from the Google Play
**Describe what happened**
The number of ANR reports in RUM is significantly more than at Google Play. However, the number of crashes is similar.

1. Integrate DataDog Android SDK
2. Check the RUM error section

1. Number of ANR is similar to Google Play's one

P.S. I was not able to find the algorithm of the ANR collection in the documentation.

xgouchet:
### What does this PR do?

Fixes #1067 

This PR prevents reporting false positive ANRs when the application is in background. 

### Motivation

In some cases (#1067) the host application is in background and the SDK detekts multiple SDKs that are not actually perceived by the end user.

### DIFF
--- a/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
+++ b/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
@@ -12,7 +12,10 @@ import android.util.Log
 import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.net.FirstPartyHostDetector
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import com.datadog.android.rum.RumErrorSource
 import com.datadog.android.rum.internal.RumFeature
+import com.datadog.android.rum.internal.anr.ANRDetectorRunnable
+import com.datadog.android.rum.internal.anr.ANRException
 import com.datadog.android.rum.internal.domain.RumContext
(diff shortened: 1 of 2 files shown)


0xnm (approved):

mariusc83 (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -355,6 +358,26 @@ internal class RumViewManagerScopeTest {
         assertThat(testedScope.childrenScopes).hasSize(0)
     }
 
+    @Test
+    fun `𝕄 not start a bg ViewScope 𝕎 handleEvent { app displayed, event is background ANR}`() {
+        // Given
+        testedScope.applicationDisplayed = true
+        val fakeEvent = RumRawEvent.AddError(
+            message = ANRDetectorRunnable.ANR_MESSAGE,
```
emartynov, line 366: Probably off topic - from names `ANRDetectorRunnable.ANR_MESSAGE` and `ANRException` I can not drive understanding it is a background ANR.

emartynov: Also, testedScope has `applicationDisplayed` as `true`, which sounds to me as if it is not in the background.

PR #1259, opened 2023-01-31
RUMM-2994: Register Log feature dynamically

0xnm:
### What does this PR do?

This change removes Logs feature initialization from the SDK initialization step in favor of registering it dynamically, using `SdkCore#registerFeature`.

Also `Configuration.Feature.Logs` class is removed.


xgouchet (approved):

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
```diff
@@ -117,13 +118,22 @@ class SampleApplication : Application() {
...
         val sessionReplayConfig = SessionReplayConfiguration.Builder().apply {
+            useSite(DatadogSite.valueOf(BuildConfig.DD_SITE_NAME))
             if (BuildConfig.DD_OVERRIDE_SESSION_REPLAY_URL.isNotBlank()) {
                 useCustomEndpoint(BuildConfig.DD_OVERRIDE_SESSION_REPLAY_URL)
             }
         }.build()
         val sessionReplayFeature = SessionReplayFeature(sessionReplayConfig)
         Datadog.registerFeature(sessionReplayFeature)
 
+        val logsFeature = LogsFeature.Builder().apply {
```
mariusc83, line 129: why do we have 2 different way of building a feature ? I see that SR feature is built by passing a config to its constructor whereas for Logs you have a builder instead. Can't we just make them the same ?

mariusc83 (changes requested):

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
0xnm: Yes, that is the plan - make SR follow the same pattern later.

mariusc83 (approved):

PR #1261, opened 2023-02-02
RUMM-2611: Extract Logs feature into dedicated module

0xnm:
### What does this PR do?

This change moves all (well, almost all) Logs feature code into the dedicated module. What is done:

* Introduced `FeatureScope#unwrap` method, so that we are able to get the original feature registered with the particular instance, this allows to access feature properties.
* Some classes which were using logs and which are deprecated long time ago (`RumWebChromeClient` and `RumWebViewClient`) are removed.
* Some internal classes/methods are made public and moved out of `internal` packages. These classes/methods are `DataConstraints`, `DatadogDataConstraints`, `Serializer`, `Sampler`, `RateBasedSampler`, `serializeToByteArray`, `Any?.toJsonElement`, `NULL_MAP_VALUE` (dedicated task created to replace this one), `MapperSerializer`, `NoOpEventMapper`. Task created to arrange packages after this move, because we can have `com.datadog.android.core.persistence` and `com.datadog.android.core.internal.persistence` instead of `com.datadog.android.core.persistence.internal` which makes more sense.
* `LogAttributes` class is kept in the core module, because it is used by other feature. `LogsRequestFactory` is also kept in the main module, because it is used by WebView Logs feature (maybe we should migrate it to the message bus instead).
* `Logger` is now associated with the particular SDK instead by making `Logger.Builder` request `SdkCore`.
* `service` property added to the `SdkCore`, because it is needed for the `Logger` creation.
* `Datadog.globalSdkCore` is made public, but this is temporary, until we implement the global registry.

This change is quite big (I tried to deliver it in the working state), so it is split in several commit to simplify the review.


mariusc83 (approved):
LGTM

xgouchet (approved):

PR #1262, opened 2023-02-02
RUMM-2943: Convert NDK plugin into a feature

0xnm:
### What does this PR do?

This change removes usages of `DatadogPlugin` and converts `NdkCrashPlugin` into the feature.

To support this new class was introduced `EnvironmentProvider` which is giving the information about tracking consent and root storage directory (previously provided by `DatadogPlugin`). Maybe we can come with smarter solution later.

Also NDK code is moved under the `libraries` folder (which later will be renamed into `features`, to better reflect the content).


library/dd-sdk-android-ndk/src/main/kotlin/com/datadog/android/ndk/NdkCrashReportsFeature.kt
```diff
@@ -4,73 +4,66 @@
...
+                InternalLogger.Target.USER,
+                "Unable to create NDK Crash Report folder $ndkCrashesDirs",
+                e
+            )
             return
         }
         registerSignalHandler(
             ndkCrashesDirs.absolutePath,
-            consentToInt(config.trackingConsent)
+            consentToInt(environmentProvider.trackingConsent)
```
mariusc83, line 56: I don't remember how we were doing this before but have you considered the case when the tracking consent is changed ?

0xnm: yes, `onConsentUpdated` callback is still there https://github.com/DataDog/dd-sdk-android/pull/1262/files#diff-f688390d6a60752a35fe77982ed01029b262b098deaf5bc55b162cc68ecabd7dR71

mariusc83 (approved):

PR #1265, opened 2023-02-03
RUMM-2885 Add `addFeatureFlagEvaluation` function for RUM

fuzzybinary:
### What does this PR do?

`addFeatureFlagEvalutaion` adds the result of evaluating a feature flag to a view, which is key / value pair. These feature flags are sent with any subsequent View updates, as well as on Errors until the view is stopped.

### DIFF
--- a/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
+++ b/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
@@ -6246,6 +6246,108 @@ internal class RumViewScopeTest {
 
     // endregion
 
+    // region Feature Flags
+
+    @Test
+    fun `M send event W handleEvent(AddFeatureFlagEvaluation) on active view`(
+        @StringForgery flagName: String,
+        @StringForgery flagValue: String
+    ) {
+        // GIVEN
+
+        // WHEN
+        testedScope.handleEvent(
+            RumRawEvent.AddFeatureFlagEvaluation(
+                name = flagName,
+                value = flagValue
+            ),
+            mockWriter
+        )
+
+        // THEN
+        argumentCaptor<ViewEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture())
+            assertThat(lastValue)
+                .apply {
+                    hasFeatureFlag(flagName, flagValue)
+                }
+        }
+    }
+
+    @Test
+    fun `M modify flag W handleEvent(AddFeatureFlagEvaluation) on active view { existing feature flag }`(
+        @StringForgery flagName: String,
+        @BoolForgery oldFlagValue: Boolean,
+        @StringForgery flagValue: String
+    ) {
+        // GIVEN
+        testedScope.handleEvent(
+            RumRawEvent.AddFeatureFlagEvaluation(
+                name = flagName,
+                value = oldFlagValue
+            ),
+            mockWriter
+        )
+
+        // WHEN
+        testedScope.handleEvent(
+            RumRawEvent.AddFeatureFlagEvaluation(
+                name = flagName,
+                value = flagValue
+            ),
+            mockWriter
+        )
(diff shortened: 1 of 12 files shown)


xgouchet (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -6246,6 +6246,108 @@ internal class RumViewScopeTest {
...
     // endregion
 
+    // region Feature Flags
+
+    @Test
+    fun `M send event W handleEvent(AddFeatureFlagEvaluation) on active view`(
+        @StringForgery flagName: String,
+        @StringForgery flagValue: String
+    ) {
+        // GIVEN
```
0xnm, line 6256: Empty `GIVEN`. Should it be like that?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -6246,6 +6246,108 @@ internal class RumViewScopeTest {
...
+                value = flagValue
+            ),
+            mockWriter
+        )
+
+        // THEN
+        argumentCaptor<ViewEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture())
+            assertThat(lastValue)
+                .apply {
```
0xnm, line 6271: `apply` is not needed here if I'm not wrong, direct call is possible. Same for all tests below.

0xnm (approved):
lgtm! I've added some minor comments to the tests

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -6246,6 +6246,108 @@ internal class RumViewScopeTest {
...
+            verify(mockWriter, times(2)).write(eq(mockEventBatchWriter), capture())
+            assertThat(lastValue)
+                .apply {
+                    hasFeatureFlag(flagName, flagValue)
+                }
+        }
+    }
+
+    // endregion
+
```
plousada, line 6350: Perhaps missing a test on not sending pre-existing Feature Flag after a new view is started.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
fuzzybinary: All the givens here are set up by the fixture setup. I'm not sure if I should keep the comment and have a blank body or remove the comment?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
fuzzybinary: Yeah, I'll probably need to see if I can add that to `DatadogRumMonitorTest` since that's what would handle changing the scope.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
fuzzybinary: 👍 Yeah, I guess I only need apply if I were checking multiple things.

0xnm: it is not needed even for multiple assertions, because each assertion returns `Assert` instance, it can be called in the chain.

0xnm (approved):
looks good!

PR #1267, opened 2023-02-06
RUMM-3005: Logs feature is using message bus to forward errors to RUM

0xnm:
### What does this PR do?

This PR makes Logs feature to use message bus to forward errors to RUM instead of having an explicit dependency on the RUM.

### DIFF
--- a/library/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/logger/DatadogLogHandler.kt
+++ b/library/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/logger/DatadogLogHandler.kt
@@ -11,8 +11,6 @@ import com.datadog.android.core.sampling.RateBasedSampler
 import com.datadog.android.core.sampling.Sampler
 import com.datadog.android.log.internal.domain.LogGenerator
 import com.datadog.android.log.model.LogEvent
-import com.datadog.android.rum.GlobalRum
-import com.datadog.android.rum.RumErrorSource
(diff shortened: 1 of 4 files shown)


library/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/logger/DatadogLogHandler.kt
```diff
@@ -77,7 +75,23 @@ internal class DatadogLogHandler(
...
+                    mapOf(
+                        "type" to "logger_error",
+                        "message" to message,
+                        "throwable" to throwable,
+                        "attributes" to attributes
+                    )
+                )
+            } else {
+                internalLogger.log(
+                    InternalLogger.Level.INFO,
```
xgouchet, line 90: Maybe `info` is a bit too much for this. Customers who don't use RUM will get "spammed" if they log many errors, wdyt?

xgouchet (approved):
Just one question but non blocking;  LGTM

library/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/logger/DatadogLogHandler.kt
0xnm: Indeed, good point. The problem is a bit wider imo, because maybe we need to notify only once that feature is missing during the SDK instance lifetime.

I created a dedicated task to tune up the logging for SDK v2.

PR #1268, opened 2023-02-07
RUMM-2864 add support to AP1

xgouchet:
### What does this PR do?

Adds support to the AP1 (Japan) datacenter

### Additional Notes

We also use subdomainless endpoints, which are not necessary since API v2 where the track is part of the path.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/DatadogSite.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/DatadogSite.kt
@@ -11,14 +11,16 @@ package com.datadog.android
  *
  * @property siteName Explicit site name property introduced in order to have a consistent SDK
  * instance ID (because this value is used there) in case if enum values are renamed.
+ * @property intakeHostName the host name for the given site.
  */
 // TODO RUMM-0000 Remove logs, traces, rum endpoint methods, because this class is for core SDK v2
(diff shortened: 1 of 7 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/DatadogSite.kt
```diff
@@ -31,51 +33,35 @@ enum class DatadogSite(val siteName: String) {
...
-            US5 -> DatadogEndpoint.LOGS_US5
-            US1_FED -> DatadogEndpoint.LOGS_US1_FED
-            EU1 -> DatadogEndpoint.LOGS_EU1
-        }
-    }
+    US1_FED("us1_fed", "browser-intake-ddog-gov.com"),
 
     /**
-     * Returns the endpoint to use to upload Spans to this site.
+     *  The STAGING site (internal usage only): [app.datad0g.com](https://app.datad0g.com).
```
0xnm, line 51: Is it useful to expose it in the public API? Isn't using custom endpoint URL API enough?

0xnm (approved):
lgtm! I like the change to switch to remove feature-specific intakes.

dd-sdk-android/src/main/kotlin/com/datadog/android/DatadogSite.kt
xgouchet: It is already exposed in the browser SDK, so it's not really a secret, and it can make our life during tests easier

PR #1278, opened 2023-02-10
RUMM-2599 ApplicationLaunch logic changes

fuzzybinary:
### What does this PR do?

ApplicationLaunch now starts immediately when the first event is sent to RUM provided we are in the Foreground. The view's start time is now the process start time (or as close as we can get it), and the `application_start` event is also triggered immediately. The "duration" of the `application_start` event is equal to the time from when the process started, until the first event sent to RUM.

### Additional Notes

This may change the timing for `application_start` for some clients, but should make it shorter in most cases since instead of timing to first view (or first event that causes ApplicationLaunch) we time until the first event sent to RUM regardless of what that event is.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/AppStartTimeProvider.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/AppStartTimeProvider.kt
@@ -1,0 +1,15 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal
+
+internal interface AppStartTimeProvider {
+    /**
+     * Provide the application start time in nanoseconds from device boot, or our best guess if
+     * the actual start time is not available
+     */
+    val appStartTimeNs: Long
+}
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/DefaultAppStartTimeProvider.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/DefaultAppStartTimeProvider.kt
@@ -1,0 +1,31 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal
+
+import android.annotation.SuppressLint
+import android.os.Build
+import android.os.Process
+import android.os.SystemClock
+import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import com.datadog.android.core.internal.system.DefaultBuildSdkVersionProvider
+import java.util.concurrent.TimeUnit
+
+internal class DefaultAppStartTimeProvider(
+    buildSdkVersionProvider: BuildSdkVersionProvider = DefaultBuildSdkVersionProvider()
+) : AppStartTimeProvider {
+
+    override val appStartTimeNs: Long by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        @SuppressLint("NewApi")
+        when {
+            buildSdkVersionProvider.version() >= Build.VERSION_CODES.N -> {
+                val diffMs = SystemClock.elapsedRealtime() - Process.getStartElapsedRealtime()
+                System.nanoTime() - TimeUnit.MILLISECONDS.toNanos(diffMs)
+            }
+            else -> RumFeature.startupTimeNs
+        }
+    }
+}
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
@@ -6,21 +6,16 @@
 
 package com.datadog.android.rum.internal.domain.scope
 
-import android.annotation.SuppressLint
 import android.app.ActivityManager
-import android.os.Build
-import android.os.Process
-import android.os.SystemClock
-import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
 import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
-import com.datadog.android.core.internal.system.BuildSdkVersionProvider
-import com.datadog.android.core.internal.system.DefaultBuildSdkVersionProvider
 import com.datadog.android.core.internal.utils.internalLogger
-import com.datadog.android.rum.internal.RumFeature
+import com.datadog.android.rum.internal.AppStartTimeProvider
+import com.datadog.android.rum.internal.DefaultAppStartTimeProvider
 import com.datadog.android.rum.internal.anr.ANRException
 import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.domain.Time
 import com.datadog.android.rum.internal.vitals.NoOpVitalMonitor
 import com.datadog.android.rum.internal.vitals.VitalMonitor
 import com.datadog.android.v2.api.InternalLogger
@@ -38,7 +33,7 @@ internal class RumViewManagerScope(
     private val cpuVitalMonitor: VitalMonitor,
     private val memoryVitalMonitor: VitalMonitor,
     private val frameRateVitalMonitor: VitalMonitor,
-    private val buildSdkVersionProvider: BuildSdkVersionProvider = DefaultBuildSdkVersionProvider(),
+    private val appStartTimeProvider: AppStartTimeProvider = DefaultAppStartTimeProvider(),
     private val contextProvider: ContextProvider
 ) : RumScope {
 
@@ -49,10 +44,18 @@ internal class RumViewManagerScope(
 
     @WorkerThread
     override fun handleEvent(event: RumRawEvent, writer: DataWriter<Any>): RumScope {
+        if (!applicationDisplayed) {
+            val isForegroundProcess = CoreFeature.processImportance ==
+                ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
+            if (isForegroundProcess) {
+                startApplicationLaunchView(event, writer)
+            }
+        }
+
         delegateToChildren(event, writer)
 
         if (event is RumRawEvent.StartView) {
-            startForegroundView(event, writer)
+            startForegroundView(event)
         } else if (childrenScopes.count { it.isActive() } == 0) {
             handleOrphanEvent(event, writer)
         }
@@ -71,6 +74,22 @@ internal class RumViewManagerScope(
     // endregion
 
     // region Internal
+
+    @WorkerThread
+    internal fun startApplicationLaunchView(event: RumRawEvent, writer: DataWriter<Any>) {
+        val applicationLaunchViewTimeNs = appStartTimeProvider.appStartTimeNs
+        val applicationLaunchViewTime = Time(
+            timestamp = TimeUnit.NANOSECONDS.toMillis(applicationLaunchViewTimeNs),
+            nanoTime = applicationLaunchViewTimeNs
+        )
+        val viewScope = createAppLaunchViewScope(applicationLaunchViewTime)
+        viewScope.handleEvent(
+            RumRawEvent.ApplicationStarted(event.eventTime, applicationLaunchViewTimeNs),
+            writer
+        )
+        childrenScopes.add(viewScope)
+        applicationDisplayed = true
+    }
 
     @WorkerThread
     private fun delegateToChildren(
@@ -96,12 +115,19 @@ internal class RumViewManagerScope(
         if (applicationDisplayed || !isForegroundProcess) {
             handleBackgroundEvent(event, writer)
         } else {
-            handleAppLaunchEvent(event, writer)
+            val isSilentOrphanEvent = event.javaClass in silentOrphanEventTypes
+            if (!isSilentOrphanEvent) {
+                internalLogger.log(
+                    InternalLogger.Level.WARN,
+                    InternalLogger.Target.USER,
+                    MESSAGE_MISSING_VIEW
+                )
+            }
         }
     }
 
     @WorkerThread
-    private fun startForegroundView(event: RumRawEvent.StartView, writer: DataWriter<Any>) {
+    private fun startForegroundView(event: RumRawEvent.StartView) {
         val viewScope = RumViewScope.fromEvent(
             this,
             sdkCore,
@@ -113,7 +139,7 @@ internal class RumViewManagerScope(
             contextProvider,
             trackFrustrations
         )
-        onViewDisplayed(event, viewScope, writer)
+        applicationDisplayed = true
         childrenScopes.add(viewScope)
     }
 
@@ -145,27 +171,6 @@ internal class RumViewManagerScope(
         }
     }
 
-    @WorkerThread
-    private fun handleAppLaunchEvent(
-        event: RumRawEvent,
-        actualWriter: DataWriter<Any>
-    ) {
-        val isValidAppLaunchEvent = event.javaClass in validAppLaunchEventTypes
-        val isSilentOrphanEvent = event.javaClass in silentOrphanEventTypes
-
-        if (isValidAppLaunchEvent) {
-            val viewScope = createAppLaunchViewScope(event)
-            viewScope.handleEvent(event, actualWriter)
-            childrenScopes.add(viewScope)
-        } else if (!isSilentOrphanEvent) {
-            internalLogger.log(
-                InternalLogger.Level.WARN,
-                InternalLogger.Target.USER,
-                MESSAGE_MISSING_VIEW
-            )
-        }
-    }
-
     private fun createBackgroundViewScope(event: RumRawEvent): RumViewScope {
         return RumViewScope(
             this,
@@ -184,13 +189,13 @@ internal class RumViewManagerScope(
         )
     }
 
-    private fun createAppLaunchViewScope(event: RumRawEvent): RumViewScope {
+    private fun createAppLaunchViewScope(time: Time): RumViewScope {
         return RumViewScope(
             this,
             sdkCore,
             RUM_APP_LAUNCH_VIEW_URL,
             RUM_APP_LAUNCH_VIEW_NAME,
-            event.eventTime,
+            time,
             emptyMap(),
             firstPartyHostHeaderTypeResolver,
             NoOpVitalMonitor(),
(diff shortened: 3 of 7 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/AppStartTimeProvider.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal
+
+internal interface AppStartTimeProvider {
+    /**
+     * Provide the application start time in nanoseconds from device boot, or our best guess if
```
plousada, line 11: > from device boot
From application boot, no?

Also, can we elaborate on the "our best guess" with the actual implementation? 
i.e. from RUM class load in Android < N

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/DefaultAppStartTimeProvider.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.annotation.SuppressLint
+import android.os.Build
+import android.os.Process
+import android.os.SystemClock
+import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import com.datadog.android.core.internal.system.DefaultBuildSdkVersionProvider
+import java.util.concurrent.TimeUnit
+
+internal class DefaultAppStartTimeProvider(
+    buildSdkVersionProvider: BuildSdkVersionProvider = DefaultBuildSdkVersionProvider()
```
mariusc83, line 18: if you're only doing this for tests there's a way in which you can avoid that and use directly the `Build.Version` static value inside. We have an extension class that we use to test these things: `ApiLevelExtension` class. Check for its usages and you will see it is pretty straightforward to use it.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -49,10 +44,18 @@ internal class RumViewManagerScope(
 
     @WorkerThread
     override fun handleEvent(event: RumRawEvent, writer: DataWriter<Any>): RumScope {
+        if (!applicationDisplayed) {
+            val isForegroundProcess = CoreFeature.processImportance ==
```
mariusc83, line 48: have you added a UnitTest for this case ? I didn't see it, maybe I missed it ?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/DefaultAppStartTimeProvider.kt
0xnm: `ApiLevelExtension` is not preferred, because it is using reflection, and with reflection after some time calls may be ignored for `static final` fields (which `Build.Version.SDK_INT` is), because such fields may be optimized (inlined I guess) by JIT.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/AppStartTimeProvider.kt
fuzzybinary: I can yes. I left it vague so that if it changes we wouldn't have to worry about updating the documentation or have it be wrong.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
fuzzybinary: Yes, it's just they didn't change. They're `M start a foreground ViewScope W handleEvent(StartView)` in RumViewManagerScopeTest

Now it's with any event but I think other tests capture that.

fuzzybinary: Yes it's just that they didn't change. They're `𝕄 start a foreground ViewScope 𝕎 handleEvent(StartView)` in `RumViewManagerScopeTest`. They only test for ViewEvents, but I think other test handle other possible events.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -243,12 +216,6 @@ internal class RumViewManagerScope(
             RumRawEvent.StartAction::class.java,
             RumRawEvent.StartResource::class.java
         )
-        internal val validAppLaunchEventTypes = arrayOf<Class<*>>(
```
0xnm, line 246: we need also to delete `Forge.validAppLaunchEvent()` in `RumRawEventExt` file in tests

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -663,52 +636,40 @@ internal class RumViewManagerScopeTest {
...
-    ) {
-        // Given
+        val appStartTimeNs = System.nanoTime() - TimeUnit.MILLISECONDS.toNanos(500)
+        whenever(mockAppStartTimeProvider.appStartTimeNs) doReturn appStartTimeNs
         CoreFeature.processImportance = RunningAppProcessInfo.IMPORTANCE_FOREGROUND
-        whenever(mockBuildSdkVersionProvider.version()) doReturn apiVersion
-        val childView: RumViewScope = mock()
-        val startViewEvent = forge.startViewEvent()
+
+        val fakeEvent = forge.addErrorEvent()
```
0xnm, line 656: we need to make sure that it is working with any event, so I would suggest to add a method in the `RumRawEventExt` file which will return any `RumRawEvent` available and use it here.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/AppStartTimeProvider.kt
fuzzybinary: Also, the time is actually nanoseconds since device boot (it's using `[System.nanoTime](https://developer.android.com/reference/java/lang/System#nanoTime())` or the `elapsedRealtime` functions, which are both timed from JVM or device boot time.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -663,52 +636,40 @@ internal class RumViewManagerScopeTest {
...
         // Then
-        argumentCaptor<RumRawEvent> {
-            verify(childView).handleEvent(capture(), same(mockWriter))
-
-            val event = firstValue as RumRawEvent.ApplicationStarted
-            assertThat(event.applicationStartupNanos)
-                .isCloseTo(System.nanoTime(), Offset.offset(TimeUnit.MILLISECONDS.toNanos(100)))
+        val appStartTimeMs = TimeUnit.NANOSECONDS.toMillis(appStartTimeNs)
+        argumentCaptor<ActionEvent> {
+            verify(mockWriter, atLeastOnce()).write(eq(mockEventBatchWriter), capture())
```
0xnm, line 664: Do we expect more calls here? If we expect only one call, we can use `times(1)`, or simply omit it and use `verify(mockWriter).write(...)`

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -49,10 +44,18 @@ internal class RumViewManagerScope(
 
     @WorkerThread
     override fun handleEvent(event: RumRawEvent, writer: DataWriter<Any>): RumScope {
+        if (!applicationDisplayed) {
+            val isForegroundProcess = CoreFeature.processImportance ==
+                ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
+            if (isForegroundProcess) {
+                startApplicationLaunchView(event, writer)
```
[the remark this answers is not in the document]
fuzzybinary: Yes. ApplicationLaunch will then take the time between `AppStartTimeProvied.appStartTimeNs` and the start time of the new View.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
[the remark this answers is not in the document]
fuzzybinary: Yes. I think that's the correct logic as that would skep application start time, but this is for sure where I'm confused about the logic.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/AppStartTimeProvider.kt
plousada: Ok, now I understand what you meant by "from device boot".
Initially I interpreted it as "we are returning the time difference between the app started and the time you booted your device"

plousada (approved):

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -514,23 +518,26 @@ internal class RumViewManagerScopeTest {
...
     @Test
-    fun `𝕄 start an AppLaunch ViewScope 𝕎 handleEvent { app not displayed, event is relevant }`(
+    fun `𝕄 start an AppLaunch ViewScope 𝕎 handleEvent { app not displayed, any event }`(
         forge: Forge
     ) {
         // Given
         CoreFeature.processImportance = RunningAppProcessInfo.IMPORTANCE_FOREGROUND
         testedScope.applicationDisplayed = false
+        val appStartTimeNs = System.nanoTime() - TimeUnit.MILLISECONDS.toNanos(500)
+        whenever(mockAppStartTimeProvider.appStartTimeNs) doReturn appStartTimeNs
```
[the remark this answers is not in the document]
fuzzybinary: Actually, because these tests use the time provided by the event (which is using System.nanoTime) I unfortunately can't forge the start time without changing how the mock events get their time.

I can do it but it might be some work...

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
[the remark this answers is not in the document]
0xnm: I'm curious if Android also can pre-warm apps (didn't dig into this topic tbh). I would suggest doing some tests using either our sample app or Reddit client. We can maybe emulate start by the user, background invocation using adb (like background invocation can maybe to send some intent to the service, etc).

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
0xnm: Oh yeah, it makes sense since we create Application Launch scope and then delegate action to it.

And I guess then the number of interactions vary from the type of the original event handled? Because some events can cause an instant write (like fatal errors).

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -72,6 +75,22 @@ internal class RumViewManagerScope(
 
     // region Internal
 
+    @WorkerThread
+    internal fun startApplicationLaunchView(event: RumRawEvent, writer: DataWriter<Any>) {
```
0xnm, line 79: Can be `private` I think

```suggestion
    private fun startApplicationLaunchView(event: RumRawEvent, writer: DataWriter<Any>) {
```

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -72,6 +75,22 @@ internal class RumViewManagerScope(
...
+            timestamp = TimeUnit.NANOSECONDS.toMillis(applicationLaunchViewTimeNs),
+            nanoTime = applicationLaunchViewTimeNs
+        )
+        val viewScope = createAppLaunchViewScope(applicationLaunchViewTime)
+        viewScope.handleEvent(
+            RumRawEvent.ApplicationStarted(event.eventTime, applicationLaunchViewTimeNs),
+            writer
+        )
+        childrenScopes.add(viewScope)
+        applicationDisplayed = true
```
0xnm, line 91: should we pull this line up? just in case if inner `handleEvent` also access this flag at some point.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -514,23 +518,27 @@ internal class RumViewManagerScopeTest {
...
 
         // When
         testedScope.handleEvent(fakeEvent, mockWriter)
 
         // Then
-        assertThat(testedScope.childrenScopes).hasSize(1)
+        val appStartTimeMs = TimeUnit.NANOSECONDS.toMillis(appStartTimeNs)
+        val scopes = if (fakeEvent is RumRawEvent.StartView) 2 else 1
+        assertThat(testedScope.childrenScopes).hasSize(scopes)
```
0xnm, line 537: minor suggestion (just to distinguish from `scopes` as collection)

```suggestion
        val scopesCount = if (fakeEvent is RumRawEvent.StartView) 2 else 1
        assertThat(testedScope.childrenScopes).hasSize(scopesCount)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -514,23 +518,27 @@ internal class RumViewManagerScopeTest {
...
-    fun `𝕄 start an AppLaunch ViewScope 𝕎 handleEvent { app not displayed, event is relevant }`(
+    fun `𝕄 start an AppLaunch ViewScope 𝕎 handleEvent { app not displayed, any event }`(
         forge: Forge
     ) {
         // Given
         CoreFeature.processImportance = RunningAppProcessInfo.IMPORTANCE_FOREGROUND
         testedScope.applicationDisplayed = false
-        val fakeEvent = forge.validAppLaunchEvent()
+        val fakeEvent = forge.anyRumEvent()
+        val appStartTimeNs = fakeEvent.eventTime.nanoTime - TimeUnit.MILLISECONDS.toNanos(500)
```
0xnm, line 528: or we can a bit more randomness and simply do `val appStartTimeNs = forge.aLong(min = 0L, max = fakeEvent.eventTime.nanoTime)`

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -663,52 +637,40 @@ internal class RumViewManagerScopeTest {
...
-    }
+        val fakeEvent = forge.anyRumEvent()
 
-    @Test
-    fun `𝕄 send ApplicationStarted event 𝕎 onViewDisplayed() {Nougat+}`(
-        forge: Forge,
-        @IntForgery(min = Build.VERSION_CODES.N) apiVersion: Int
-    ) {
-        // Given
+        val appStartTimeNs = fakeEvent.eventTime.nanoTime - TimeUnit.MILLISECONDS.toNanos(500)
```
0xnm, line 655:
```suggestion
        val appStartTimeNs = forge.aLong(min = 0L, max = fakeEvent.eventTime.nanoTime)
```

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/ActivityTrackingTest.kt
```diff
@@ -29,9 +29,18 @@ internal abstract class ActivityTrackingTest :
 
         instrumentation.waitForIdleSync()
 
-        // one for the Application start action
+        expectedEvents.add(ExpectedApplicationStart())
```
0xnm, line 32: should we align the naming used? Seems we have `ApplicationStart` and `ApplicationLaunch` at the same time.

UPD: seems we are a bit late, both definitions already live in the codebase.

0xnm (commented):
I've left few suggestions. However, it seems that unit-tests are failing, so we probably have something wrong in the logic.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
fuzzybinary: 👍 Though I think we probably want to go min, say fakeTime - 1000ms second to fakeTime - 100ms? What do you think?

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/ActivityTrackingTest.kt
fuzzybinary: I'll fix so it's "ApplicationStartEvent" and "ApplicaitonLaunchView" which I know doesn't make things that much clearer but at least it will be consistent.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
0xnm: It is ok for me. We don't have thresholds to filter out too big times, so we can accept any positive value as min (we just try to avoid hardcoded values in our tests).

0xnm (approved):
I think it looks good now, let's wait for the product decision first then before merging.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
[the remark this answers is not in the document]
xgouchet: Indeed that's the expected logic as we don't have any other event than the process start time to compute for the warm app launch. There's no catchable event except for the Activity's lifecycle which is already too late.

xgouchet (approved):

PR #1279, opened 2023-02-13
RUMM-3041 The RumEventMapper checks ViewEvents by reference

mariusc83:
### What does this PR do?

There was a bug in the `RumEventMapper` because for the `ViewEvent` we were comparing the original object with the mapped one by value and not by reference as required.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMapperTest.kt
+++ b/dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMapperTest.kt
@@ -363,7 +363,7 @@ internal class RumEventMapperTest {
         val mappedRumEvent = testedRumEventMapper.map(fakeRumEvent)
 
         // THEN
-        assertThat(mappedRumEvent).isEqualTo(fakeRumEvent)
+        assertThat(mappedRumEvent).isSameAs(fakeRumEvent)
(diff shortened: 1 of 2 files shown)


dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMapperTest.kt
```diff
@@ -456,6 +456,109 @@ internal class RumEventMapperTest {
...
+        // GIVEN
+        val fakeRumEvent = forge.getForgery<ViewEvent>()
+        whenever(mockViewEventMapper.map(fakeRumEvent))
+            .thenReturn(fakeRumEvent.copy())
+
+        // WHEN
+        val mappedRumEvent = testedRumEventMapper.map(fakeRumEvent)
+
+        // THEN
+        assertThat(mappedRumEvent).isEqualTo(fakeRumEvent)
```
0xnm, line 470: we need to use `isSameAs` here, because this line will be green even if another instance is returned (`isEqualTo` will call `equals`).

```suggestion
        assertThat(mappedRumEvent).isEqualTo(fakeRumEvent)
```

The `log` verification below helps though to avoid this test to pass-by.

0xnm (changes requested):

0xnm (approved):

PR #1283, opened 2023-02-14
RUMM-2613: Extract traces into the dedicated module

0xnm:
### What does this PR do?

This change moves all the APM-related code into the dedicated `dd-sdk-android-trace` module. Main points:

* Package name for tracing related code is now `com.datadog.android.trace` to align with naming used around other projects. This package name affects not only code in `dd-sdk-android-trace` module, but also in the main and `okhttp` modules from public API perspective.
* `io.opentracing` dependency now exists only for the `dd-sdk-android-trace` module. This means that we have to access `io.opentracing.util.GlobalTracer` for the configuration telemetry using reflection, because neither main module, nor RUM module which will be created in the future and will have telemetry code won't have `io.opentracing` dependency. Relying on the feature context which can be added during `AndroidTracer.Builder().build()` as a flag saying that there is a tracer is not relevant, because it may be a local tracer call, not global one. However, we will review this topic once again once we have tracer associated with SDK instance.
* Main module has one tracing class remaining, which is `TracingHeaderType`. This is because in the constructor of `DefaultFirstPartyHostHeaderTypeResolver` we have to provide hosts and header types and this class is also used in RUM. While it makes no sense to have header types for integration without tracing enabled, for now we will keep it like that and maybe will be able to change this API later (on the other hand having two different resolver class with and without headers, but with same hosts list for RUM and Traces makes no sense as well).


dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -235,13 +234,41 @@ internal class TelemetryEventHandler(
...
                     mobileVitalsUpdatePeriod = rumConfig?.vitalsMonitorUpdateFrequency?.periodInMs,
-                    useTracing = traceFeature != null && GlobalTracer.isRegistered(),
+                    useTracing = traceFeature != null && isGlobalTracerRegistered(),
                     trackNetworkRequests = trackNetworkRequests
                 )
             )
         )
     }
 
+    private fun isGlobalTracerRegistered(): Boolean {
```
mariusc83, line 244: Do you think this is really needed as long as the `traceFeature!=null` ?

dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -235,13 +234,41 @@ internal class TelemetryEventHandler(
...
+        // TODO RUMM-0000 we don't reference io.opentracing from RUM directly, so using this.
+        // alternatively we can afford maybe to reference it, because it seems transitive size
+        // of io.opentracing is like 30 KBs in total?
+        // Would be nice to add the test with the flavor which is has no io.opentracing and test
+        // for obfuscation enabled case.
+        return try {
+            val globalTracerClass = Class.forName("io.opentracing.util.GlobalTracer")
```
mariusc83, line 251: This scares me :)

library/dd-sdk-android-trace/build.gradle.kts
```diff
@@ -43,6 +43,8 @@ android {
         minSdk = AndroidConfig.MIN_SDK
         targetSdk = AndroidConfig.TARGET_SDK
         setLibraryVersion()
+
+        consumerProguardFiles("consumer-rules.pro")
```
mariusc83, line 47: what is the difference from :
`
proguardFiles(...)
`
Is this the new way of adding a file or this is just merging with the default project file ?

library/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/utils/config/DatadogSingletonTestConfiguration.kt
```diff
@@ -0,0 +1,30 @@
...
+internal class DatadogSingletonTestConfiguration :
+    MockTestConfiguration<SdkCore>(SdkCore::class.java) {
+
+    override fun setUp(forge: Forge) {
+        super.setUp(forge)
+        Datadog::class.java.setStaticValue("globalSdkCore", mockInstance)
+    }
+
+    override fun tearDown(forge: Forge) {
+        Datadog::class.memberFunctions.first { it.name == "stop" }
```
mariusc83, line 26: you do not want also to invalidate the `globalSdkCore` instance here ?

library/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/utils/forge/SpanEventForgeryFactory.kt
```diff
@@ -77,20 +76,20 @@ internal class SpanEventForgeryFactory : ForgeryFactory<SpanEvent> {
...
-    fun Forge.exhaustiveMetrics(): MutableMap<String, Number> {
+    private fun Forge.exhaustiveMetrics(): MutableMap<String, Number> {
         return listOf(
             aLong(),
             anInt(),
             aFloat(),
             aDouble()
         ).map { anAlphabeticalString() to it as Number }
-            .toMutableMap()
+            .toMap(mutableMapOf())
```
mariusc83, line 86: what is the advantage ?

0xnm: no advantage, the original method is just private in the main module, so it was easier to to the replacement than to share it.

library/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/utils/config/DatadogSingletonTestConfiguration.kt
0xnm: `Datadog.stop` method will set `NoOpSdkCore` to the `globalSdkCore` internally.

library/dd-sdk-android-trace/build.gradle.kts
0xnm: This makes that file to be shipped with `AAR`, so it is consumed later on the app side and rules are taken into the account. We already doing the similar for main module.

dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
0xnm: Let's keep it as it is for now, because originally both feature check and tracer check were done.

mariusc83 (approved):

PR #1286, opened 2023-02-16
RUMM-3054 Correctly report viewport size in SR

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SystemInformation.kt
```diff
@@ -6,4 +6,9 @@
 
 package com.datadog.android.sessionreplay.internal.recorder
 
-internal data class OrientationChanged(val width: Int, val height: Int)
+import android.content.res.Configuration
+
+internal data class SystemInformation(
+    val screenBounds: ViewGlobalBounds,
```
0xnm, line 12: maybe we need to rename `ViewGlobalBounds` class? Because here it is used in the context of the screen, and internally in the `ViewGlobalBounds` there is nothing specific for the view, it is just area description with starting point + width + height.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/MiscUtils.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun resolveSystemInformation(activity: Activity): SystemInformation {
+        return SystemInformation(
+            screenBounds = resolveScreenBounds(activity),
+            screenOrientation = activity.resources.configuration.orientation
+        )
+    }
+
+    @Suppress("DEPRECATION")
+    private fun resolveScreenBounds(activity: Activity): ViewGlobalBounds {
+        val windowManager = activity.windowManager ?: return ViewGlobalBounds(0, 0, 0, 0)
```
0xnm, line 42: in which case `windowManager` can be `null`?

0xnm (approved):
lgtm!

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SystemInformation.kt
mariusc83: yes...good point, I will just name it GlobalBounds

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/MiscUtils.kt
mariusc83: well not sure because I could not get this information from under the hood but I remember it happened to me at some point so I just wanted to make sure just in case. That method is not marked as `@NonNull` in Java.

mariusc83: yeah...only if I would have that available in the code yet :). I have a big TODO task to add all the logs in the `session-replay` module. I will add a todo here too.

PR #1288, opened 2023-02-17
RUMM-3051: Remove okhttp3.internal package usage

### LINKED ISSUE #1282: Use of an internal okHttp package causes issues/crashes when using a more recent version of okHttp(4+).
https://github.com/DataDog/dd-sdk-android/blob/dde54aa25954bcdcaffb45b89a84d1ee5450e7ff/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RequestUniqueIdentifier.kt#L10

According to their own docs this is a little bit [naughty](https://square.github.io/okhttp/changelogs/upgrading_to_okhttp_4/#internal-api-changes)

0xnm:
### What does this PR do?

Fixes #1282. This PR removes `okhttp3.internal` package usage, we were using `okhttp3.internal.Util.EMPTY_REQUEST`.

There is a [warning](https://square.github.io/okhttp/changelogs/upgrading_to_okhttp_4/#internal-api-changes) about `okhttp3.internal` package usage for OkHttp 4.x, but with OkHttp 4.x this usage didn't cause any problems - the class and field were still there.

However, it seems things are finally moved for OkHttp 5.x (which is in alpha for now, changelog is [here](https://square.github.io/okhttp/changelogs/changelog/)), so to be on the safe side we are getting rid of reference to `okhttp3.internal`.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RequestUniqueIdentifier.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RequestUniqueIdentifier.kt
@@ -7,7 +7,7 @@
 package com.datadog.android.core.internal.net
 
 import okhttp3.Request
-import okhttp3.internal.Util
+import java.io.IOException
 
 /**
  * Generates an identifier to uniquely track requests.
@@ -16,11 +16,25 @@ internal fun identifyRequest(request: Request): String {
     val method = request.method()
     val url = request.url()
     val body = request.body()
-    return if (body == null || body == Util.EMPTY_REQUEST) {
+    return if (body == null) {
         "$method•$url"
     } else {
-        val contentLength = body.contentLength()
+        val contentLength = try {
+            body.contentLength()
+        } catch (@Suppress("SwallowedException") ioe: IOException) {
+            0
+        }
         val contentType = body.contentType()
-        "$method•$url•$contentLength•$contentType"
+        // TODO RUMM-3062 It is possible that if requests are say GZIPed (as an example), or maybe
+        //  streaming case (?), they all will have contentLength = -1, so if they target the same URL
+        //  they are going to have same identifier, messing up reporting.
+        //  -1 is valid return value for contentLength() call according to the docs and stands
+        //  for "unknown" case.
+        //  We need to have a more precise identification.
+        if (contentType != null || contentLength != 0L) {
+            "$method•$url•$contentLength•$contentType"
+        } else {
+            "$method•$url"
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RequestUniqueIdentifier.kt
```diff
@@ -16,11 +16,25 @@ internal fun identifyRequest(request: Request): String {
...
     } else {
-        val contentLength = body.contentLength()
+        val contentLength = try {
+            body.contentLength()
+        } catch (@Suppress("SwallowedException") ioe: IOException) {
+            0
+        }
         val contentType = body.contentType()
-        "$method•$url•$contentLength•$contentType"
+        // TODO RUMM-3062 It is possible that if requests are say GZIPed (as an example, or maybe
```
mariusc83, line 28: I thin you are missing something after `as an example` ?

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RequestUniqueIdentifier.kt
```diff
@@ -16,11 +16,25 @@ internal fun identifyRequest(request: Request): String {
...
+        //  streaming case?), they all will have contentLength = -1, so if they target the same URL
+        //  they are going to have same identifier, messing up reporting.
+        //  -1 is valid return value for contentLength() call according to the docs and stands
+        //  for "unknown" case.
+        //  We need to have a more precise identification.
+        if (contentType != null || contentLength != 0L) {
```
mariusc83, line 34: Aren't we worried that we will have too many collisions if we follow what you stated above ? I guess there are many apps there with GZIPed requests.

0xnm: No, this is the same behavior what we have right now (because `EMPTY_REQUEST` replace was the body with `contentType == null` and `contentLength == 0`), so we don't introduce any changes to the current behavior.

And later we can improve, this issue exists today as well.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/net/RequestUniqueIdentifier.kt
0xnm: GZIP is an example :) I put the bracket at the wrong place, will update it.

mariusc83 (approved):

xgouchet (approved):

PR #1293, opened 2023-02-20
RUMM-3064: Register WebView integration RUM and Logs features dynamically

0xnm:
### What does this PR do?

* `webViewTrackingHosts` is removed from the Core configuration, they can now only be provided in the `DatadogEventBridge`. This makes sense, because it was used only in `DatadogEventBridge`.
* To follow-up change above, `DatadogEventBridge` constructor without specifying hosts tracked is removed.
* Also `DatadogEventBridge` is going to sanitize hosts which it receives (it turns out there was a discrepancy, because we sanitized them in the Core configuration, during `setWebViewTrackingHosts` call, but didn't do the same for hosts provided to the `DatadogEventBridge` directly).
* WebView RUM and Logs features are registered dynamically now (using `Datadog#registerFeature`), during `DatadogEventBridge` construction time, also they re-use request factories of RUM and Logs, meaning to make things work a) RUM and Logs feature should be registered; b) `DatadogEventBridge#setup` is called after SDK initialization. This is fine, because current logic is exactly the same - `DatadogEventBridge` was expecting RUM and Logs to be enabled by the time it is constructed. We may enrich support of the dynamic configuration later (check features each time we received message from WebView), but it is not a priority for now (and we didn't receive any request for it yet).
* WebView RUM and Logs features won't become public and are going to reuse the same request factories from RUM and Logs, allowing them to be private in their own packages as well (since we need only the interfaces) and aligning network configuration (like target site) for WebView RUM and Logs features without the need to introduce their own configs. Also there won't be need in the explicit dependency on the RUM and Logs packages, because we are going to use only interfaces defined in the core package.

### Motivation

### Additional Notes


xgouchet (approved):

kayayarai (approved):
👍  for the docs updates

mariusc83 (approved):

PR #1297, opened 2023-02-22
RUMM-2996: Extract WebView RUM and Logs into dedicated module

0xnm:
### What does this PR do?

This change does extraction of WebView RUM and Logs into the dedicated module.

Notable changes:

* Removed the usage of `RumDataWriter`, because events coming were `JsonObjects` anyway, so there was nothing specific inside for them.


library/dd-sdk-android-webview/src/test/kotlin/com/datadog/android/webview/internal/MixedWebViewEventConsumerTest.kt
```diff
@@ -94,7 +94,9 @@ internal class MixedWebViewEventConsumerTest {
         // Then
         argumentCaptor<Pair<JsonObject, String>> {
             verify(mockLogsEventConsumer).consume(capture())
-            assertThat(firstValue.first).isEqualTo(fakeBundledEvent)
+            // toString call because of how Gson is comparing Float/Double vs LazilyParsedNumber
+            // for JsonPrimitive https://github.com/google/gson/issues/1864
```
mariusc83, line 98: I think there is a solution for that:
https://github.com/DataDog/dd-sdk-android/blob/144d7babd3928157bf417f08a01cf85042c8f7a3/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/sessionreplay/SrTest.kt#L44

mariusc83 (approved):
LGTM

library/dd-sdk-android-webview/src/test/kotlin/com/datadog/android/webview/internal/MixedWebViewEventConsumerTest.kt
0xnm: it can be used, but I would rather just compare strings here, given that anyway during the WebView ingerstion flow we don't work with Java/Kotlin types.

PR #1304, opened 2023-02-27
RUMM-3082: Remove CoreFeature usage from RumMonitor/DatadogRumMonitor

0xnm:
### What does this PR do?

This change removes references to `CoreFeature` (which will stay as internal) from the `RumMonitor` and `DatadogRumMonitor` (meaning from `RumViewScope`, `RumActionScope`, etc.).

This is mostly achieved by the properties of `SdkCore`. But for the action/resource scopes we need to know the network info at the scope start, and since I don't think we should expose `networkInfo` in the `SdkCore` (quite specific property), it is exposed in the `AdvancedSdkCore` instead. Later however, we may promote this property to the `SdkCore`.

Expecting SDK instance to implement `AdvancedSdkCore` is not an issue, because it is not created outside anyway and managed by us.

Also `RumFeature.Builder` now requires `applicationId` explicitly, this will allow later to remove `rumApplicationId` from `Credentials` object.

### DIFF
--- a/dd-sdk-android/apiSurface
+++ b/dd-sdk-android/apiSurface
@@ -539,6 +539,8 @@ data class com.datadog.android.v2.api.context.TimeInfo
   constructor(Long, Long, Long, Long)
 data class com.datadog.android.v2.api.context.UserInfo
   constructor(kotlin.String? = null, kotlin.String? = null, kotlin.String? = null, Map<kotlin.String, kotlin.Any?> = emptyMap())
+interface com.datadog.android.v2.core.InternalSdkCore : com.datadog.android.v2.api.SdkCore
+  val networkInfo: com.datadog.android.v2.api.context.NetworkInfo
 interface com.datadog.android.v2.core.storage.DataWriter<T>
   fun write(com.datadog.android.v2.api.EventBatchWriter, T): Boolean
 class com.datadog.tools.annotation.NoOpImplementation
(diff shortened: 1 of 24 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/AdvancedSdkCore.kt
```diff
@@ -0,0 +1,21 @@
...
+
+package com.datadog.android.v2.core
+
+import com.datadog.android.v2.api.SdkCore
+import com.datadog.android.v2.api.context.NetworkInfo
+
+/**
+ * FOR INTERNAL USAGE ONLY. THIS INTERFACE CONTENT MAY CHANGE WITHOUT NOTICE.
+ */
+interface AdvancedSdkCore : SdkCore {
```
xgouchet, line 15: Maybe instead of Advanced we can rename it InternalSdkCore to discourage people from using it?

xgouchet (approved):

mariusc83 (approved):
LGTM

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/AdvancedSdkCore.kt
0xnm: done. I will also introduce something similar (I mean public, but internal interface) later for writing last view event in RUM. Later we maybe can review such internal APIs usage and find a better approach.

PR #1305, opened 2023-02-27
Fix timing of ApplicationLaunch view and application_start events

fuzzybinary:
### What does this PR do?

Fix the timing of ApplicaitonLaunch vie and application_start events. The events were incorrectly using the nano time for their timestamp instead of using an ms timestamp. We use the difference between the triggering event's timestamp and nano time to determine how far apart they are, and add that to the application start time.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
@@ -91,9 +91,10 @@ internal class RumViewManagerScope(
             nanoTime = processStartTime
         )
         val viewScope = createAppLaunchViewScope(applicationLaunchViewTime)
+        val startupTime = event.eventTime.nanoTime - processStartTime
         applicationDisplayed = true
         viewScope.handleEvent(
-            RumRawEvent.ApplicationStarted(event.eventTime, processStartTime),
+            RumRawEvent.ApplicationStarted(applicationLaunchViewTime, startupTime),
             writer
         )
         childrenScopes.add(viewScope)
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
@@ -77,20 +77,23 @@ internal class RumViewManagerScope(
 
     @WorkerThread
     private fun startApplicationLaunchView(event: RumRawEvent, writer: DataWriter<Any>) {
-        val applicationLaunchViewTimeNs = appStartTimeProvider.appStartTimeNs
-        // appStartTimeNs is based on elapsed system time. Timestamp needs to be currentMs.
-        // Figure out the difference then apply that to the applicationLaunchView to get the
-        // timestamp
-        val timestamp = (event.eventTime.timestamp - TimeUnit.NANOSECONDS.toMillis(event.eventTime.nanoTime)) +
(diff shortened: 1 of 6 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -78,8 +78,13 @@ internal class RumViewManagerScope(
     @WorkerThread
     private fun startApplicationLaunchView(event: RumRawEvent, writer: DataWriter<Any>) {
         val applicationLaunchViewTimeNs = appStartTimeProvider.appStartTimeNs
+        // appStartTimeNs is based on elapsed system time. Timestamp needs to be currentMs.
```
0xnm, line 81: I find this description confusing. `timestamp` doesn't need to be _current_ time.

Basically `Time#timestamp` and `Time#nanoTime` are pointing to the same instant (moment of the time), while having different base - `timestamp` is based on the Unix epoch start, `nanoTime` is based on the JVM start time (for Android I think it is efficiently the process start/fork time). 

So I guess the comment wanted to say that `timestamp` should have the same base as `System.currentTimeMillis` (just the shortened abbreviation is used)?

So what we are doing below is first calculating the different between 2 instant representations, which is `event.eventTime.timestamp - TimeUnit.NANOSECONDS.toMillis(event.eventTime.nanoTime)` to find out when actually process has started in terms of epoch time, and then apply the time since this process start (which is `applicationLaunchViewTimeNs`) to find when even happened in the epoch time format.

So I would suggest for the clarity to introduce the variable `processStartNs` with the description what is this (also maybe we need to keep calculation in the ns and convert to the ms at the very end, better probably precision-wise (although 10^6 different between ns and ms will erase this precision gain anyway)), because I think without it the expression below is not quite clear.

fuzzybinary: 👍 I'll take another stab at it.

fuzzybinary:
@0xnm took another stab at it. Let me know if that's clearer.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -77,15 +77,21 @@ internal class RumViewManagerScope(
 
     @WorkerThread
     private fun startApplicationLaunchView(event: RumRawEvent, writer: DataWriter<Any>) {
-        val applicationLaunchViewTimeNs = appStartTimeProvider.appStartTimeNs
+        val processStartTime = appStartTimeProvider.appStartTimeNs
+        // processStartTime is the time in nanoseconds since VM start. To get a timestamp, we want
+        // to convert it to milliseconds since epoch provided by System.currentTimeMillis.
+        // To do so, we take the offset of those times in the event time, which should be consistent,
+        // then add that to our processStartTime to get the correct value.
+        val timestampNs = (TimeUnit.MILLISECONDS.toNanos(event.eventTime.timestamp - event.eventTime.nanoTime)) +
```
0xnm, line 85: shouldn't it be like `(TimeUnit.MILLISECONDS.toNanos(event.eventTime.timestamp) - event.eventTime.nanoTime)` here? given that `nanoTime` is actually in nanoseconds.

fuzzybinary: 🤦‍♂️ Yes. Fixed.

0xnm (approved):

xgouchet (approved):

PR #1310, opened 2023-03-01
Add internal telemetry configuration sampling rate

fuzzybinary:
### What does this PR do?

Add a property that can be supplied through additionalConfiguration which sets the extra sample rate for configuration telemetry. This serves as a way for the cross platform SDKs to modify whether telemetry gets sent in order to test our own telemetry logic.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -256,6 +256,7 @@ object Datadog {
     internal const val DD_SOURCE_TAG = "_dd.source"
     internal const val DD_SDK_VERSION_TAG = "_dd.sdk_version"
     internal const val DD_APP_VERSION_TAG = "_dd.version"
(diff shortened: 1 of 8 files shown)


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -34,7 +34,7 @@ import com.datadog.android.telemetry.model.TelemetryConfigurationEvent.ViewTrack
 internal class TelemetryEventHandler(
     internal val sdkCore: SdkCore,
     internal val eventSampler: Sampler,
-    internal val configurationExtraSampler: Sampler = RateBasedSampler(20f.percent()),
+    internal var configurationExtraSampler: Sampler = RateBasedSampler(20f.percent()),
```
0xnm, line 37: it doesn't need to be `var` it seems?

fuzzybinary: Yes, thanks. I changed it when I thought I would take a different approach but found one that worked better.

0xnm (approved):

PR #1312, opened 2023-03-02
RUMM-2903 handle device refresh rate with navigation strategy

xgouchet:
### What does this PR do?

In order to get the device's refresh rate to scale down or up the detected app's refresh rate , we need to grab the activity from the navigation controller.

### Motivation

Ensure that the refresh rate is properly reported in all use cases

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -13,6 +13,7 @@ import android.os.Build
 import android.view.WindowManager
 import androidx.annotation.WorkerThread
 import androidx.fragment.app.Fragment
+import androidx.navigation.NavController
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
 import com.datadog.android.core.internal.system.DefaultBuildSdkVersionProvider
@@ -34,6 +35,7 @@ import com.datadog.android.rum.model.ActionEvent
 import com.datadog.android.rum.model.ErrorEvent
 import com.datadog.android.rum.model.LongTaskEvent
 import com.datadog.android.rum.model.ViewEvent
+import com.datadog.android.rum.tracking.NavigationViewTrackingStrategy
 import com.datadog.android.v2.api.InternalLogger
 import com.datadog.android.v2.api.SdkCore
 import com.datadog.android.v2.core.internal.ContextProvider
@@ -990,6 +992,10 @@ internal open class RumViewScope(
             is Activity -> key
             is Fragment -> key.activity
             is android.app.Fragment -> key.activity
+            is NavigationViewTrackingStrategy.NavigationKey -> {
+                navControllerActivityField?.isAccessible = true
+                navControllerActivityField?.get(key.controller) as? Activity
+            }
             else -> null
         } ?: return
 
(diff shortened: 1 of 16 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1101,5 +1107,9 @@ internal open class RumViewScope(
         private fun invertValue(value: Double): Double {
             return if (value == 0.0) 0.0 else 1.0 / value
         }
+
+        val navControllerActivityField = NavController::class.java.declaredFields.firstOrNull {
```
fuzzybinary, line 1111: Why did this get put at the bottom of the companion object instead of with the other `val`s?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -6477,6 +6480,101 @@ internal class RumViewScopeTest {
...
+                        fakeDatadogContext.deviceInfo.architecture
+                    )
+                    hasOsInfo(
+                        fakeDatadogContext.deviceInfo.osName,
+                        fakeDatadogContext.deviceInfo.osVersion,
+                        fakeDatadogContext.deviceInfo.osMajorVersion
+                    )
+                    hasConnectivityInfo(fakeDatadogContext.networkInfo)
+                    hasServiceName(fakeDatadogContext.service)
+                    hasVersion(fakeDatadogContext.version)
```
fuzzybinary, line 6571: Should we be checking all of these properties in this test?

fuzzybinary (approved):

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
xgouchet: Mostly because I added it there, I could move it up but it doesn't really change anything IMO

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
xgouchet: Yes, just to make sure that the event is complete as expected

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1005,6 +1007,10 @@ internal open class RumViewScope(
             is Activity -> key
             is Fragment -> key.activity
             is android.app.Fragment -> key.activity
+            is NavigationViewTrackingStrategy.NavigationKey -> {
+                navControllerActivityField?.isAccessible = true
```
0xnm, line 1011: maybe we can add some log entry when `navControllerActivityField` is not found (because of the private API changes, for example)? this will simplify the debugging if things don't work.

0xnm (approved):
lgtm! I've added one comment in order to simplify debugging.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
xgouchet: Yes indeed, I'll add one

PR #1314, opened 2023-03-03
REPLAY-1422 SR records all the decor views at any moment

mariusc83:
### What does this PR do?

There was a problem in our SR recording strategy as we were only recording the `PhoneWindow` as the `Activity` main window. 
This approach is not covering all the cases in Android as most of the `Menu` or `Spinner` lists are using the `PopUpWindow` to display their content. The `PopUpWindow` uses its proper `PopUpDecorView` as the root view and this was not taken into account so far.
This PR doesn't solve the problem entirely because I could not find a way to grab the `PopUpWindow` reference yet in order to attach my `TouchEventListener` and intercept the `TouchEvents` so currently we are recording the views but not the user interaction on those pop - ups. The `PopUpWindow` is not accessible, not even through reflection so for now we are stuck here.
On the other hand we found a way to mitigate this issue for `DialogFragments`, in this case we could get a hand on `dialog.window` and solve at least this part. 

Please note that this problem exists also in `Rum` as we cannot record user interaction on those `PopUpWindows`.

### Motivation

### Additional Notes


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -0,0 +1,48 @@
...
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            WindowInspector.getGlobalWindowViews()
+        } else {
+            getGlobalWindowViewApi29()
+        }
+    }
+
+    @Suppress("SwallowedException", "TooGenericExceptionCaught")
+    @SuppressLint("PrivateApi", "DiscouragedPrivateApi")
+    private fun getGlobalWindowViewApi29(): List<View> {
```
xgouchet, line 27: Nitpicking: I'd rename that `getGlobalWindowViewLegacy` or `getGlobalWindowViewReflection` or `getGlobalWindowViewPreApi29` as its a method used *before* API 29.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -0,0 +1,48 @@
...
+        try {
+            val globalWmClass = Class.forName("android.view.WindowManagerGlobal")
+            val globalWmInstance = globalWmClass.getMethod("getInstance").invoke(null)
+            val viewsField: Field = globalWmClass.getDeclaredField("mViews")
```
xgouchet, line 31: It'd be better to have a lazy (static?) property to hold `globalWmClass` and `viewsField` as those will never change. That way you only compute them once

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayLifecycleCallback.kt
```diff
@@ -41,25 +41,25 @@ internal class SessionReplayLifecycleCallback(
...
-        RecordedDataProcessor(
-            rumContextProvider,
-            timeProvider,
-            processorExecutorService,
-            recordWriter,
-            recordCallback
-        ),
-        SnapshotProducer(privacy.mapper()),
-        timeProvider
+    val processor = RecordedDataProcessor(
```
0xnm, line 44: can be `internal` it seems.

```suggestion
    internal val processor = RecordedDataProcessor(
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewOnDrawInterceptor.kt
```diff
@@ -0,0 +1,59 @@
...
+        val onDrawListener = WindowsOnDrawListener(
+            ownerActivity,
+            decorViews,
+            processor,
+            snapshotProducer
+        )
+        decorViews.forEach { decorView ->
+            decorOnDrawListeners[decorView] = onDrawListener
+            decorView.setOnTouchListener { v, _ ->
+                v?.performClick()
```
0xnm, line 34: `view` can be `null` here? or it is just because of Java API. Also not quite clear why we need this call here, won't we steal user focus/input by doing it?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -0,0 +1,81 @@
...
+import android.view.View
+import android.view.inspector.WindowInspector
+import java.lang.reflect.Field
+
+@SuppressLint("PrivateApi")
+internal object WindowInspector {
+
+    private val GLOBAL_WM_CLASS by lazy {
+        try {
+            return@lazy Class.forName("android.view.WindowManagerGlobal")
```
0xnm, line 20: this will probably be obfuscated in the release build, so we need to add Proguard rules to be shipped with the library to prevent this obfuscation. Same for the reflection calls below.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspectorTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        assertThat(WindowInspector.getGlobalWindowViewsLegacy(null, mock())).isEmpty()
+    }
+
+    @Test
+    fun `M return emptyList W getGlobalWindowViewsLegacy { null mViews field}`() {
+        assertThat(WindowInspector.getGlobalWindowViewsLegacy(mock(), null)).isEmpty()
+    }
+
+    @Test
+    fun `M return emptyList W getGlobalWindowViewsLegacy { null WN instance and null mViews }`() {
```
0xnm, line 46:
```suggestion
    fun `M return emptyList W getGlobalWindowViewsLegacy { null WM instance and null mViews }`() {
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
mariusc83: Good point

xgouchet: Classes from the android framework are not embedded in the apk, but provided by the OS, so they are not obfuscated

0xnm: indeed, good point.

xgouchet (approved):

0xnm (approved):

PR #1325, opened 2023-03-07
REPLAY-1447 Introduce the ExtensionSupport public API

mariusc83:
### What does this PR do?

In this PR we are introducing a new capability in the SR Public API which will allow users to provide an `ExtensionSupport` implementation to support UI frameworks different than the default ones (e.g. Material). For now the `ExtensionSupport` has only one method/property -> ` getCustomViewMappers(): Map<Class<*>, WireframeMapper<View, *>>` which could be used to specify a custom `WireframeMapper` for a specific `View` type that is not supported by default by the core SR sdk. 
This method will be leveraged in the next PR to provide a `SliderWireframeMapper` for the Material `Slider` component.

Please note that a fix was added for the `ApiSurfaceGenerator` in order to support wildcard generics in functions.

### Motivation

### Additional Notes

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/ExtensionSupport.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/ExtensionSupport.kt
@@ -1,0 +1,25 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay
+
+import android.view.View
+import com.datadog.android.sessionreplay.internal.recorder.mapper.WireframeMapper
+
+/**
+ * In case you need to provide different configuration for a specific Android UI framework that
+ * is not supported by our SR instrumentation layer (e.g Material elements)
+ * you can implement this class.
+ * @see [SessionReplayConfiguration.Builder.addExtensionSupport]
+ */
+interface ExtensionSupport {
+    /**
+     * Use this method if you want to apply a custom [WireframeMapper] for a specific [View] type
+     * and for a specific [SessionReplayPrivacy].
+     * @return the [View] type to custom [WireframeMapper] map grouped by [SessionReplayPrivacy].
+     */
+    fun getCustomViewMappers(): Map<SessionReplayPrivacy, Map<Class<*>, WireframeMapper<View, *>>>
+}
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
@@ -6,9 +6,11 @@
 
 package com.datadog.android.sessionreplay
 
+import android.view.View
 import com.datadog.android.DatadogEndpoint
 import com.datadog.android.DatadogSite
 import com.datadog.android.core.configuration.Configuration
+import com.datadog.android.sessionreplay.internal.recorder.mapper.WireframeMapper
 
 /**
  * An object describing the configuration of the Session Replay.
@@ -16,7 +18,8 @@ import com.datadog.android.core.configuration.Configuration
 data class SessionReplayConfiguration
 internal constructor(
     internal val endpointUrl: String,
-    internal val privacy: SessionReplayPrivacy
+    internal val privacy: SessionReplayPrivacy,
+    internal val extensionSupport: ExtensionSupport
 ) {
 
     /**
@@ -25,6 +28,18 @@ internal constructor(
     class Builder {
         private var endpointUrl = DatadogEndpoint.SESSION_REPLAY_US1
         private var privacy = SessionReplayPrivacy.MASK_ALL
+        private var extensionSupport: ExtensionSupport = NoOpExtensionSupport()
+
+        /**
+         * Adds an extension support implementation. This is mostly used when you want to provide
+         * different behaviour of the Session Replay when using other Android UI frameworks
+         * than the default ones.
+         * @see [ExtensionSupport.getCustomViewMappers]
+         */
+        fun addExtensionSupport(extensionSupport: ExtensionSupport): Builder {
+            this.extensionSupport = extensionSupport
+            return this
+        }
 
         /**
          * Let the Session Replay target your preferred Datadog's site.
@@ -59,8 +74,22 @@ internal constructor(
         fun build(): SessionReplayConfiguration {
             return SessionReplayConfiguration(
                 endpointUrl = endpointUrl,
-                privacy = privacy
+                privacy = privacy,
+                extensionSupport = extensionSupport
             )
         }
     }
+
+    // region Internal
+
+    internal fun customMappers(): Map<Class<*>, WireframeMapper<View, *>> {
+        extensionSupport.getCustomViewMappers().entries.forEach {
+            if (it.key.name == privacy.name) {
+                return it.value
+            }
+        }
+        return emptyMap()
+    }
+
+    // endregion
 }
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/GlobalBounds.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/GlobalBounds.kt
@@ -6,7 +6,18 @@
 
 package com.datadog.android.sessionreplay.internal.recorder
 
-internal data class GlobalBounds(
+/**
+ * Defines the dimension and positions in Global coordinates for a geometry. By Global we mean that
+ * the View position will not be relative to its parent but to the Device screen.
+ * These dimensions are already normalized according with the current device screen density.
+ * Example: if a device has a DPI = 2, the value of the dimension or position is divided by 2 to get
+ * a normalized value.
+ * @param x as the position on X axis
+ * @param y as the position on Y axis
+ * @param width as the width
+ * @param height as the height
+ */
+data class GlobalBounds(
     val x: Long,
     val y: Long,
     val width: Long,
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/GenericWireframeMapper.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/GenericWireframeMapper.kt
@@ -17,60 +17,95 @@ import android.widget.TextView
 import androidx.appcompat.widget.SwitchCompat
 import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
 import com.datadog.android.sessionreplay.model.MobileSegment
+import java.util.LinkedList
+
+typealias MapperCondition = (View) -> Boolean
 
 internal abstract class GenericWireframeMapper(
-    private val viewWireframeMapper: ViewWireframeMapper,
+    viewWireframeMapper: ViewWireframeMapper,
     internal val imageMapper: ViewScreenshotWireframeMapper,
-    private val textMapper: TextWireframeMapper,
-    private val buttonMapper: ButtonMapper,
-    private val editTextViewMapper: EditTextViewMapper,
-    private val checkedTextViewMapper: CheckedTextViewMapper,
-    private val decorViewMapper: DecorViewMapper,
-    private val checkBoxMapper: CheckBoxMapper,
-    private val radioButtonMapper: RadioButtonMapper,
-    private val switchCompatMapper: SwitchCompatMapper
+    textMapper: TextWireframeMapper,
+    buttonMapper: ButtonMapper,
+    editTextViewMapper: EditTextViewMapper,
+    checkedTextViewMapper: CheckedTextViewMapper,
+    decorViewMapper: DecorViewMapper,
+    checkBoxMapper: CheckBoxMapper,
+    radioButtonMapper: RadioButtonMapper,
+    switchCompatMapper: SwitchCompatMapper,
+    customMappers: Map<Class<*>, WireframeMapper<View, *>>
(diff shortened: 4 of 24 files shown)


library/dd-sdk-android-session-replay-material/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayMaterialConfiguration.kt
```diff
@@ -0,0 +1,26 @@
...
+package com.datadog.android.sessionreplay
+
+import android.view.View
+import com.datadog.android.sessionreplay.internal.recorder.mapper.WireframeMapper
+import com.google.android.material.slider.Slider
+
+/**
+ * An object describing the configuration of the Session Replay.
+ */
+class SessionReplayMaterialConfiguration {
```
mariusc83, line 16: I mean I could add here a new feature:
```
class SessionReplayMaterialFeature internal constructor(
    configuration: SessionReplayMaterialConfiguration,   
):SessionReplayFeature(configuration) 
```

mariusc83: And have the 
```class SessionReplayMaterialConfiguration internal constructor(val configuration:SessionReplayConfiguration) {
     Builder {
      ....
     }
}
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -53,10 +58,17 @@ internal constructor(
             return this
         }
 
+        protected fun setViewMapper(
+                type:Class<*>,
+                mapper: WireframeMapper<View,*>) {
+                this.customViewMappers[type] = mapper
+        }
```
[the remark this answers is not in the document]
mariusc83: yes I went this way first part even further `fun <T:View, S:Wireframe> setViewMapper(type:Class<T>, mapper:WireframeMapper<T,S>)`
but problem is that I have the map declared as:
` internal val customViewMappers: Map<Class<*>, WireframeMapper<View, *>> = emptyMap()`
and Kotlin as Java is stupid and doesn't know how to map `T` to `View` and `S` to `*` which in Kotlin means `? extends Wireframe`. So I will end up with force casting it but I guess if we go with the public api method I will better go this way. It will force the user to provide the right mapper for the right type at least.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/ExtensionSupport.kt
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * In case you need to provide different configuration for a specific Android UI framework that
+ * is not supported by our SR instrumentation layer (e.g Material elements)
+ * you can implement this class.
+ * @see [SessionReplayConfiguration.Builder.addExtensionSupport]
+ */
+interface ExtensionSupport {
+    /**
+     * Use this method if you want to apply a custom [WireframeMapper] for a specific [View] type.
+     * @return the [View] type to custom [WireframeMapper] as a [Map].
```
0xnm, line 21: we will need to explain what is `WireframeMapper` later in the docs

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -25,6 +26,18 @@ internal constructor(
...
         private var privacy = SessionReplayPrivacy.MASK_ALL
+        private var extensionSupport: ExtensionSupport = NoOpExtensionSupport()
+
+        /**
+         * Adds an extension support implementation. This is mostly used when you want to provide
+         * different behaviour of the Session Replay when using other Android UI frameworks
+         * than the default ones.
+         * @see [ExtensionSupport.getCustomViewMappers]
+         */
+        fun addExtensionSupport(extensionSupport: ExtensionSupport): Builder {
```
0xnm, line 37: `add` has a connotation of the multiple times usage, so I guess it should be backed by `List`. In the end it is a valid case imo when multiple extensions are added.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/ExtensionSupport.kt
```diff
@@ -0,0 +1,24 @@
...
+     */
+    fun getCustomViewMappers(): Map<Class<*>, WireframeMapper<View, *>>
```
0xnm, line 23: I'm missing some context, but why the signature of `WireframeMapper` here is like that, while at the declaration site it has some boundaries set `<T : View, S : MobileSegment.Wireframe>`?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/GlobalBounds.kt
```diff
@@ -6,7 +6,17 @@
 
 package com.datadog.android.sessionreplay.internal.recorder
 
-internal data class GlobalBounds(
+/**
+ * Defines the dimension and positions in Global coordinates for a geometry.
```
0xnm, line 10: imo we need to say what is the meaning of global here, like (0,0) is the corner of the screen, not app decor view, etc.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/GenericWireframeMapper.kt
```diff
@@ -28,10 +28,26 @@ internal abstract class GenericWireframeMapper(
...
     private val checkBoxMapper: CheckBoxMapper,
     private val radioButtonMapper: RadioButtonMapper,
-    private val switchCompatMapper: SwitchCompatMapper
+    private val switchCompatMapper: SwitchCompatMapper,
+    private val customMappers: Map<Class<*>, WireframeMapper<View, *>>
 ) : WireframeMapper<View, MobileSegment.Wireframe> {
 
     override fun map(view: View, systemInformation: SystemInformation):
+        List<MobileSegment.Wireframe> {
+        return applyCustomMappers(view, systemInformation)
```
0xnm, line 37: minor: wouldn't it be simpler to just have a list of custom + default mappers (in this order) and just return `firstNotNull` while iterating over it? but probably it doesn't bring a lot of added value.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapperTest.kt
```diff
@@ -96,6 +97,10 @@ internal abstract class BaseWireframeMapperTest {
         return textView.text.toString()
     }
 
+    protected fun Forge.aClass(): Class<*> {
+        return anElementFrom()
```
0xnm, line 101: it will crash if not overriden, right? maybe we just need to throw explicit exception here? because if it throws by calling this, it will be quite non-evident why it is like that.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/ExtensionSupport.kt
mariusc83: It is hard to explain so I will explain you at the desk :). In a nutshell it is because those boundaries are set by default and allows also wildcard. Basically in Kotlin `WireframeMapper<View,*>` relates to `WireframeMapper<View, T:Wireframe>`. The idea is that I prefer to cast my mappers here where I provide than later. So if I would provide a `Map<Class<T>, WireframeMapper<T,S>` from here later I would have to cast it to `Map<Class<*>, WireframeMapper<View, *>>` in order to use it in my `GenericMapper`. I'd rather provided it directly like that and do the casting in the provider method.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/GenericWireframeMapper.kt
mariusc83: yeah good idea...I am going to give it a try.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapperTest.kt
mariusc83: this is left over 🤦

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
mariusc83: yes but what I had in mind is to actually leverage this later ...and to back it by a list in case it is really needed. I can change that later without modifying the public api. What do you think ?

mariusc83: The idea is that for now I can only see this be using from an App or from an external SR module. In that sense I can only see it apply it from when the configuration is being made and only one `ExtensionSupport` implementation / app or module. Maybe later it will need more but I don't see the case for now.

0xnm: Ok, let's have a TODO + task then for that. For me `addXXX` allows multiple calls, while single call is for the `setXXX`.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/GenericWireframeMapper.kt
```diff
@@ -17,60 +17,95 @@ import android.widget.TextView
...
-                    !View::class.java.isAssignableFrom(viewParent::class.java)
-                ) {
-                    decorViewMapper.map(view, systemInformation)
-                } else {
-                    viewWireframeMapper.map(view, systemInformation)
-                }
-            }
+        return mappers.firstOrNull {
+            it.first(view)
+        }?.second?.map(view, systemInformation) ?: emptyList()
```
0xnm, line 92: I would suggest to use an explicit class instead of `Pair`, because here it is not that obvious what is `first` and `second`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/GenericWireframeMapper.kt
```diff
@@ -17,60 +17,95 @@ import android.widget.TextView
...
+        return {
+            type.isAssignableFrom(it::class.java)
+        }
+    }
+    private fun decorViewCheckCondition(): MapperCondition {
+        return {
+            val viewParent = it.parent
+            viewParent == null || !View::class.java.isAssignableFrom(viewParent::class.java)
         }
     }
```
0xnm, line 105: minor: seems like formatting can be simplified:

```suggestion
    private fun defaultTypeCheckCondition(type: Class<*>): MapperCondition {
        return type.isAssignableFrom(it::class.java)
    }
    
    private fun decorViewCheckCondition(): MapperCondition {
        val viewParent = it.parent
        return viewParent == null || !View::class.java.isAssignableFrom(viewParent::class.java)
    }
```

0xnm (approved):
lgtm, I've added few suggestions.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/GenericWireframeMapper.kt
mariusc83: actually for keeping it simpler I would rather keep the Pairs as you cannot do a mistake when applying them. If you use `second` instead of `first` it will not compile as it will not match the signature.

0xnm: I meant to create a data class with 2 named fields. But it is minor anyway, just for reviews, because in IDE there are type suggestions, unlike here.

xgouchet (approved):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/ExtensionSupport.kt
```diff
@@ -0,0 +1,25 @@
...
+ * you can implement this class.
+ * @see [SessionReplayConfiguration.Builder.addExtensionSupport]
+ */
+interface ExtensionSupport {
+    /**
+     * Use this method if you want to apply a custom [WireframeMapper] for a specific [View] type
+     * and for a specific [SessionReplayPrivacy].
+     * @return the [View] type to custom [WireframeMapper] map grouped by [SessionReplayPrivacy].
+     */
+    fun getCustomViewMappers(): Map<SessionReplayPrivacy, Map<Class<*>, WireframeMapper<View, *>>>
```
0xnm, line 24: side note: I'm curious if `Wireframe` mapper would have `SessionReplayPrivacy` as `map` method argument. Would it make architecture better or worse? wdyt?

0xnm (approved):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/ExtensionSupport.kt
mariusc83: oh...that will make it worse as i will introduce a new state in the `WireframeMapper`. At least this way I am benefiting from the strategy pattern.

PR #1329, opened 2023-03-07
Stabilize integration tests

fuzzybinary:
### What does this PR do?

This attempts to remove flakeyness from our integration tests by fixing 3 major issues:

1. The TelemetryTest was expecting a very specific set of telemetry in a very specific order. There is a new piece of telemetry that is getting sent on Activity launch that can be sent anywhere in that order. The new test checks that all expected telemetry is there, but ignores any extra telemetry.
2. Activity and Fragment tests could attempt to capture RUM context while other events were processing, giving them either an incorrect View ID during the capture or "null". Add a (private) method that waits for the RUM queue to process events before continuing to capture the RUM context
3. The ApplicationLaunchView does not stop until the `application_start` action is sent, which is highly dependent on how long it takes for it to get written to storage. Because of this, I've split out checking all the regular events from checking any events associated with tracking application launch.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
@@ -41,6 +41,7 @@ import com.datadog.android.v2.api.SdkCore
 import com.datadog.android.v2.core.internal.ContextProvider
 import com.datadog.android.v2.core.internal.storage.DataWriter
 import java.util.Locale
+import java.util.concurrent.CountDownLatch
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
 import java.util.concurrent.RejectedExecutionException
@@ -432,6 +433,25 @@ internal class DatadogRumMonitor(
                         e
                     )
                 }
+            }
+        }
+    }
+
+    /**
+     * Wait for any pending events. This is mostly for integration tests to ensure that the
+     * RUM context is in the correct state before proceeding.
+     */
+    @Suppress("unused")
+    private fun waitForPendingEvents() {
+        if (!executorService.isShutdown) {
+            val latch = CountDownLatch(1)
+            // Submit an empty task, and wait for it to complete
+            executorService.submit {
+                latch.countDown()
+            }
+            try {
+                latch.await()
+            } catch (_: InterruptedException) {
             }
         }
     }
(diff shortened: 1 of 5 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -436,6 +437,25 @@ internal class DatadogRumMonitor(
...
+    /**
+     * Wait for any pending events. This is mostly for integration tests to ensure that the
+     * RUM context is in the correct state before proceeding.
+     */
+    @Suppress("unused")
+    private fun waitForPendingEvents() {
+        if (!executorService.isShutdown) {
+            val latch = CountDownLatch(1)
+            // Submit an empty task, and wait for it to complete
+            executorService.submit {
```
0xnm, line 449: nice trick

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -436,6 +437,25 @@ internal class DatadogRumMonitor(
...
+                latch.countDown()
+            }
+            try {
+                latch.await()
```
xgouchet, line 453: Maybe we could add a timeout here in case something is blocking the executor indefinitely?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -436,6 +437,25 @@ internal class DatadogRumMonitor(
...
+            } catch (_: InterruptedException) {
+            }
```
xgouchet, line 455: Let's not swallow the exception here and at least send it to the Maintainers via internalLogger

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/telemetry/TelemetryTest.kt
```diff
@@ -61,16 +62,28 @@ internal open class TelemetryTest {
...
+        Assertions.assertThat(telemetryEvents.size)
             .overridingErrorMessage(
-                "Recorded telemetry has different number of events" +
-                    " (${telemetryEvents.size}) than expected" +
-                    " telemetry (${expectedTelemetry.size})"
+                "Recorded telemetry expected at least ${expectedTelemetry.size} events, " +
+                    " got (${telemetryEvents.size}) which is less than expected."
             )
-            .hasSameSizeAs(expectedTelemetry)
+            .isGreaterThanOrEqualTo(expectedTelemetry.size)
```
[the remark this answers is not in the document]
fuzzybinary: It's more that we sometimes can't control what will send telemetry during the integration test. Right now I only ever see 4, with the 4th being something from the Activity. But that doesn't mean other telemetry couldn't sneak its way in.

0xnm (approved):

xgouchet (approved):

PR #1331, opened 2023-03-08
RUMM-2984 global registry

xgouchet:
### What does this PR do?

Implement the registry to allow multiple SDK initialization

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -21,54 +21,67 @@ import com.datadog.android.v2.api.context.UserInfo
 import com.datadog.android.v2.core.DatadogCore
 import com.datadog.android.v2.core.NoOpSdkCore
 import com.datadog.android.v2.core.internal.HashGenerator
+import com.datadog.android.v2.core.internal.SdkCoreRegistry
 import com.datadog.android.v2.core.internal.Sha256HashGenerator
 import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * This class initializes the Datadog SDK, and sets up communication with the server.
  */
+@Suppress("DEPRECATION") // TODO RUMM-3103 remove deprecated references
 @SuppressWarnings("TooManyFunctions")
 object Datadog {
 
-    // TODO RUMM-2984 Implement global registry
+    internal val registry = SdkCoreRegistry(internalLogger)
+
     /**
      * Temporary thing, until global registry is implemented.
      */
+    @Deprecated("Will be removed in RUMM-3103")
     var globalSdkCore: SdkCore = NoOpSdkCore()
         internal set
 
     internal var hashGenerator: HashGenerator = Sha256HashGenerator()
 
+    @Deprecated("Will be removed in RUMM-3103")
     internal val initialized = AtomicBoolean(false)
 
     // region Initialization
 
     /**
-     * Initializes the Datadog SDK.
+     * Initializes a named instance of the Datadog SDK.
+     * @param instanceName the name of the instance (or null to initialize the default instance
      * @param context your application context
      * @param credentials your organization credentials
      * @param configuration the configuration for the SDK library
-     * @param trackingConsent as the initial state of the tracking consent flag.
+     * @param trackingConsent as the initial state of the tracking consent flag
+     * @return the initialized SDK instance, or null if something prevents the SDK from
+     * being initialized
      * @see [Credentials]
      * @see [Configuration]
      * @see [TrackingConsent]
      * @throws IllegalArgumentException if the env name is using illegal characters and your
      * application is in debug mode otherwise returns false and stops initializing the SDK
      */
+    @Suppress("ReturnCount")
     @JvmStatic
     fun initialize(
+        instanceName: String?,
         context: Context,
         credentials: Credentials,
         configuration: Configuration,
         trackingConsent: TrackingConsent
-    ) {
-        if (initialized.get()) {
-            internalLogger.log(
-                InternalLogger.Level.WARN,
-                InternalLogger.Target.USER,
-                MESSAGE_ALREADY_INITIALIZED
-            )
-            return
+    ): SdkCore? {
+        synchronized(registry) {
+            val existing = registry.getInstance(instanceName)
+            if (existing != null) {
+                internalLogger.log(
+                    InternalLogger.Level.WARN,
+                    InternalLogger.Target.USER,
+                    MESSAGE_ALREADY_INITIALIZED
+                )
+                return existing
+            }
         }
 
         val sdkInstanceId = hashGenerator.generate(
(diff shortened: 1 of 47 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -81,24 +94,64 @@ object Datadog {
...
     }
 
     /**
-     * Checks if the Datadog SDK was already initialized.
-     * @return true if the SDK was initialized, false otherwise
+     * Checks if a default instance of the Datadog SDK was already initialized.
+     * @return true if a default instance of the SDK was initialized, false otherwise
      */
     @JvmStatic
     fun isInitialized(): Boolean {
```
fuzzybinary, line 140: Should this be deprecated as well?

buraizu (approved):
A couple of minor edits, but otherwise LGTM

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -21,54 +21,67 @@ import com.datadog.android.v2.api.context.UserInfo
...
 
+    @Deprecated("Will be removed in RUMM-3103")
     internal val initialized = AtomicBoolean(false)
 
     // region Initialization
 
     /**
-     * Initializes the Datadog SDK.
+     * Initializes a named instance of the Datadog SDK.
+     * @param instanceName the name of the instance (or null to initialize the default instance
```
0xnm, line 53:
```suggestion
     * @param instanceName the name of the instance (or null to initialize the default instance)
```

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -21,54 +21,67 @@ import com.datadog.android.v2.api.context.UserInfo
...
      * @param configuration the configuration for the SDK library
-     * @param trackingConsent as the initial state of the tracking consent flag.
+     * @param trackingConsent as the initial state of the tracking consent flag
+     * @return the initialized SDK instance, or null if something prevents the SDK from
+     * being initialized
      * @see [Credentials]
      * @see [Configuration]
      * @see [TrackingConsent]
      * @throws IllegalArgumentException if the env name is using illegal characters and your
      * application is in debug mode otherwise returns false and stops initializing the SDK
```
0xnm, line 64: now `null` is returned

```suggestion
     * application is in debug mode otherwise returns null and stops initializing the SDK
```

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -81,24 +94,64 @@ object Datadog {
...
     }
 
+    /**
+     * Retrieve the initialized SDK instance attached to the given name,
+     * or the default instance if the name is null.
+     * @param instanceName the name of the instance to retrieve,
+     * or null to get the default instance
+     * @return the existing instance linked with the given name, or null
+     */
+    fun getInstance(instanceName: String? = null): SdkCore? {
```
0xnm, line 151:
```suggestion
    @JvmStatic
    fun getInstance(instanceName: String? = null): SdkCore? {
```

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -81,24 +94,64 @@ object Datadog {
...
+     * @param credentials your organization credentials
+     * @param configuration the configuration for the SDK library
+     * @param trackingConsent as the initial state of the tracking consent flag
+     * @return the initialized SDK instance, or null if something prevents the SDK from
+     * being initialized
+     * @see [Credentials]
+     * @see [Configuration]
+     * @see [TrackingConsent]
+     * @throws IllegalArgumentException if the env name is using illegal characters and your
+     * application is in debug mode otherwise returns false and stops initializing the SDK
```
0xnm, line 123:
```suggestion
     * application is in debug mode otherwise returns null and stops initializing the SDK
```

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -81,24 +94,64 @@ object Datadog {
...
+     */
+    @JvmStatic
+    fun initialize(
+        context: Context,
+        credentials: Credentials,
+        configuration: Configuration,
+        trackingConsent: TrackingConsent
+    ): SdkCore? {
+        return initialize(null, context, credentials, configuration, trackingConsent)
```
0xnm, line 132: I think we should give a name to the default instance. In the end it will be useful for internal logging, for example, where logs can come from multiple instances and it is need to distinguish them.

Upd: Ah, I see the default one is in the `SdkCoreRegistry`.

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -81,24 +94,64 @@ object Datadog {
...
             )
-            return
+            return null
         }
 
-        globalSdkCore = DatadogCore(context, credentials, configuration, sdkInstanceId)
-        globalSdkCore.setTrackingConsent(trackingConsent)
+        val sdkCore = DatadogCore(context, credentials, configuration, sdkInstanceId)
+        sdkCore.setTrackingConsent(trackingConsent)
+        registry.register(instanceName, sdkCore)
```
0xnm, line 102: should we make a double-check (synchronise it and do `getInstance` check again) here to protect from concurrency issues? It is possible that 2 threads are calling `initialize` method at the same time with same instance name, both go through initial `getInstance` check and come to this point, creating 2 `DatadogCore` instances, but only one of them will be in registry in the end (but both will be returned). wdyt?

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -81,24 +94,64 @@ object Datadog {
...
 
-        initialized.set(true)
```
0xnm, line 90: we still need to set it, there is some code relying on it.

dd-sdk-android/src/test/kotlin/com/datadog/android/DatadogTest.kt
```diff
@@ -99,8 +94,291 @@ internal class DatadogTest {
...
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.USER,
+            Datadog.CANNOT_CREATE_SDK_INSTANCE_ID_ERROR
+        )
+        assertThat(instance).isNull()
+    }
+
+    // endregion
+
+    /*
```
0xnm, line 380: should this be commented? methods below are there, just deprecated.

0xnm (changes requested):
nice! lgtm to me overall, I left some questions/suggestions.

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -21,54 +21,67 @@ import com.datadog.android.v2.api.context.UserInfo
...
      * @see [Credentials]
      * @see [Configuration]
      * @see [TrackingConsent]
      * @throws IllegalArgumentException if the env name is using illegal characters and your
-     * application is in debug mode otherwise returns false and stops initializing the SDK
+     * application is in debug mode otherwise returns null and stops initializing the SDK
      */
+    @Suppress("ReturnCount")
     @JvmStatic
     fun initialize(
```
0xnm, line 68: should we have just one `initialize` method with default value for `instanceName` and `JvmOverloads` annotation? Basically the same way `getInstance` is implemented for the consistency.

0xnm (approved):
lgtm, I've added one suggestion.

PR #1337, opened 2023-03-14
REPLAY-1334 SR - add capability to record Material Slider

mariusc83:
### What does this PR do?

In this PR we are introducing a new module `dd-sdk-android-session-replay-material` module from which we are providing the `MaterialExtensionSupport`. This new extension support implementation will provide the required `WireframeMappers` for material `Slider` component.

### Motivation

### Additional Notes


library/dd-sdk-android-session-replay-material/src/test/kotlin/com/datadog/android/sessionreplay/material/forge/WireframeForgeryFactory.kt
```diff
@@ -0,0 +1,20 @@
...
+
+package com.datadog.android.sessionreplay.forge
+
+import com.datadog.android.sessionreplay.model.MobileSegment
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+internal class WireframeForgeryFactory : ForgeryFactory<MobileSegment.Wireframe> {
+    override fun getForgery(forge: Forge): MobileSegment.Wireframe {
+        return when (forge.anInt(min = 0, max = 2)) {
```
0xnm, line 15: minor: can be simplified just to `forge.aBool`

library/dd-sdk-android-session-replay/build.gradle.kts
```diff
@@ -72,7 +72,7 @@ android {
     }
 }
 dependencies {
-    implementation(project(":dd-sdk-android"))
```
0xnm, line 75: side note: this is something to be reviewed later, but I think we should use `implementation` for our libraries when referencing the core module. This would allow to be more explicit and always require customer to add core module in their setup. But we will discuss it later.

sample/kotlin/build.gradle.kts
```diff
@@ -112,7 +112,7 @@ android {
 dependencies {
 
     implementation(project(":dd-sdk-android"))
-    implementation(project(":library:dd-sdk-android-session-replay"))
+    implementation(project(":library:dd-sdk-android-session-replay-material"))
```
0xnm, line 115: shouldn't we have both? it will make more clear which is the core and which one is just the extension.

library/dd-sdk-android-session-replay-material/src/main/kotlin/com/datadog/android/sessionreplay/material/SliderWireframeMapper.kt
```diff
@@ -0,0 +1,153 @@
...
+        val trackYPos = viewGlobalBounds.y + trackTopPadding +
+            (viewGlobalBounds.height - trackHeight) / 2
+
+        // thumb dimensions
+        val thumbHeight = (view.thumbRadius.toLong().densityNormalized(screenDensity)) * 2
+
+        // thumb positions
+        val thumbXPos = (trackXPos + trackWidth * normalizedSliderValue).toLong()
+        val thumbYPos = viewGlobalBounds.y + trackTopPadding +
+            (viewGlobalBounds.height - thumbHeight) / 2
```
0xnm, line 80: will it work for vertical slider as well?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/utils/UniqueIdentifierGenerator.kt
```diff
@@ -4,16 +4,32 @@
...
 
     internal const val DATADOG_UNIQUE_IDENTIFIER_KEY_PREFIX = "DATADOG_UNIQUE_IDENTIFIER_"
     private val secureRandom = SecureRandom()
 
+    /**
+     * Generates a persistent unique identifier for a virtual child view based on its unique
+     * name and its physical parent. The identifier will only be created once and persisted in
+     * the parent [View] tag to provide consistency. In case there was already a value with the
+     * same key in the tags and this was used by a different party we will try to use this value
+     * as identifier if it's a Long, in other case we will return null. This last scenario is
```
0xnm, line 26:
```suggestion
     * as identifier if it's a [Long], in other case we will return null. This last scenario is
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewUtilsInternal.kt
```diff
@@ -12,7 +12,7 @@ import android.view.ViewStub
 import androidx.appcompat.widget.ActionBarContextView
 import androidx.appcompat.widget.Toolbar
 
-internal class ViewUtils {
+internal class ViewUtilsInternal {
```
0xnm, line 15: are we splitting into 2 classes just because one has to stay simple `class` and another `object class`? I guess making this class `public` with some public members makes no sense, because then we have to use `object class` to follow the common convention for utility classes?

xgouchet (approved):

kayayarai:
The docs page in this is kinda empty. Is there a plan to fill it?

kayayarai (commented):
No docs review required at this time.

library/dd-sdk-android-session-replay-material/build.gradle.kts
```diff
@@ -0,0 +1,90 @@
...
+        abortOnError = true
+        checkReleaseBuilds = false
+        checkGeneratedSources = true
+        ignoreTestSources = true
+    }
+}
+dependencies {
+    api(project(":library:dd-sdk-android-session-replay"))
+    implementation(libs.kotlin)
+    implementation(libs.gson)
```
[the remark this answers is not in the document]
mariusc83: good question...I need to check with local maven to see.

library/dd-sdk-android-session-replay/build.gradle.kts
mariusc83: yeah need to test it with local maven to see how it behaves with or without

library/dd-sdk-android-session-replay-material/src/main/kotlin/com/datadog/android/sessionreplay/material/SliderWireframeMapper.kt
mariusc83: good point...and that's a SeekBar for which I will have another PR.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewUtilsInternal.kt
mariusc83: nope...is just because one is meant for internal usage only as those method will not be needed externally and it also has a state due to the `Build.Version` check.

library/dd-sdk-android-session-replay-material/src/test/kotlin/com/datadog/android/sessionreplay/material/forge/WireframeForgeryFactory.kt
mariusc83: I foresee more types later so I would rather keep it this way.

PR #1338, opened 2023-03-14
RUMM-3009: Leave useSite configuration only in the core

0xnm:
### What does this PR do?

Before this change we had `useSite` configuration options for every feature, but it makes setup more verbose and unlikely it can be the case when different features are using different sites (especially considering that `clientToken` is site-specific and is defined only in the core configuration).

To address that we are bringing back `DatadogContext#site` property and request factories will use it unless custom URL is given.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/NoOpContextProvider.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/NoOpContextProvider.kt
@@ -6,6 +6,7 @@
 
 package com.datadog.android.v2.core.internal
 
+import com.datadog.android.DatadogSite
 import com.datadog.android.privacy.TrackingConsent
 import com.datadog.android.v2.api.context.DatadogContext
 import com.datadog.android.v2.api.context.DeviceInfo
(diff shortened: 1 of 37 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/NoOpContextProvider.kt
```diff
@@ -19,6 +20,7 @@ internal class NoOpContextProvider : ContextProvider {
     // TODO RUMM-0000 this one is quite ugly. Should return type be nullable?
     override val context: DatadogContext
```
mariusc83, line 21: should we unit test this class ?

mariusc83 (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/NoOpContextProvider.kt
0xnm: we have `DatadogContextProviderTest`

xgouchet (approved):

PR #1339, opened 2023-03-15
RUMM-3103 remove global sdk core

xgouchet:
### What does this PR do?

Remove all references to the globalSdkCore and general static fields of the Datadog class.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -145,11 +145,6 @@ object Datadog {
 
     // region Global methods
 
-    // Stop all Datadog work (for test purposes).
-    internal fun stop() {
-        stopInstance(null)
-    }
-
     // Executes all the pending queues in the upload/persistence executors.
     // Tries to send all the granted data for each feature and then clears the folders and shuts
     // down the persistence and the upload executors.
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -14,34 +14,20 @@ import com.datadog.android.privacy.TrackingConsent
 import com.datadog.android.v2.api.Feature
 import com.datadog.android.v2.api.InternalLogger
 import com.datadog.android.v2.api.SdkCore
-import com.datadog.android.v2.api.context.UserInfo
 import com.datadog.android.v2.core.DatadogCore
-import com.datadog.android.v2.core.NoOpSdkCore
 import com.datadog.android.v2.core.internal.HashGenerator
 import com.datadog.android.v2.core.internal.SdkCoreRegistry
 import com.datadog.android.v2.core.internal.Sha256HashGenerator
-import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * This class initializes the Datadog SDK, and sets up communication with the server.
  */
-@Suppress("DEPRECATION") // TODO RUMM-3103 remove deprecated references
 @SuppressWarnings("TooManyFunctions")
 object Datadog {
 
     internal val registry = SdkCoreRegistry(internalLogger)
 
-    /**
-     * Temporary thing, until global registry is implemented.
-     */
-    @Deprecated("Will be removed in RUMM-3103")
-    var globalSdkCore: SdkCore = NoOpSdkCore()
-        internal set
-
     internal var hashGenerator: HashGenerator = Sha256HashGenerator()
-
-    @Deprecated("Will be removed in RUMM-3103")
-    internal val initialized = AtomicBoolean(false)
 
     // region Initialization
 
@@ -79,31 +65,26 @@ object Datadog {
                 )
                 return existing
             }
-        }
 
-        val sdkInstanceId = hashGenerator.generate(
-            "${credentials.clientToken}${configuration.coreConfig.site.siteName}"
-        )
-
-        if (sdkInstanceId == null) {
-            internalLogger.log(
-                InternalLogger.Level.ERROR,
-                InternalLogger.Target.USER,
-                CANNOT_CREATE_SDK_INSTANCE_ID_ERROR
+            val sdkInstanceId = hashGenerator.generate(
+                "${credentials.clientToken}${configuration.coreConfig.site.siteName}"
             )
-            return null
-        }
 
-        val sdkCore = DatadogCore(context, credentials, configuration, sdkInstanceId)
-        sdkCore.setTrackingConsent(trackingConsent)
-        registry.register(instanceName, sdkCore)
+            if (sdkInstanceId == null) {
+                internalLogger.log(
+                    InternalLogger.Level.ERROR,
+                    InternalLogger.Target.USER,
+                    CANNOT_CREATE_SDK_INSTANCE_ID_ERROR
+                )
+                return null
+            }
 
-        // TODO RUMM-3103 remove this
-        if (instanceName == null) {
-            globalSdkCore = sdkCore
-            initialized.set(true)
+            val sdkCore = DatadogCore(context, credentials, configuration, sdkInstanceId)
+            sdkCore.setTrackingConsent(trackingConsent)
+            registry.register(instanceName, sdkCore)
+
+            return sdkCore
         }
-        return sdkCore
     }
 
     /**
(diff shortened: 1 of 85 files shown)


urseberry (approved):

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/ResourceTrackingTest.kt
```diff
@@ -92,6 +93,7 @@ internal class ResourceTrackingTest {
     }
 
     @Test
+    @Ignore
```
0xnm, line 96: should this be reverted?

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/rules/MockServerActivityTestRule.kt
```diff
@@ -81,8 +80,10 @@ internal open class MockServerActivityTestRule<T : Activity>(
 
                     return if (request.path == CONNECTION_ISSUE_PATH) {
                         MockResponse().setSocketPolicy(SocketPolicy.NO_RESPONSE)
-                    } else {
+                    } else if (request.path.endsWith("nyan-cat.gif")) {
```
0xnm, line 83: seems leftover from debugging?

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/log/ConsentPendingGrantedLogsTest.kt
```diff
@@ -33,7 +32,7 @@ internal class ConsentPendingGrantedLogsTest : LogsTest() {
     @Test
     fun verifyActivityLogs() {
         // update the tracking consent
-        Datadog.setTrackingConsent(TrackingConsent.GRANTED)
+        Datadog.getInstance()?.setTrackingConsent(TrackingConsent.GRANTED)
```
0xnm, line 35: I'm wondering if we can also have some non-null variant of getting the instance, because for example in nightly-tests we have to use `!!`, which is not very nice.

To me it is the same situation as getting the `Activity` in Android, where `getActivity` and `requireActivity` exist (which looks to me as a hacky workaround).

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/logs/LoggerE2ETests.kt
```diff
@@ -62,14 +62,12 @@ class LoggerE2ETests {
...
      */
     @Before
     fun setUp() {
-        initializeSdk(
-            InstrumentationRegistry.getInstrumentation().targetContext,
-            forgeSeed = forge.seed
-        )
-        logger = Logger.Builder(Datadog.globalSdkCore)
+        initializeSdk(InstrumentationRegistry.getInstrumentation().targetContext)
+        logger = Logger.Builder(Datadog.getInstance()!!)
```
0xnm, line 70: Passing the seed it missing and we can use `sdkCore` return value.

```suggestion
        val sdkCore = initializeSdk(InstrumentationRegistry.getInstrumentation().targetContext, forgeSeed = forge.seed)
        checkNotNull(sdkCore)
        logger = Logger.Builder(sdkCore)
```

library/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/AndroidTracer.kt
```diff
@@ -226,6 +227,7 @@ class AndroidTracer internal constructor(
             val propagationStyles = tracingHeaderTypes.joinToString(",")
             properties.setProperty(Config.PROPAGATION_STYLE_EXTRACT, propagationStyles)
             properties.setProperty(Config.PROPAGATION_STYLE_INJECT, propagationStyles)
+            properties.setProperty(Config.PRIORITY_SAMPLING, "false")
```
0xnm, line 230: why this change is needed?

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -149,36 +121,33 @@ object Datadog {
...
+            instance?.stop()
+        }
     }
 
+    // endregion
+
+    // region Global methods
+
     // Stop all Datadog work (for test purposes).
     internal fun stop() {
```
0xnm, line 149: seems like this method can be removed, given that we can call `stopInstance` directly from public API.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/rules/MockServerActivityTestRule.kt
```diff
@@ -141,13 +145,13 @@ internal open class MockServerActivityTestRule<T : Activity>(
     // region MockServerRule
 
     fun getRequests(): List<HandledRequest> {
-        Log.i(TAG, "Caught ${requests.size} requests")
+//        Log.i(TAG, "Caught ${requests.size} requests")
```
0xnm, line 148: should be reverted here and below it seems

0xnm (commented):
nice work! I added few comments.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadWorker.kt
```diff
@@ -10,21 +10,18 @@ import android.content.Context
...
-import com.datadog.android.v2.core.DatadogCore
+import com.datadog.android.v2.core.InternalSdkCore
 import com.datadog.android.v2.core.internal.net.DataUploader
 import java.util.LinkedList
 import java.util.Queue
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 
-@Suppress("DEPRECATION") // TODO RUMM-3103 remove deprecated references
 internal class UploadWorker(
```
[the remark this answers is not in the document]
xgouchet: Oh indeed that's an issue :/

dd-sdk-android/src/main/kotlin/com/datadog/android/error/internal/DatadogExceptionHandler.kt
```diff
@@ -73,9 +71,8 @@ internal class DatadogExceptionHandler(
         // TODO RUMM-0000 If DatadogExceptionHandler goes into dedicated module (module of 1 class
         //  only?), we have to wait for the write in some other way
         // give some time to the persistence executor service to finish its tasks
-        val coreFeature = (Datadog.globalSdkCore as? DatadogCore)?.coreFeature
-        if (coreFeature != null) {
-            val idled = (coreFeature.persistenceExecutorService as? ThreadPoolExecutor)
+        if (sdkCore is InternalSdkCore) {
```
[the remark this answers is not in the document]
0xnm: Let's keep it like that then, I'm not quite a fan of pulling this internal thing to the public `SdkCore`.

Regarding throwing an exception: my assumption is that if it is thrown, it will always be caught by our integration tests before leaking to the release, because it is us only who are creating the cores.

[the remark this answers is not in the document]
0xnm: My concern is just to avoid places in the SDK where something gets unnoticed if it is not as what we expect, this can lead to the hard debugging, especially if it happens on the client side.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/log/ConsentPendingGrantedLogsTest.kt
0xnm: Yes, I'm not a fan of such API either. I'm just curious how we can reduce the number of nullable states in our codebase.

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadWorker.kt
```diff
@@ -33,38 +32,34 @@ internal class UploadWorker(
...
     override fun doWork(): Result {
-        if (!Datadog.isInitialized()) {
+        // the idea behind upload is the following:
+        // 1. we shuffle features list to randomize initial upload task sequence. It is done to
+        // avoid the possible bottleneck when some feature has big batches which are uploaded
+        // slowly, so that next time other features don't wait and have a chance to go before.
+        // 2. we introduce FIFO queue also to avoid the bottleneck: if some feature batch cannot
+        // be uploaded we put retry task to the end of queue, so that batches of other features
+        // have a chance to go.
+        val sdkCore = Datadog.getInstance() as? InternalSdkCore
```
0xnm, line 42: When we create `uploadWorkRequest` we can add the name of the instance it is created for and then read it here and get the instance needed.

0xnm (approved):
lgtm! I've added a comment regarding `UploadWorker`, so that it is not associated with default instance only.

PR #1347, opened 2023-03-20
RUMM-3105 add SdkCore dependency in Rum Monitor builder

xgouchet:
### What does this PR do?

Update the `RumMonitor.Builder` constructor to have an `SdkCore` instance, and avoid refering to the default instance


library/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -348,7 +349,7 @@ interface RumMonitor {
             }
         }
 
-        internal companion object {
```
0xnm, line 351: seems like redundant change, all the member are `internal` anyway

library/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -42,7 +42,8 @@ internal class RumViewManagerScope(
     @WorkerThread
     override fun handleEvent(event: RumRawEvent, writer: DataWriter<Any>): RumScope {
         if (!applicationDisplayed) {
-            val isForegroundProcess = sdkCore.processImportance ==
+            val processImportance = sdkCore.getDatadogContext()?.processInfo?.processImportance
```
0xnm, line 45: I was actually thinking to remove process info from `DatadogContext`, because it seems it is something very specific and needed only in 1 place. Also if we are moving `processImportance` to the instance variable instead of global, can we have any issues popping up? Say we have 2 SDK instances, each got different values, depending on the time it was initialized. wdyt?

xgouchet: Indeed, another option would be to keep this static but not in the CoreFeature, as it's only needed for RUM. What about having a ContentProvider in RUM that is initialized with the app's process, and reads the info on startup once, and make it available for everyone?

library/dd-sdk-android-rum/apiSurface
```diff
@@ -99,9 +107,11 @@ interface com.datadog.android.rum.RumMonitor
   fun addFeatureFlagEvaluation(String, Any)
   fun _getInternal(): _RumInternalProxy?
   class Builder
+    constructor(com.datadog.android.v2.api.SdkCore)
     fun sampleRumSessions(Float): Builder
     fun setSessionListener(RumSessionListener): Builder
     fun build(): RumMonitor
+    companion object
```
0xnm, line 114: we can make `companion` back `internal`

library/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/DdRumContentProvider.kt
```diff
@@ -0,0 +1,70 @@
...
+    override fun update(
+        uri: Uri,
+        values: ContentValues?,
+        selection: String?,
+        selectionArgs: Array<out String>?
+    ): Int {
+        return 0
+    }
+
+    companion object {
```
0xnm, line 65: minor: can be `internal`

library/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScopeTest.kt
```diff
@@ -634,24 +633,20 @@ internal class RumViewManagerScopeTest {
...
-        val appStartTimestamp = TimeUnit.NANOSECONDS.toMillis(
-            (
-                TimeUnit.MILLISECONDS.toNanos(fakeEvent.eventTime.timestamp) -
-                    fakeEvent.eventTime.nanoTime
-                ) +
-                appStartTimeNs
-        )
+        val eventTimeNanos = TimeUnit.MILLISECONDS.toNanos(fakeEvent.eventTime.timestamp)
+        val timestampNs = (eventTimeNanos - fakeEvent.eventTime.nanoTime) + appStartTimeNs
+        val timestampMs = TimeUnit.NANOSECONDS.toMillis((timestampNs))
```
0xnm, line 644: format: redundant brackets

```suggestion
        val timestampMs = TimeUnit.NANOSECONDS.toMillis(timestampNs)
```

0xnm (approved):
very nice work! lgtm!

PR #1356, opened 2023-03-27
RUMM-2873 Allow users to stop a RUM session

fuzzybinary:
### What does this PR do?

This adds a method to manually end a RUM session with stopRumSession. This immediately stops the active view and sets the current session to inactive. A new session starts automatically when a user action is sent (startView or addUserAction).

RUM events now have a new property: session.isActive. This can be null, and is only currently set in RUMViewEvents. The state of this member is pulled from the parent session context.

I've modified RumApplicationScope to support holding multiple RumSessionScopes in a similar way to how RumViewManager can handle multiple Views until all of their events have been processed.

When a RUMStopSession event is received, the Application Scope forwards it to the Session, which marks itself as inactive, then forwards the event to the all views, which mark themselves as inactive and send a ViewUpdate event which captures the session now being inactive.

Provided views have finished processing all of their user actions and resources, the session is removed from available sessions on the ApplicationScope. Otherwise, it sicks around (marked as inactive) until all events are complete.


instrumented/integration/src/main/res/layout/kiosk_splash_layout.xml
```diff
@@ -0,0 +1,34 @@
...
+        app:layout_constraintTop_toBottomOf="@+id/end_session" />
+
+    <Button
+        android:id="@+id/end_session"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="140dp"
+        android:text="@string/end_session"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintHorizontal_bias="0.5"
```
0xnm, line 31: minor: probably we don't need `ConstraintLayout` with bias `0.5`/`0.498` (and `0.498`/`0.501` for another layout). Good old `LinearLayout` with gravity should work.

instrumented/integration/src/main/kotlin/com/datadog/android/sdk/integration/rum/KioskTrackedPlaygroundActivity.kt
```diff
@@ -0,0 +1,29 @@
...
+    lateinit var button: Button
+    lateinit var backButton: Button
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.kiosk_tracked_layout)
+
+        button = findViewById(R.id.kiosk_button)
+        backButton = findViewById(R.id.kiosk_back_button)
+        backButton.setOnClickListener {
```
0xnm, line 25: can be a bit shorter

```suggestion

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.kiosk_tracked_layout)
        
        val backButton = findViewById(R.id.kiosk_back_button)
        backButton.setOnClickListener {
```

instrumented/integration/src/main/kotlin/com/datadog/android/sdk/integration/rum/KioskSplashPlaygroundActivity.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.datadog.android.rum.GlobalRum
+import com.datadog.android.rum.RumMonitor
+import com.datadog.android.rum.tracking.ActivityViewTrackingStrategy
+import com.datadog.android.sdk.integration.R
+import com.datadog.android.sdk.integration.RuntimeConfig
+
+internal class KioskSplashPlaygroundActivity : AppCompatActivity() {
+
+    lateinit var endSessionButton: Button
+    lateinit var startKioskButton: Button
```
0xnm, line 24: no need to have properties, can be just variables inside method.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -742,7 +751,8 @@ internal open class RumViewScope(
                     session = ViewEvent.ViewEventSession(
                         id = rumContext.sessionId,
                         type = ViewEvent.ViewEventSessionType.USER,
-                        hasReplay = hasReplay
+                        hasReplay = hasReplay,
+                        isActive = rumContext.isSessionActive
```
0xnm, line 755: I believe we are hitting the problem here by relying on the `rumContext`. Imagine the following situation: session was stopped, but view needs to wait for the child events to complete. But before they complete, another session was started, meaning by the time this view will receive terminal view update, it will be another session in the `RumContext` (`active=true` with another session ID), different from the one which view was attributed to originally.

I think the only solution in this case is to take session ID at the time view is constructed and rely on it, instead of relying on the shared `RumContext`.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/RumContext.kt
```diff
@@ -13,6 +13,7 @@ import java.util.UUID
 internal data class RumContext(
     val applicationId: String = NULL_UUID,
     val sessionId: String = NULL_UUID,
+    val isSessionActive: Boolean = true,
```
0xnm, line 16: I believe it should be `false` by default, to have a consistency with default `sessionId` which is `NULL_UUID`, meaning there is no session.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/RumTestUtils.kt
```diff
@@ -123,6 +123,12 @@ private fun JsonObject.verifyEventMatches(event: ExpectedViewEvent) {
     if (viewArguments.isNotEmpty()) {
         assertThat(this.getAsJsonObject(CONTEXT_KEY)).containsAttributes(viewArguments)
     }
+    if (event.sessionIsActive != null) {
```
0xnm, line 126: shouldn't they always have this field?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -65,4 +87,52 @@ internal class RumApplicationScope(
...
+                    val startViewEvent = RumRawEvent.StartView(
+                        key = it.keyRef.get()!!,
+                        name = it.name,
+                        attributes = it.attributes
+                    )
+                    newSession.handleEvent(startViewEvent, writer)
+                }
+            }
+        }
+        lastActiveViewScope = null
```
0xnm, line 136: this looks like a hack, can you please elaborate the purpose of this? I have concerns about mutable and public property `lastActiveViewScope`, looks like it is leaking encapsulation and the nesting concept.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -37,13 +37,15 @@ internal class RumSessionScope(
...
     contextProvider: ContextProvider,
+    applicationDisplayed: Boolean,
     appStartTimeProvider: AppStartTimeProvider = DefaultAppStartTimeProvider(),
     private val sessionInactivityNanos: Long = DEFAULT_SESSION_INACTIVITY_NS,
     private val sessionMaxDurationNanos: Long = DEFAULT_SESSION_MAX_DURATION_NS
 ) : RumScope {
 
     internal var sessionId = RumContext.NULL_UUID
     internal var sessionState: State = State.NOT_TRACKED
+    internal var isActive: Boolean = true
```
0xnm, line 48: should be `false` by default to be aligned with `RumContext.NULL_UUID` for `sessionId` imo.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -254,6 +254,14 @@ interface RumMonitor {
         value: Any
     )
 
+    /**
+     * Stops the current session.
+     * A new session will start in response to a call to `startView` or `addUserAction`.
```
0xnm, line 259: `startUserAction` as well according to the code.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScopeTest.kt
```diff
@@ -121,16 +150,196 @@ internal class RumApplicationScopeTest {
...
+        testedScope.childScopes.add(mockChildScope)
 
         testedScope.handleEvent(mockEvent, mockWriter)
 
         verify(mockChildScope).handleEvent(mockEvent, mockWriter)
         verifyZeroInteractions(mockWriter)
     }
+
+    @Test
+    fun `M have return active session W activeSession`() {
```
0xnm, line 168:
```suggestion
    fun `M return active session W activeSession`() {
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScopeTest.kt
```diff
@@ -121,16 +150,196 @@ internal class RumApplicationScopeTest {
...
+        // Then
+        val activeSession = testedScope.activeSession
+        check(activeSession is RumSessionScope)
```
0xnm, line 171:
```suggestion
        assertThat(activeSession).isInstanceOf(RumSessionScope::class.java)
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScopeTest.kt
```diff
@@ -121,16 +150,196 @@ internal class RumApplicationScopeTest {
...
+        val newSession = testedScope.childScopes.first()
+        check(newSession is RumSessionScope)
+        assertThat(newSession).isNotSameAs(initialSession)
+        assertThat(newSession.samplingRate).isEqualTo(fakeSamplingRate)
+        assertThat(newSession.backgroundTrackingEnabled).isEqualTo(fakeBackgroundTrackingEnabled)
+        assertThat(newSession.firstPartyHostHeaderTypeResolver).isSameAs(mockResolver)
+    }
+
+    @Test
+    fun `M create a new session with last known view W handleEvent { no active sessions, start action } `(
```
0xnm, line 229: I'm not quite clear why do we create a new session with last known view. Can you please elaborate?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -213,6 +206,84 @@ internal class RumSessionScopeTest {
 
     // endregion
 
+    // region Stopping Sessions
+
+    @Test
+    fun `M set session active to false W hanldeEvent { StopSession }`() {
```
0xnm, line 212:
```suggestion
    fun `M set session active to false W handleEvent { StopSession }`() {
```

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -213,6 +206,84 @@ internal class RumSessionScopeTest {
...
+//            verify(mockSdkCore, times(2)).updateFeatureContext(eq(RumFeature.RUM_FEATURE_NAME), capture())
+//
+//            val rumContext = mutableMapOf<String, Any?>()
+//            lastValue.invoke(rumContext)
+//
+//            assertThat(rumContext["application_id"]).isEqualTo(initialContext.applicationId)
+//            assertThat(rumContext["session_id"]).isEqualTo(RumContext.NULL_UUID)
+//            assertThat(rumContext["session_state"]).isEqualTo(RumSessionScope.State.NOT_TRACKED)
+//        }
+//    }
```
0xnm, line 283: should be uncommented?

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7090,6 +7090,34 @@ internal class RumViewScopeTest {
...
+        )
+
+        // Then
+        assertThat(testedScope.isActive()).isFalse()
+
+        argumentCaptor<ViewEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture())
+            assertThat(lastValue)
+                .apply {
+                    hasSessionActive(false)
```
0xnm, line 7114: we also need to add this assertion to other tests in this file.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/KioskTrackingTest.kt
```diff
@@ -0,0 +1,139 @@
...
+        expectedEvents.add(
+            ExpectedViewEvent(
+                firstViewUrl,
+                docVersion = 2,
+                viewArguments = mapOf(),
+                sessionIsActive = false
+            )
+        )
+
+        onView(withId(R.id.end_session)).perform(click())
```
0xnm, line 134: why do we expect `sessionIsActive = false` before we click on `End Session`? shouldn't `onView(withId(R.id.end_session)).perform(click())` be above?

0xnm (commented):
I've added few comments, most of them are quite minor and not related to the functional part of the change, but some are for clarifying the functional part.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/KioskTrackingTest.kt
fuzzybinary: This is a weird thing about how these tests are written. I need to capture current scope before I actually end the session, so I have to create the expected event before I click the button. The same reason the other tests add the expected "stop view" before actually stopping the view.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/RumTestUtils.kt
fuzzybinary: I guess for ViewEvents yes they should always have this field. I'll fix.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -18,41 +19,62 @@ import com.datadog.android.v2.core.internal.storage.DataWriter
...
+            backgroundTrackingEnabled,
+            trackFrustrations,
+            firstPartyHostHeaderTypeResolver,
+            cpuVitalMonitor,
+            memoryVitalMonitor,
+            frameRateVitalMonitor,
+            sessionListener,
+            contextProvider,
+            false
+        )
```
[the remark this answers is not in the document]
fuzzybinary: Yeah. Might make the creation of the first session more concise as well.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
fuzzybinary: Yeah, this is the part I'm most concerned about as well.

The requirements are that when start a new session from a user action, it reactivates the last known view in thew new session.  The only good way I've found to do this is to reach down to get the currently active view in scope and save it until the new session is started.

The alternative would be either to have session scope keep its own copy of the currently active view and it's properties modified by events (which I worry could get out of sync), or broadcast up the relevant information on change, which also requires a store of the property. I'm open to either approach though I think the second is likely better.

But, `lastActiveViewScope` shouldn't be public either way.

dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScopeTest.kt
fuzzybinary: I think that was just what we decided to do. The requirements require that we start a new session in response to a user interaction, but in order to get any information from a user action, they have to be attached to a view. So, if the session is starting from an action, we take the last known view and restart it.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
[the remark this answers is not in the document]
fuzzybinary: Actually -- there's a difference in requirements here between the first session and subsequent sessions. The first session should be started by default during initialization, where it captures things like Application Launch. Subsequent sessions only start on a user interaction.

So I think it makes sense to create this default session session here. Let me know if you agree.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
fuzzybinary: It appears there are actually a lot of tests that assume a session is active by default, even without a SessionId. (`activeSession` on `RumApplicationScope` just returned the session it had.)

I'm curious why we don't go the opposite direction to make it consistent, and have the session generate its id on creation?

xgouchet (approved):

0xnm (commented):
lgtm! I've left some comments, but they are mostly not related to the main functionality.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -52,19 +55,27 @@ internal class RumSessionScope(
...
         cpuVitalMonitor,
         memoryVitalMonitor,
         frameRateVitalMonitor,
         appStartTimeProvider,
-        contextProvider
+        contextProvider,
+        applicationDisplayed
     )
 
+    internal val lastActiveViewScope: RumViewScope?
```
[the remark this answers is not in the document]
fuzzybinary: Yes, good catch.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -65,4 +87,57 @@ internal class RumApplicationScope(
...
+            memoryVitalMonitor,
+            frameRateVitalMonitor,
+            sessionListener,
+            contextProvider,
+            true
+        )
+        childScopes.add(newSession)
+        if (event !is RumRawEvent.StartView) {
+            lastActiveViewInfo?.let {
+                it.keyRef.get()?.let { key ->
```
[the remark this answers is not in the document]
fuzzybinary: Added a message. Can you review that it makes sense?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -65,4 +89,69 @@ internal class RumApplicationScope(
...
+                        key = key,
+                        name = it.name,
+                        attributes = it.attributes
+                    )
+                    newSession.handleEvent(startViewEvent, writer)
+                } else {
+                    internalLogger.log(
+                        InternalLogger.Level.WARN,
+                        InternalLogger.Target.USER,
+                        LAST_ACTIVE_VIEW_GONE_WARNING_MESSAGE.format(it.name)
```
0xnm, line 146: minor: better to explicitly specify the `Locale`

```suggestion
                        LAST_ACTIVE_VIEW_GONE_WARNING_MESSAGE.format(Locale.US, it.name)
```

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -8,51 +8,75 @@ package com.datadog.android.rum.internal.domain.scope
...
+            frameRateVitalMonitor,
+            sessionListener,
+            contextProvider,
+            false
+        )
     )
 
+    val activeSession: RumScope?
+        get() {
+            return childScopes.find { it.isActive() }
```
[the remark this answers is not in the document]
fuzzybinary: Yes, I actually have a similar telemetry in iOS. Maybe not here, though as it could get noisy. I'll see if I can find a better place to add it.

fuzzybinary:
@0xnm I added telemetry for checking if we create multiple active sessions during session start. I think that should handle the case you're talking about.

0xnm (approved):

PR #1362, opened 2023-03-30
RUMM-2948: Share internal logger with modules

0xnm:
### What does this PR do?

This change encapsulates internal logger and binds it and its usage to the particular SDK instance where possible.

PR may look big, but actual changes are almost the same everywhere.


dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -71,15 +70,22 @@ object Datadog {
...
                 )
                 return null
             }
 
-            val sdkCore = DatadogCore(context, credentials, configuration, sdkInstanceId)
+            val sdkCore = DatadogCore(
+                context,
+                credentials,
+                sdkInstanceId,
+                instanceName ?: SdkCoreRegistry.DEFAULT_INSTANCE_NAME
```
xgouchet, line 85: Maybe we could extract this in a local var `sdkInstanceName` and reuse it in the `registry.register(instanceName, sdkCore)` below. 
Right now the implementation ensures that the `instanceName`, if null, will be converted to the same `DEFAULT_INSTANCE_NAME`, but it might cause a bug in the future if we change the registry implementation. WDYT?

xgouchet (approved):
Nice work, left one question regarding the default instance name consistency

mariusc83 (approved):
LGTM. Nice job !!

dd-sdk-android/src/main/kotlin/com/datadog/android/Datadog.kt
0xnm: Indeed, good point. I pushed the changed in https://github.com/DataDog/dd-sdk-android/pull/1362/commits/8d45e9a30f51eeac882bae72c7fe619c553cc07a.

PR #1374, opened 2023-04-11
REPLAY-1540 Do not add translucent background wireframe for PopUpDecorView

mariusc83:
### What does this PR do?

The Android system draws a gray translucent layer over the main `Window` whenever using a `DialogWindow`. This layer is not found in the view tree and is somehow drew directly on the canvas so in order to mitigate this we were trying to manually (artificially) adding it to below the `DecorView` wireframe. The problem with this is approach is that this was applicable for all the decor view types including the `PopUpWindow$PopUpDecorView` which should not have this layer. In this PR we are addressing this issue.

### Motivation

### Additional Notes


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/DecorViewMapper.kt
```diff
@@ -27,27 +28,36 @@ internal class DecorViewMapper(
...
-                shapeStyle = MobileSegment.ShapeStyle(
-                    backgroundColor = WINDOW_WIREFRAME_COLOR,
-                    opacity = WINDOW_WIREFRAME_OPACITY
+        // effect in the player. We will only do this for `DialogWindow@DecorView` and not for
+        // `PopUpWindow@PopUpDecorView` types in order to follow what the system does. We do not have
+        // a way to access that type because it is private in the Android system so we will try to assert
+        // the class name in this case. Quite a nasty workaround but given the limitations is the best
+        // we could find. We know that the system classes are not obfuscated by Proguard so in case
+        // this class name will be changed in the future we will see it in our replays and fix it.
+        if (!decorClassName.lowercase(Locale.ENGLISH)
```
0xnm, line 41: Not sure if we need lowercase though here, normally class names follow the convention.

```suggestion
        if (!decorClassName.lowercase(Locale.US)
```

0xnm (approved):
lgtm, I've added a minor suggestion

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/DecorViewMapper.kt
mariusc83: yeah...good point.

PR #1382, opened 2023-04-13
RUMM-3197: Lazy usage of SDK instance in the OkHttp instrumentation

0xnm:
### What does this PR do?

This PR makes changes to the OkHttp instrumentation in a way that SDK instance is queried lazily now, when the actual call is made. This is needed to support the cases when OkHttp client is initialized before the SDK (for the remote config fetch, for example) and since it is immutable and can be used in the DI graph, re-building it is not an easy and convenient thing.

The goal is achieved by explicitly passing down the SDK instance in the `DatadogInterceptor`, `TracingInterceptor`, `RumInterceptor`, `DatadogEventListener` classes, which is taken from `SdkReference` class - it tries to acquire the SDK instance when it is available and does it only once (to avoid the possible thread bottleneck during the sync lock in the `getInstance` calls).

### DIFF
--- a/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
+++ b/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
@@ -10,11 +10,11 @@ import androidx.annotation.FloatRange
 import com.datadog.android.core.configuration.Configuration
 import com.datadog.android.core.configuration.HostsSanitizer
 import com.datadog.android.core.internal.net.DefaultFirstPartyHostHeaderTypeResolver
-import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.core.internal.utils.loggableStackTrace
 import com.datadog.android.core.internal.utils.percent
 import com.datadog.android.core.sampling.RateBasedSampler
 import com.datadog.android.core.sampling.Sampler
+import com.datadog.android.okhttp.utils.SdkReference
 import com.datadog.android.trace.AndroidTracer
 import com.datadog.android.trace.TracingHeaderType
 import com.datadog.android.v2.api.Feature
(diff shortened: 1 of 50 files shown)


dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -228,14 +240,27 @@ internal constructor(
 
     // region Internal
 
-    private fun isRequestTraceable(request: Request): Boolean {
+    internal open fun onSdkInstanceReady(sdkCore: SdkCore) {
```
mariusc83, line 243: good call. that might help a lot during the support investigations.

dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/utils/SdkReference.kt
```diff
@@ -0,0 +1,38 @@
...
+    private val onSdkInstanceCaptured: (SdkCore) -> Unit = {}
+) {
+
+    private val reference = AtomicReference<SdkCore>(null)
+
+    fun get(): SdkCore? {
+        val current = reference.get()
+        // TODO RUMM-3195 Check if instance is stopped and remove it?
+        @Suppress("IfThenToElvis") // less readable
+        return if (current == null) {
```
mariusc83, line 24: I think you still need a `synchronized` block here. That `AtomicReference` will not guarantee you that `Datadog.getInstance(sdkInstanceName)` will not be called multiple times from different threads. That means you will return 2 references even though in the `AtomicReference` you only keep one.
I would add:
```
if(current == null){
 synchronized(this){
     current = reference.get()
     if(current == null) {
                val sdkCore = Datadog.getInstance(sdkInstanceName)
                if (reference.compareAndSet(null, sdkCore)) {
                    onSdkInstanceCaptured(sdkCore)
                }
                sdkCore
     }
 }
```
}

0xnm: We don't need to synchronize here, because `Datadog.initialize` / `Datadog.getInstance` are already synchronized internally. This means that each time the same instance will be returned.

I'm using `compareAndSet` just to make sure that `onSdkInstanceCaptured` will be called once in case of multiple threads.

mariusc83: Good to know, tks for the explanation ;)

mariusc83 (approved):
LGTM

xgouchet (approved):

PR #1386, opened 2023-04-14
REPLAY-1552 Add the MaskInputEditTextViewMapper

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/MaskInputEditTextViewMapper.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.mapper
+
+import com.datadog.android.sessionreplay.internal.recorder.obfuscator.FixedLengthStringObfuscator
+import com.datadog.android.sessionreplay.utils.StringUtils
+import com.datadog.android.sessionreplay.utils.UniqueIdentifierGenerator
+import com.datadog.android.sessionreplay.utils.ViewUtils
+
+internal class MaskInputEditTextViewMapper(
```
0xnm, line 14: I have a concern here (maybe it is related to the architecture) - class which does nothing, but calls another class, even though with some custom parameters look a bit strange and non-obvious. For example here, having a values which can be set + default arguments imply that we can for example pass `textViewMapper` which won't mask anything and then class name (`MaskInputEditTextViewMapper`) won't match the actual functionality?

0xnm (approved):
lgtm, but I left once comment regarding the mappers - it is becoming a bit hard to navigate between them and to understand how they work (in general and also for exact classes), maybe we need to see if we can make things simpler.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/DefaultStringObfuscatorTest.kt
```diff
@@ -4,11 +4,10 @@
  * Copyright 2016-Present Datadog, Inc.
  */
 
-package com.datadog.android.sessionreplay.internal.recorder.mapper
+package com.datadog.android.sessionreplay.internal.recorder.obfuscator
```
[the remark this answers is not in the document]
mariusc83: yup good catch

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/MaskInputEditTextViewMapper.kt
mariusc83: yes...the non-default arguments are meant for testing only, this class should only encapsulate that `TextViewMapper` with those `FixedLengthObfuscators`. Would it be more visible if instead of using a generic `TextViewMapper` I create one `MaskInputTextViewMapper` which will be used by the `MaskInputEditTextViewMapper` ? 

Have to admit that I am wondering myself what kind of architecture I should use in this situation so any suggestion would help with the small note that I don't want to change this too much as the rules are not yet 100% detailed/approved and these might change later.

mariusc83: I actually decided to try to change it from the ground. I am going to come up with a new commit on top of this one with a proposal later. Will keep you posted.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
```diff
@@ -18,6 +18,7 @@ import android.widget.RadioButton
 import android.widget.SeekBar
 import android.widget.TextView
 import androidx.appcompat.widget.SwitchCompat
+import com.datadog.android.sessionreplay.internal.recorder.mapper.*
```
0xnm, line 21: shouldn't use star import

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/StringObfuscatorTest.kt
```diff
@@ -4,11 +4,10 @@
  * Copyright 2016-Present Datadog, Inc.
  */
 
-package com.datadog.android.sessionreplay.internal.recorder.mapper
+package com.datadog.android.sessionreplay.internal.recorder.obfuscator
```
0xnm, line 7: should be still `mapper` package according to the location? Or file should be moved?

0xnm (approved):
lgtm, I've added few comments

0xnm (approved):

xgouchet (approved):

PR #1388, opened 2023-04-17
RUMM-2986 update sdk instance id hash to use only instance name and datacenter

xgouchet:
### What does this PR do?

Remove the usage of the Client token in the instance id. This allows an application to rotate their token and still send previous data.

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/SdkCoreRegistry.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/SdkCoreRegistry.kt
@@ -69,6 +69,6 @@ internal class SdkCoreRegistry(
     // endregion
 
     companion object {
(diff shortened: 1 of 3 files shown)


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/SdkCoreRegistry.kt
```diff
@@ -69,6 +69,6 @@ internal class SdkCoreRegistry(
     // endregion
 
     companion object {
-        const val DEFAULT_INSTANCE_NAME = "main"
+        const val DEFAULT_INSTANCE_NAME = "_dd.sdk_core.default"
```
0xnm, line 72: do we change it to avoid possible clash with user provided name? This name will be used in the logcat output btw.

0xnm (approved):

mariusc83 (approved):

PR #1392, opened 2023-04-18
RUMM-3207: Support of dynamic sampling rate in OkHttp instrumentation

0xnm:
### What does this PR do?

This PR introduces the support of dynamic sampling in OkHttp instrumentation by changing the signature of all interceptors to accept `Sample` directly, exposing `RateBasedSampler` and making it to also accept a provider which will be called for the rate each time sampling decision needs to be made.

Also `RumInterceptor` is removed, because it was just a shorthand alias for the `DatadogInterceptor`.

The scale of sampling rate is now `[0;100]` everywhere, which means there is a change for the sampling in the `Logger.Builder` (method stays the same, but the range of expected values is different now).


xgouchet (approved):

mariusc83 (approved):
LGTM

PR #1393, opened 2023-04-18
RUMM-3194 add tracer sampling rate

xgouchet:
### What does this PR do?

Adds a sampling rate for local traces. The sampling will be applied automatically to the root span if it's not set manually on it.


dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/AndroidTracer.kt
```diff
@@ -193,6 +194,15 @@ class AndroidTracer internal constructor(
             return this
         }
 
+        /**
+         * Sets the sampling rate of spans.
+         * @param samplingRate the sampling rate as a percentage between 0 and 100 (default is 100%)
+         */
+        fun setSamplingRate(samplingRate: Double): Builder {
```
0xnm, line 201: we can also add `FloatRange` annotation for the clarity.

0xnm (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/AndroidTracer.kt
xgouchet: Makes sense indeed

mariusc83 (approved):
Nice work !

PR #1396, opened 2023-04-19
REPLAY-1598 Introduce the TextValueObfuscationRule

mariusc83:
### What does this PR do?

We are extracting the text obfuscation rules in separated logic outside the `WireframeMapper`. For this we are introducing the `TextValueObfuscationRule` implementations which are using the `TextTypeResolver` to determine the `TextType` for each `TextView`. The implementation rules of `TextValueObfuscationRule` type are following the rules defined in the document: [Privacy Options Rules](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2945942237/Privacy+Options+in+Mobile+Session+Replay)

### Motivation

### Additional Notes


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/obfuscator/rules/TextValueObfuscationRule.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import android.widget.TextView
+
+/**
+ * Will be used to apply the masking strategy for a [TextView] following the rules defined
+ * in the document: [Privacy Options Rules](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2945942237/Privacy+Options+in+Mobile+Session+Replay)
+ */
+internal interface TextValueObfuscationRule {
+
+    fun resolveObfuscatedValue(textView: TextView): String
```
ncreated, line 17: question/ Can we assume uniform behaviour for all `TextView` instances? Two negative examples:
* what if we want to mask differently `TextView` elements displayed in calendar vs in scroll view?
* in "fine-grained masking" feature, we will likely decide to [allow overriding masking behaviour in child elements](https://docs.datadoghq.com/real_user_monitoring/session_replay/privacy_options/#override-an-html-element) from their container - how would we know it here, given that we only have the access to `TextView` instance, not its container?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/obfuscator/rules/MaskInputObfuscationRule.kt
```diff
@@ -0,0 +1,29 @@
...
+        TextTypeResolver(),
+    private val textValueResolver: TextValueResolver = TextValueResolver()
+) : TextValueObfuscationRule {
+    override fun resolveObfuscatedValue(textView: TextView): String {
+        val textType = textTypeResolver.resolveTextType(textView)
+        val textValue = textValueResolver.resolveTextValue(textView)
+        return when (textType) {
+            TextType.SENSITIVE_TEXT -> fixedLengthStringObfuscator.obfuscate(textValue)
+            TextType.INPUT_TEXT -> fixedLengthStringObfuscator.obfuscate(textValue)
+            TextType.OPTION_TEXT -> fixedLengthStringObfuscator.obfuscate(textValue)
```
ncreated, line 25: question/ Can `OPTION_TEXT` appear in `TextView` element? From `TextTypeResolver` implementation I see it can not. It can mean we lack cohesion in this design.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/obfuscator/rules/MaskAllObfuscationRule.kt
```diff
@@ -0,0 +1,33 @@
...
+internal class MaskAllObfuscationRule(
+    private val defaultStringObfuscator: DefaultStringObfuscator =
+        DefaultStringObfuscator(),
+    private val fixedLengthStringObfuscator: FixedLengthStringObfuscator =
+        FixedLengthStringObfuscator(),
+    private val textTypeResolver: TextTypeResolver =
+        TextTypeResolver(),
+    private val textValueResolver: TextValueResolver = TextValueResolver()
+) : TextValueObfuscationRule {
+    override fun resolveObfuscatedValue(textView: TextView): String {
```
ncreated, line 22: question/ How would it scale to obfuscating text in other elements (e.g. options in value pickers or dropdown lists)? Will we keep adding more methods to `TextValueObfuscationRule` interface? That could indicate very strong coupling and a need for [shotgun surgery](https://refactoring.guru/smells/shotgun-surgery) to change behaviour of a single component.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/obfuscator/rules/TextValueObfuscationRule.kt
mariusc83: I thought of that too and I am going to introduce later a new type `OPTION_IN_PARENT_SELECTOR` which I am going to be able to resolve it from the `TextTypeResolver` by checking the parent type of the `TextView`. In Android the `TextView` has a `parent` property that I can leverage for this.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/obfuscator/rules/MaskInputObfuscationRule.kt
mariusc83: Not yet, coming in a new PR where like explained above I am going to use the `TextView#parent` property to check for its container.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/obfuscator/rules/MaskAllObfuscationRule.kt
mariusc83: By applying the strategy I explained above using the `parent` my plan is to always stick to these rules and only have 4 text types. The `parent` property will be used to check if the current `TextView` is part of a `SELECTOR` meaning it will be an `OPTION_TEXT`

mariusc83: The only thing that I need to add later will be probably a `HideObfuscationRule` for fine grained but will see when I get there.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/obfuscator/rules/TextValueObfuscationRule.kt
ncreated: That would mean we'll be effectively iterating the view tree not only "top (root) → bottom (leafs)" (`O(N)`) but also "bottom → up (root)" **for each view** in the hierarchy, no? We would need to recursively search `parents` in the hierarchy to find the ones that apply mask override (or traverse the entire tree up if none does it). This sounds like switching from `O(N)` to something much heavier for the main thread operation, no?

PR #1399, opened 2023-04-24
Force new session at sdk initialization

plousada:
Introduces the SdkInit event at the initialization of RumMonitor in order to make sure a session is started before the first RUM interaction. 
This makes it so that logs and traces are correctly correlated to a valid session if they occur before the first view or action takes place.

### What does this PR do?

### Motivation

### Additional Notes


xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -507,6 +511,13 @@ internal class DatadogRumMonitor(
         }
     }
 
+    internal fun sendSdkInitEvent() {
+        val eventTime = getEventTime(emptyMap())
```
0xnm, line 515: I think we don't need this line. `getEventTime(emptyMap())` will return `Time()` object, but the same object will be created at the `SdkInit` constructor with default argument.

0xnm (approved):
lgtm! I've added a comment about redundant line.

PR #1400, opened 2023-04-25
REPLAY-1598 Apply masking rule for texts in calendars, time pickers, dropdowns

mariusc83:
### What does this PR do?

Text fields used in the time/date pickers or dropdown lists should be considered as `input texts` and masked accordingly. Because we could not use a property on these fields that could tell us if these are part of these special components we needed to find an optimised way to pass the information from their parents when traversing the tree. That is the reason why we introduced the `MappingContext` type which will later on be leveraged also for passing the fine grained masking rule from any ancestor. Because those special type components are not visible we had to use a string assertion at the `className` level to detect them.

### Motivation

### Additional Notes


library/dd-sdk-android-session-replay/consumer-rules.pro
```diff
@@ -0,0 +1,5 @@
+# Keep the optional selector class name. We need this in the SR recorder.
+-keepnames class android.widget.DropDownListView
```
0xnm, line 2: this is a system class, it won't be obfuscated, so I think this line can be removed.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/optionselectormocks/DropDownListView.kt
```diff
@@ -0,0 +1,16 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package android.widget
+
+import android.content.Context
+import android.view.ViewGroup
+
+internal class DropDownListView(context: Context) : ViewGroup(context) {
```
0xnm, line 12: do we need this class? `android.widget.DropDownListView` will always be in the classpath, because it is a system API.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/optionselectormocks/MaterialCalendarGridView.kt
```diff
@@ -0,0 +1,16 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.google.android.material.datepicker
+
+import android.content.Context
+import android.view.ViewGroup
+
+internal class MaterialCalendarGridView(context: Context) : ViewGroup(context) {
```
0xnm, line 12: for `com.google.android.material.datepicker.MaterialCalendarGridView`, `com.google.android.material.timepicker.TimePickerView` and `androidx.appcompat.widget.DropDownListView ` you can add the necessary libraries directly to the test classpath using `testImplementation` dependency.

This will have an advantage of properly tracking name/package changes for these classes (instead of having our own versions).

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/OptionSelectorDetector.kt
```diff
@@ -0,0 +1,34 @@
...
+internal class OptionSelectorDetector {
+    fun isOptionSelector(view: ViewGroup): Boolean {
+        val parentClassName = view.javaClass.canonicalName ?: ""
+        return parentClassName in OPTION_SELECTORS_CLASS_NAMES_SET ||
+            AppCompatSpinner::class.java.isAssignableFrom(view::class.java)
+    }
+
+    companion object {
+        private const val DROP_DOWN_LIST_CLASS_NAME = "android.widget.DropDownListView"
+        private const val APPCOMPAT_DROP_DOWN_LIST_CLASS_NAME =
```
0xnm, line 21: if we already have `androidx.appcompat` library in the runtime classpath, cannot we simply do something like `DropDownListView.javaClass == view.javaClass`? this will remove the need to depend on the name and have name entry in the Proguard rules.

library/dd-sdk-android-session-replay/consumer-rules.pro
mariusc83: good point ;)

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/optionselectormocks/DropDownListView.kt
mariusc83: yeah the main issue is that all these classes are either internal to their packages or `hidden` so you cannot use them in your code. I even tried to look for them in the `ClassLoader` at runtime and you can't find them.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/optionselectormocks/MaterialCalendarGridView.kt
mariusc83: yes same as above you can't use them as they are either not public or `hidden`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/OptionSelectorDetector.kt
mariusc83: same as above

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/optionselectormocks/MaterialCalendarGridView.kt
0xnm: Here I'm not talking about using them directly, we are referencing them by name anyway. Here I mean that we don't need to create mocks, we can have them in the classpath by adding the libraries to the test classpath only.

PR #1405, opened 2023-04-28
RUMM-3216 use jank stats

xgouchet:
### What does this PR do?

Use the JankStats library to get the actual frame durations. 

### Motivation

Using the Choreographer like we did forced a redraw of all frames, which was not necessary if the content didn't require it. 

### Additional Notes


dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -231,30 +230,16 @@ internal class RumFeature(
     }
 
     private fun initializeVitalReaders(periodInMs: Long) {
-        @Suppress("UnsafeThirdPartyFunctionCall") // pool size can't be <= 0
+        @Suppress(
+            "UnsafeThirdPartyFu" +
+                "nctionCall"
+        ) // pool size can't be <= 0
```
0xnm, line 236: seems like a wrong change

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    override fun onActivityDestroyed(activity: Activity) {
+    }
+
+    // endregion
+
+    // region JankStats.OnFrameListener
+
+    override fun onFrame(volatileFrameData: FrameData) {
+        println("$volatileFrameData")
```
0xnm, line 63: to be removed

gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 # Commons
-kotlin = "1.6.10"
-kotlinSP = "1.6.10-1.0.4"
+kotlin = "1.8.10"
+kotlinSP = "1.8.10-1.0.9"
```
0xnm, line 4: do we need it as a part of this PR? I would rather handle it in the SDK v2 branch with Gradle 8 update (which is based on Kotlin 1.8 as well, so that we don't have conflicts for the build classpath), and also checking API/ABI compatibility.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/home/HomeFragment.kt
```diff
@@ -15,7 +15,7 @@ import androidx.core.view.children
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.ViewModelProviders
 import androidx.navigation.NavController
-import androidx.navigation.fragment.NavHostFragment.findNavController
+import androidx.navigation.fragment.NavHostFragment.Companion.findNavController
```
0xnm, line 18: Why do we need to specify `Companion` now? Is it related to the Kotlin 1.8 change?

0xnm (commented):
nice solution! I've added few comments to clarify.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -0,0 +1,82 @@
...
+internal class JankStatsActivityLifecycleListener(
+    private val vitalObserver: VitalObserver,
+    private val internalLogger: InternalLogger
+) : ActivityLifecycleCallbacks, JankStats.OnFrameListener {
+
+    // region ActivityLifecycleCallbacks
+
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        try {
+            JankStats.createAndTrack(activity.window, this)
```
mariusc83, line 29: this will work only for activity windows but will not measure the frame rate for `DialogWindow` or `PopUpWindow`. I guess once we add that ticket to support that in RUM we will have to find a different way to plugin in the `JankStats` into those windows too.

gradle/libs.versions.toml
xgouchet: Unfortunately I need to use this version otherwise there's an incompatibility with the Metrics library.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -236,25 +235,8 @@ internal class RumFeature(
...
-            )
-            internalLogger.log(
-                InternalLogger.Level.WARN,
-                InternalLogger.Target.USER,
-                "It seems you initialized the SDK on a thread without a Looper: " +
-                    "we won't be able to track your Views' refresh rate."
-            )
-        }
+        jankStatsActivityLifecycleListener = JankStatsActivityLifecycleListener(frameRateVitalMonitor, internalLogger)
+        (appContext as? Application)?.registerActivityLifecycleCallbacks(jankStatsActivityLifecycleListener)
```
0xnm, line 239: should we modify the way we are checking refresh rate? currently we are using `RumViewScope#detectRefreshRateScale`, which relies on the view key and the presence of activity/fragment there (which may be missing in case of manual instrumentation or calls coming from cross-platform).

But now, since we are listening for activities anyway, we can access UI context at any time and basically pass this info with a sample, making things work properly even for cross-platform.

wdyt?

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -0,0 +1,81 @@
...
+    override fun onActivityDestroyed(activity: Activity) {
+    }
+
+    // endregion
+
+    // region JankStats.OnFrameListener
+
+    override fun onFrame(volatileFrameData: FrameData) {
+        val durationNs = volatileFrameData.frameDurationUiNanos
+        if (durationNs > 0.0) {
```
0xnm, line 64: in which case it can be negative or zero?

0xnm (commented):
nice! I've left a comment wondering if we should change also the way we detect refresh rate.

0xnm (approved):

PR #1408, opened 2023-05-02
RUMM-3250: Keep bytecode produced compatible with Java 11

0xnm:
### What does this PR do?

Rolling back some changes done in https://github.com/DataDog/dd-sdk-android/pull/1406: we need to keep bytecode produced at the Java 11 level, because `aar` files are distributed non-dexed (meaning they carry simple `.class` jvm files), so old AGP versions may be not able to handle bytecode produced by Java 17.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -79,7 +79,8 @@ gradlePlugin {
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-    kotlinOptions.jvmTarget = JavaVersion.VERSION_17.toString()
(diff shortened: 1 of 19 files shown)


fuzzybinary (approved):

xgouchet (approved):
Maybe we could have a constant set up in our `buildSrc` and reference it from all modules instead of changing the constant everywhere?

That would make the next updates easier

0xnm:
@xgouchet good suggestion. I extracted it to the `CompileOptions.java11`, so that it is now centralized.

PR #1413, opened 2023-05-10
Add extra error output in case of a failed upload

fuzzybinary:
### What does this PR do?

We currently send failed upload information to the maintainer logger, but not to the user console. I've added a line at the `verbose` level for users to help diagnose network errors.

### Motivation

I am trying to diagnose an error on batch upload and can't seem to get enough information to figure out what's happening. Hoping this will help in the future.

### Additional Notes

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
@@ -56,6 +56,12 @@ internal class DataOkHttpUploader(
                 "Unable to upload batch data.",
                 e
             )
+            internalLogger.log(
+                InternalLogger.Level.DEBUG,
+                InternalLogger.Target.USER,
+                "Unable to upload batch data.",
+                e
+            )
             UploadStatus.NETWORK_ERROR
         }
 


dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
```diff
@@ -56,6 +56,12 @@ internal class DataOkHttpUploader(
                 "Unable to upload batch data.",
                 e
             )
+            internalLogger.log(
+                InternalLogger.Level.DEBUG,
+                InternalLogger.Target.USER,
+                "Unable to upload batch data.",
+                e
+            )
```
xgouchet, line 64: you can combine the targets and make just one call: 
```
            internalLogger.log(
                InternalLogger.Level.ERROR,
                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.USER)
                "Unable to upload batch data.",
                e
            )
```

fuzzybinary: The only reason we might not want to do that is because of the difference in level -- this isn't really an error from the user's perspective, as we're just going to retry. But we need the extra info in the log if the user is experiencing significant trouble uploading data.

Alternately we can set the level to Debug across the board for this error? Thoughts @0xnm ?

0xnm: imo let's keep it different: debug - for the user, error for us. Having error at least is more visible/noticeable in logcat.

0xnm (approved):

PR #1414, opened 2023-05-12
RUMM-3195: SdkReference will release SDK instance if it is stopped

0xnm:
### What does this PR do?

This PR makes `SdkReference` to release SDK instance if it is stopped.

On top of that 2 methods are removed from `SdkCore` interface:

* `stop` - because we already have `Datadog.stopInstance` which does the right thing, we don't want users to call `SdkCore.stop`, because it won't remove instance from the registry.
* `flushStoredData` - there is no usage of it outside of `core` module and unlikely it will be needed for the users.

`SdkReference` class was moved to the `core` module, because it will be used in other modules.


xgouchet (approved):
Great job

PR #1420, opened 2023-05-16
REPLAY-1301 Provide placeholder for Action Bar

jonathanmos:
### What does this PR do?
Unsupported view placeholder support

- Provide an unsupported view placeholder for android.widget.Toolbar and androidx.appcompat.widget.Toolbar
- Refactor unit tests for SessionReplayPrivacyTest

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
@@ -125,10 +125,7 @@ enum class SessionReplayPrivacy {
             MapperTypeWrapper(SwitchCompat::class.java, switchCompatMapper.toGenericMapper()),
             MapperTypeWrapper(RadioButton::class.java, radioButtonMapper.toGenericMapper()),
             MapperTypeWrapper(CheckBox::class.java, checkBoxMapper.toGenericMapper()),
-            MapperTypeWrapper(
-                CheckedTextView::class.java,
-                checkedTextViewMapper.toGenericMapper()
-            ),
+            MapperTypeWrapper(CheckedTextView::class.java, checkedTextViewMapper.toGenericMapper()),
             MapperTypeWrapper(Button::class.java, buttonMapper.toGenericMapper()),
             MapperTypeWrapper(EditText::class.java, editTextViewMapper.toGenericMapper()),
             MapperTypeWrapper(TextView::class.java, textMapper.toGenericMapper()),
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
@@ -17,6 +17,7 @@ import android.widget.NumberPicker
 import android.widget.RadioButton
 import android.widget.SeekBar
 import android.widget.TextView
+import android.widget.Toolbar
 import androidx.appcompat.widget.SwitchCompat
 import com.datadog.android.sessionreplay.internal.recorder.mapper.BasePickerMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.ButtonMapper
@@ -37,9 +38,11 @@ import com.datadog.android.sessionreplay.internal.recorder.mapper.RadioButtonMap
 import com.datadog.android.sessionreplay.internal.recorder.mapper.SeekBarWireframeMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.SwitchCompatMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.TextViewMapper
+import com.datadog.android.sessionreplay.internal.recorder.mapper.UnsupportedViewMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.ViewScreenshotWireframeMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.ViewWireframeMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.WireframeMapper
+import androidx.appcompat.widget.Toolbar as AppCompatToolbar
 
 /**
  * Defines the Session Replay privacy policy when recording the sessions.
@@ -73,6 +76,7 @@ enum class SessionReplayPrivacy {
     @Suppress("LongMethod")
     internal fun mappers(): List<MapperTypeWrapper> {
         val viewWireframeMapper = ViewWireframeMapper()
+        val unsupportedViewMapper = UnsupportedViewMapper()
         val imageMapper: ViewScreenshotWireframeMapper
         val textMapper: TextViewMapper
         val buttonMapper: ButtonMapper
@@ -129,8 +133,14 @@ enum class SessionReplayPrivacy {
             MapperTypeWrapper(Button::class.java, buttonMapper.toGenericMapper()),
             MapperTypeWrapper(EditText::class.java, editTextViewMapper.toGenericMapper()),
             MapperTypeWrapper(TextView::class.java, textMapper.toGenericMapper()),
-            MapperTypeWrapper(ImageView::class.java, imageMapper.toGenericMapper())
+            MapperTypeWrapper(ImageView::class.java, imageMapper.toGenericMapper()),
+            MapperTypeWrapper(AppCompatToolbar::class.java, unsupportedViewMapper.toGenericMapper())
         )
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            mappersList.add(0, MapperTypeWrapper(Toolbar::class.java, unsupportedViewMapper.toGenericMapper()))
+        }
+
         seekBarMapper?.let {
             mappersList.add(0, MapperTypeWrapper(SeekBar::class.java, it.toGenericMapper()))
         }
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
@@ -1,0 +1,88 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.mapper
+
+import android.view.View
+import com.datadog.android.sessionreplay.internal.recorder.MappingContext
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal class UnsupportedViewMapper :
+    BaseWireframeMapper<View, MobileSegment.Wireframe.TextWireframe>() {
+
+    override fun map(view: View, mappingContext: MappingContext):
+        List<MobileSegment.Wireframe.TextWireframe> {
+        val viewGlobalBounds = resolveViewGlobalBounds(
+            view,
+            mappingContext.systemInformation.screenDensity
+        )
+
+        return listOf(
+            MobileSegment.Wireframe.TextWireframe(
+                id = resolveViewId(view),
+                x = viewGlobalBounds.x,
+                y = viewGlobalBounds.y,
+                width = viewGlobalBounds.width,
+                height = viewGlobalBounds.height,
+                shapeStyle = resolveShapeStyle(view),
+                border = resolveBorder(),
+                text = view::class.java.name,
+                textStyle = resolveTextStyle(),
+                textPosition = resolveTextPosition(),
+            )
+        )
+    }
+
+    // region Internal
+
+    private fun resolveTextStyle():
+            MobileSegment.TextStyle{
+        return MobileSegment.TextStyle(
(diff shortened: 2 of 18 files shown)


library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacyTest.kt
```diff
@@ -20,277 +22,167 @@ import com.datadog.android.sessionreplay.internal.recorder.mapper.ButtonMapper
...
-    }
+    @ParameterizedTest(name = "M return mappers W rule({1}, API Level {0}")
+    @MethodSource("provideTestArguments")
+    fun `M return mappers W rule({1}, API Level {0}`(
+            apiLevel: Int,
+            maskLevel: String,
+            expectedMappers: List<MapperTypeWrapper>
+    ) {
+        // Given
+        Build.VERSION::class.java.setStaticValue("SDK_INT", apiLevel)
```
mariusc83, line 63: I think we need to put this back to the original value once the test suite is finished otherwise it might affect other tests outside of this class.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacyTest.kt
```diff
@@ -20,277 +22,167 @@ import com.datadog.android.sessionreplay.internal.recorder.mapper.ButtonMapper
...
-        assertThat(mappers[6].type).isEqualTo(EditText::class.java)
-        assertThat(mappers[6].mapper).isInstanceOf(EditTextViewMapper::class.java)
-        assertThat(mappers[7].type).isEqualTo(TextView::class.java)
-        assertThat(mappers[7].mapper).isInstanceOf(MaskAllTextViewMapper::class.java)
-        assertThat(mappers[8].type).isEqualTo(ImageView::class.java)
-        assertThat(mappers[8].mapper).isInstanceOf(ViewScreenshotWireframeMapper::class.java)
-    }
+    @ParameterizedTest(name = "M return mappers W rule({1}, API Level {0}")
+    @MethodSource("provideTestArguments")
+    fun `M return mappers W rule({1}, API Level {0}`(
```
mariusc83, line 57: Really good job here, this is much more readable now 👏

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
```diff
@@ -0,0 +1,88 @@
...
+        return listOf(
+            MobileSegment.Wireframe.TextWireframe(
+                id = resolveViewId(view),
+                x = viewGlobalBounds.x,
+                y = viewGlobalBounds.y,
+                width = viewGlobalBounds.width,
+                height = viewGlobalBounds.height,
+                shapeStyle = resolveShapeStyle(view),
+                border = resolveBorder(),
+                text = view::class.java.name,
```
0xnm, line 32: so instead of the screenshot we are going to show the box with class name? What will be a solution for the case when class name is obfuscated?

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapperTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    @Test
+    fun `M resolve with the class name of the View W map()`() {
+        // When
+        val wireframes = testedUnsupportedViewMapper.map(
+            mockUnsupportedView,
+            fakeMappingContext
+        )
+
+        // Then
+        Assertions.assertThat(wireframes[0].text)
```
0xnm, line 67: very minor: normally we import `Assertions.assertThat` directly, so that no need to type `Assertions.assertThat` and we can can use `assertThat` directly. but not an issue

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapperTest.kt
```diff
@@ -0,0 +1,70 @@
...
+            .isEqualTo(mockUnsupportedView::class.java.name)
```
0xnm, line 68: should we assert some other properties?

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacyTest.kt
```diff
@@ -15,282 +17,176 @@ import android.widget.NumberPicker
...
+        apiLevel: Int,
+        maskLevel: String,
+        expectedMappers: List<MapperTypeWrapper>
+    ) {
+        // Given
+        Build.VERSION::class.java.setStaticValue("SDK_INT", apiLevel)
```
0xnm, line 63: ideally we need to reset this value back to the original after the test (this is what `ApiLevelExtension` does), because it is a static value. Maybe it is still better to have parametrized test for below API 26 and another parametrized test for API above 26? wdyt?

0xnm (commented):
nice work! I've added some comments, let me know what you think.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
```diff
@@ -129,8 +133,14 @@ enum class SessionReplayPrivacy {
             MapperTypeWrapper(Button::class.java, buttonMapper.toGenericMapper()),
             MapperTypeWrapper(EditText::class.java, editTextViewMapper.toGenericMapper()),
             MapperTypeWrapper(TextView::class.java, textMapper.toGenericMapper()),
-            MapperTypeWrapper(ImageView::class.java, imageMapper.toGenericMapper())
+            MapperTypeWrapper(ImageView::class.java, imageMapper.toGenericMapper()),
+            MapperTypeWrapper(AppCompatToolbar::class.java, unsupportedViewMapper.toGenericMapper())
```
mariusc83, line 137: let's try to do this more generic like we discussed. Maybe to be applied for all the implementations of the `View` type that do not have an explicit mapper yet.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
mariusc83: well this is only used right now explicitly for system classes so in theory those will not be obfuscated.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacyTest.kt
```diff
@@ -87,13 +89,16 @@ internal class SessionReplayPrivacyTest {
...
-                MapperTypeWrapper(Button::class.java, mockButtonMapper.toGenericMapper()),
-                MapperTypeWrapper(EditText::class.java, mockEditTextViewMapper.toGenericMapper()),
-                MapperTypeWrapper(ImageView::class.java, mockImageMapper.toGenericMapper()),
-        )
+            MapperTypeWrapper(Button::class.java, mockButtonMapper.toGenericMapper()),
+            MapperTypeWrapper(EditText::class.java, mockEditTextViewMapper.toGenericMapper()),
+            MapperTypeWrapper(ImageView::class.java, mockImageMapper.toGenericMapper()),
 
+            MapperTypeWrapper(AppCompatToolbar::class.java, mockUnsupportedViewMapper.toGenericMapper()),
+            MapperTypeWrapper(Toolbar::class.java, mockUnsupportedViewMapper.toGenericMapper())
```
mariusc83, line 100: shouldn't you add a test case for when the SDK is above `LOLLIPOP` ? The way I see it here you are not testing that `if` branch because you are adding this mapper in the `baseMappers` by default.

mariusc83 (changes requested):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
0xnm: cannot we have this class to be used with say views coming from `androidx` or any other non-system package?

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacyTest.kt
jonathanmos: Yes, good point. I've added an afterEach hook to reset the api level

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapperTest.kt
jonathanmos: I've added additional tests to check the other properties

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacyTest.kt
jonathanmos: Added additional testing for Kitkat

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
jonathanmos: Changed to have a static text - for toolbars this will be "Toolbar", and as a fallback for unsupported views that are not explicitly mapped it will be "Unsupported view"

xgouchet (approved):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
```diff
@@ -38,6 +41,24 @@ internal class UnsupportedViewMapper :
...
 
+    private fun resolveViewTitle(view: View): String {
+        return if (isToolbar(view)) {
+            return TOOLBAR_LABEL
+        } else {
+            DEFAULT_LABEL
+        }
+    }
+
+    private fun isToolbar(view: View): Boolean {
```
mariusc83, line 52: nice ;)

mariusc83 (approved):
Nice work and thank you for all the updates ;)

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/UnsupportedViewMapper.kt
```diff
@@ -0,0 +1,111 @@
...
+        )
+    }
+
+    // region Internal
+
+    private fun resolveViewTitle(view: View): String {
+        return if (isToolbar(view)) {
+            return TOOLBAR_LABEL
+        } else {
+            DEFAULT_LABEL
```
0xnm, line 48: just an idea: btw, maybe we can do a trick here: check if class name contains `View` - it it is there, probably it is non-obfuscated class name and we can use it directly, if it is not there - probably it is obfuscated class name (or some other naming convention) and we can use default label.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacyTest.kt
```diff
@@ -15,282 +17,205 @@ import android.widget.NumberPicker
...
-        assertThat(mappers[5].type).isEqualTo(Button::class.java)
-        assertThat(mappers[5].mapper).isInstanceOf(ButtonMapper::class.java)
-        assertThat(mappers[6].type).isEqualTo(EditText::class.java)
-        assertThat(mappers[6].mapper).isInstanceOf(EditTextViewMapper::class.java)
-        assertThat(mappers[7].type).isEqualTo(TextView::class.java)
-        assertThat(mappers[7].mapper).isInstanceOf(TextViewMapper::class.java)
-        assertThat(mappers[8].type).isEqualTo(ImageView::class.java)
-        assertThat(mappers[8].mapper).isInstanceOf(ViewScreenshotWireframeMapper::class.java)
+    private fun setApiLevel(newApiLevel: Int) {
+        Build.VERSION::class.java.setStaticValue("SDK_INT", newApiLevel)
```
0xnm, line 92: fyi: beware that these calls may have no effect at some point. Because after dozens of calls most likely JVM will inline this property value (because from JVM point of view it has the right to do so - it is `static final`), this is what we experienced in the past. Ideally in the future we shouldn't touch `static final` things with reflection.

0xnm (approved):

xgouchet (approved):

xgouchet (approved):

mariusc83 (approved):

PR #1421, opened 2023-05-16
REPLAY-1614 Mask calendar header for MASK_ALL or MASK_USER_INPUT

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ComposedOptionSelectorDetector.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ComposedOptionSelectorDetector.kt
@@ -1,0 +1,16 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import android.view.ViewGroup
+
+internal class ComposedOptionSelectorDetector(private val detectors: List<OptionSelectorDetector>) :
+    OptionSelectorDetector {
+    override fun isOptionSelector(view: ViewGroup): Boolean {
+        return detectors.any { it.isOptionSelector(view) }
+    }
+}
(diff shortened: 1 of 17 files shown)


library/dd-sdk-android-session-replay/build.gradle.kts
```diff
@@ -78,6 +78,7 @@ dependencies {
     implementation(libs.kotlin)
     implementation(libs.gson)
     implementation(libs.androidXAppCompat)
+    implementation(libs.googleMaterial)
```
0xnm, line 81: doesn't it then make `dd-sdk-android-session-replay-material` module useless? I though the purpose of that module was to avoid having `material` dependency in the main module.

mariusc83: ohhhh sorry that was a leftover 🤦

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ComposedOptionSelectorDetector.kt
```diff
@@ -0,0 +1,21 @@
...
+
+internal class ComposedOptionSelectorDetector(private val detectors: List<OptionSelectorDetector>) :
+    OptionSelectorDetector {
+    override fun isOptionSelector(view: ViewGroup): Boolean {
+        detectors.forEach {
+            if (it.isOptionSelector(view)) {
+                return true
+            }
+        }
+        return false
```
0xnm, line 19: can be even simpler I think:

```suggestion
        return detectors.any { it.isOptionSelector(view) }
```

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/ComposedOptionSelectorDetectorTest.kt
```diff
@@ -0,0 +1,61 @@
...
+            mock {
+                whenever(it.isOptionSelector(mockViewGroup))
+                    .thenReturn(fakeBundledDetectorValue)
+            }
+        }
+        testedComposedOptionSelectorDetector = ComposedOptionSelectorDetector(mockBundledDetectors)
+    }
+
+    @Test
+    fun `M delegate to bundled detectors W isOptionSelector`() {
```
0xnm, line 57: this test checks the case when all detectors return the same value. 

shouldn't we also add a test for the case when only one detector return `true` as well?

0xnm (approved):
lgtm, I've added few suggestions.

0xnm (approved):
lgtm, I've added few suggestions.

0xnm (approved):
lgtm, I've added few suggestions.

0xnm (approved):

PR #1423, opened 2023-05-17
RUMM-2537: Apply configuration standard

0xnm:
### What does this PR do?

This PR renames some configuration options in the public API to conform with the SDK config standardisation RFC. What was changed:

* `envName` property renamed to `env` (only in the public API)
* `serviceName` property renamed to `service` (only in the public API)
* `trackInteractions` / `disableInteractionTracking` were renamed to `trackUserInteractions` / `disableUserInteractionTracking` accordingly.
* removed RUM mention from the `setRumXXXEventMapper` methods (because these methods are already in the builder dedicated to RUM).
* renamed `sampling rate` to `sample rate` everywhere (to be aligned with the RFC and also to simplify text search).
* updated names of the methods defining sample rate
* removes deprecated properties from `DatadogEndpoint`

This PR doesn't rename properties related to the first party / traced hosts, because using term `traced` will be wrong in case of RUM (where we need to set `provider` property which can be `first party` according to schema), and anyway in the browser SDK the term different from the one proposed in RFC is used already - `allowedTracingUrls`. 


mariusc83 (approved):
LGTM

xgouchet (approved):

PR #1448, opened 2023-06-01
RUMM-3314: Keep old viewIds for view scope

0xnm:
### What does this PR do?

We need to keep old RUM view scope IDs, because `viewId` can be changed on the [session renewal](https://github.com/DataDog/dd-sdk-android/blob/c9ab67cf0a1565f94a0ed3002eb5a6a78ad0cde0/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L199), meaning that events like `ActionSent`, etc. which do the match by `viewId` won't match, because they will carry old `viewId`.


xgouchet (approved):
Nice work

PR #1458, opened 2023-06-05
RUMM-3323: Create lint check for the internal API usage

0xnm:
### What does this PR do?

This PR creates an Android Lint check which will report lint error is some publicly accessible internal API is used outside of Datadog packages.

It is implemented by checking the following sources:

* https://googlesamples.github.io/android-custom-lint-rules/api-guide
* https://github.com/googlesamples/android-custom-lint-rules
* [CheckResultDetector](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/CheckResultDetector.kt;drc=c12132327840ec10b640bdba1962d7cf54d5a0a9)

In the next commit I will add `InternalApi` annotation to the core module and put it on the publicly accessible internal API.

**NOTE**: this is very basic implementation and to make things simple annotation will be supposed to be used only for classes, methods and properties (not on the interface directly, because in this case usage site is reported as a whole class implementing it).


xgouchet:
Nice work, the question is why are we creating a custom lint rule instead of a custom detekt rule?

xgouchet (approved):

jonathanmos (approved):

PR #1472, opened 2023-06-14
RUMM-3340: Create a dedicated internal interface for feature-related methods of SDK core instance

0xnm:
### What does this PR do?

This PR creates a dedicated interface for SDK core instance for hosting feature-related method. Now the inheritance chain looks like the following: `SdkCore` -> `FeatureSdkCore` -> `InternalSdkCore`.

First two interfaces are publicly available.

SDK core is **guaranteed** to implement `FeatureSdkCore` (even in case of `NoOpSdkCore`), so inside the features context we can cast to this interface explicitly (this also removes the need to come up with any fallback if needed).


buraizu (approved):

mariusc83 (approved):
LGTM

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/NoOpSdkCore.kt
```diff
@@ -21,7 +22,7 @@ import java.util.concurrent.TimeUnit
 /**
  * A no-op implementation of [SdkCore].
  */
-internal class NoOpSdkCore : SdkCore {
+internal class NoOpSdkCore : FeatureSdkCore {
```
xgouchet, line 25: NitPick, I'd rename this as `NoOpFeatureSdkCore`

xgouchet (approved):

dd-sdk-android/src/main/kotlin/com/datadog/android/v2/core/NoOpSdkCore.kt
0xnm: indeed, makes sense. I will do it in another PR, to avoid rebasing other stacked PRs.

PR #1481, opened 2023-06-18
REPLAY-1722 Implement recorded data queue

jonathanmos:
### What does this PR do?

The traversal of nodes and production of snapshots (and touch events) was, up until now, synchronous. The recorded data queue enables asynchronous tasks to be performed during the traversal and for snapshots and touch events to still be generated in order and sent. 

In general terms what happens is this (for a snapshot item):
1) An item is generated and inserted into the queue, however this item has no nodes yet and so cannot be processed. A reference to this item will be passed during the traversal through all the mappers.
2) The traversal of the node tree finishes. If we didn't find any images on the way and if we successfully got a list of nodes then we update the item and try to process the queue. In a future pr the logic will be: If we encounter an image in a mapper during the traversal we increment the pendingImages counter for the item that we inserted. If we finish processing an image in a mapper we decrement this counter asynchronously (and potentially trigger the queue processing if pendingImages has reached zero).
3) If a request to process the queue arrives, we start going through the queue from the head. If the next item is invalid (no nodes or has been in the queue too long) then we remove it. Since the traversals happen on the main thread, we shouldn't be able to have a situation where a later traversal interferes with a previous traversal and ejects it from the queue before it's finished. Even if the previous traversal has pendingImages, it will have it's nodes already populated before the later traversal can awaken the queue. Valid items without pending images are sent to the processor similar to the way it used to work.
4) We continue processing the queue until either it is empty or the next item is not ready for processing.

### Motivation

Base64 requires async work on drawables and bitmaps during the traversal

### DIFF
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
@@ -13,10 +13,9 @@ import android.view.Window
 import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import com.datadog.android.sessionreplay.internal.LifecycleCallback
-import com.datadog.android.sessionreplay.internal.NoOpRecordCallback
-import com.datadog.android.sessionreplay.internal.RecordCallback
 import com.datadog.android.sessionreplay.internal.RecordWriter
 import com.datadog.android.sessionreplay.internal.SessionReplayLifecycleCallback
+import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueHandler
 import com.datadog.android.sessionreplay.internal.processor.RecordedDataProcessor
 import com.datadog.android.sessionreplay.internal.processor.RumContextDataHandler
 import com.datadog.android.sessionreplay.internal.recorder.ComposedOptionSelectorDetector
@@ -31,9 +30,6 @@ import com.datadog.android.sessionreplay.internal.recorder.callback.OnWindowRefr
 import com.datadog.android.sessionreplay.internal.recorder.mapper.MapperTypeWrapper
 import com.datadog.android.sessionreplay.internal.utils.RumContextProvider
 import com.datadog.android.sessionreplay.internal.utils.TimeProvider
-import java.util.concurrent.LinkedBlockingDeque
-import java.util.concurrent.ThreadPoolExecutor
-import java.util.concurrent.TimeUnit
 
 internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
 
@@ -42,24 +38,15 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
     private val privacy: SessionReplayPrivacy
     private val recordWriter: RecordWriter
     private val timeProvider: TimeProvider
-    private val recordCallback: RecordCallback
     private val customMappers: List<MapperTypeWrapper>
     private val customOptionSelectorDetectors: List<OptionSelectorDetector>
     private val windowInspector: WindowInspector
     private val windowCallbackInterceptor: WindowCallbackInterceptor
     private val sessionReplayLifecycleCallback: LifecycleCallback
-    private val processor: RecordedDataProcessor
+    private val recordedDataQueueHandler: RecordedDataQueueHandler
     private val viewOnDrawInterceptor: ViewOnDrawInterceptor
     private val uiHandler: Handler
 
-    @Suppress("UnsafeThirdPartyFunctionCall") // workQueue can't be null
-    private val processorExecutorService = ThreadPoolExecutor(
-        CORE_DEFAULT_POOL_SIZE,
-        CORE_DEFAULT_POOL_SIZE,
-        THREAD_POOL_MAX_KEEP_ALIVE_MS,
-        TimeUnit.MILLISECONDS,
-        LinkedBlockingDeque()
-    )
     private var shouldRecord = false
 
     constructor(
@@ -68,7 +55,6 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
         privacy: SessionReplayPrivacy,
         recordWriter: RecordWriter,
         timeProvider: TimeProvider,
-        recordCallback: RecordCallback = NoOpRecordCallback(),
         customMappers: List<MapperTypeWrapper> = emptyList(),
         customOptionSelectorDetectors: List<OptionSelectorDetector> = emptyList(),
         windowInspector: WindowInspector = WindowInspector
@@ -78,23 +64,25 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
             timeProvider
         )
 
+        val processor = RecordedDataProcessor(
+            recordWriter
+        )
+
         this.appContext = appContext
         this.rumContextProvider = rumContextProvider
         this.privacy = privacy
         this.recordWriter = recordWriter
         this.timeProvider = timeProvider
-        this.recordCallback = recordCallback
         this.customMappers = customMappers
         this.customOptionSelectorDetectors = customOptionSelectorDetectors
         this.windowInspector = windowInspector
-        this.processor = RecordedDataProcessor(
-            rumContextDataHandler,
-            processorExecutorService,
-            recordWriter,
-            recordCallback
+        this.recordedDataQueueHandler = RecordedDataQueueHandler(
+            processor = processor,
+            rumContextDataHandler = rumContextDataHandler,
+            timeProvider = timeProvider
         )
         this.viewOnDrawInterceptor = ViewOnDrawInterceptor(
-            processor,
+            recordedDataQueueHandler = recordedDataQueueHandler,
             SnapshotProducer(
                 TreeViewTraversal(customMappers + privacy.mappers()),
                 ComposedOptionSelectorDetector(
@@ -102,7 +90,11 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
                 )
             )
         )
-        this.windowCallbackInterceptor = WindowCallbackInterceptor(processor, viewOnDrawInterceptor, timeProvider)
+        this.windowCallbackInterceptor = WindowCallbackInterceptor(
+            recordedDataQueueHandler,
+            viewOnDrawInterceptor,
+            timeProvider
+        )
         this.sessionReplayLifecycleCallback = SessionReplayLifecycleCallback(this)
         this.uiHandler = Handler(Looper.getMainLooper())
     }
@@ -114,14 +106,13 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
         privacy: SessionReplayPrivacy,
         recordWriter: RecordWriter,
         timeProvider: TimeProvider,
-        recordCallback: RecordCallback = NoOpRecordCallback(),
         customMappers: List<MapperTypeWrapper> = emptyList(),
         customOptionSelectorDetectors: List<OptionSelectorDetector>,
         windowInspector: WindowInspector = WindowInspector,
         windowCallbackInterceptor: WindowCallbackInterceptor,
         sessionReplayLifecycleCallback: LifecycleCallback,
         viewOnDrawInterceptor: ViewOnDrawInterceptor,
-        processor: RecordedDataProcessor,
+        recordedDataQueueHandler: RecordedDataQueueHandler,
         uiHandler: Handler
     ) {
         this.appContext = appContext
@@ -129,11 +120,10 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
         this.privacy = privacy
         this.recordWriter = recordWriter
         this.timeProvider = timeProvider
-        this.recordCallback = recordCallback
         this.customMappers = customMappers
         this.customOptionSelectorDetectors = customOptionSelectorDetectors
         this.windowInspector = windowInspector
-        this.processor = processor
+        this.recordedDataQueueHandler = recordedDataQueueHandler
         this.viewOnDrawInterceptor = viewOnDrawInterceptor
         this.windowCallbackInterceptor = windowCallbackInterceptor
         this.sessionReplayLifecycleCallback = sessionReplayLifecycleCallback
@@ -186,10 +176,5 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
             windowCallbackInterceptor.stopIntercepting(windows)
             viewOnDrawInterceptor.intercept(decorViews, appContext)
         }
-    }
-
-    companion object {
-        private val THREAD_POOL_MAX_KEEP_ALIVE_MS = TimeUnit.SECONDS.toMillis(5)
-        private const val CORE_DEFAULT_POOL_SIZE = 1 // Only one thread will be kept alive
     }
 }
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
@@ -1,0 +1,209 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.async
+
+import androidx.annotation.MainThread
+import androidx.annotation.VisibleForTesting
+import androidx.annotation.WorkerThread
+import com.datadog.android.sessionreplay.internal.processor.RecordedDataProcessor
+import com.datadog.android.sessionreplay.internal.processor.RumContextDataHandler
+import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
+import com.datadog.android.sessionreplay.internal.utils.TimeProvider
+import com.datadog.android.sessionreplay.model.MobileSegment
+import java.lang.ClassCastException
+import java.lang.NullPointerException
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.LinkedBlockingDeque
+import java.util.concurrent.RejectedExecutionException
+import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.TimeUnit
+
+/**
+ * Responsible for storing the Snapshot and Interaction events in a queue.
+ * Allows for asynchronous enrichment, which still preserving the event order.
+ * The items are added to the queue from the main thread and processed on a background thread.
+ */
+internal class RecordedDataQueueHandler {
+
+    private var executorService: ExecutorService
+    private var processor: RecordedDataProcessor
+    private var rumContextDataHandler: RumContextDataHandler
+    private var timeProvider: TimeProvider
+
+    internal constructor(
+        processor: RecordedDataProcessor,
+        rumContextDataHandler: RumContextDataHandler,
+        timeProvider: TimeProvider
+    ) : this(
+        processor = processor,
+        rumContextDataHandler = rumContextDataHandler,
+        timeProvider = timeProvider,
+
+        /**
+         * TODO: RUMM-0000 consider change to LoggingThreadPoolExecutor once V2 is merged.
+         * if we ever decide to make the poolsize greater than 1, we need to ensure
+         * synchronization works correctly in the triggerProcessingLoop method below
+         */
+        executorService = ThreadPoolExecutor(
+            CORE_DEFAULT_POOL_SIZE,
+            CORE_DEFAULT_POOL_SIZE,
+            THREAD_POOL_MAX_KEEP_ALIVE_MS,
+            TimeUnit.MILLISECONDS,
+            LinkedBlockingDeque()
+        )
+    )
+
+    @VisibleForTesting
+    internal constructor(
+        processor: RecordedDataProcessor,
+        rumContextDataHandler: RumContextDataHandler,
+        timeProvider: TimeProvider,
+        executorService: ExecutorService
+    ) {
+        this.processor = processor
+        this.rumContextDataHandler = rumContextDataHandler
+        this.executorService = executorService
+        this.timeProvider = timeProvider
+    }
+
+    // region internal
+    internal val recordedDataQueue = ConcurrentLinkedQueue<RecordedDataQueueItem>()
+
+    @MainThread
+    internal fun addTouchEventItem(
+        pointerInteractions: List<MobileSegment.MobileRecord>
+    ): TouchEventRecordedDataQueueItem? {
+        val rumContextData = rumContextDataHandler.createRumContextData()
+            ?: return null
+
+        val item = TouchEventRecordedDataQueueItem(
+            rumContextData = rumContextData,
+            touchData = pointerInteractions
+        )
+
+        insertIntoRecordedDataQueue(item)
+
+        return item
+    }
+
+    @MainThread
+    internal fun addSnapshotItem(
+        systemInformation: SystemInformation
+    ): SnapshotRecordedDataQueueItem? {
+        val rumContextData = rumContextDataHandler.createRumContextData()
+            ?: return null
+
+        val item = SnapshotRecordedDataQueueItem(
+            rumContextData = rumContextData,
+            systemInformation = systemInformation
+        )
+
+        insertIntoRecordedDataQueue(item)
+
+        return item
+    }
+
+    /**
+     * Goes through the queue one item at a time for as long as there are items in the queue.
+     * If an item is ready to be consumed, it is processed.
+     * If an invalid item is encountered, it is removed (invalid items are possible
+     * for example if a snapshot failed to traverse the tree).
+     * If neither of the previous conditions occurs, the loop breaks.
+     */
+    @MainThread
+    internal fun tryToConsumeItems() {
+        // no need to create a thread if the queue is empty
+        if (recordedDataQueue.isEmpty()) {
+            return
+        }
+
+        // currentTime needs to be obtained on the uithread
+        val currentTime = timeProvider.getDeviceTimestamp()
+
+        @Suppress("SwallowedException", "TooGenericExceptionCaught")
+        try {
+            executorService.execute {
+                triggerProcessingLoop(currentTime)
+            }
+        } catch (e: RejectedExecutionException) {
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+        } catch (e: NullPointerException) {
+            // in theory will never happen but we'll log it
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+        }
+    }
+
+    /**
+     * The threadpool that triggers this method has a size/maxsize of 1.
+     * This makes it implicitly synchronised so we should never have any multithreading issues
+     * where one thread peeks while another removes the element.
+     */
+    @WorkerThread
+    @Synchronized
+    private fun triggerProcessingLoop(currentTime: Long) {
+        while (recordedDataQueue.isNotEmpty()) {
+            val nextItem = recordedDataQueue.peek()
+
+            if (nextItem != null) {
+                if (shouldRemoveItem(nextItem, currentTime)) {
+                    recordedDataQueue.poll()
+                } else if (nextItem.isReady()) {
+                    processItem(recordedDataQueue.poll())
+                } else {
+                    break
+                }
+            }
+        }
+    }
+
+    private fun processItem(nextItem: RecordedDataQueueItem?) {
+        when (nextItem) {
+            is SnapshotRecordedDataQueueItem ->
+                processSnapshotEvent(nextItem)
+            is TouchEventRecordedDataQueueItem ->
+                processTouchEvent(nextItem)
+        }
+    }
+
+    private fun processSnapshotEvent(item: SnapshotRecordedDataQueueItem) {
+        processor.processScreenSnapshots(item)
+    }
+
+    private fun processTouchEvent(item: TouchEventRecordedDataQueueItem) {
+        processor.processTouchEventsRecords(item)
+    }
+
+    private fun shouldRemoveItem(recordedDataQueueItem: RecordedDataQueueItem, currentTime: Long) =
+        !recordedDataQueueItem.isValid() || isTooOld(currentTime, recordedDataQueueItem)
+
+    private fun isTooOld(currentTime: Long, recordedDataQueueItem: RecordedDataQueueItem): Boolean =
+        (currentTime - recordedDataQueueItem.rumContextData.timestamp) > MAX_DELAY_MS
+
+    private fun insertIntoRecordedDataQueue(recordedDataQueueItem: RecordedDataQueueItem) {
+        @Suppress("SwallowedException", "TooGenericExceptionCaught")
+        try {
+            recordedDataQueue.offer(recordedDataQueueItem)
+        } catch (e: IllegalArgumentException) {
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+        } catch (e: ClassCastException) {
+            // in theory will never happen but we'll log it
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+        } catch (e: NullPointerException) {
+            // in theory will never happen but we'll log it
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+        }
+    }
+
+    // end region
+
+    internal companion object {
+        internal const val MAX_DELAY_MS = 200L
+        private val THREAD_POOL_MAX_KEEP_ALIVE_MS = TimeUnit.SECONDS.toMillis(5)
+        private const val CORE_DEFAULT_POOL_SIZE = 1 // Only one thread will be kept alive
+    }
+}
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/SnapshotRecordedDataQueueItem.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/SnapshotRecordedDataQueueItem.kt
@@ -1,0 +1,31 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.async
+
+import com.datadog.android.sessionreplay.internal.processor.RumContextData
+import com.datadog.android.sessionreplay.internal.recorder.Node
+import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
+import java.util.concurrent.atomic.AtomicInteger
+
+internal class SnapshotRecordedDataQueueItem(
+    rumContextData: RumContextData,
+    internal val systemInformation: SystemInformation
+) : RecordedDataQueueItem(rumContextData) {
+    internal var nodes = emptyList<Node>()
+    internal var pendingImages = AtomicInteger(0)
+
+    override fun isValid(): Boolean {
+        return nodes.isNotEmpty()
+    }
+
+    override fun isReady(): Boolean {
+        return pendingImages.get() == 0
+    }
+
+    internal fun incrementPendingImages() = pendingImages.incrementAndGet()
+    internal fun decrementPendingImages() = pendingImages.decrementAndGet()
+}
--- a/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/TouchEventRecordedDataQueueItem.kt
+++ b/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/TouchEventRecordedDataQueueItem.kt
@@ -1,0 +1,24 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.async
+
+import com.datadog.android.sessionreplay.internal.processor.RumContextData
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal class TouchEventRecordedDataQueueItem(
+    rumContextData: RumContextData,
+    internal val touchData: List<MobileSegment.MobileRecord> = emptyList()
+) : RecordedDataQueueItem(rumContextData) {
+
+    override fun isValid(): Boolean {
+        return touchData.isNotEmpty()
+    }
+
+    override fun isReady(): Boolean {
+        return true
+    }
+}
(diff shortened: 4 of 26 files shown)


library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
```diff
@@ -0,0 +1,135 @@
...
+                systemInformation = blockingQueueItem.systemInformation
+            )
+        }
+    }
+
+    private fun removeInvalidItemsFromHead(currentTime: Long) {
+        while (workQueue.isNotEmpty()) {
+            val blockingQueueItem = workQueue.peek()
+
+            if (shouldRemoveItem(blockingQueueItem, currentTime)) {
```
mariusc83, line 107: why do you break if you should remove the item ?

mariusc83 (changes requested):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            workQueue.poll()
+        }
+    }
+
+    private fun shouldRemoveItem(blockingQueueItem: BlockingQueueItem?, currentTime: Long) =
+        blockingQueueItem == null ||
+            (
+                blockingQueueItem.isValid() &&
```
mariusc83, line 118: shouldn't be here `isNotValid && isTooOld`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
jonathanmos: Good catch. I flipped the method logic due to detekt rules and forgot to rename

mariusc83 (changes requested):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
jonathanmos: originally there was one loop that did exactly that, but I had issues with detekt and tried to split it up. I'll do some refactor and commenting

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
```diff
@@ -187,6 +187,16 @@ internal class RecordedDataProcessor(
     ): Runnable? {
         val rumContextData = rumContextDataHandler.createRumContextData() ?: return null
 
+        updateViewSent(rumContextData.newRumContext)
+
+        return runnableFactory(
```
mariusc83, line 192: the whole idea about runnable here was to be able to move some operations outside of the `WorkerThread` like the timestamp and rum context creations. Right now all these operations are in your `BlockingQueueHandler` and when you reach this method you are already in the `WorkerThread`. You will not need any runnable here and we should remove this part to simplify this class.

mariusc83 (changes requested):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
jonathanmos: doesn't timestamp need to be on the uithread to avoid skew?

jonathanmos: actually this part is only relevant to the touch actions and doesn't relate to the snapshots so it isn't going through the blockinghandler and still follows the original logic (with the exception that runcontextdata has been extracted to its own class)

mariusc83: Hmm what timestamp ? The way I see it you are passing the `timestamp` from the `BlockingQueueItem` which was added in the `MainThread`. When you reach the `RecordedDataProcessor#processScreenSnapshots` you are already in the `WorkerThread` as you call this from your executor in the `BlockingQueueHandler` so everything you do in this class is already in the `WorkerThread`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
mariusc83: We should always favour optimisation to `detekt`. Detekt rules can be suppressed if they alter optimisation.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
mariusc83: I see, well this part is a bit worrying as we are mixing things a bit here. Actually if you delay the snapshots but execute the touch actions you will have issues in the player. Also we are mixing a bit too much things in here. Instead I would assess how we can add the touch actions into the blocking queue also and then have the processor only working with the queue. Will allow you to remove a lot of complexity from here and get rid of executor service and runnables.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
```diff
@@ -48,7 +49,7 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
     private val windowInspector: WindowInspector
     private val windowCallbackInterceptor: WindowCallbackInterceptor
     private val sessionReplayLifecycleCallback: LifecycleCallback
-    private val processor: RecordedDataProcessor
+    private val blockingQueueHandler: BlockingQueueHandler
```
0xnm, line 52: minor nit: probably we don't need word `blocking` in the `blockingQueueExecutorService` and `blockingQueueHandler` names, because what matters is that it is queue, and the actual implementation doesn't really matter.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
```diff
@@ -60,6 +61,15 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
         TimeUnit.MILLISECONDS,
         LinkedBlockingDeque()
     )
+
+    private val blockingQueueExecutorService = ThreadPoolExecutor(
```
0xnm, line 65: I would rather use `LoggingThreadPoolExecutor`. The thing is that if task submitted to `ThreadPoolExecutor` throws and we don't unwrap the resulting `Future`, exception will be lost. While with `LoggingThreadPoolExecutor` it will be logged.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
```diff
@@ -0,0 +1,128 @@
...
+
+internal class BlockingQueueHandler(
+    private val processor: RecordedDataProcessor,
+    private val rumContextDataHandler: RumContextDataHandler,
+    private val executorService: ExecutorService,
+    private val timeProvider: TimeProvider
+) {
+    // region internal
+
+    internal var workQueue = LinkedBlockingQueue<BlockingQueueItem>()
```
0xnm, line 29: does it need to be `var`? I don't see any other assignment. Probably can be just `val`.

```suggestion
    internal val workQueue = LinkedBlockingQueue<BlockingQueueItem>()
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
```diff
@@ -0,0 +1,128 @@
...
+        } catch (e: NullPointerException) {
+            // in theory will never happen but we'll log it
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+        }
+
+        return blockingQueueItem
+    }
+
+    @MainThread
+    internal fun update() {
```
0xnm, line 60: I think this method needs some description (or maybe rename it?), it is not obvious what is its purpose and what it is doing.

0xnm (commented):
Happy to see the improvement in the processing! I gave a light review without focusing on the logic itself.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
mariusc83, line 29: I actually wonder if you need a `LinkedBlockingQueue` here if you don't have an infinite `while` loop in the consumer thread ? The whole point of a blocking queue is to wait/block when there's nothing to read and notify when there's something new added.

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandlerTest.kt
```diff
@@ -0,0 +1,243 @@
...
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Future
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(ForgeConfigurator::class)
+internal class BlockingQueueHandlerTest {
```
0xnm, line 44: it makes sense also to add some multi-threaded tests here

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
jonathanmos: I've added a todo to change to `LoggingThreadPoolExecutor` once V2 is merged

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
```diff
@@ -0,0 +1,128 @@
...
+        while (workQueue.isNotEmpty()) {
+            val blockingQueueItem = workQueue.peek()
+
+            if (blockingQueueItem != null &&
+                (
+                    !shouldRemoveItem(blockingQueueItem, currentTime) &&
+                        blockingQueueItem.isReady()
+                    )
+            ) {
+                workQueue.poll()
```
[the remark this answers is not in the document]
jonathanmos: I'll make `wakeUpProcessor` synchronised to avoid race conditions

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
jonathanmos: hmm you're absolutely right that blocking isn't adding anything meaningful here. I'll change the implementation to LinkedList and do a bit of renaming

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
jonathanmos: I'm also renaming all blockingQueue classes/items to make the names less connected to the implementation

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandler.kt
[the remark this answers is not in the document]
mariusc83: I would first try and see if the blocking queue already handles this internally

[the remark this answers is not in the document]
mariusc83: also why not using `peek` and `remove(item)` ...this way you can avoid doing a `poll` accidentally if the `peek` is empty. Also I would add the condition first like:
```
if(item!=null){
     if(shouldRemoveItem(item)){
        queue.remove(item)
     } else if(item.isReady()){
        queue.remove(item)
        processor.processScreenSnapshots(...)
     } else {
        break
     }
   } else {
     break
   }
   ```
   And theoretically if you have the condition `while(queue.isNotEmpty)` the `item` should not be null anyway so you could drop that if you synchronise on the `blockingQueue`

File: library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt (hunk 3 above)
Comment (jonathanmos) [reply library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt]: I've added touch events to the flow

File: library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandlerTest.kt (hunk 8 above)
Comment (jonathanmos) [reply library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/BlockingQueueHandlerTest.kt]: I've now added some

File: library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt [hunk 10]
```diff
@@ -0,0 +1,177 @@
...
+        } catch (e: NullPointerException) {
+            // in theory will never happen but we'll log it
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+        }
+    }
+
+    // end region
+
+    internal companion object {
+        internal const val MAX_DELAY_MS = 200L
```
mariusc83, line 175: are you sure you are able to load all the bitmaps in `200ms` ? I guess you will see once you add that logic and you can play with this number.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
```diff
@@ -165,37 +159,6 @@ internal class RecordedDataProcessor(
...
-
-        return runnableFactory(
-            rumContextData.timestamp,
-            rumContextData.newRumContext,
-            rumContextData.prevRumContext
-        )
-    }
-
     private fun updateViewSent(rumContext: SessionReplayRumContext) {
         // Because the runnable will be executed in another thread it can happen in case there is
```
mariusc83, line 163: THis comment needs to be adjusted now -> `because this code is executed in the worker thread...`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
```diff
@@ -123,7 +123,15 @@ internal class RecorderWindowCallback(
         if (pointerInteractions.isEmpty()) {
             return
         }
-        processor.processTouchEventsRecords(ArrayList(pointerInteractions))
+
+        val item = recordedDataQueueHandler.addTouchEventItem(
+            ArrayList(pointerInteractions)
+        ) ?: return
+
+        if (item.isReady()) {
```
mariusc83, line 131: why do you check here ? I think this value is `true` by default.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/listener/WindowsOnDrawListener.kt
```diff
@@ -50,8 +51,10 @@ internal class WindowsOnDrawListener(
             }
 
         if (nodes.isNotEmpty()) {
-            blockingQueueItem.nodes = nodes
-            blockingQueueHandler.update()
+            item.nodes = nodes
+            if (item.isReady()) {
```
mariusc83, line 55: also why do you check here if `item.isReady()` if we already check it when we loop through the queue ?

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandlerTest.kt
```diff
@@ -0,0 +1,455 @@
...
+            mockRumContextDataHandler,
+            mockExecutorService,
+            mockTimeProvider
+        )
+    }
+
+    @Test
+    fun `M use executorService W tryToConsumeItems() { queue has items }`() {
+        // Given
+        testedHandler.recordedDataQueue.offer(fakeSnapshotQueueItem)
```
mariusc83, line 108: why not using the public methods here ? 
`addSnapshotItem` and `addTouchEventItem` ?

mariusc83: I think you need a test for each

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandlerTest.kt
```diff
@@ -0,0 +1,455 @@
...
+            nodes = spy.nodes,
+            systemInformation = spy.systemInformation,
+            newContext = spy.newRumContext,
+            prevContext = spy.prevRumContext,
+            timestamp = spy.timestamp
+        )
+    }
+
+    @Test
+    fun `M call processor W tryToConsumeItems() { valid Touch Event item }`() {
```
mariusc83, line 309: I understand that this way you can assert the actual item in the `mockProcessor` but I would also find useful to test this logic when calling the actual public methods:
`addSnapshotItem` or `addTouchEventItem` even if you cannot assert the actual item you could test that: 
```
verify(mockProcessor).processTouchEventRecords(argumentCaptor<SessionReplayContext>.capture(), argumentCaptor<List<MobileSegment.MobileRecord>>().capture())
```
And then you can assert the value in those `argumentCaptors`.
Same thing for the above test for `processScreenSnapshots`

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+     * Goes through the queue one item at a time for as long as there are items in the queue.
+     * If an item is ready to be consumed, it is processed.
+     * If an invalid item is encountered, it is removed (invalid items are possible
+     * for example if a snapshot failed to traverse the tree).
+     * If neither of the previous conditions occurs, the loop breaks.
+     */
+    @MainThread
+    internal fun tryToConsumeItems() {
+        // no need to create a thread if the queue is empty
+        if (recordedDataQueue.isEmpty()) {
```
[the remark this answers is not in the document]
0xnm: Ok, you can ignore this comment. I checked the implementation in the Android codebase and it seems we are good here, because `isEmpty` implementation for the `ConcurrentLinkedQueue` [relies on the `first/peek` call](https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/java/util/concurrent/ConcurrentLinkedQueue.java;l=446;drc=260545b78f2f71d32f1208c90f77220729591f7b).

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
jonathanmos: This was for future proofing. My assumption was that if in future the isReady has any sort of logic it would be better to already wrap this so that we don't forget to do so later on

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    @WorkerThread
+    private fun triggerProcessingLoop(currentTime: Long) {
+        while (recordedDataQueue.isNotEmpty()) {
+            val nextItem = recordedDataQueue.peek()
+
+            if (nextItem != null) {
+                if (shouldRemoveItem(nextItem, currentTime)) {
+                    recordedDataQueue.poll()
```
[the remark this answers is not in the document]
mariusc83: that's the reason I suggested to use `remove(item)` instead of `poll` and on top of that I also agree with the synchronisation.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/SnapshotRecordedDataQueueItem.kt
```diff
@@ -0,0 +1,33 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+internal class SnapshotRecordedDataQueueItem(
+    timestamp: Long,
+    prevRumContext: SessionReplayRumContext,
+    newRumContext: SessionReplayRumContext,
+    internal var systemInformation: SystemInformation
+) : RecordedDataQueueItem(timestamp, prevRumContext, newRumContext) {
+    internal var nodes = emptyList<Node>()
+    internal var pendingImages = AtomicInteger(0)
```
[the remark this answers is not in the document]
jonathanmos: You're absolutely right. I'm refactoring and moving these

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+                } else if (nextItem.isReady()) {
+                    processItem()
```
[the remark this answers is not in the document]
jonathanmos: same as prev - I think synchronising should solve these edge cases.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
```diff
@@ -48,18 +49,19 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
...
     private val windowCallbackInterceptor: WindowCallbackInterceptor
     private val sessionReplayLifecycleCallback: LifecycleCallback
-    private val processor: RecordedDataProcessor
+    private val recordedDataQueueHandler: RecordedDataQueueHandler
     private val viewOnDrawInterceptor: ViewOnDrawInterceptor
     private val uiHandler: Handler
 
-    @Suppress("UnsafeThirdPartyFunctionCall") // workQueue can't be null
-    private val processorExecutorService = ThreadPoolExecutor(
+    // TODO: change to LoggingThreadPoolExecutor once V2 is merged
```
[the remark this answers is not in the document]
maciejburda: maybe `(.*)/sessionreplay/` is excluded from `detekt`?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+    internal fun addTouchEventItem(
+        pointerInteractions: List<MobileSegment.MobileRecord>
+    ): TouchEventRecordedDataQueueItem? {
+        val rumContextData = rumContextDataHandler.createRumContextData()
+            ?: return null
+
+        val item = TouchEventRecordedDataQueueItem(
+            timestamp = rumContextData.timestamp,
+            prevRumContext = rumContextData.prevRumContext,
+            newRumContext = rumContextData.newRumContext,
```
maciejburda, line 47: minor: maybe it could leverage convenience constructor that just takes rumContextData and touchData as a parameters

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+    internal fun addSnapshotItem(
+        systemInformation: SystemInformation
+    ): SnapshotRecordedDataQueueItem? {
+        val rumContextData = rumContextDataHandler.createRumContextData()
+            ?: return null
+
+        val item = SnapshotRecordedDataQueueItem(
+            timestamp = rumContextData.timestamp,
+            prevRumContext = rumContextData.prevRumContext,
+            newRumContext = rumContextData.newRumContext,
```
maciejburda, line 66: minor: maybe it could leverage convenience constructor that just takes rumContextData and systemInformation as a parameters

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+        when (val item = recordedDataQueue.poll()) {
+            is SnapshotRecordedDataQueueItem ->
+                processSnapshotEvent(item)
+            is TouchEventRecordedDataQueueItem ->
+                processTouchEvent(item)
+        }
+    }
+
+    private fun processSnapshotEvent(item: SnapshotRecordedDataQueueItem) {
+        processor.processScreenSnapshots(
```
maciejburda, line 133: probably could just take item as a parameter

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+        processor.processTouchEventsRecords(
+            newContext = item.newRumContext,
+            touchEventsRecords = item.touchData
+        )
+    }
+
+    private fun shouldRemoveItem(recordedDataQueueItem: RecordedDataQueueItem, currentTime: Long) =
+        !recordedDataQueueItem.isValid() || isTooOld(currentTime, recordedDataQueueItem)
+
+    private fun isTooOld(currentTime: Long, recordedDataQueueItem: RecordedDataQueueItem): Boolean =
```
maciejburda, line 152: I'm wondering if we should ever consider item as "too old"

Should we process all of them regardless the time? What the strategy here?

maciejburda (commented):
Nice work! Added some minor comments 🙌

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
jonathanmos: Yes, this is just an arbitrary assumption. We'll need to test on an image-rich application to figure out what the threshold should be

library/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandlerTest.kt
jonathanmos: I've now added tests for each and moved to using the internal methods

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/TouchEventRecordedDataQueueItem.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import com.datadog.android.sessionreplay.internal.utils.SessionReplayRumContext
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal class TouchEventRecordedDataQueueItem(
+    timestamp: Long,
+    prevRumContext: SessionReplayRumContext,
+    newRumContext: SessionReplayRumContext
+) : RecordedDataQueueItem(timestamp, prevRumContext, newRumContext) {
+    internal var touchData = emptyList<MobileSegment.MobileRecord>()
```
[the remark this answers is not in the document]
jonathanmos: I'm keeping touchData internal for the unit tests, but otherwise I've made it val and moved it inside the constructor

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
jonathanmos: I like this suggestion. I'll refactor

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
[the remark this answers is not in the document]
0xnm: if method is synchronised - yes. But instead we can put a synchronised lock only on the parts of the method which modify data and do a double check (inside and outside synchronised block), don't know which one will be better. I guess synchronising the whole method is not needed, since we are loosing the efficiency of the parallelism for the parts where queue is not modified.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
jonathanmos: The thing is that if we don't define a threshold for an item to be too old then an item that for whatever reason had a failure resolving it's bitmaps (pendingImages > 0) could get stuck at the head of the queue forever

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
```diff
@@ -165,28 +157,7 @@ internal class RecordedDataProcessor(
...
-    private fun buildRunnable(
-        runnableFactory: (
-            timestamp: Long,
-            newContext: SessionReplayRumContext,
-            prevRumContext: SessionReplayRumContext
-        ) -> Runnable
-    ): Runnable? {
-        val rumContextData = rumContextDataHandler.createRumContextData() ?: return null
-
+    private fun updateViewSent(rumContext: SessionReplayRumContext) {
```
mariusc83, line 160: Looks good but we missed something. This method was before called from `MainThread` intentionally because we want to update the `Rum` context in time to mark the `has_replay` on the `RumViewEvent` in progress. Right now all these methods are  called in the `WorkerThread` so we will have to move this call from here and mainly the `recordCallback` somewhere else. I'd say a good moment will be when we put the item in the queue, probably in the `QueueHandler`

mariusc83 (changes requested):

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
jonathanmos: When we initially put the item in the queue we don't know for sure that the traversal will succeed and we will have nodes. What about putting it right after the end of the traversal where we set the nodes, before the call to awakening the queue?

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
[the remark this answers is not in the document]
jonathanmos: actually we can be sure that we can never encounter this sort of issue, because the threadpool has a size/maxsize of 1 so there will never be more than one thread executing at any one time. In fact, we don't even need to synchronise the method because this makes it implicitly synchronised

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
jonathanmos: As we discussed, I've moved recordCallback inside the writer

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,166 @@
...
+    @WorkerThread
+    private fun triggerProcessingLoop(currentTime: Long) {
+        while (recordedDataQueue.isNotEmpty()) {
+            val nextItem = recordedDataQueue.peek()
+
+            if (nextItem != null) {
+                if (shouldRemoveItem(nextItem, currentTime)) {
+                    recordedDataQueue.poll()
+                } else if (nextItem.isReady()) {
+                    processItem()
```
mariusc83, line 115: I would call `processItem(recordDataQueue.poll())` here just for visibility.

mariusc83 (approved):
Nice job, it looks good to me now, I left one more comment regarding visibility in a function signature

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayRecorder.kt
```diff
@@ -42,24 +41,24 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
...
     private val windowCallbackInterceptor: WindowCallbackInterceptor
     private val sessionReplayLifecycleCallback: LifecycleCallback
-    private val processor: RecordedDataProcessor
+    private val recordedDataQueueHandler: RecordedDataQueueHandler
     private val viewOnDrawInterceptor: ViewOnDrawInterceptor
     private val uiHandler: Handler
 
-    @Suppress("UnsafeThirdPartyFunctionCall") // workQueue can't be null
-    private val processorExecutorService = ThreadPoolExecutor(
+    // TODO: RUMM-0000 change to LoggingThreadPoolExecutor once V2 is merged
```
maciejburda, line 53: probably a good idea to log actual ticket

maciejburda (approved):
LGTM

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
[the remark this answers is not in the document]
xgouchet: > the threadpool has a size/maxsize of 1 so there will never be more than one thread executing at any one time

Technically, the maximum pool size can be increased with a simple `ThreadPoolExecutor.setMaximumPoolSize()`. Also from this class's point of view, this method is triggered by an `ExecutorService`, so we have no guarantee of what the actual implementation is going to be.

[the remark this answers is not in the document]
jonathanmos: That's a good point. I've discussed this with Marius, and what I'm going to do is to create two constructors for the class - a regular constructor where we internally create the executorService, and another constructor visible only for testing where we pass in the executor. In this way we can be fairly sure that we encapsulate the executor and that we're only ever using a single thread, and avoid the added overhead of synchronising the method

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -0,0 +1,204 @@
...
+
+        // TODO: RUMM-0000 change to LoggingThreadPoolExecutor once V2 is merged
+        this.executorService = ThreadPoolExecutor(
+            CORE_DEFAULT_POOL_SIZE,
+            CORE_DEFAULT_POOL_SIZE,
+            THREAD_POOL_MAX_KEEP_ALIVE_MS,
+            TimeUnit.MILLISECONDS,
+            LinkedBlockingDeque()
+        )
+    }
```
0xnm, line 55: minor: probably for simplification this one can call another one:

```suggestion
    ): this(
    processor,
      rumContextDataHandler,
      timeProvider,
      ThreadPoolExecutor(
            CORE_DEFAULT_POOL_SIZE,
            CORE_DEFAULT_POOL_SIZE,
            THREAD_POOL_MAX_KEEP_ALIVE_MS,
            TimeUnit.MILLISECONDS,
            LinkedBlockingDeque()
        )
      )
```

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriter.kt
```diff
@@ -7,16 +7,21 @@
...
+internal class SessionReplayRecordWriter(
+    private val sdkCore: SdkCore,
+    private val recordCallback: RecordCallback
+) : RecordWriter {
     private var lastRumContextId: String = ""
     override fun write(record: EnrichedRecord) {
         val forceNewBatch = resolveForceNewBatch(record)
         sdkCore.getFeature(SessionReplayFeature.SESSION_REPLAY_FEATURE_NAME)
             ?.withWriteContext(forceNewBatch) { _, eventBatchWriter ->
+                updateViewSent(record)
```
0xnm, line 24: probably it better to place this call below in the following way:

```
if (eventBatchWriter.write(serializedRecord, null)) {
  updateViewSent(record)
}
```
that way it will be invoked only if record was really written. Otherwise, I guess, you may have some discrepancy if write operation failed.

library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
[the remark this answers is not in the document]
0xnm: I guess the concern here is that if at any point someone does the change to the max number of threads in the pool by changing the constant or by calling the `setMaximumPoolSize` things will start to be buggy and will be tricky to catch the root cause.

This algorithm is working fine if there is only a single thread accessing the queue at the time. I'm okay with that if we put some checks/restrictions to make sure that will always be the case. Otherwise we need to change the way we are working with the queue.

For the moment, to make sure that this piece of code will be fine in any scenario, you can just make top-level method `synchronised`, but for sure if we at some point allow multiple threads to work with the queue, better locking approach should be developed.

PR #1482, opened 2023-06-19
RUMM-3021 internal logger 1/2

xgouchet:
### What does this PR do?

Update the usage of the Internal Logger to adjust level, target(s) and/or message.

Next PR will add a lazy logging API


dd-sdk-android-core/src/test/kotlin/com/datadog/android/v2/core/DatadogCoreTest.kt
```diff
@@ -299,13 +300,14 @@ internal class DatadogCoreTest {
     ) {
         // Given
         val fakeReceiver = FeatureEventReceiver { }
+        reset(mockInternalLogger)
```
0xnm, line 303: why do we need to reset it?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -485,6 +486,7 @@ internal class RumViewScopeTest {
             type = expectedViewType,
             trackFrustrations = fakeTrackFrustrations
         )
+        reset(mockInternalLogger)
```
0xnm, line 489: why do we need a `reset` call here as well?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/api/InternalLogger.kt
```diff
@@ -11,6 +11,29 @@ import com.datadog.tools.annotation.NoOpImplementation
...
+ *         a feature from working, or for an issue resulting in unexpected data loss;
+ *     - Level.WARN: to inform of an actionable misconfiguration or missuses of the SDK, resulting
+ *         in delayed or incomplete data;
+ *     - Level.INFO: information about important expected event (e.g.: successful upload);
+ * - Target.TELEMETRY: any event that need to be tracked for usage monitoring or for error
+ *     diagnostic.
+ *     - Level.ERROR, Level.WARN: for any critical error that is unexpected enough and actionable;
+ *     - Level.INFO, Level.DEBUG, Level.VERBOSE: important information about critical parts of the
+ *         SDK we want to monitor;
+ * - Target.MAINTAINER: can be anything relevant about the moving parts of the core SDK or any
```
0xnm, line 29: I was always curious if we really need `MAINTAINER` target, but I guess even if we remove it, some errors which are for `MAINTAINER` target are not really actionable on the user side if logged with `USER` target. But at the same time some things logged to `MAINTAINER` can be also useful to see on the `USER` target, so that they give more context for the debugging.

Anyway, this is just some thoughts, we can keep it for now.

0xnm (commented):
Nice work! I added some questions to clarify.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/internal/net/DataOkHttpUploader.kt
```diff
@@ -151,7 +145,10 @@ internal class DataOkHttpUploader(
         return c == '\t' || c in '\u0020' until '\u007F'
     }
 
-    private fun responseCodeToUploadStatus(code: Int): UploadStatus {
+    private fun responseCodeToUploadStatus(
+        code: Int,
+        request: com.datadog.android.v2.api.Request
```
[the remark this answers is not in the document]
xgouchet: Yes that's the reason why. Just noticed there was an alias there, I'll use that

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/RuntimeUtils.kt
```diff
@@ -7,52 +7,6 @@
...
 internal var unboundInternalLogger: InternalLogger = InternalLogger.UNBOUND
-
-/**
- * Warns the user that they're using a deprecated feature.
- * @param target the target feature (e.g. method name)
- * @param deprecatedSince the version when the feature was deprecated
- * @param removedInVersion the version in which the feature will disappear
- * @param alternative an alternative option to get the same effect
- */
-internal fun warnDeprecated(
```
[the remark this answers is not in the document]
xgouchet: True, but then we might want to use a specific internal logger instead of the unbound one, so my thought was we can create a helper method then if need be.
One that can also warn us of usage of deprecated features to know when we can actually remove them

dd-sdk-android-core/src/test/kotlin/com/datadog/android/v2/core/DatadogCoreTest.kt
xgouchet: There was some flakiness because some steps during the setup could send logs to the internal logger

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
xgouchet: Same thing, some flakiness rose up during the setup

PR #1484, opened 2023-06-20
RUMM-3351: Provide default value of SDK instance in public APIs

0xnm:
### What does this PR do?

This PR makes all the methods of public API to have a default value of SDK instance if not provided explicitly.

`JvmOverrides` annotation is added to such methods/constructors only for sites which are supposed to be called from Java, meaning it is not added to the extension functions (while technically it is possible to call them, unlikely it will be the case) and to the code in `-compose` and `-ktx` modules.

Code in the sample app was updated to reflect that it is not needed to pass SDK instance to such methods, code in the nightly tests and integration tests was left untouched mostly.

More PRs regarding public API changes to follow.


mariusc83 (approved):
Nice, good call ;)

PR #1485, opened 2023-06-20
RUMM-3021 internal logger only once

xgouchet:
### What does this PR do?

Add an optional setting to send an internal log only once. This can be used to limit the spam in the Logcat/Telemetry.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/api/InternalLogger.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/api/InternalLogger.kt
@@ -82,7 +82,7 @@ interface InternalLogger {
      * @param message the log message
      * @param throwable an optional throwable error
      * @param onlyOnce whether only one instance of the message should be sent per lifetime of the
-     * logger (default is `false`)
+     * logger (default is `false`, onlyOnce applies to each target independently)
      */
     fun log(
         level: Level,
@@ -96,7 +96,7 @@ interface InternalLogger {
 
         /**
          * Logger for the cases when SDK instance is not yet available. Try to use the logger
-         * provided by [SdkCore._internalLogger] instead if possible.
+         * provided by [FeatureSdkCore.internalLogger] instead if possible.
          */
         val UNBOUND: InternalLogger = SdkInternalLogger(null)
     }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/api/InternalLogger.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/api/InternalLogger.kt
@@ -64,12 +64,15 @@ interface InternalLogger {
      * @param target the target handler for the log
      * @param message the log message
      * @param throwable an optional throwable error
+     * @param onlyOnce whether only one instance of the message should be sent per lifetime of the
+     * logger (default is `false`)
      */
     fun log(
         level: Level,
         target: Target,
         message: String,
-        throwable: Throwable? = null
+        throwable: Throwable? = null,
+        onlyOnce: Boolean = false
     )
 
     /**
@@ -78,12 +81,15 @@ interface InternalLogger {
      * @param targets list of the target handlers for the log
      * @param message the log message
      * @param throwable an optional throwable error
+     * @param onlyOnce whether only one instance of the message should be sent per lifetime of the
+     * logger (default is `false`)
      */
     fun log(
         level: Level,
         targets: List<Target>,
         message: String,
-        throwable: Throwable? = null
+        throwable: Throwable? = null,
+        onlyOnce: Boolean = false
     )
 
     companion object {
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
@@ -42,9 +42,9 @@ internal class SdkInternalLogger(
      */
     internal val maintainerLogger = maintainerLogHandlerFactory.invoke()
 
-    private val onlyOnceUserMessages = mutableListOf<String>()
-    private val onlyOnceMaintainerMessages = mutableListOf<String>()
-    private val onlyOnceTelemetryMessages = mutableListOf<String>()
+    private val onlyOnceUserMessages = mutableSetOf<String>()
+    private val onlyOnceMaintainerMessages = mutableSetOf<String>()
+    private val onlyOnceTelemetryMessages = mutableSetOf<String>()
 
     // region InternalLogger
 
@@ -118,7 +118,7 @@ internal class SdkInternalLogger(
         message: String,
         error: Throwable?,
         onlyOnce: Boolean,
-        knownSingleMessages: MutableList<String>
+        knownSingleMessages: MutableSet<String>
     ) {
         if (onlyOnce) {
             if (knownSingleMessages.contains(message)) {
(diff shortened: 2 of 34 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
```diff
@@ -30,38 +30,44 @@ internal class SdkInternalLogger(
...
-     * Global SDK Logger. This logger is meant for internal debugging purposes.
+     * This logger is meant for internal debugging purposes.
      * Logcat logs are conditioned by a BuildConfig flag (set to false for releases).
      */
-    internal val sdkLogger = sdkLogHandlerFactory.invoke()
+    internal val maintainerLogger = maintainerLogHandlerFactory.invoke()
+
+    private val onlyOnceUserMessages = mutableListOf<String>()
+    private val onlyOnceMaintainerMessages = mutableListOf<String>()
+    private val onlyOnceTelemetryMessages = mutableListOf<String>()
```
0xnm, line 47: should it be `mutableSetOf` instead? otherwise we are hitting O(N) for `contains` calls.

0xnm, line 47: minor: Telemetry already has "already seen" filter, but imo it is fine to have it here as well to reduce the amount redundant work.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/api/InternalLogger.kt
```diff
@@ -64,12 +64,15 @@ interface InternalLogger {
      * @param target the target handler for the log
      * @param message the log message
      * @param throwable an optional throwable error
+     * @param onlyOnce whether only one instance of the message should be sent per lifetime of the
```
0xnm, line 67: should it be per target as well? otherwise if same message has different targets, it will go to the first, but not to the second probably?

EDIT: ok, filters are per target in the implementation. Maybe we should update doc then to make it clear.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/v2/core/SdkInternalLoggerTest.kt
```diff
@@ -60,13 +61,34 @@ internal class SdkInternalLoggerTest {
...
+        }
+    }
+
+    fun callWithString(i: Int, str: String) {
+        if (i == 0) {
+            print(str)
+        }
+    }
+
+//    Simple string
```
0xnm, line 81: some benchmarking leftover? :)

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -224,7 +224,8 @@ internal constructor(
             (sdkCore?.internalLogger ?: InternalLogger.UNBOUND).log(
                 InternalLogger.Level.WARN,
                 InternalLogger.Target.USER,
-                WARN_RUM_DISABLED.format(Locale.US, prefix)
+                WARN_RUM_DISABLED.format(Locale.US, prefix),
+                onlyOnce = true
```
0xnm, line 228: here maybe we should use `INFO` and allow multiple logs? wdyt? the idea is that `DatadogInterceptor` can start working before SDK is initialized and we need to let user know that __some__ calls will not be tracked.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
```diff
@@ -75,33 +81,71 @@ internal class SdkInternalLogger(
...
+        error: Throwable?,
+        onlyOnce: Boolean,
+        knownSingleMessages: MutableList<String>
+    ) {
+        if (onlyOnce) {
+            if (knownSingleMessages.contains(message)) {
+                // drop the message… wait should we log that we dropped it?
+                return
+            } else {
+                knownSingleMessages.add(message)
```
0xnm, line 128: I think it may be useful to add message attributed to the particular SDK instance. Because in the situation when 2 SDK instances are logging the same message, only message associated with first instance will be shown, masking the message/issue with the second instance as well.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
xgouchet: Good call

xgouchet: Yes agreed if we can catch duplicates early that's better

xgouchet: Also the Telemetry filter is about "seen in the current RUM session", here we filter message seen once in the lifetyme of the logger

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
xgouchet: The Internal logger is already tied to a specific instance, meaning that message seen in one instance won't impact messages seen in another instance, or in the UNBOUND logger

xgouchet: The Internal logger is already tied to a specific instance, meaning that message seen in one instance won't impact messages seen in another instance, or in the UNBOUND logger

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
xgouchet: That makes sense indeed.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
0xnm: true, I missed that.

0xnm (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/SdkInternalLogger.kt
```diff
@@ -30,38 +30,44 @@ internal class SdkInternalLogger(
...
-     * Global SDK Logger. This logger is meant for internal debugging purposes.
+     * This logger is meant for internal debugging purposes.
      * Logcat logs are conditioned by a BuildConfig flag (set to false for releases).
      */
-    internal val sdkLogger = sdkLogHandlerFactory.invoke()
+    internal val maintainerLogger = maintainerLogHandlerFactory.invoke()
+
+    private val onlyOnceUserMessages = mutableSetOf<String>()
+    private val onlyOnceMaintainerMessages = mutableSetOf<String>()
+    private val onlyOnceTelemetryMessages = mutableSetOf<String>()
```
mariusc83, line 47: do you want to keep the strings or maybe a hashcode will be lighter with the caveat that you may have some collisions ? I wonder in case of the String if we should add a backup logic here in case of high cardinality in logs (I am thinking of sending a log with an id every time an operation is performed)

xgouchet: We don't have that many logs that need to be sent only once, so I think using String won't have a bad impact.

mariusc83 (approved):
LGTM

PR #1491, opened 2023-06-22
RUMM-3359: Move position of `sdkCore` argument in case of being default one to the tail

0xnm:
### What does this PR do?

This change moves `sdkCore` argument to the end of method/constructor arguments in case if it is default one, because it will simplify the migration: in case if value is default, there won't be need to update old method/constructor signature on the client side.

The only exception to this rule is made when the last argument is lambda, in this case `sdkCore` comes right before the lambda. It creates some discrepancy in case of `registerIfAbsent` method:

* `fun registerIfAbsent(RumMonitor, com.datadog.android.v2.api.SdkCore = Datadog.getInstance()): Boolean`
* `fun registerIfAbsent(com.datadog.android.v2.api.SdkCore = Datadog.getInstance(), java.util.concurrent.Callable<RumMonitor>): Boolean`

But in the latter case we cannot move `Callable` to the first position, because something with named argument like `registerIfAbsent(sdkCore) { ... }` to make lambda last in the call won't compile.


mariusc83 (approved):
LGTM

xgouchet (approved):

PR #1492, opened 2023-06-23
RUMM-3021 create lazy logging

xgouchet:
### What does this PR do?

Create a lazy message for the internal Logger for better performances when the target logger is disabled

### DIFF
--- a/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
+++ b/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
@@ -30,6 +30,7 @@ import com.datadog.android.rum.RumActionType
 import com.datadog.android.rum.RumAttributes
 import com.datadog.android.rum.RumMonitor
 import com.datadog.android.v2.api.SdkCore
+import kotlinx.coroutines.flow.FlowCollector
 import java.lang.Exception
 import kotlin.coroutines.cancellation.CancellationException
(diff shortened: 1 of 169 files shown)


integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/InteractionTracking.kt
```diff
@@ -244,35 +246,38 @@ internal suspend fun <T> trackDragInteraction(
...
-                            // give a fake start
-                            Log.e(
-                                LOG_TAG,
-                                "Unexpected branch reached" +
-                                    " for the drag interaction start"
-                            )
-                            DragInteraction.Start()
-                        }
+        interactionSource.interactions.collect(
+            FlowCollector { interaction ->
```
0xnm, line 250: strange that it compiled. probably it is pure recent Android Studio thing (for sure it didn't complain before), which is not capable to understanding that `FlowCollector` is functional interface (but with `suspend` method), while Kotlin compiler is ok with that.

local_ci.sh
```diff
@@ -146,6 +146,9 @@ if [[ $TEST == 1 ]]; then
 
   echo "---- Unit tests (Debug)"
   ./gradlew uTD
+
+  echo "---- Unit tests (Relase)"
```
0xnm, line 150: small typo

```suggestion
  echo "---- Unit tests (Release)"
```

0xnm (approved):
nice job! I've added some comments, but they are not blocking for me.

PR #1499, opened 2023-06-26
Add persistence executor service factory for unit tests

0xnm:
### What does this PR do?

Some tests are flaky because there is another thread involved in the tests which may interact with certain parts like `internalLogger`. This PR introduces a factory for the persistence executor service, so that we can provide "same thread" implementation for unit tests in order to avoid concurrency.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
```diff
@@ -281,7 +284,11 @@ internal class DatadogCore(
         }
 
         // always initialize Core Features first
-        coreFeature = CoreFeature(internalLogger)
+        coreFeature = if (persistenceExecutorServiceFactory != null) {
+            CoreFeature(internalLogger, persistenceExecutorServiceFactory)
```
mariusc83, line 288: why not just have a default value instead and make it mandatory in the `CoreFeature` constructor ?

mariusc83 (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/v2/core/DatadogCore.kt
0xnm: If I pull the default value to `DatadogCore`, then I'm leaking the implementation details with some properties coming from `CoreFeature`.

Issue #1469, opened 2023-06-08
One CPU fully occupied by Datadog when device offline


torbjorn-humane:
We're using the Datadog android java SDK to collect both traces and logs on an android device. The device may be offline for an extended period, which means logs/traces cannot be uploaded but are stored on device. After a while in this state, the datadog agent is fully occupying 1 CPU. We have reproduced the issue both on device and using the android emulator. We have tested both version 1.9.0 and 1.19.2 and the issue persists. We're using the Datadog-Timber integration for log/trace capturing.

**Steps to reproduce the issue:**

1. Create an android instrumented test that loops over a log statement say 100k times. 
2. Start the android emulator and enable airplane mode on the emulated device.
3. Run the test and observe the folder on device (using e.g. adb shell) where the cached files are stored (e.g. `/data/user/0/<myapplication>/cache/datadog-.../logs-v2`
4. When the number of cached files reaches ~200, stop the test.
5. Create another instrumented test that logs in a loop. For each iteration of the loop, sleep 10ms (emulating 100 log entries/second).
6. Run the second test and attach the profiler, e.g. android studio. One of the worker threads will occupy one CPU 100%. Breaking in the debugger indicates that the Datadog agent is using one of the worker threads to iterate over every stored file repeatedly, causing the high CPU usage.

**Describe what you expected:**
Running the same test with datadog enabled, emulating 100 log writes/second, occupies ~8% of one CPU when the device has network access. I'd expect no more CPU usage even if the device is offline.

**Additional context**
 - Android OS version: 31
 - Emulator: Pixel 6 Pro
 - Datadog SDK version: 1.9.0, 1.19.2
 - Versions of any other relevant dependencies (OkHttp, …): Timber 5.0.1, okhttp 4.9.3
 - Proguard configuration: N/A
 - Gradle Plugins: N/A

torbjorn-humane:
Hi and thanks for getting back to me! You're of course right, my example is doing some excessive logging but it's mostly a quick way to demonstrate the problem. Our real logging rate is also somewhat excessive but not like above. Still, because the device can be offline for hours and hours the problem occurs eventually. Below are a couple of callstacks I get from the worker thread when breaking the test:

Example 1
```
isFileSafe:85, FileExtKt (com.datadog.android.core.internal.persistence.file)
accept:285, BatchFileOrchestrator$BatchFileFilter (com.datadog.android.core.internal.persistence.file.batch)
listFiles:1298, File (java.io)
[...]

PR #1506, opened 2023-06-28
RUMM-3338: Introduce known file cache and cleanup throttling in `BatchFileOrchestrator` in order to reduce the number of syscalls

0xnm:
### What does this PR do?

The goal of this PR is to reduce CPU consumption during the scenarios like described in https://github.com/DataDog/dd-sdk-android/issues/1469.

The issue comes from the fact that when we have many batch files on the disk calling `listSortedBatchFiles` as-is (which is called several times during `getWritableFile`, for example) becomes quite expensive because:

* syscalls (like `isFile`) are expensive
* pattern matching becomes expensive as well.

In order to optimize that, the following is done:

* Introduced the cache for the files already known to be batch files, so that we don't need to call `isFile` and use regex.
* Introduce throttling for the cleanup (only for the `getWritableFile` path, because a) it can be called much more often than `getReadableFile`, for every event; b) cleanup actions are a bit different between `getReadableFile` and `getWritableFile`, so having the same timestamp is not quite correct).

Test scenario: 300 batch files on the disk, disabled network, writing 1 log entry every 100 ms.

CPU usage after the change (same scenario and duration is used):

image

It gives a very good bump if the total number files on the disk is below the cache max size. If cache is not enough, the problem comes back (but maybe in a bit better state, because we have cleanup throttling)

CPU usage before the change:

image

### DIFF
--- a/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
+++ b/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
@@ -7,6 +7,7 @@
 package com.datadog.android.core.internal.persistence.file.batch
 
 import androidx.annotation.WorkerThread
+import androidx.collection.LruCache
 import com.datadog.android.core.internal.persistence.file.FileOrchestrator
 import com.datadog.android.core.internal.persistence.file.FilePersistenceConfig
 import com.datadog.android.core.internal.persistence.file.canWriteSafe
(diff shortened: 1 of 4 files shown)


mariusc83 (approved):
Nice job ;)

dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -67,6 +77,7 @@ internal class BatchFileOrchestrator(
         }
 
         deleteObsoleteFiles()
+        lastCleanupTimestamp = System.currentTimeMillis()
```
xgouchet, line 80: maybe the update of `lastCleanupTimestamp` can be part of the `deleteObsoleteFiles()` method?

xgouchet (approved):

PR #1511, opened 2023-06-29
REPLAY-1817 Require sampleRate when building the SR configuration

mariusc83:
### What does this PR do?

In the iOS sdk we are currently explicitly require the `sampleRate` value when enabling the Session Replay feature. We need to align on the Android side also and require this when building the SR configuration object.

Please also note that given this change the `setSessionReplayRate` method doesn't make sense anymore and was removed.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
@@ -24,12 +24,13 @@ data class SessionReplayConfiguration internal constructor(
 
     /**
      * A Builder class for a [SessionReplayConfiguration].
+     * @param sampleRate must be a value between 0 and 100. A value of 0
+     * means no session will be recorded, 100 means all sessions will be recorded.
      */
-    class Builder {
+    class Builder(@FloatRange(from = 0.0, to = 100.0)private val sampleRate: Float) {
         private var customEndpointUrl: String? = null
         private var privacy = SessionReplayPrivacy.MASK_ALL
         private var extensionSupport: ExtensionSupport = NoOpExtensionSupport()
-        private var sampleRate: Float = DEFAULT_SAMPLE_RATE
 
         /**
          * Adds an extension support implementation. This is mostly used when you want to provide
@@ -62,22 +63,7 @@ data class SessionReplayConfiguration internal constructor(
         }
 
         /**
-         * Sets the sample rate for Session Replay recorded Sessions. Please note that this
-         * sample rate will be applied on top of the already sampled in RUM session.
-         *
-         * @param sampleRate must be a value between 0 and 100. A value of 0
-         * means no session will be recorded, 100 means all sessions will be recorded.
-         * The default value for the sample rate will be 0 meaning that no session replay will be
-         * recorded if the sample rate will not be explicitly set.
-         */
-        fun setSessionReplaySampleRate(@FloatRange(from = 0.0, to = 100.0) sampleRate: Float):
-            Builder {
-            this.sampleRate = sampleRate
-            return this
-        }
-
-        /**
-         * * Builds a [SessionReplayConfiguration] based on the current state of this Builder.
+         * Builds a [SessionReplayConfiguration] based on the current state of this Builder.
          */
         fun build(): SessionReplayConfiguration {
             return SessionReplayConfiguration(
@@ -100,9 +86,5 @@ data class SessionReplayConfiguration internal constructor(
             }
             return emptyList()
         }
-    }
-
-    internal companion object {
-        internal const val DEFAULT_SAMPLE_RATE = 0f
     }
 }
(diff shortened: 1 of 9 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -62,24 +60,13 @@ data class SessionReplayConfiguration internal constructor(
...
-         * Sets the sample rate for Session Replay recorded Sessions. Please note that this
-         * sample rate will be applied on top of the already sampled in RUM session.
-         *
+         * Builds a [SessionReplayConfiguration] based on the current state of this Builder.
          * @param sampleRate must be a value between 0 and 100. A value of 0
          * means no session will be recorded, 100 means all sessions will be recorded.
          * The default value for the sample rate will be 0 meaning that no session replay will be
          * recorded if the sample rate will not be explicitly set.
          */
-        fun setSessionReplaySampleRate(@FloatRange(from = 0.0, to = 100.0) sampleRate: Float):
```
mariusc83, line 73: Because this value is required when building the config there is no reason to have this method.

maciejburda (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
ncreated: Makes sense 👍. 

By removing this method we make the sample rate immutable, which is also the way we define it on iOS:
```swift
let sampleRate: Float // immutable & non-optional
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -62,24 +60,13 @@ data class SessionReplayConfiguration internal constructor(
...
-            Builder {
-            this.sampleRate = sampleRate
-            return this
-        }
-
-        /**
-         * * Builds a [SessionReplayConfiguration] based on the current state of this Builder.
-         */
-        fun build(): SessionReplayConfiguration {
+        fun build(sampleRate: Float): SessionReplayConfiguration {
```
xgouchet, line 69: In the Builder pattern, it's common to have the mandatory parameters in the constructor, and keep the build method a no argument method.

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
mariusc83: Yes...didn't know where to put it actually but you are right.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -24,12 +23,15 @@ data class SessionReplayConfiguration internal constructor(
 
     /**
      * A Builder class for a [SessionReplayConfiguration].
+     * @param sampleRate must be a value between 0 and 100. A value of 0
+     * means no session will be recorded, 100 means all sessions will be recorded.
+     * The default value for the sample rate will be 0 meaning that no session replay will be
+     * recorded if the sample rate will not be explicitly set.
      */
-    class Builder {
+    class Builder(private val sampleRate: Float) {
```
0xnm, line 31: let's add `@FloatRange(from = 0.0, to = 100.0)` annotation as well here, like it was in the `setSessionReplaySampleRate` method.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -24,12 +23,15 @@ data class SessionReplayConfiguration internal constructor(
 
     /**
      * A Builder class for a [SessionReplayConfiguration].
+     * @param sampleRate must be a value between 0 and 100. A value of 0
+     * means no session will be recorded, 100 means all sessions will be recorded.
+     * The default value for the sample rate will be 0 meaning that no session replay will be
```
0xnm, line 28: there is no default value, this comment should be updated

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -62,23 +64,12 @@ data class SessionReplayConfiguration internal constructor(
...
         /**
-         * Sets the sample rate for Session Replay recorded Sessions. Please note that this
-         * sample rate will be applied on top of the already sampled in RUM session.
-         *
+         * Builds a [SessionReplayConfiguration] based on the current state of this Builder.
          * @param sampleRate must be a value between 0 and 100. A value of 0
          * means no session will be recorded, 100 means all sessions will be recorded.
          * The default value for the sample rate will be 0 meaning that no session replay will be
          * recorded if the sample rate will not be explicitly set.
          */
```
0xnm, line 72: this to be removed, there is no `sampleRate` param for the `build` method.

0xnm (changes requested):

0xnm (approved):

PR #1512, opened 2023-06-29
RUMM-3356 fix jankstats

xgouchet:
### What does this PR do?

This PR aims at solving some issues found when using the JankStats tool to track Mobile Vitals

- Delay the initialization to `onActivityStarted()`: before that the window is not fully created and the JankStats creation would often fail
- Add some tracking of the active windows: several activities could theoretically share the same window. This PR tracks this to avoid registering more than one listener to the same window
- Furthermore, we now listen to `onActivityStopped()` to prevent keeping a listener on a Window not being used anymore. 

### Additional Notes

- Since the JankStats doesn't measure the actual frame rate, but the time needed to draw the frame, scaling by the device's maximum frame rate doesn't make sense anymore.
- A custom view in the sample app is added to illustrate what a slow framerate can look like; it uses a random image generation (avoids possible optimization by the compiler or JIT), and performs a lot of bitmap operation that makes it very slow. 


dd-sdk-android/src/test/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListenerTest.kt
```diff
@@ -62,25 +62,164 @@ internal class JankStatsActivityLifecycleListenerTest {
...
+
+        // Then
+        inOrder(mockJankStatsProvider, mockJankStats) {
+            verify(mockJankStatsProvider).createJankStatsAndTrack(mockWindow, testedJankListener, mockInternalLogger)
+            verify(mockJankStats).isTrackingEnabled = false
+        }
+    }
+
+    @Test
+    fun `𝕄 foNothing 𝕎 onActivityStopped() {}`() {
```
0xnm, line 145:
```suggestion
    fun `𝕄 do nothing 𝕎 onActivityStopped() {}`() {
```

detekt_custom.yml
```diff
@@ -613,6 +613,7 @@ datadog:
       - "kotlin.collections.List.toSet()"
       - "kotlin.collections.List.withIndex()"
       - "kotlin.collections.Map.containsKey(kotlin.String)"
+      - "kotlin.collections.MutableMap.containsKey(android.view.Window)"
```
0xnm, line 616: probably should be placed below, around line 673.

dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -70,6 +76,27 @@ internal class JankStatsActivityLifecycleListener(
...
+            it.get() == null || it.get() == activity
+        }
+        activeActivities[window] = list
+        if (list.isEmpty()) {
+            internalLogger.log(
+                InternalLogger.Level.DEBUG,
+                InternalLogger.Target.MAINTAINER,
+                "Disabling jankStats for window $window"
+            )
+            activeWindowsListener[window]?.isTrackingEnabled = false
```
0xnm, line 98: I'm wondering if we should use `WeakHashMap` for `activeActivities` and `activeWindowsListener`, in case if the window was temporary and will be destroyed once a particular activity is destroyed.

I can see that `PolicyManager.makeNewWindow` can be called from different places, `Dialog` constructor included. So if there is a dialog shown and then destroyed, we will be keeping `Window` instance in the memory forever, preventing it from be GCed.

0xnm (commented):
nice work, thanks for the help! I added few comments/suggestions.

PR #1520, opened 2023-07-05
REPLAY-1862 Increate batchSizeLimit to 10MB for SR feature

mariusc83:
### What does this PR do?

In this PR we are increasing the SR batch size limit to 10MB into the core to support large payloads containing embedded base64 images. This follows the same approach as in iOS introduced with the Image recording feature added lately. 

Please note that the SR intake unlike RUM accepts payloads up to 10MB size.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
@@ -100,7 +100,7 @@ internal class SessionReplayFeature constructor(
         SessionReplayRequestFactory(customEndpointUrl)
 
     override val storageConfiguration: FeatureStorageConfiguration =
-        FeatureStorageConfiguration.DEFAULT
+        STORAGE_CONFIGURATION
 
     override fun onStop() {
         stopRecording()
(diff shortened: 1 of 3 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -211,6 +211,25 @@ internal class SessionReplayFeature constructor(
...
     internal companion object {
+
+        /**
+         * Session Replay storage configuration with the following parameters:
+         * * max item size = 4 MB,
+         * * max items per batch = 500,
+         * * max batch size = 10 MB, SR intake batch limit is 10MB
+         * * old batch threshold = 18 hours.
+         */
+        internal val STORAGE_CONFIGURATION: FeatureStorageConfiguration =
```
0xnm, line 222: isn't it better to take `DEFAULT` and call `copy` here, only overriding what is needed for SR?

mariusc83: Hmmm....good question, not sure which one is better. I think I will do that.

0xnm (approved):

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -211,6 +211,17 @@ internal class SessionReplayFeature constructor(
...
+            FeatureStorageConfiguration.DEFAULT.copy(maxBatchSize = 10 * 1024 * 1024)
```
xgouchet, line 223: Shouldn't the maxItemSize be increased as well?

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
mariusc83: I am actually assessing this...I think 512KB should be enough...let me check with Maciej

mariusc83: Yes, you were right, they increased that too so right now we are 10MB, 10MB for both item and bath size to align with them for SR

jonathanmos (approved):

PR #1528, opened 2023-07-06
Reduce flakiness of RUM View Tracking tests

0xnm:
### What does this PR do?

This change reduces flakiness of RUM View Tracking tests. The issue is that `launch` [exists once Activity transitions to the `RESUMED` state](https://developer.android.com/reference/androidx/test/core/app/ActivityScenario#launch(java.lang.Class%3CA%3E)), meaning if there is no more interaction test completes and `waitForIdleSync` in `NightlyTestRule` can be executed before activity actually transitions to the `DESTROYED` state (when `stopView` is called).

This change ensures activity explicitly transitions to the `DESTROYED` state, so that `stopView` event is fired before test actually completes.

### DIFF
--- a/instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumViewTrackingE2ETests.kt
+++ b/instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumViewTrackingE2ETests.kt
@@ -8,6 +8,7 @@ package com.datadog.android.nightly.rum
 
 import android.app.Activity
 import androidx.fragment.app.Fragment
+import androidx.lifecycle.Lifecycle
 import androidx.navigation.NavDestination
 import androidx.test.core.app.ActivityScenario.launch
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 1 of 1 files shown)


mariusc83 (approved):
Nice catch

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumViewTrackingE2ETests.kt
```diff
@@ -64,6 +65,7 @@ internal class RumViewTrackingE2ETests {
             )
         }
         launch(ViewTrackingActivity::class.java)
+            .moveToState(Lifecycle.State.DESTROYED)
```
mariusc83, line 68: I wonder if this is only needed in these test suite 🤔

mariusc83 (approved):

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/rum/RumViewTrackingE2ETests.kt
0xnm: Other tests have some UI interactions after `launch` call, so they have view updates before test is finished.

xgouchet (approved):

PR #1533, opened 2023-07-10
REPLAY-1887 Provide the uploadFrequency per feature

mariusc83:
### What does this PR do?

This PR is follow - up on the [ios PR](https://github.com/DataDog/dd-sdk-ios/pull/1363) where we are providing a custom (batchSize, uploadFrequency) configuration for Session Replay feature to decrease the number of sessions with corrupted/not-enough records.
### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/FeatureStorageConfiguration.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/FeatureStorageConfiguration.kt
@@ -6,6 +6,9 @@
 
 package com.datadog.android.api.storage
 
+import com.datadog.android.core.configuration.BatchSize
+import com.datadog.android.core.configuration.UploadFrequency
+
 /**
  * Contains the storage configuration for an [FeatureScope] instance.
  * @property maxItemSize the maximum size (in bytes) for a single item in a batch
@@ -13,12 +16,18 @@ package com.datadog.android.api.storage
  * @property maxBatchSize the maximum size (in bytes) of a complete batch
  * @property oldBatchThreshold the duration (in milliseconds) after which a batch is considered too
  * old to be uploaded (usually because it'll be discarded at ingestion by the backend)
+ * @property uploadFrequency the desired upload frequency policy. If not explicitly provided this
+ * value will be taken from core configuration.
+ * @property batchSize the desired batch size policy.If not explicitly provided this
+ * value will be taken from core configuration.
  */
 data class FeatureStorageConfiguration(
     val maxItemSize: Long,
     val maxItemsPerBatch: Int,
     val maxBatchSize: Long,
-    val oldBatchThreshold: Long
+    val oldBatchThreshold: Long,
+    val uploadFrequency: UploadFrequency?,
+    val batchSize: BatchSize?
 ) {
     companion object {
 
@@ -36,7 +45,9 @@ data class FeatureStorageConfiguration(
             // 4 MB
             maxBatchSize = 4L * 1024 * 1024,
             // 18 hours
-            oldBatchThreshold = 18L * 60L * 60L * 1000L
+            oldBatchThreshold = 18L * 60L * 60L * 1000L,
+            uploadFrequency = null,
+            batchSize = null
         )
     }
 }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
@@ -33,7 +33,7 @@ internal class DataUploadRunnable(
     private val contextProvider: ContextProvider,
     private val networkInfoProvider: NetworkInfoProvider,
     private val systemInfoProvider: SystemInfoProvider,
-    uploadFrequency: UploadFrequency,
+    internal val uploadFrequency: UploadFrequency,
     private val batchUploadWaitTimeoutMs: Long = CoreFeature.NETWORK_TIMEOUT_MS,
     private val internalLogger: InternalLogger
 ) : UploadRunnable {
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
@@ -15,13 +15,17 @@ import com.datadog.android.api.feature.FeatureEventReceiver
 import com.datadog.android.api.feature.StorageBackedFeature
 import com.datadog.android.api.storage.EventBatchWriter
 import com.datadog.android.api.storage.FeatureStorageConfiguration
+import com.datadog.android.core.configuration.BatchSize
+import com.datadog.android.core.configuration.UploadFrequency
 import com.datadog.android.core.internal.data.upload.DataOkHttpUploader
 import com.datadog.android.core.internal.data.upload.NoOpUploadScheduler
 import com.datadog.android.core.internal.data.upload.UploadScheduler
 import com.datadog.android.core.internal.data.upload.v2.DataUploadScheduler
 import com.datadog.android.core.internal.data.upload.v2.NoOpDataUploader
+import com.datadog.android.core.internal.persistence.ConsentAwareStorage
 import com.datadog.android.core.internal.persistence.NoOpStorage
 import com.datadog.android.core.internal.persistence.Storage
+import com.datadog.android.core.internal.persistence.file.FilePersistenceConfig
 import com.datadog.android.core.internal.persistence.file.NoOpFileOrchestrator
 import com.datadog.android.privacy.TrackingConsent
 import com.datadog.android.privacy.TrackingConsentProviderCallback
@@ -32,6 +36,7 @@ import com.datadog.android.utils.verifyLog
 import com.datadog.tools.unit.annotations.TestConfigurationsProvider
 import com.datadog.tools.unit.extensions.TestConfigurationExtension
 import com.datadog.tools.unit.extensions.config.TestConfiguration
+import fr.xgouchet.elmyr.Forge
 import fr.xgouchet.elmyr.annotation.BoolForgery
 import fr.xgouchet.elmyr.annotation.Forgery
 import fr.xgouchet.elmyr.annotation.StringForgery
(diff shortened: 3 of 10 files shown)


xgouchet (changes requested):
It works, my main concern is making the core configuration's upload frequency optional. 

It would be better to keep the core configuration mandatory, and make individual features override optional.  If you flip the `resolveUploadFrequency` logic to the following, it would work better (that way people can just override the one feature they want a different upload frequency)

```kotlin
    private fun resolveUploadFrequency(
        coreFeature: CoreFeature,
        storageConfiguration: FeatureStorageConfiguration
    ): UploadFrequency {
        return storageConfiguration.uploadFrequency ?: coreFeature.uploadFrequency
    }
```

mariusc83:
> It works, my main concern is making the core configuration's upload frequency optional.
> 
> It would be better to keep the core configuration mandatory, and make individual features override optional. If you flip the `resolveUploadFrequency` logic to the following, it would work better (that way people can just override the one feature they want a different upload frequency)
> 
> ```kotlin
>     private fun resolveUploadFrequency(
>         coreFeature: CoreFeature,
>         storageConfiguration: FeatureStorageConfiguration
>     ): UploadFrequency {
>         return storageConfiguration.uploadFrequency ?: coreFeature.uploadFrequency
>     }
> ```

mariusc83:
> It works, my main concern is making the core configuration's upload frequency optional.
> 
> It would be better to keep the core configuration mandatory, and make individual features override optional. If you flip the `resolveUploadFrequency` logic to the following, it would work better (that way people can just override the one feature they want a different upload frequency)
> 
> ```kotlin
>     private fun resolveUploadFrequency(
>         coreFeature: CoreFeature,
>         storageConfiguration: FeatureStorageConfiguration
>     ): UploadFrequency {
>         return storageConfiguration.uploadFrequency ?: coreFeature.uploadFrequency
>     }
> ```

Tks @xgouchet, I think that's even better for now. My idea was that at some point we will drop this API at the core level as it doesn't make sense to have it anymore there. We could simplify it by having this only in the `StorageConfiguration`. But to keep a backward compatibility we could go this way indeed. I would like to hear also @0xnm thoughts on that.

ncreated (commented):
The override logic looks good 👌

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/BatchingDelayResolver.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core.internal
+
+import com.datadog.android.api.storage.FeatureStorageConfiguration
+
+internal class BatchingDelayResolver {
```
0xnm, line 11: imo it is a bit too much of abstractions, let's maybe just mock values returned by the `wrappedFeature` in the `SdkFeature`?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/FeatureStorageConfiguration.kt
```diff
@@ -6,19 +6,26 @@
...
  * old to be uploaded (usually because it'll be discarded at ingestion by the backend)
  */
 data class FeatureStorageConfiguration(
     val maxItemSize: Long,
     val maxItemsPerBatch: Int,
     val maxBatchSize: Long,
-    val oldBatchThreshold: Long
+    val oldBatchThreshold: Long,
+    val uploadFrequency: UploadFrequency? = null,
+    val batchSize: BatchSize? = null
```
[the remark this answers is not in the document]
mariusc83: actually not...I prefer NULL, it's easier when I resolve them (if it is set here I take this one if not I take the core one)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -33,7 +33,7 @@ internal class DataUploadRunnable(
     private val contextProvider: ContextProvider,
     private val networkInfoProvider: NetworkInfoProvider,
     private val systemInfoProvider: SystemInfoProvider,
-    uploadFrequency: UploadFrequency,
+    internal val uploadFrequency: UploadFrequency,
```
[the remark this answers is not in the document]
mariusc83: tests

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/BatchingDelayResolver.kt
mariusc83: I get your point, in the same time I prefer to have them separated and do not mix responsibilities. It's way easier for unit testing. I also suspect that this logic will change later so in that case `SdkFeature` will not change anymore and we will just change the resolver. What do you think ?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/FeatureStorageConfiguration.kt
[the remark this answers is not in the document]
0xnm: Yes, what I mean is to keep nullable type here, but put default `null` value to the `DEFAULT` object.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/UploadFrequencyResolverTest.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    @BeforeEach
+    fun `set up`(forge: Forge) {
+        fakeCoreUploadFrequency = forge.aValueFrom(UploadFrequency::class.java)
+        whenever(mockCoreFeature.uploadFrequency).thenReturn(fakeCoreUploadFrequency)
+        testedUploadFrequencyResolver = UploadFrequencyResolver()
+    }
+
+    @Test
+    fun `M provide the core frequency W resolveUploadFrequency{not provide in storage config}`() {
```
0xnm, line 53:
```suggestion
    fun `M provide the core frequency W resolveUploadFrequency{not provided in storage config}`() {
```

0xnm (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/BatchingDelayResolver.kt
xgouchet: not a fan of the overhead introduced by the resolvers either, it would be cleaner to not use them IMO

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/FeatureStorageConfiguration.kt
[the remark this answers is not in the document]
mariusc83: hmm let me look into that, not sure I understand :)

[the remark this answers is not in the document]
mariusc83: oh yeah, I see your point, done ;)

xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/BatchingDelayResolver.kt
mariusc83: I will try to add these as private methods instead, just need to find a better way to test it without too much noise.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/FeatureStorageConfiguration.kt
```diff
@@ -6,19 +6,26 @@
...
+import com.datadog.android.core.configuration.UploadFrequency
+
 /**
  * Contains the storage configuration for an [FeatureScope] instance.
  * @property maxItemSize the maximum size (in bytes) for a single item in a batch
  * @property maxItemsPerBatch the maximum number of individual items in a batch
  * @property maxBatchSize the maximum size (in bytes) of a complete batch
  * @property oldBatchThreshold the duration (in milliseconds) after which a batch is considered too
+ * @property uploadFrequency the desired upload frequency policy
+ * @property batchSize the desired batch size policy
```
0xnm, line 19: minor: maybe we should say, that if not explicitly provided, then value will be taken from the core configuration.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -130,6 +149,61 @@ internal class SdkFeatureTest {
...
+    fun `𝕄 use the storage frequency if set 𝕎 initialize()`(forge: Forge) {
+        // Given
+        val fakeUploadFrequency = forge.aValueFrom(UploadFrequency::class.java)
+        val fakeStorageConfig = mockWrappedFeature.storageConfiguration
+            .copy(uploadFrequency = fakeUploadFrequency)
+        whenever(mockWrappedFeature.storageConfiguration)
+            .thenReturn(fakeStorageConfig)
+        // When
+
+        testedFeature.initialize(appContext.mockInstance)
```
0xnm, line 162:
```suggestion
        whenever(mockWrappedFeature.storageConfiguration)
            .thenReturn(fakeStorageConfig)

        // When
        testedFeature.initialize(appContext.mockInstance)
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -130,6 +149,61 @@ internal class SdkFeatureTest {
...
+
+        // Then
+        assertThat(testedFeature.uploadScheduler)
+            .isInstanceOf(DataUploadScheduler::class.java)
+        val dataUploadRunnable = (testedFeature.uploadScheduler as DataUploadScheduler).runnable
+        assertThat(dataUploadRunnable.uploadFrequency).isEqualTo(fakeUploadFrequency)
+    }
+
+    @Test
+    fun `𝕄 initialize the storage 𝕎 initialize()`() {
```
0xnm, line 172: with core batch size probably is missing in the name?

0xnm (approved):
lgtm, added few comments

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
mariusc83: I am considering this as default initialization so wanted to keep it as that in the test description

PR #1534, opened 2023-07-10
REPLAY-1727: Base64 Caching Mechanism

jonathanmos:
### What does this PR do?
Adds an LRUCache for base64 strings - cache base64 strings against drawable hashcodes. 
Evict all items in the LRUCache every time we change views. 

### Motivation
Without a caching mechanism we create and garbage collect an excessive number of bitmaps. 

### Additional Notes
This PR contains a revert commit of the revert of imagebuttonmapper (iow it adds base64 functionality back)

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
@@ -6,12 +6,15 @@
 
 package com.datadog.android.sessionreplay.internal.async
 
+import android.graphics.drawable.Drawable
 import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
 import com.datadog.android.sessionreplay.internal.processor.RecordedDataProcessor
 import com.datadog.android.sessionreplay.internal.processor.RumContextDataHandler
 import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
+import com.datadog.android.sessionreplay.internal.recorder.base64.Base64LRUCache
+import com.datadog.android.sessionreplay.internal.recorder.base64.Cache
 import com.datadog.android.sessionreplay.internal.utils.TimeProvider
 import com.datadog.android.sessionreplay.model.MobileSegment
 import java.lang.ClassCastException
@@ -34,15 +37,18 @@ internal class RecordedDataQueueHandler {
     private var processor: RecordedDataProcessor
     private var rumContextDataHandler: RumContextDataHandler
     private var timeProvider: TimeProvider
+    private var cache: Cache<Drawable, String>
 
     internal constructor(
         processor: RecordedDataProcessor,
         rumContextDataHandler: RumContextDataHandler,
-        timeProvider: TimeProvider
+        timeProvider: TimeProvider,
+        cache: Cache<Drawable, String> = Base64LRUCache
     ) : this(
         processor = processor,
         rumContextDataHandler = rumContextDataHandler,
         timeProvider = timeProvider,
+        cache = cache,
 
         /**
          * TODO: RUMM-0000 consider change to LoggingThreadPoolExecutor once V2 is merged.
@@ -64,12 +70,14 @@ internal class RecordedDataQueueHandler {
         processor: RecordedDataProcessor,
         rumContextDataHandler: RumContextDataHandler,
         timeProvider: TimeProvider,
-        executorService: ExecutorService
+        executorService: ExecutorService,
+        cache: Cache<Drawable, String>
     ) {
         this.processor = processor
         this.rumContextDataHandler = rumContextDataHandler
         this.executorService = executorService
         this.timeProvider = timeProvider
+        this.cache = cache
     }
 
     // region internal
@@ -98,6 +106,11 @@ internal class RecordedDataQueueHandler {
     ): SnapshotRecordedDataQueueItem? {
         val rumContextData = rumContextDataHandler.createRumContextData()
             ?: return null
+
+        // if the view changed then clear the drawable cache
+        if (rumContextData.prevRumContext != rumContextData.newRumContext) {
+            cache.clear()
+        }
 
         val item = SnapshotRecordedDataQueueItem(
             rumContextData = rumContextData,
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
@@ -1,0 +1,134 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.content.ComponentCallbacks2
+import android.content.res.Configuration
+import android.graphics.drawable.AnimationDrawable
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.DrawableContainer
+import android.graphics.drawable.LayerDrawable
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+
+internal object Base64LRUCache : Cache<Drawable, String>, ComponentCallbacks2 {
+    @Suppress("MagicNumber")
+    val MAX_CACHE_MEMORY_SIZE_BYTES = 4 * 1024 * 1024 // 4MB
+    @Suppress("MagicNumber")
+    private val ON_LOW_MEMORY_SIZE_BYTES = MAX_CACHE_MEMORY_SIZE_BYTES / 2 // 50% size
+    @Suppress("MagicNumber")
+    private val ON_MODERATE_MEMORY_SIZE_BYTES = (MAX_CACHE_MEMORY_SIZE_BYTES / 4) * 3 // 75% size
+
+    private var cache: LruCache<String, ByteArray> = object :
+        LruCache<String, ByteArray>(MAX_CACHE_MEMORY_SIZE_BYTES) {
+        override fun sizeOf(key: String?, value: ByteArray): Int {
+            return value.size
+        }
+    }
+
+    override fun onTrimMemory(level: Int) {
+        when (level) {
+            ComponentCallbacks2.TRIM_MEMORY_BACKGROUND -> {
+                cache.evictAll()
+            }
+
+            ComponentCallbacks2.TRIM_MEMORY_COMPLETE -> {
+                cache.evictAll()
+            }
+
+            ComponentCallbacks2.TRIM_MEMORY_MODERATE -> {
+                cache.trimToSize(ON_MODERATE_MEMORY_SIZE_BYTES)
+            }
+
+            ComponentCallbacks2.TRIM_MEMORY_RUNNING_CRITICAL -> {
+                cache.evictAll()
+            }
+
+            ComponentCallbacks2.TRIM_MEMORY_RUNNING_LOW -> {
+                cache.trimToSize(ON_LOW_MEMORY_SIZE_BYTES)
+            }
+
+            ComponentCallbacks2.TRIM_MEMORY_RUNNING_MODERATE -> {
+                cache.trimToSize(ON_MODERATE_MEMORY_SIZE_BYTES)
+            }
+
+            ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN -> {
+                cache.evictAll()
+            }
+
+            else -> {
+                cache.evictAll()
+            }
+        }
+    }
+
+    override fun onConfigurationChanged(newConfig: Configuration) {}
+
+    override fun onLowMemory() {
+        cache.evictAll()
+    }
+
+    @VisibleForTesting
+    internal fun setBackingCache(cache: LruCache<String, ByteArray>) {
+        this.cache = cache
+    }
+
+    @Synchronized
+    override fun put(element: Drawable, value: String) {
+        val key = generateKey(element)
+        val byteArray = value.toByteArray(Charsets.UTF_8)
+        cache.put(key, byteArray)
+    }
+
+    @Synchronized
+    override fun get(element: Drawable): String? =
+        cache.get(generateKey(element))?.let {
+            String(it)
+        }
+
+    @Synchronized
+    override fun size(): Int = cache.size()
+
+    @Synchronized
+    override fun clear() {
+        cache.evictAll()
+    }
+
+    private fun generateKey(drawable: Drawable): String =
+        generatePrefix(drawable) + System.identityHashCode(drawable)
+
+    private fun generatePrefix(drawable: Drawable): String {
+        return when (drawable) {
+            is DrawableContainer -> getPrefixForDrawableContainer(drawable)
+            is LayerDrawable -> getPrefixForLayerDrawable(drawable)
+            else -> ""
+        }
+    }
+
+    private fun getPrefixForDrawableContainer(drawable: DrawableContainer): String {
+        if (drawable !is AnimationDrawable) {
+            return drawable.state.joinToString(separator = "", postfix = "-")
+        }
+
+        return ""
+    }
+
+    private fun getPrefixForLayerDrawable(drawable: LayerDrawable): String {
+        return if (drawable.numberOfLayers > 1) {
+            val sb = StringBuilder()
+            for (index in 0 until drawable.numberOfLayers) {
+                val layer = drawable.getDrawable(index)
+                val layerHash = System.identityHashCode(layer).toString()
+                sb.append(layerHash)
+                sb.append("-")
+            }
+            "$sb"
+        } else {
+            ""
+        }
+    }
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
@@ -6,12 +6,15 @@
 
 package com.datadog.android.sessionreplay.internal.recorder.base64
 
+import android.content.ComponentCallbacks2
+import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.drawable.Drawable
 import android.util.DisplayMetrics
 import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.internal.AsyncImageProcessingCallback
 import com.datadog.android.sessionreplay.internal.utils.Base64Utils
 import com.datadog.android.sessionreplay.internal.utils.DrawableUtils
@@ -25,22 +28,32 @@ import java.util.concurrent.TimeUnit
 @Suppress("UndocumentedPublicClass")
 class Base64Serializer private constructor(
     private val threadPoolExecutor: ExecutorService,
-    private val drawableUtils: DrawableUtils = DrawableUtils(),
-    private val base64Utils: Base64Utils = Base64Utils(),
-    private val webPImageCompression: ImageCompression = WebPImageCompression()
+    private val drawableUtils: DrawableUtils,
+    private val base64Utils: Base64Utils,
+    private val webPImageCompression: ImageCompression,
+    private val base64LruCache: Cache<Drawable, String>,
+    private val logger: InternalLogger
 ) {
-
     private var asyncImageProcessingCallback: AsyncImageProcessingCallback? = null
 
     // region internal
 
     @MainThread
     internal fun handleBitmap(
+        applicationContext: Context,
         displayMetrics: DisplayMetrics,
         drawable: Drawable,
         imageWireframe: MobileSegment.Wireframe.ImageWireframe
     ) {
+        registerCacheForCallbacks(applicationContext)
+
         asyncImageProcessingCallback?.startProcessingImage()
+
+        val cachedBase64 = base64LruCache.get(drawable)
+        if (cachedBase64 != null) {
+            finalizeRecordedDataItem(cachedBase64, imageWireframe, asyncImageProcessingCallback)
+            return
+        }
 
         val bitmap = drawableUtils.createBitmapFromDrawable(drawable, displayMetrics)
 
@@ -51,7 +64,7 @@ class Base64Serializer private constructor(
 
         Runnable {
             @Suppress("ThreadSafety") // this runs inside an executor
-            serialiseBitmap(bitmap, imageWireframe, asyncImageProcessingCallback)
+            serialiseBitmap(drawable, bitmap, imageWireframe, asyncImageProcessingCallback)
         }.let { executeRunnable(it) }
     }
 
@@ -78,30 +91,54 @@ class Base64Serializer private constructor(
 
     @WorkerThread
     private fun serialiseBitmap(
+        drawable: Drawable,
         bitmap: Bitmap,
         imageWireframe: MobileSegment.Wireframe.ImageWireframe,
         asyncImageProcessingCallback: AsyncImageProcessingCallback?
     ) {
-        val base64String = convertBmpToBase64(bitmap)
+        val base64String = convertBmpToBase64(drawable, bitmap)
         finalizeRecordedDataItem(base64String, imageWireframe, asyncImageProcessingCallback)
     }
 
+    @MainThread
+    private fun registerCacheForCallbacks(applicationContext: Context) {
+        if (isCacheRegisteredForCallbacks) return
+
+        if (base64LruCache is ComponentCallbacks2) {
+            applicationContext.registerComponentCallbacks(base64LruCache)
+            isCacheRegisteredForCallbacks = true
+        } else {
+            // Temporarily use UNBOUND logger
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            logger.log(
+                level = InternalLogger.Level.WARN,
+                target = InternalLogger.Target.MAINTAINER,
+                messageBuilder = { DOES_NOT_IMPLEMENT_COMPONENTCALLBACKS }
+            )
+        }
+    }
+
     @WorkerThread
-    private fun convertBmpToBase64(bitmap: Bitmap): String {
+    private fun convertBmpToBase64(drawable: Drawable, bitmap: Bitmap): String {
         val byteArrayOutputStream = webPImageCompression.compressBitmapToStream(bitmap)
 
         if (isOverSizeLimit(byteArrayOutputStream.size())) {
             return ""
         }
 
-        val base64: String
+        val base64String: String
         try {
-            base64 = base64Utils.serializeToBase64String(byteArrayOutputStream)
+            base64String = base64Utils.serializeToBase64String(byteArrayOutputStream)
+
+            if (base64String.isNotEmpty()) {
+                // if we got a base64 string then cache it
+                base64LruCache.put(drawable, base64String)
+            }
         } finally {
             bitmap.recycle()
         }
 
-        return base64
+        return base64String
     }
 
     private fun finalizeRecordedDataItem(
@@ -137,13 +174,18 @@ class Base64Serializer private constructor(
             threadPoolExecutor: ExecutorService = THREADPOOL_EXECUTOR,
             drawableUtils: DrawableUtils = DrawableUtils(),
             base64Utils: Base64Utils = Base64Utils(),
-            webPImageCompression: ImageCompression = WebPImageCompression()
+            webPImageCompression: ImageCompression = WebPImageCompression(),
+            base64LruCache: Cache<Drawable, String> = Base64LRUCache,
+            // Temporarily use UNBOUND until we handle the loggers
+            logger: InternalLogger = InternalLogger.UNBOUND
         ) =
             Base64Serializer(
                 threadPoolExecutor = threadPoolExecutor,
                 drawableUtils = drawableUtils,
                 base64Utils = base64Utils,
-                webPImageCompression = webPImageCompression
+                webPImageCompression = webPImageCompression,
+                base64LruCache = base64LruCache,
+                logger = logger
             )
 
         private companion object {
@@ -168,5 +210,12 @@ class Base64Serializer private constructor(
     internal companion object {
         @VisibleForTesting
         internal const val BITMAP_SIZE_LIMIT_BYTES = 15000 // 15 kbs
+
+        internal const val DOES_NOT_IMPLEMENT_COMPONENTCALLBACKS =
+            "Cache instance does not implement ComponentCallbacks2"
+
+        // The cache is a singleton, so we want to share this flag among
+        // all instances so that it's registered only once
+        private var isCacheRegisteredForCallbacks: Boolean = false
     }
 }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
@@ -1,0 +1,14 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+internal interface Cache<K : Any, V : Any> {
+    fun put(element: K, value: V)
+    fun get(element: K): V?
+    fun size(): Int
+    fun clear()
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
@@ -6,6 +6,7 @@
 
 package com.datadog.android.sessionreplay.internal.recorder.mapper
 
+import android.content.Context
 import android.graphics.drawable.ColorDrawable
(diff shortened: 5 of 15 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -42,6 +43,12 @@ class Base64Serializer private constructor(
     ) {
         asyncImageProcessingCallback?.startProcessingImage()
 
+        val cachedBase64 = base64LruCache.get(drawable.hashCode())
```
mariusc83, line 46: in order to avoid collisions you could better rely on `System.identityHashCode(drawable)` which always returns the memory address of that reference.

mariusc83: as we discussed this will work for non `BitmapDrawables`  but for those you will have to use something like `System.identityHashCode(bitmap)`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+
+internal object Base64LRUCache : Cache {
+    private const val MAX_ITEMS = 1000
+    private var cache = LruCache<Int, String>(MAX_ITEMS)
```
mariusc83, line 14: are you sure you want to base it on number of items ? Why not on bytes ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,42 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+
+internal object Base64LRUCache : Cache {
```
mariusc83, line 12: This is nice but let's make it implement also the `ComponentCallbacks2` and listen to the system resources status. Based on the trim level you can decide how eagerly you much memory you could clear from the cache. For more info:
https://developer.android.com/topic/performance/memory
https://github.com/bumptech/glide/blob/fd96de2dd20541c46b3eee24e312969830bc17fb/library/src/main/java/com/bumptech/glide/load/engine/bitmap_recycle/LruBitmapPool.java#L244

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
mariusc83: You could leverage the :
`Runtime.maxMemory` which returns the max amount of memory in bytes allocated for current JVM and then take a part of it (e.g. Runtime.maxMemory/8)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: If I understand correctly, if we go by memory address then two separate drawables that resolve to the same bitmap would require separate cache entries? The contract for hashcode specifies that if the objects are equal then they get the same id.

mariusc83: hmm, if you check the `hashCode` function into the `Object` class you will see that they delegate the same `identityHashCode` static function which is also called by `System.identityHashCode`. What I am suggesting here is to use that directly in order to avoid situations where users will provide their own implementation of a `Drawable` where they override the `hashCode` function and by doing that they could create collisions in your cache.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -99,6 +106,11 @@ internal class RecordedDataQueueHandler {
         val rumContextData = rumContextDataHandler.createRumContextData()
             ?: return null
 
+        // if the view changed then clear the drawable lrucache
+        if (rumContextData.prevRumContext != rumContextData.newRumContext) {
```
mariusc83, line 110: good call

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Synchronized
+    override fun isEmpty(): Boolean =
+        cache.size() == 0
+
+    private fun generateKey(drawable: Drawable): Int =
+        System.identityHashCode(drawable)
+
+    override fun onTrimMemory(level: Int) {
```
mariusc83, line 63: niceee

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,99 @@
...
+
+import android.content.ComponentCallbacks2
+import android.content.res.Configuration
+import android.graphics.drawable.Drawable
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+
+@Suppress("MagicNumber")
+internal object Base64LRUCache : Cache, ComponentCallbacks2 {
+    private val CACHE_MEMORY_SIZE = (Runtime.getRuntime().maxMemory() / 8).toInt()
```
jonathanmos, line 17: What is everyone's opinion about these memory thresholds + the eviction strategy for the cache? Should we be less aggressive in evicting - e.g. only evict 50% of entries in critical memory situations? I see that Glide, for example, only evicts 50% when critical.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
mariusc83: you want to keep the `hashCode` here ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,100 @@
...
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.content.ComponentCallbacks2
+import android.content.res.Configuration
+import android.graphics.drawable.Drawable
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+
+@Suppress("MagicNumber")
+internal object Base64LRUCache : Cache, ComponentCallbacks2 {
```
mariusc83, line 16: I honestly do not see where do you register this as a `ComponentCallback`. It needs to be registered in the `Context.registerComponentCallback`. Also please make sure you test this feature:
https://github.com/schedjoules/android-event-discovery-sdk/issues/330

mariusc83 (changes requested):
We need to test and register that callback

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -34,15 +36,18 @@ internal class RecordedDataQueueHandler {
...
     private var rumContextDataHandler: RumContextDataHandler
     private var timeProvider: TimeProvider
+    private var base64LruCache: Cache
 
     internal constructor(
         processor: RecordedDataProcessor,
         rumContextDataHandler: RumContextDataHandler,
-        timeProvider: TimeProvider
+        timeProvider: TimeProvider,
+        base64LruCache: Cache = Base64LRUCache
```
0xnm, line 45: minor: `Cache` type doesn't say that it is `LRU`, so to not leak the exact implementation and avoid confusion in the future (maybe at some point we switch from `LRU` do something else), I propose to rename `base64LruCache` to just `base64Cache` everywhere in this class.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
0xnm: is it needed? given that we convert small icons which can be re-used between the views, maybe we can keep them?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
0xnm: regarding max memory size: That is quite a lot for our cache imo, given that our SDK is not the only citizen in the application process and there are obviously image libraries which create cache as well. Should we maybe go lower?

Also I guess this constant should be called `MAX_CACHE_MEMORY_SIZE_BYTES` (to highlight that it is the max value and also the unit)?.

Regarding eviction strategy: it is not that important, so I guess 50% is okay.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,100 @@
...
+    @Suppress("UnsafeThirdPartyFunctionCall") // maxMemory does not throw
+    private val CACHE_MEMORY_SIZE = (Runtime.getRuntime().maxMemory() / 8).toInt()
+    private val ON_LOW_MEMORY_SIZE = CACHE_MEMORY_SIZE / 2 // 50% size
+    private val ON_MODERATE_MEMORY_SIZE = (CACHE_MEMORY_SIZE / 4) * 3 // 75% size
+
+    private var cache: LruCache<Int, String> = object : LruCache<Int, String>(CACHE_MEMORY_SIZE) {
+        override fun sizeOf(key: Int?, value: String): Int {
+            return value.toByteArray().size
```
0xnm, line 24: `toByteArray()` will involve new byte array allocation each time (according to `String.getBytes`), so effectively increasing memory footprint of our cache 2x if this method is called frequently.

Can we optimize it somehow? Not have a solution in my head, just throwing in the idea for optimization.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,100 @@
...
+    override fun clear() {
+        cache.evictAll()
+    }
+
+    @Synchronized
+    override fun isEmpty(): Boolean =
+        cache.size() == 0
+
+    private fun generateKey(drawable: Drawable): Int =
+        System.identityHashCode(drawable)
```
0xnm, line 63: I'm wondering how we handle mutable drawables, like `ColorStateDrawable`. Obviously look of such drawable may be different (depending on the state), but hash code will be the same.

Or all the drawables coming to this layer are immutable?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    override fun onConfigurationChanged(newConfig: Configuration) {}
+
+    override fun onLowMemory() {
+        cache.evictAll()
+    }
+
+    @VisibleForTesting
+    internal fun setBackingCache(cache: LruCache<Int, String>) {
```
0xnm, line 35: minor: maybe just a provider `(Int) -> LruCache` in the constructor with default value for prod and mock value for tests?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.graphics.drawable.Drawable
+
+internal interface Cache {
+    fun put(drawable: Drawable, base64String: String)
+    fun get(key: Int): String?
```
0xnm, line 13: Should we make `key: Drawable` here to be aligned with the signature of `put` method?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
```diff
@@ -0,0 +1,17 @@
...
+    fun size(): Int
+    fun clear()
+    fun isEmpty(): Boolean
```
0xnm, line 16: is it used?

0xnm (commented):
nice idea! I've added some comments to discuss.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
mariusc83: yes completely agree with @0xnm here. I think you need a reliable identifier / Drawable type.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
mariusc83: I am more worried about the big ones here but that's a good question indeed.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
jonathanmos: Yes, I understand the direction. Base64 takes up 6 of 8 bits, so 0.75 * the length of the string would give us the size without creating a bytearray

0xnm: I think it will be the simply the length of the string, because even if we take only only lower half of the byte space, the atomic unit is still the byte.

Moreover, internally JVM at least (not sure about ART though) stores Strings in UTF-16 format, so probably the native size of the object will be `2 * length`? ignoring the fields of the `String` object itself (which is just few bytes).

jonathanmos: Thanks - it is indeed string length due to utf-8. I'll make these changes and it should improve memory quite a lot

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
jonathanmos: I'll go over the types of drawables and see if I can find a strategy of prepending data to the key. There are going to be some complicated cases like `AnimationDrawable` and `LayerDrawable`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
jonathanmos: I missed that - I'll register for callbacks in handleBitmap

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,100 @@
...
+
+import android.content.ComponentCallbacks2
+import android.content.res.Configuration
+import android.graphics.drawable.Drawable
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+
+@Suppress("MagicNumber")
+internal object Base64LRUCache : Cache, ComponentCallbacks2 {
+    @Suppress("UnsafeThirdPartyFunctionCall") // maxMemory does not throw
```
xgouchet, line 17: Do not `@Suppress` this warning, instead you need to add the `Runtime.getRuntime()` and `Runtime.maxMemory()` methods in the `detekt_custom.yml` in the `knownSafeCalls` section

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
jonathanmos: not used. I put it in as a possibly useful utility for the class, but I'll remove it until we actually need it

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
jonathanmos: With image downscaling it will be possible to use a small cache size and still handle a relatively large number of images, so I've set the cache to 4mb (as we did on ios)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
jonathanmos: I've added prefixing to the keys for LayerDrawables and DrawableContainers

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,135 @@
...
+    override fun put(drawable: Drawable, base64String: String) {
+        val key = generateKey(drawable)
+        val value = base64String.toByteArray()
+        cache.put(key, value)
+    }
+
+    @Synchronized
+    override fun get(drawable: Drawable): String? =
+        cache.get(generateKey(drawable))?.let {
+            String(it)
```
0xnm, line 96: seems like unnecessary allocation here - we don't need a string down the line, because anyway we store all the data in the byte array, so I guess we create it here for nothing. It is better to return `ByteArray` from here directly imo.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    @VisibleForTesting
+    internal fun setBackingCache(cache: LruCache<String, ByteArray>) {
+        this.cache = cache
+    }
+
+    @Synchronized
+    override fun put(drawable: Drawable, base64String: String) {
+        val key = generateKey(drawable)
+        val value = base64String.toByteArray()
```
0xnm, line 89: minor: it is better to be precise which encoding to use:

```suggestion
        val value = base64String.toByteArray(Charsets.UTF_8)
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    private fun generatePrefix(drawable: Drawable): String {
+        return when (drawable) {
+            is DrawableContainer -> getPrefixForDrawableContainer(drawable)
+            is LayerDrawable -> getPrefixForLayerDrawable(drawable)
+            else -> ""
+        }
+    }
+
+    private fun getPrefixForDrawableContainer(drawable: DrawableContainer) =
```
0xnm, line 118: just a thought: in theory with something like `AnimationDrawable` we can fill our cache very fast, should we support drawable like that?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,135 @@
...
+import android.content.ComponentCallbacks2
+import android.content.res.Configuration
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.DrawableContainer
+import android.graphics.drawable.LayerDrawable
+import android.util.LruCache
+import android.view.View
+import androidx.annotation.VisibleForTesting
+
+@Suppress("MagicNumber")
```
0xnm, line 18: it is better to apply this annotation on the individual fields, rather than on a class.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -27,7 +28,8 @@ class Base64Serializer private constructor(
     private val threadPoolExecutor: ExecutorService,
     private val drawableUtils: DrawableUtils = DrawableUtils(),
     private val base64Utils: Base64Utils = Base64Utils(),
-    private val webPImageCompression: ImageCompression = WebPImageCompression()
+    private val webPImageCompression: ImageCompression = WebPImageCompression(),
+    private val base64LruCache: Cache = Base64LRUCache
```
0xnm, line 32: I guess default parameters are not needed here? because this is a private constructor which is only invoked from `Base64Serializer.Builder.build` with all the necessary parameters provided, it is not invoked from the tests.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
```diff
@@ -103,10 +103,12 @@ abstract class BaseWireframeMapper<T : View, S : MobileSegment.Wireframe>(
      */
     @MainThread
     protected fun handleBitmap(
+        view: View,
```
0xnm, line 106: maybe rather pass `Context` than a `View`? It seems more reasonable. Or `View` will be needed at some point in future?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
jonathanmos: I'm returning `String` here because we're using that eventually in the wireframe property, however I can change to`ByteArray` so that the cache won't perform the conversion

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,135 @@
...
+internal object Base64LRUCache : Cache, ComponentCallbacks2 {
```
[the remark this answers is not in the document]
0xnm: Indeed, seems like `ImageWireframe#base64` requires it to be a string. In this case let's return a `String` indeed (while it will create a copy of byte array, but probably we don't have a way to create a String pointing to existing byte array without allocating a copy, because `String` is immutable and our array can be mutable).

We could store things as `String`, to avoid this allocation, but then it will be 2x memory if Android stores internally strings as UTF-16 as desktop JVMs do.

Some people had the similar idea about string storage https://stackoverflow.com/questions/54393179/does-android-use-string-compacting-behind-the-scenes, but no reply.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.graphics.drawable.Drawable
+import android.view.View
+
+internal interface Cache {
+    fun registerComponentCallbacksIfNecessary(view: View)
+    fun put(drawable: Drawable, base64String: String)
```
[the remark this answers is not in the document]
jonathanmos: I'll modify `Cache` to use generics

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
jonathanmos: Good catch - I'll exclude AnimationDrawable explicitly for now

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,135 @@
...
+                cache.evictAll()
+            }
+
+            else -> {
+                cache.evictAll()
+            }
+        }
+    }
+
+    override fun registerComponentCallbacksIfNecessary(view: View) {
```
mariusc83, line 68: you should mark this method as @MainThread if it is only called from main otherwise you should synchronise it.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.graphics.drawable.Drawable
+import android.view.View
+
+internal interface Cache {
+    fun registerComponentCallbacksIfNecessary(view: View)
```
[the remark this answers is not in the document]
mariusc83: yup I do agree with this, this is outside of its responsibility.

mariusc83 (approved):
LGTM now. Nicely done.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
```diff
@@ -27,7 +27,7 @@ internal class ImageButtonMapper(
         view: ImageButton,
         mappingContext: MappingContext
     ): List<MobileSegment.Wireframe.ImageWireframe> {
-        val drawable = view.drawable
+        val drawable = view.drawable?.constantState?.newDrawable()
```
0xnm, line 30: Why this call is needed? As I see, `constantState` can return `null`, won't we loose drawables from some drawables? Also `newDrawable` is saying the following:

```
Note that using this means the density-dependent drawables (like bitmaps) will not be able to update their target density correctly. One should use {@link #newDrawable(Resources)} instead to provide a resource.
```
Shouldn't we use `newDrawable(Resources)` instead?

0xnm (commented):
I've added some questions and suggestions, but overall I think it is good to go.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
jonathanmos: Needed because drawing to canvas using the original drawable seems to mutate the view in the app - this became quite visible with downscaling

0xnm: I believe that for __certain__ drawables (which have the state) call to `draw` can indeed modify their state (drawable itself don't have a link to the view and the canvas passed to the `draw` method is owned by us, so it is just about the state of drawable). I don't think we need this allocation for the static (read __stateless__) drawables, so I guess this is the field where we can optimize later.

Let's create a task for optimization to look deeper into this topic for later and keep it like that for now maybe?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -0,0 +1,134 @@
...
+    private fun generatePrefix(drawable: Drawable): String {
+        return when (drawable) {
+            is DrawableContainer -> getPrefixForDrawableContainer(drawable)
+            is LayerDrawable -> getPrefixForLayerDrawable(drawable)
+            else -> ""
+        }
+    }
+
+    private fun getPrefixForDrawableContainer(drawable: DrawableContainer): String {
+        if (drawable !is AnimationDrawable) {
```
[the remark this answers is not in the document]
jonathanmos: Yes, its just a fix to avoid having `AnimationDrawable` take up too much space in the lrucache. We'll need to have a discussion/brainstorming if we decide we want to send animations with base64, and this will in any case cease to be an issue once we have the resources endpoint.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Cache.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+internal interface Cache<A : Any, B : Any, V : Any> {
```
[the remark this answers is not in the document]
jonathanmos: I'll change to K,V

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -100,8 +103,16 @@ class Base64Serializer private constructor(
 
     @MainThread
     private fun registerCacheForCallbacks(applicationContext: Context) {
-        if (!isCacheRegisteredForCallbacks) {
-            applicationContext.registerComponentCallbacks(base64LruCache as ComponentCallbacks2)
+        if (base64LruCache !is ComponentCallbacks2) {
```
0xnm, line 106: minor: probably we can do the check for `ComponentCallbacks2` only once:

```
if (isCacheRegisteredForCallbacks) return

if (base64LruCache is ComponentCallbacks2) {
  applicationContext.registerComponentCallbacks(base64LruCache)
  isCacheRegisteredForCallbacks = true
} else {
  internalLogger.log(...)
}

```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -168,5 +208,10 @@ class Base64Serializer private constructor(
     internal companion object {
         @VisibleForTesting
         internal const val BITMAP_SIZE_LIMIT_BYTES = 15000 // 15 kbs
+
+        internal const val DOES_NOT_IMPLEMENT_COMPONENTCALLBACKS =
+            "LruCache does not implement ComponentCallbacks2"
```
0xnm, line 213: The type is just `Cache`, so no necessary it is `LRU`.

```suggestion
            "Cache instance does not implement ComponentCallbacks2"
```

0xnm (commented):
I added a comment about `internalLogger` storage, it is a potential memory leak.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -225,6 +227,8 @@ internal class SessionReplayFeature constructor(
                 maxBatchSize = 10 * 1024 * 1024
             )
 
+        internal var logger: InternalLogger? = null
```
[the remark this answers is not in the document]
jonathanmos: I'm setting temporarily to unbound logger as we discussed

0xnm (approved):

PR #1538, opened 2023-07-12
RUMM-3445: Support `navHost` hosted by `FragmentContainerView` for `NavigationViewTrackingStrategy`

0xnm:
### What does this PR do?

This PR addresses the issue described [here](https://github.com/DataDog/dd-sdk-android/issues/760#issuecomment-1511050310). Basically now the [recommendation](https://developer.android.com/guide/navigation/get-started#add-navhostfragment) is to use `FragmentContainerView` for hosting `NavHostFragment`, this now will be the priority for lookup, and if not found we fall back to the old strategy.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategy.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategy.kt
@@ -180,12 +180,24 @@ class NavigationViewTrackingStrategy(
     @Suppress("SwallowedException")
     private fun Activity.findNavControllerOrNull(@IdRes viewId: Int): NavController? {
         return try {
-            Navigation.findNavController(this, viewId)
+            val navController = if (this is FragmentActivity) {
+                findNavControllerFromNavHostFragmentOrNull(viewId)
+            } else {
+                null
+            }
+            navController ?: Navigation.findNavController(this, viewId)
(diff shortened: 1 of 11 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategy.kt
```diff
@@ -180,14 +180,26 @@ class NavigationViewTrackingStrategy(
     @Suppress("SwallowedException")
     private fun Activity.findNavControllerOrNull(@IdRes viewId: Int): NavController? {
         return try {
-            Navigation.findNavController(this, viewId)
+            val navController = if (this is FragmentActivity) {
```
mariusc83, line 183: can you name it more specifically -> fragmentNavController

mariusc83 (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/NavigationViewTrackingStrategy.kt
0xnm: nav controller is in the fragment in any case, the difference is just that with old approach it is in user-created fragment, and with new approach it is in the lib-created fragment which is controlled by `FragmentContainerView`, so not sure if rename is worth it.

xgouchet (approved):

PR #1546, opened 2023-07-16
REPLAY- 1893: Implement bitmap downscaling

jonathanmos:
### What does this PR do?
Implement bitmap downscaling similar to what was done on iOS, aiming to end up with a bitmap of 10kb or less. We calculate the expected dimensions sizes for a perfect square with the required filesize, and then work out the dimensions of the resized bitmap from that.

Things to consider: 

- We are later compressing the bitmaps to webp and this gives additional savings in file size, but it's difficult to predict in advance how much the savings will be. Should we be taking this into account in the scaling of the raw bitmap?

Some numbers from testing:
Initial bitmap size: 26544 
Size of bitmap after downscaling: 9400
After webp compression: 3136

### Motivation
This pr will allow us to more accurately size the lrucache

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
@@ -55,7 +55,7 @@ class Base64Serializer private constructor(
             return
         }
 
-        val bitmap = drawableUtils.createBitmapFromDrawable(drawable, displayMetrics)
+        val bitmap = drawableUtils.createBitmapOfApproxSizeFromDrawable(drawable, displayMetrics)
 
         if (bitmap == null) {
             asyncImageProcessingCallback?.finishProcessingImage()
@@ -73,10 +73,6 @@ class Base64Serializer private constructor(
     ) {
         this.asyncImageProcessingCallback = asyncImageProcessingCallback
     }
-
-    @VisibleForTesting
-    internal fun isOverSizeLimit(bitmapSize: Int): Boolean =
-        bitmapSize > BITMAP_SIZE_LIMIT_BYTES
 
     // endregion
 
@@ -120,25 +116,22 @@ class Base64Serializer private constructor(
 
     @WorkerThread
     private fun convertBmpToBase64(drawable: Drawable, bitmap: Bitmap): String {
+        val base64Result: String
+
         val byteArrayOutputStream = webPImageCompression.compressBitmapToStream(bitmap)
 
-        if (isOverSizeLimit(byteArrayOutputStream.size())) {
-            return ""
-        }
-
-        val base64String: String
         try {
-            base64String = base64Utils.serializeToBase64String(byteArrayOutputStream)
+            base64Result = base64Utils.serializeToBase64String(byteArrayOutputStream)
 
-            if (base64String.isNotEmpty()) {
+            if (base64Result.isNotEmpty()) {
                 // if we got a base64 string then cache it
-                base64LruCache.put(drawable, base64String)
+                base64LruCache.put(drawable, base64Result)
             }
         } finally {
             bitmap.recycle()
         }
 
-        return base64String
+        return base64Result
     }
 
     private fun finalizeRecordedDataItem(
@@ -208,9 +201,6 @@ class Base64Serializer private constructor(
     // endregion
 
     internal companion object {
-        @VisibleForTesting
-        internal const val BITMAP_SIZE_LIMIT_BYTES = 15000 // 15 kbs
-
         internal const val DOES_NOT_IMPLEMENT_COMPONENTCALLBACKS =
             "Cache instance does not implement ComponentCallbacks2"
 
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
@@ -6,7 +6,9 @@
 
 package com.datadog.android.sessionreplay.internal.recorder.mapper
 
+import android.graphics.drawable.Drawable
 import android.widget.ImageButton
+import com.datadog.android.sessionreplay.internal.recorder.GlobalBounds
 import com.datadog.android.sessionreplay.internal.recorder.MappingContext
 import com.datadog.android.sessionreplay.internal.recorder.base64.Base64Serializer
 import com.datadog.android.sessionreplay.internal.recorder.base64.ImageCompression
@@ -18,7 +20,7 @@ internal class ImageButtonMapper(
     webPImageCompression: ImageCompression = WebPImageCompression(),
     base64Serializer: Base64Serializer = Base64Serializer.Builder().build(),
     uniqueIdentifierGenerator: UniqueIdentifierGenerator = UniqueIdentifierGenerator
-) : BaseWireframeMapper<ImageButton, MobileSegment.Wireframe.ImageWireframe>(
+) : BaseWireframeMapper<ImageButton, MobileSegment.Wireframe>(
     webPImageCompression = webPImageCompression,
     base64Serializer = base64Serializer,
     uniqueIdentifierGenerator = uniqueIdentifierGenerator
@@ -26,7 +28,7 @@ internal class ImageButtonMapper(
     override fun map(
         view: ImageButton,
         mappingContext: MappingContext
-    ): List<MobileSegment.Wireframe.ImageWireframe> {
+    ): List<MobileSegment.Wireframe> {
         val resources = view.resources
         val drawable = view.drawable?.constantState?.newDrawable(resources)
         val id = resolveChildDrawableUniqueIdentifier(view)
@@ -39,22 +41,51 @@ internal class ImageButtonMapper(
         val (shapeStyle, border) = view.background?.resolveShapeStyleAndBorder(view.alpha)
             ?: (null to null)
 
+        val wireframes = mutableListOf<MobileSegment.Wireframe>()
+
+        // if the drawable has no width/height then there's no point trying to get a bitmap
+        if (drawable.intrinsicWidth > 0 && drawable.intrinsicHeight > 0) {
+            val imageWireframe = resolveImageWireframe(
+                view,
+                id,
+                bounds,
+                shapeStyle,
+                border,
+                drawable
+            )
+            wireframes.add(imageWireframe)
+        }
+
+        return wireframes
+    }
+
+    // region internal
+
+    private fun resolveImageWireframe(
+        view: ImageButton,
+        id: Long,
+        bounds: GlobalBounds,
+        shapeStyle: MobileSegment.ShapeStyle?,
+        border: MobileSegment.ShapeBorder?,
+        drawable: Drawable
+    ): MobileSegment.Wireframe.ImageWireframe {
         val mimeType = getWebPMimeType()
         val displayMetrics = view.resources.displayMetrics
         val applicationContext = view.context.applicationContext
 
-        val imageWireframe = MobileSegment.Wireframe.ImageWireframe(
-            id = id,
-            x = bounds.x,
-            y = bounds.y,
-            width = bounds.width,
-            height = bounds.height,
-            shapeStyle = shapeStyle,
-            border = border,
-            base64 = "",
-            mimeType = mimeType,
-            isEmpty = true
-        )
+        val imageWireframe =
+            MobileSegment.Wireframe.ImageWireframe(
+                id = id,
+                x = bounds.x,
+                y = bounds.y,
+                width = bounds.width,
+                height = bounds.height,
+                shapeStyle = shapeStyle,
+                border = border,
+                base64 = "",
+                mimeType = mimeType,
+                isEmpty = true
+            )
 
         @Suppress("ThreadSafety") // TODO REPLAY-1861 caller thread of .map is unknown?
         handleBitmap(
@@ -64,6 +95,8 @@ internal class ImageButtonMapper(
             imageWireframe = imageWireframe
         )
 
-        return listOf(imageWireframe)
+        return imageWireframe
     }
+
+    // endregion
 }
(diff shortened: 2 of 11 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -31,4 +32,46 @@ internal class DrawableUtils(
         drawable.draw(canvas)
         return bitmap
     }
+
+    internal fun downscaleBitmapToApproxSize(
```
jonathanmos, line 36: perhaps this should be moved to a new BitmapUtils class, because it doesn't actually involve drawables

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -62,8 +62,8 @@ class Base64Serializer private constructor(
     }
 
     @VisibleForTesting
-    internal fun isOverSizeLimit(bitmapSize: Int): Boolean =
-        bitmapSize > BITMAP_SIZE_LIMIT_BYTES
```
0xnm, line 66: probably this change can be reverted, we can even name argument here just `size`.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -87,21 +87,24 @@ class Base64Serializer private constructor(
...
     @WorkerThread
-    private fun convertBmpToBase64(bitmap: Bitmap): String {
-        val byteArrayOutputStream = webPImageCompression.compressBitmapToStream(bitmap)
+    private fun convertBmpToBase64(initialBitmap: Bitmap): String {
+        var base64Result = ""
 
-        if (isOverSizeLimit(byteArrayOutputStream.size())) {
-            return ""
-        }
+        val scaledBitmap = drawableUtils.downscaleBitmapToApproxSize(initialBitmap)
```
0xnm, line 93: I would add what `approximate` means in this context, maybe it is worth adding the doc / explanation to the method?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -87,21 +87,24 @@ class Base64Serializer private constructor(
     }
 
     @WorkerThread
-    private fun convertBmpToBase64(bitmap: Bitmap): String {
-        val byteArrayOutputStream = webPImageCompression.compressBitmapToStream(bitmap)
+    private fun convertBmpToBase64(initialBitmap: Bitmap): String {
```
0xnm, line 90: in this flow we create 2 bitmaps: first bitmap to draw content of drawable, second bitmap to downscale. I'm wondering if we can do this in 1 pass: can we create a canvas which will apply affine transformation (scale)?

In theory downscaling is just applying the matrix, so maybe we can create a downscaled bitmap straight away (we know size of the drawable, so we should be able to do it), and then provide a canvas which will apply the necessary transformation? This would greatly reduce memory usage.

0xnm (commented):
I left one comment wondering if we can use only 1 bitmap during the image conversion flow, I guess it should be possible.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -87,21 +87,24 @@ class Base64Serializer private constructor(
...
-        try {
-            base64 = base64Utils.serializeToBase64String(byteArrayOutputStream)
-        } finally {
-            bitmap.recycle()
+        scaledBitmap?.let {
+            val byteArrayOutputStream = webPImageCompression.compressBitmapToStream(it)
+
+            try {
+                if (!isOverSizeLimit(byteArrayOutputStream.size())) {
+                    base64Result = base64Utils.serializeToBase64String(byteArrayOutputStream)
```
mariusc83, line 100: maybe we want a log here in case is over the limit ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
jonathanmos: In the end not necessary, since refactoring to perform the downscaling when the bitmap is first created removes this method

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: I'm removing the method, since we now know the bitmap is already downscaled under the size limit, and compressing to webp can only make it smaller. We could also move the size limit method to later in order to test the serialized string for size, however, base64 adds a constant 30% to the bitmap size so we can already predict it and handle it by making the scaled bitmap size at least 30% smaller than the limit.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/wrappers/BitmapWrapper.kt
```diff
@@ -27,4 +27,25 @@ internal class BitmapWrapper {
             null
         }
     }
+
+    internal fun createScaledBitmap(
```
mariusc83, line 31: I wonder why do you need the `createBitmap` above ? As that one only delegates to this one.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtilsTest.kt
```diff
@@ -105,26 +109,57 @@ internal class DrawableUtilsTest {
...
+            assertThat(it).isEqualTo(50)
+        }
+
+        assertThat(displayMetricsCaptor.firstValue).isEqualTo(mockDisplayMetrics)
+
+        assertThat(result).isEqualTo(mockBitmap)
+    }
+
+    @Test
+    fun `M return null W createBitmapFromDrawableOfApproxSize() { failed to create bmp }`() {
```
mariusc83, line 150: I think you are missing the test where the `canvasWrapper` returns null ?

mariusc83 (approved):
LGTM

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -13,22 +13,67 @@ import android.util.DisplayMetrics
...
+            displayMetrics,
+            dimensions.first,
+            dimensions.second,
+            Bitmap.Config.ARGB_8888
+        )
             ?: return null
 
         val canvas = canvasWrapper.createCanvas(bitmap) ?: return null
         drawable.setBounds(0, 0, canvas.width, canvas.height)
         drawable.draw(canvas)
```
0xnm, line 46: so how does writing the scaled bitmap works? I'm asking, because it seems we don't specify any transformation matrix. Say we have `BitmapDrawable` and its bitmap is 1000x1000, will it be properly recorded to the bitmap of size 100x100?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -13,22 +13,67 @@ import android.util.DisplayMetrics
...
+                width = (maxSize * bitmapRatio).toInt()
+            }
+        }
+
+        return Pair(width, height)
+    }
+
+    internal companion object {
+        internal const val DESIRED_BITMAP_SIZE_IN_BYTES = 10240 // 10kb
+        private const val ARGB_8888_DEPTH = 4
```
0xnm, line 77: a bit more descriptive naming:

```suggestion
        private const val ARGB_8888_PIXEL_SIZE_BYTES = 4
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -13,22 +13,67 @@ import android.util.DisplayMetrics
...
+        drawable: Drawable,
+        requestedSizeInBytes: Int
+    ): Pair<Int, Int> {
+        var width = if (drawable.intrinsicWidth <= 0) 1 else drawable.intrinsicWidth
+        var height = if (drawable.intrinsicHeight <= 0) 1 else drawable.intrinsicHeight
+        val sizeAfterCreation = width * height * ARGB_8888_DEPTH
+
+        if (sizeAfterCreation > requestedSizeInBytes) {
+            val bitmapRatio = width.toDouble() / height.toDouble()
+            val perfectSquareSize = (requestedSizeInBytes / ARGB_8888_DEPTH).toDouble()
```
0xnm, line 60:
```suggestion
            val totalMaxPixels = (requestedSizeInBytes / ARGB_8888_DEPTH).toDouble()
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -13,22 +13,67 @@ import android.util.DisplayMetrics
...
         drawable.setBounds(0, 0, canvas.width, canvas.height)
         drawable.draw(canvas)
         return bitmap
     }
+
+    private fun getScaledWidthAndHeight(
+        drawable: Drawable,
+        requestedSizeInBytes: Int
+    ): Pair<Int, Int> {
+        var width = if (drawable.intrinsicWidth <= 0) 1 else drawable.intrinsicWidth
```
0xnm, line 54: in which case `intrinsicWidth` can be 0 or negative? if we get `width` and `height` equal to 1, we are going to send this 1 pixel to backend? seems a bit strange

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -13,22 +13,67 @@ import android.util.DisplayMetrics
...
+            } else {
+                width = (maxSize * bitmapRatio).toInt()
+            }
+        }
+
+        return Pair(width, height)
+    }
+
+    internal companion object {
+        internal const val DESIRED_BITMAP_SIZE_IN_BYTES = 10240 // 10kb
```
0xnm, line 76: maybe max bitmap size? because `desired` has a connotation that we may upscale bitmap if size is below

```suggestion
        internal const val MAX_BITMAP_SIZE_IN_BYTES = 10240 // 10kb
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
jonathanmos: It seems to me that there's is automatic scaling applied. `Drawable.draw` is calling `drawBitmap`, and the comment for this method says: "Draw the specified bitmap, scaling/translating automatically to fill the destination rectangle."

0xnm: I'm wondering if scaling also applies to other draw operation (`drawRect`, `drawCircle`, etc.) which take coordinates as parameters.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
0xnm: yes, I would pull it up probably, because if we don't have drawable size we cannot do anything anyway.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
```diff
@@ -39,22 +43,58 @@ internal class ImageButtonMapper(
...
+            val imageWireframe = resolveImageWireframe(
+                view,
+                id,
+                bounds,
+                shapeStyle,
+                border,
+                drawable
+            )
+            wireframes.add(imageWireframe)
+        } else {
```
jonathanmos, line 58: This raises the question - is it better not to return a wireframe at all in this case, or should we return a shape with the bounds of the view?

mariusc83: we should not return as it will not have any visible effect in the player.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
mariusc83: This can be tested inside the sample app by adding these particular cases.

jonathanmos: I've added some shapes to the sample app

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
```diff
@@ -18,15 +20,17 @@ internal class ImageButtonMapper(
...
     webPImageCompression = webPImageCompression,
     base64Serializer = base64Serializer,
     uniqueIdentifierGenerator = uniqueIdentifierGenerator
 ) {
     override fun map(
         view: ImageButton,
         mappingContext: MappingContext
-    ): List<MobileSegment.Wireframe.ImageWireframe> {
+    ): List<MobileSegment.Wireframe> {
+        val wireframes = mutableListOf<MobileSegment.Wireframe>()
```
0xnm, line 32: minor: it is better to move declaration closer to the usage site (L46)

0xnm (approved):

PR #1554, opened 2023-07-25
REPLAY-1890: Implement pool of reusable bitmaps

jonathanmos:
### What does this PR do?
Adds a reusable bitmap pool for base64.

### Motivation
Reduce the number of new bitmaps created.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/build.gradle.kts
+++ b/features/dd-sdk-android-session-replay/build.gradle.kts
@@ -27,6 +27,7 @@ plugins {
     id("com.github.ben-manes.versions")
 
     // Tests
+    id("de.mobilej.unmock")
     id("org.jetbrains.kotlinx.kover")
 
     // Internal Generation
@@ -110,6 +111,10 @@ dependencies {
 
 apply(from = "clone_session_replay_schema.gradle.kts")
 apply(from = "generate_session_replay_models.gradle.kts")
+
+unMock {
+    keep("android.util.LruCache")
+}
 
 kotlinConfig(jvmBytecodeTarget = JvmTarget.JVM_11)
 junitConfig()
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
@@ -6,15 +6,12 @@
 
 package com.datadog.android.sessionreplay.internal.async
 
-import android.graphics.drawable.Drawable
 import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
 import com.datadog.android.sessionreplay.internal.processor.RecordedDataProcessor
 import com.datadog.android.sessionreplay.internal.processor.RumContextDataHandler
 import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
-import com.datadog.android.sessionreplay.internal.recorder.base64.Base64LRUCache
-import com.datadog.android.sessionreplay.internal.recorder.base64.Cache
 import com.datadog.android.sessionreplay.internal.utils.TimeProvider
 import com.datadog.android.sessionreplay.model.MobileSegment
 import java.lang.ClassCastException
@@ -37,18 +34,15 @@ internal class RecordedDataQueueHandler {
     private var processor: RecordedDataProcessor
     private var rumContextDataHandler: RumContextDataHandler
     private var timeProvider: TimeProvider
-    private var cache: Cache<Drawable, String>
 
     internal constructor(
         processor: RecordedDataProcessor,
         rumContextDataHandler: RumContextDataHandler,
-        timeProvider: TimeProvider,
-        cache: Cache<Drawable, String> = Base64LRUCache
+        timeProvider: TimeProvider
     ) : this(
         processor = processor,
         rumContextDataHandler = rumContextDataHandler,
         timeProvider = timeProvider,
-        cache = cache,
 
         /**
          * TODO: RUMM-0000 consider change to LoggingThreadPoolExecutor once V2 is merged.
@@ -70,14 +64,12 @@ internal class RecordedDataQueueHandler {
         processor: RecordedDataProcessor,
         rumContextDataHandler: RumContextDataHandler,
         timeProvider: TimeProvider,
-        executorService: ExecutorService,
-        cache: Cache<Drawable, String>
+        executorService: ExecutorService
     ) {
         this.processor = processor
         this.rumContextDataHandler = rumContextDataHandler
         this.executorService = executorService
         this.timeProvider = timeProvider
-        this.cache = cache
     }
 
     // region internal
@@ -106,11 +98,6 @@ internal class RecordedDataQueueHandler {
     ): SnapshotRecordedDataQueueItem? {
         val rumContextData = rumContextDataHandler.createRumContextData()
             ?: return null
-
-        // if the view changed then clear the drawable cache
-        if (rumContextData.prevRumContext != rumContextData.newRumContext) {
-            cache.clear()
-        }
 
         val item = SnapshotRecordedDataQueueItem(
             rumContextData = rumContextData,
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
@@ -1,0 +1,191 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.content.ComponentCallbacks2
+import android.content.res.Configuration
+import android.graphics.Bitmap
+import android.graphics.Bitmap.Config
+import android.os.Build
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.sessionreplay.internal.utils.CacheUtils
+import com.datadog.android.sessionreplay.internal.utils.InvocationUtils
+import java.util.concurrent.atomic.AtomicInteger
+
+@Suppress("TooManyFunctions")
+internal object BitmapPool : Cache<String, Bitmap>, ComponentCallbacks2 {
+    private const val BITMAP_OPERATION_FAILED = "operation failed for bitmap pool"
+
+    @VisibleForTesting
+    @Suppress("MagicNumber")
+    internal val MAX_CACHE_MEMORY_SIZE_BYTES = 4 * 1024 * 1024 // 4MB
+
+    private var bitmapsBySize = HashMap<String, HashSet<Bitmap>>()
+    private var usedBitmaps = HashSet<Bitmap>()
+    private val logger = InternalLogger.UNBOUND
+    private val invocationUtils = InvocationUtils()
+    private var bitmapIndex = AtomicInteger(0)
+
+    private var cache: LruCache<String, Bitmap> = object :
+        LruCache<String, Bitmap>(MAX_CACHE_MEMORY_SIZE_BYTES) {
+        override fun sizeOf(key: String?, bitmap: Bitmap): Int {
+            return bitmap.allocationByteCount
+        }
+
+        override fun entryRemoved(
+            evicted: Boolean,
+            key: String?,
+            oldValue: Bitmap?,
+            newValue: Bitmap?
+        ) {
+            super.entryRemoved(evicted, key, oldValue, newValue)
+
+            if (oldValue != null) {
+                val dimensionsKey = generateKey(oldValue)
+                val bitmapGroup = bitmapsBySize[dimensionsKey] ?: HashSet()
+
+                invocationUtils.safeCallWithErrorLogging(
+                    logger = logger,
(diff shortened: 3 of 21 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -72,8 +74,52 @@ internal class DrawableUtils(
...
+        config: Config
+    ): Bitmap? {
+        val newBitmap = bitmapWrapper.createBitmap(displayMetrics, width, height, config)
+        newBitmap?.let {
+            bitmapPool.put(it)
+        }
+        return newBitmap
+    }
+
+    private fun registerBitmapPoolForCallbacks(applicationContext: Context) {
```
jonathanmos, line 110: Need to fix issues with pool registration (should be similar to base64serializer logic)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -107,9 +114,10 @@ internal class RecordedDataQueueHandler {
         val rumContextData = rumContextDataHandler.createRumContextData()
             ?: return null
 
-        // if the view changed then clear the drawable cache
+        // if the view changed then clear the drawable cache and bitmap pool
         if (rumContextData.prevRumContext != rumContextData.newRumContext) {
             cache.clear()
+            bitmapPool.clear()
```
mariusc83, line 120: do we need to clear it here ? What if those bitmaps can be reused for the next View ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
```diff
@@ -0,0 +1,59 @@
...
+    override fun put(value: Bitmap) {
+        val key = generateKey(value)
+        cache.put(key, value)
+    }
+
+    override fun size(): Int = cache.size()
+
+    override fun clear() = cache.evictAll()
+
+    override fun get(element: String): Bitmap? = cache.get(element)
```
mariusc83, line 36: the way I see this `BitmapPool` will be shared between threads. How will you make sure that you do not have race issues (a bitmap needs only to be re - used by one single thread at a time) ?

mariusc83 (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -72,8 +80,47 @@ internal class DrawableUtils(
...
+        config: Config
+    ): Bitmap? {
+        var bitmap = getReusableBitmapFromPool(width, height, config)
+        if (bitmap == null) {
+            bitmap = createNewBitmap(displayMetrics, width, height, config)
+        }
+        return bitmap
+    }
+
+    private fun getReusableBitmapFromPool(
```
mariusc83, line 98: why this extra method ?

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    @Test
+    fun `M return a bitmap W get() { free bitmap in cache }`() {
+        // Given
+        val width = mockBitmap.width
+        val height = mockBitmap.height
+        val config = mockConfig
+        val key = "$width-$height-$config"
+        whenever(mockLruCache.get(key)).thenReturn(mockBitmap)
```
mariusc83, line 95: why mocking the cache here and not using the logic inside by leveraging the `put` method ? This way you can test also the add to cache logic into this method.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -0,0 +1,216 @@
...
+
+        // When
+        val cacheItem = testedCache.getBitmapByProperties(width, height, config)
+
+        // Then
+        assertThat(cacheItem).isEqualTo(mockBitmap)
+    }
+
+    @Test
+    fun `M return null W get() { bitmap in cache but it's already in use }`() {
```
mariusc83, line 108: again here you should use in the `Given`: `put` and `get` methods and then verify that at the next `get` the returned value is null

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -0,0 +1,216 @@
...
+
+        // When
+        testedCache.put(mockBitmap)
+
+        // Then
+        verify(mockLruCache).put(key + "-${testedCache.bitmapIndex}", mockBitmap)
+    }
+
+    @Test
+    fun `M call LruCache put with index 2 W put() { bitmapsBySize has an item }`() {
```
mariusc83, line 145: you should not test these methods, they are part of your internal logic and here you need to test the public API. Instead here you could have tested that if you `put` 2 values in the pool and you call `get` 2 times you will get 2 valid bitmaps back. Same for all the other similar tests here.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -0,0 +1,216 @@
...
+import org.mockito.kotlin.whenever
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(ForgeConfigurator::class)
+internal class BitmapPoolTest {
```
mariusc83, line 36: There is no concurrency test for this class and we need quite a few here in order to make sure it is thread safe.

mariusc83 (changes requested):
I like the idea of backing up the `BitmapPool` with an `LruCache`. There are still things to address in here but the idea works.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
```diff
@@ -0,0 +1,134 @@
...
+import android.util.LruCache
+import androidx.annotation.VisibleForTesting
+import com.datadog.android.sessionreplay.internal.utils.CacheUtils
+import java.util.concurrent.atomic.AtomicInteger
+
+internal object BitmapPool : Cache<String, Bitmap>, ComponentCallbacks2 {
+    @Suppress("MagicNumber")
+    private val MAX_CACHE_MEMORY_SIZE_BYTES = 4 * 1024 * 1024 // 4MB
+
+    private var bitmapsBySize = HashMap<String, HashSet<Bitmap>>()
```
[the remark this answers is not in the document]
jonathanmos: It's a problem to inject  for testing because BitmapPool is a singleton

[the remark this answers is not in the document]
mariusc83: yup I just noticed that, I think @0xnm will challenge you on that and ask why don't you use a non - singleton instance here and just make sure you inject one single instance when you initialize the SR feature.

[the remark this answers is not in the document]
jonathanmos: yes, I agree although it's the same issue that we had with base64lrucache. To use a single instance I would have to hold that instance higher up, and then it needs to be injected into base64serializer. Base64serializer is constructed in the baseMapper so that all mappers can make use of it, without us modifying the signature of map() - but since it's constructed there we have a problem injecting dependencies.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
```diff
@@ -0,0 +1,134 @@
...
+    @Synchronized
+    override fun clear() = cache.evictAll()
+
+    @Synchronized
+    override fun get(element: String): Bitmap? {
+        var bitmap: Bitmap? = null
+        val bitmapsOfRightSize = bitmapsBySize.get(element)
+        if (!bitmapsOfRightSize.isNullOrEmpty()) {
+            // filter out all the used bitmaps
+            val freeBitmaps = bitmapsOfRightSize.filterNot {
```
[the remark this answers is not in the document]
jonathanmos: You're right, there is find() - I'll refactor the block

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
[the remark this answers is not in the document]
mariusc83: I see your point and it might be the easiest way. On the other hand if we would find a way to use non - singleton instances here we will be able to solve couple of things:
- get rid of those testing methods
- eliminate the problem you raised about having collisions if you use 2 datadog instances (2 SR instances sharing same singleton). 
I am not saying that you should do this here but just try to assess the effort and evaluate if it's really doable here or if you want to keep it for later.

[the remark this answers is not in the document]
jonathanmos: I'll refactor the cache objects to single class instances in the next pr

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
jonathanmos: I wanted to make the intent of what was happening here clearer

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
```diff
@@ -0,0 +1,197 @@
...
+    @VisibleForTesting
+    @Suppress("MagicNumber")
+    internal val MAX_CACHE_MEMORY_SIZE_BYTES = 4 * 1024 * 1024 // 4MB
+
+    private var bitmapsBySize = HashMap<String, HashSet<Bitmap>>()
+    private var usedBitmaps = HashSet<Bitmap>()
+    private val logger = InternalLogger.UNBOUND
+
+    @VisibleForTesting
+    internal var bitmapIndex = AtomicInteger(0)
```
jonathanmos, line 33: Change to private access

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    private fun generateKey(bitmap: Bitmap) =
+        generateKey(bitmap.width, bitmap.height, bitmap.config)
+
+    private fun generateKey(width: Int, height: Int, config: Config) =
+        "$width-$height-$config"
+
+    @Suppress("SwallowedException", "TooGenericExceptionCaught")
+    private inline fun<R> safeCall(
```
jonathanmos, line 154: Perhaps worthwhile to extract the safecall methods to a global util

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            val cacheKey = "$key-${bitmapIndex.incrementAndGet()}"
+            cache.put(cacheKey, value)
+            if (bitmapsBySize[key] == null) bitmapsBySize[key] = HashSet()
+            bitmapsBySize[key]?.add(value)
+        }
+    }
+
+    @Synchronized
+    override fun size(): Int = cache.size()
```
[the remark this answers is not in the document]
jonathanmos: I've opened a task to investigate this

features/dd-sdk-android-session-replay/build.gradle.kts
```diff
@@ -103,6 +104,7 @@ dependencies {
     }
     testImplementation(libs.bundles.jUnit5)
     testImplementation(libs.bundles.testTools)
+    unmock(libs.robolectric)
```
mariusc83, line 107: why do you need robolectric libs here ?

mariusc83 (dismissed):

features/dd-sdk-android-session-replay/build.gradle.kts
jonathanmos: I thought we would need this to do the unmock. I've removed it

mariusc83 (approved):

PR #1574, opened 2023-08-09
REPLAY-1895 use internalLogger in SR modules

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
@@ -9,6 +9,7 @@ package com.datadog.android.sessionreplay.internal.async
 import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.internal.processor.RecordedDataProcessor
 import com.datadog.android.sessionreplay.internal.processor.RumContextDataHandler
 import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
@@ -16,6 +17,7 @@ import com.datadog.android.sessionreplay.internal.utils.TimeProvider
 import com.datadog.android.sessionreplay.model.MobileSegment
 import java.lang.ClassCastException
 import java.lang.NullPointerException
+import java.util.Queue
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.LinkedBlockingDeque
@@ -28,21 +30,25 @@ import java.util.concurrent.TimeUnit
  * Allows for asynchronous enrichment, which still preserving the event order.
  * The items are added to the queue from the main thread and processed on a background thread.
  */
+@Suppress("TooManyFunctions")
 internal class RecordedDataQueueHandler {
-
     private var executorService: ExecutorService
     private var processor: RecordedDataProcessor
     private var rumContextDataHandler: RumContextDataHandler
     private var timeProvider: TimeProvider
+    private val internalLogger: InternalLogger
+    internal val recordedDataQueue: Queue<RecordedDataQueueItem>
 
     internal constructor(
         processor: RecordedDataProcessor,
         rumContextDataHandler: RumContextDataHandler,
-        timeProvider: TimeProvider
+        timeProvider: TimeProvider,
+        internalLogger: InternalLogger
     ) : this(
         processor = processor,
         rumContextDataHandler = rumContextDataHandler,
         timeProvider = timeProvider,
+        internalLogger = internalLogger,
 
         /**
          * TODO: RUMM-0000 consider change to LoggingThreadPoolExecutor once V2 is merged.
@@ -56,7 +62,8 @@ internal class RecordedDataQueueHandler {
             THREAD_POOL_MAX_KEEP_ALIVE_MS,
             TimeUnit.MILLISECONDS,
             LinkedBlockingDeque()
-        )
+        ),
+        recordedQueue = ConcurrentLinkedQueue()
     )
 
     @VisibleForTesting
@@ -64,16 +71,17 @@ internal class RecordedDataQueueHandler {
         processor: RecordedDataProcessor,
         rumContextDataHandler: RumContextDataHandler,
         timeProvider: TimeProvider,
-        executorService: ExecutorService
+        executorService: ExecutorService,
+        internalLogger: InternalLogger,
+        recordedQueue: Queue<RecordedDataQueueItem> = ConcurrentLinkedQueue()
     ) {
         this.processor = processor
         this.rumContextDataHandler = rumContextDataHandler
         this.executorService = executorService
         this.timeProvider = timeProvider
+        this.internalLogger = internalLogger
+        this.recordedDataQueue = recordedQueue
     }
-
-    // region internal
-    internal val recordedDataQueue = ConcurrentLinkedQueue<RecordedDataQueueItem>()
 
     @MainThread
     internal fun addTouchEventItem(
@@ -133,10 +141,9 @@ internal class RecordedDataQueueHandler {
                 triggerProcessingLoop(currentTime)
             }
         } catch (e: RejectedExecutionException) {
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            logConsumeQueueException(e)
         } catch (e: NullPointerException) {
-            // in theory will never happen but we'll log it
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            logConsumeQueueException(e)
         }
     }
 
@@ -149,6 +156,9 @@ internal class RecordedDataQueueHandler {
     @Synchronized
     private fun triggerProcessingLoop(currentTime: Long) {
         while (recordedDataQueue.isNotEmpty()) {
+            // peeking is safe here because we are in a synchronized block
+            // and we check for isEmpty first
+            @SuppressWarnings("UnsafeThirdPartyFunctionCall")
             val nextItem = recordedDataQueue.peek()
 
             if (nextItem != null) {
@@ -168,6 +178,7 @@ internal class RecordedDataQueueHandler {
         when (nextItem) {
             is SnapshotRecordedDataQueueItem ->
                 processSnapshotEvent(nextItem)
+
             is TouchEventRecordedDataQueueItem ->
                 processTouchEvent(nextItem)
         }
@@ -194,14 +205,30 @@ internal class RecordedDataQueueHandler {
         try {
             recordedDataQueue.offer(recordedDataQueueItem)
         } catch (e: IllegalArgumentException) {
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            logAddToQueueException(e)
         } catch (e: ClassCastException) {
-            // in theory will never happen but we'll log it
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            logAddToQueueException(e)
         } catch (e: NullPointerException) {
-            // in theory will never happen but we'll log it
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            logAddToQueueException(e)
         }
+    }
+
+    private fun logAddToQueueException(e: Exception) {
+        internalLogger.log(
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.MAINTAINER,
+            { FAILED_TO_ADD_RECORDS_TO_QUEUE_ERROR_MESSAGE },
+            e
+        )
+    }
+
+    private fun logConsumeQueueException(e: Exception) {
+        internalLogger.log(
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.MAINTAINER,
+            { FAILED_TO_CONSUME_RECORDS_QUEUE_ERROR_MESSAGE },
+            e
+        )
     }
 
     // end region
@@ -212,5 +239,9 @@ internal class RecordedDataQueueHandler {
 
         private val THREAD_POOL_MAX_KEEP_ALIVE_MS = TimeUnit.SECONDS.toMillis(5)
         private const val CORE_DEFAULT_POOL_SIZE = 1 // Only one thread will be kept alive
+        internal const val FAILED_TO_CONSUME_RECORDS_QUEUE_ERROR_MESSAGE =
+            "SR RecordedDataQueueHandler: failed to consume records from queue"
+        internal const val FAILED_TO_ADD_RECORDS_TO_QUEUE_ERROR_MESSAGE =
+            "SR RecordedDataQueueHandler: failed to add records into the queue"
     }
 }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/gson/GsonExt.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/gson/GsonExt.kt
@@ -6,37 +6,76 @@
 
 package com.datadog.android.sessionreplay.internal.gson
 
+import com.datadog.android.api.InternalLogger
 import com.google.gson.JsonArray
 import com.google.gson.JsonElement
 import com.google.gson.JsonObject
 import com.google.gson.JsonPrimitive
+import java.util.Locale
+
+internal const val BROKEN_JSON_ERROR_MESSAGE_FORMAT =
+    "SR GsonExt: Unable parse the batch data into a JsonObject: expected to parse [%s] as %s"
+internal const val JSON_OBJECT_TYPE = "JsonObject"
+internal const val JSON_ARRAY_TYPE = "JsonArray"
+internal const val JSON_PRIMITIVE_TYPE = "JsonPrimitive"
 
 @Suppress("SwallowedException")
-internal fun JsonElement.safeGetAsJsonObject(): JsonObject? {
-    return try {
+internal fun JsonElement.safeGetAsJsonObject(internalLogger: InternalLogger): JsonObject? {
+    return if (isJsonObject) {
         asJsonObject
-    } catch (e: IllegalStateException) {
-        // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
+    } else {
+        internalLogger.log(
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.TELEMETRY,
+            {
+                BROKEN_JSON_ERROR_MESSAGE_FORMAT.format(
+                    Locale.ENGLISH,
+                    this.toString(),
+                    JSON_OBJECT_TYPE
+                )
+            }
+        )
         null
     }
 }
 
 @Suppress("SwallowedException")
-internal fun JsonPrimitive.safeGetAsLong(): Long? {
+internal fun JsonPrimitive.safeGetAsLong(internalLogger: InternalLogger): Long? {
     return try {
         asLong
     } catch (e: NumberFormatException) {
-        // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
+        internalLogger.log(
+            InternalLogger.Level.ERROR,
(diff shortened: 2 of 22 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -194,13 +211,26 @@ internal class RecordedDataQueueHandler {
         try {
             recordedDataQueue.offer(recordedDataQueueItem)
         } catch (e: IllegalArgumentException) {
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            internalLogger.log(
+                InternalLogger.Level.WARN,
```
mariusc83, line 215: should it be `WARN` or `ERROR` ? 🤔

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
```diff
@@ -72,7 +72,6 @@ internal class RecordedDataProcessor(
         val wireframes = snapshots.flatMap { nodeFlattener.flattenNode(it) }
 
         if (wireframes.isEmpty()) {
-            // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
```
plousada, line 75: Just making sure we don't want to log anything here if there are no wireframes.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RumContextDataHandler.kt
```diff
@@ -36,4 +42,9 @@ internal class RumContextDataHandler(
 
         return rumContextData
     }
+
+    companion object {
+        const val INVALID_RUM_CONTEXT_ERROR_MESSAGE = "SR RumContextDataHandler: Invalid RUM " +
```
plousada, line 47: Maybe we could add the context in question to better understand what is not present and make the telemetry more actionable?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -10,54 +10,56 @@ import android.annotation.SuppressLint
...
 @SuppressLint("PrivateApi")
 internal object WindowInspector {
 
-    @Suppress("SwallowedException", "TooGenericExceptionCaught")
+    internal const val FAILED_TO_RETRIEVE_DECOR_VIEWS_ERROR_MESSAGE =
+        "SR WindowInspector failed to retrieve the decor views"
+
+    @Suppress("SwallowedException", "TooGenericExceptionCaught", "UnsafeThirdPartyFunctionCall")
     private val GLOBAL_WM_CLASS by lazy {
-        try {
```
plousada, line 20: Why remove the try catches here?

plousada (dismissed):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/gson/GsonExt.kt
```diff
@@ -6,37 +6,56 @@
...
-internal fun JsonElement.safeGetAsJsonObject(): JsonObject? {
+internal fun JsonElement.safeGetAsJsonObject(internalLogger: InternalLogger): JsonObject? {
     return try {
         asJsonObject
     } catch (e: IllegalStateException) {
-        // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
+        internalLogger.log(
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.TELEMETRY,
+            { BROKEN_JSON_ERROR_MESSAGE },
```
xgouchet, line 26: Since you want to send this to telemetry, you should consider adding a more explicit message, as we're gonna parse lot of messages : 
e.g.
```
internal const val BROKEN_JSON_ERROR_MESSAGE =
    "SR GsonExt: Unable parse the batch data into a JsonObject: expected to parse %s as %s "
```

and then 

```suggestion
            { BROKEN_JSON_ERROR_MESSAGE.format(Locale.US, this.toString(), "JsonObject" },
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/gson/GsonExt.kt
```diff
@@ -6,37 +6,56 @@
...
+            e
+        )
         null
     }
 }
 
 @Suppress("SwallowedException")
-internal fun JsonPrimitive.safeGetAsLong(): Long? {
+internal fun JsonPrimitive.safeGetAsLong(internalLogger: InternalLogger): Long? {
```
xgouchet, line 34: Same comments as above

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/gson/GsonExt.kt
```diff
@@ -6,37 +6,56 @@
...
+            { BROKEN_JSON_ERROR_MESSAGE },
+            e
+        )
         null
     }
 }
 
 @Suppress("SwallowedException")
-internal fun JsonElement.safeGetAsJsonArray(): JsonArray? {
+internal fun JsonElement.safeGetAsJsonArray(internalLogger: InternalLogger): JsonArray? {
```
xgouchet, line 49: Same comments as above

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/BatchesToSegmentsMapper.kt
```diff
@@ -114,31 +131,38 @@ internal class BatchesToSegmentsMapper {
...
+        return getAsJsonPrimitive(TIMESTAMP_KEY)?.safeGetAsLong(internalLogger)
     }
 
     private fun JsonObject.rumContext(): SessionReplayRumContext? {
         val applicationId = get(EnrichedRecord.APPLICATION_ID_KEY)?.asString
         val sessionId = get(EnrichedRecord.SESSION_ID_KEY)?.asString
         val viewId = get(EnrichedRecord.VIEW_ID_KEY)?.asString
         if (applicationId == null || sessionId == null || viewId == null) {
-            // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
+            internalLogger.log(
```
xgouchet, line 159: Is this telemetry log actionable? Specifically, do we already know what we could do/check if we see those events appearing in the telemetry org? Do we have enough info to deal with this?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -10,54 +10,56 @@ import android.annotation.SuppressLint
...
-        return@lazy null
+        return@lazy GLOBAL_WM_CLASS?.getDeclaredField("mViews")
     }
 
-    fun getGlobalWindowViews(): List<View> {
-        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-            WindowInspector.getGlobalWindowViews()
-        } else {
-            getGlobalWindowViewsLegacy(GLOBAL_WM_INSTANCE, VIEWS_FIELD)
+    @SuppressWarnings("TooGenericExceptionCaught")
```
xgouchet, line 37: What are the exceptions we are expecting here? I'd rather we actually took the effort to know what the reflection calls might throw and only catch those, as catching any throwable could lead to keeping the JVM in a bad state (e.g.: catching an OOME)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -10,54 +10,56 @@ import android.annotation.SuppressLint
...
+                e,
+                true
+            )
+            emptyList()
         }
     }
 
     @Suppress("SwallowedException", "TooGenericExceptionCaught")
-    internal fun getGlobalWindowViewsLegacy(globalWmInstance: Any?, viewsField: Field?): List<View> {
+    internal fun getGlobalWindowViewsLegacy(
```
xgouchet, line 58: Same here, we should not catch a generic exception if we can. Only exceptions (no pun intended) would be if we called a customer's callback (and even then, we should only catch `Exception`  and not `Throwable`)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -10,54 +10,56 @@ import android.annotation.SuppressLint
...
-            WindowInspector.getGlobalWindowViews()
-        } else {
-            getGlobalWindowViewsLegacy(GLOBAL_WM_INSTANCE, VIEWS_FIELD)
+    @SuppressWarnings("TooGenericExceptionCaught")
+    fun getGlobalWindowViews(internalLogger: InternalLogger): List<View> {
+        return try {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                WindowInspector.getGlobalWindowViews()
+            } else {
+                getGlobalWindowViewsLegacy(internalLogger, GLOBAL_WM_INSTANCE, VIEWS_FIELD)
```
xgouchet, line 43: We already catch exceptions in that method, shouldn't we instead have two methods (e.g.: `getGlobalWindowViewsLegacy` and `getGlobalWindowViewsAPI29`) throwing exceptions, and having a global catch here? Or having each method catch their own exceptions?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
```diff
@@ -10,54 +10,56 @@ import android.annotation.SuppressLint
...
+            }
+        } catch (e: Throwable) {
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                InternalLogger.Target.TELEMETRY,
```
xgouchet, line 48: Same as above, is this actionable? Do we have enough info to diagnose what went wrong?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
```diff
@@ -54,16 +56,24 @@ internal class RecorderWindowCallback(
                 copy.recycle()
             }
         } else {
-            // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
-            // sdkLogger.errorWithTelemetry("Received MotionEvent=null")
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                InternalLogger.Target.TELEMETRY,
```
xgouchet, line 61: I feel that because we don't have any control on how we're being called by the Window, there's nothing that we could do even if we go a high volume of events telling us that MotionEvent are null. Unless we dug into the AOSP source code and know when such a situation can occur, I think this should just be a `MAINTAINER`  log

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
```diff
@@ -54,16 +56,24 @@ internal class RecorderWindowCallback(
...
 
         @Suppress("SwallowedException")
         return try {
             wrappedCallback.dispatchTouchEvent(event)
         } catch (e: Throwable) {
-            // TODO: RUMM-2397 Add the proper logs here once the sdkLogger will be added
-            // sdkLogger.errorWithTelemetry("Wrapped callback failed processing MotionEvent", e)
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                InternalLogger.Target.TELEMETRY,
```
xgouchet, line 73: Same thing the wrapped callback is customer code, so there's nothing we can do if we detect some errors. In this case, probably a `USER` message would be best.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
```diff
@@ -54,16 +56,24 @@ internal class RecorderWindowCallback(
...
+                { MOTION_EVENT_WAS_NULL_ERROR_MESSAGE },
+                null
+            )
         }
 
         @Suppress("SwallowedException")
         return try {
             wrappedCallback.dispatchTouchEvent(event)
         } catch (e: Throwable) {
```
xgouchet, line 70: Also let's only wrap `Exception`  and not `Throwable`

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
mariusc83: because they are all handled in that method there `getGlobalViews`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/gson/GsonExt.kt
mariusc83: I was thinking about that but then I thought I am going to create a very high cardinality with this. What do you think ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/BatchesToSegmentsMapper.kt
mariusc83: well, means that we have a problem in the way we asynchronously transmit the RUM context and wanted to make sure we are safe there. Without this we are never going to be sure.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
mariusc83: The idea is that is a global one because we are not very sure of the reflection exception in general but also because of the `WindowInspector.getGlobalWindowViews` which is a system call and calls `WindowManagerGlobal.getInstance().getWindowViews()` for which we do not have any information. That might throw also but we do not know what.

mariusc83: Also this method is called every time we take a snapshot so it is very important to have it not to throw anything.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
mariusc83: same as above

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
mariusc83: yes we do not as it might happen and it is normal...and also there is no actionable thing I can think of right now

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
mariusc83: yes I forgot to remove the one from `getGlobalWindowLegacy` . My idea was the same as your but actually there is no point of `getGlobalWindowViewsAPI29` as that is only calling the system method. We can catch everything here but indeed need to remove the catch from below.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
mariusc83: yeah you are right, I was not sure of this one either. I can make it a `MAINTAINER` one only for visibility in debug.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
mariusc83: I was thinking that the stacktrace should be enough. We do not send this through telemetry ? Ah damn...I guess not, in that case this is not useful as some others in my code. I thought we are logging also the stacktrace :(. In the same time I need some visibility here as this is quite important, not sure what else I would want to send as there's nothing here. Ideas ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/gson/GsonExt.kt
mariusc83: so it is my bad here as I was not aware that we are not logging the throwable stacktrace in telemetry :(. In that case I think the throwable message would make sense along with the Log message. What do you think ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/WindowInspector.kt
mariusc83: Yup, I think that stacktrace will be enough to get an idea what is going on.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -194,13 +214,26 @@ internal class RecordedDataQueueHandler {
         try {
             recordedDataQueue.offer(recordedDataQueueItem)
         } catch (e: IllegalArgumentException) {
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            internalLogger.log(
```
jonathanmos, line 217: same as previous - perhaps we could extract these

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -133,10 +140,19 @@ internal class RecordedDataQueueHandler {
                 triggerProcessingLoop(currentTime)
             }
         } catch (e: RejectedExecutionException) {
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
+            internalLogger.log(
```
jonathanmos, line 143: minor: since we're using essentially the same parameters for each of these, perhaps we could extract to a method and pass the exception

jonathanmos (approved):

plousada (approved):

xgouchet (approved):

PR #1576, opened 2023-08-16
REPLAY-1891: Optimize bitmap processing

jonathanmos:
### What does this PR do?
• Gets the bitmap from BitmapDrawable if available (use new EnrichedBitmap class to know if we should later cache it)
• Extraction of the base64 logic from the ImageButtonMapper to ImageWireframeHelper
• Gets background images if they exist (added an example to the sample app)
• Support scaletypes (atm only fitXY and center-crop)
• Move registration of bitmapPool for callbacks from DrawableUtils to Base64Serializer

### Motivation
Additional improvements necessary to rollout base64 imagebutton support

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
@@ -9,8 +9,10 @@ package com.datadog.android.sessionreplay.internal.recorder.base64
 import android.content.ComponentCallbacks2
 import android.content.Context
 import android.graphics.Bitmap
+import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import android.util.DisplayMetrics
+import android.widget.ImageView
 import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
@@ -18,6 +20,7 @@ import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.internal.AsyncImageProcessingCallback
 import com.datadog.android.sessionreplay.internal.recorder.base64.Cache.Companion.DOES_NOT_IMPLEMENT_COMPONENTCALLBACKS
 import com.datadog.android.sessionreplay.internal.utils.Base64Utils
+import com.datadog.android.sessionreplay.internal.utils.DrawableDimensions
 import com.datadog.android.sessionreplay.internal.utils.DrawableUtils
 import com.datadog.android.sessionreplay.model.MobileSegment
 import java.util.concurrent.ExecutorService
@@ -27,7 +30,7 @@ import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 
 @Suppress("UndocumentedPublicClass")
-class Base64Serializer private constructor(
+internal class Base64Serializer private constructor(
     private val threadPoolExecutor: ExecutorService,
     private val drawableUtils: DrawableUtils,
     private val base64Utils: Base64Utils,
@@ -37,6 +40,8 @@ class Base64Serializer private constructor(
     private val logger: InternalLogger
 ) {
     private var asyncImageProcessingCallback: AsyncImageProcessingCallback? = null
+    private var isCacheRegisteredForCallbacks: Boolean = false
+    private var isBitmapPoolRegisteredForCallbacks: Boolean = false
 
     // region internal
 
@@ -48,20 +53,32 @@ class Base64Serializer private constructor(
         imageWireframe: MobileSegment.Wireframe.ImageWireframe
     ) {
         registerCacheForCallbacks(applicationContext)
+        registerBitmapPoolForCallbacks(applicationContext)
 
         asyncImageProcessingCallback?.startProcessingImage()
 
+        var shouldCacheBitmap = false
         val cachedBase64 = base64LRUCache?.get(drawable)
         if (cachedBase64 != null) {
             finalizeRecordedDataItem(cachedBase64, imageWireframe, asyncImageProcessingCallback)
             return
         }
 
-        val bitmap = drawableUtils.createBitmapOfApproxSizeFromDrawable(
-            applicationContext,
-            drawable,
-            displayMetrics
-        )
+        val bitmap = if (
+            drawable is BitmapDrawable &&
+            drawable.bitmap != null &&
+            !drawable.bitmap.isRecycled
+        ) {
+            drawable.bitmap
+        } else {
+            drawableUtils.createBitmapOfApproxSizeFromDrawable(
+                drawable,
+                displayMetrics
+            )?.let {
+                shouldCacheBitmap = true
+                it
+            }
+        }
 
         if (bitmap == null) {
             asyncImageProcessingCallback?.finishProcessingImage()
@@ -70,7 +87,7 @@ class Base64Serializer private constructor(
 
         Runnable {
             @Suppress("ThreadSafety") // this runs inside an executor
-            serialiseBitmap(drawable, bitmap, imageWireframe, asyncImageProcessingCallback)
+            serialiseBitmap(drawable, bitmap, shouldCacheBitmap, imageWireframe, asyncImageProcessingCallback)
         }.let { executeRunnable(it) }
     }
 
@@ -79,6 +96,12 @@ class Base64Serializer private constructor(
     ) {
         this.asyncImageProcessingCallback = asyncImageProcessingCallback
     }
+
+    internal fun getDrawableScaledDimensions(
+        view: ImageView,
+        drawable: Drawable,
+        density: Float
+    ): DrawableDimensions = drawableUtils.getDrawableScaledDimensions(view, drawable, density)
 
     // endregion
 
@@ -95,10 +118,11 @@ class Base64Serializer private constructor(
     private fun serialiseBitmap(
         drawable: Drawable,
         bitmap: Bitmap,
+        shouldCacheBitmap: Boolean,
         imageWireframe: MobileSegment.Wireframe.ImageWireframe,
         asyncImageProcessingCallback: AsyncImageProcessingCallback?
     ) {
-        val base64String = convertBmpToBase64(drawable, bitmap)
+        val base64String = convertBmpToBase64(drawable, bitmap, shouldCacheBitmap)
         finalizeRecordedDataItem(base64String, imageWireframe, asyncImageProcessingCallback)
     }
 
@@ -120,8 +144,16 @@ class Base64Serializer private constructor(
         }
     }
 
+    @MainThread
+    private fun registerBitmapPoolForCallbacks(applicationContext: Context) {
+        if (isBitmapPoolRegisteredForCallbacks) return
+
+        applicationContext.registerComponentCallbacks(bitmapPool)
+        isBitmapPoolRegisteredForCallbacks = true
+    }
+
     @WorkerThread
-    private fun convertBmpToBase64(drawable: Drawable, bitmap: Bitmap): String {
+    private fun convertBmpToBase64(drawable: Drawable, bitmap: Bitmap, shouldCacheBitmap: Boolean): String {
         val base64Result: String
 
         val byteArray = webPImageCompression.compressBitmap(bitmap)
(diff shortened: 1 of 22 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -133,7 +160,9 @@ class Base64Serializer private constructor(
             base64LRUCache?.put(drawable, base64Result)
         }
 
-        bitmapPool?.put(bitmap)
+        if (enrichedBitmap.shouldCache) {
+            bitmapPool?.put(enrichedBitmap.bitmap)
```
0xnm, line 164: I have a feeling that we don't really need `EnrichedBitmap` maybe? Like we can have just 2 arguments of the methods - bitmap and caching flag. Anyway this class seems to be used only inside `Base64Serializer`. wdyt?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -80,6 +93,12 @@ class Base64Serializer private constructor(
         this.asyncImageProcessingCallback = asyncImageProcessingCallback
     }
 
+    internal fun getDrawableScaledDimensions(
+        view: ImageView,
+        drawable: Drawable,
+        density: Float
+    ): Pair<Long, Long> = drawableUtils.getDrawableScaledDimensions(view, drawable, density)
```
0xnm, line 100: Don't we have any class for dimensions? just to be explicit.

0xnm (commented):
nice! I left some comments, let me know wdyt.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: 👍  I'll change this to use `android.util.Size`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
```diff
@@ -6,97 +6,74 @@
...
-            imageWireframe = imageWireframe
-        )
+        imageWireframeHelper.createImageWireframe(
+            view = view,
+            index = wireframes.size,
+            x = centerX,
+            y = centerY,
+            width = scaledDrawableWidth,
+            height = scaledDrawableHeight,
+            drawable = foregroundDrawable.constantState?.newDrawable(resources),
```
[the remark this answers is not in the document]
mariusc83: I have a question here? If this is an `ImageButton` why do you resolve it as `backround` + `foreground` wireframe and not just take one `screenshot` of the whole element ?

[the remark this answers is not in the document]
jonathanmos: It's calling the drawable constructor per drawable type so I think not necessarily so heavy. Regarding taking a screenshot, we could go that route but it would mean having two different strategies on the mappers instead of having the same logic for all going forward. Also potentially it would be more resource intensive to take screenshots because you're implicitly creating a new bitmap, while you could potentially have a bitmapDrawable on the imagebutton and get the bitmap for free in the wireframe.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -56,6 +54,50 @@ internal class DrawableUtils(
         return bitmap
     }
 
+    internal fun getDrawableScaledDimensions(
+        view: ImageView,
+        drawable: Drawable,
+        density: Float
+    ): Pair<Long, Long> {
```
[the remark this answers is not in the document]
jonathanmos: Changing it to `android.util.Size`

mariusc83 (dismissed):
Apart from @0xnm comments LGTM, very nice job there.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64SerializerTest.kt
```diff
@@ -317,6 +319,131 @@ internal class Base64SerializerTest {
...
+        // When
+        testedBase64Serializer.handleBitmap(
+            applicationContext = mockApplicationContext,
+            displayMetrics = mockDisplayMetrics,
+            drawable = mockBitmapDrawable,
+            imageWireframe = fakeImageWireframe
+        )
+
+        // Then
+        verify(mockDrawableUtils, times(1)).createBitmapOfApproxSizeFromDrawable(
```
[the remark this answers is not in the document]
jonathanmos: This is to make it explicit that the call should happen only once. If I omit times(1) then the test will pass regardless of how many calls were made to the method (>0)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
```diff
@@ -31,25 +28,48 @@ abstract class BaseWireframeMapper<T : View, S : MobileSegment.Wireframe>(
...
+                0,
+                x = bounds.x,
+                y = bounds.y,
+                width = view.width.densityNormalized(density).toLong(),
+                height = view.height.densityNormalized(density).toLong(),
+                drawable = view.background.constantState?.newDrawable(resources),
+                shapeStyle = null,
+                border = null
+            )?.let {
+                wireframes.add(it as S)
```
jonathanmos, line 69: we need to discuss this unsafe cast. Because of type erasure it can't be checked, and this may rule out implementing the background drawable logic in the basewireframemapper

xgouchet: The issue is that the unsafe cast can lead to many issues: 
The `ImageWireframeHelper.createImageWireframe()` will always return `MobileSegment.Wireframe.ImageWireframe` when subclasses of `BaseWireframeMapper` might expect different Wireframe types. 
An alternative would be to remove the `S` generic and let map return a generic `MobileSegment.Wireframe` instead. I'm not sure we need to have the exact type of wireframe a in the end it's going to be serialised.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -133,7 +166,9 @@ class Base64Serializer private constructor(
             base64LRUCache?.put(drawable, base64Result)
         }
 
-        bitmapPool?.put(bitmap)
+        if (shouldCacheBitmap) {
```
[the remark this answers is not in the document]
jonathanmos: All the bitmaps will be after resizing, so they would at most be around 15kbs each. The original bitmaps that aren't resized won't be cached

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/EnrichedBitmap.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.graphics.Bitmap
+
+internal data class EnrichedBitmap(
```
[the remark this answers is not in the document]
jonathanmos: You're right - I've eliminated the class and moved it to being just an additional boolean flag

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
jonathanmos: I've removed the generic S and am now returning Wireframe

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
```diff
@@ -6,97 +6,61 @@
...
     override fun map(
         view: ImageButton,
         mappingContext: MappingContext
     ): List<MobileSegment.Wireframe> {
-        val resources = view.resources
-        val drawable = view.drawable?.constantState?.newDrawable(resources)
-        val id = resolveChildDrawableUniqueIdentifier(view)
-
-        if (drawable == null || id == null) return emptyList()
+        val wireframes = mutableListOf<MobileSegment.Wireframe>()
```
mariusc83, line 27: don't you need to call super here ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
```diff
@@ -6,97 +6,61 @@
...
-                id,
-                bounds,
-                shapeStyle,
-                border,
-                drawable
-            )
-            wireframes.add(imageWireframe)
+        // add background wireframes if any
+        if (shapeStyle == null && border == null) {
+            wireframes.addAll(super.map(view, mappingContext))
```
mariusc83, line 39: calling super here looks quite weird, it means you are using `super` just to handle your background as an `ImageWireframe`. Reason why we moved this logic in the `BaseWireframeMapper` is to be able to be shared with other mappers later. If you want to merge this as it is is fine but I would re - think this architecture...maybe just have a method in the `BaseWireframeMapper` -> `resolveBackgroundAsScreenshot` or even better maybe you could think to a way to resolve the background from a single method:
- in case in case it can be resolved as a `shapeStyle` and `border` you return an empty `ShapeWireframe` which lays at the bottom of your wireframes and contains these attributes
- in case you cannot resolve as a `shapeStyle` and `border` you return an `ImageWireframe` as you do now

mariusc83 (dismissed):
I am letting this go but I strongly recommend that you refactor the way you resolve your `View` backgrounds in the `BaseWireframeMapper`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/ImageButtonMapper.kt
jonathanmos: I've refactored this code

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
```diff
@@ -104,45 +107,90 @@ abstract class BaseWireframeMapper<T : View, S : MobileSegment.Wireframe>(
...
-        imageWireframe
-    )
-
     internal fun registerAsyncImageProcessingCallback(
         asyncImageProcessingCallback: AsyncImageProcessingCallback
     ) {
         base64Serializer?.registerAsyncLoadingCallback(asyncImageProcessingCallback)
     }
 
+    private fun resolveViewBackground(
```
mariusc83, line 116: looks way better now, tks for this change.

mariusc83 (approved):

xgouchet (approved):

0xnm (approved):

PR #1578, opened 2023-08-16
RUMM-3500 View tracking gap

xgouchet:
### What does this PR do?

Delay the view stopping event to prevent gaps between views

### Motivation

A non zero gap (~100ms) exists when navigating between fragments/activities. This is because a Fragment/Activity's `onResume` is called some time after the previous one's `onPause`. This would leave a short period of time during which errors and resources would not be tracked because no Active view was available. 

### Additional Notes

Thoughts about additional implementation of this feature : 
- mark a view as stopped as it is, and allow adding new children events (errors, resources), until a start view is detected or only after a specific period of time. 


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -59,6 +61,15 @@ internal class RumViewManagerScope(
 
         if (event is RumRawEvent.StartView && !stopped) {
             startForegroundView(event, writer)
+            lastStoppedViewTime?.let {
```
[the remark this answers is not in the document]
xgouchet: This would indeed, but the tracking strategy and view scope are completely independent from one another. If we were to move this detection in the view tracking strategy itself, we'd have to duplicate the logic three times. 
We do have the information of the view tracking strategy in the configuration telemetry event, meaning we can exclude orgs using manual tracking if need be to know where the gap is coming from.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -121,7 +132,13 @@ internal class RumViewManagerScope(
         val iterator = childrenScopes.iterator()
         @Suppress("UnsafeThirdPartyFunctionCall") // next/remove can't fail: we checked hasNext
         while (iterator.hasNext()) {
-            val result = iterator.next().handleEvent(event, writer)
+            val childScope = iterator.next()
+            if (event is RumRawEvent.StopView) {
```
mariusc83, line 136: shouldn't this be before the `while` ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/tracking/AndroidXFragmentLifecycleCallbacks.kt
```diff
@@ -77,11 +87,19 @@ internal open class AndroidXFragmentLifecycleCallbacks(
         }
     }
 
-    override fun onFragmentPaused(fm: FragmentManager, f: Fragment) {
-        super.onFragmentPaused(fm, f)
-        componentPredicate.runIfValid(f, internalLogger) {
-            val key = resolveKey(it)
-            rumMonitor.stopView(key)
+    @MainThread
+    override fun onFragmentStopped(fm: FragmentManager, f: Fragment) {
```
mariusc83, line 91: if we do this here, shouldn't we move the `startView` on the `onStart` also ? This will lower the gap.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/tracking/AndroidXFragmentLifecycleCallbacks.kt
```diff
@@ -77,11 +87,19 @@ internal open class AndroidXFragmentLifecycleCallbacks(
...
+        super.onFragmentStopped(fm, f)
+        executor.scheduleSafe(
+            "Delayed view stop",
+            STOP_VIEW_DELAY_MS,
+            TimeUnit.MILLISECONDS,
+            sdkCore.internalLogger
+        ) {
+            componentPredicate.runIfValid(f, internalLogger) {
```
mariusc83, line 99: what happens if we delay the new view resumes/starts in before the 150ms interval ? In that case this scheduled action will not just stop it ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/tracking/AndroidXFragmentLifecycleCallbacks.kt
xgouchet: No, because the Fragment can be started but not visible yet (e.g.: in a view pager, the fragment is not resumed until visible but it can be started long before that)

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/tracking/AndroidXFragmentLifecycleCallbacks.kt
xgouchet: I'm not sure I understand the question you're asking. Why would we delay the new view resume/start?

mariusc83: let me rephrase this, if I am correct here you are postponing the `stopView` method call for the current view with 150ms. In case the next view `onResume` is called during this interval, your `stopView` will be called just after. I wanted to make sure in that case you are not going to stop the `just started new view`. The way I remember we are doing a validation on the event.key but I am not sure anymore that's the reason wanted to make sure you verified that.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
mariusc83: oh yeah I see that, nevermind my question.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/tracking/AndroidXFragmentLifecycleCallbacks.kt
xgouchet: The stop view can only stop a view created with the same key, meaning we can't stop the newly created view.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/FragmentTrackingTest.kt
```diff
@@ -101,16 +101,16 @@ internal abstract class FragmentTrackingTest :
...
 
         // view stopped
-        expectedEvents.add(
-            ExpectedViewEvent(
-                fragmentAViewUrl,
-                3,
-                currentFragmentExtras(activity)
-            )
-        )
+//        expectedEvents.add(
```
xgouchet, line 104: This expected events is removed. Apparently the JUnit Runner's `callActivityOnStop` doesn't properly forward the fragment's `onStop()` to the fragment lifecycle listeners.

mariusc83 (approved):
LGTM

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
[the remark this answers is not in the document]
plousada: Makes sense. 
It would be better if we could do it before sending the telemetry in the first place but detecting if we have an automated view tracking strategy in place here would create unwanted dependencies.

plousada (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/tracking/AndroidXFragmentLifecycleCallbacks.kt
```diff
@@ -94,4 +112,8 @@ internal open class AndroidXFragmentLifecycleCallbacks(
     }
 
     // endregion
+
+    companion object {
+        private const val STOP_VIEW_DELAY_MS = 200L
```
jonathanmos, line 117: I see that this constant is defined here but also in other places. Should we unify to a SSOT incase we want to modify the delay later on?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/tracking/OreoFragmentLifecycleCallbacks.kt
```diff
@@ -104,6 +117,7 @@ internal class OreoFragmentLifecycleCallbacks(
 
     private companion object {
         private const val REPORT_FRAGMENT_NAME = "androidx.lifecycle.ReportFragment"
+        private const val STOP_VIEW_DELAY_MS = 200L
```
jonathanmos, line 120: see previous

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityViewTrackingStrategy.kt
```diff
@@ -84,4 +99,8 @@ class ActivityViewTrackingStrategy @JvmOverloads constructor(
     }
 
     // endregion
+
+    internal companion object {
+        private const val STOP_VIEW_DELAY_MS = 200L
```
jonathanmos, line 104: see previous

jonathanmos (approved):

PR #1581, opened 2023-08-22
RUMM-924 Introduce the new InternalLogger#metric API

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/api/dd-sdk-android-core.api
+++ b/dd-sdk-android-core/api/dd-sdk-android-core.api
@@ -77,6 +77,7 @@ public abstract interface class com/datadog/android/api/InternalLogger {
 	public static final field Companion Lcom/datadog/android/api/InternalLogger$Companion;
 	public abstract fun log (Lcom/datadog/android/api/InternalLogger$Level;Lcom/datadog/android/api/InternalLogger$Target;Lkotlin/jvm/functions/Function0;Ljava/lang/Throwable;ZLjava/util/Map;)V
 	public abstract fun log (Lcom/datadog/android/api/InternalLogger$Level;Ljava/util/List;Lkotlin/jvm/functions/Function0;Ljava/lang/Throwable;ZLjava/util/Map;)V
+	public abstract fun logMetric (Lkotlin/jvm/functions/Function0;Ljava/util/Map;)V
 }
 
 public final class com/datadog/android/api/InternalLogger$Companion {
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
@@ -96,6 +96,15 @@ interface InternalLogger {
         additionalProperties: Map<String, Any?>? = null
     )
 
+    /**
+     * Logs a specific metric from the internal implementation. The metric values will be sent
+     * as key-value pairs in the additionalProperties and as part of the
+     * [com.datadog.android.telemetry.model.TelemetryDebugEvent.Telemetry] event.
+     * @param messageBuilder the lambda building the metric message
+     * @param additionalProperties additional properties to add to the metric
+     */
+    fun logMetric(messageBuilder: () -> String, additionalProperties: Map<String, Any?>)
+
     companion object {
 
         /**
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/SdkInternalLogger.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/SdkInternalLogger.kt
@@ -88,6 +88,18 @@ internal class SdkInternalLogger(
         }
     }
 
+    override fun logMetric(messageBuilder: () -> String, additionalProperties: Map<String, Any?>) {
+        val rumFeature = sdkCore?.getFeature(Feature.RUM_FEATURE_NAME) ?: return
+        val message = messageBuilder()
+        val telemetryEvent =
+            mapOf(
+                TYPE_KEY to "mobile_metric",
+                MESSAGE_KEY to message,
+                ADDITIONAL_PROPERTIES_KEY to additionalProperties
+            )
+        rumFeature.sendEvent(telemetryEvent)
+    }
+
     // endregion
 
     // region Internal
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/SdkInternalLoggerTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/SdkInternalLoggerTest.kt
@@ -14,6 +14,7 @@ import com.datadog.android.api.feature.FeatureScope
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.utils.forge.Configurator
 import com.datadog.tools.unit.forge.aThrowable
+import com.datadog.tools.unit.forge.exhaustiveAttributes
 import fr.xgouchet.elmyr.Forge
 import fr.xgouchet.elmyr.annotation.IntForgery
 import fr.xgouchet.elmyr.annotation.StringForgery
@@ -23,6 +24,7 @@ import fr.xgouchet.elmyr.junit5.ForgeExtension
 import org.assertj.core.api.Assertions
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
 import org.mockito.Mock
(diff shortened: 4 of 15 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
```diff
@@ -96,6 +96,13 @@ interface InternalLogger {
         additionalProperties: Map<String, Any?>? = null
     )
 
+    /**
+     * Logs a metric from the internal implementation.
```
0xnm, line 100: what is the meaning of `internal implementation` here? Also maybe we need to add the statement that metric will be sent to the telemetry (and say how), because from the method arguments it is not evident where is the metric here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/SdkInternalLogger.kt
```diff
@@ -188,6 +192,21 @@ internal class SdkInternalLogger(
         }
         rumFeature.sendEvent(telemetryEvent)
     }
+    private fun logMetric(
```
0xnm, line 195:
```suggestion

    private fun logMetric(
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/SdkInternalLoggerTest.kt
```diff
@@ -444,6 +445,51 @@ internal class SdkInternalLoggerTest {
...
+        // Given
+        whenever(mockSdkCore.getFeature(Feature.RUM_FEATURE_NAME)) doReturn null
+        val fakeAdditionalProperties = forge.exhaustiveAttributes()
+        val mockLambda: () -> String = mock()
+        whenever(mockLambda.invoke()) doReturn fakeMessage
+        // When
+        testedInternalLogger.metric(
+            mockLambda,
+            fakeAdditionalProperties
+        )
```
0xnm, line 490: What is the point of the test without verification part?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/Telemetry.kt
```diff
@@ -39,4 +39,8 @@ internal class Telemetry(
         (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor)
             ?.sendDebugTelemetryEvent(message, additionalProperties)
     }
+    fun metric(message: String, additionalProperties: Map<String, Any?>? = null) {
```
0xnm, line 42:
```suggestion

    fun metric(message: String, additionalProperties: Map<String, Any?>? = null) {
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -113,7 +113,7 @@ internal class TelemetryEventHandler(
 
         val eventIdentity = event.identity
 
-        if (seenInCurrentSession.contains(eventIdentity)) {
+        if (!event.isMetric && seenInCurrentSession.contains(eventIdentity)) {
```
0xnm, line 116: do we have a risk of sending too much metrics by disabling this check and thus adding a lot of pressure on the storage and also on the Datadog servers?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/RumFeatureTest.kt
```diff
@@ -1124,6 +1124,49 @@ internal class RumFeatureTest {
...
+        testedFeature.onReceive(event)
+
+        // Then
+        verify(mockRumMonitor).sendMetricEvent(fakeMessage, fakeAdditionalProperties)
+        verifyNoMoreInteractions(mockRumMonitor)
+        verifyNoInteractions(mockInternalLogger)
+    }
+
+    @Test
+    fun `𝕄 handle metric event 𝕎 onReceive(){no additionalProperties}`(
```
0xnm, line 1151: probably test with missing message is missing

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/SdkInternalLoggerTest.kt
mariusc83: I cannot verify anything there, I wanted to make sure is not throwing anything...maybe there is a method to assert that `isNotThrowing`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
mariusc83: I was following what was done in iOS here. I guess the risk was assessed already there, in the end we are going to have an extra sampler in the `MetricsDispatcher` which will come in the next PR. We have a `15%` extra sampler there.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/RumFeatureTest.kt
```diff
@@ -1124,6 +1124,49 @@ internal class RumFeatureTest {
...
+    @Test
+    fun `𝕄 handle metric event 𝕎 onReceive()`(
+        @StringForgery fakeMessage: String,
+        forge: Forge
+    ) {
+        // Given
+        val fakeAdditionalProperties = forge.exhaustiveAttributes()
+        testedFeature.onInitialize(appContext.mockInstance)
+        val event = mapOf(
+            "type" to "mobile_metric",
```
jonathanmos, line 1136: minor: "mobile_metric" is hardcoded in the tests in several places - perhaps it could be taken from a constant so as to only change in one place in future if the value changed?

jonathanmos (dismissed):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/SdkInternalLoggerTest.kt
mariusc83: yup, that's what I also have in mind ;)

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/SdkInternalLoggerTest.kt
```diff
@@ -444,6 +446,53 @@ internal class SdkInternalLoggerTest {
             )
     }
 
+    @Test
+    fun `𝕄 send metric 𝕎 metric`(
```
xgouchet, line 450:
```suggestion
    fun `𝕄 send metric 𝕎 metric()`(
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
```diff
@@ -96,6 +96,16 @@ interface InternalLogger {
...
 
+    /**
+     * Logs a specific metric from the internal implementation. The metric will leverage the new
+     * added additionalProperties in the
+     * [com.datadog.android.telemetry.model.TelemetryDebugEvent.Telemetry] and will be sent as
+     * [com.datadog.android.telemetry.model.TelemetryDebugEvent].
+     * @param messageBuilder the lambda building the metric message
+     * @param additionalProperties additional properties to add to the metric
+     */
+    fun metric(messageBuilder: () -> String, additionalProperties: Map<String, Any?>)
```
xgouchet, line 107: Function name should be action verbs. We should consider `logMetric`. Alternatively, we could instead use an `InternalLogger.Level.METRIC` instead, and reuse the existing top level methods. WDYT?

mariusc83: i'd rather keep a new method here to match what `iOS` added in the `InternalLogger` but will change it to `logMetric`

dd-sdk-android-core/api/dd-sdk-android-core.api
```diff
@@ -77,6 +77,7 @@ public abstract interface class com/datadog/android/api/InternalLogger {
 	public static final field Companion Lcom/datadog/android/api/InternalLogger$Companion;
 	public abstract fun log (Lcom/datadog/android/api/InternalLogger$Level;Lcom/datadog/android/api/InternalLogger$Target;Lkotlin/jvm/functions/Function0;Ljava/lang/Throwable;ZLjava/util/Map;)V
 	public abstract fun log (Lcom/datadog/android/api/InternalLogger$Level;Ljava/util/List;Lkotlin/jvm/functions/Function0;Ljava/lang/Throwable;ZLjava/util/Map;)V
+	public abstract fun metric (Lkotlin/jvm/functions/Function0;Ljava/util/Map;)V
```
0xnm, line 80: this has to be updated

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
```diff
@@ -96,6 +96,16 @@ interface InternalLogger {
         additionalProperties: Map<String, Any?>? = null
     )
 
+    /**
+     * Logs a specific metric from the internal implementation. The metric will leverage the new
+     * added additionalProperties in the
```
0xnm, line 101: I'm still not quite clear with the description. Does `internal implementation` mean from SDK-related code only? Also I guess there is a typo in `new added`, and not quite clear how `additionalProperties` will be leveraged. Maybe we can say that metric data (type, values, etc.) will be taken from `additionalProperties`? This way it is more clear imo.

0xnm (approved):
lgtm. I left some comments, but they are not blocking.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
mariusc83: the `internal implementation` is present in all the other methods description so I just followed the pattern. And yes for me it means that is meant for internal usage inside the SDK code. I will try to update the last part there ;)

dd-sdk-android-core/api/dd-sdk-android-core.api
mariusc83: yeah it is something wrong with the api generation as I already ran that method 🤔

xgouchet (approved):

jonathanmos (approved):

PR #1591, opened 2023-08-29
RUM-971: Report ApplicationLaunch view even if first RUM event is not interaction

### LINKED ISSUE #1523: App startup time missing when using version 1.19.2
**Describe what happened**
We upgraded to SDK version 1.19.2 and I noticed that the startup time metric is now gone for any app versions that use this SDK version. When I downgrade the SDK locally to 1.19.1, I can get it back.

Use SDK version 1.19.2 and notice that the app startup time is empty on the DD dashboard.

Screenshot 2023-07-05 at 13 04 13

I expect the startup time to be tracked.

**Additional context**
 - Android OS version:
 - Device Model:
 - Datadog SDK version: 1.19.2
 - Versions of any other relevant dependencies (OkHttp, …): 
 - Proguard configuration: 
 - Gradle Plugins:

### LINKED ISSUE #1582: application_start action event won't be called when AddLongTask event triggered as a first RUM event
**Describe what happened**

Usually Datadog will trigger `application_start` action event at first.
But when `AddLongTask` event was triggered as a first RUM event, `application_start` action event won't be triggered.

Trigger any `AddLongTask` quickly when just Datadog initialized.

Sample: https://github.com/ken-kentan/dd-sdk-android/pull/2

`application_start` action event should be triggered even if the first RUM event was `AddLongTask`.

I investigated around `RumSessionScope` and found than when `AddLongTask` event was triggered as a first event, `sessionState` was changed to `State.EXPIRED`, then `RumSessionScope::handleEvent` passed `noOpWriter` as an `actualWriter` for `ApplicationStarted` event.
I guess `sessionState` shouldn't be `State.EXPIRED` in this case.

**Additional context**
 - Android OS version: N/A
 - Device Model: N/A
 - Datadog SDK version: 1.19.0-2.0.0
 - Versions of any other relevant dependencies (OkHttp, …):  N/A
 - Proguard configuration: N/A
 - Gradle Plugins: N/A

Note: This issue might be same with https://github.com/DataDog/dd-sdk-android/issues/1523

0xnm:
### What does this PR do?

Fixes #1582 and maybe fixes #1523.

This PR fixes the issue when `ApplicationLaunch` is not reported in case if first RUM event coming is not interaction event. In this case session is not yet active, so we create `ApplicationLaunch` and `ApplicationStarted` action for nothing - they won't be written, because writer is no-op.

This PR moves the logic of creating `ApplicationStarted` event up to the application scope, which makes sense - we have to report it only once in the application lifecycle. Also session we be activated once `ApplicationStarted` event is received by the session scope, meaning that session is created not only by start view or start action events (which makes sense, because we introduced `ApplicationLaunch` view to be independent from background tracking feature).

There is one leftover which I don't personally enjoy - having `applicationDisplayed` property in the scopes (I think this property is vague), but getting rid of it is tricky.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
@@ -6,16 +6,22 @@
 
 package com.datadog.android.rum.internal.domain.scope
 
+import android.app.ActivityManager
 import androidx.annotation.WorkerThread
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.storage.DataWriter
 import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
+import com.datadog.android.rum.DdRumContentProvider
 import com.datadog.android.rum.RumSessionListener
+import com.datadog.android.rum.internal.AppStartTimeProvider
+import com.datadog.android.rum.internal.DefaultAppStartTimeProvider
 import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.domain.Time
 import com.datadog.android.rum.internal.vitals.VitalMonitor
 import java.util.Locale
+import java.util.concurrent.TimeUnit
 
 @Suppress("LongParameterList")
 internal class RumApplicationScope(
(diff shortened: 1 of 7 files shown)


xgouchet (dismissed):
Nice work

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -155,6 +169,34 @@ internal class RumApplicationScope(
         }
     }
 
+    @WorkerThread
+    private fun sendApplicationStartEvent(eventTime: Time, writer: DataWriter<Any>) {
+        val processImportance = DdRumContentProvider.processImportance
+        val isForegroundProcess = processImportance ==
+            ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
+        if (isForegroundProcess) {
+            val processStartTime = appStartTimeProvider.appStartTimeNs
```
jonathanmos, line 178: minor: since the processStartTime is in ns, should we not rename to processStartTimeNs?

jonathanmos (dismissed):

mariusc83 (dismissed):

PR #1592, opened 2023-08-30
RUM-936: Serialize TextViews/Buttons to base64

jonathanmos:
### What does this PR do?
Implement serialization of TextViews and Buttons to base64.

### Motivation
Continuation of base64 support for different views.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
@@ -8,19 +8,25 @@ package com.datadog.android.sessionreplay.internal.recorder.base64
 
 import android.graphics.drawable.Drawable
 import android.view.View
+import android.widget.TextView
 import androidx.annotation.MainThread
+import androidx.annotation.VisibleForTesting
+import com.datadog.android.sessionreplay.internal.recorder.MappingContext
+import com.datadog.android.sessionreplay.internal.recorder.ViewUtilsInternal
+import com.datadog.android.sessionreplay.internal.recorder.densityNormalized
 import com.datadog.android.sessionreplay.model.MobileSegment
 import com.datadog.android.sessionreplay.utils.UniqueIdentifierGenerator
 
 internal class ImageWireframeHelper(
     private val imageCompression: ImageCompression = WebPImageCompression(),
     private val uniqueIdentifierGenerator: UniqueIdentifierGenerator = UniqueIdentifierGenerator,
-    private val base64Serializer: Base64Serializer
+    private val base64Serializer: Base64Serializer,
+    private val viewUtilsInternal: ViewUtilsInternal = ViewUtilsInternal()
 ) {
     @MainThread
     internal fun createImageWireframe(
         view: View,
-        index: Int,
+        currentWireframeIndex: Int,
         x: Long,
         y: Long,
         width: Long,
@@ -30,10 +36,15 @@ internal class ImageWireframeHelper(
         border: MobileSegment.ShapeBorder? = null,
         prefix: String = DRAWABLE_CHILD_NAME
     ): MobileSegment.Wireframe.ImageWireframe? {
-        val id = uniqueIdentifierGenerator.resolveChildUniqueIdentifier(view, prefix + index)
+        val id = uniqueIdentifierGenerator.resolveChildUniqueIdentifier(view, prefix + currentWireframeIndex)
 
         @Suppress("ComplexCondition")
-        if (drawable == null || id == null || drawable.intrinsicWidth < 0 || drawable.intrinsicHeight < 0) {
+        if (
+            drawable == null ||
+            id == null ||
+            drawable.intrinsicWidth <= 0 ||
+            drawable.intrinsicHeight <= 0
+        ) {
             return null
         }
 
@@ -65,7 +76,78 @@ internal class ImageWireframeHelper(
         return imageWireframe
     }
 
-    private companion object {
-        private const val DRAWABLE_CHILD_NAME = "drawable"
+    @Suppress("NestedBlockDepth")
+    internal fun createCompoundDrawableWireframes(
+        view: TextView,
+        mappingContext: MappingContext,
+        prevWireframeIndex: Int
+    ): MutableList<MobileSegment.Wireframe> {
+        val result = mutableListOf<MobileSegment.Wireframe>()
+        var wireframeIndex = prevWireframeIndex
+        val density = mappingContext.systemInformation.screenDensity
+
+        // CompoundDrawables returns an array of indexes in the following order:
+        // left, top, right, bottom
+        view.compoundDrawables.forEachIndexed { compoundDrawableIndex, _ ->
+            if (compoundDrawableIndex > CompoundDrawablePositions.values().size) {
+                return@forEachIndexed
+            }
+
+            val compoundDrawablePosition = convertIndexToCompoundDrawablePosition(
+                compoundDrawableIndex
+            ) ?: return@forEachIndexed
+
+            val drawable = view.compoundDrawables[compoundDrawableIndex]
+
+            if (drawable != null) {
+                val drawableCoordinates = viewUtilsInternal.resolveCompoundDrawableBounds(
+                    view = view,
+                    drawable = drawable,
+                    pixelsDensity = density,
+                    position = compoundDrawablePosition
+                )
+
+                @Suppress("ThreadSafety") // TODO REPLAY-1861 caller thread of .map is unknown?
+                createImageWireframe(
+                    view = view,
+                    currentWireframeIndex = ++wireframeIndex,
+                    x = drawableCoordinates.x,
+                    y = drawableCoordinates.y,
+                    width = drawable.intrinsicWidth
+                        .densityNormalized(density).toLong(),
+                    height = drawable.intrinsicHeight
+                        .densityNormalized(density).toLong(),
+                    drawable = drawable,
+                    shapeStyle = null,
+                    border = null
+                )?.let { resultWireframe ->
+                    result.add(resultWireframe)
+                }
+            }
+        }
+
+        return result
+    }
+
+    @Suppress("MagicNumber")
+    private fun convertIndexToCompoundDrawablePosition(compoundDrawableIndex: Int): CompoundDrawablePositions? {
+        return when (compoundDrawableIndex) {
+            0 -> CompoundDrawablePositions.LEFT
+            1 -> CompoundDrawablePositions.TOP
+            2 -> CompoundDrawablePositions.RIGHT
+            3 -> CompoundDrawablePositions.BOTTOM
+            else -> null
+        }
+    }
+
+    internal enum class CompoundDrawablePositions {
+        LEFT,
+        TOP,
+        RIGHT,
+        BOTTOM
+    }
+
+    internal companion object {
+        @VisibleForTesting internal const val DRAWABLE_CHILD_NAME = "drawable"
     }
 }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/TextViewMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/TextViewMapper.kt
@@ -9,11 +9,15 @@ package com.datadog.android.sessionreplay.internal.recorder.mapper
 import android.graphics.Typeface
 import android.view.Gravity
 import android.widget.TextView
+import com.datadog.android.sessionreplay.internal.recorder.GlobalBounds
 import com.datadog.android.sessionreplay.internal.recorder.MappingContext
+import com.datadog.android.sessionreplay.internal.recorder.base64.Base64Serializer
+import com.datadog.android.sessionreplay.internal.recorder.base64.ImageWireframeHelper
 import com.datadog.android.sessionreplay.internal.recorder.densityNormalized
 import com.datadog.android.sessionreplay.internal.recorder.obfuscator.rules.AllowObfuscationRule
 import com.datadog.android.sessionreplay.internal.recorder.obfuscator.rules.TextValueObfuscationRule
 import com.datadog.android.sessionreplay.model.MobileSegment
+import com.datadog.android.sessionreplay.utils.UniqueIdentifierGenerator
 
 /**
  * A [WireframeMapper] implementation to map a [TextView] component.
@@ -23,26 +27,86 @@ import com.datadog.android.sessionreplay.model.MobileSegment
  */
 @Suppress("TooManyFunctions")
 open class TextViewMapper :
-    BaseWireframeMapper<TextView, MobileSegment.Wireframe.TextWireframe> {
+    BaseWireframeMapper<TextView, MobileSegment.Wireframe> {
 
-    internal val textValueObfuscationRule: TextValueObfuscationRule
+    internal var textValueObfuscationRule: TextValueObfuscationRule = AllowObfuscationRule()
+    private var base64Serializer: Base64Serializer? = null
+    private var imageWireframeHelper: ImageWireframeHelper? = null
+    private var uniqueIdentifierGenerator: UniqueIdentifierGenerator? = null
 
-    constructor() {
-        textValueObfuscationRule = AllowObfuscationRule()
+    constructor()
+
+    internal constructor(
+        base64Serializer: Base64Serializer,
+        imageWireframeHelper: ImageWireframeHelper,
+        uniqueIdentifierGenerator: UniqueIdentifierGenerator,
+        textValueObfuscationRule: TextValueObfuscationRule? = null
+    ) : super(base64Serializer, imageWireframeHelper, uniqueIdentifierGenerator) {
+        this.base64Serializer = base64Serializer
+        this.imageWireframeHelper = imageWireframeHelper
+        this.uniqueIdentifierGenerator = uniqueIdentifierGenerator
+        textValueObfuscationRule?.let {
+            this.textValueObfuscationRule = it
+        }
     }
 
-    internal constructor(textValueObfuscationRule: TextValueObfuscationRule) {
+    internal constructor(
+        textValueObfuscationRule: TextValueObfuscationRule
+    ) {
         this.textValueObfuscationRule = textValueObfuscationRule
     }
 
     override fun map(view: TextView, mappingContext: MappingContext):
-        List<MobileSegment.Wireframe.TextWireframe> {
+        List<MobileSegment.Wireframe> {
+        val wireframes = mutableListOf<MobileSegment.Wireframe>()
+
+        wireframes.addAll(super.map(view, mappingContext))
+
+        val density = mappingContext.systemInformation.screenDensity
         val viewGlobalBounds = resolveViewGlobalBounds(
             view,
-            mappingContext.systemInformation.screenDensity
+            density
         )
-        val (shapeStyle, border) = view.background?.resolveShapeStyleAndBorder(view.alpha)
-            ?: (null to null)
+
+        resolveTextElements(
+            view,
+            mappingContext,
+            viewGlobalBounds
+        ).let(wireframes::addAll)
+
+        resolveImages(
+            view,
+            mappingContext,
+            wireframes.size
+        ).let(wireframes::addAll)
+
+        return wireframes
+    }
+
+    // region Internal
+
+    private fun resolveImages(
+        view: TextView,
+        mappingContext: MappingContext,
+        currentIndex: Int
+    ): List<MobileSegment.Wireframe> {
+        val wireframes = mutableListOf<MobileSegment.Wireframe>()
+
+        imageWireframeHelper?.createCompoundDrawableWireframes(
+            view,
+            mappingContext,
+            currentIndex
+        )?.let { result ->
+            wireframes.addAll(result)
+        }
+
+        return wireframes
+    }
+    private fun resolveTextElements(
+        view: TextView,
+        mappingContext: MappingContext,
+        viewGlobalBounds: GlobalBounds
+    ): List<MobileSegment.Wireframe> {
         return listOf(
             MobileSegment.Wireframe.TextWireframe(
                 id = resolveViewId(view),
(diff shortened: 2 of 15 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -33,7 +39,12 @@ internal class ImageWireframeHelper(
         val id = uniqueIdentifierGenerator.resolveChildUniqueIdentifier(view, prefix + index)
 
         @Suppress("ComplexCondition")
-        if (drawable == null || id == null || drawable.intrinsicWidth < 0 || drawable.intrinsicHeight < 0) {
+        if (
+            drawable == null ||
+            id == null ||
+            drawable.intrinsicWidth < 0 ||
```
0xnm, line 45: should it be `<=` here?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+    @Suppress("NestedBlockDepth")
+    internal fun createCompoundDrawableWireframes(
+        view: TextView,
+        mappingContext: MappingContext,
+        startIndex: Int
+    ): MutableList<MobileSegment.Wireframe> {
+        val result = mutableListOf<MobileSegment.Wireframe>()
+        var currentIndex = startIndex
+
+        view.compoundDrawables.indices.forEach {
```
0xnm, line 88: minor: I think we can use `forEachIndexed` here

```suggestion
        view.compoundDrawables.forEachIndexed {
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+            if (drawable != null) {
+                val compoundDrawableLocationMap = createCompoundDrawableLocationMap(
+                    view = view,
+                    drawable = drawable,
+                    density = density,
+                    index = it
+                )
+
+                val drawableX = compoundDrawableLocationMap[it]?.first
+                val drawableY = compoundDrawableLocationMap[it]?.second
```
0xnm, line 101: I think you can use `android.graphics.Point` class in order to avoid using `Pair`.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+
+        return result
+    }
+
+    private fun createCompoundDrawableLocationMap(
+        view: TextView,
+        drawable: Drawable,
+        density: Float,
+        index: Int
+    ): HashMap<Int, Pair<Long, Long>> {
```
0xnm, line 132: return type can be just `Map`

```suggestion
    ): Map<Int, Pair<Long, Long>> {
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+        val compoundDrawableLocationMap = HashMap<Int, Pair<Long, Long>>()
```
0xnm, line 133:
```suggestion
        val compoundDrawableLocationMap = mutableMapOf<Int, Pair<Long, Long>>()
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+            .densityNormalized(density)
+        val viewPaddingTop = view.paddingTop
+            .densityNormalized(density)
+        val viewPaddingBottom = view.paddingBottom
+            .densityNormalized(density)
+        val viewPaddingEnd = view.paddingEnd
+            .densityNormalized(density)
+
+        @Suppress("MagicNumber")
+        compoundDrawableLocationMap[index] = when (index) {
```
0xnm, line 159: maybe enum would be here more descriptive than just integers? Or we can use `@IntDef` as well.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+            )
+            3 -> getCompoundDrawableBottomCoords(
+                viewPaddingBottom,
+                viewWidth,
+                viewHeight,
+                drawableWidth,
+                drawableHeight,
+                globalBounds
+            )
+            else -> error("Unexpected drawable index: $index")
```
0xnm, line 188: who handles this exception? SDK shouldn't crash host app.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+        }
+
+        return compoundDrawableLocationMap
+    }
+    private fun getCompoundDrawableStartCoords(
```
0xnm, line 193: what are the start/top/end/bottom coordinates of the compound drawable? what do they describe?

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelperTest.kt
```diff
@@ -243,4 +276,156 @@ internal class ImageWireframeHelperTest {
...
+        whenever(mockTextView.compoundDrawables)
+            .thenReturn(arrayOf(mockDrawable, null, null, null))
+
+        // When
+        val wireframes = testedHelper.createCompoundDrawableWireframes(
+            mockTextView,
+            mockMappingContext,
+            0
+        )
+        val actualWireframe = wireframes[0] as MobileSegment.Wireframe.ImageWireframe
```
0xnm, line 309: minor: I would move `actualWireframe` into the `// Then` section (everywhere in this class)

0xnm (commented):
nice feature! I added few questions/suggestions

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+        view: TextView,
+        mappingContext: MappingContext,
+        startIndex: Int
+    ): MutableList<MobileSegment.Wireframe> {
+        val result = mutableListOf<MobileSegment.Wireframe>()
+        var currentIndex = startIndex
+
+        view.compoundDrawables.indices.forEach {
+            val drawable = view.compoundDrawables[it]
+            val density = mappingContext.systemInformation.screenDensity
```
mariusc83, line 90: why do you resolve this in the for loop ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+
+            if (drawable != null) {
+                val compoundDrawableLocationMap = createCompoundDrawableLocationMap(
```
mariusc83, line 93: I wonder why do you need a `Map` here if later you just get the entry for each drawable in the loop ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+    ): HashMap<Int, Pair<Long, Long>> {
+        val compoundDrawableLocationMap = HashMap<Int, Pair<Long, Long>>()
+
+        val globalBounds = viewUtilsInternal.resolveDrawableBounds(
+            view,
+            drawable,
+            density
+        )
+
+        val drawableWidth = drawable.intrinsicWidth
```
mariusc83, line 141: why do you still need these values here as you already have the `drawableWidth` and `drawableHeight` in the `globalBounds` above

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
mariusc83: definitely an Enum or @IntDef here, that index is very very hard to read

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,179 @@ internal class ImageWireframeHelper(
...
+                        result.add(resultWireframe)
+                    }
+                }
+            }
+        }
+
+        return result
+    }
+
+    private fun createCompoundDrawableLocationMap(
```
mariusc83, line 127: overall I think this logic needs to be simplified here, right now is quite hard to read and as @0xnm mentioned I am not sure you need all this logic

mariusc83: In the end the problem that you are trying to solve is to get the bounds of a drawable depending on its parent (view) global bounds + padding + position (left, right, top, bottom) . I think you can find a cleaner logic to achieve this.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: I'm changing it to an enum

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: changed to logging/returning empty coords

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: changed to compare to a minimum dimensions constant

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: I've removed the map and changed to return the points directly

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: I've simplified and moved the bounds logic into `InternalViewUtils`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: It describes the absolute positions of the drawables surrounding the `TextView`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,57 @@ internal class ImageWireframeHelper(
...
+            }
+        }
+
+        return result
+    }
+
+    internal companion object {
+        @VisibleForTesting internal const val DRAWABLE_CHILD_NAME = "drawable"
+
+        @VisibleForTesting internal const val DRAWABLE_MIN_DP = 10
```
jonathanmos, line 130: I've set this as 10 - does this seem reasonable?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,57 @@ internal class ImageWireframeHelper(
         return imageWireframe
     }
 
-    private companion object {
-        private const val DRAWABLE_CHILD_NAME = "drawable"
+    @Suppress("NestedBlockDepth")
```
mariusc83, line 79: much much better this way, tks !!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/TextViewMapper.kt
```diff
@@ -23,35 +27,95 @@ import com.datadog.android.sessionreplay.model.MobileSegment
...
+        val density = mappingContext.systemInformation.screenDensity
         val viewGlobalBounds = resolveViewGlobalBounds(
             view,
-            mappingContext.systemInformation.screenDensity
+            density
         )
-        val (shapeStyle, border) = view.background?.resolveShapeStyleAndBorder(view.alpha)
-            ?: (null to null)
+
+        resolveTextElements(
```
mariusc83, line 71: I like this and is quite scalable

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
mariusc83: hmm should we add a limit at all ?
I would start without actually as there is no content image in a TextView drawable usually. What do you think ?

mariusc83 (dismissed):
Nice work, LGTM !!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: alright, I'll remove this for now and set to <= 0 and later on we can set a limit if needed

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,55 @@ internal class ImageWireframeHelper(
...
+    internal fun createCompoundDrawableWireframes(
+        view: TextView,
+        mappingContext: MappingContext,
+        startIndex: Int
+    ): MutableList<MobileSegment.Wireframe> {
+        val result = mutableListOf<MobileSegment.Wireframe>()
+        var currentIndex = startIndex
+        val density = mappingContext.systemInformation.screenDensity
+
+        view.compoundDrawables.forEachIndexed { index, _ ->
```
0xnm, line 89: probably we need to add a comment saying that `compoundDrawables` returns an array of drawables in the following order: left, top, right, bottom. It will be easier to understand what is the index.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -65,7 +76,55 @@ internal class ImageWireframeHelper(
...
+                )
+
+                val drawableX: Long? = drawableCoordinates?.x
+                val drawableY: Long? = drawableCoordinates?.y
+
+                if (drawableX != null && drawableY != null) {
+                    @Suppress("ThreadSafety") // TODO REPLAY-1861 caller thread of .map is unknown?
+                    createImageWireframe(
+                        view = view,
+                        index = ++currentIndex,
```
0xnm, line 107: I would also suggest to convert index to enum in this method instead of doing it inside `createImageWireframe`, this has several advantages:

* we don't use `index` in the `createImageWireframe`, it removes confusion because when inside `createImageWireframe` it is not clear what is the index.
* `createImageWireframe` signature is cleaner, it will get an enum.
* it will remove the need for `CompoundDrawablePositions` to have `position` property, because you can create it right in the beginning of the `forEachIndexed` lambda in the following way:

```
val drawablePosition = when(index) {
   0 -> ...
   1 -> ...
   ...
}
```
since it will be close to the comment describing `compoundDrawables` property, it will be clear what is the index.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewUtilsInternalTest.kt
```diff
@@ -6,25 +6,35 @@
...
 
+import android.graphics.drawable.Drawable
 import android.view.View
 import android.view.ViewStub
+import android.widget.TextView
 import androidx.appcompat.widget.ActionBarContextView
 import androidx.appcompat.widget.Toolbar
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.forge.ForgeConfigurator
+import com.datadog.android.sessionreplay.internal.recorder.ViewUtilsInternal.Companion.COMPOUND_DRAWABLE_INDEX_OUT_OF_BOUNDS
```
0xnm, line 17: it should be possible to reference it directly, without `Companion`

```suggestion
import com.datadog.android.sessionreplay.internal.recorder.ViewUtilsInternal.COMPOUND_DRAWABLE_INDEX_OUT_OF_BOUNDS
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewUtilsInternalTest.kt
```diff
@@ -182,5 +225,149 @@ internal class ViewUtilsInternalTest {
...
+            (mockTextView.width / 2 - mockDrawable.intrinsicWidth / 2).toLong()
+        )
+        assertThat(actualBounds?.y).isEqualTo(
+            (
+                mockTextView.height -
+                    (
+                        mockDrawable.intrinsicHeight +
+                            mockTextView.paddingBottom
+                        )
+                ).toLong()
```
0xnm, line 370: seems like bad formatting

0xnm (commented):
looks good! I left some final suggestions.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: I understand the confusion here - this is a different index. The index for compoundDrawable represents the left, top, right, bottom drawable, however the index for createImageWireframe represents the index of the wireframe in the stack. Since we are creating potentially multiple imagewireframes off the same view we need a way to generate unique ids for each wireframe to avoid collisions - this can be seen in line39 of imageWireframeHelper where the index is appended to the resolveChildUniqueIdentifier call.

0xnm: Then still maybe it is better to remove this confusion, because here https://github.com/DataDog/dd-sdk-android/pull/1592/files#diff-7870c4af39004135266ff0b27eb0e6ee96774f0443572f43f1225b716dc9a0d3R80 it is still 0 to left, etc. So we can create enum in the current method, and let then index to be used in `createImageWireframe` only for ID generation.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewUtilsInternalTest.kt
jonathanmos: yes, ktlint keeps "fixing" this line to look like this. I'll find something that it won't modify

0xnm (approved):
lgtm! I added few suggestions, but they are not blocking.

PR #1596, opened 2023-09-01
RUM-986: Fix crash when disabling JankStats tracking

0xnm:
### What does this PR do?

The following crash was reported by the customer:

```
java.lang.IllegalArgumentException: attempt to remove OnFrameMetricsAvailableListener that was never added
        at android.view.View.removeFrameMetricsListener(View.java:7797)
        at android.view.Window.removeOnFrameMetricsAvailableListener(Window.java:995)
        at androidx.metrics.performance.DelegatingFrameMetricsListener.remove(JankStatsApi24Impl.kt:253)
        at androidx.metrics.performance.JankStatsApi24Impl.removeFrameMetricsListenerDelegate(JankStatsApi24Impl.kt:126)
        at androidx.metrics.performance.JankStatsApi24Impl.setupFrameTimer(JankStatsApi24Impl.kt:107)
        at androidx.metrics.performance.JankStats.setTrackingEnabled(JankStats.kt:129)
        at com.datadog.android.rum.internal.vitals.JankStatsActivityLifecycleListener.onActivityStopped(JankStatsActivityLifecycleListener.kt:99)
        ....
```

This PR does the following:

* Checks that tracking is enabled before disabling it. If it is already disabled - we report it to the telemetry, because probably the issue is on our side.
* If still there is a crash when we disabling jank stats even in a proper state (when it is enabled) - we still report it to the telemetry, but in this case the problem is in the `JankStats` class itself probably.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
@@ -75,6 +75,7 @@ internal class JankStatsActivityLifecycleListener(
     override fun onActivityPaused(activity: Activity) {
     }
 
+    @Suppress("NestedBlockDepth")
     @MainThread
     override fun onActivityStopped(activity: Activity) {
         val window = activity.window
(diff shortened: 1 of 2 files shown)


xgouchet (approved):
Nice catch :)

mariusc83 (approved):

jonathanmos (approved):

PR #1600, opened 2023-09-06
RUM-1012 Fix RippleDrawables

jonathanmos:
### What does this PR do?
Fixes RippleDrawables not converting correctly to base64 (caused by it's dimensions being -1, and RippleDrawable not being the actual drawable from which we can create a bitmap). In this fix, the InsetDrawable is extracted from the RippleDrawable, and the GradientDrawable is extracted from that - we then use the dimensions of the view as the dimensions of the drawable (this doesn't affect the created bitmap so much, since the bitmap is rescaled to 10kbs or less before caching). 

### Motivation

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -129,6 +135,17 @@ internal class ImageWireframeHelper(
         return result
     }
 
+    private fun resolveDrawableProperties(view: View, drawable: Drawable?): DrawableProperties {
```
mariusc83, line 138: nice ;)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
```diff
@@ -180,7 +180,7 @@ abstract class BaseWireframeMapper<T : View, S : MobileSegment.Wireframe>(
             y = bounds.y,
             width,
             height,
-            view.background,
+            view.background?.constantState?.newDrawable(),
```
mariusc83, line 183: why do you create a new drawable here ? isn't this costly for each view ?

jonathanmos: preferably we wouldn't need to do this, however if the drawable is not moved to a new instance before we create the bitmap, then drawing the bitmap on the canvas modifies the view in the app itself.

mariusc83: ok I see so you need to take a copy ...hmmm...I wonder how expensive this is for each view. Sorry for asking but why do you need to modify the Drawable to create the bitmap from it ?

jonathanmos: I need to set the bounds before I can draw to the canvas

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -129,6 +135,17 @@ internal class ImageWireframeHelper(
...
+        if (drawable == null) return DrawableProperties(null, 0, 0)
+
+        return when (drawable) {
+            is LayerDrawable -> resolveDrawableProperties(view, drawable.getDrawable(0))
+            is InsetDrawable -> resolveDrawableProperties(view, drawable.drawable)
```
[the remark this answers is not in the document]
mariusc83: for me it looks fine like this also

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseWireframeMapper.kt
mariusc83: I'd say that if profiling it shows decent number we can go with this like that. @0xnm, @xgouchet waiting for your thoughts on this one. I searched for a way to go around it, but apparently there's not, the only other way is to `setBounds` when before `drawing` and try to put the drawable back into its previous state after. Not sure how reliable this is and honestly I think I would rather go with the copy.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -37,13 +40,14 @@ internal class ImageWireframeHelper(
         prefix: String = DRAWABLE_CHILD_NAME
     ): MobileSegment.Wireframe.ImageWireframe? {
         val id = uniqueIdentifierGenerator.resolveChildUniqueIdentifier(view, prefix + currentWireframeIndex)
+        val (currentDrawable, drawableWidth, drawableHeight) = resolveDrawableProperties(view, drawable)
```
xgouchet, line 43: Why unbox the components here when. you have an object storing the info? You could then remove the `@Suppress("ComplexCondition")` below by having a `DrawableProperties.isValid(): Boolean` method

xgouchet (changes requested):
I'm not sure I understand the need to have the ripple drawables converted to base64? 
By definition a RippleDrawable is an animated drawable used to show feedback when tapping on interactive views (e.g. buttons). 
Saving those as bitmaps means we'll either have to save the whole animation because everytime we draw them the animation state will be different, or it's going to only store a single ripple state which will look weird in the replay. 
Either way, should'n we instead save them as a solid background for now, and wait for a better solution to record/replay the ripple effect?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
mariusc83: @xgouchet ripple drawable are by default used as backgrounds for a View (button, text, etc).

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/wrappers/BitmapWrapper.kt
```diff
@@ -27,4 +27,21 @@ internal class BitmapWrapper {
...
+        dstHeight: Int,
+        filter: Boolean
+    ): Bitmap? {
+        @Suppress("SwallowedException")
+        return try {
+            Bitmap.createScaledBitmap(src, dstWidth, dstHeight, filter)
+        } catch (e: IllegalArgumentException) {
+            // should never happen since config is given as valid type and width/height are
+            // normalized to be at least 1
+            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
```
jonathanmos, line 43: I can add the logger in this pr, but because the logger is not passed at the moment in base64 it will require a change to several classes that may be better to do separately (and if so, for all necessary places in base64)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -129,6 +128,23 @@ internal class ImageWireframeHelper(
...
 
+    private fun resolveDrawableProperties(view: View, drawable: Drawable?): DrawableProperties {
+        if (drawable == null) return DrawableProperties(null, 0, 0)
+
+        return when (drawable) {
+            is LayerDrawable -> {
+                if (drawable.numberOfLayers > 0) {
+                    resolveDrawableProperties(view, drawable.getDrawable(0))
+                } else {
+                    DrawableProperties(drawable, drawable.intrinsicWidth, drawable.intrinsicHeight)
```
jonathanmos, line 139: Perhaps the drawable should be returned here as null

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: @xgouchet regarding whether we should be rendering rippledrawable as imagewireframes, I understand you discussed with marius - we have a problem getting the color of the drawable in order to use it in a shape, so need to render the bitmap (internally the rippledrawables are using gradientdrawable)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -50,29 +50,42 @@ internal class Base64Serializer private constructor(
...
 
-        var shouldCacheBitmap = false
         val cachedBase64 = base64LRUCache?.get(drawable)
         if (cachedBase64 != null) {
             finalizeRecordedDataItem(cachedBase64, imageWireframe, asyncImageProcessingCallback)
             return
         }
 
+        var shouldCacheBitmap = false
+
```
mariusc83, line 69: I don't see where is the unit test related with this fix

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
mariusc83: so basically we do not care anymore about `RippleDrawable` but about his super class. Much more reliable this way I'd say.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/wrappers/BitmapWrapper.kt
mariusc83: you do not have to change it now but for situations like this one we usually prefer having extension methods as:
```
fun Bitmap.safeCreateScaledBitmap(..)
```

mariusc83 (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/wrappers/BitmapWrapper.kt
jonathanmos: Specifically in this case, I'm not sure that its supported because these are static methods of the class. I see that [support for extending static methods](https://github.com/Kotlin/KEEP/issues/348) is still in the [proposal stage](https://github.com/Kotlin/KEEP/blob/statics/proposals/statics.md#static-extensions) for kotlin.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: There are tests for the bitmapDrawable scaling [here](https://github.com/DataDog/dd-sdk-android/blob/jmoskovich/rum-1012/fix-rippledrawables/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64SerializerTest.kt#L467)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/wrappers/CanvasWrapper.kt
```diff
@@ -8,19 +8,38 @@ package com.datadog.android.sessionreplay.internal.recorder.wrappers
 
 import android.graphics.Bitmap
 import android.graphics.Canvas
+import com.datadog.android.api.InternalLogger
 
-internal class CanvasWrapper {
+internal class CanvasWrapper(
+    private val logger: InternalLogger = InternalLogger.UNBOUND
```
mariusc83, line 14: why do you use an unbound logger here ?

jonathanmos: At the moment the internalLogger is not passed into the base64 structure. The logging change can be made in this PR but it will increase the scope quite a bit, and I preferred to leave it for a later PR where we can fix all the base64 loggers at once

mariusc83 (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/LayerDrawableExt.kt
```diff
@@ -0,0 +1,27 @@
...
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.LayerDrawable
+import com.datadog.android.api.InternalLogger
+
+@Suppress("TooGenericExceptionCaught")
+internal fun LayerDrawable.safeGetDrawable(index: Int, logger: InternalLogger = InternalLogger.UNBOUND): Drawable? {
+    return try {
+        this.getDrawable(index)
+    } catch (e: IndexOutOfBoundsException) {
+        // this should never happen
```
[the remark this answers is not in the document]
jonathanmos: yes, I can add this. Should we be logging these cases? My concern is that if we just return null here we won't know that we're passing invalid indexes in the logic

xgouchet (dismissed):

xgouchet (approved):

mariusc83 (approved):

PR #1613, opened 2023-09-12
RUM-1042 Fix base64 issues with multithreading

jonathanmos:
### What does this PR do?
Fixes two issues that were discovered during functional tests. 

The first issue is having the `AsyncImageProcessingCallback` held by the `Base64Serializer` instance, and this can lead to the next bitmap replacing the callback before the first one has completed. This can lead to the decrement of pending images happening on the wrong node. Solution: passing the asyncCallback in every call to handleBitmap.

The second issue is having the previous `RumContext` held by the `RumContextData` object. This can lead to an item being removed from the queue (lost - not processed) with prevContext A and newContext B, and the next item with prevContext B newContext B will no longer be seen as a new view - so no meta data. Solution: the RecordedDataProcessor now holds it's own prevRumContext variable and replaces it only after processing an item. 

### Motivation
Issues found during functional tests.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
@@ -24,28 +24,28 @@ internal class RecordedDataProcessor(
     private val mutationResolver: MutationResolver,
     private val nodeFlattener: NodeFlattener = NodeFlattener()
 ) : Processor {
-
     private var prevSnapshot: List<MobileSegment.Wireframe> = emptyList()
     private var lastSnapshotTimestamp = 0L
     private var previousOrientation = Configuration.ORIENTATION_UNDEFINED
+    private var prevRumContext: SessionReplayRumContext = SessionReplayRumContext()
 
     @WorkerThread
     override fun processScreenSnapshots(
         item: SnapshotRecordedDataQueueItem
     ) {
         handleSnapshots(
-            newRumContext = item.rumContextData.newRumContext,
-            prevRumContext = item.rumContextData.prevRumContext,
-            timestamp = item.rumContextData.timestamp,
+            newRumContext = item.recordedQueuedItemContext.newRumContext,
+            timestamp = item.recordedQueuedItemContext.timestamp,
             snapshots = item.nodes,
             systemInformation = item.systemInformation
         )
+        prevRumContext = item.recordedQueuedItemContext.newRumContext
     }
 
     @WorkerThread
     override fun processTouchEventsRecords(item: TouchEventRecordedDataQueueItem) {
         handleTouchRecords(
-            rumContext = item.rumContextData.newRumContext,
+            rumContext = item.recordedQueuedItemContext.newRumContext,
             touchData = item.touchData
         )
     }
@@ -64,7 +64,6 @@ internal class RecordedDataProcessor(
     @WorkerThread
     private fun handleSnapshots(
         newRumContext: SessionReplayRumContext,
-        prevRumContext: SessionReplayRumContext,
         timestamp: Long,
         snapshots: List<Node>,
         systemInformation: SystemInformation
@@ -76,13 +75,13 @@ internal class RecordedDataProcessor(
         }
 
         val records: MutableList<MobileSegment.MobileRecord> = LinkedList()
-        val isNewView = isNewView(prevRumContext, newRumContext)
+        val isNewView = isNewView(newRumContext)
         val isTimeForFullSnapshot = isTimeForFullSnapshot()
         val screenOrientationChanged = systemInformation.screenOrientation != previousOrientation
         val fullSnapshotRequired = isNewView || isTimeForFullSnapshot || screenOrientationChanged
 
         if (isNewView) {
-            handleViewEndRecord(prevRumContext, timestamp)
+            handleViewEndRecord(timestamp)
             val screenBounds = systemInformation.screenBounds
             val metaRecord = MobileSegment.MobileRecord.MetaRecord(
                 timestamp,
@@ -142,7 +141,7 @@ internal class RecordedDataProcessor(
         }
     }
 
-    private fun handleViewEndRecord(prevRumContext: SessionReplayRumContext, timestamp: Long) {
+    private fun handleViewEndRecord(timestamp: Long) {
         if (prevRumContext.isValid()) {
             // send first the ViewEndRecord for the previous RUM context (View)
             val viewEndRecord = MobileSegment.MobileRecord.ViewEndRecord(timestamp)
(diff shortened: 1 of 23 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
```diff
@@ -24,22 +24,22 @@ internal class RecordedDataProcessor(
     private val mutationResolver: MutationResolver,
     private val nodeFlattener: NodeFlattener = NodeFlattener()
 ) : Processor {
-
     private var prevSnapshot: List<MobileSegment.Wireframe> = emptyList()
     private var lastSnapshotTimestamp = 0L
     private var previousOrientation = Configuration.ORIENTATION_UNDEFINED
+    private var prevRumContext: SessionReplayRumContext = SessionReplayRumContext()
```
0xnm, line 30: I'm not quite clear what this change does: if `RecordedDataProcessor#processScreenSnapshots` is called always from the same thread, then it shouldn't be the difference before and after this change. If it is called from multiple threads, then how do we avoid concurrency issues here?

 Also is there any impact on storing this as a property from the fact that we have early return in `RecordedDataProcessor#handleSnapshots` (when wireframes are empty).

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
```diff
@@ -48,6 +48,7 @@ internal class RecordedDataProcessor(
             rumContext = item.rumContextData.newRumContext,
             touchData = item.touchData
         )
+        prevRumContext = item.rumContextData.newRumContext
```
mariusc83, line 51: if you add this here and by any chance you have a touch record that is send before a snapshot record when the view is activity/fragment is opened, this will break your logic as you do not send the `isNewView` records in the `handleTouchRecords`

mariusc83: These fixes will require tests to cover all cases here

mariusc83 (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
0xnm: Thanks for the explanation!

> It's called on a synchronized method that launches a new background thread each time (although the threadpool size is 1, so at most one thread alive at any time)

`RecordedDataQueueHandler` has a `ThreadPoolExecutor` as executor service, this implementation won't create a **new** thread each time, it will have one and the same thread running always unless it dies - in this case it will create a new one.

Anyway, it won't change anything, because this method is called from the synchronised one.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RumContextData.kt
```diff
@@ -10,6 +10,5 @@ import com.datadog.android.sessionreplay.internal.utils.SessionReplayRumContext
 
 internal data class RumContextData(
     internal val timestamp: Long,
-    internal val newRumContext: SessionReplayRumContext,
-    internal val prevRumContext: SessionReplayRumContext
+    internal val newRumContext: SessionReplayRumContext
```
0xnm, line 13: I guess after removal of `prevRumContext` having the naming `newRumContext` can be a bit confusing. Maybe this class should be called `TimedSessionReplayRumContext`? it seems it carries only `SessionReplayRumContext` and `timestamp` after all. WDYT? Not pushing for it anyway.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessorTest.kt
```diff
@@ -860,6 +852,39 @@ internal class RecordedDataProcessorTest {
...
+        testedProcessor.processScreenSnapshots(snapshot)
+
+        currentRumContextData = mockRumContextDataHandler.createRumContextData() ?: fail("RumContextData is null")
+        val snapshot2 = createSnapshotItem(fakeSnapshot2)
+        testedProcessor.processScreenSnapshots(snapshot2)
+
+        // Then
+        val captor = argumentCaptor<EnrichedRecord>()
+        verify(mockWriter, times(3)).write(captor.capture())
+        val relevantItem = captor.allValues[2]
```
0xnm, line 881:
```suggestion
        val relevantItem = captor.lastValue
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessorTest.kt
```diff
@@ -908,6 +932,226 @@ internal class RecordedDataProcessorTest {
...
+        val endSnapshot = captor.allValues[1].records
+        val newSnapshot = captor.allValues[2].records
+        val touchEvent = captor.allValues[3].records
+
+        assertThat(initialSnapshot[0] is MobileSegment.MobileRecord.MetaRecord).isTrue
+        assertThat(initialSnapshot[1] is MobileSegment.MobileRecord.FocusRecord).isTrue
+        assertThat(endSnapshot[0] is MobileSegment.MobileRecord.ViewEndRecord).isTrue
+        assertThat(newSnapshot[0] is MobileSegment.MobileRecord.MetaRecord).isTrue
+        assertThat(newSnapshot[1] is MobileSegment.MobileRecord.FocusRecord).isTrue
+        assertThat(touchEvent[0] is MobileSegment.MobileRecord.MetaRecord).isFalse
```
0xnm, line 990: I think it is better to use `assertThat(...).isInstanceOf(...)`, because in this case the message will be more explicit in case of failure. Same applies for all tests below.

mariusc83 (dismissed):
LGTM

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RumContextData.kt
jonathanmos: 👍 I'll rename this to `RecordedQueuedItemContext`

0xnm (approved):

mariusc83 (approved):

PR #1619, opened 2023-09-15
Fix the async image loading logic inside the SR view mappers

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/DecorViewMapper.kt
```diff
@@ -15,11 +16,16 @@ import java.util.Locale
 internal class DecorViewMapper(
     private val viewWireframeMapper: ViewWireframeMapper,
     private val uniqueIdentifierGenerator: UniqueIdentifierGenerator = UniqueIdentifierGenerator
-) : BaseWireframeMapper<View, MobileSegment.Wireframe.ShapeWireframe>() {
+) : WireframeMapper<View, MobileSegment.Wireframe> {
```
mariusc83, line 19: we do not need the `BaseWireframeMapper` anymore here

features/dd-sdk-android-session-replay/api/apiSurface
```diff
@@ -22,25 +25,26 @@ interface com.datadog.android.sessionreplay.internal.recorder.OptionSelectorDete
   fun isOptionSelector(android.view.ViewGroup): Boolean
 data class com.datadog.android.sessionreplay.internal.recorder.SystemInformation
   constructor(GlobalBounds, Int = Configuration.ORIENTATION_UNDEFINED, Float, String? = null)
-abstract class com.datadog.android.sessionreplay.internal.recorder.mapper.BaseWireframeMapper<T: android.view.View, S: com.datadog.android.sessionreplay.model.MobileSegment.Wireframe> : WireframeMapper<T, com.datadog.android.sessionreplay.model.MobileSegment.Wireframe>, com.datadog.android.sessionreplay.internal.AsyncImageProcessingCallback
+abstract class com.datadog.android.sessionreplay.internal.recorder.mapper.BaseAsyncBackgroundWireframeMapper<T: android.view.View> : BaseWireframeMapper<T, com.datadog.android.sessionreplay.model.MobileSegment.Wireframe>
+  constructor(com.datadog.android.sessionreplay.utils.StringUtils = StringUtils, com.datadog.android.sessionreplay.utils.ViewUtils = ViewUtils)
+  override fun map(T, com.datadog.android.sessionreplay.internal.recorder.MappingContext, com.datadog.android.sessionreplay.internal.AsyncJobStatusCallback): List<com.datadog.android.sessionreplay.model.MobileSegment.Wireframe>
+  companion object
```
0xnm, line 31: if it doesn't expose public members, then we can make it `internal`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseAsyncBackgroundWireframeMapper.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Maps the [View] into a list of [MobileSegment.Wireframe].
+     */
+    override fun map(
+        view: T,
+        mappingContext: MappingContext,
+        asyncJobStatusCallback: AsyncJobStatusCallback
+    ):
+        List<MobileSegment.Wireframe> {
```
0xnm, line 48:
```suggestion
    ): List<MobileSegment.Wireframe> {
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseAsyncBackgroundWireframeMapper.kt
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * Maps the [View] into a list of [MobileSegment.Wireframe].
+     */
+    override fun map(
+        view: T,
+        mappingContext: MappingContext,
+        asyncJobStatusCallback: AsyncJobStatusCallback
+    ):
```
0xnm, line 47: we need to add a good documentation for this method in the future, because it is not clear why do we have such API - we return some result instantly and at the same time have an async callback with some notification mechanism. Usually APIs have either of the two - return result in a blocking way, or give it in the callback.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/QueueableViewMapper.kt
```diff
@@ -7,25 +7,26 @@
...
 
 internal class QueueableViewMapper(
     private val mapper: WireframeMapper<View, *>,
     private val recordedDataQueueRefs: RecordedDataQueueRefs
-) : BaseWireframeMapper<View, MobileSegment.Wireframe>(), AsyncImageProcessingCallback {
-    override fun map(view: View, mappingContext: MappingContext):
+) : AsyncJobStatusCallback {
+    fun map(view: View, mappingContext: MappingContext):
         List<MobileSegment.Wireframe> {
-        (mapper as? BaseWireframeMapper<View, *>)?.registerAsyncImageProcessingCallback(this)
```
mariusc83, line 21: this was the main issue here as we were still registering a callback

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/WireframeMapper.kt
```diff
@@ -22,9 +24,16 @@ interface WireframeMapper<in T : View, out S : MobileSegment.Wireframe> {
      * Maps a [View] to a [List<Wireframe>] in order to be rendered in the Session Replay player.
      * @param view as the [View] instance that will be mapped
      * @param mappingContext in which we provide useful information regarding the current
-     * system state.
+     * @param asyncJobStatusCallback a callback that can be called when the mapper starts or
+     * finishes processing an async job. This is useful to keep track of the current async jobs
```
0xnm, line 28: since both `jobStarted` and `jobFinished` doesn't have any arguments like job ID and result, maybe we can elaborate why it is useful? Without any arguments passed it can probably be served as a counter only.

0xnm (approved):
looks good!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseAsyncBackgroundWireframeMapper.kt
mariusc83: I do agree, maybe once we will have a stable version, meaning the new resources endpoint in please (which will remove the base64 logic) and the fine grained masking we could simplify a bit this part. My plan is to assess after if we still need this queue or not. We already copy the drawable now so I am thinking to actually create this `ViewPlacehdolder` model that will be processed into wireframe at the `Processor` level. For now we are not ready for this switch as we do not know if copying the drawable will always be possible and also need to see how the logic will look like with the new resources endpoint in place.

features/dd-sdk-android-session-replay/api/apiSurface
mariusc83: actually not as we have the `TextViewMapper` class which extends it and it is public to be used in the other module

jonathanmos (dismissed):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseAsyncBackgroundWireframeMapper.kt
```diff
@@ -0,0 +1,151 @@
...
+import com.datadog.android.sessionreplay.internal.recorder.base64.ImageWireframeHelper
+import com.datadog.android.sessionreplay.internal.recorder.base64.ImageWireframeHelperCallback
+import com.datadog.android.sessionreplay.internal.recorder.densityNormalized
+import com.datadog.android.sessionreplay.model.MobileSegment
+import com.datadog.android.sessionreplay.utils.StringUtils
+import com.datadog.android.sessionreplay.utils.UniqueIdentifierGenerator
+import com.datadog.android.sessionreplay.utils.ViewUtils
+
+@Suppress("UndocumentedPublicClass")
+abstract class BaseAsyncBackgroundWireframeMapper<T : View>(
```
0xnm, line 22: minor: does it make sense to use word `async` here in the class naming? `WireframeMapper` interface already hints that it supports async jobs by having async callback argument in `map` method.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/WireframeMapper.kt
```diff
@@ -22,9 +24,18 @@ interface WireframeMapper<in T : View, out S : MobileSegment.Wireframe> {
      * Maps a [View] to a [List<Wireframe>] in order to be rendered in the Session Replay player.
      * @param view as the [View] instance that will be mapped
      * @param mappingContext in which we provide useful information regarding the current
-     * system state.
+     * @param asyncJobStatusCallback a callback that can be called when the mapper starts or
+     * finishes processing an async job. By calling this whenever a job started (in the main thread)
+     * and finished (in the background thread) you will make sure that the `List<Wireframe>` will
+     * not be consumed until all the wireframes are updated by the async jobs. The main reason we
+     * needed this functionality was to be able to unblock the main thread whenever we needed
+     * heavy operations like image capture or serialisation.
```
0xnm, line 32: I would re-word this sentence about why it is needed. Not to say "we needed this functionality", but something like "it can be used to offload heavy work from the calling thread (like main) ... ", in such fashion.

0xnm (dismissed):
lgtm. I've added few minor suggestions.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseAsyncBackgroundWireframeMapper.kt
mariusc83: yeah it tells that the background will be handled in async way, hmm maybe I can find a better name for this if not will leave it like this.

0xnm (approved):

PR #1625, opened 2023-09-19
RUM-985 propagate session state and view type as Strings

xgouchet:
### What does this PR do?

The RUM `session_state` and `view_type` were propagated as internal enums in the shared RUM feature context. This means that other modules could not use those values (except with a call to `toString()`, which could have resulted in obfuscated results). 

This PR ensures the data is shared as a String, to be manipulated unobfuscated from any module using it. 

### Motivation

Part of the work done to make the RUM sampling consistent between Webviews and the native app.


0xnm (approved):
nice catch! another option btw is to add the rules to the `proguard` file we ship with the library, was it considered? anyway, I'm fine with this solution as well.

mariusc83 (approved):

xgouchet:
@0xnm 
> another option btw is to add the rules to the proguard file we ship with the library, was it considered? anyway, I'm fine with this solution as well.
I thought about it, but since those enums are internal anyway, it felt wrong to add a proguard rule for it

PR #1626, opened 2023-09-19
RUM-985 Only send Webview RUM events when Native Session exists and is tracked

xgouchet:
### What does this PR do?

The rule is that a browser RUM session should be tracked as part of the host application's session. This means that if the native session is not tracked, or if the RUM feature is not enabled, then the webview RUM events should be dropped.


0xnm (approved):

mariusc83 (approved):
LGTM

0xnm (approved):

PR #1632, opened 2023-09-21
RUM-1094: Let exceptions from `Window.Callback` to propagate

0xnm:
### What does this PR do?

After checking our telemetry I found some customer-specific exceptions reported. Turns out we are swallowing exceptions thrown from the wrapped `Window.Callback`, which it seems is not the right approach - we need to let them propagate, so that customer is aware of them. This PR basically reverts what was done in https://github.com/DataDog/dd-sdk-android/pull/732

However, we need to catch exceptions related to the delegate calling - argument can be `null`, while delegate can exception only non-null values. Such exceptions were the case in the past, but it seems that at least Kotlin 1.8 is now smarter and doesn't generate the null-parameter check as before. But I still left such exception handling just to avoid regressions in the future Kotlin compiler releases.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/WindowCallbackWrapper.kt
```diff
@@ -26,7 +26,7 @@ internal class WindowCallbackWrapper(
     window: Window,
     val sdkCore: SdkCore,
     val wrappedCallback: Window.Callback,
-    val gesturesDetector: GesturesDetectorWrapper,
+    private val gesturesDetector: GesturesDetectorWrapper,
     val interactionPredicate: InteractionPredicate = NoOpInteractionPredicate(),
     val copyEvent: (MotionEvent) -> MotionEvent = {
```
mariusc83, line 31: Can we add the same fixes in the `WindowCallbackWrapper` for SR please ? Maybe add it in this PR

mariusc83: ah no...sorry didn't go through all the PR 😢 . My bad, tks a tons for the update will review the whole PR later.

xgouchet (approved):

mariusc83 (approved):

PR #1633, opened 2023-09-21
RUM-1095 Fix the batch duration value in batch_close telemetry event

mariusc83:
### What does this PR do?

Before we were computing the batch duration value for the `batch_close` telemetry event as: `System.currentTimeInMillis - lastTimeBatchWasAccessed` which was completely wrong. In this PR we are addressing this issue.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
@@ -11,6 +11,7 @@ import com.datadog.android.api.feature.Feature
 import com.datadog.android.core.internal.configuration.DataUploadConfiguration
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleMonitor
 import com.datadog.android.core.internal.persistence.file.FilePersistenceConfig
+import com.datadog.android.core.internal.persistence.file.existsSafe
 import com.datadog.android.core.internal.persistence.file.lengthSafe
 import com.datadog.android.core.internal.time.TimeProvider
 import com.datadog.android.core.sampling.RateBasedSampler
@@ -47,7 +48,7 @@ internal class BatchMetricsDispatcher(
     }
 
     override fun sendBatchClosedMetric(batchFile: File, batchMetadata: BatchClosedMetadata) {
-        if (trackName == null || !sampler.sample()) {
+        if (trackName == null || !sampler.sample() || !batchFile.existsSafe(internalLogger)) {
             return
         }
         resolveBatchClosedMetricAttributes(batchFile, batchMetadata)?.let {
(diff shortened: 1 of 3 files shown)


0xnm (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -123,17 +121,17 @@ internal class BatchMetricsDispatcher(
         )
     }
 
-    private fun resolveFileAge(file: File): Long {
+    private fun File.nameAsTimestampSafe(): Long? {
         return try {
-            dateTimeProvider.getDeviceTimestamp() - file.name.toLong()
+            this.name.toLong()
```
xgouchet, line 126: we could use `.toLongOrNull()` to avoid throwing an exception

xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
mariusc83: I wanted to have the exception just for us though...to see in case we have a weird filename in there. What do you think ?

0xnm (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
xgouchet: You can still log the filename when the result is null, but throwing an exception costs resources, it's more efficient to avoid it when we can

jonathanmos (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -47,7 +48,7 @@ internal class BatchMetricsDispatcher(
     }
 
     override fun sendBatchClosedMetric(batchFile: File, batchMetadata: BatchClosedMetadata) {
-        if (trackName == null || !sampler.sample()) {
+        if (!batchFile.existsSafe(internalLogger) || trackName == null || !sampler.sample()) {
```
0xnm, line 51: small performance gain - let's put the check if file exists to the right - this way we won't do it if metric is not sampled, we are saving some I/O capacity and block the thread less

```suggestion
        if (trackName == null || !sampler.sample() || !batchFile.existsSafe(internalLogger)) {
```

jonathanmos (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
mariusc83: Good catch

0xnm (approved):

PR #1640, opened 2023-09-26
RUM-938 Implement heuristic image classification

jonathanmos:
### What does this PR do?
Implement a heuristic image classification to determine whether an image is PII or not. If the image is PII then send a placeholder wireframe identifying the image as a content image. 

The identification is by checking that the dimensions of the image do not exceed 48dp (the largest icon in the material design guidelines). In the case of GradientDrawable we ignore this check, since GradientDrawables cannot be PII and constitute the backgrounds for RippleDrawables. 

### Motivation
Avoid sending PII as part of base64.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageTypeResolver.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageTypeResolver.kt
@@ -1,0 +1,25 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.GradientDrawable
+import androidx.annotation.VisibleForTesting
+
+internal class ImageTypeResolver {
+    fun isDrawablePII(drawable: Drawable, widthInDp: Int, heightInDp: Int): Boolean =
+        drawable !is GradientDrawable &&
+            (
+                widthInDp >= IMAGE_DIMEN_CONSIDERED_PII_IN_DP ||
+                    heightInDp >= IMAGE_DIMEN_CONSIDERED_PII_IN_DP
+                )
+
+    internal companion object {
+        // material design icon size is up to 48x48
+        @VisibleForTesting internal const val IMAGE_DIMEN_CONSIDERED_PII_IN_DP = 49
+    }
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
@@ -17,7 +17,6 @@ import androidx.annotation.VisibleForTesting
 import com.datadog.android.sessionreplay.internal.recorder.MappingContext
 import com.datadog.android.sessionreplay.internal.recorder.ViewUtilsInternal
 import com.datadog.android.sessionreplay.internal.recorder.densityNormalized
-import com.datadog.android.sessionreplay.internal.recorder.safeGetDrawable
 import com.datadog.android.sessionreplay.model.MobileSegment
 import com.datadog.android.sessionreplay.utils.UniqueIdentifierGenerator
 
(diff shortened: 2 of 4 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageTypeResolver.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.GradientDrawable
+import androidx.annotation.VisibleForTesting
+
+internal class ImageTypeResolver {
+    fun isDrawablePII(drawable: Drawable, widthInDp: Int, heightInDp: Int): Boolean =
+        drawable !is GradientDrawable &&
+            (
+                widthInDp >= IMAGE_DIMEN_CONSIDERED_PII_IN_DP ||
```
mariusc83, line 17: here you are using DP but I see that you are calling it with a `densityNormalized` value.

mariusc83 (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageTypeResolver.kt
jonathanmos: `DensityNormalized` divides the input by the density. Does this not mean dp? (pixels/density = dp)

mariusc83 (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
```diff
@@ -55,6 +57,25 @@ internal class ImageWireframeHelper(
         val displayMetrics = view.resources.displayMetrics
         val applicationContext = view.context.applicationContext
         val mimeType = imageCompression.getMimeType()
+        val density = displayMetrics.density
+
+        // in case we suspect the image is PII, return a placeholder
+        @Suppress("UnsafeCallOnNullableType") // drawable already checked for null in isValid
+        if (imageTypeResolver.isDrawablePII(
+                drawableProperties.drawable!!,
```
0xnm, line 65: minor thought: maybe we can avoid this `!!` usage? I can see that `DrawableProperties#isValid` above is checking that drawable is not null, but compiler doesn't know about it and moreover `drawable` getter may do something different.

Maybe we can get rid of `isValid` property and checking things here? This way compiler will know that `drawable` is not null. Or even change the type of `drawable` in `DrawableProperties` to be non-null (is `DrawablePropeties` useful without drawable)?

this is the only usage of `!!` in production codebase and this is a bit confusing.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageTypeResolver.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
+
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.GradientDrawable
+import androidx.annotation.VisibleForTesting
+
+internal class ImageTypeResolver {
+    fun isDrawablePII(drawable: Drawable, widthInDp: Int, heightInDp: Int): Boolean =
+        drawable !is GradientDrawable &&
```
0xnm, line 15: what is special about `GradientDrawable`?

jonathanmos: GradientDrawable can't be PII because it's just a color gradient, but it receives the dimensions of the view so it would fall foul of this heuristic

0xnm: in theory everything is possible - the most naive counter-example is encoding of some PII information as color. Like QR-code carrying PII, but as Gradient; we don't know what apps can pop up in the market.

jonathanmos: That's true, on the other hand if we don't exclude this drawable type we may lose the backgrounds of most RippleDrawables since GradientDrawable is at their base

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageTypeResolver.kt
```diff
@@ -0,0 +1,25 @@
...
+            (
+                widthInDp >= IMAGE_DIMEN_CONSIDERED_PII_IN_DP ||
+                    heightInDp >= IMAGE_DIMEN_CONSIDERED_PII_IN_DP
+                )
+
+    internal companion object {
+        // material design icon size is up to 48x48
+        @VisibleForTesting internal const val IMAGE_DIMEN_CONSIDERED_PII_IN_DP = 49
```
jonathanmos, line 23: Does this seem reasonable? Should we increase to 100dp?

0xnm: I would say let's keep it and then just collect feedback/watch replays of our apps.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ImageWireframeHelper.kt
jonathanmos: I'll refactor the code to make DrawableProperties drawable non-nullable

0xnm (approved):

mariusc83 (approved):

PR #1641, opened 2023-09-26
RUM-1187 Add more information into the batch telemetry

mariusc83:
### What does this PR do?

Following the latest discussions regarding the problem with the batch telemetry data for Android SDK we decided to add more information into the related logs:

- thread name (might be useful for potential concurrency debug)
- the batch file name 
- the tracking consent according with the batch origin folder

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
@@ -11,11 +11,13 @@ import com.datadog.android.api.feature.Feature
 import com.datadog.android.core.internal.configuration.DataUploadConfiguration
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleMonitor
 import com.datadog.android.core.internal.persistence.file.FilePersistenceConfig
+import com.datadog.android.core.internal.persistence.file.advanced.FeatureFileOrchestrator
 import com.datadog.android.core.internal.persistence.file.existsSafe
 import com.datadog.android.core.internal.persistence.file.lengthSafe
 import com.datadog.android.core.internal.time.TimeProvider
 import com.datadog.android.core.sampling.RateBasedSampler
 import com.datadog.android.core.sampling.Sampler
+import com.datadog.android.privacy.TrackingConsent
 import java.io.File
 import java.util.Locale
 import java.util.concurrent.atomic.AtomicBoolean
@@ -99,7 +101,10 @@ internal class BatchMetricsDispatcher(
             UPLOADER_WINDOW_KEY to filePersistenceConfig.recentDelayMs,
 
             BATCH_REMOVAL_KEY to deletionReason.toString(),
-            IN_BACKGROUND_KEY to isInBackground.get()
+            IN_BACKGROUND_KEY to isInBackground.get(),
+            TRACKING_CONSENT_KEY to file.resolveFileOriginAsConsent(),
+            FILE_NAME to file.name,
+            THREAD_NAME to Thread.currentThread().name
         )
     }
 
@@ -118,7 +123,10 @@ internal class BatchMetricsDispatcher(
             // be sent as a batch_delete telemetry later
             BATCH_SIZE_KEY to file.lengthSafe(internalLogger),
             BATCH_EVENTS_COUNT_KEY to batchMetadata.eventsCount,
-            FORCE_NEW_KEY to batchMetadata.forcedNew
+            FORCE_NEW_KEY to batchMetadata.forcedNew,
+            TRACKING_CONSENT_KEY to file.resolveFileOriginAsConsent(),
+            FILE_NAME to file.name,
+            THREAD_NAME to Thread.currentThread().name
         )
     }
 
@@ -141,6 +149,17 @@ internal class BatchMetricsDispatcher(
             Feature.TRACING_FEATURE_NAME -> TRACE_TRACK_NAME
             Feature.SESSION_REPLAY_FEATURE_NAME -> SR_TRACK_NAME
             else -> null
+        }
+    }
+
+    private fun File.resolveFileOriginAsConsent(): String? {
+        val fileDirectory = this.parentFile?.name ?: return null
+        return if (fileDirectory.matches(FeatureFileOrchestrator.IS_PENDING_DIR_REG_EX)) {
+            TrackingConsent.PENDING.toString().lowercase(Locale.US)
+        } else if (fileDirectory.matches(FeatureFileOrchestrator.IS_GRANTED_DIR_REG_EX)) {
+            TrackingConsent.GRANTED.toString().lowercase(Locale.US)
+        } else {
+            null
         }
     }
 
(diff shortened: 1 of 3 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -218,6 +235,15 @@ internal class BatchMetricsDispatcher(
...
         internal const val BATCH_CLOSED_TYPE_VALUE = "batch closed"
 
+        /* The value of the tracking consent according with this file origin.*/
+        internal const val TRACKING_CONSENT_KEY = "consent"
+
+        /* The file name.*/
+        internal const val FILE_NAME = "name"
+
+        /* The thread name from which the current metric was sent.*/
+        internal const val THREAD_NAME = "thread"
```
0xnm, line 245: how it can be useful? is it actionable attribute and what knowledge can it give?

mariusc83: we decided that is better to have more and remove it later if not needed. This is for tracking any concurrency issue. We know that theoretically a `batch_close` should be sent with more or less same thread (depending of the delay between them as the executor could kill that thread and spawn a new one).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -144,6 +152,15 @@ internal class BatchMetricsDispatcher(
         }
     }
 
+    private fun File.resolveFileOriginAsConsent(): String? {
+        val fileDirectory = this.parentFile?.name ?: return null
+        return if (fileDirectory.endsWith(FeatureFileOrchestrator.PENDING_DIR_SUFFIX)) {
```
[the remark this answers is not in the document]
mariusc83: I was not sure about this one either...is good that you added the comment. I am wondering what should we use for granted specifically in the assertion of the folder name ? We do not have any `granted` in the granted folder name. Would a check for `doenNotEndWith(PENDING_DIR_SUFFIX)` be enough ?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
0xnm: so we are going to assert is 2 values between 2 metrics from the same source are equal or not? Not sure if it is possible in automatic way, but okay, fair enough (not still sure that it can help us).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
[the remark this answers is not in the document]
0xnm: good point that we don't have `granted` explicitly. Maybe we that can check that folder name matches `granted` folder name pattern? It should be an easy and cheap regex.

Anyway, we should have a solution which doesn't allow wrong consent folders to be categorized as `granted` (avoid false positives here).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -218,6 +238,15 @@ internal class BatchMetricsDispatcher(
         /* The value for the type of the metric.*/
         internal const val BATCH_CLOSED_TYPE_VALUE = "batch closed"
 
+        /* The value of the tracking consent according with this file origin.*/
```
jonathanmos, line 241:
```suggestion
        /* The tracking consent according to the file origin.*/
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -218,6 +238,15 @@ internal class BatchMetricsDispatcher(
...
+        internal const val TRACKING_CONSENT_KEY = "consent"
+
+        /* The file name.*/
+        internal const val FILE_NAME = "name"
+
+        /* The thread name from which the current metric was sent.*/
+        internal const val THREAD_NAME = "thread"
```
jonathanmos, line 248: minor: are these key names taken from a telemetry rfc doc? If we have flexibility it may be better to have more explicit map keys (e.g. it looks like we're sending a semiambiguous `name` in the map, where it might be preferable to send `filename`, and to a lesser extent for thread/threadname). wdyt?

jonathanmos (dismissed):

0xnm (approved):
lgtm! I've added one suggestions

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -144,10 +152,22 @@ internal class BatchMetricsDispatcher(
...
+        } else {
+            null
+        }
+    }
+
     // endregion
 
     companion object {
 
+        internal val IS_GRANTED_DIR_REG_EX = Regex("[a-z]+-v[0-9]+")
```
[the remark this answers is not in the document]
mariusc83: actually this is not good enough :)...need to find a better one. We also have `session-replay-v1` or `ndk_crash_reports_intermediary-v1`

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
mariusc83: wanted to use a shorter key for having lighter logs but yeah...3 chars will not kill

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/FeatureFileOrchestrator.kt
```diff
@@ -64,8 +64,13 @@ internal class FeatureFileOrchestrator(
...
 
     companion object {
+        private const val BASE_DIR_NAME_REG_EX = "([a-z]+[-|_])+"
+        internal val IS_GRANTED_DIR_REG_EX = Regex("${BASE_DIR_NAME_REG_EX}v[0-9]+")
+        internal val IS_PENDING_DIR_REG_EX = Regex("${BASE_DIR_NAME_REG_EX}pending-v[0-9]+")
+
         internal const val VERSION = 2
-        internal const val PENDING_DIR = "%s-pending-v$VERSION"
+        internal const val PENDING_DIR_SUFFIX = "-pending-v$VERSION"
+        internal const val PENDING_DIR = "%s$PENDING_DIR_SUFFIX"
```
0xnm, line 73: it seems that this change can be reverted

0xnm (approved):

0xnm (approved):

jonathanmos (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/FeatureFileOrchestrator.kt
```diff
@@ -64,6 +64,10 @@ internal class FeatureFileOrchestrator(
     )
 
     companion object {
+        private const val BASE_DIR_NAME_REG_EX = "([a-z]+[-|_])+"
```
xgouchet, line 67: Do we have a list of base dire name used? It seems this regex is allowing a lot more than it should (e.g.: `abcdef|`)

xgouchet (changes requested):

PR #1645, opened 2023-09-27
Fix RegEx in FeatureFileOrchestrator to resolve file consent type

mariusc83:
### What does this PR do?

The NDK folders do not produce batching telemetry so we had to simplify the rules a bit by removing the `|` separator case.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/FeatureFileOrchestrator.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/FeatureFileOrchestrator.kt
@@ -65,7 +65,7 @@ internal class FeatureFileOrchestrator(
     )
 
     companion object {
-        private const val BASE_DIR_NAME_REG_EX = "([a-z]+[-|_])+"
+        private const val BASE_DIR_NAME_REG_EX = "([a-z]+-)+"
         internal val IS_GRANTED_DIR_REG_EX = Regex("${BASE_DIR_NAME_REG_EX}v[0-9]+")
         internal val IS_PENDING_DIR_REG_EX = Regex("${BASE_DIR_NAME_REG_EX}pending-v[0-9]+")
 
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcherTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcherTest.kt
@@ -608,13 +608,13 @@ internal class BatchMetricsDispatcherTest {
     }
 
     private fun Forge.forgeAGrantedDirName(): String {
-        val separator = if (aBool()) "_" else "-"
+        val separator = "-"
         return aList(anInt(min = 1, max = 10)) { anAlphabeticalString() }
             .joinToString(separator) + "-v" + aNumericalString()
     }
 
     private fun Forge.forgeAPendingDirName(): String {
-        val separator = if (aBool()) "_" else "-"
+        val separator = "-"
         return aList(anInt(min = 1, max = 10)) { anAlphabeticalString() }
             .joinToString(separator) + "-pending-v" + aNumericalString()
     }


0xnm (approved):

jonathanmos (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/FeatureFileOrchestrator.kt
```diff
@@ -64,7 +64,7 @@ internal class FeatureFileOrchestrator(
     )
 
     companion object {
-        private const val BASE_DIR_NAME_REG_EX = "([a-z]+[-|_])+"
+        private const val BASE_DIR_NAME_REG_EX = "([a-z]+(-|_))+"
```
xgouchet, line 67: Do we have a list of all the available names so far? Why are we including the `_` character?

xgouchet (changes requested):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/FeatureFileOrchestrator.kt
mariusc83: the `_` is used in the NDK folders for some reason...I see it was not aligned with the new folders name format. For now all the other features are using the `-` separator. That's the list so far:
```
logs-pending-v2
logs-v2
ndk_crash_reports_intermediary_v2
ndk_crash_reports_v2
rum-pending-v2
rum-v2
session-replay-pending-v2
session-replay-v2
tracing-pending-v2
tracing-v2
```

0xnm: NDK feature doesn't use storage for batches, so we may exclude folders belonging to it, it is outside of the scope for file orchestrator

mariusc83: ok that's great...and the way I see it neither the uploader is using it as the ndk_crash_file is just used to write the equivalent RUM event in their (pending|granted) folder at runtime ? Can you confirm this @0xnm ...this way I could remove that extra rule from there

xgouchet: so it seems you can simplify that regex to `([a-z]+-)+`

0xnm: yes, NDK folders are out of scope of the upload pipeline

mariusc83: @xgouchet yup doing this now after the confirmation ;)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/FeatureFileOrchestrator.kt
```diff
@@ -64,7 +64,7 @@ internal class FeatureFileOrchestrator(
     )
 
     companion object {
-        private const val BASE_DIR_NAME_REG_EX = "([a-z]+[-|_])+"
+        private const val BASE_DIR_NAME_REG_EX = "([a-z]+-)+"
```
0xnm, line 67: I'm not sure if it is simpler, but probably it just can be `[a-z\-]+`. Yes, it allows to have multiple `-` or even only `-`, but maybe evaluation will be simpler/faster. I checked `([a-z]+-)+v[0-9]+` and for test string `session-replay-v2` it captures 2 groups (well, not a bit deal, since overall string still matches).

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcherTest.kt
```diff
@@ -608,13 +608,13 @@ internal class BatchMetricsDispatcherTest {
     }
 
     private fun Forge.forgeAGrantedDirName(): String {
-        val separator = if (aBool()) "_" else "-"
+        val separator = "-"
```
0xnm, line 611: can be just as `joinToString` argument like `joinToString(separator = "-")`. Same below

0xnm (approved):

jonathanmos (approved):

xgouchet (approved):

PR #1647, opened 2023-09-28
RUM-1194 DataUploadWorker is scheduled every time and on non - roaming network

mariusc83:
### What does this PR do?

Previously our data upload worker was scheduled only if the device was offline and on any type of network. We are currently scheduling this all the time and only if the network is unmetered (WIFI or any other form of unmetered network)
We also updated the default `inBackground` state in the `BatchMetricsDispatcher` as `true` to handle correctly the metrics sent from the scheduled upload worker.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/WorkManagerUtils.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/WorkManagerUtils.kt
@@ -44,7 +44,7 @@ internal fun triggerUploadWorker(context: Context, internalLogger: InternalLogge
     try {
         val workManager = WorkManager.getInstance(context)
         val constraints = Constraints.Builder()
-            .setRequiredNetworkType(NetworkType.CONNECTED)
+            .setRequiredNetworkType(NetworkType.NOT_ROAMING)
             .build()
         val uploadWorkRequest = OneTimeWorkRequest.Builder(UploadWorker::class.java)
             .setConstraints(constraints)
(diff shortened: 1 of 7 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/WorkManagerUtils.kt
```diff
@@ -44,7 +45,7 @@ internal fun triggerUploadWorker(context: Context, internalLogger: InternalLogge
     try {
         val workManager = WorkManager.getInstance(context)
         val constraints = Constraints.Builder()
-            .setRequiredNetworkType(NetworkType.CONNECTED)
+            .setRequiredNetworkType(NetworkType.UNMETERED)
```
mariusc83, line 48: should we let it go no matter what the network type is ? I am thinking in situations where the client is on ROAMING and we are sending data under the hood which might not be wise.

jonathanmos: Looks like `WorkManager` has a constraint of [NOT_ROAMING](https://developer.android.com/reference/androidx/work/NetworkType#NOT_ROAMING) - could we use this?

mariusc83: oh good catch...we could indeed use that which is better, we care about ROAMING specifically here

xgouchet (approved):

0xnm (approved):

PR #1648, opened 2023-09-28
feat: Add `isEnabled` to Logs

fuzzybinary:
### What does this PR do?

Added a static method to Logs for checking if the feature had been enabled in a core.  This method is needed to properly setup Flutter when using `attachToExisting`.

refs: RUM-1066

### DIFF
--- a/features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/Logs.kt
+++ b/features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/Logs.kt
@@ -8,6 +8,7 @@ package com.datadog.android.log
 
 import com.datadog.android.Datadog
 import com.datadog.android.api.SdkCore
+import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.log.internal.LogsFeature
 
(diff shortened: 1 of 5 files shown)


xgouchet (approved):

features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/Logs.kt
```diff
@@ -34,4 +35,20 @@ object Logs {
...
+     * This check is useful in scenarios where more than one component may be responsible
+     * for enabling the feature
+     *
+     * @param sdkCore the [SdkCore] instance to check against. If not provided, default instance
+     * will be checked.
+     * @return whether Logs has been enabled
+     */
+    @JvmOverloads
+    @JvmStatic
+    fun isEnabled(sdkCore: SdkCore = Datadog.getInstance()): Boolean {
```
0xnm, line 51: Seems like implementation is different from iOS, where this method is public, but is still in the internal extension? Here we expose it directly, without any proxies.

fuzzybinary: Yes, I can make it internal but I based the decision based on the documentation on `GlobalRumMonitor`. I figured Android might have more of a need for such a method than iOS, but if you think not so much I can make the internal proxy.

0xnm: I'm ok with having it as public, just was curious about the consistency of public API with iOS.

0xnm (approved):

PR #1653, opened 2023-10-03
RUM-1236: Use enum for HTTP method parameter of `RumMonitor#startResource API`

0xnm:
### What does this PR do?

In the [RumMonitor#startResource](https://github.com/DataDog/dd-sdk-android/blob/c6d8de8bc48f3e7dce7bcff41b928cc9d380c187/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt#L116) API we accept HTTP methods as String, however this leads to many errors reported in telemetry from customers who try to supply freeform values for this parameter, because they cannot be matched to the [RUM Schema enum](https://github.com/DataDog/rum-events-format/blob/f69ca4664ed6e69c929855d02c4ce3d4b85d0bb4/schemas/rum/resource-schema.json#L43). It means we also need to use enum, which aligns closer with what iOS has - similar API [doesn’t allow freeform HTTP methods](https://github.com/DataDog/dd-sdk-ios/blob/463a71ff40da79261c45fb5abbe97a96c8980027/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L312) and in case of conversion to RUM schema enum fallback to get is [used](https://github.com/DataDog/dd-sdk-ios/blob/463a71ff40da79261c45fb5abbe97a96c8980027/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L16).

It may be some discrepancy if span and RUM resource reported for the same call, because in case of span freeform `http.method` is allowed and RUM resource expects the value from the enum (so if value cannot be matched, `GET` will be used), but anyway it is the same what iOS SDK has right now, and at least RUM resource will be reported instead of loosing it due to exception.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumEventExt.kt
```diff
@@ -22,31 +23,25 @@ import com.datadog.android.rum.model.ResourceEvent
...
-        ResourceEvent.Method.valueOf(this.uppercase(Locale.US))
-    } catch (e: IllegalArgumentException) {
-        internalLogger.log(
-            InternalLogger.Level.ERROR,
-            listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-            { "Unable to convert [$this] to a valid http method" },
-            e
-        )
-        ResourceEvent.Method.GET
+internal fun RumHttpMethod.toMethod(): ResourceEvent.Method {
```
xgouchet, line 26: > [!NOTE]
> Maybe we should rename this method to `toResourceMethod()` for consistency with the `toErrorMethod()`. below?

instrumented/nightly-tests/src/androidTest/kotlin/com/datadog/android/nightly/utils/ForgeExtensions.kt
```diff
@@ -52,8 +53,8 @@ fun Forge.anErrorMessage(prefix: String = RUM_ERROR_MESSAGE_PREFIX): String {
     return prefix + this.aStringMatching("[a-zA-z](.+)")
 }
 
-fun Forge.aResourceMethod(): String {
-    return this.anElementFrom(listOf("GET", "POST"))
+fun Forge.aResourceMethod(): RumHttpMethod {
+    return this.anElementFrom(RumHttpMethod.GET, RumHttpMethod.POST)
```
xgouchet, line 57: > [!NOTE]
> Shouldn't we use all possible values of the enum? My guess is that as a string we only used the most popular values (and maybe we should have added more), but now that it's an enum, we can use 
```
return this.aValueFrom(RumHttpMethod::class.java)
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -354,6 +355,25 @@ internal constructor(
         }
     }
 
+    private fun toHttpMethod(method: String, internalLogger: InternalLogger): RumHttpMethod {
+        return when (method.uppercase(Locale.US)) {
```
xgouchet, line 359: > [!IMPORTANT]
> The official [HTTP rfc](https://www.rfc-editor.org/rfc/rfc9110.html#section-9) mentions three additional known methods: `CONNECT`, `OPTIONS` and `TRACE`. 
> Although they are rarely used in mobile apps, maybe it's worth adding them as supported values (but this means also adding them to the json schema). Some customers are using the `OTPIONS` value (cf [telemetry](https://dd-rum-telemetry.datadoghq.com/logs?query=status%3Aerror%20service%3Add-sdk-android%20%22valid%20http%20method%22%20&agg_q=status%2Cservice&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=sliding&sort_m=%2C&sort_t=%2C&stream_sort=desc&top_n=10%2C10&top_o=top%2Ctop&viz=pattern&x_missing=true%2Ctrue&from_ts=1695721329642&to_ts=1696326129642&live=true))

> [!NOTE]
> On a side note, there's also a much [wider list of method extensions](https://www.iana.org/assignments/http-methods/http-methods.xhtml), which should be considered "part of HTTP, as long as they’re listed in the appropriate registries". It might be overkill to add those as well, though, and I can't find any of those being reported in the telemetry org.

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorNotSendingSpanTest.kt
```diff
@@ -1324,19 +1325,23 @@ internal open class TracingInterceptorNotSendingSpanTest {
...
 
-    private fun forgeRequest(
+    protected fun forgeRequest(
         forge: Forge,
         configure: (Request.Builder) -> Unit = {}
     ): Request {
         val builder = Request.Builder().url(fakeUrl)
         if (forge.aBool()) {
-            fakeMethod = "POST"
+            fakeMethod = RumHttpMethod.POST
```
xgouchet, line 1334: > [!IMPORTANT]
Same as above, we might want to fake all possible kinds of requests, so maybe have either `POST`, `PATCH` or `PUT` here

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/DatadogInterceptorWithoutTracesTest.kt
```diff
@@ -329,13 +378,17 @@ internal class DatadogInterceptorWithoutTracesTest {
         fakeUrl = "$protocol://$host/$path"
         val builder = Request.Builder().url(fakeUrl)
         if (forge.aBool()) {
-            fakeMethod = "POST"
+            fakeMethod = RumHttpMethod.POST
```
xgouchet, line 381: > [!IMPORTANT]
> We might want to fake all possible kinds of requests, so maybe have either `POST`, `PATCH` or `PUT` here

xgouchet (changes requested):

xgouchet (approved):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
0xnm: Completely agree. For now I would propose to keep things as they are, since the goal of this PR is just to change the API and we can deal with RUM schema in the future.

xgouchet (approved):
Good work indeed, this will limit us spamming the telemetry and can help guide the customers in using the proper values

mariusc83 (approved):

PR #1659, opened 2023-10-04
RUM-891 Unit test to confirm SR records order is kept when having same timestamps

mariusc83:
### What does this PR do?

We have situations where we have to send records with same timestamp (dummy records) when the view is changed to comply with the browser sdk - player way of working (FocusRecord, MetaRecord, ViewPortRecord) and we wanted to make sure that whenever they are read back from the batch to be sent the order is kept.

### Motivation

### Additional Notes


features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/net/BatchesToSegmentMapperTest.kt
0xnm, line 1: minor: this file should be called `BatchesToSegmentsMapperTest` instead, because class under test is called `BatchesToSegmentsMapper`

0xnm (approved):

0xnm (approved):

PR #1661, opened 2023-10-04
Update testing conventions

xgouchet:
### What does this PR do?

Follow up after today's meeting to improve our testing guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,24 +22,37 @@ In addition, to be able to run the static analysis tools locally, you should run
 This project hosts the following modules:
 
   - `dd-sdk-android-core`: the main library implementing the core functionality of SDK (storage and upload of data, core APIs);
-  - `features/***`: a set of libraries implementing Datadog products: Logs, RUM, Traces, etc.
-  - `integrations/dd-sdk-android-coil`: a lightweight library providing a bridge integration between Datadog SDK and [Coil](https://coil-kt.github.io/coil/);
-  - `integrations/dd-sdk-android-compose`: a lightweight library providing a bridge integration between Datadog SDK and [Jetpack Compose](https://developer.android.com/jetpack/compose);
-  - `integrations/dd-sdk-android-fresco`: a lightweight library providing a bridge integration between Datadog SDK and [Fresco](https://frescolib.org/);
-  - `integrations/dd-sdk-android-okhttp`: a lightweight library providing an instrumentation for [OkHttp](https://square.github.io/okhttp/);
-  - `integrations/dd-sdk-android-rx`: a lightweight library providing a bridge integration between Datadog SDK and [RxJava](https://github.com/ReactiveX/RxJava);
-  - `integrations/dd-sdk-android-sqldelight`: a lightweight library providing a bridge integration between Datadog SDK and [SQLDelight](https://cashapp.github.io/sqldelight/);
-  - `integrations/dd-sdk-android-tv`: a lightweight library providing extensions for [Android TV](https://www.android.com/tv/)
-  - `integrations/dd-sdk-android-ktx`: a set of Kotlin extensions to make the Datadog SDK more Kotlin friendly;
-  - `integrations/dd-sdk-android-glide`: a lightweight library providing a bridge integration between Datadog SDK and [Glide](https://bumptech.github.io/glide/);
-  - `integrations/dd-sdk-android-timber`: a lightweight library providing a bridge integration between Datadog SDK and [Timber](https://github.com/JakeWharton/timber);
-  - `instrumented/integration`: a test module with integration tests using Espresso;
-  - `instrumented/nightly-tests`: a test module with E2E tests using Espresso;
-  - `tools/detekt`: a few custom [Detekt](https://github.com/arturbosch/detekt) static analysis rules;
-  - `tools/lint`: a custom [Lint](https://developer.android.com/studio/write/lint) static analysis rule;
-  - `tools/noopfactory`: an annotation processor generating no-op implementation of interfaces;
-  - `tools/unit`: a utility library with code to help writing unit tests;
+  - `features/***`: a set of libraries implementing Datadog products: 
+    - `features/dd-sdk-android-logs`: a library to send logs to Datadog;
+    - `features/dd-sdk-android-rum`: a library to track user navigation and interaction;
+    - `features/dd-sdk-android-ndk`: a lightweight library to track crashes from NDK libraries;
+    - `features/dd-sdk-android-session-replay`: a library to capture the window content;
+    - `features/dd-sdk-android-session-replay-material`: an extension for Session Replay to integrate with the Material Design library;
+    - `features/dd-sdk-android-session-trace`: a library to measure performance of operations locally;
+    - `features/dd-sdk-android-session-webview`: a library to forward logs and RUM events captured in a webview to be linked with the mobile session;
+  - `integrations/***`: a set of libraries integrating Datadog products in third party libraries:
+    - `integrations/dd-sdk-android-coil`: a lightweight library providing a bridge integration between Datadog SDK and [Coil](https://coil-kt.github.io/coil/);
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,19 +22,27 @@ In addition, to be able to run the static analysis tools locally, you should run
...
+    - `integrations/dd-sdk-android-rx`: a lightweight library providing a bridge integration between Datadog SDK and [RxJava](https://github.com/ReactiveX/RxJava);
+    - `integrations/dd-sdk-android-sqldelight`: a lightweight library providing a bridge integration between Datadog SDK and [SQLDelight](https://cashapp.github.io/sqldelight/);
+    - `integrations/dd-sdk-android-tv`: a lightweight library providing extensions for [Android TV](https://www.android.com/tv/)
+    - `integrations/dd-sdk-android-ktx`: a set of Kotlin extensions to make the Datadog SDK more Kotlin friendly;
+    - `integrations/dd-sdk-android-glide`: a lightweight library providing a bridge integration between Datadog SDK and [Glide](https://bumptech.github.io/glide/);
+    - `integrations/dd-sdk-android-timber`: a lightweight library providing a bridge integration between Datadog SDK and [Timber](https://github.com/JakeWharton/timber);
+    - `instrumented/integration`: a test module with integration tests using Espresso;
+    - `instrumented/nightly-tests`: a test module with E2E tests using Espresso;
```
0xnm, line 45: these two shouldn't be a part of `integrations` group

```suggestion
  - `instrumented/integration`: a test module with integration tests using Espresso;
  - `instrumented/nightly-tests`: a test module with E2E tests using Espresso;
```

CONTRIBUTING.md
```diff
@@ -22,19 +22,27 @@ In addition, to be able to run the static analysis tools locally, you should run
...
-  - `integrations/dd-sdk-android-glide`: a lightweight library providing a bridge integration between Datadog SDK and [Glide](https://bumptech.github.io/glide/);
-  - `integrations/dd-sdk-android-timber`: a lightweight library providing a bridge integration between Datadog SDK and [Timber](https://github.com/JakeWharton/timber);
-  - `instrumented/integration`: a test module with integration tests using Espresso;
-  - `instrumented/nightly-tests`: a test module with E2E tests using Espresso;
+  - `features/***`: a set of libraries implementing Datadog products: 
+    - `features/dd-sdk-android-logs`: a library to send logs to Datadog;
+    - `features/dd-sdk-android-rum`: a library to track user navigation and interaction;
+    - `features/dd-sdk-android-ndk`: a lightweight library to track crashes from NDK libraries;
+    - `features/dd-sdk-android-session-replay`: a library to capture the window content;
+    - `features/dd-sdk-android-session-replay-material`: ;
```
0xnm, line 30: missing description

CONTRIBUTING.md
```diff
@@ -233,38 +247,61 @@ In order to make the test classes more readable, here are a set of naming conven
...
+#### Fields & Test Method parameters
 
 Fields should appear in the following order, and be named as explained by these rules:
 
-- The object(s) under test must be named from their class, and prefixed by `tested`. E.g.: `testedListener: Listener`, `testedVisitor: KotlinFileVisitor`.
-- Mocked objects must be named from their class (with an optional qualifier), and prefixed by `mock`. E.g.: `mockListener: Listener`, `mockService: ExecutorService`).
-- Fake data must be named from their class (with an optional qualifier), and prefixed by `fake`. E.g.: `fakeContext: Context`, `fakeApplicationId: UUID`, `fakeRequest: NetworkRequest`.
+- The object(s) under test must be named from their class, and prefixed by `tested`. 
+    E.g.: `testedListener: Listener`, `testedHandler: Handler`.
+- Stubed objects (mocks with predefined behavior) must be named from their class (with an optional qualifier), and prefixed by `stub`.
```
0xnm, line 267: not sure here, but I guess the correct spelling should be `stubbed`

```suggestion
- Stubbed objects (mocks with predefined behavior) must be named from their class (with an optional qualifier), and prefixed by `stub`.
```

CONTRIBUTING.md
```diff
@@ -233,38 +247,61 @@ In order to make the test classes more readable, here are a set of naming conven
...
+Parameters shall have simple local names reflecting their intent (see above), whether they use an `@Forgery` or `@Mock` annotation (or none).
 
 Here's a test method following those conventions: 
 
 ```kotlin
     @Test
-    fun `𝕄 forward boolean attribute to handler 𝕎 addAttribute()`(
-        @StringForgery(StringForgeryType.ALPHABETICAL) key : String,
-        @BoolForgery value : Boolean
+    fun `M forward boolean attribute to handler W addAttribute()`(
```
0xnm, line 296: we don't use unicode variants for `M` and `W` anymore?

CONTRIBUTING.md
```diff
@@ -281,11 +318,11 @@ Here's a test method following those conventions:
...
 ```
 
-### Test Utility Methods
+#### Test Utility Methods
 
-Because we sometimes need to reuse some setup or assertions in our tests, we tend to write utility methods. Those methods should be private (or internal in a dedicated class/file if they need to be shared across tests).
+Because we sometimes need to reuse some setup or assertions in our tests, we tend to write utility methods. 
+Those methods should be private (or internal in a dedicated class/file if they need to be shared across tests).
 
-- `fun mockSomething([args]): T`: methods setting up a mock. These methods must return the mocked instance;
```
0xnm, line 288: why do we remove this line?

0xnm (commented):
looks good! I've added some questions/suggestions

CONTRIBUTING.md
xgouchet: A lot of tests aren't using them (they're not simple enough to type), and they do raise a bunch of warning in AS that add noise to the editor. It's still readable without the variants, so let's make it simpler to use the convention

CONTRIBUTING.md
xgouchet: As per our discussions, mocks are only listeners that shouldn't need any configuration. Only stubs should require configuration, so this line isn't relevant anymore

CONTRIBUTING.md
0xnm: They are also causing the following Kotlin compiler bug I've opened in the past https://youtrack.jetbrains.com/issue/KT-60748. Although it can be solved.

jonathanmos (approved):

CONTRIBUTING.md
```diff
@@ -233,38 +252,61 @@ In order to make the test classes more readable, here are a set of naming conven
...
 
-- The object(s) under test must be named from their class, and prefixed by `tested`. E.g.: `testedListener: Listener`, `testedVisitor: KotlinFileVisitor`.
-- Mocked objects must be named from their class (with an optional qualifier), and prefixed by `mock`. E.g.: `mockListener: Listener`, `mockService: ExecutorService`).
-- Fake data must be named from their class (with an optional qualifier), and prefixed by `fake`. E.g.: `fakeContext: Context`, `fakeApplicationId: UUID`, `fakeRequest: NetworkRequest`.
+- The object(s) under test must be named from their class, and prefixed by `tested`. 
+    E.g.: `testedListener: Listener`, `testedHandler: Handler`.
+- Stubbed objects (mocks with predefined behavior) must be named from their class (with an optional qualifier), and prefixed by `stub`.
+    E.g.: `stubDataProvider: DataProvider`, `stubReader: Reader`.
+- Mocked objects (mocks being verified) must be named from their class (with an optional qualifier), and prefixed by `mock`. 
+    E.g.: `mockListener: Listener`, `mockLogger: Logger`).
```
0xnm, line 275: extra bracket

```suggestion
    E.g.: `mockListener: Listener`, `mockLogger: Logger`.
```

0xnm (approved):
minor suggestion only, otherwise lgtm

0xnm (dismissed):

0xnm (approved):

PR #1663, opened 2023-10-09
RUM-501 Handle Android Strict Mode

xgouchet:
### What does this PR do?

This tries to handle some StrictMode warnings happening during the SDK initialization, which can't easily be deferred to background threads. 

### Motivation

Prevent the SDK from spamming warnings in host application when StrictMode feature is enabled.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StrictModeExt.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StrictModeExt.kt
@@ -1,0 +1,30 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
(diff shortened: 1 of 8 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StrictModeExt.kt
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * This utility function wraps a call to a method that needs to perform a disk read operation
+ * on the main thread.
+ * This prevents adding LogCat noise when customer enable StrictMode logging.
+ * @param T the type returned by the operation
+ * @param operation the operation
+ * @return the value returned by the operation
+ */
+fun <T> allowThreadDiskReads(
```
0xnm, line 19: Let's mark it as `InternalApi`.

0xnm (approved):

mariusc83 (dismissed):

0xnm (approved):

PR #1665, opened 2023-10-10
RUM-1330 Make sure we use try-locks in our NDK signal catcher

mariusc83:
### What does this PR do?

We recently had some reported issues in the flutter sdk regarding some ANRs caused by our signal catcher. Basically what was happening is that our code inside the signal catcher was trying to acquire a `mutex` lock which was already acquired by another thread (we could not really reproduce this). We chose to go with the safest mechanism here in order to avoid any other issues and use `try locks` instead of simple `locks`.

https://github.com/DataDog/dd-sdk-flutter/issues/487

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
+++ b/features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
@@ -53,10 +53,8 @@ class NdkTests {
 
     @Test
     fun mustWriteAnErrorLog_whenHandlingSignal_whenConsentUpdatedToGranted() {
-        val fakeSignal = forge.anInt(min = 1, max = 32)
-        // we need to keep this this size because we are using a buffer of [30] size in c++ for
-        // the error.signal attribute
-        val fakeSignalName = forge.anAlphabeticalString(size = 20)
+        val fakeSignal = forge.aPositiveInt(true)
+        val fakeSignalName = forge.anAlphabeticalString()
         val fakeErrorMessage = forge.anAlphabeticalString()
         val fakeErrorStack = forge.anAlphabeticalString()
         initNdkErrorHandler(temporaryFolder.root.absolutePath)
@@ -95,25 +93,23 @@ class NdkTests {
 
     @Test
     fun mustNotWriteAnyLog_whenHandlingSignal_whenConsentUpdatedToPending() {
-        val signal = forge.anInt(min = 1, max = 32)
-        // we need to keep this this size because we are using a buffer of [30] size in c++ for
-        // the error.signal attribute
-        val signalName = forge.anAlphabeticalString(size = 20)
-        val errorMessage = forge.anAlphabeticalString()
-        val errorStack = forge.anAlphabeticalString()
+        val fakeSignal = forge.aPositiveInt(true)
+        val fakeSignalName = forge.anAlphabeticalString()
+        val fakeErrorMessage = forge.anAlphabeticalString()
+        val fakeErrorStack = forge.anAlphabeticalString()
         initNdkErrorHandler(temporaryFolder.root.absolutePath)
         updateTrackingConsent(0)
         simulateSignalInterception(
-            signal,
-            signalName,
-            errorMessage,
-            errorStack
+            fakeSignal,
+            fakeSignalName,
+            fakeErrorMessage,
+            fakeErrorStack
         )
 
         // we need to give time to native part to write the file
         // otherwise we will get into race condition issues
         ConditionWatcher {
-            // assert the log file
+            // assert the log file is not written
             Assertions.assertThat(temporaryFolder.root.listFiles()).isEmpty()
             true
         }.doWait(timeoutMs = 5000)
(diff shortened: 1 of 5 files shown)


features/dd-sdk-android-ndk/src/test/cpp/integration-tests.cpp
```diff
@@ -1,11 +1,19 @@
...
+#include <thread>
+
 
 #include <android/log.h>
 #include <datadog-native-lib.h>
 #include "greatest/greatest.h"
 #include "utils/string-utils.h"
 
+
+#ifndef NDEBUG
```
mariusc83, line 13: I actually don't think we need this directive check here but just in case

features/dd-sdk-android-ndk/src/main/cpp/datadog-native-lib.cpp
```diff
@@ -29,7 +29,13 @@ static struct Context {
 
 
 static const char *LOG_TAG = "DatadogNdkCrashReporter";
+
+#ifndef NDEBUG
```
fuzzybinary, line 33: I feel like this is a lot of changes just to support the test. Maybe it makes sense to leave the mutex as `static` and instead create an `NDEBUG` wrapped function just to lock it.

features/dd-sdk-android-ndk/src/main/cpp/datadog-native-lib.cpp
```diff
@@ -43,9 +49,10 @@ void write_crash_report(int signum,
     static const std::string crash_log_filename = "crash_log";
 
     // sync everything
-    pthread_mutex_lock(&handler_mutex);
     if (tracking_consent != tracking_consent_granted) {
```
fuzzybinary, line 52: Is there any reason we can't move this check up before the mutex lock? I wouldn't think we'd care about a race condition here.

features/dd-sdk-android-ndk/src/test/cpp/integration-tests.cpp
fuzzybinary: We probably do, in fact the whole test would probably need to be `NDEBUG` specific

features/dd-sdk-android-ndk/src/main/cpp/datadog-native-lib.cpp
```diff
@@ -43,9 +49,10 @@ void write_crash_report(int signum,
...
-        pthread_mutex_unlock(&handler_mutex);
+        return;
+    }
+    if(pthread_mutex_trylock(&handler_mutex)!=0){
```
0xnm, line 55:
```suggestion
    if(pthread_mutex_trylock(&handler_mutex) != 0){
```

features/dd-sdk-android-ndk/src/main/cpp/datadog-native-lib.cpp
```diff
@@ -87,7 +94,9 @@ void write_crash_report(int signum,
 void update_main_context(JNIEnv *env,
                          jstring storage_path) {
     using namespace stringutils;
-    pthread_mutex_lock(&handler_mutex);
+    if(pthread_mutex_trylock(&handler_mutex)!=0){
```
0xnm, line 97:
```suggestion
    if(pthread_mutex_trylock(&handler_mutex) != 0){
```

features/dd-sdk-android-ndk/src/main/cpp/utils/signal-monitor.c
```diff
@@ -93,7 +93,9 @@ void free_up_memory() {
 void invoke_previous_handler(int signum, siginfo_t *info, void *user_context) {
     // It may happen that the process is killed during this function execution.
     // Therefore this function may never return.
-    pthread_mutex_lock(&mutex);
+    if(pthread_mutex_trylock(&mutex)!=0){
```
0xnm, line 96:
```suggestion
    if(pthread_mutex_trylock(&mutex) != 0){
```

0xnm (approved):

features/dd-sdk-android-ndk/src/main/cpp/datadog-native-lib.cpp
mariusc83: I usually apply ktlintFormat at the end of my PR but of course for C code doesn't work :)

features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
```diff
@@ -54,7 +54,7 @@ class NdkTests {
     @Test
     fun mustWriteAnErrorLog_whenHandlingSignal_whenConsentUpdatedToGranted() {
         val fakeSignal = forge.anInt(min = 1, max = 32)
-        // we need to keep this this size because we are using a buffer of [30] size in c++ for
+        // we need to keep this size because we are using a buffer of [30] size in c++ for
```
xgouchet, line 57: > [!NOTE] 
> You're talking about the `(size = 20)` below right? Wouldn't it be better to have a constant `const val SIGNAL_SIZE = 20` (declared in the `companion object` for example) and reuse it everywhere?
> Or because we know the size can be at most 30, user something like `val signalSize = forge.anInt(1, 30)`?

features/dd-sdk-android-ndk/src/main/cpp/datadog-native-lib.cpp
xgouchet: > [!WARNING]
> According to the [documentation](https://linux.die.net/man/3/pthread_mutex_trylock) for `pthread_mutex_trylock`, there might be several return value. Maybe it'd be relevant to actually look at the return value and act accordingly.
> 
> > If successful, the pthread_mutex_lock() and pthread_mutex_unlock() functions shall return zero; otherwise, an error number shall be returned to indicate the error.

xgouchet (changes requested):

mariusc83:
@xgouchet cannot reply on those comments for some reason so I will reply here, I've read the docs before adding this logic and actually we do not want to do anything in case we cannot acquire the lock, we just do not write the error log as there is no other action that we can take in that case. We could not find out who's acquiring that lock in the tests we performed (me and @fuzzybinary) and we do not want to block the process in the signal catcher in this case.

xgouchet:
@mariusc83 in that case, then maybe add a comment explaining that 👍

features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
mariusc83: I actually need to check this but I think this is not needed anymore and I need to just forge a positive int there as the serialization was changed

features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
```diff
@@ -95,45 +93,65 @@ class NdkTests {
...
     @Test
     fun mustNotWriteAnyLog_whenHandlingSignal_whenConsentUpdatedToPending() {
-        val signal = forge.anInt(min = 1, max = 32)
-        // we need to keep this this size because we are using a buffer of [30] size in c++ for
-        // the error.signal attribute
-        val signalName = forge.anAlphabeticalString(size = 20)
-        val errorMessage = forge.anAlphabeticalString()
-        val errorStack = forge.anAlphabeticalString()
+        val fakeSignal = forge.aPositiveInt(true)
+        val fakeSignalName = forge.anAlphabeticalString(size = 20)
```
xgouchet, line 97: > [!NOTE]
> Can you use a more random thing than 20, or at least use a constant explaining the 30 characters constraint

mariusc83: We actually do not need these limitations anymore as this was needed before when that part of code was written in pure C and we used fixed length memory allocation for that. Currently that code is C++ and allows dynamic allocation.

xgouchet (approved):

fuzzybinary (approved):

PR #1666, opened 2023-10-11
(1/3) RUM-392 create core testFixtures source set

xgouchet:
### What does this PR do?

This PR adds a `testFixtures` source set to the Core module, and adds test fixture classes that will be shared with other modues (in the next PR)

### Motivation

This is part of our effort on improving our test suites across the mobile teams

### DIFF
--- a/dd-sdk-android-core/build.gradle.kts
+++ b/dd-sdk-android-core/build.gradle.kts
@@ -170,9 +170,11 @@ dependencies {
     testImplementation(libs.bundles.testTools)
     unmock(libs.robolectric)
 
-    // Static Analysis
-    // TODO MTG-12 detekt(project(":tools:detekt"))
-    // TODO MTG-12 detekt(libs.detektCli)
+    // Test Fixtures
+    testFixturesImplementation(libs.kotlin)
+    testFixturesImplementation(libs.bundles.jUnit5)
+    testFixturesImplementation(libs.okHttp)
+    testFixturesImplementation(libs.bundles.testTools)
 }
 
 unMock {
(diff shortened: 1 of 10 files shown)


dd-sdk-android-core/build.gradle.kts
```diff
@@ -170,9 +170,11 @@ dependencies {
     testImplementation(libs.bundles.testTools)
     unmock(libs.robolectric)
 
-    // Static Analysis
-    // TODO MTG-12 detekt(project(":tools:detekt"))
-    // TODO MTG-12 detekt(libs.detektCli)
+    // Test Fixtures
+    testFixturesImplementation(libs.kotlin)
```
0xnm, line 174: why it is needed? `implementation` dependencies don't cover `testFixtures` source set (like they cover `test` source set)?

dd-sdk-android-core/build.gradle.kts
```diff
@@ -170,9 +170,11 @@ dependencies {
...
+    testFixturesImplementation(libs.bundles.jUnit5)
+    testFixturesImplementation(libs.okHttp)
```
0xnm, line 176: it seems no class in `testFixtures` source set is using it, why do we need it?

0xnm (approved):
just two minor questions, otherwise lgtm.

mariusc83 (commented):
Nice ;)

mariusc83 (approved):

dd-sdk-android-core/build.gradle.kts
xgouchet: For Gradle, no it's not necessary but AndroidStudio needed that to be happy.

dd-sdk-android-core/build.gradle.kts
xgouchet: it's needed for something that's going to be added next but not in this PR yet

PR #1667, opened 2023-10-11
(2/3) RUM-392 refactor shared android library build script

xgouchet:
### What does this PR do?

This PR refactor the `build.gradle.kts` buildscript for all our libraries to extract shared configurations and make it easier to: 
- create new libraries
- add build features to all libraries

### Motivation

This is part of our effort on improving our test suites across the mobile teams

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
@@ -7,9 +7,10 @@
 package com.datadog.gradle.config
 
 import com.android.build.api.dsl.CompileOptions
-import com.android.build.api.dsl.LibraryDefaultConfig
+import com.android.build.gradle.LibraryExtension
 import com.datadog.gradle.utils.Version
 import org.gradle.api.JavaVersion
+import org.gradle.api.Project
 
 object AndroidConfig {
 
@@ -21,7 +22,7 @@ object AndroidConfig {
     val VERSION = Version(2, 3, 0, Version.Type.Snapshot)
 }
 
-// TODO RUMM-3263 Switch to Java 17 bytecode
+// TODO RUM-628 Switch to Java 17 bytecode
 fun CompileOptions.java11() {
     sourceCompatibility = JavaVersion.VERSION_11
     targetCompatibility = JavaVersion.VERSION_11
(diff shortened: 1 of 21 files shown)


buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
0xnm, line 1: In the future probably it will be better to switch from `buildSrc` to the composite build usage (like mobile app is doing). I tried to do a quick migration, but then the problem is that some extension symbols are not available and the only way to make things work is to re-write everything to plugins, which is quite a work.

buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
```diff
@@ -32,11 +33,60 @@ fun CompileOptions.java17() {
...
-    buildConfigField("String", "SDK_VERSION_NAME", "\"$versionName\"")
+fun Project.androidLibraryConfig() {
+    extensionConfig<LibraryExtension> {
+        compileSdk = AndroidConfig.TARGET_SDK
+        buildToolsVersion = AndroidConfig.BUILD_TOOLS_VERSION
+
+        defaultConfig {
+            minSdk = AndroidConfig.MIN_SDK
+            buildFeatures {
+                buildConfig = true
```
0xnm, line 44: in fact we need it only in the core module. Not sure how it affects compilation speed by enabling it for all modules.

features/dd-sdk-android-logs/api/dd-sdk-android-logs.api
```diff
@@ -1,3 +1,10 @@
+public final class com/datadog/android/log/BuildConfig {
```
0xnm, line 1: probably we don't need this to be generated for most of the library modules.

integrations/dd-sdk-android-sqldelight/api/dd-sdk-android-sqldelight.api
```diff
@@ -1,3 +1,10 @@
+public final class com/datadog/android/sqldelight/BuildConfig {
```
mariusc83, line 1: where is this class being generated if it was picked up by apiSurface ?

buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
xgouchet: That might be an idea, we need to see the pros and cons of that. A quick read on the composite build docs left a lot of questions, so I think this might be better addressed as an exploration/ADR.

buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
xgouchet: Yep, changing that

features/dd-sdk-android-logs/api/dd-sdk-android-logs.api
xgouchet: True

integrations/dd-sdk-android-sqldelight/api/dd-sdk-android-sqldelight.api
xgouchet: This was comming from the snippet in the shared `androidLibraryConfig()` I created. 
```
        buildFeatures {
            buildConfig = true
        }
```

I removed this to cleanup this change

mariusc83 (approved):

0xnm (approved):

integrations/dd-sdk-android-rx/api/apiSurface
0xnm, line 1: this file should be restored

integrations/dd-sdk-android-sqldelight/api/apiSurface
0xnm, line 1: this one should be restored as well

integrations/dd-sdk-android-timber/api/apiSurface
0xnm, line 1: this file should be restored

0xnm (changes requested):
it seems few `apiSurface` files should are deleted by some occasion, they should be restored.

0xnm (approved):

PR #1678, opened 2023-10-19
RUM-321: Introduce view event filtering in upload pipeline, remove view event throttling in write pipeline

0xnm:
### What does this PR do?

This is the counterpart of https://github.com/DataDog/dd-sdk-ios/pull/1328 for Android.

This PR filters out RUM View events during the upload process following the logic that batch should have only one RUM view event for a given ID and this event should have maximum `documentVersion` property value. To support that we introduce RUM View event metadata which will be attached to every view event written and read back during the upload cycle.

View throttling logic in `RumViewScope` is removed.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMeta.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    open fun toJson(): JsonObject {
+        val model = JsonObject()
+
+        model.addProperty(TYPE_KEY, type)
+
+        return model
+    }
+
+    companion object {
```
xgouchet, line 25: > [!NOTE]
> Nitpicking: the `companion object` is usually at the very end of it's class, it feels weird to have other declarations after it.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMetaDeserializer.kt
```diff
@@ -0,0 +1,60 @@
...
+
+internal class RumEventMetaDeserializer(
+    private val internalLogger: InternalLogger
+) : Deserializer<ByteArray, Any> {
+    override fun deserialize(model: ByteArray): Any? {
+        if (model.isEmpty()) return null
+
+        return try {
+            RumEventMeta.fromJson(model.toJson())
+        } catch (@Suppress("TooGenericExceptionCaught") e: NullPointerException) {
```
xgouchet, line 23: > [!NOTE]
> What would be the cause of an NPE in the `fromJson`? We don't throw it ourselves, is this something that happened during your tests? If so, maybe we can prevent it from throwing instead of catching it?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMeta.kt
```diff
@@ -0,0 +1,77 @@
...
+        const val TYPE_KEY = "type"
+        const val VIEW_TYPE_VALUE = "view"
+        const val VIEW_ID_KEY = "viewId"
+        const val DOCUMENT_VERSION_KEY = "documentVersion"
+
+        @Throws(
+            NullPointerException::class,
+            ClassCastException::class,
+            IllegalStateException::class,
+            NumberFormatException::class
```
xgouchet, line 38: > [!NOTE]
> It might be relevant to add as a comment what situation those exception could rise from?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMeta.kt
```diff
@@ -0,0 +1,77 @@
...
+                }
+
+                else ->
+                    @Suppress("ThrowingInternalException")
+                    throw IllegalStateException(
+                        UNKNOWN_RUM_EVENT_META_TYPE_ERROR.format(
+                            Locale.US,
+                            type
+                        )
+                    )
```
xgouchet, line 56: > [!IMPORTANT]
> Instead of throwing an exception that we don't know _for sure_ is going to be caught, maybe we could let `fromJson` return a nullable `RumEventMeta?`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMeta.kt
```diff
@@ -0,0 +1,77 @@
...
+            IllegalStateException::class,
+            NumberFormatException::class
+        )
+        fun fromJson(model: JsonObject): RumEventMeta {
+            return when (val type = model.get(TYPE_KEY).asString) {
+                VIEW_TYPE_VALUE -> {
+                    val viewId = model.get(VIEW_ID_KEY).asString
+                    val docVersion = model.get(DOCUMENT_VERSION_KEY).asLong
+
+                    View(viewId, docVersion)
```
xgouchet, line 46: > [!NOTE]
> It might be more readable (especially when we have more metadata types) to have a `RumEventMeta.View.fromJson(JsonObject)` companion method instead? That way the `RumEventMeta` companion doesn't need to know about the `View` structure?

xgouchet (changes requested):

mariusc83 (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMetaSerializer.kt
```diff
@@ -0,0 +1,19 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.domain.event
+
+import com.datadog.android.core.persistence.Serializer
+import com.google.gson.JsonObject
+
+internal class RumEventMetaSerializer : Serializer<Any> {
```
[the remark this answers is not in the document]
0xnm: Good catch, it should be `Serializer<RumEventMeta>` indeed.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMetaDeserializer.kt
0xnm: It comes from Java interop with `gson` library, because `get` call can return `null`.

Basically the list of exceptions caught is the same as we use in RUM events parsing.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMeta.kt
0xnm: I changed the way we are doing the parsing, so that it looks now the same as for other RUM events and the only exception thrown is `JsonParseException`. Let me know what you think.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMeta.kt
0xnm: I would rather keep it in the companion of `RumEventMeta`, because if we add more meta types in the future we then leak responsibility - external caller should decide companion method of which class exactly to call, and to know that it needs to parse meta into JSON to know exact metadata type, so it means we are doing JSON parsing for the same meta binary in 2 places.

xgouchet (approved):

PR #1681, opened 2023-10-26
RUM-1519 Api coverage

xgouchet:
### What does this PR do?

This PR attemps to get an "API coverage" for our integration tests suites: 

- Using Detekt (and it's type resolution feature), we can generate an apiSurface and apiUsage log files. The API Surface is generated from the public class/functions of the public libraries, and API Usage is generated from the test classes in the integration test suites
- Using a script we can count the number of lines in API Surface also present in API Usage and get a coverage number
- For now the coverage will be reported as a warning, but eventually it will become an error if not 100%

### Motivation

As part of our updated test pyramid definition, we need to have a reliable coverage reporting for integration tests


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -275,7 +275,7 @@ internal class BatchFileOrchestrator(
 
     private fun freeSpaceIfNeeded() {
         val files = listSortedBatchFiles()
-        val sizeOnDisk = files.sumOf { it.lengthSafe(internalLogger) }
+        val sizeOnDisk = files.sumOf { f: File -> f.lengthSafe(internalLogger) }
```
0xnm, line 278: is explicit type needed?

0xnm (approved):
great work!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
xgouchet: I think it helped it resolving a type at some point

PR #1684, opened 2023-10-31
RUM-1613 Add mapper interface for traversing all children

jonathanmos:
### What does this PR do?
Add an internal interface for mappers that need to traverse all children during tree traversal.

### Motivation
This interface is to be used by React Native for traversing RN view groups, as preparation for implementing Session Replay there.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
@@ -11,6 +11,7 @@ import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueRefs
 import com.datadog.android.sessionreplay.internal.recorder.mapper.DecorViewMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.MapperTypeWrapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.QueueableViewMapper
+import com.datadog.android.sessionreplay.internal.recorder.mapper.TraverseAllChildrenMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.ViewWireframeMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.WireframeMapper
 import com.datadog.android.sessionreplay.model.MobileSegment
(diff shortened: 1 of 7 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/WireframeMapper.kt
```diff
@@ -20,6 +20,10 @@ import com.datadog.android.sessionreplay.model.MobileSegment
  * Session Replay representation for a specific View type you can implement this on your end.
  */
 interface WireframeMapper<in T : View, out S : MobileSegment.Wireframe> {
+    /**
+     * Determines the behavior during the traversal of the view tree.
+     * @see TraversalStrategy
+     */
     val traversalStrategy: TraversalStrategy
```
mariusc83, line 27: I think last time I discussed with @louiszawadzki I mentioned that this should not be part of the `WireframeMapper` interface. Instead we could make it part only of the `NonExhaustiveTypeMapper` and check in the traversal strategy if the mapper is of this type in order to resolve the strategy.

mariusc83 (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/WireframeMapper.kt
louiszawadzki: Yes I agree with @mariusc83, sorry if the requirements were unclear. 
The idea is to leave the `WireframeMapper` interface untouched, and only touch `NonExhaustiveTypeMapper`.

jonathanmos: I've made this change - now `TraversalStrategy` is internal within `TreeViewTraversal` (it doesn't seem possible to make it private due to usage within the `TraversedTreeView` data class)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
```diff
@@ -38,11 +39,16 @@ internal class TreeViewTraversal(
...
+        val mapper = mappers.findFirstForType(view::class.java)
 
-        if (exhaustiveTypeMapper != null) {
-            val queueableViewMapper = QueueableViewMapper(exhaustiveTypeMapper, recordedDataQueueRefs)
-            traversalStrategy = TraversalStrategy.STOP_AND_RETURN_NODE
+        if (mapper != null) {
+            val queueableViewMapper =
+                QueueableViewMapper(mapper, recordedDataQueueRefs)
+            traversalStrategy = if (mapper is NonExhaustiveTypeMapper) {
+                TraversalStrategy.TRAVERSE_ALL_CHILDREN
```
xgouchet, line 48: > [!NOTE]
> I'm confused about the naming of the interface: It says the mapper is Non exhaustive, but it would imply using a `TRAVERSE_ALL_CHILDREN` strategy. Can we be a bit more explicit about what is being non exhaustive here?

jonathanmos: Until now we've referred to mappers that return a node whenever they map to a specific class as being "exhaustive", and so the naming follows from this. How about "TraverseAllChildrenMapper"?

xgouchet (approved):

louiszawadzki (approved):

mariusc83 (approved):

PR #1686, opened 2023-11-02
RUM-1983 Introduce the BatchProcessingLevel API

mariusc83:
### What does this PR do?

Modifies the core logic to get a list of batches instead of single batch. It's done through a new configuration called BatchProcessingLevel that allows controlling the amount of batches processed sequentially without a delay within one reading/uploading cycle. Currently it exposed 3 levels: low, medium and high that translate to 1, 10 and 100 of batches processed. By default it's taking up to 10 batches in a cycle.

This logic improves the data upload when batch back pressure occurs.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
@@ -44,38 +44,31 @@ internal class DataUploadRunnable(
     internal var currentDelayIntervalMs = uploadConfiguration.defaultDelayMs
     internal val minDelayMs = uploadConfiguration.minDelayMs
     internal val maxDelayMs = uploadConfiguration.maxDelayMs
+    internal val maxBatchesPerJob = uploadConfiguration.maxBatchesPerUploadJob
 
     //  region Runnable
 
     @WorkerThread
-    @Suppress("UnsafeThirdPartyFunctionCall") // called inside a dedicated executor
     override fun run() {
         if (isNetworkAvailable() && isSystemReady()) {
             val context = contextProvider.context
             // TODO RUMM-0000 it should be already on the worker thread and if readNextBatch is async,
             //  we should wait until it completes before scheduling further
-            val lock = CountDownLatch(1)
-            storage.readNextBatch(
-                noBatchCallback = {
-                    increaseInterval()
-                    lock.countDown()
-                }
-            ) { batchId, reader ->
-                try {
-                    val batch = reader.read()
-                    val batchMeta = reader.currentMetadata()
-
-                    consumeBatch(
-                        context,
-                        batchId,
-                        batch,
-                        batchMeta
-                    )
-                } finally {
-                    lock.countDown()
-                }
+            var batchConsumerAvailableAttempts = maxBatchesPerJob
+            var lastBatchUploadStatus: UploadStatus?
+            do {
+                batchConsumerAvailableAttempts--
+                lastBatchUploadStatus = handleNextBatch(context)
+            } while (batchConsumerAvailableAttempts > 0 &&
+                lastBatchUploadStatus is UploadStatus.Success
+            )
+            if (lastBatchUploadStatus != null) {
+                handleBatchConsumingJobFrequency(lastBatchUploadStatus)
+            } else {
+                // there was no batch left or there was a problem reading the next batch
+                // in the storage so we increase the interval
+                increaseInterval()
             }
-            lock.await(batchUploadWaitTimeoutMs, TimeUnit.MILLISECONDS)
         }
 
         scheduleNextUpload()
@@ -84,6 +77,42 @@ internal class DataUploadRunnable(
     // endregion
 
     // region Internal
+
+    private fun handleBatchConsumingJobFrequency(lastBatchUploadStatus: UploadStatus) {
+        if (lastBatchUploadStatus.shouldRetry) {
+            increaseInterval()
+        } else {
+            decreaseInterval()
+        }
+    }
+
+    @WorkerThread
+    @Suppress("UnsafeThirdPartyFunctionCall") // called inside a dedicated executor
+    private fun handleNextBatch(context: DatadogContext): UploadStatus? {
+        var uploadStatus: UploadStatus? = null
+        val lock = CountDownLatch(1)
+        storage.readNextBatch(
+            noBatchCallback = {
+                lock.countDown()
+            }
+        ) { batchId, reader ->
+            try {
+                val batch = reader.read()
+                val batchMeta = reader.currentMetadata()
+
+                uploadStatus = consumeBatch(
+                    context,
+                    batchId,
+                    batch,
+                    batchMeta
+                )
+            } finally {
+                lock.countDown()
+            }
+        }
+        lock.await(batchUploadWaitTimeoutMs, TimeUnit.MILLISECONDS)
+        return uploadStatus
+    }
 
     private fun isNetworkAvailable(): Boolean {
         val networkInfo = networkInfoProvider.getLatestNetworkInfo()
@@ -115,7 +144,7 @@ internal class DataUploadRunnable(
         batchId: BatchId,
         batch: List<RawBatchEvent>,
         batchMeta: ByteArray?
-    ) {
+    ): UploadStatus {
         val status = dataUploader.upload(context, batch, batchMeta)
         val removalReason = if (status is UploadStatus.RequestCreationError) {
             RemovalReason.Invalid
@@ -123,14 +152,9 @@ internal class DataUploadRunnable(
             RemovalReason.IntakeCode(status.code)
         }
         storage.confirmBatchRead(batchId, removalReason) {
-            if (status.shouldRetry) {
-                it.markAsRead(false)
-                increaseInterval()
-            } else {
-                it.markAsRead(true)
-                decreaseInterval()
-            }
+            it.markAsRead(deleteBatch = !status.shouldRetry)
         }
+        return status
     }
 
     @Suppress("UnsafeThirdPartyFunctionCall") // rounded Double isn't NaN
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
@@ -36,12 +36,14 @@ import org.junit.jupiter.api.extension.Extensions
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.MethodSource
 import org.mockito.Mock
+import org.mockito.invocation.InvocationOnMock
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
 import org.mockito.kotlin.any
 import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.doAnswer
 import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.doReturnConsecutively
 import org.mockito.kotlin.doThrow
 import org.mockito.kotlin.eq
 import org.mockito.kotlin.mock
@@ -52,9 +54,12 @@ import org.mockito.kotlin.verifyNoInteractions
 import org.mockito.kotlin.verifyNoMoreInteractions
 import org.mockito.kotlin.whenever
 import org.mockito.quality.Strictness
+import org.mockito.stubbing.Answer
 import java.util.concurrent.ScheduledThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 
+// TODO: RUM-2014 Simplify / Refactor these tests when we're going to switch to the
+// sync API for uploading data
 @Extensions(
     ExtendWith(MockitoExtension::class),
     ExtendWith(ForgeExtension::class)
@@ -90,10 +95,14 @@ internal class DataUploadRunnableTest {
     @Forgery
     lateinit var fakeDataUploadConfiguration: DataUploadConfiguration
 
+    private var expectedBatchesHandled: Int = 0
+
     private lateinit var testedRunnable: DataUploadRunnable
 
     @BeforeEach
     fun `set up`(forge: Forge) {
+        // to make sure the existing tests based only on 1 batch are not broken
+        fakeDataUploadConfiguration = fakeDataUploadConfiguration.copy(maxBatchesPerUploadJob = 1)
         val fakeNetworkInfo =
             NetworkInfo(
                 forge.aValueFrom(
@@ -101,6 +110,7 @@ internal class DataUploadRunnableTest {
                     exclude = listOf(NetworkInfo.Connectivity.NETWORK_NOT_CONNECTED)
                 )
             )
+        expectedBatchesHandled = fakeDataUploadConfiguration.maxBatchesPerUploadJob
         whenever(mockNetworkInfoProvider.getLatestNetworkInfo()) doReturn fakeNetworkInfo
         val fakeSystemInfo = SystemInfo(
             batteryFullOrCharging = true,
@@ -191,8 +201,8 @@ internal class DataUploadRunnableTest {
         // Then
         verify(batchConfirmation).markAsRead(true)
         verifyNoMoreInteractions(batchConfirmation)
-        verify(batchReader).read()
-        verify(mockDataUploader).upload(fakeContext, batch, batchMetadata)
+        verify(batchReader, times(expectedBatchesHandled)).read()
+        verify(mockDataUploader, times(expectedBatchesHandled)).upload(fakeContext, batch, batchMetadata)
         verify(mockThreadPoolExecutor).schedule(
             same(testedRunnable),
             any(),
@@ -243,10 +253,11 @@ internal class DataUploadRunnableTest {
         testedRunnable.run()
 
         // Then
-        verify(batchConfirmation).markAsRead(true)
+        verify(batchConfirmation, times(expectedBatchesHandled)).markAsRead(true)
         verifyNoMoreInteractions(batchConfirmation)
-        verify(batchReader).read()
-        verify(mockDataUploader).upload(fakeContext, batch, batchMetadata)
+        verify(batchReader, times(expectedBatchesHandled)).read()
+        verify(mockDataUploader, times(expectedBatchesHandled))
+            .upload(fakeContext, batch, batchMetadata)
         verify(mockThreadPoolExecutor).schedule(
             same(testedRunnable),
             any(),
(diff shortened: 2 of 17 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -54,28 +55,33 @@ internal class DataUploadRunnable(
...
-                        context,
-                        batchId,
-                        batch,
-                        batchMeta
-                    )
-                } finally {
-                    lock.countDown()
+            var batchesSent = 0
+            while (batchesSent < maxBatchesPersUpload) {
+                val lock = CountDownLatch(1)
```
mariusc83, line 60: TODO: extract this in a method

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -54,28 +55,40 @@ internal class DataUploadRunnable(
...
-                    lock.countDown()
+            var batchesSent = 0
+            var lastUploadWasSuccess = false
+            do  {
+                val lock = CountDownLatch(1)
+                storage.readNextBatch(
+                    noBatchCallback = {
+                        increaseInterval()
+                        lock.countDown()
+                        batchesSent = maxBatchesPersUpload
```
0xnm, line 66: this line is misleading and not actually true, the line with `lastUploadWasSuccess` below should be enough. Btw, is it possible to just use `break` here?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -54,28 +55,40 @@ internal class DataUploadRunnable(
...
+                        val batchMeta = reader.currentMetadata()
+
+                        val status = consumeBatch(
+                            context,
+                            batchId,
+                            batch,
+                            batchMeta
+                        )
+                        if (status !is UploadStatus.Success) {
+                            batchesSent = maxBatchesPersUpload
```
0xnm, line 81: this is not quite true, why relying only on `lastUploadWasSuccess` is not enough?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -54,28 +55,40 @@ internal class DataUploadRunnable(
...
+                            lastUploadWasSuccess = false
+                        }
+                    } finally {
+                        batchesSent++
+                        lastUploadWasSuccess = true
```
0xnm, line 86: is this to try reading next batch?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -54,28 +55,40 @@ internal class DataUploadRunnable(
...
+                        lock.countDown()
+                        batchesSent = maxBatchesPersUpload
+                        lastUploadWasSuccess = false
+                    }
+                ) { batchId, reader ->
+                    try {
+                        val batch = reader.read()
+                        val batchMeta = reader.currentMetadata()
+
+                        val status = consumeBatch(
```
0xnm, line 74: are we ok that each call to `consumeBatch` will decrease scheduling interval in case of upload success? Should we decrease interval only when the whole queue is processed?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -54,28 +55,40 @@ internal class DataUploadRunnable(
...
-                noBatchCallback = {
-                    increaseInterval()
-                    lock.countDown()
+            var batchesSent = 0
+            var lastUploadWasSuccess = false
+            do  {
+                val lock = CountDownLatch(1)
+                storage.readNextBatch(
+                    noBatchCallback = {
+                        increaseInterval()
```
mariusc83, line 64: Question for reviewers: Should we increase the interval only once at the end of the queue if all the batches were sent or should we keep it like this and increase every time a batch was sent (meaning that will increase up to the max limit if all the batches are sent). 
If we go with first option there's less complexity and IMO we are going to keep exactly the same rule as the success rate of the uploads is dictating that. 
If we go with the second option we need to change a bit more this logic and we will also introduce a skew in our upload frequency (if 10 uploads were successful we will only increase the interval once whereas it should be to max frequency)
@0xnm , @xgouchet , @jonathanmos need your opinion here

0xnm: IMO we are changing the definition of the unit-of-work in the uploader, not it will be not a single batch, but a pack of 1-10-100 batches, so we need to change the interval only when unit-of-work is processed.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
mariusc83: yes a break would have been nice, unfortunately the compiler complains about using a break outside of the function/class :(. I wonder if there is any workaround for this in Kotlin, I could not find any except using a flag.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
mariusc83: yes was a leftover, sorry for the trouble ;)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
mariusc83: yes...do not mind the logic there it is a bit broken that flag should be marked as true only if the status was ok. I just wanted to get a review on the approach, sorry for the inconvenience.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
mariusc83: @0xnm I agree but on the other hand if we say handle 100 batches in one job then we will increase the interval only once ? We should have the highest frequency in that case, that's the reason I am willing to keep the first approach in place as it kind of keeps same functionality.

xgouchet: I do agree with Nikita, we should keep thinking about unit of work. Extreme case, we upload 100 batches each time, this means we don't need to do an upload cycle as often, since chances are there won't be a 100 batches to upload next.

mariusc83: ok...tks all, decision taken, we will go with the unit of work interval decrease/increase. Tks !!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -54,28 +55,13 @@ internal class DataUploadRunnable(
...
-                        batchId,
-                        batch,
-                        batchMeta
-                    )
-                } finally {
-                    lock.countDown()
-                }
-            }
-            lock.await(batchUploadWaitTimeoutMs, TimeUnit.MILLISECONDS)
+            var batchConsumerCoins = maxBatchesPerAsyncJob
```
jonathanmos, line 58: What are batchConsumerCoins? Does this mean remaining batch upload attempts?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -85,6 +71,40 @@ internal class DataUploadRunnable(
 
     // region Internal
 
+    private fun handleBatchConsumingJobFrequency(lastBatchUploadStatus: UploadStatus) {
+        if (lastBatchUploadStatus is UploadStatus.Success) {
+            decreaseInterval()
```
jonathanmos, line 76: Since we decrease and increase at a constant rate of X%, doesn't this expose us to the thundering herd problem? Should we add some sort of random seed here?

Also, there might be a potential issue here. Let's say a client can send X batches in a time period, but X+1 batches in that period will cause a network issue for that user. The user sends X, receives success and scales up, then sends X+1, receives failure so scales down, then X and scales up, then X+1 and scales down - in a loop. About half of all network requests will therefore fail causing wasted network for the client. Perhaps we could optimize this?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
mariusc83: yes...need to find a better name there

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -125,12 +153,11 @@ internal class DataUploadRunnable(
         storage.confirmBatchRead(batchId, removalReason) {
             if (status.shouldRetry) {
```
0xnm, line 154: minor: then the whole lambda can be simplified as `it.markAsRead(deleteBatch = !status.shouldRetry)`

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
```diff
@@ -60,7 +60,7 @@ import java.util.concurrent.TimeUnit
     ExtendWith(ForgeExtension::class)
 )
 @MockitoSettings(strictness = Strictness.LENIENT)
-@ForgeConfiguration(Configurator::class)
+@ForgeConfiguration(Configurator::class, seed = 0xf4c7ed013062L)
```
0xnm, line 63: debugging leftover? :)

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
```diff
@@ -90,17 +90,24 @@ internal class DataUploadRunnableTest {
...
+    private var expectedBatchesHandled: Int = 0
+
     private lateinit var testedRunnable: DataUploadRunnable
 
     @BeforeEach
     fun `set up`(forge: Forge) {
+        // to make sure the basic tests are not flaky
+        // the cases where the maxBatchesPerUploadJob is variable are tested in dedicated test
+        // methods
+        fakeDataUploadConfiguration = fakeDataUploadConfiguration.copy(maxBatchesPerUploadJob = 1)
```
0xnm, line 102: where the flakiness comes from? In fact it seems we are always testing the case where `maxBatchesPerUploadJob = = 1`, because I don't see any place where it would have a different value

mariusc83: maybe it's a mistake on my end I will update those tests where it should have a different value. I wanted to make sure that the basics tests that are testing the increase/decrease interval are only based on a single value to avoid problems and complexity. I will add a better explanation

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
```diff
@@ -651,7 +668,13 @@ internal class DataUploadRunnableTest {
 
         // Then
         val captor = argumentCaptor<Long>()
-        verify(mockThreadPoolExecutor, times(5))
+        verify(
+            mockThreadPoolExecutor,
+            times(
+                5 *
+                    fakeDataUploadConfiguration.maxBatchesPerUploadJob
```
0xnm, line 675:
```suggestion
                    expectedBatchesHandled
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
```diff
@@ -917,6 +977,156 @@ internal class DataUploadRunnableTest {
...
+        verifyNoMoreInteractions(mockDataUploader)
+        verify(mockThreadPoolExecutor).schedule(
+            same(testedRunnable),
+            any(),
+            eq(TimeUnit.MILLISECONDS)
+        )
+    }
+
+    @Test
+    fun `𝕄 exhaust the available batches W run {maxBatchesPerJob higher availableBatches}`(
```
0xnm, line 1041: seems like `higher or equal` according to the test

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/DataUploadConfigurationForgeryFactory.kt
```diff
@@ -14,6 +14,7 @@ import fr.xgouchet.elmyr.ForgeryFactory
 internal class DataUploadConfigurationForgeryFactory : ForgeryFactory<DataUploadConfiguration> {
     override fun getForgery(forge: Forge): DataUploadConfiguration {
         val frequency: UploadFrequency = forge.getForgery()
-        return DataUploadConfiguration(frequency)
+        // we limit the size to avoid OOM errors inside our tests
+        return DataUploadConfiguration(frequency, forge.anInt(min = 1, max = 50))
```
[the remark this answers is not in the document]
mariusc83: the `maxBatchesPerUpload` gives the size of the batches list in some of our tests. That number can go up to `Int.MAX` and it creates problems ...at least I had those on my end. I guess a rule of thumb would be to try to stick with low size numbers when using lists of objects in our tests.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -44,38 +44,30 @@ internal class DataUploadRunnable(
...
+            do {
+                batchConsumerAvailableAttempts--
+                lastBatchUploadStatus = handleNextBatch(context)
+            } while (batchConsumerAvailableAttempts > 0 &&
+                lastBatchUploadStatus is UploadStatus.Success
+            )
+            if (lastBatchUploadStatus != null) {
+                handleBatchConsumingJobFrequency(lastBatchUploadStatus)
+            } else {
+                // there was no batch left in the storage so we increase the interval
```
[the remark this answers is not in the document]
mariusc83: so do you think we should not increase the interval or not add the comment ? For me it makes sense to increase the interval

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
```diff
@@ -44,38 +44,30 @@ internal class DataUploadRunnable(
...
-                    )
-                } finally {
-                    lock.countDown()
-                }
+            var batchConsumerAvailableAttempts = maxBatchesPerJob
+            var lastBatchUploadStatus: UploadStatus?
+            do {
+                batchConsumerAvailableAttempts--
+                lastBatchUploadStatus = handleNextBatch(context)
+            } while (batchConsumerAvailableAttempts > 0 &&
```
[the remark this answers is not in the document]
mariusc83: I think it will add more complexity ...I would rather wait and see how this will behave in telemetry and in the case we are going to see a problem add this in place. I am afraid that we will have to change the way we handle our interval after as if we have 2 failures in one loop ...which one to take into account  for interval decrease/increase ? The last one ?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
```diff
@@ -802,33 +842,53 @@ internal class DataUploadRunnableTest {
     @MethodSource("retryBatchStatusValues")
     fun `𝕄 increase delay between runs 𝕎 batch fails and should be retried`(
         status: UploadStatus,
-        @IntForgery(16, 64) runCount: Int,
-        @Forgery batch: List<RawBatchEvent>,
-        @StringForgery batchMeta: String,
-        forge: Forge
+        @IntForgery(1, 10) runCount: Int,
```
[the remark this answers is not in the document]
mariusc83: I did not se the point to have it that high ? Have in mind that with the maxBatchesProcessingLevel now we are increasing the number of iterations so the `runCount` could be smaller. Any particular reason you wanted it that high ?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/DataUploadConfigurationForgeryFactory.kt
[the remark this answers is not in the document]
0xnm: yes, but since we have a definition `HIGH = 100`, we need to set at least 100 as max to make sure things work for `HIGH` as well (although the allocation pattern would eventually be different in unit-tests compared to the on-device run, it is still better to align imo).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnable.kt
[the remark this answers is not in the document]
0xnm: I mean this comment should be updated, currently it may be misleading.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/v2/DataUploadRunnableTest.kt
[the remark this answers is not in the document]
0xnm: no, I was just curious what was the reason for this change

jonathanmos (approved):

0xnm (approved):

xgouchet (approved):

PR #1687, opened 2023-11-02
RUM-1951 Improve performance with base64

jonathanmos:
### What does this PR do?
Benchmarks that I've done show that performance with base64 was around 20ms on the main thread. This PR moves bitmap resizing tasks to a background thread in order to improve performance to around 7ms on the main thread. In addition the PR handles a couple of additional issues:
• Wrap the call to `bitmap.compress` in`WebPImageCompression` in try/catch to prevent an exception if the bitmap was recycled while we were working on it (can happen with the system-cached bitmaps that we take from `BitmapDrawable`)
• Rename some variables that were just called "callback" to make it clearer when reading the code what callback is involved

### Motivation
SDK should not impact the main thread more than 8ms.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
@@ -16,13 +16,13 @@ import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
 import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.internal.utils.executeSafe
 import com.datadog.android.sessionreplay.internal.recorder.base64.Cache.Companion.DOES_NOT_IMPLEMENT_COMPONENTCALLBACKS
 import com.datadog.android.sessionreplay.internal.utils.Base64Utils
 import com.datadog.android.sessionreplay.internal.utils.DrawableUtils
 import com.datadog.android.sessionreplay.model.MobileSegment
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.LinkedBlockingDeque
-import java.util.concurrent.RejectedExecutionException
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 
@@ -49,21 +49,29 @@ internal class Base64Serializer private constructor(
         drawableWidth: Int,
         drawableHeight: Int,
         imageWireframe: MobileSegment.Wireframe.ImageWireframe,
-        callback: Base64SerializerCallback? = null
+        base64SerializerCallback: Base64SerializerCallback
     ) {
         registerCallbacks(applicationContext)
 
-        tryToGetBase64FromCache(drawable, imageWireframe, callback)
-            ?: tryToGetBitmapFromBitmapDrawable(drawable, imageWireframe, callback)
-            ?: tryToDrawNewBitmap(
-                drawable,
-                drawableWidth,
-                drawableHeight,
-                displayMetrics,
-                imageWireframe,
-                callback
+        tryToGetBase64FromCache(drawable, imageWireframe, base64SerializerCallback)
+            ?: tryToGetBitmapFromBitmapDrawable(
+                drawable = drawable,
+                displayMetrics = displayMetrics,
+                imageWireframe = imageWireframe,
+                base64SerializerCallback = base64SerializerCallback
             )
-            ?: callback?.onReady()
+            ?: tryToDrawNewBitmap(
+                drawable = drawable,
+                drawableWidth = drawableWidth,
+                drawableHeight = drawableHeight,
+                displayMetrics = displayMetrics,
+                imageWireframe = imageWireframe,
+                base64SerializerCallback = base64SerializerCallback,
+
+                // this parameter is used to avoid infinite recursion
+                // basically we only allow one attempt to recreate the bitmap
+                didCallOriginateFromFailover = false
+            )
     }
 
     // endregion
@@ -78,15 +86,31 @@ internal class Base64Serializer private constructor(
     // region private
 
     @WorkerThread
-    private fun serialiseBitmap(
+    private fun serializeBitmap(
         drawable: Drawable,
+        displayMetrics: DisplayMetrics,
         bitmap: Bitmap,
         shouldCacheBitmap: Boolean,
         imageWireframe: MobileSegment.Wireframe.ImageWireframe,
-        base64SerializerCallback: Base64SerializerCallback?
+        base64SerializerCallback: Base64SerializerCallback,
+
+        // this parameter is used to avoid infinite recursion
+        // basically we only allow one attempt to recreate the bitmap
+        didCallOriginateFromFailover: Boolean
     ) {
-        val base64String = convertBmpToBase64(drawable, bitmap, shouldCacheBitmap)
-        finalizeRecordedDataItem(base64String, imageWireframe, base64SerializerCallback)
+        val base64String = convertBitmapToBase64(
+            drawable = drawable,
+            displayMetrics = displayMetrics,
+            bitmap = bitmap,
+            shouldCacheBitmap = shouldCacheBitmap,
+            imageWireframe = imageWireframe,
+            base64SerializerCallback = base64SerializerCallback,
+            didCallOriginateFromFailover = didCallOriginateFromFailover
+        )
+
+        if (base64String.isNotEmpty()) {
+            finalizeRecordedDataItem(base64String, imageWireframe, base64SerializerCallback)
+        }
     }
 
     @MainThread
@@ -116,10 +140,37 @@ internal class Base64Serializer private constructor(
     }
 
     @WorkerThread
-    private fun convertBmpToBase64(drawable: Drawable, bitmap: Bitmap, shouldCacheBitmap: Boolean): String {
+    private fun convertBitmapToBase64(
+        drawable: Drawable,
+        displayMetrics: DisplayMetrics,
+        bitmap: Bitmap,
+        shouldCacheBitmap: Boolean,
+        imageWireframe: MobileSegment.Wireframe.ImageWireframe,
+        base64SerializerCallback: Base64SerializerCallback,
+
+        // this parameter is used to avoid infinite recursion
+        // basically we only allow one attempt to recreate the bitmap
+        didCallOriginateFromFailover: Boolean
+    ): String {
         val base64Result: String
 
         val byteArray = webPImageCompression.compressBitmap(bitmap)
+
+        // failed to get byteArray
+        // Try once to recreate bitmap from the drawable
+        if (byteArray.isEmpty() && bitmap.isRecycled && !didCallOriginateFromFailover) {
+            tryToDrawNewBitmap(
+                drawable = drawable,
+                drawableWidth = bitmap.width,
+                drawableHeight = bitmap.height,
+                displayMetrics = displayMetrics,
+                imageWireframe = imageWireframe,
+                base64SerializerCallback = base64SerializerCallback,
+                didCallOriginateFromFailover = true
+            )
+
+            return ""
+        }
 
         base64Result = base64Utils.serializeToBase64String(byteArray)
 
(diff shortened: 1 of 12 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -37,27 +41,30 @@ internal class DrawableUtils(
...
 
-        // erase the canvas
-        // needed because overdrawing an already used bitmap causes unusual visual artifacts
-        canvas.drawColor(Color.TRANSPARENT, PorterDuff.Mode.MULTIPLY)
+        lambdaExecutor.execute(
+            backgroundWork,
+            foregroundWork,
+            callback = object :
+                LambdaExecutor.LambdaExecutorCallback {
+                override fun onSuccess() {}
```
mariusc83, line 58: why `onSuccess` if you do not use it ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/LambdaExecutor.kt
```diff
@@ -0,0 +1,71 @@
...
+package com.datadog.android.sessionreplay.internal.utils
+
+import android.os.Handler
+import android.os.Looper
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.LinkedBlockingDeque
+import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.TimeUnit
+
+internal class LambdaExecutor<T>(
```
mariusc83, line 16: this executor is quite complex and a bit over - engineered for me. I am not sure what you are trying to achieve with it ? If you need some work to be done on the result why not adding this in the caller thread (Main Thread in your case) in the `onSuccess` callback ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -37,27 +41,30 @@ internal class DrawableUtils(
...
+        config: Config = Config.ARGB_8888,
+        base64SerializerCallback: Base64SerializerCallback,
+        bitmapCreationCallback: Base64Serializer.BitmapCreationCallback
+    ) {
+        val backgroundWork =
+            createBackgroundLambda(drawableWidth, drawableHeight, requestedSizeInBytes, displayMetrics, config)
 
-        val bitmap = getBitmapBySize(displayMetrics, width, height, config) ?: return null
-        val canvas = canvasWrapper.createCanvas(bitmap) ?: return null
+        val foregroundWork = createForegroundLambda(drawable, bitmapCreationCallback)
```
mariusc83, line 51: why do you need this ? Can't you just use an executor that executes the background work and in the `onSuccess` of that you can go on with your `foregroundWork` ? Also I would make sure I am using the same executor in the `DrawableUtils` and `Base64Serializer` class otherwise you will create too many threads and also we are going to have problems when debugging this code.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -142,25 +142,26 @@ internal class Base64Serializer private constructor(
...
-                imageWireframe,
-                base64SerializerCallback
-            )
-            return resizedBitmap
-        }
-
-        return null
+            displayMetrics,
+            base64SerializerCallback = base64SerializerCallback,
+            bitmapCreationCallback = object : BitmapCreationCallback {
```
mariusc83, line 153: I see that `Base64Serializer` already has an executor. I wonder why don't you use this instead to executor call/execute the `drawableUtils.createBitmapOfApproxSizeFromDrawable` and leave that method synchronous and remove the `LambdaExecutor`

mariusc83 (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/LambdaExecutor.kt
jonathanmos: Indeed, you're correct that it can be done in a simpler way using the threadpoolexecutor from `Base64Serializer`. I've refactored and removed the `LambdaExecutor` class

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -142,25 +141,26 @@ internal class Base64Serializer private constructor(
...
+                override fun onReady(bitmap: Bitmap) {
+                    serializeBitmapAsynchronously(
+                        drawable,
+                        bitmap,
+                        shouldCacheBitmap = true,
+                        imageWireframe,
+                        base64SerializerCallback
```
0xnm, line 159: design question: why do we have to pass `base64SerializerCallback` twice? first time as method argument, second time wrapped in `bitmapCreationCallback`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -169,25 +169,31 @@ internal class Base64Serializer private constructor(
...
-                    drawable,
-                    scaledBitmap,
-                    shouldCacheBitmap,
-                    imageWireframe,
-                    base64SerializerCallback
-                )
-
-                result = scaledBitmap
-            }
+            val bitmap = (drawable as BitmapDrawable).bitmap
```
0xnm, line 173: Is it a safe cast? Maybe we should change a signature of `shouldUseDrawableBitmap` method to accept only `BitmapDrawable`? This has the following benefits: it helps compiler to use smart-cast, no cast is needed; also the name `shouldUseDrawableBitmap` gets more sense, because it accepts only `BitmapDrawable`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -70,6 +87,69 @@ internal class DrawableUtils(
         return bitmapWrapper.createScaledBitmap(bitmap, width, height, false)
     }
 
+    internal interface ResizeBitmapCallback {
+        fun onSuccess(bitmap: Bitmap)
+        fun onFailure()
+    }
+
+    private fun executeRunnable(runnable: Runnable) {
```
mariusc83, line 95: there is already an extension method `executeSafe` that you can use in the `ConcurrencyExt` class

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -169,25 +169,31 @@ internal class Base64Serializer private constructor(
...
+
+            Runnable {
+                @Suppress("ThreadSafety") // this runs inside an executor
+                drawableUtils.createScaledBitmap(
+                    bitmap
+                )?.let { scaledBitmap ->
+                    val shouldCacheBitmap = scaledBitmap != drawable.bitmap
```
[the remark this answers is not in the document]
mariusc83: I have the same question here...it seems that `createScaledBitmap` always creates a new instance so how can this condition be false ?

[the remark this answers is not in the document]
jonathanmos: If createScaledBitmap is called with the same dimensions as the existing bitmap then it returns the instance of the existing bitmap. I can also compare width/height of the bitmaps if it would be more performant than comparing by reference

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/WebPImageCompression.kt
```diff
@@ -25,8 +25,17 @@ internal class WebPImageCompression : ImageCompression {
...
-        @Suppress("UnsafeThirdPartyFunctionCall") // stream is not null and image quality is between 0 and 100
-        bitmap.compress(imageFormat, IMAGE_QUALITY, byteArrayOutputStream)
+
+        @Suppress("SwallowedException")
+        try {
+            // stream is not null and image quality is between 0 and 100
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            bitmap.compress(imageFormat, IMAGE_QUALITY, byteArrayOutputStream)
+        } catch (e: IllegalStateException) {
+            // if the bitmap was recycled while we were working on it
```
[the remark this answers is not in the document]
jonathanmos: The bitmaps are recycled by the system based on the view lifecycle, and I can reproduce it fairly easily. I can add a fallback in base64serializer that creates the bitmap from the drawable in the case of an empty bytearray, so that we don't lose the bitmap entirely, at the cost of some added complexity to the flow

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -78,15 +83,28 @@ internal class Base64Serializer private constructor(
...
-    private fun serialiseBitmap(
+    private fun serializeBitmap(
         drawable: Drawable,
+        displayMetrics: DisplayMetrics,
         bitmap: Bitmap,
         shouldCacheBitmap: Boolean,
         imageWireframe: MobileSegment.Wireframe.ImageWireframe,
-        base64SerializerCallback: Base64SerializerCallback?
+        base64SerializerCallback: Base64SerializerCallback,
+        isFailover: Boolean
```
0xnm, line 93: should we rename this property or add any explanation comment at least? Currently it is not clear what it is doing tbh.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -78,15 +83,28 @@ internal class Base64Serializer private constructor(
...
     ) {
-        val base64String = convertBmpToBase64(drawable, bitmap, shouldCacheBitmap)
-        finalizeRecordedDataItem(base64String, imageWireframe, base64SerializerCallback)
+        val base64String = convertBmpToBase64(
```
0xnm, line 95: Let's try to use a common lingvo and use `bitmap` everywhere

```suggestion
        val base64String = convertBitmapToBase64(
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
0xnm: I believe there is no need to mock `executeSafe` in tests, because it is an extension method - it will be called already on the mock of the executor and the only other mock it needs it is the logger. So I guess the reason to not use it here is just the need to provide the logger which is not here yet.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64SerializerTest.kt
```diff
@@ -654,6 +654,55 @@ internal class Base64SerializerTest {
...
+        val emptyByteArray = ByteArray(0)
+        whenever(mockWebPImageCompression.compressBitmap(any()))
+            .thenReturn(emptyByteArray)
+            .thenReturn(fakeByteArray)
+
+        whenever(mockDrawableUtils.createScaledBitmap(any(), anyOrNull()))
+            .thenReturn(mockBitmap)
+            .thenReturn(mockCreatedBitmap)
+
+        whenever(mockBase64Utils.serializeToBase64String(any()))
```
0xnm, line 679: probably for input parameters of the functions called it is possible to use `mockBitmap`, `mockBitmapDrawable` where possible instead of just `any`? or other objects we already know.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64SerializerTest.kt
```diff
@@ -654,6 +654,55 @@ internal class Base64SerializerTest {
...
+        // Then
+        verify(mockDrawableUtils, times(1)).createBitmapOfApproxSizeFromDrawable(
+            drawable = any(),
+            drawableWidth = any(),
+            drawableHeight = any(),
+            displayMetrics = any(),
+            requestedSizeInBytes = anyOrNull(),
+            config = anyOrNull(),
+            base64SerializerCallback = any(),
+            bitmapCreationCallback = any()
```
0xnm, line 702: is it possible to provide parameters we know for sure instead of just generic `any`?

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/WebPImageCompressionTest.kt
```diff
@@ -85,5 +86,17 @@ internal class WebPImageCompressionTest {
...
+    @Test
+    fun `M return empty bytearray W compressBitmap { bitmap was already recycled }`() {
+        // Given
+        whenever(mockBitmap.isRecycled).thenReturn(true)
+
+        // When
+        val result = testedImageCompression.compressBitmap(mockBitmap)
+
+        // Then
+        assertThat(result).isEqualTo(ByteArray(0))
```
0xnm, line 98: probably `isEmpty` assertion will be more descriptive here

```suggestion
        assertThat(result).isEmpty()
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtilsTest.kt
```diff
@@ -65,19 +71,49 @@ internal class DrawableUtilsTest {
...
+        whenever(mockCanvasWrapper.createCanvas(any()))
             .thenReturn(mockCanvas)
         whenever(mockBitmap.config).thenReturn(mockConfig)
         whenever(mockBitmapPool.getBitmapByProperties(any(), any(), any())).thenReturn(null)
 
+        doAnswer { invocation ->
+            val work = invocation.getArgument(0) as Runnable
+            work.run()
+            null
+        }.`when`(mockMainThreadHandler).post(
```
0xnm, line 99:
```suggestion
        }.whenever(mockMainThreadHandler).post(
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtilsTest.kt
```diff
@@ -65,19 +71,49 @@ internal class DrawableUtilsTest {
...
+            any()
+        )
+
+        doAnswer { invocation ->
+            val work = invocation.getArgument(0) as Runnable
+            work.run()
+            null
+        }.`when`(mockExecutorService).submit(
```
0xnm, line 107:
```suggestion
        }.whenever(mockExecutorService).submit(
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtilsTest.kt
```diff
@@ -185,78 +227,86 @@ internal class DrawableUtilsTest {
...
         assertThat(displayMetricsCaptor.firstValue).isEqualTo(mockDisplayMetrics)
-
-        assertThat(result).isEqualTo(mockBitmap)
     }
 
+    // endregion
+
     @Test
-    fun `M return null W createBitmapFromDrawableOfApproxSize() { failed to create bmp }`() {
+    fun `M use bitmap from pool W createBitmapFromDrawable() { exists in pool }`(
```
0xnm, line 235: the method beign called is `createBitmapOfApproxSizeFromDrawable`, same for the tests below

```suggestion
    fun `M use bitmap from pool W createBitmapOfApproxSizeFromDrawable() { exists in pool }`(
```

0xnm (commented):
lgtm, I just added some comments/suggestions

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: yes, there is a comment in convertBmpToBase64, but I'll try to make this clearer

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
jonathanmos: I spoke with Xavier about how it would be possible to test while using this extension. He suggested a solution so I will push executeSafe with the next changes and remove executeRunnable

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64SerializerTest.kt
jonathanmos: I don't know the value of the last callback since it's instantiated when DrawableUtils is called. However, I can use argumentCaptors and assert the values afterwards

0xnm: yes, I mean some parameters we probably already have, like mock of drawable, bitmaps, etc., so that we can use them instead  of `any`.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtilsTest.kt
jonathanmos: good catch - I missed that when renaming

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtilsTest.kt
```diff
@@ -185,78 +230,86 @@ internal class DrawableUtilsTest {
...
     }
 
-    // endregion
-
-    fun `M use bitmap from pool W createBitmapFromDrawable() { exists in pool }`(
-        @IntForgery(min = 1, max = 1000) viewWidth: Int,
-        @IntForgery(min = 1, max = 1000) viewHeight: Int
-    ) {
+    @Test
+    fun `M return onReady W createBitmapOfApproxSizeFromDrawable { failed to create canvas }`() {
```
0xnm, line 292:
```suggestion
    fun `M call onReady W createBitmapOfApproxSizeFromDrawable { failed to create canvas }`() {
```

0xnm (approved):

mariusc83 (approved):

PR #1690, opened 2023-11-08
RUM-1654 Add resourceId to ImageWireframe

jonathanmos:
### What does this PR do?
Adds the resourceId property to ImageWireframe, and additionally removes base64 from the MutationResolver diffing.  

### Motivation
Removing base64 from diffing improved performance on ios and in this PR we align on this change. Additionally, resourceId is a necessary prerequisite for the resource endpoint change.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
@@ -19,16 +19,18 @@ import com.datadog.android.sessionreplay.internal.utils.CacheUtils
 import com.datadog.android.sessionreplay.internal.utils.InvocationUtils
 
 internal class Base64LRUCache(
-    private val cacheUtils: CacheUtils<String, ByteArray> = CacheUtils(),
+    private val cacheUtils: CacheUtils<String, CacheData> = CacheUtils(),
     private val invocationUtils: InvocationUtils = InvocationUtils(),
-    private var cache: LruCache<String, ByteArray> =
+    private var cache: LruCache<String, CacheData> =
         object :
-            LruCache<String, ByteArray>(MAX_CACHE_MEMORY_SIZE_BYTES) {
-            override fun sizeOf(key: String, value: ByteArray): Int {
-                return value.size
+            LruCache<String, CacheData>(MAX_CACHE_MEMORY_SIZE_BYTES) {
+            override fun sizeOf(key: String, value: CacheData): Int {
+                val base64Size = value.base64Encoding.size
+                val hashSize = value.resourceId?.size ?: 0
+                return base64Size + hashSize
             }
         }
-) : Cache<Drawable, String>, ComponentCallbacks2 {
+) : Cache<Drawable, CacheData>, ComponentCallbacks2 {
 
     override fun onTrimMemory(level: Int) {
         cacheUtils.handleTrimMemory(level, cache)
@@ -45,24 +47,27 @@ internal class Base64LRUCache(
     }
 
     @Synchronized
-    override fun put(element: Drawable, value: String) {
+    override fun put(element: Drawable, value: CacheData) {
         val key = generateKey(element)
-        val byteArray = value.toByteArray(Charsets.UTF_8)
+        val base64Encoding = value.base64Encoding
+        val resourceId = value.resourceId
 
         @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
         invocationUtils.safeCallWithErrorLogging(
-            call = { cache.put(key, byteArray) },
+            call = { cache.put(key, CacheData(base64Encoding, resourceId)) },
             failureMessage = FAILURE_MSG_PUT_CACHE
         )
     }
 
     @Synchronized
-    override fun get(element: Drawable): String? =
+    override fun get(element: Drawable): CacheData? =
         @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
         invocationUtils.safeCallWithErrorLogging(
             call = {
                 cache.get(generateKey(element))?.let {
-                    String(it)
+                    val base64Encoding = it.base64Encoding
+                    val resourceId = it.resourceId
+                    CacheData(base64Encoding, resourceId)
                 }
             },
             failureMessage = FAILURE_MSG_GET_CACHE
@@ -117,5 +122,29 @@ internal class Base64LRUCache(
         private const val FAILURE_MSG_EVICT_CACHE_CONTENTS = "Failed to evict cache entries"
         private const val FAILURE_MSG_PUT_CACHE = "Failed to put item in cache"
         private const val FAILURE_MSG_GET_CACHE = "Failed to get item from cache"
+    }
+}
+
+internal data class CacheData(val base64Encoding: ByteArray, var resourceId: ByteArray?) {
+    // we must override these methods because we are using arrays as properties
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as CacheData
+
+        if (!base64Encoding.contentEquals(other.base64Encoding)) return false
+        if (resourceId != null) {
+            if (other.resourceId == null) return false
+            if (!resourceId.contentEquals(other.resourceId)) return false
+        } else if (other.resourceId != null) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = base64Encoding.contentHashCode()
+        result = 31 * result + (resourceId?.contentHashCode() ?: 0)
+        return result
     }
 }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
@@ -32,11 +32,12 @@ internal class Base64Serializer private constructor(
     private val drawableUtils: DrawableUtils,
     private val base64Utils: Base64Utils,
     private val webPImageCompression: ImageCompression,
-    private val base64LRUCache: Cache<Drawable, String>?,
+    private val base64LRUCache: Cache<Drawable, CacheData>,
     private val bitmapPool: BitmapPool?,
-    private val logger: InternalLogger
+    private val logger: InternalLogger,
+    private val md5HashGenerator: MD5HashGenerator
 ) {
-    private var isCacheRegisteredForCallbacks: Boolean = false
+    private var isBase64CacheRegisteredForCallbacks: Boolean = false
     private var isBitmapPoolRegisteredForCallbacks: Boolean = false
 
     // region internal
@@ -53,7 +54,11 @@ internal class Base64Serializer private constructor(
     ) {
         registerCallbacks(applicationContext)
 
-        tryToGetBase64FromCache(drawable, imageWireframe, base64SerializerCallback)
+        tryToGetBase64FromCache(
+            drawable = drawable,
+            imageWireframe = imageWireframe,
+            base64SerializerCallback = base64SerializerCallback
+        )
             ?: tryToGetBitmapFromBitmapDrawable(
                 drawable = drawable,
                 displayMetrics = displayMetrics,
@@ -98,31 +103,54 @@ internal class Base64Serializer private constructor(
         // basically we only allow one attempt to recreate the bitmap
         didCallOriginateFromFailover: Boolean
     ) {
-        val base64String = convertBitmapToBase64(
-            drawable = drawable,
-            displayMetrics = displayMetrics,
-            bitmap = bitmap,
-            shouldCacheBitmap = shouldCacheBitmap,
-            imageWireframe = imageWireframe,
-            base64SerializerCallback = base64SerializerCallback,
-            didCallOriginateFromFailover = didCallOriginateFromFailover
-        )
+        val byteArray = webPImageCompression.compressBitmap(bitmap)
 
-        if (base64String.isNotEmpty()) {
-            finalizeRecordedDataItem(base64String, imageWireframe, base64SerializerCallback)
+        // failed to get byteArray potentially because the bitmap was recycled before imageCompression
+        // Try once to recreate bitmap from the drawable
+        if (byteArray.isEmpty() && bitmap.isRecycled && !didCallOriginateFromFailover) {
+            tryToDrawNewBitmap(
+                drawable = drawable,
+                drawableWidth = bitmap.width,
+                drawableHeight = bitmap.height,
+                displayMetrics = displayMetrics,
+                imageWireframe = imageWireframe,
+                base64SerializerCallback = base64SerializerCallback,
+                didCallOriginateFromFailover = true
+            )
+
+            return
         }
+
+        if (byteArray.isEmpty()) {
+            // failed to get image data
+            base64SerializerCallback.onReady()
+            return
+        }
+
+        val base64String = convertBitmapToBase64(
+            byteArray = byteArray,
+            bitmap = bitmap,
+            shouldCacheBitmap = shouldCacheBitmap
+        ).toByteArray(Charsets.UTF_8)
+        val resourceId = md5HashGenerator.generate(byteArray)?.toByteArray(Charsets.UTF_8)
+        val cacheData = CacheData(base64String, resourceId)
+
+        if (base64String.isNotEmpty() || resourceId != null) {
+            base64LRUCache.put(drawable, cacheData)
+        }
+
+        finalizeRecordedDataItem(cacheData, imageWireframe)
+        base64SerializerCallback.onReady()
     }
 
     @MainThread
-    private fun registerCacheForCallbacks(applicationContext: Context) {
-        if (isCacheRegisteredForCallbacks) return
+    private fun registerBase64LruCacheForCallbacks(applicationContext: Context) {
+        if (isBase64CacheRegisteredForCallbacks) return
 
         if (base64LRUCache is ComponentCallbacks2) {
             applicationContext.registerComponentCallbacks(base64LRUCache)
-            isCacheRegisteredForCallbacks = true
+            isBase64CacheRegisteredForCallbacks = true
         } else {
-            // Temporarily use UNBOUND logger
-            // TODO: REPLAY-1364 Add logs here once the sdkLogger is added
             logger.log(
                 level = InternalLogger.Level.WARN,
                 target = InternalLogger.Target.MAINTAINER,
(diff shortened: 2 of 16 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -93,17 +96,51 @@ internal class Base64Serializer private constructor(
...
     ) {
-        val base64String = convertBmpToBase64(drawable, bitmap, shouldCacheBitmap)
-        finalizeRecordedDataItem(base64String, imageWireframe, base64SerializerCallback)
+        val byteArray = compressBitmap(bitmap)
+        val base64String = convertBmpToBase64(byteArray, drawable, bitmap, shouldCacheBitmap)
+        val hash = calculateHash(drawable, byteArray)
+        finalizeRecordedDataItem(base64String, hash, imageWireframe, base64SerializerCallback)
+    }
+
+    private fun calculateHash(
```
jonathanmos, line 105: There's already an SHA256 hash generator in the core, so it could be better to add an MD5 generator there and then use it here

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -93,17 +95,51 @@ internal class Base64Serializer private constructor(
...
+        }
+
+        return hashValue
     }
 
     @MainThread
-    private fun registerCacheForCallbacks(applicationContext: Context) {
-        if (isCacheRegisteredForCallbacks) return
+    private fun registerBase64LruCacheForCallbacks(applicationContext: Context) {
+        if (isBase64CacheRegisteredForCallbacks) return
```
mariusc83, line 120: why do we still keep/use the base64LruCache ? Can't we just keep a cache of the binary instead with a key which is the `resourceId` ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/MD5HashGenerator.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.security.MessageDigest
+import java.security.NoSuchAlgorithmException
+import java.util.Locale
+
+internal class MD5HashGenerator(
+    private val logger: InternalLogger
+) : HashGenerator {
+    override fun generate(input: ByteArray): String {
+        return try {
+            val messageDigest = MessageDigest.getInstance("MD5")
```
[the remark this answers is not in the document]
mariusc83: I think you can surpass this as we are not using this to secure any sort of information, it is only for generating an unique fingerprint from a binary

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -93,17 +95,51 @@ internal class Base64Serializer private constructor(
...
+        val base64String = convertBmpToBase64(byteArray, drawable, bitmap, shouldCacheBitmap)
+        val hash = calculateHash(drawable, byteArray)
+        finalizeRecordedDataItem(base64String, hash, imageWireframe, base64SerializerCallback)
+    }
+
+    private fun calculateHash(
+        drawable: Drawable,
+        byteArray: ByteArray
+    ): String {
+        val md5Generator = MD5HashGenerator(logger)
```
0xnm, line 108: Is usage of `MD5` a requirement by the backend?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/MD5HashGenerator.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import com.datadog.android.api.InternalLogger
+import java.security.MessageDigest
+import java.security.NoSuchAlgorithmException
+import java.util.Locale
+
+internal class MD5HashGenerator(
+    private val logger: InternalLogger
+) : HashGenerator {
+    override fun generate(input: ByteArray): String {
```
0xnm, line 17: I would propose to make return type as `String?`. Returning empty string makes a little sense semantically (as we have to check if it is not empty later), because proper hash can never be an empty string. I understand, that this is a fallback in case of an exception, but having `null` as a fallback looks more logical to me.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -93,17 +95,51 @@ internal class Base64Serializer private constructor(
...
+        val hashValue = md5Generator.generate(byteArray)
+
+        if (hashValue.isNotEmpty()) {
```
0xnm, line 111: having hash equal to empty string is something strange, it is better to make `generate` method return `null` in case of an error.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -93,17 +95,51 @@ internal class Base64Serializer private constructor(
...
-        if (isCacheRegisteredForCallbacks) return
+    private fun registerBase64LruCacheForCallbacks(applicationContext: Context) {
+        if (isBase64CacheRegisteredForCallbacks) return
 
         if (base64LRUCache is ComponentCallbacks2) {
             applicationContext.registerComponentCallbacks(base64LRUCache)
-            isCacheRegisteredForCallbacks = true
+            isBase64CacheRegisteredForCallbacks = true
+        } else {
+            // Temporarily use UNBOUND logger
```
0xnm, line 126: seems this comment is wrong, there is no `UNBOUND` logger here, and there is no need for `TODO` item.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -203,10 +244,16 @@ internal class Base64Serializer private constructor(
         imageWireframe: MobileSegment.Wireframe.ImageWireframe,
         base64SerializerCallback: Base64SerializerCallback?
     ): String? {
-        return base64LRUCache?.get(drawable)?.let { base64String ->
-            finalizeRecordedDataItem(base64String, imageWireframe, base64SerializerCallback)
-            base64String
+        val base64String = base64LRUCache?.get(drawable)
+        // get hash from cache
```
0xnm, line 248: I think line below is self-explanatory, no need for this comment.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
0xnm: even if it is needed, would it make sense to use a single cache for both base64 and hash? or it will make things more complicated?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -230,9 +277,14 @@ internal class Base64Serializer private constructor(
 
     private fun finalizeRecordedDataItem(
         base64String: String,
+        hash: String,
         wireframe: MobileSegment.Wireframe.ImageWireframe,
         base64SerializerCallback: Base64SerializerCallback?
     ) {
+        if (hash.isNotEmpty()) {
+            wireframe.resourceId = hash
```
0xnm, line 285: So what backend does if there is no `resourceId`, but other payload like `base64`, etc. is there. How `resourceId` is used?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/MD5HashGenerator.kt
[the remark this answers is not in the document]
jonathanmos: yes, I've dismissed this. It was causing a failure in the build steps

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: @mariusc83 we still send base64 and will only stop sending it when the rss endpoint is up, so I think we still need to cache the values to prevent unnecessary computations. 

@0xnm You mean a single cache that maps a drawable to a pair of base64 and md5? The issue that I see is that it couples the two hashes and forces us to change the cache implementation now, and then change it again when we want to remove b64 hash.

mariusc83: I would rather wait on this PR and remove base64 all together when we will be able to send binaries. What blocks you ?

jonathanmos: RSS endpoint is still a wip on the backend. On iOS they are already sending resourceId alongside b64 as preparation for the next stage, and this pr is mainly to align on that

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: afaik we can use any hash. MD5 just because it's fast, and is [what is being used on ios](https://github.com/DataDog/dd-sdk-ios/blob/da1f1b99257e36edc114964651fa4a33f7d35d03/DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift#L107)

0xnm: Ok good to know. My point was just that we already have SHA-256 generator and maybe it is worth to share it.

And we are using MD5 for non-cryptographic purpose, so the security alert is obviously a false positive.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: at the moment backend is using base64, and eventually will map rssId as the key to pull the binary for use in the replays. In the future when the rss endpoint is done, there will have to be some logic there for ending support for b64 and moving over entirely to rssId (we'll start to send b64 as an empty field in the wireframes and only populate rssId)

0xnm: So what will be our logic then if we cannot generate `resourceId`? We won't send a resource itself then? This 99.999% won't happen (all devices for sure can generate MD5 hash), but I'm just curious.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: We could use that, but I'm assuming our primary consideration is speed (and potential collisions) and md5 is almost certainly going to be faster and less computationally intensive

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: If we can't generate the hash then I think we'll need to throw away the binary because backend won't be able to use it in a replay. We can check for this in the future by having the rssId inside the recordedDataQueueItem holding the binary and marking it as invalid if the field isn't populated, then the dataQueueHandler can purge the record

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/MD5HashGenerator.kt
jonathanmos: np, I'll make this change

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
mariusc83: are we sure we are not going to do this PR for nothing now just to align with iOS ? In the end once the endpoint will be in place we will have to get back and remove this and re - factor. Why should we not wait ? How much of this PR work is going to help you when the endpoint will be ready ?

plousada: There's a benefit in already removing base64 from MutationResolver. We expect to have the same level of benefit in terms of diffing speed that iOS achieve. Remember, this GitHub [issue](https://github.com/DataDog/dd-sdk-ios/issues/1482) originally triggered the investigation. 

Yes we'll have to revisit it in the future, but we still don't know when. Likely only next year and on the mean time it's beneficial to have some performance improvements and alignment with iOS.

plousada (approved):

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCacheTest.kt
```diff
@@ -40,20 +41,30 @@ import org.mockito.quality.Strictness
...
+    @Mock
+    lateinit var mockInvocationUtils: InvocationUtils
+
     @StringForgery
     lateinit var fakeBase64: String
 
+    @StringForgery
+    lateinit var fakeResourceId: String
+
+    var fakeHashPair: Pair<String, String?> = Pair("", null)
```
0xnm, line 58:
```suggestion
    lateinit var fakeHashPair: Pair<String, String?>
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCacheTest.kt
```diff
@@ -40,20 +41,30 @@ import org.mockito.quality.Strictness
...
+
     val argumentCaptor = argumentCaptor<String>()
 
     @BeforeEach
     fun setup() {
-        internalCache = LruCache<String, ByteArray>(MAX_CACHE_MEMORY_SIZE_BYTES)
+        fakeHashPair = Pair(fakeBase64, fakeResourceId)
```
0xnm, line 64:
```suggestion
        fakeHashPair = Pair(fakeBase64, forge.aNullable { fakeResourceId })
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCacheTest.kt
```diff
@@ -140,13 +147,12 @@ internal class Base64LRUCacheTest {
     @Test
     fun `M not generate key prefix W put() { layerDrawable with only one layer }`(
         @Mock mockRippleDrawable: RippleDrawable,
-        @Mock mockBgLayer: Drawable,
-        @StringForgery fakeBase64: String
+        @Mock mockBgLayer: Drawable
```
0xnm, line 150: nit: better not to shorten

```suggestion
        @Mock mockBackgroundLayer: Drawable
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64SerializerTest.kt
```diff
@@ -71,11 +71,18 @@ internal class Base64SerializerTest {
...
+    lateinit var mockMD5HashGenerator: MD5HashGenerator
+
     @Mock
     lateinit var mockSerializerCallback: Base64SerializerCallback
 
     private lateinit var fakeBase64String: String
 
+    private lateinit var fakeHashString: String
+
+    private var fakeHashPair: Pair<String, String?> = Pair("", null)
```
0xnm, line 84:
```suggestion
    private lateinit var fakeHashPair: Pair<String, String?>
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64LRUCache.kt
```diff
@@ -19,16 +19,18 @@ import com.datadog.android.sessionreplay.internal.utils.CacheUtils
 import com.datadog.android.sessionreplay.internal.utils.InvocationUtils
 
 internal class Base64LRUCache(
-    private val cacheUtils: CacheUtils<String, ByteArray> = CacheUtils(),
+    private val cacheUtils: CacheUtils<String, Pair<ByteArray, ByteArray?>> = CacheUtils(),
```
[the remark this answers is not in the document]
jonathanmos: I'll create an internal class for this 👍

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -32,11 +32,12 @@ internal class Base64Serializer private constructor(
     private val drawableUtils: DrawableUtils,
     private val base64Utils: Base64Utils,
     private val webPImageCompression: ImageCompression,
-    private val base64LRUCache: Cache<Drawable, String>?,
+    private val base64LRUCache: Cache<Drawable, Pair<String, String?>>?,
```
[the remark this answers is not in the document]
jonathanmos: Good point. I'll do the same for the bitmapPool

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -139,45 +167,17 @@ internal class Base64Serializer private constructor(
         isBitmapPoolRegisteredForCallbacks = true
     }
 
+    private fun compressBitmap(bitmap: Bitmap): ByteArray {
```
0xnm, line 170: very minor: seems like there is no need for this method, all call sites can use `webPImageCompression.compressBitmap` directly.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -270,23 +273,38 @@ internal class Base64Serializer private constructor(
         imageWireframe: MobileSegment.Wireframe.ImageWireframe,
         base64SerializerCallback: Base64SerializerCallback
     ): String? {
-        return base64LRUCache?.get(drawable)?.let { base64String ->
-            finalizeRecordedDataItem(base64String, imageWireframe, base64SerializerCallback)
-            base64String
+        val encodings = base64LRUCache.get(drawable)
```
0xnm, line 276: `resourceId` cannot be classified as encoding, so better to use other name.

```suggestion
        val cacheData = base64LRUCache.get(drawable)
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64SerializerTest.kt
```diff
@@ -71,12 +71,19 @@ internal class Base64SerializerTest {
...
+    @Mock
+    lateinit var mockMD5HashGenerator: MD5HashGenerator
+
     @Mock
     lateinit var mockSerializerCallback: Base64SerializerCallback
 
-    private lateinit var fakeBase64String: String
+    private lateinit var fakeBase64Encoding: ByteArray
+
+    private var fakeResourceIdByteArray: ByteArray? = null
```
0xnm, line 82:
```suggestion
    private lateinit var fakeResourceIdByteArray: ByteArray?
```

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: 👍  I'll change in all relevant places

0xnm (approved):

PR #1694, opened 2023-11-13
RUM-1924: Use tracestate header to supply vendor-specific information

0xnm:
### What does this PR do?

This PR adds `tracestate` header with Datadog-specific information in case of `TRACECONTEXT` tracing. This header will contain the following Datadog-specific tags:

* `p` - parent ID
* `s` - sampling decision
* `o` - origin, if exists. In case of instrumentation with RUM, it will have `rum` value.

Full spec https://www.w3.org/TR/trace-context/#tracestate-header

iOS counterpart https://github.com/DataDog/dd-sdk-ios/pull/1536

### DIFF
--- a/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/W3CHttpCodec.java
+++ b/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/W3CHttpCodec.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import io.opentracing.SpanContext;
 import io.opentracing.propagation.TextMapExtract;
 import io.opentracing.propagation.TextMapInject;
+import kotlin.text.StringsKt;
 
 /**
  * A codec designed for HTTP transport via headers using W3C traceparent header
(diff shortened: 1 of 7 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -649,7 +661,13 @@ internal constructor(
 
         // taken from W3CHttpCodec
         internal const val W3C_TRACEPARENT_KEY = "traceparent"
-        internal const val W3C_DROP_SAMPLING_DECISION = "00-%s-%s-00"
+        internal const val W3C_TRACESTATE_KEY = "tracestate"
+
+        // https://www.w3.org/TR/trace-context/#traceparent-header
+        internal const val W3C_TRACEPARENT_DROP_SAMPLING_DECISION = "00-%s-%s-00"
```
mariusc83, line 667: these are very weird as they look similar with the `TRACE_PARENT_VALUE` declared in the class above but here the name says that is about sampling ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -649,7 +661,13 @@ internal constructor(
...
+        internal const val W3C_TRACESTATE_DROP_SAMPLING_DECISION = "dd=p:%s;s:0"
```
mariusc83, line 668: Also this one ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
0xnm: yes, because it is the same, but with sampling decision already included - which is a `drop`, meaning `00` in the end.

The class above it not public, it cannot be referenced/used from OkHttp package.

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/W3CHttpCodec.java
```diff
@@ -31,11 +32,21 @@
 class W3CHttpCodec {
 
   private static final String TRACEPARENT_KEY = "traceparent";
-  private static final String TRACEPARENT_VALUE = "00-0000000000000000%s-%s-0%s";
+  private static final String TRACESTATE_KEY = "tracestate";
+
+  private static final String TRACEPARENT_VALUE = "00-%s-%s-0%s";
+
+  private static final int TRACECONTEXT_PARENT_ID_LENGTH = 16;
```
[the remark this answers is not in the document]
0xnm: no, the length of the trace ID and span ID (or parent ID) is different per spec https://www.w3.org/TR/trace-context/#version-format

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
0xnm: same reason as above

mariusc83 (approved):

jonathanmos (approved):

xgouchet (approved):

PR #1717, opened 2023-11-15
(4/9) RUM-2127 Add synthetics info to RUM Views

xgouchet:
_This PR has been created automatically by the CI_

Add synthetics info to RUM Views.
The info is read from the RUM context and converted into the relevant attributes at write time.


features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -903,77 +907,6 @@ internal class RumViewScopeTest {
         assertThat(result2).isNull()
     }
 
-    @Test
-    fun `𝕄 send event 𝕎 handleEvent(StopView) on active view`(
```
0xnm, line 907: why do we remove this test? I can see similar test is added with synthetics session, but then shouldn't this test assert user session type then instead of removing it?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -697,6 +697,23 @@ internal open class RumViewScope(
                     currentViewId
                 )
                 val replayStats = ViewEvent.ReplayStats(recordsCount = sessionReplayRecordsCount)
+                val syntheticsAttribute = if (
```
mariusc83, line 700: maybe we should extract these methods ?

mariusc83 (approved):

0xnm (approved):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
xgouchet: That must have been a cut-paste instead of a copy paste, fixed

PR #1722, opened 2023-11-15
(9/9) RUM-2127 track synthetics info from activity extras

xgouchet:
_This PR has been created automatically by the CI_

track synthetics info from activity extras

All the provided ViewTrackingStrategy implementation extend the ActivityLifecycleTrackingStrategy, meaning that all supports the Synthetics use case.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityLifecycleTrackingStrategy.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityLifecycleTrackingStrategy.kt
@@ -15,7 +15,9 @@ import androidx.annotation.MainThread
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
 import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.rum.GlobalRumMonitor
 import com.datadog.android.rum.internal.RumFeature
+import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
 
 /**
  * The ActivityLifecycleTrackingStrategy as an [Application.ActivityLifecycleCallbacks]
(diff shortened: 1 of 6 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityLifecycleTrackingStrategy.kt
```diff
@@ -89,7 +91,17 @@ abstract class ActivityLifecycleTrackingStrategy :
 
     @MainThread
     override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
-        // No Op
+        val intent = activity.intent
+        val extras = intent.extras
+        val testId = extras?.getString("_dd.synthetics.test_id")
+        val resultId = extras?.getString("_dd.synthetics.result_id")
+        if (!testId.isNullOrBlank()) {
```
0xnm, line 98: why do we check only `testId` here, if later on in the attributes injection logic we check both `testId` and `resultId` to be able to add them? Would it make more more sense for the `SetSyntheticsTestAttribute` to have both properties as non-null then?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityLifecycleTrackingStrategy.kt
```diff
@@ -89,7 +91,17 @@ abstract class ActivityLifecycleTrackingStrategy :
 
     @MainThread
     override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
-        // No Op
+        val intent = activity.intent
+        val extras = intent.extras
+        val testId = extras?.getString("_dd.synthetics.test_id")
```
mariusc83, line 96: I am missing the context here...who's going to add these in the extras ? are we going to add this every time we run a synthetic test as runtime arguments ?

xgouchet: The synthetics team is the one adding those

0xnm (approved):

mariusc83 (approved):

PR #1724, opened 2023-11-15
Fix flaky test in WebViewRumEventConsumerTest

xgouchet:
### What does this PR do?

Fix a test that was randomly failing locally and in CI 

### Motivation

Test and CI reliability

### Additional Notes

The test relied on several threads sending events to the WebViewRumEventConsumerTest. But the assertions expected the events to be processed in the original order, which, on many occasion (10% to 25% of test runs) wasn't true, meaning the assertion would fail.


0xnm (approved):
thanks for taking care of this!

mariusc83 (approved):

PR #1732, opened 2023-11-17
Fix flaky test in BatchFileOrchestratorTest

xgouchet:
### What does this PR do?

Fix flaky test in BatchFileOrchestratorTest

The main issue here is that were using Kotlin's `mutableListOf()` which is inherently not thread safe, and try to insert elements from multiple concurrent threads. What happens 10% of the time (locally) is that two thread end up appending at the same time, and one of the value is overridden, resulting in one less value in the list than expected.


0xnm (approved):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestratorTest.kt
```diff
@@ -1207,15 +1207,17 @@ internal class BatchFileOrchestratorTest {
         repeat(repeatCount) {
             Thread {
                 val result = testedOrchestrator.getRootDir()
-                results.add(result)
+                synchronized(results) { results.add(result) }
```
mariusc83, line 1210: ohh nice catch...good to know

mariusc83 (approved):

PR #1736, opened 2023-11-21
RUM-2245 Generate wireframe ids as 32bit Int

jonathanmos:
### What does this PR do?
Generates wireframe IDs as `Int` instead of as `Long`. Also modify the forgery factories to do the same so that tests will reflect this change.

### Motivation
Generating the IDs as `Long` caused an issue on the backend because there the IDs are being parsed as `Int`. When the ID was out of bounds it would be rounded and this could cause problems with removals of wireframes or duplicated wireframes.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/utils/UniqueIdentifierGeneratorTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/utils/UniqueIdentifierGeneratorTest.kt
@@ -115,4 +115,23 @@ internal class UniqueIdentifierGeneratorTest {
 
         assertThat(generatedUniqueNumbers).isEmpty()
     }
+
+    @Test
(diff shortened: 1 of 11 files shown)


features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/utils/UniqueIdentifierGeneratorTest.kt
```diff
@@ -115,4 +115,23 @@ internal class UniqueIdentifierGeneratorTest {
...
+        // When
+        val results = forge.aList<View>(numberOfCalls) { mock() }
+            .map {
+                UniqueIdentifierGenerator.resolveChildUniqueIdentifier(mock(), forge.aString())
+            }
+
+        // Then
+        results.forEach {
+            if (it != null) {
+                assertThat(it in Int.MIN_VALUE..Int.MAX_VALUE).isTrue
```
0xnm, line 133: this will make failure message more descriptive

```suggestion
                assertThat(it).isBetween(Int.MIN_VALUE, Int.MAX_VALUE)
```

0xnm (approved):

PR #1742, opened 2023-11-22
RUM-1930 Add regression test for Gson toString method

mariusc83:
### What does this PR do?

As a followup on the previous action taken on iOS: https://datadoghq.atlassian.net/browse/RUM-1929, I investigated all the potential issues on our end regarding this topic and came to the following conclusions:

- we are using indeed `.` attributes on our end but most of them are correctly handled in our serializers. 
- it will be quite hard and very intrusive to try to remove all these attributes in the sdk and for no reason momentarily because the serializers are handling this. There is only one problem that could appear if by any chance we adopt a new `Gson` version which changes the implementation in adding the properties into the `toString` method. Right now they are using a `LinkedList` which suites us very well.

In this PR I added a regression test for our Gson dependency to make sure it will always add the properties in the order they were added in the Json object when using the `toString` method.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/serializer/GsonCompatibilityTests.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/serializer/GsonCompatibilityTests.kt
@@ -1,0 +1,72 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core.serializer
+
+import com.google.gson.JsonObject
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.RepeatedTest
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+
+/**
+ * This class is used to test the compatibility of the Gson serializer with our current `.`
+ * keys in the JSON payloads. This should make sure that Json properties are added into the
+ * `toString` representation in the order they were added in the JSON object.
+ */
+
+@Extensions(ExtendWith(ForgeExtension::class))
+internal class GsonCompatibilityTests {
+
+    @RepeatedTest(5)
+    fun `M add the properties in order in the string W toString()`(forge: Forge) {
+        val propertiesKeysValues = forge.aMap(size = forge.anInt(min = 5, max = 50)) {
+            val key = forge.aList(size = forge.anInt(min = 1, max = 10)) {
+                anAlphabeticalString()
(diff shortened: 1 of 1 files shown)


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/serializer/GsonCompatibilityTests.kt
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+@Extensions(ExtendWith(ForgeExtension::class), )
+internal class GsonCompatibilityTests {
+
+    @RepeatedTest(5)
+    fun `M add the properties in order in the string W toString()`(forge:Forge){
+        val propertiesKeysValues = forge.aMap(size = forge.anInt(min=5,max=50)){
+                val key = forge.aList(size = forge.anInt(min = 1, max = 10)) {
+                    anAlphabeticalString() }.joinToString(".")
```
0xnm, line 33: Do we need to have dots in keys? We simply test that keys are added in order, we don't care about keys format.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/serializer/GsonCompatibilityTests.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.jupiter.api.RepeatedTest
+import org.junit.jupiter.api.Test
+
+/**
+ * This class is used to test the compatibility of the Gson serializer with our current `.`
+ * keys in the JSON payloads. This should make sure that Json properties are added into the
+ * `toString` representation in the order they were added in the JSON object.
+ */
+
+@Extensions(ExtendWith(ForgeExtension::class), )
```
0xnm, line 26:
```suggestion
@Extensions(ExtendWith(ForgeExtension::class))
```

0xnm (approved):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/serializer/GsonCompatibilityTests.kt
mariusc83: yes but I wanted to just make sure it is robust enough. It doesn't hurt to have the dots there.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/serializer/GsonCompatibilityTests.kt
```diff
@@ -0,0 +1,73 @@
...
+            )
+            val value = forge.anElementFrom(values)
+
+            key to value
+        }
+        val expectedJsonRepresentation = buildString {
+            append("{")
+            propertiesKeysValues.entries.forEachIndexed { index, entry ->
+                when (entry.value) {
+                    is Int -> append("\"${entry.key}\":${entry.value}")
```
jonathanmos, line 47: Since Int and Boolean look like they have the same logic, would it make sense to do
```suggestion
                    is Int, is Boolean -> append("\"${entry.key}\":${entry.value}")
```
to avoid duplication?

PR #1746, opened 2023-11-27
(2/4) RUM-2129 let customer set custom persistence strategy in configuration

xgouchet:
_This PR has been created automatically by splitting a large branch_

## Pull request 2 out of 4

> The overall goal is to create a feature allowing customers to store unsent data any way they want (instead of using the official disk batch files)

This PR adds the necessary methods to set a custom `PersistenceStrategy.Factory` in the core configuration.

When set, this will be reported in the configuration telemetry, and
the persistenceStrategyFactory will be stored in the CoreFeature instance

### DIFF
--- a/dd-sdk-android-core/api/apiSurface
+++ b/dd-sdk-android-core/api/apiSurface
@@ -177,6 +177,7 @@ data class com.datadog.android.core.configuration.Configuration
     fun setAdditionalConfiguration(Map<String, Any>): Builder
     fun setProxy(java.net.Proxy, okhttp3.Authenticator?): Builder
     fun setEncryption(com.datadog.android.security.Encryption): Builder
+    fun setPersistenceStrategyFactory(com.datadog.android.core.persistence.PersistenceStrategy.Factory?): Builder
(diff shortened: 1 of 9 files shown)


dd-sdk-android-core/api/apiSurface
```diff
@@ -177,6 +177,7 @@ data class com.datadog.android.core.configuration.Configuration
     fun setAdditionalConfiguration(Map<String, Any>): Builder
     fun setProxy(java.net.Proxy, okhttp3.Authenticator?): Builder
     fun setEncryption(com.datadog.android.security.Encryption): Builder
+    fun setPersistenceStrategyFactory(com.datadog.android.core.persistence.PersistenceStrategy.Factory?): Builder
```
0xnm, line 180: Should we hide it behind some internal proxy? I was thinking that we are not going to show this method to everyone.

xgouchet: The thing is that we don't have an internal proxy for the configuration. The fact that the interface to implement is not trivial at all makes me think that no one (if they discover it) will bother implementing it

PR #1748, opened 2023-11-27
(4/4) RUM-2129 Use AbstractStorage when custom persistence strategy provided

xgouchet:
_This PR has been created automatically by splitting a large branch_

## Pull request 4 out of 4

> The overall goal is to create a feature allowing customers to store unsent data any way they want (instead of using the official disk batch files)

This final PR updates the SDK Feature initialisation by choosing the default storage or the new `AbstractStorage` one depending on the provided `PersistenceStrategy.Factory`

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
@@ -31,6 +31,7 @@ import com.datadog.android.core.internal.lifecycle.ProcessLifecycleMonitor
 import com.datadog.android.core.internal.metrics.BatchMetricsDispatcher
 import com.datadog.android.core.internal.metrics.MetricsDispatcher
 import com.datadog.android.core.internal.metrics.NoOpMetricsDispatcher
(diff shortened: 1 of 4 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -242,7 +239,47 @@ internal class SdkFeature(
...
+        }
+    }
+
+    private fun createCustomStorage(
+        featureName: String,
+        storageConfiguration: FeatureStorageConfiguration,
+        persistenceStrategyFactory: PersistenceStrategy.Factory
+    ): Storage {
+        return AbstractStorage(
+            coreFeature.variant,
```
0xnm, line 272: it seems it should be `sdkCoreId` here rather than `coreFeature.variant`?

xgouchet: Good catch indeed

jonathanmos (approved):

0xnm (approved):

PR #1754, opened 2023-12-06
Remove resourceId field from e2e payloads

jonathanmos:
### What does this PR do?
Adds resourceIds to the e2e payloads

### Motivation
We now send resourceIds as part of the ImageWireframes

### Additional Notes

### DIFF
--- a/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/sessionreplay/BaseSessionReplayTest.kt
+++ b/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/sessionreplay/BaseSessionReplayTest.kt
@@ -241,7 +241,7 @@ internal abstract class BaseSessionReplayTest<R : Activity> {
         // will be executed in Bitrise and currently Bitrise does not own a specific model for the
         // API 33. They only have a standard emulator for this API with the required screen size and
         // X,Y positions are different from the ones we have in our local emulator.
-        // Also the base64 encoded images values are inconsistent from one run to another so will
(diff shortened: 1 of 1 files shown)


0xnm:
I'm not sure this is the right approach, because when we currently validating data we drop `base64` property [here](https://github.com/DataDog/dd-sdk-android/blob/f38f6beabc2aef5355ae68f89efa277aa359676b/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/sessionreplay/BaseSessionReplayTest.kt#L256). `resourceId` property also relies on the drawable content, the same way as `base64` (essentially they are using the same input, correct me if I'm wrong), so since input for `base64` is considered as inconsistent, we should also remove `resourceId` from the validation instead of adding it to the `json`. WDYT?

jonathanmos:
> I'm not sure this is the right approach, because when we currently validate data we drop `base64` property [here](https://github.com/DataDog/dd-sdk-android/blob/f38f6beabc2aef5355ae68f89efa277aa359676b/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/sessionreplay/BaseSessionReplayTest.kt#L256). `resourceId` property also relies on the drawable content, the same way as `base64` (essentially they are using the same input, correct me if I'm wrong), so since input for `base64` is considered as inconsistent, we should also remove `resourceId` from the validation instead of adding it to the `json`. WDYT?

Yes, I recall now. Marius had failures on CI because of the variability - good catch. I've changed to remove the resourceId field

0xnm (approved):

PR #1755, opened 2023-12-08
RUM-1336 Add session start reason to events

xgouchet:
### What does this PR do?

Adds a `session_precondition` attribute to all RUM event to track the reason the session was created.

### Motivation

Get more information on what leads to a new session starting, as part of our Session Management work


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumContext.kt
```diff
@@ -58,8 +60,12 @@ internal data class RumContext(
         fun fromFeatureContext(featureContext: Map<String, Any?>): RumContext {
             val applicationId = featureContext[APPLICATION_ID] as? String
             val sessionId = featureContext[SESSION_ID] as? String
-            val sessionState =
-                RumSessionScope.State.fromString(featureContext[SESSION_STATE] as? String)
+            val sessionState = RumSessionScope.State.fromString(
+                featureContext[SESSION_STATE] as? String
+            )
+            val sessionStartReason = RumSessionScope.StartReason.fromString(
```
0xnm, line 66: we need to update `toMap` method then as well, to write it to the context.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -149,25 +169,33 @@ internal class RumSessionScope(
...
+                } else {
+                    StartReason.MAX_DURATION
+                }
+                renewSession(nanoTime, reason)
             }
             lastUserInteractionNs.set(nanoTime)
         } else if (isExpired) {
             if (backgroundTrackingEnabled && isBackgroundEvent) {
-                renewSession(nanoTime)
+                renewSession(nanoTime, StartReason.INACTIVITY_TIMEOUT)
```
0xnm, line 184: I think here we can qualify it as `background_launch`, maybe? If I'm not wrong, we can hit this line in case if application/session was started by some background event (but I may have outdated memories of this logic).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumContext.kt
xgouchet: Good catch

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
xgouchet: As per the [ticket](https://datadoghq.atlassian.net/browse/RUM-1336): 

> [!Note]
> The following will not be implemented yet because we don’t track non-interactive sessions.

0xnm (approved):

jonathanmos (approved):

PR #1756, opened 2023-12-08
RUM-1823: Add buildId to the RUM error and Log events

0xnm:
### What does this PR do?

This PR adds `buildId` support for RUM error events and Log events.

`buildId` will be generated by the Datadog Gradle Plugin (see https://github.com/DataDog/dd-sdk-android-gradle-plugin/pull/211) only for builds which are minified.

Once attached to RUM error events and Logs events, it will be used for the symbolication.

**NB**: although `build_id` is declared in the common schema for RUM events, this PR adds it only for error events.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -332,6 +334,21 @@ internal class CoreFeature(
         serviceName = configuration.service ?: appContext.packageName
         envName = configuration.env
         variant = configuration.variant
+        appBuildId = with(appContext.assets) {
+            try {
+                open(BUILD_ID_FILE_NAME).bufferedReader().use {
+                    it.readText().trim()
+                }
+            } catch (@Suppress("SwallowedException") e: IOException) {
```
xgouchet, line 342: > [!NOTE]
> We might want to distinguish `FileNotFoundException` from other kinds of `IOException`
> The first might mean that the app was built with an older plugin (or no plugin at all) in which case 
> it's expected to not have a Build ID File. In that case an `INFO` log is indeed enough.
> The second means an actual IO error happened with a file that exists, and this should report at least to `WARNING` or `ERROR` (and maybe be sent to telemetry).

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -464,6 +473,44 @@ internal class CoreFeatureTest {
...
+        // Then
+        assertThat(testedFeature.appBuildId).isEqualTo(fakeBuildId.toString())
+    }
+
+    @Test
+    fun `𝕄 initializes build ID 𝕎 initialize() { build ID is missing }`() {
+        // Given
+        whenever(
+            appContext.mockInstance.assets.open(CoreFeature.BUILD_ID_FILE_NAME)
+        ) doThrow IOException()
```
xgouchet, line 495: > [!NOTE]
> This should be a `FileNotFoundException` to properly mock an asset not being present.

xgouchet: You can add other tests for other scenarios: 
- `RuntimeException` could be thrown if the `AssetManager` has been closed (cf: `AssetManager.ensureOpenLocked()`
- `IOException` if the `read()` operation fails

mariusc83 (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
0xnm: Indeed, good point. Makes sense to distinguish between the case when file is not there and when something went wrong during the file read.

xgouchet (approved):

xgouchet (approved):

PR #1760, opened 2023-12-11
RUM-2441 Print RUM app, session and view ID in LogCat

xgouchet:
### What does this PR do?

Print the RUM app, session and view ID in LogCat

### Motivation

Let the Synthetics runner know what those ids are to create a link from a test run to the linked RUM events


0xnm (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -77,6 +78,7 @@ internal class RumApplicationScope(
                 syntheticsTestId = event.testId,
                 syntheticsResultId = event.resultId
             )
+            Log.i(RumScope.SYNTHETICS_LOGCAT_TAG, "_dd.application.id=${rumContext.applicationId}")
```
mariusc83, line 81: Is this safe to expose it in the Logcat ? The session Id and view Id is something we generated but the `applicationId` is the actual RUM application ID.

xgouchet: It's only exposed when being ran via Synthetics. 
Also the RUM Application ID is an identifier, not a token or a key. Just having the App id gives no permission by itself

jonathanmos (approved):

mariusc83 (approved):

PR #1761, opened 2023-12-13
RUM-2376: Fix duplicate wireframes issue

jonathanmos:
### What does this PR do?
This pr fixes the issue of duplicate wireframes appearing in the replays, and also adds in telemetry so that we'll know if items are dropped in future from the `RecordedDataQueue`. 

What was happening was that `RecordedDataQueueRefs` was held as a class-level variable inside `RecordedDataQueueHandler`. As a result, for every traversal the value of the `SnapshotRecordedDataQueueItem` held by the ref object was mutating, and this caused asynchronous tasks to occasionally return their callback to the wrong `SnapshotRecordedDataQueueItem`. This further led to the number of pending asynchronous tasks on the previous item never reaching zero, so it became stuck in the queue and eventually cleaned after exceeding the 200ms timeout. 

### Motivation
To handle the issue of duplicate wireframes. 

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandlerTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandlerTest.kt
@@ -8,6 +8,7 @@ package com.datadog.android.sessionreplay.internal.async
 
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.forge.ForgeConfigurator
+import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueHandler.Companion.ITEM_DROPPED_FROM_QUEUE_ERROR_MESSAGE
 import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueHandler.Companion.MAX_DELAY_MS
 import com.datadog.android.sessionreplay.internal.processor.RecordedDataProcessor
 import com.datadog.android.sessionreplay.internal.processor.RecordedQueuedItemContext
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandlerTest.kt
```diff
@@ -359,6 +380,16 @@ internal class RecordedDataQueueHandlerTest {
 
         // Then
         assertThat(testedHandler.recordedDataQueue.size).isEqualTo(0)
```
0xnm, line 382: side note: we can use this PR to change `assertThat(testedHandler.recordedDataQueue.size).isEqualTo(0)` assertions in this file to `assertThat(testedHandler.recordedDataQueue).isEmpty`

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/listener/WindowsOnDrawListenerTest.kt
```diff
@@ -102,12 +102,12 @@ internal class WindowsOnDrawListenerTest {
...
-        fakeMockedDecorViews.forEachIndexed { index, decorView ->
+        fakeMockedDecorViews.forEachIndexed { index, _ ->
             whenever(
                 mockSnapshotProducer.produce(
-                    decorView,
-                    fakeSystemInformation,
-                    mockRecordedDataQueueRefs
+                    any(),
+                    any(),
+                    any()
```
0xnm, line 110: I would still keep the assertion of the arguments here:

```suggestion
                    eq(decorView),
                    eq(fakeSystemInformation),
                    any()
```

0xnm (approved):

xgouchet (approved):

PR #1764, opened 2023-12-14
RUM-2467: Fix `ConcurrentModificationException` during `ConsentAwareStorage.dropAll` call

0xnm:
### What does this PR do?

We were removing items from collection during its traversal, this PR fixes that.

The issue doesn't reproduce with only single item in the collection, because modification check it done on the next iteration, so I also updated test to have multiple locked batches.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
```diff
@@ -149,8 +149,8 @@ internal class ConsentAwareStorage(
         synchronized(lockedBatches) {
             lockedBatches.forEach {
                 deleteBatch(it, RemovalReason.Flushed)
-                lockedBatches.remove(it)
             }
+            lockedBatches.clear()
```
mariusc83, line 153: good catch

mariusc83 (approved):

Issue #1762, opened 2023-12-14
RumViewScope creates memory leaks


pyricau:
**Describe what happened**

RumViewScope prevents destroyed activities from being garbage collected, causing an increase in memory pressure.

This leak was hard to find: the destroyed activities were not being garbage collected, however every time myself or LeakCanary would look at a heap dump, they were always weakly reachable. I ended up tweaking the shortest path algorithm that LeakCanary uses to include weak references that aren't set by the Android Framework.

I eventually figured out that the culprit is the way the SDK abuses weak references and revives the references by frequently invoking `WeakReference.get()` and recreating temporary strong local references to the destroyed activities.

**Leak Trace:**

```
┬───
│ GC Root: Global variable in native code
│
├─ dalvik.system.PathClassLoader instance
│    Leaking: NO (GlobalRumMonitor↓ is not leaking and A ClassLoader is never leaking)
│    ↓ ClassLoader.runtimeInternalObjects
├─ java.lang.Object[] array
│    Leaking: NO (GlobalRumMonitor↓ is not leaking)
│    ↓ Object[23579]
├─ com.datadog.android.rum.GlobalRumMonitor class
│    Leaking: NO (a class is never leaking)
│    ↓ static GlobalRumMonitor.registeredMonitors
│                              ~~~~~~~~~~~~~~~~~~
├─ java.util.LinkedHashMap instance
│    ↓ LinkedHashMap[instance @378137616 of com.datadog.android.core.DatadogCore]
│                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
├─ com.datadog.android.rum.internal.monitor.DatadogRumMonitor instance
│    ↓ DatadogRumMonitor.rootScope
│                        ~~~~~~~~~
├─ com.datadog.android.rum.internal.domain.scope.RumApplicationScope instance
│    ↓ RumApplicationScope.childScopes
│                          ~~~~~~~~~~~
├─ java.util.ArrayList instance
│    ↓ ArrayList[0]
│               ~~~
├─ com.datadog.android.rum.internal.domain.scope.RumSessionScope instance
│    ↓ RumSessionScope.childScope
│                      ~~~~~~~~~~
├─ com.datadog.android.rum.internal.domain.scope.RumViewManagerScope instance
│    ↓ RumViewManagerScope.childrenScopes
│                          ~~~~~~~~~~~~~~
[...]

PR #1767, opened 2023-12-14
Fix WeakReference leak

tcmulcahy:
See https://github.com/DataDog/dd-sdk-android/issues/1762 for details. According to https://developer.android.com/reference/java/lang/ref/Reference#refersTo(T) we should prefer using refersTo over a comparison with the result of get. refersTo avoids creating a strong ref even temporarily, thereby avoiding this leak.


tcmulcahy:
Closing because refersTo requires API 33+

PR #1769, opened 2023-12-15
RUM-1520 Add all Logs Feature integration tests

xgouchet:
### What does this PR do?

Add integration tests covering all the Logs Feature apis

### Motivation

Reliability, reliability, reliability


reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubSDKCore.kt
```diff
@@ -91,6 +91,9 @@ class StubSDKCore(
     override val internalLogger: InternalLogger = StubInternalLogger()
 
     override fun registerFeature(feature: Feature) {
+        // Stop previous registered
+        featureScopes[feature.name]?.unwrap<Feature>()?.onStop()
```
0xnm, line 95: shouldn't we just stop core instead after each test? and stop the features in the `stop` method of the code.

0xnm (approved):

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubSDKCore.kt
xgouchet: This part is only required because in one of the tests I override the previously started feature: in the `@BeforeEach` set up, we register a Logs feature, but in some tests I need to register a feature with a different configuration.

PR #1776, opened 2023-12-19
RUM-2566: Convert pending resource to pending error when Resource scope completes with an error

0xnm:
### What does this PR do?

There is a following bug: we increate `RumViewScope#pendingResourceCount` when we create resource scope, but resource scope can be finalized with 2 outcomes: either resource event or error event. In latter case we should instead decrease `pendingResourceCount` and increase `pendingErrorCount`, because `ResourceSent` / `ResourceDropped` will never be sent and we get instead `ErrorSent` / `ErrorDropped`.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -591,6 +591,14 @@ internal open class RumViewScope(
             val entry = iterator.next()
             val scope = entry.value.handleEvent(event, writer)
             if (scope == null) {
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -591,6 +591,14 @@ internal open class RumViewScope(
             val entry = iterator.next()
             val scope = entry.value.handleEvent(event, writer)
             if (scope == null) {
+                // if we finalized this scope and it was by error, we won't have resource
```
mariusc83, line 594: good catch

mariusc83 (approved):

jonathanmos (approved):

xgouchet (approved):
Nice job

PR #1777, opened 2023-12-19
Enable recording session if first RUM message happened before init

louiszawadzki:
### What does this PR do?

This fixes a race condition that was quite present in RN - see description below.

The solution I choose was to move the responsibility of checking for initialization from `startRecording` into `checkStatusAndApplySample`. 
The `startRecording` method is easier to work with if it just performs an atomic action (like `stopRecording`).

It seems fit to put the check in `checkStatusAndApplySample` since it's the method that registers the `currentRumSessionId` that was creating the lock before.

#### Description of the race condition
This can happen in any Android app, it’s just that conditions for creating the race condition are more present in RN since we collect JS framerate through RUM events on every frame.

The race condition happens in SessionReplayFeature.kt, in the onInitialize function:

```kotlin
        // first line making the race condition possible
        sdkCore.setEventReceiver(SESSION_REPLAY_FEATURE_NAME, this)
        dataWriter = createDataWriter()
        sessionReplayRecorder = sessionReplayRecorderProvider(dataWriter, appContext)
        @Suppress("ThreadSafety") // TODO REPLAY-1861 can be called from any thread
        sessionReplayRecorder.registerCallbacks()
        initialized.set(true) 
        // end of race condition
```

This is called when we call SessionReplay.enable.

By setting the event receiver, we listen on all RUM events. Now if a RUM event happens before we call initialized.set(true), here is what is going to happen in SessionReplayFeature:

- [onReceive](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt#L136) is called
- onReceive [calls handleRumSession](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt#L146)
- handleRumSession [calls checkStatusAndApplySample](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt#L157)
- checkStatusAndApplySample:
  - [calls startRecording](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt#L191) since we haven’t started recording
    - startRecording will [immediately return](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt#L208) since we haven’t called initialized.set(true)
  - calls currentRumSessionId.set(sessionId) [at the end of the ](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt#L201)function to mark the session as started

So far the Session Replay is not recording, and on the next RUM event, even if initialized.set(true) has been called:

- onReceive is called
- onReceive calls handleRumSession
- handleRumSession calls checkStatusAndApplySample
- [checkStatusAndApplySample returns before starting the replay](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt#L185C13-L185C51) since currentRumSessionId.get() == sessionId

So we fail to ever initialize the SDK.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
@@ -222,7 +222,8 @@ internal class SessionReplayFeature(
      * Resumes the replay recorder.
      */
     internal fun startRecording() {
-        if (!isRecording.getAndSet(true)) {
+        // Check initialization again so we don't forget to do it when this method is made public
+        if (checkIfInitialized() && !isRecording.getAndSet(true)) {
             @Suppress("ThreadSafety") // TODO REPLAY-1861 can be called from any thread
             sessionReplayRecorder.resumeRecorders()
         }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
@@ -188,24 +188,11 @@ internal class SessionReplayFeature(
             return
         }
 
-        val recordNewSession = keepSession && rateBasedSampler.sample()
-
-        if (!initialized.get()) {
-            sdkCore.internalLogger.log(
-                InternalLogger.Level.WARN,
-                InternalLogger.Target.USER,
-                {
-                    if (recordNewSession) {
-                        CANNOT_START_RECORDING_NOT_INITIALIZED
-                    } else {
-                        CANNOT_STOP_RECORDING_NOT_INITIALIZED
-                    }
-                }
-            )
+        if (!checkIfInitialized()) {
             return
         }
 
-        if (recordNewSession) {
+        if (keepSession && rateBasedSampler.sample()) {
             startRecording()
         } else {
             sdkCore.internalLogger.log(
@@ -217,6 +204,18 @@ internal class SessionReplayFeature(
         }
 
         currentRumSessionId.set(sessionId)
+    }
+
+    private fun checkIfInitialized(): Boolean {
+        if (!initialized.get()) {
+            sdkCore.internalLogger.log(
+                InternalLogger.Level.WARN,
+                InternalLogger.Target.USER,
+                { CANNOT_START_RECORDING_NOT_INITIALIZED }
+            )
+            return false
+        }
+        return true
     }
 
     /**
@@ -276,8 +275,6 @@ internal class SessionReplayFeature(
             " are either missing or have wrong type."
         internal const val CANNOT_START_RECORDING_NOT_INITIALIZED =
             "Cannot start session recording, because Session Replay feature is not initialized."
-        internal const val CANNOT_STOP_RECORDING_NOT_INITIALIZED =
-            "Cannot stop session recording, because Session Replay feature is not initialized."
         const val SESSION_REPLAY_FEATURE_NAME = "session-replay"
         const val SESSION_REPLAY_BUS_MESSAGE_TYPE_KEY = "type"
         const val RUM_SESSION_RENEWED_BUS_MESSAGE = "rum_session_renewed"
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -187,7 +187,24 @@ internal class SessionReplayFeature(
...
+                InternalLogger.Level.WARN,
+                InternalLogger.Target.USER,
+                {
+                    if (recordNewSession) {
+                        CANNOT_START_RECORDING_NOT_INITIALIZED
+                    } else {
+                        CANNOT_STOP_RECORDING_NOT_INITIALIZED
+                    }
+                }
+            )
```
louiszawadzki, line 203: Here I can use a single message (not making a difference between start and stop) or move this to a dedicated method if this is too verbose for this function.

louiszawadzki: If we use a single message I can even move this check above in the function

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeatureTest.kt
```diff
@@ -323,20 +323,6 @@ internal class SessionReplayFeatureTest {
...
-
-        // Then
-        mockInternalLogger.verifyLog(
-            InternalLogger.Level.WARN,
-            InternalLogger.Target.USER,
-            SessionReplayFeature.CANNOT_START_RECORDING_NOT_INITIALIZED
-        )
-        verifyNoInteractions(mockRecorder)
-    }
-
```
louiszawadzki, line 339: Now that the `startRecording` function does not perform the check, this test does not make sense anymore. 
It is replaced by the other tests below.

louiszawadzki:
Other solutions I had:

### Moving `sdkCore.setEventReceiver(SESSION_REPLAY_FEATURE_NAME, this)` call after `initialized.set(true)` in `onInitialize`

This alone does not fix the underlying issue of the lock happening because of the session Id being set. This might be a second step we can take, but to me it makes sense to have `initialized.set(true)` at the very end.

### Returning whether the recording started from `startRecording`

In this solution we lose the early return syntax and it makes both functions harder to read in my opinion.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -187,7 +188,24 @@ internal class SessionReplayFeature(
             return
         }
 
-        if (keepSession && rateBasedSampler.sample()) {
+        val recordNewSession = keepSession && rateBasedSampler.sample()
+
+        if (!initialized.get()) {
```
mariusc83, line 193: I think the fix is good but in order not to duplicate logic I would extract this part of code in a:
`checkIfInitialized()` function and use it also in the `startRecording` function. The reason why we need this check also in the `startRecording` is that this method will be called later from the manual API that we will expose to the users.

louiszawadzki: That makes sense, thanks for clarifying!

I'm thinking we might want to do other checks in the manual API as well (like checking the current session is not already tracked, maybe sampling?) and thus not expose `startRecording` directly. I'm usually not a big fan of adding code "because we'll need it later" because requirements might change by then and we might forget why we added it in the first place.
I'd be in favour of keeping `startRecording` as it is right now, but I have no strong opinion here, I can also add the check now if that's your preference.

mariusc83: I would keep it like it is now, the only thing we need is to extract that code in a new function.

mariusc83 (approved):

jonathanmos (approved):

PR #1793, opened 2023-12-28
RUM-2487 Provide the WebViewRecordConsumer for the JS bridge

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-webview/api/apiSurface
+++ b/features/dd-sdk-android-webview/api/apiSurface
@@ -1,5 +1,5 @@
 object com.datadog.android.webview.WebViewTracking
   fun enable(android.webkit.WebView, List<String>, Float = 100f, com.datadog.android.api.SdkCore = Datadog.getInstance())
   class _InternalWebViewProxy
-    constructor(com.datadog.android.api.SdkCore)
+    constructor(com.datadog.android.api.SdkCore, String? = null)
     fun consumeWebviewEvent(String)
--- a/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
+++ b/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
@@ -22,6 +22,10 @@ import com.datadog.android.webview.internal.NoOpWebViewEventConsumer
 import com.datadog.android.webview.internal.WebViewEventConsumer
 import com.datadog.android.webview.internal.log.WebViewLogEventConsumer
 import com.datadog.android.webview.internal.log.WebViewLogsFeature
+import com.datadog.android.webview.internal.replay.WebViewReplayEventConsumer
+import com.datadog.android.webview.internal.replay.WebViewReplayEventMapper
+import com.datadog.android.webview.internal.replay.WebViewReplayFeature
+import com.datadog.android.webview.internal.rum.TimestampOffsetProvider
 import com.datadog.android.webview.internal.rum.WebViewRumEventConsumer
 import com.datadog.android.webview.internal.rum.WebViewRumEventContextProvider
 import com.datadog.android.webview.internal.rum.WebViewRumFeature
@@ -76,7 +80,11 @@ object WebViewTracking {
         val featureContext = sdkCore.getFeatureContext(Feature.SESSION_REPLAY_FEATURE_NAME)
         val privacyLevel = (featureContext[SESSION_REPLAY_PRIVACY_KEY] as? String)
             ?: SESSION_REPLAY_MASK_ALL_PRIVACY
-        val webViewEventConsumer = buildWebViewEventConsumer(featureSdkCore, logsSampleRate)
+        val webViewEventConsumer = buildWebViewEventConsumer(
+            featureSdkCore,
+            logsSampleRate,
+            System.identityHashCode(webView).toString()
+        )
         webView.addJavascriptInterface(
             DatadogEventBridge(webViewEventConsumer, allowedHosts, privacyLevel),
             DATADOG_EVENT_BRIDGE_NAME
@@ -85,48 +93,38 @@ object WebViewTracking {
 
     private fun buildWebViewEventConsumer(
         sdkCore: FeatureSdkCore,
-        logsSampleRate: Float
+        logsSampleRate: Float,
+        webViewId: String?
     ): WebViewEventConsumer<String> {
-        val rumFeature = sdkCore.getFeature(Feature.RUM_FEATURE_NAME)
-            ?.unwrap<StorageBackedFeature>()
-        val logsFeature = sdkCore.getFeature(Feature.LOGS_FEATURE_NAME)
-            ?.unwrap<StorageBackedFeature>()
-
-        val webViewRumFeature = if (rumFeature != null) {
-            WebViewRumFeature(sdkCore, rumFeature.requestFactory)
-                .apply { sdkCore.registerFeature(this) }
-        } else {
-            sdkCore.internalLogger.log(
-                InternalLogger.Level.INFO,
-                InternalLogger.Target.USER,
-                { RUM_FEATURE_MISSING_INFO }
-            )
-            null
-        }
-
-        val webViewLogsFeature = if (logsFeature != null) {
-            WebViewLogsFeature(sdkCore, logsFeature.requestFactory)
-                .apply { sdkCore.registerFeature(this) }
-        } else {
-            sdkCore.internalLogger.log(
-                InternalLogger.Level.INFO,
-                InternalLogger.Target.USER,
-                { LOGS_FEATURE_MISSING_INFO }
-            )
-            null
-        }
-
-        val contextProvider = WebViewRumEventContextProvider(sdkCore.internalLogger)
-
+        val webViewRumFeature = resolveRumFeature(sdkCore)
+        val webViewLogsFeature = resolveLogsFeature(sdkCore)
+        val webViewReplayFeature = resolveReplayFeature(sdkCore)
         if (webViewLogsFeature == null && webViewRumFeature == null) {
             return NoOpWebViewEventConsumer()
         } else {
+            // it is very important that the timestamp offset provider is shared between the
+            // different consumers, otherwise we might end up with different offsets for the replay
+            // and rum browser events for the same view id.
+            val timestampOffsetProvider = TimestampOffsetProvider(sdkCore.internalLogger)
+            val contextProvider = WebViewRumEventContextProvider(sdkCore.internalLogger)
             return MixedWebViewEventConsumer(
                 WebViewRumEventConsumer(
                     sdkCore = sdkCore,
+                    offsetProvider = timestampOffsetProvider,
                     dataWriter = webViewRumFeature?.dataWriter ?: NoOpDataWriter(),
                     contextProvider = contextProvider
                 ),
+                webViewId?.let {
+                    WebViewReplayEventConsumer(
+                        sdkCore = sdkCore,
+                        dataWriter = webViewReplayFeature?.dataWriter ?: NoOpDataWriter(),
+                        contextProvider = contextProvider,
+                        webViewReplayEventMapper = WebViewReplayEventMapper(
+                            it,
+                            timestampOffsetProvider
+                        )
+                    )
+                } ?: NoOpWebViewEventConsumer(),
                 WebViewLogEventConsumer(
                     sdkCore = sdkCore,
                     userLogsWriter = webViewLogsFeature?.dataWriter ?: NoOpDataWriter(),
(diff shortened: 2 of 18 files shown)


features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventMapper.kt
```diff
@@ -0,0 +1,46 @@
...
+        ClassCastException::class,
+        IllegalStateException::class,
+        NumberFormatException::class
+    )
+    fun mapEvent(
+        event: JsonObject,
+        rumContext: RumContext,
+        timeOffset: Long
+    ): JsonObject {
+        event.get("event")?.asJsonObject?.let { record ->
```
mariusc83, line 27: @reviewers should we add here some telemetry ? it might be overkill but just wondering ?

features/dd-sdk-android-webview/api/apiSurface
```diff
@@ -1,5 +1,5 @@
 object com.datadog.android.webview.WebViewTracking
   fun enable(android.webkit.WebView, List<String>, Float = 100f, com.datadog.android.api.SdkCore = Datadog.getInstance())
   class _InternalWebViewProxy
-    constructor(com.datadog.android.api.SdkCore)
```
0xnm, line 4: even though it is for the internal usage, we need to check that both Flutter and RN:

* are able to provide `WebView` instance
* are not using dynamic versioning for native SDK dependencies (otherwise existing releases of RN/Flutter will be broken)

If there is a blocker for that change, we need to have 2 constructors then: with and without `WebView`.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
```diff
@@ -121,12 +125,25 @@ object WebViewTracking {
         if (webViewLogsFeature == null && webViewRumFeature == null) {
             return NoOpWebViewEventConsumer()
         } else {
+            // it is very important that the timestamp offset provider is shared between the
+            // different consumers, otherwise we might end up with different offsets for the replay
+            // and rum browser events for the same view id.
+            val timestampOffsetProvider = TimestampOffsetProvider(sdkCore.internalLogger)
```
0xnm, line 131: we should also probably apply it to the `WebViewLogEventConsumer`? we do time correction there.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventConsumer.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    private fun sendEventThroughBus(
+        datadogContext: DatadogContext,
+        event: JsonObject,
+        sessionReplayFeatureScope: FeatureScope
+    ) {
+        val rumContext = contextProvider.getRumContext(datadogContext)
+        if (rumContext != null && rumContext.sessionState == SESSION_TRACKED_STATE) {
+            map(event, datadogContext, rumContext)?.let { mappedEvent ->
+                sessionReplayFeatureScope.sendEvent(
```
0xnm, line 49: why approach here is different from webview tracking of Logs/RUM? Also it seems that nothing is going to respond to this event so far?

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventConsumer.kt
```diff
@@ -0,0 +1,123 @@
...
+            sendEventThroughBus(datadogContext, event, sessionReplayFeatureScope)
+        }
+    }
+
+    private fun sendEventThroughBus(
+        datadogContext: DatadogContext,
+        event: JsonObject,
+        sessionReplayFeatureScope: FeatureScope
+    ) {
+        val rumContext = contextProvider.getRumContext(datadogContext)
```
0xnm, line 46: since we are sending the message through the message bus, wouldn't it be simpler than to do the mapping inside `SessionReplayFeature`? as a benefit there we can get `DatadogContext` directly from `withWriteContext`.

features/dd-sdk-android-webview/api/apiSurface
mariusc83: yes...but for SR to work we need that unique id for the WebView, this is a very important part in the logic. I need to discuss with @louiszawadzki and @fuzzybinary about this.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
mariusc83: yes we do but we do not have the same constraints and no `viewId` involved.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventConsumer.kt
mariusc83: the logic to respond will come in the next PRs. The reason why we need to send everything through the bridge is that all events needs to be consumed by the `RecordedDataProcessor` and handled into the queue. This way we are making sure that we are not recording when RUM session is sampled out or SR is dropped because of SR sampling. Also later on we will have a START/STOP replay manual API which will make even more necessary that all SR should be consumed through the same pipeline. You will see it better in the next PRs.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventConsumer.kt
mariusc83: nope because the interest is to add this context as fast as possible. By the time it reaches the `SessionReplayFeature` the context might already change. Also is following the same approach as for RUM events.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventConsumer.kt
mariusc83: I am actually re - thinking this now, it might make sense what you're asking here. I am summoning a short brainstorming session with you and @xgouchet

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
xgouchet: > and no viewId involved.
Isn't there a bundled RUM info in the logs in the browser SDK ?

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
```diff
@@ -76,23 +79,58 @@ object WebViewTracking {
...
     private fun buildWebViewEventConsumer(
+        webView: WebView,
         sdkCore: FeatureSdkCore,
         logsSampleRate: Float
     ): WebViewEventConsumer<String> {
+        val webViewRumFeature = resolveRumFeature(sdkCore)
+        val webViewLogsFeature = resolveLogsFeature(sdkCore)
+        val webViewReplayFeature = resolveReplayFeature(sdkCore)
+        val contextProvider = WebViewRumEventContextProvider(sdkCore.internalLogger)
+        if (webViewReplayFeature == null && webViewRumFeature == null) {
```
0xnm, line 98: there is a mistake here, we need also to check `webViewLogsFeature` here. Also it makes sense to create `contentProvider` only inside `else` block.

features/dd-sdk-android-webview/api/apiSurface
xgouchet: Essentially, since the only info you need is a Webview identifier, and not the webview itself, you could delegate that and allow a null value for the Cross Platform SDKs, and maybe not init the SessionReplay consumer if this value is missing.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventConsumer.kt
```diff
@@ -0,0 +1,103 @@
...
+import com.google.gson.JsonObject
+import java.lang.IllegalStateException
+import java.lang.NumberFormatException
+import java.lang.UnsupportedOperationException
+
+internal class WebViewReplayEventConsumer(
+    private val sdkCore: FeatureSdkCore,
+    webViewId: Long,
+    internal val dataWriter: DataWriter<JsonObject>,
+    offsetProvider: TimestampOffsetProvider,
```
0xnm, line 27: minor: maybe it makes sense to simplify constructor by removing `webViewId` and `offsetProvider` here, removing the default value `webViewReplayEventMapper` and just creating it at the construction site? This way constructor will be simpler and shorter.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventMapper.kt
```diff
@@ -0,0 +1,77 @@
...
+        val viewId = viewDataObject?.get(VIEW_ID_KEY)?.asString
+            ?: throw IllegalStateException(BROWSER_EVENT_MISSING_VIEW_DATA_ERROR_MESSAGE)
+        event.get(EVENT_KEY)?.asJsonObject?.let { record ->
+            val timeOffset = offsetProvider.getOffset(viewId, datadogContext)
+            record.get(TIMESTAMP_KEY)?.let { timestamp ->
+                val asLong = timestamp.asLong
+                val correctedTimestamp = asLong + timeOffset
+                record.addProperty(TIMESTAMP_KEY, correctedTimestamp)
+            }
+            record.addProperty(SLOT_ID_KEY, webViewId.toString())
```
0xnm, line 44: How Slot ID is used later?

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventMapper.kt
```diff
@@ -0,0 +1,77 @@
...
+    companion object {
+        const val EVENT_KEY = "event"
+        const val VIEW_OBJECT_KEY = "view"
+        const val VIEW_ID_KEY = "id"
+        const val TIMESTAMP_KEY = "timestamp"
+        const val ENRICHED_RECORD_APPLICATION_ID_KEY = "application_Id"
+        const val ENRICHED_RECORD_SESSION_ID_KEY = "session_id"
+        const val ENRICHED_RECORD_VIEW_ID_KEY = "view_id"
+        const val RECORDS_KEY = "records"
+        const val SLOT_ID_KEY = "slotId"
```
0xnm, line 71: should it be like that? Other keys above are using snake case instead of came case

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventMapper.kt
```diff
@@ -0,0 +1,77 @@
...
+            add(RECORDS_KEY, JsonArray().apply { add(record) })
+        }
+    }
+
+    companion object {
+        const val EVENT_KEY = "event"
+        const val VIEW_OBJECT_KEY = "view"
+        const val VIEW_ID_KEY = "id"
+        const val TIMESTAMP_KEY = "timestamp"
+        const val ENRICHED_RECORD_APPLICATION_ID_KEY = "application_Id"
```
0xnm, line 67: probably a typo here

```suggestion
        const val ENRICHED_RECORD_APPLICATION_ID_KEY = "application_id"
```

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
mariusc83: I need to check this actually.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventMapper.kt
mariusc83: yeah because in our `EnrichedRecord` we are using snake case and in the generated records we are using camelcase

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
mariusc83: yes something was wrong during a `renaming` :(. Nice spot, I need to see why this was not caught in the unit tests actually.

features/dd-sdk-android-webview/api/apiSurface
mariusc83: yeah...I might have to do something to support a null there

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
```diff
@@ -85,14 +93,53 @@ object WebViewTracking {
 
     private fun buildWebViewEventConsumer(
         sdkCore: FeatureSdkCore,
-        logsSampleRate: Float
+        logsSampleRate: Float,
+        webViewId: String? = null
```
0xnm, line 97: we can remove default value here, probably? it is a private method.

features/dd-sdk-android-webview/src/test/kotlin/com/datadog/android/webview/internal/replay/WebViewReplayEventConsumerTest.kt
```diff
@@ -0,0 +1,288 @@
...
+            time = fakeDatadogContext.time.copy(
+                serverTimeOffsetMs = fakeServerTimeOffsetInMillis
+            ),
+            featuresContext = forge.aMap {
+                Feature.SESSION_REPLAY_FEATURE_NAME to fakeSessionReplayFeatureContext
+            }
+        )
+        whenever(
+            mockSdkCore.getDatadogContext()
+        ) doReturn fakeDatadogContext
```
0xnm, line 110: this shouldn't be needed, because you get `DatadogContext` through the `withWriteContext` call.

0xnm (approved):

jonathanmos (approved):
LGTM - I left a few minor comments

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
```diff
@@ -168,4 +213,6 @@ object WebViewTracking {
         "RUM feature is not registered, will ignore RUM events from WebView."
     internal const val LOGS_FEATURE_MISSING_INFO =
         "Logs feature is not registered, will ignore Log events from WebView."
+    internal const val SESSION_REPLAY_FEATURE_MISSING_INFO =
```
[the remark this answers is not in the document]
mariusc83: good spot, it should have been used in the SR feature resolving. Fixed now

features/dd-sdk-android-webview/src/test/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumerTest.kt
```diff
@@ -49,8 +48,6 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
```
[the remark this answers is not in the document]
mariusc83: weird that ktlint format did not remove this already 👍

0xnm (approved):

PR #1798, opened 2024-01-01
RUM-2314 Fix ConcurrentModificationException in BitmapPool

jonathanmos:
### What does this PR do?
Fixes a ConcurrentModificationException that can occur due a race condition between low memory evicting items from the cache, and threads attempting to query the cache.

### Motivation

### Additional Notes
3000 iterations is the lower bound to reliably reproduce the issue in a reasonable number of unit test runs when it exists. This number of iterations takes around 4 seconds to run locally.

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
@@ -103,6 +103,21 @@ internal class BitmapPool(
         }?.apply { markBitmapAsUsed(this) }
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration) {}
+
+    @Synchronized
+    override fun onLowMemory() {
+        bitmapPoolHelper.safeCall {
+            @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
+            cache.evictAll()
+        }
+    }
+
+    @Synchronized
+    override fun onTrimMemory(level: Int) {
+        cacheUtils.handleTrimMemory(level, cache)
+    }
+
     internal fun getBitmapByProperties(width: Int, height: Int, config: Config): Bitmap? {
         val key = bitmapPoolHelper.generateKey(width, height, config)
         return get(key)
@@ -144,19 +159,6 @@ internal class BitmapPool(
             @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
             bitmapsBySize[key]?.add(bitmap)
         }
-    }
-
-    override fun onConfigurationChanged(newConfig: Configuration) {}
-
-    override fun onLowMemory() {
-        bitmapPoolHelper.safeCall {
-            @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
-            cache.evictAll()
-        }
-    }
-
-    override fun onTrimMemory(level: Int) {
-        cacheUtils.handleTrimMemory(level, cache)
     }
 
     internal companion object {
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
@@ -17,7 +17,9 @@ import fr.xgouchet.elmyr.junit5.ForgeConfiguration
 import fr.xgouchet.elmyr.junit5.ForgeExtension
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.RepeatedTest
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
 import org.mockito.Mock
@@ -32,6 +34,8 @@ import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.mockito.quality.Strictness
 import java.util.concurrent.CountDownLatch
+import java.util.concurrent.Executors
+import java.util.concurrent.Future
 import java.util.concurrent.TimeUnit
 
 @Extensions(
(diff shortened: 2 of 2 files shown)


features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -358,6 +361,53 @@ internal class BitmapPoolTest {
         assertThat(captor.firstValue).isEqualTo(0)
     }
 
+    @RepeatedTest(3000)
```
0xnm, line 364: let's reduce the number of repetitions to just few, probably 3000 runs takes too long and we can accumulate them in different runs on CI anyway.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -358,6 +361,53 @@ internal class BitmapPoolTest {
...
+                }
+                )
+            }
+        }
+
+        threadList.forEach { it.start() }
+        allThreadsFinishedLatch.await()
+
+        // Then
+        MatcherAssert.assertThat("Received ConcurrentModificationException when getting bitmap", !failed.get())
```
0xnm, line 408: maybe we can just do `assertThat(failed.get).overridingErrorMessage(...).isFalse` without `MatcherAssert` usage.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -358,6 +361,53 @@ internal class BitmapPoolTest {
...
+        val numberOfThreads = 3
+        val allThreadsStartedLatch = CountDownLatch(numberOfThreads)
+        val allThreadsFinishedLatch = CountDownLatch(numberOfThreads)
+        testedCache.put(mockBitmap)
+        val bitmapKey = spyBitmapPoolHelper.generateKey(mockBitmap)
+        val failed = AtomicBoolean(false)
+
+        // When
+        repeat(numberOfThreads) {
+            if (Math.random() < 0.5) {
```
0xnm, line 379:
```suggestion
            if (forge.aBool()) {
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -358,6 +361,53 @@ internal class BitmapPoolTest {
...
+                threadList.add(
+                Thread {
+                    allThreadsStartedLatch.countDown()
+                    allThreadsStartedLatch.await()
+                    try {
+                        testedCache.get(bitmapKey)
+                    } catch (e: ConcurrentModificationException) {
+                        failed.set(true)
```
0xnm, line 387: The disadvantage of this approach is that we are loosing the original exception and if test fails we have to somehow run it with debugger to be able to see the exception and its stack.

Slightly better approach, in my opinion at least, will be to use `Future` + `ExecutorService`. The you can do something like:

```
val executor = Executors.newFixedThreadPool(3)
val tasks = mutableListOf<Future<*>>()
repeat(...) {
   tasks += executor.submit { ... }
}

tasks.forEach { assertNotThrows { it.get() } }
```

This way you get original exception and a nice related message.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPool.kt
```diff
@@ -103,6 +103,21 @@ internal class BitmapPool(
         }?.apply { markBitmapAsUsed(this) }
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration) {}
+
+    @Synchronized
```
mariusc83, line 108: are you sure it is caused by these methods ? Were you able to reproduce it locally ?

jonathanmos: yes, I reproduced it with the test for onLowMemory. For onTrimMemory it's would probably also occur, but since the tests don't have access to the backing cache and have to call clear - clear is already synchronized so it prevents the issue from being visible.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
jonathanmos: It takes about 4 seconds for 3000 runs (locally). I'll reduce it to 30

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
jonathanmos: Looks good - I'll make this change

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/BitmapPoolTest.kt
```diff
@@ -358,6 +362,40 @@ internal class BitmapPoolTest {
...
+            tasks += executor.submit {
+                if (forge.aBool()) {
+                    allThreadsStartedLatch.countDown()
+                    allThreadsStartedLatch.await()
+                    testedCache.get(bitmapKey)
+                } else {
+                    allThreadsStartedLatch.countDown()
+                    allThreadsStartedLatch.await()
+                    testedCache.onLowMemory()
+                }
```
0xnm, line 391: common part can be lifted out.

```suggestion
                allThreadsStartedLatch.countDown()
                allThreadsStartedLatch.await()
                if (forge.aBool()) {
                    testedCache.get(bitmapKey)
                } else {
                    testedCache.onLowMemory()
                }
```

0xnm (approved):

PR #1804, opened 2024-01-02
Use automatic Gradle daemon instrumentation with CI Visibility instead of manual test tasks instrumentation

nikita-tkachenko-datadog:
### What does this PR do?

Updates CI Visibility instrumentation: rather than configuring each test task to fork a JVM with the DD Java tracer injected, now the tracer is injected into the Gradle daemon that runs the build.

### Motivation

Instrumenting the Gradle daemon has the following advantages:
- No need to manually change the build files, every test task is instrumented automatically.
- All tests executed within a single build are reported as part of a single test session, and it is possible to see the duration and status for the entire build or individual modules.
- Some background operations (such as reporting Git data or building an index of the repository) are done once for the entire build and not for every test task.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,15 +4,15 @@ include:
 # SETUP
 
 variables:
-  CURRENT_CI_IMAGE: "5"
+  CURRENT_CI_IMAGE: "6"
   CI_IMAGE_DOCKER: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/dd-sdk-android:$CURRENT_CI_IMAGE
   GIT_DEPTH: 5
 
   DD_SERVICE: "dd-sdk-android"
   DD_ENV_TESTS: "ci"
-  DD_INTEGRATION_JUNIT_5_ENABLED: "true"
   DD_CIVISIBILITY_ENABLED: "true"
   DD_INSIDE_CI: "true"
+  DD_COMMON_AGENT_CONFIG: "dd.env=ci,dd.trace.enabled=false,dd.jmx.fetch.enabled=false"
 
   KUBERNETES_MEMORY_REQUEST: "8Gi"
   KUBERNETES_MEMORY_LIMIT: "16Gi"
@@ -104,7 +104,7 @@ test:debug:
   script:
     - rm -rf ~/.gradle/daemon/
     - export DD_AGENT_HOST="$BUILDENV_HOST_IP"
-    - GRADLE_OPTS="-Xmx3072m" ./gradlew :unitTestDebug --stacktrace --no-daemon --build-cache --gradle-user-home cache/
+    - GRADLE_OPTS="-Xmx3072m" DD_TAGS="test.configuration.variant:debug" ./gradlew :unitTestDebug --stacktrace --no-daemon --build-cache --gradle-user-home cache/ -Dorg.gradle.jvmargs=-javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=$DD_COMMON_AGENT_CONFIG
   artifacts:
     when: always
     expire_in: 1 week
@@ -125,7 +125,7 @@ test:release:
   script:
     - rm -rf ~/.gradle/daemon/
     - export DD_AGENT_HOST="$BUILDENV_HOST_IP"
-    - GRADLE_OPTS="-Xmx3072m" ./gradlew :unitTestRelease --stacktrace --no-daemon --build-cache --gradle-user-home cache/
+    - GRADLE_OPTS="-Xmx3072m" DD_TAGS="test.configuration.variant:release" ./gradlew :unitTestRelease --stacktrace --no-daemon --build-cache --gradle-user-home cache/ -Dorg.gradle.jvmargs=-javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=$DD_COMMON_AGENT_CONFIG
   artifacts:
     when: always
     expire_in: 1 week
@@ -146,7 +146,7 @@ test:tools:
   script:
     - rm -rf ~/.gradle/daemon/
     - export DD_AGENT_HOST="$BUILDENV_HOST_IP"
-    - GRADLE_OPTS="-Xmx3072m" ./gradlew :unitTestTools --stacktrace --no-daemon --build-cache --gradle-user-home cache/
+    - GRADLE_OPTS="-Xmx3072m" DD_TAGS="test.configuration.variant:jvmRelease" ./gradlew :unitTestTools --stacktrace --no-daemon --build-cache --gradle-user-home cache/ -Dorg.gradle.jvmargs=-javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=$DD_COMMON_AGENT_CONFIG
 
 test:kover:
   tags: [ "arch:amd64" ]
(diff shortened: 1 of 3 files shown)


libs/dd-java-agent-0.98.1.jar
nikita-tkachenko-datadog, line 1: Not sure why the agent Jar was stored directly in the repo - I updated the Docker image to download the agent from Maven. If there's a reason it had to be done that particular way, I'm happy to revert this part of the changes

.gitlab-ci.yml
```diff
@@ -104,7 +101,7 @@ test:debug:
   script:
     - rm -rf ~/.gradle/daemon/
     - export DD_AGENT_HOST="$BUILDENV_HOST_IP"
-    - GRADLE_OPTS="-Xmx3072m" ./gradlew :unitTestDebug --stacktrace --no-daemon --build-cache --gradle-user-home cache/
```
nikita-tkachenko-datadog, line 107: I added `DD_ENV`, `DD_TAGS` and other config params as agent properties rather than env vars because previously they were only set for the test tasks. The idea was to err on the side of caution, since I was not sure what the implications of setting them for the entire pipeline will be. 

If there are no other jobs that need their own `DD_ENV` and `DD_TAGS`, then I'd rather move these settings to the `variables` section.

Dockerfile.gitlab
```diff
@@ -96,3 +96,9 @@ RUN set -x \
 # Install Datadog CI
 RUN npm install -g npm@9.6.5
 RUN npm install -g @datadog/datadog-ci
+
+# Install Datadog Java tracer
+ENV DD_TRACER_FOLDER $PWD/dd-java-tracer
```
mariusc83, line 101: To not make a confusion here with the `dd-java-tracer` module that we are using for APM in our code now I would change the name of this folder to `dd-hava-agent`. What do you think @0xnm

.gitlab-ci.yml
```diff
@@ -4,15 +4,12 @@ include:
...
-  CURRENT_CI_IMAGE: "5"
+  CURRENT_CI_IMAGE: "6"
   CI_IMAGE_DOCKER: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/dd-sdk-android:$CURRENT_CI_IMAGE
   GIT_DEPTH: 5
 
   DD_SERVICE: "dd-sdk-android"
-  DD_ENV_TESTS: "ci"
-  DD_INTEGRATION_JUNIT_5_ENABLED: "true"
   DD_CIVISIBILITY_ENABLED: "true"
-  DD_INSIDE_CI: "true"
```
0xnm, line 15: I would propose to remove only `DD_INTEGRATION_JUNIT_5_ENABLED` variable, other two seems not to be related to the change made.

.gitlab-ci.yml
```diff
@@ -104,7 +101,7 @@ test:debug:
   script:
     - rm -rf ~/.gradle/daemon/
     - export DD_AGENT_HOST="$BUILDENV_HOST_IP"
-    - GRADLE_OPTS="-Xmx3072m" ./gradlew :unitTestDebug --stacktrace --no-daemon --build-cache --gradle-user-home cache/
+    - GRADLE_OPTS="-Xmx3072m" ./gradlew :unitTestDebug --stacktrace --no-daemon --build-cache --gradle-user-home cache/ -Dorg.gradle.jvmargs=-javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=dd.env=ci,dd.trace.enabled=false,dd.jmx.fetch.enabled=false,dd.tags=test.configuration.variant:debug
```
0xnm, line 104: maybe we can extract the common part of agent arguments inside a variable to be stored in the `variables` section do avoid the repetition. Ex. `DD_COMMON_AGENT_CONFIG="-Dorg.gradle.jvmargs=-javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=dd.env=ci,dd.trace.enabled=false,dd.jmx.fetch.enabled=false"`, so that the line looks like:

```suggestion
    - GRADLE_OPTS="-Xmx3072m" ./gradlew :unitTestDebug --stacktrace --no-daemon --build-cache --gradle-user-home cache/ -D$DD_COMMON_AGENT_CONFIG -Ddd.tags=test.configuration.variant:debug
```

Dockerfile.gitlab
0xnm: I'm fine with any, it won't be inside a project folder anyway (it will be inside the entry directory of the docker container).

Dockerfile.gitlab
```diff
@@ -96,3 +96,9 @@ RUN set -x \
...
+RUN mkdir -p $DD_TRACER_FOLDER
+RUN wget -O $DD_TRACER_FOLDER/dd-java-agent.jar 'https://repo1.maven.org/maven2/com/datadoghq/dd-java-agent/1.26.1/dd-java-agent-1.26.1.jar'
```
0xnm, line 103: Can you please extract the version of the agent inside the env variable in the block above (around L27-L32)? So that we have a single place to manage the versions.

.gitlab-ci.yml
nikita-tkachenko-datadog: Sure, reverted the extra changes

.gitlab-ci.yml
nikita-tkachenko-datadog: Makes sense, done

Dockerfile.gitlab
nikita-tkachenko-datadog: Extracted

.gitlab-ci.yml
nikita-tkachenko-datadog: That didn't seem to work, I think because of referencing an env var defined inside the Docker image from an env var defined inside the pipeline. So I had to change it slightly, to only have the actual agent args inside the "common config" variable.

buildSrc/src/main/kotlin/com/datadog/gradle/config/KotlinConfig.kt
```diff
@@ -26,24 +24,4 @@ fun Project.kotlinConfig(
...
-    val moduleName = this@kotlinConfig.name
-    val javaAgentJar = File(File(rootDir, "libs"), "dd-java-agent-1.26.1.jar")
-    taskConfig<AndroidUnitTest> {
-        if (environment["DD_INTEGRATION_JUNIT_5_ENABLED"] == "true") {
-            val variant = variantName.substringBeforeLast("UnitTest")
-
-            // set the `env` tag for the test spans
-            environment("DD_ENV", "ci")
-            // add custom tags based on the module and variant (debug/release, flavors, …)
-            environment("DD_TAGS", "test.module:$moduleName,test.variant:$variant")
```
0xnm, line 39: It seems that we are losing these tags? I can see that `test.configuration.variant` was added in the `gitlab-ci.yml` file, but the name of the tag is different, and `test.module` tag is completely missing. Are these 2 added automatically by the agent by some chance?

nikita-tkachenko-datadog: Yes, the module tag is added automatically, its value is slightly different though (the path of the Gradle task).

I changed the name of the variant tag to make it a part of the [test configuration](https://docs.datadoghq.com/tests/#custom-configurations) - that way the same test executed via a different task will be considered a different test case in the CI Vis backend. I wasn't exactly sure if this is a possible case - if not, I can revert the tag to its original name.

0xnm: ok, clear enough, let's try to have these changes then.

0xnm (approved):

PR #1806, opened 2024-01-03
[RUM-2258] Improve vital support for higher refresh rate devices

plousada:
### What does this PR do?

CPU Vitals start to take into account refresh rate of device obtained either through `OnFrameMetricsAvailableListener` or `Display.getRefreshRate()`.

### Motivation

### Additional Notes


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -24,17 +32,45 @@ import java.util.concurrent.TimeUnit
...
 
+    @RequiresApi(Build.VERSION_CODES.N)
+    interface DDFrameMetricsListener : Window.OnFrameMetricsAvailableListener {
+        override fun onFrameMetricsAvailable(
+            window: Window,
+            frameMetrics: FrameMetrics,
+            dropCountSinceLastInvocation: Int
+        )
+
+        var frameDeadline: Long
```
0xnm, line 56: I believe this one can be removed

```suggestion
    interface DDFrameMetricsListener : Window.OnFrameMetricsAvailableListener {

        var frameDeadline: Long
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -160,6 +215,26 @@ internal class JankStatsActivityLifecycleListener(
...
+            window.addOnFrameMetricsAvailableListener(DDFrameMetricsListener.DEFAULT, handler)
+        } else {
+        // Fallback - Android 30 allows apps to not run at a fixed 60hz, but didn't yet have
+        // Frame Metrics callbacks available
+        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.R) {
+            val displayManager = activity.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
+            display = displayManager.getDisplay(Display.DEFAULT_DISPLAY)
+        }
+    }
+}
```
0xnm, line 230: latest version of `ktlint` we are using have indent formatting turned off (for a while, we have a ticket for it), so it is better to format code with Android Studio to fix the indent.

```suggestion
    private fun registerMetricListener(window: Window, activity: Activity) {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
            val handler = Handler(Looper.getMainLooper())
            window.addOnFrameMetricsAvailableListener(DDFrameMetricsListener.DEFAULT, handler)
        } else if (Build.VERSION.SDK_INT == Build.VERSION_CODES.R && display == null) {
            // Fallback - Android 30 allows apps to not run at a fixed 60hz, but didn't yet have
            // Frame Metrics callbacks available
            val displayManager = activity.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
            display = displayManager.getDisplay(Display.DEFAULT_DISPLAY)
        }
    }
```

Also I think we may set `display` only once (if `DisplayManager` doesn't depend on a particular UI context though).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -160,6 +215,26 @@ internal class JankStatsActivityLifecycleListener(
...
+
+    private fun unregisterMetricListener(window: Window) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+                window.removeOnFrameMetricsAvailableListener(DDFrameMetricsListener.DEFAULT)
+        }
+    }
```
0xnm, line 236: formatting:

```suggestion
    private fun unregisterMetricListener(window: Window) {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
            window.removeOnFrameMetricsAvailableListener(DDFrameMetricsListener.DEFAULT)
        }
    }
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -143,8 +182,24 @@ internal class JankStatsActivityLifecycleListener(
...
+                screenRefreshRate = display?.refreshRate?.toDouble() ?: SIXTY_FPS
+            }
+
+            frameRate *= (SIXTY_FPS / screenRefreshRate)
+
+            // If normalized frame rate is still at over 60fps it means the frame rendered
+            // quickly enough for the devices refresh rate.
+            if (frameRate > MAX_FPS) {
+                frameRate = MAX_FPS
+            }
```
0xnm, line 200: very minor, just looks shorter

```suggestion
            // If normalized frame rate is still at over 60fps it means the frame rendered
            // quickly enough for the devices refresh rate.
            frameRate *= (SIXTY_FPS / screenRefreshRate).coerceAtMost(MAX_FPS)
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -24,17 +32,45 @@ import java.util.concurrent.TimeUnit
...
     internal val activeActivities = WeakHashMap<Window, MutableList<WeakReference<Activity>>>()
+    internal var display: Display? = null
 
     // region ActivityLifecycleCallbacks
     @MainThread
     override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
     }
 
+    @RequiresApi(Build.VERSION_CODES.N)
+    interface DDFrameMetricsListener : Window.OnFrameMetricsAvailableListener {
```
0xnm, line 49: we should move it to the bottom of the file, just before the `companion`, because the ordering is `propeties` > `methods` > `inner classes/interfaces`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -24,17 +32,45 @@ import java.util.concurrent.TimeUnit
 internal class JankStatsActivityLifecycleListener(
     private val vitalObserver: VitalObserver,
     private val internalLogger: InternalLogger,
-    private val jankStatsProvider: JankStatsProvider = JankStatsProvider.DEFAULT
+    private val jankStatsProvider: JankStatsProvider = JankStatsProvider.DEFAULT,
+    private var screenRefreshRate: Double = 60.0
 ) : ActivityLifecycleCallbacks, JankStats.OnFrameListener {
```
0xnm, line 38: We can remove it from the constructor, I don't see any call sites where we pass custom `screenRefreshRate` at the `JankStatsActivityLifecycleListener` construction time

```suggestion
) : ActivityLifecycleCallbacks, JankStats.OnFrameListener {
    private var screenRefreshRate: Double = 60.0
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListenerTest.kt
```diff
@@ -356,10 +359,41 @@ internal class JankStatsActivityLifecycleListenerTest {
         verify(mockObserver, never()).onNewSample(any())
     }
 
+    @Test
+    fun `𝕄 adjust sample value to refresh rate 𝕎 doFrame() {refresh rate over 60hz}`(
```
0xnm, line 363: ideally, it would be nice to see the tests for different Android API versions (since there are different code paths in the class under test depending on API version), but this may explode the number of tests, so it is not a blocker for me.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -65,6 +101,8 @@ internal class JankStatsActivityLifecycleListener(
                 activeWindowsListener[window] = jankStats
```
mariusc83, line 101: Not related with this PR but maybe we should add a task here to not use anymore the `window` as a key here. See the previous problems we had with using the activity as a key for view scopes.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
mariusc83, line 49: any particular reason you need an interface here ? I think just a simple class would be simpler.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -24,17 +32,45 @@ import java.util.concurrent.TimeUnit
...
     private val vitalObserver: VitalObserver,
     private val internalLogger: InternalLogger,
-    private val jankStatsProvider: JankStatsProvider = JankStatsProvider.DEFAULT
+    private val jankStatsProvider: JankStatsProvider = JankStatsProvider.DEFAULT,
+    private var screenRefreshRate: Double = 60.0
 ) : ActivityLifecycleCallbacks, JankStats.OnFrameListener {
 
     internal val activeWindowsListener = WeakHashMap<Window, JankStats>()
     internal val activeActivities = WeakHashMap<Window, MutableList<WeakReference<Activity>>>()
+    internal var display: Display? = null
```
mariusc83, line 41: we should not keep the `Display` as a reference here as it can create memory leaks.

mariusc83 (changes requested):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
0xnm: if it is not bound to the particular UI-context and there is just a single instance for the whole app, it should be fine I guess.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListenerTest.kt
```diff
@@ -356,10 +359,41 @@ internal class JankStatsActivityLifecycleListenerTest {
...
+            verify(mockObserver, never()).onNewSample(any())
+        }
+    }
+
     companion object {
         const val ONE_MILLISECOND_NS: Long = 1000L * 1000L
         const val ONE_SECOND_NS: Long = 1000L * 1000L * 1000L
         const val TEN_SECOND_NS: Long = 10L * ONE_SECOND_NS
         const val ONE_MINUTE_NS: Long = 60L * ONE_SECOND_NS
+        const val MIN_FPS: Double = 1.0
```
jonathanmos, line 396: minor: If we in the future modify min/max fps we would have to change this in two places. We could make the class variables internal to share with the test

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
plousada: That was my assumption, I can replace with a weak reference if you prefer.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
plousada: It is used in the test ``𝕄 adjust sample value to refresh rate 𝕎 doFrame() {refresh rate over 60hz}`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
plousada: Originally it was meant to be provided as parameter to aid in mocking, but ultimately didn't need to. I'll change it.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
mariusc83: if you are certain that it will be the main window `display` instance then it will be fine.

mariusc83 (commented):
I am going to change to `Comment` from `Request Changes` to not block you while I am in PTO

mariusc83 (approved):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListenerTest.kt
plousada: We just need to test pre-R, R and post-R. 
I can include those.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListenerTest.kt
plousada: I think it's unlikely to change, but I can follow this pattern if we're using it elsewhere.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -160,6 +195,42 @@ internal class JankStatsActivityLifecycleListener(
...
+            }
+            val handler = Handler(Looper.getMainLooper())
+            frameMetricsListener?.let { window.addOnFrameMetricsAvailableListener(it, handler) }
+        } else {
+        // Fallback - Android 30 allows apps to not run at a fixed 60hz, but didn't yet have
+        // Frame Metrics callbacks available
+        if (display == null && buildSdkVersionProvider.version() == Build.VERSION_CODES.R) {
+            val displayManager = activity.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
+            display = displayManager.getDisplay(Display.DEFAULT_DISPLAY)
+        }
```
0xnm, line 212: maybe we can pull this out from this method? it may be confusing, because there is no listener registration in this block. Also by pulling it out we will have a symmetry at the calls site between `registerMetricListener` and `unregisterMetricListener` calls in terms of the build version required, because currently the former is guarded by `Build.VERSION.SDK_INT >= Build.VERSION_CODES.R` and the latter `Build.VERSION.SDK_INT >= Build.VERSION_CODES.S` - this raises the obvious question initially: we don't unregister listener registered if `Build.VERSION.SDK_INT == Build.VERSION_CODES.R`?

After this change, the following:

```
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
            registerMetricListener(window, activity)
        }
```

will become:

```
        if (buildSdkVersionProvider.version() >= Build.VERSION_CODES.S) {
            registerMetricListener(window, activity)
        } else if (display == null && buildSdkVersionProvider.version() == Build.VERSION_CODES.R) {
            val displayManager = activity.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
            display = displayManager.getDisplay(Display.DEFAULT_DISPLAY)
        }
```

which is cleaner to me

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -160,6 +195,42 @@ internal class JankStatsActivityLifecycleListener(
...
+                frameMetricsListener = DDFrameMetricsListener()
+            }
+            val handler = Handler(Looper.getMainLooper())
+            frameMetricsListener?.let { window.addOnFrameMetricsAvailableListener(it, handler) }
+        } else {
+        // Fallback - Android 30 allows apps to not run at a fixed 60hz, but didn't yet have
+        // Frame Metrics callbacks available
+        if (display == null && buildSdkVersionProvider.version() == Build.VERSION_CODES.R) {
+            val displayManager = activity.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
+            display = displayManager.getDisplay(Display.DEFAULT_DISPLAY)
```
0xnm, line 211: I believe holding `Display` object shouldn't cause memory leaks by looking on its structure https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/Display.java;l=1?q=Display.java&sq= - it _seems_ it doesn't have references to the UI-related context (Activity), the only concern is `Resources` object which may hold such reference. So maybe it is worth to check this change for the leaks.

0xnm (commented):
I've added some comments/thoughts, let me know what you think. But overall the change is lgtm.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
```diff
@@ -160,6 +195,42 @@ internal class JankStatsActivityLifecycleListener(
         activeActivities[window] = list
     }
 
+    @RequiresApi(Build.VERSION_CODES.N)
```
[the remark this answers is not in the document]
plousada: I was relying on the API required for the method itself, just to be on the safe side.
But I don't have a strong opinion here, I can change it to `.R` if we feel it's better.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsActivityLifecycleListener.kt
plousada: Didn't see anything on my local profiles. 
Also did a quick analysis on the activity allocation/deallocation pattern in the sample app with and without these changes and found no difference.

xgouchet (approved):

0xnm (approved):

PR #1809, opened 2024-01-05
Add testable core wrapper behind property

louiszawadzki:
### What does this PR do?

This is a follow-up on the [initial POC](https://github.com/DataDog/dd-sdk-ios/pull/1570).
I've extracted the functionality in a method of the internal core proxy and added a condition on a build config field to make it harder to use it by mistake.

I'd like to get feedback on the implementation before moving on to add some unit tests to the feature.

### Motivation

The goal of the testable core wrapper is to enable UI feature testing. With this we can make assertions as to what fields were called by the SDK.
This is going to enable us to test the React Native Session Replay implementation, as the work to correctly mock UI RN components in unit test is much larger.

### Additional Notes

On iOS we can get rid of the wrapper altogether with compiler flags, but I don't think such a thing exists on Android. As far as I understand we cannot totally hide or remove the feature for other users. The best I could find was to put it behind an internal proxy and require a build config field to be set.

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/config/GradlePropertiesKeys.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/config/GradlePropertiesKeys.kt
@@ -9,4 +9,6 @@ object GradlePropertiesKeys {
 
     // you can set this property from your gradle.properties as: forceEnableLogcat = true | false
     const val FORCE_ENABLE_LOGCAT = "forceEnableLogcat"
+    // you can set this property from your gradle.properties as: buildForCoreTesting = true | false
+    const val BUILD_FOR_CORE_TESTING = "buildForCoreTesting"
 }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
@@ -1,0 +1,103 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core
+
+import com.datadog.android.api.context.DatadogContext
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureScope
+import com.datadog.android.api.storage.EventBatchWriter
+import com.datadog.android.api.storage.RawBatchEvent
+import com.google.gson.Gson
+import com.google.gson.JsonElement
+
+class DatadogCoreProxy(
+    val core: InternalSdkCore,
+) : InternalSdkCore by core {
+    private val featureScopes = mutableMapOf<String, FeatureScopeInterceptor>()
+
+    fun eventsWritten(featureName: String): String {
+        val events = featureScopes[featureName]?.eventsWritten()?.toList() ?: emptyList<Any>()
+        return Gson().toJson(events)
+    }
+
+    fun clearData(featureName: String) {
+        featureScopes[featureName]?.clearData()
+    }
+
+    override fun registerFeature(feature: Feature) {
+        core.registerFeature(feature)
+        val featureScope = core.getFeature(feature.name)
+        featureScopes[feature.name] = FeatureScopeInterceptor(featureScope!!, core, feature)
+    }
+
+    override fun getFeature(featureName: String): FeatureScope? {
+        core.getFeature(featureName)
+        return featureScopes[featureName]
+    }
+}
+
+internal class FeatureScopeInterceptor(
+    private val featureScope: FeatureScope,
+    private val core: InternalSdkCore,
+    private val feature: Feature,
+    ) : FeatureScope by featureScope {
+
+    private val eventsBatchInterceptor = EventBatchInterceptor()
+
+    fun eventsWritten(): List<String> {
+        return eventsBatchInterceptor.events
+    }
+
+    fun clearData() {
+        eventsBatchInterceptor.clearData()
+    }
+
+    // region FeatureScope
+
+    override fun withWriteContext(
+        forceNewBatch: Boolean,
+        callback: (DatadogContext, EventBatchWriter) -> Unit
+    ) {
+        featureScope.withWriteContext(forceNewBatch, callback)
+
+        val context = core.getDatadogContext()!!
+        callback(context, eventsBatchInterceptor)
+    }
+
+    override fun <T : Feature> unwrap(): T {
+        @Suppress("UNCHECKED_CAST")
+        return feature as T
+    }
+
+    // endregion
+}
+
+
+internal class EventBatchInterceptor: EventBatchWriter {
+    internal val events = mutableListOf<String>()
+
+    override fun currentMetadata(): ByteArray? {
+        TODO("Not yet implemented")
+    }
+
+    fun clearData() {
+        events.clear()
+    }
+
+    override fun write(
+        event: RawBatchEvent,
+        batchMetadata: ByteArray?
+    ): Boolean {
+        val eventContent = String(event.data)
+
+        events.add(events.size,
+            eventContent
+        )
+
+        return false
+    }
+}
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkCoreRegistry.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkCoreRegistry.kt
@@ -6,6 +6,7 @@
 
 package com.datadog.android.core.internal
 
+import com.datadog.android.BuildConfig
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
 import com.datadog.android.core.DatadogCoreProxy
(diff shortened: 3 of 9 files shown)


buildSrc/src/main/kotlin/com/datadog/gradle/config/GradlePropertiesKeys.kt
```diff
@@ -9,4 +9,6 @@ object GradlePropertiesKeys {
 
     // you can set this property from your gradle.properties as: forceEnableLogcat = true | false
     const val FORCE_ENABLE_LOGCAT = "forceEnableLogcat"
+    // you can set this property from your gradle.properties as: buildForCoreTesting = true | false
+    const val BUILD_FOR_CORE_TESTING = "buildForCoreTesting"
```
0xnm, line 13: I'm not quite clear how this is going to be used from the RN side, given it either consumes release artifacts or snapshot artifacts (which are release variants as well). Which one of the two is supposed to have this flag enabled?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+
+import com.datadog.android.api.context.DatadogContext
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureScope
+import com.datadog.android.api.storage.EventBatchWriter
+import com.datadog.android.api.storage.RawBatchEvent
+import com.google.gson.Gson
+import com.google.gson.JsonElement
+
+class DatadogCoreProxy(
```
0xnm, line 17: I think this is very similar to https://github.com/DataDog/dd-sdk-android/blob/develop/reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubSDKCore.kt we have (but it is not published).

If this one is needed for the testing, maybe we can simply inject it into the necessary features without registering in the registry? In fact, very few components rely on the registry internally (like OkHttp instrumentation, which needs to capture the core in the lazy manner, hence it is using name -> registry), and probably none of them are used by the RN.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+    val core: InternalSdkCore,
+) : InternalSdkCore by core {
+    private val featureScopes = mutableMapOf<String, FeatureScopeInterceptor>()
+
+    fun eventsWritten(featureName: String): String {
+        val events = featureScopes[featureName]?.eventsWritten()?.toList() ?: emptyList<Any>()
+        return Gson().toJson(events)
```
0xnm, line 24: `Gson` instance is thread-safe and stateless, it can be created only once and stored in the class field.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+import com.google.gson.Gson
+import com.google.gson.JsonElement
+
+class DatadogCoreProxy(
+    val core: InternalSdkCore,
+) : InternalSdkCore by core {
+    private val featureScopes = mutableMapOf<String, FeatureScopeInterceptor>()
+
+    fun eventsWritten(featureName: String): String {
+        val events = featureScopes[featureName]?.eventsWritten()?.toList() ?: emptyList<Any>()
```
0xnm, line 23: It is better to have `toList` call in the `FeatureScopeInterceptor#eventsWritten` instead, so it can provide immutable copy by design. Also there is handy `orEmpty` method.

```suggestion
        val events = featureScopes[featureName]?.eventsWritten().orEmpty()
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    fun clearData(featureName: String) {
+        featureScopes[featureName]?.clearData()
+    }
+
+    override fun registerFeature(feature: Feature) {
+        core.registerFeature(feature)
+        val featureScope = core.getFeature(feature.name)
+        featureScopes[feature.name] = FeatureScopeInterceptor(featureScope!!, core, feature)
```
0xnm, line 34: Unless we want things to crash with `!!` call, we can do an early return.

```suggestion
        val featureScope = core.getFeature(feature.name) ?: return null
        featureScopes[feature.name] = FeatureScopeInterceptor(featureScope, core, feature)
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    override fun getFeature(featureName: String): FeatureScope? {
+        core.getFeature(featureName)
```
0xnm, line 38: why this call is needed? we don't use return value here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+        return featureScopes[featureName]
+    }
+}
+
+internal class FeatureScopeInterceptor(
+    private val featureScope: FeatureScope,
+    private val core: InternalSdkCore,
+    private val feature: Feature,
```
0xnm, line 46: It seems `feature` property is not needed here, it is used only in the `unwrap` method below, but this method can be removed and handled by the delegate.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+        forceNewBatch: Boolean,
+        callback: (DatadogContext, EventBatchWriter) -> Unit
+    ) {
+        featureScope.withWriteContext(forceNewBatch, callback)
+
+        val context = core.getDatadogContext()!!
+        callback(context, eventsBatchInterceptor)
+    }
+
+    override fun <T : Feature> unwrap(): T {
```
0xnm, line 71: this one can be removed, it is handled by the delegate.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+    // region FeatureScope
+
+    override fun withWriteContext(
+        forceNewBatch: Boolean,
+        callback: (DatadogContext, EventBatchWriter) -> Unit
+    ) {
+        featureScope.withWriteContext(forceNewBatch, callback)
+
+        val context = core.getDatadogContext()!!
+        callback(context, eventsBatchInterceptor)
```
0xnm, line 68: This will only work if the callback provided is pure (has no side-effects). If the callback is modifying some external state, then calling it twice may lead to an unexpected state (and there are cases for that, for example in the RUM view scope which modifies the number of pending events in the callback). Can you check maybe how the things are done in `StubSDKCore`?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    override fun write(
+        event: RawBatchEvent,
+        batchMetadata: ByteArray?
+    ): Boolean {
+        val eventContent = String(event.data)
+
+        events.add(events.size,
+            eventContent
+        )
```
0xnm, line 99: `MutableList#add` always adds to the end.

```suggestion
        events.add(eventContent)
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+
+        return false
```
0xnm, line 101: It should be `true` here instead, because if write failed, callback may log some unwanted telemetry/messages.

0xnm (commented):
I gave my initial feedback on the change, let me know what you think.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCoreProxy.kt
xgouchet: I think it would make sense to rely on the existing Stub we are using internally, instead of duplicating the same feature. In addition, it'd be better to keep test purpose classes in a test artifact that would not be shipped in an app

louiszawadzki: I believe this is the most important comment so I'll focus on this one before jumping to the others :) 

The goal of this Core Proxy is to be able to run assertions on events that were received by the core in the context of a real app (for instance to test React Native Session Replay). 

Also as I am using the RN SDK in the testing app I cannot interfere too much on how the SDK is going to be initialized except through config fields.
I can potentially change some gradle files but I'd like to avoid having to touch actual code of the RN SDK.

Maybe there is a way to release this as a separate artifact that could be used by the RN SDK without making too many changes to it, if so let me know, I can take time to explore more solutions :) 

Let me know if this is clearer and if you see a better alternative here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkCoreRegistry.kt
```diff
@@ -66,6 +69,23 @@ internal class SdkCoreRegistry(
         instances.clear()
     }
 
+    /**
+     * Wraps the core with a proxy to start listening to feature events.
+     * This must be called before any feature is registered.
+     */
+    fun wrapCoreWithProxy(name: String?): DatadogCoreProxy? {
```
mariusc83, line 76: this method looks a bit weird...at least the naming. If you want to keep the same logic I would name it `try...` if not maybe the condition could be checked outside of the method in another function: `isBuildForCoreTesting` and then call this method inside:
```
coreProxy = if(isBuildForCoreTesting()){wrapWithCoreProxy(name)}
 ```

PR #1810, opened 2024-01-05
feat: Add accessor for current session id

fuzzybinary:
### What does this PR do?

There are use cases (usually support use cases) where clients what to know the current session id without depending on the `sessionStarted` event. This provides that accessor.

refs: RUM-1962

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
@@ -20,6 +20,11 @@ import com.datadog.tools.annotation.NoOpImplementation
 @Suppress("ComplexInterface", "TooManyFunctions")
 @NoOpImplementation
 interface RumMonitor {
+    /**
+     * The current active session id. May return null if a session has not been started yet.
+     * Will return a "null session id" (all zeros) if the current session is not tracked.
+     */
+    val currentSessionId: String?
 
     /**
      * Notifies that a View is being shown to the user, linked with the [key] instance.
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
@@ -21,10 +21,15 @@ import com.datadog.tools.annotation.NoOpImplementation
 @NoOpImplementation
 interface RumMonitor {
     /**
-     * The current active session id. May return null if a session has not been started yet.
-     * Will return a "null session id" (all zeros) if the current session is not tracked.
+     * Get the current active session ID. The session ID will be null if no session is active or
+     * if the session has been sampled out.
+     *
+     * This method uses an asynchronous callback to ensure all pending RUM events have been processed
+     * up to the moment of the call.
+     *
+     * @param callback the callback to be invoked with the current session id.
      */
-    val currentSessionId: String?
+    fun getCurrentSessionId(callback: (String?) -> Unit)
 
     /**
      * Notifies that a View is being shown to the user, linked with the [key] instance.
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
@@ -99,16 +99,21 @@ internal class DatadogRumMonitor(
 
     // region RumMonitor
 
-    override val currentSessionId: String?
-        get() {
-            val applicationScope = rootScope as? RumApplicationScope
-            val sessionScope = applicationScope?.activeSession as? RumSessionScope
-            if (sessionScope != null && sessionScope.sessionState == RumSessionScope.State.NOT_TRACKED) {
-                // Aligns with iOS which assigns a NULL session id when the session is not tracked
-                return RumContext.NULL_UUID
+    override fun getCurrentSessionId(callback: (String?) -> Unit) {
+        executorService.submitSafe(
+            "Get current session ID",
+            sdkCore.internalLogger
+        ) {
+            val activeSession = rootScope
+                .getRumContext()
+                .sessionId
+            if (activeSession == RumContext.NULL_UUID) {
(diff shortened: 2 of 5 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -20,6 +20,11 @@ import com.datadog.tools.annotation.NoOpImplementation
 @Suppress("ComplexInterface", "TooManyFunctions")
 @NoOpImplementation
 interface RumMonitor {
+    /**
+     * The current active session id. May return null if a session has not been started yet.
+     * Will return a "null session id" (all zeros) if the current session is not tracked.
```
0xnm, line 25:
```suggestion
     * The current active session ID. May return null if a session has not been started yet.
     * Will return a "null session ID" (all zeros) if the current session is not tracked.
```

0xnm, line 25: To be precise, it won't be all zeros, but `00000000-0000-0000-0000-000000000000`, so user cannot write something like `if (currentSessionId.all {it == '0'})`, so maybe for the sake of comparison with `null session` mentioned in the documentation, we need to expose such value?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -98,6 +99,17 @@ internal class DatadogRumMonitor(
 
     // region RumMonitor
 
+    override val currentSessionId: String?
```
0xnm, line 102: This method is not thread-safe, because all the values inside RUM scopes hierarchy are set from the dedicated RUM thread, but `currentSessionId` can be called from any user thread. IMO it is not a big deal here, but if use-case from the user side requires consistency and precision, we may need to make this one thread-safe to avoid concurrency issues.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitorTest.kt
```diff
@@ -193,6 +193,27 @@ internal class DatadogRumMonitorTest {
         assertThat(rootScope.backgroundTrackingEnabled).isEqualTo(fakeBackgroundTrackingEnabled)
     }
 
+    @Test
+    fun `M returns null id for current session W no session started`() {
```
0xnm, line 197:
```suggestion
    fun `M return null ID for current session W currentSessionId { no session started }`() {
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitorTest.kt
```diff
@@ -212,6 +233,64 @@ internal class DatadogRumMonitorTest {
         verifyNoMoreInteractions(mockScope, mockWriter)
     }
 
+    @Test
+    fun `M currentSessionId returns correct sessionId W session started { sampled in }`(
```
0xnm, line 237:
```suggestion
    fun `M return correct sessionId W currentSessionId { session started, sampled in }`(
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitorTest.kt
```diff
@@ -193,6 +193,27 @@ internal class DatadogRumMonitorTest {
...
+            mockHandler,
+            mockTelemetryEventHandler,
+            mockResolver,
+            mockCpuVitalMonitor,
+            mockMemoryVitalMonitor,
+            mockFrameRateVitalMonitor,
+            mockSessionListener
+        )
+
+        assertThat(testedMonitor.currentSessionId).isEqualTo(RumContext.NULL_UUID)
```
0xnm, line 214: This is not aligned to what documentation of `currentSessionId` is saying:

```
May return null if a session has not been started yet.
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitorTest.kt
```diff
@@ -212,6 +233,64 @@ internal class DatadogRumMonitorTest {
...
+
+        argumentCaptor<String> {
+            verify(mockSessionListener).onSessionStarted(capture(), any())
+
+            assertThat(testedMonitor.currentSessionId).isEqualTo(firstValue)
+        }
+    }
+
+    @Test
+    fun `M currentSessionId returns null session W session started { sampled out }`(
```
0xnm, line 268:
```suggestion
    fun `M return null session W currentSessionId { session started, sampled out }`(
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -98,6 +99,17 @@ internal class DatadogRumMonitor(
...
+        get() {
+            val applicationScope = rootScope as? RumApplicationScope
+            val sessionScope = applicationScope?.activeSession as? RumSessionScope
+            if (sessionScope != null && sessionScope.sessionState == RumSessionScope.State.NOT_TRACKED) {
+                // Aligns with iOS which assigns a NULL session id when the session is not tracked
+                return RumContext.NULL_UUID
```
xgouchet, line 108: For "readability" purposes, in link with Nikita's comment above, maybe we should return `null` instead here. This would make it easier for customers to not use a `"000…"` value which doesn't make sense?

fuzzybinary:
@xgouchet  / @0xnm Replaced the accessor with an async callback.  Let me know what you think

0xnm (approved):
lgtm! I've added a minor suggestion

mariusc83 (approved):

xgouchet (approved):

xgouchet (approved):

PR #1812, opened 2024-01-08
RUM-2503 Use internal key for View Scopes

xgouchet:
### What does this PR do?

Use an Internal Key representation instead of a (weak) reference to actual customer instances.

### Motivation

Prevent leaking memory with UI related instances (Activity, Fragment, …)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumRawEvent.kt
```diff
@@ -22,14 +22,14 @@ internal sealed class RumRawEvent {
     abstract val eventTime: Time
 
     internal data class StartView(
-        val key: Any,
+        val key: RumScopeKey,
```
0xnm, line 25: Should we rename `RumScopeKey` to `RumViewScopeKey`? Because `key` notation is used for some other classes of `RumRawEvent`, but we are changing it to `RumScopeKey` only for `StartView` and `StopView`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumScopeKey.kt
```diff
@@ -0,0 +1,92 @@
...
+package com.datadog.android.rum.internal.domain.scope
+
+import android.app.Activity
+import android.content.ComponentName
+import androidx.navigation.ActivityNavigator
+import androidx.navigation.fragment.DialogFragmentNavigator
+import androidx.navigation.fragment.FragmentNavigator
+
+/**
+ * Uniquely identifier to link multiple events relating to the same entity.
```
0xnm, line 16:
```suggestion
 * Unique identifier to link multiple events relating to the same entity.
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumScopeKey.kt
```diff
@@ -0,0 +1,92 @@
...
+ * This fixes an issue where we would keep (weak) references to activities/fragments
+ *
+ */
+internal data class RumScopeKey(
+    val id: String,
+    val url: String,
+    val name: String?
```
0xnm, line 23: It seems that this property is not used?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumScopeKey.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        private fun resolveUrl(key: Any): String {
+            return when (key) {
+                is String -> key
+
+                is DialogFragmentNavigator.Destination -> key.className
+                is FragmentNavigator.Destination -> key.className
+
+                else -> key.javaClass.name
```
0xnm, line 60: In the current code we are using `canonicalName` with fallback to `simpleName` here instead of `name`. Won't it cause any regressions?

0xnm (commented):
Nice work! I left few questions.

jonathanmos (approved):

mariusc83 (commented):
Nice work, same concerns as @0xnm regarding the url resolver.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumScopeKey.kt
xgouchet: 👍

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumScopeKey.kt
xgouchet: Missed an additional commit

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumScopeKey.kt
xgouchet: 👍

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumScopeKey.kt
```diff
@@ -0,0 +1,92 @@
...
+
+        private fun resolveName(key: Any): String {
+            return when (key) {
+                is String -> key
+
+                is DialogFragmentNavigator.Destination -> key.className
+                is FragmentNavigator.Destination -> key.className
+
+                else -> key.javaClass.name
+            }
```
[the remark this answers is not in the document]
xgouchet: They could be different in the future indeed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -114,7 +113,7 @@ internal class RumViewManagerScope(
         while (iterator.hasNext()) {
             val childScope = iterator.next()
             if (event is RumRawEvent.StopView) {
-                if (childScope.isActive() && (childScope as? RumViewScope)?.keyRef?.get() == event.key) {
+                if (childScope.isActive() && (childScope as? RumViewScope)?.key == event.key) {
```
0xnm, line 116: Should we compare `key.id` here instead? in theory 2 different APIs can create keys in a different manner: `RumScopeKey.from(key)` and `RumScopeKey.from(key, name)` and while they are backed by the same `key` (hence same `id`), the `name` component will be different, leading to `(childScope as? RumViewScope)?.key != event.key`. Such misuse will be hard to catch.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -246,9 +241,7 @@ internal open class RumViewScope(
         writer: DataWriter<Any>
     ) {
         delegateEventToChildren(event, writer)
-        if (stopped) return
-        val startedKey = keyRef.get()
-        val shouldStop = (event.key == startedKey) || (startedKey == null)
+        val shouldStop = (event.key == key)
```
0xnm, line 244: Same question as above `event.key == key` vs `event.key.id == key.id` comparison

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -114,7 +113,7 @@ internal class RumViewManagerScope(
         while (iterator.hasNext()) {
             val childScope = iterator.next()
             if (event is RumRawEvent.StopView) {
-                if (childScope.isActive() && (childScope as? RumViewScope)?.keyRef?.get() == event.key) {
+                if (childScope.isActive() && (childScope as? RumViewScope)?.key?.id == event.key.id) {
```
0xnm, line 116: In order to make sure that we check `id` property in the test and not the whole `key`, maybe we can modify the tests for this class (and for `RumViewScope`) to use different `RumScopeKey` for different calls (same ID, but different name/url), but it is not a blocker for me, more like something useful to add.

0xnm (approved):
lgtm! I've added one minor thought.

mariusc83 (approved):

PR #1818, opened 2024-01-11
RUM-2459: Safe serialization of user-provided attributes

0xnm:
### What does this PR do?

This PR brings safer serialization of user-provided attributes, so that any exception thrown during this process won't discard the whole event (span, log, RUM).

It consists of 2 improvements:

* Serialization of `additionalProperties` (typically exists in the `usr` data and in the top-level event context) won't break the serialization of the parent event. Instead, if some particular item failed serialization, it will be dropped and serialization will continue. Both key and value will be dropped, this is a consistent behavior with drop of the reserved/non-conformant attributes.
* We will make a copy of user-provided attributes in `RumMonitor` at the call site. This is to prevent `ConcurrentModificationException` which sometimes happen when we iterate over them later. Making a copy at the call-site has a benefit of doing it on the user-thread, which lowers the chance that concurrent modification is done at the same time, and if there is such case, exception will be thrown on the user thread and will be more visible.

Both improvements operate only on the top-level items and don't go deep, this is enough to bring the necessary safety.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -149,28 +149,28 @@ internal class DatadogRumMonitor(
...
-            RumRawEvent.StartAction(type, name, false, attributes, eventTime)
+            RumRawEvent.StartAction(type, name, false, attributes.toMap(), eventTime)
         )
     }
 
     override fun startAction(type: RumActionType, name: String, attributes: Map<String, Any?>) {
         val eventTime = getEventTime(attributes)
         handleEvent(
-            RumRawEvent.StartAction(type, name, true, attributes, eventTime)
+            RumRawEvent.StartAction(type, name, true, attributes.toMap(), eventTime)
```
fuzzybinary, line 173: Why are these `toMap` calls needed now?

fuzzybinary (approved):

fuzzybinary:
I may want to pull similar functionality into the Flutter SDK, specifically for calling the Mapper. Is the `safeMapValuesToJson` available to me or will I have to copy it?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/MiscUtils.kt
```diff
@@ -93,6 +96,28 @@ object JsonSerializer {
...
         }
     }
+
+    /**
+     * This method will convert all values to JSON in a safe way, meaning if serialization fails
+     * for the particular value, the process will continue and faulty value will be dropped.
+     */
+    @InternalApi
+    fun Map<String, Any?>.safeMapValuesToJson(internalLogger: InternalLogger): Map<String, Any?> {
+        val result = mutableMapOf<String, Any?>()
```
xgouchet, line 106: > [!NOTE]
> Maybe we could mark the return value as `Map<String, JsonElement>` instead.

xgouchet (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
0xnm: This is to avoid having `ConcurrentModificationException` when we iterate over them later on another thread and they can be modified on the user-thread in the meantime. Quote from the PR description:

> We will make a copy of user-provided attributes in RumMonitor at the call site. This is to prevent ConcurrentModificationException which sometimes happen when we iterate over them later. Making a copy at the call-site has a benefit of doing it on the user-thread, which lowers the chance that concurrent modification is done at the same time, and if there is such case, exception will be thrown on the user thread and will be more visible.

0xnm:
@fuzzybinary `safeMapValuesToJson` is `public-internal` API, so you can use it.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/MiscUtils.kt
0xnm: Indeed, good point. It should still be compatible with `Map<String, Any?>` type required by the `additionalProperties` normally.

mariusc83 (approved):
Great work

PR #1820, opened 2024-01-12
POC Add Otel support using the APM Java agent dependencies

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


features/dd-sdk-android-trace/build.gradle.kts
```diff
@@ -50,8 +50,14 @@ android {
     }
 }
 
+repositories{
+    mavenLocal()
+}
+
 dependencies {
     api(project(":dd-sdk-android-core"))
+    api("io.opentelemetry:opentelemetry-api:1.4.0")
```
0xnm, line 59: why `api` and not `implementation`?

features/dd-sdk-android-trace/build.gradle.kts
```diff
@@ -50,8 +50,14 @@ android {
...
+    implementation("com.datadoghq:dd-java-agent:1.28.0-SNAPSHOT")
```
0xnm, line 60: why `SNAPSHOT`?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/AndroidTracer.kt
```diff
@@ -41,8 +57,11 @@ class AndroidTracer internal constructor(
     writer: Writer,
     random: Random,
     private val logsHandler: LogHandler,
-    private val bundleWithRum: Boolean
-) : DDTracer(config, writer, random) {
+    private val bundleWithRum: Boolean,
+    coreTracer: CoreTracer
+) : DDTracer(config, writer, random), Tracer {
+
+    private val otelTracer = OtelTracer("otel",coreTracer)
```
0xnm, line 64: are we going to support both OpenTracing and Otel in the same tracer?

sample/kotlin/build.gradle.kts
```diff
@@ -139,7 +139,13 @@ android {
     }
 }
 
+repositories{
+    mavenLocal()
+}
+
 dependencies {
+    implementation("io.opentelemetry:opentelemetry-api:1.4.0")
```
0xnm, line 147: it is already a part of `dd-sdk-android-trace` with `api` visibility though, so not needed here.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/AndroidTracer.kt
mariusc83: That's the idea @ganeshnj is following on iOS so we need to stay consistent here. Later we can deprecate all the methods from old tracer interface and then remove it.

sample/kotlin/build.gradle.kts
mariusc83: yes it will not be like that in the implementation anyway, this is just me playing with things here. In the implementation we will only provide the `opentelemetry` at compile time and the clients will add the dependency on their end. That's also the way APM does.

features/dd-sdk-android-trace/build.gradle.kts
mariusc83: me playing with things ...like I mentioned below in the real implementation will be `compile` and clients will add the dependency on their end.

features/dd-sdk-android-trace/build.gradle.kts
mariusc83: it is a local maven version that I modified for this project. In APM they are following the same approach for local/dev releases so they add a `SNAPSHOT` at the end.

PR #1827, opened 2024-01-16
RUM-2704: Add a request builder for resources

jonathanmos:
### What does this PR do?
Adds a request builder for resources, and renames the existing request builder factories to make it clear that they handle segments.

This is the first of several prs and at the moment is detached code. 

### Motivation

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,82 @@
...
+import com.datadog.android.sessionreplay.internal.recorder.SessionReplayResource
+import com.datadog.android.sessionreplay.internal.recorder.SessionReplayResourceContext
+import com.google.gson.JsonIOException
+import okhttp3.MediaType.Companion.toMediaTypeOrNull
+import okhttp3.MultipartBody
+import okhttp3.RequestBody
+import okhttp3.RequestBody.Companion.toRequestBody
+
+internal class ResourceRequestBodyFactory {
+    fun create(
```
0xnm, line 20: It seems that this method is useless, the only thing it does it the call to another method with exactly the same arguments. We need either to remove this one or inline the body.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,82 @@
...
+                messageBuilder = { CONVERT_TO_JSON_ERROR_MESSAGE },
+                throwable = e
+            )
+            ""
+        }
+    }
+
+    companion object {
+        internal const val CONTENT_TYPE_IMAGE = "image/png"
+        internal const val CONTENT_TYPE_APPLICATION = "application/json"
```
0xnm, line 77: we can hold `MediaType` object directly, without the need to doing the conversion on each call.

```suggestion
        internal val CONTENT_TYPE_IMAGE = "image/png".toMediaTypeOrNull()
        internal val CONTENT_TYPE_APPLICATION = "application/json".toMediaTypeOrNull()
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,82 @@
...
+            val data = it.data
+
+            builder.addFormDataPart(
+                name = "image",
+                filename,
+                data.toRequestBody(CONTENT_TYPE_IMAGE.toMediaTypeOrNull())
+            )
+        }
+
+        if (resources.isNotEmpty()) {
```
0xnm, line 45: What if `resources` is empty? we then will return empty `RequestBody`? Does it make sense to return empty object/use it?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,82 @@
...
+            val context = resources.first().context
```
0xnm, line 46: How do we make sure that all `resources` we get share the same `context`? Is it some invariant that is satisfied upper in the call stack?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,82 @@
...
+        return try {
+            context.toJson()
+        } catch (e: JsonIOException) {
+            internalLogger.log(
+                target = InternalLogger.Target.TELEMETRY,
+                level = InternalLogger.Level.WARN,
+                messageBuilder = { CONVERT_TO_JSON_ERROR_MESSAGE },
+                throwable = e
+            )
+            ""
```
0xnm, line 71: will backend accept resources with empty `resource` blob?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayResourceContext.kt
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import com.google.gson.JsonObject
+
+internal data class SessionReplayResourceContext(
+    var applicationId: String,
+    var type: String = "resource"
```
0xnm, line 13:
```suggestion
    val applicationId: String,
    val type: String = "resource"
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayResource.kt
```diff
@@ -0,0 +1,40 @@
...
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import java.io.Serializable
+
+// An individual resource. It is used to describe binary representation of heavy resources such as images.
+internal data class SessionReplayResource(
+    var identifier: String,
+    var data: ByteArray,
+    var context: SessionReplayResourceContext
```
0xnm, line 15:
```suggestion
    val identifier: String,
    val data: ByteArray,
    val context: SessionReplayResourceContext
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/SessionReplayResourceContextFactory.kt
```diff
@@ -0,0 +1,19 @@
...
+package com.datadog.android.sessionreplay.forge
+
+import com.datadog.android.sessionreplay.internal.recorder.SessionReplayResourceContext
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+internal class SessionReplayResourceContextFactory : ForgeryFactory<SessionReplayResourceContext> {
+    override fun getForgery(forge: Forge): SessionReplayResourceContext {
+        return SessionReplayResourceContext(
+            applicationId = forge.anAsciiString()
```
0xnm, line 16:
```suggestion
            applicationId = getForgery<UUID>().toString()
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/SessionReplayResourceFactory.kt
```diff
@@ -0,0 +1,24 @@
...
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+internal class SessionReplayResourceFactory : ForgeryFactory<SessionReplayResource> {
+    override fun getForgery(forge: Forge): SessionReplayResource {
+        val sessionReplayResourceContext = forge.getForgery<SessionReplayResourceContext>()
+
+        return SessionReplayResource(
+            identifier = forge.anAsciiString(),
+            data = forge.anAsciiString().toByteArray(),
```
0xnm, line 20: Just for the consistency with what is used in most of the places

```suggestion
            identifier = forge.anAlphabeticalString(),
            data = forge.anAlphabeticalString().toByteArray(),
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayResource.kt
```diff
@@ -0,0 +1,40 @@
...
+) : Serializable {
```
0xnm, line 16: why does it need to be `Serializable`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+import okhttp3.RequestBody
+import okio.Buffer
+import java.io.IOException
+import java.io.ObjectInputStream
+import java.io.StreamCorruptedException
+import java.lang.NullPointerException
+import java.util.Locale
+import java.util.UUID
+
+internal class ResourceRequestFactory(
```
0xnm, line 26: if I understand correctly, this class is not used right now, but it will be used in the future with another feature?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+        batchData: List<RawBatchEvent>,
+        batchMetadata: ByteArray?
+    ): Request {
+        val resources = batchData.mapNotNull { rawBatchEvent ->
+            extractSessionReplayResource(rawBatchEvent)
+        }
+
+        if (resources.isEmpty()) {
+            @Suppress("ThrowingInternalException")
+            throw InvalidPayloadFormatException(INVALID_PAYLOAD_FORMAT_ERROR)
```
0xnm, line 44: Is there a possibility to add actionable information here? like when it happens we need to know why exactly it happened and how to solve that

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+            rawEventsToResourcesMapper
+                .map(ois)
+        }.getOrElse {
+            when (it) {
+                is StreamCorruptedException,
+                is NullPointerException,
+                is SecurityException,
+                is IOException -> {
+                    throw InvalidPayloadFormatException(
+                        INVALID_PAYLOAD_FORMAT_ERROR
```
0xnm, line 98: it makes sense to attach the cause as well

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+            RequestFactory.HEADER_EVP_ORIGIN_VERSION to datadogContext.sdkVersion,
+            RequestFactory.HEADER_REQUEST_ID to requestId
+        )
+    }
+
+    @SuppressWarnings("TooGenericExceptionCaught", "ThrowingInternalException")
+    private fun extractSessionReplayResource(rawBatchEvent: RawBatchEvent): SessionReplayResource? {
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        return runCatching {
+            val ois = ObjectInputStream(rawBatchEvent.data.inputStream())
```
0xnm, line 88: why do we need to operate on the stream here? what is the rationale for that, given that all the bytes are already in the memory?

0xnm (commented):
I've reviewed production files (didn't review tests yet) and added few questions.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayResource.kt
```diff
@@ -0,0 +1,40 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import java.io.Serializable
+
+// An individual resource. It is used to describe binary representation of heavy resources such as images.
+internal data class SessionReplayResource(
```
mariusc83, line 12: Because you provide your own implementation of the `equals` class this class needs to be unit tests. See the `BatchData` class I added recently and check the `BatchDataTest` class.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
jonathanmos: It's the SessionReplayResourceContext - essentially the applicationId, which would be the same for all of them. I'll make this clearer in the code

0xnm: It is only `applicationId` there for now, but if the structure of context changes (by adding something else), we may have bugs.

jonathanmos: Following our discussion, I'm changing this to extract the applicationId from the datadogContext

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/SessionReplayResourceFactory.kt
jonathanmos: what is the advantage of making this change?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
jonathanmos: Yes, exactly. This will be connected to the ResourceFeature in the next pr

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
jonathanmos: I'm changing to be more specific errors

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayResource.kt
jonathanmos: I did this because we will need to serialize this object in order to insert it in the `RawBatchEvent` data field, and in order to do so it would need to be serializable

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
```diff
@@ -0,0 +1,97 @@
...
+        ?: throw InvalidPayloadFormatException(COULD_NOT_GET_APPLICATION_ID_ERROR)
+
+        val requestBody = resourceRequestBodyFactory
+            .create(applicationId, batchData)
+
+        return resolveRequest(context, requestBody)
+    }
+
+    private fun getApplicationId(datadogContext: DatadogContext): String? {
+        val rumContext = datadogContext.featuresContext[Feature.RUM_FEATURE_NAME]
```
0xnm, line 42: A random thought: should we store `application_id` when we write resource? Reading it at the upload cycle implies that RUM should be enabled at the time Session Replay is doing the upload and that application ID is the same as it was when resource was written.

Can it be the case that any of these statements can be false?

Not a blocker for me, just thinking.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactoryTest.kt
```diff
@@ -0,0 +1,161 @@
...
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(ForgeConfigurator::class)
+internal class ResourceRequestBodyFactoryTest {
+    private lateinit var testedSegmentRequestBodyFactory: SegmentRequestBodyFactory
```
0xnm, line 43: it seems that a completely different class is tested here, it should be `ResourceRequestBodyFactory`

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactoryTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    @Forgery
+    lateinit var fakeRawBatchEvent: RawBatchEvent
+
+    private lateinit var fakeMediaType: MediaType
+
+    @StringForgery
+    lateinit var fakeApplicationId: String
+
+    @BeforeEach
+    fun `set up`(forge: Forge) {
```
0xnm, line 54: why the tests are missing?

0xnm (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        if (uniqueApplicationIds.size == 0) {
+            @Suppress("ThrowingInternalException")
+            throw InvalidPayloadFormatException(UNABLE_GET_APPLICATION_ID_ERROR)
+        }
+
+        if (uniqueApplicationIds.size > 1) {
+            @Suppress("ThrowingInternalException")
+            throw InvalidPayloadFormatException(MULTIPLE_APPLICATION_ID_ERROR)
```
jonathanmos, line 54: This is a bit of a problem, because it means that a list of resources with two different applicationIds won't send correctly. OTOH, I'm not sure we can prevalidate and split this before it reaches the factory

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
jonathanmos: after checking with them, in case of being unable to resolve applicationId the request won't be sent

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
jonathanmos: I've modified back to storing the applicationId along with the resource. The filename and application will then be properties of a JsonObject stored as the bytearray in rawBatchEvent metadata

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
jonathanmos: I've changed to check for the number of applicationIds - if 0 then throw an invalidPayloadException, if greater than 1 then determine the most "popular" applicationId, filter the list of events by this id and send them in the request, and log the issue through the internalLogger

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/MiscUtils.kt
```diff
@@ -16,11 +16,25 @@ import com.datadog.android.sessionreplay.internal.recorder.GlobalBounds
...
     internal const val OPAQUE_ALPHA_VALUE = 255
 
+    @Suppress("SwallowedException")
+    internal fun safeDeserializeToJsonObject(data: ByteArray): JsonObject? {
+        val jsonString = String(data)
+        return try {
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            JsonParser.parseString(jsonString) as? JsonObject
+        } catch (e: JsonSyntaxException) {
+            null
```
0xnm, line 34: we should log

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/MiscUtils.kt
```diff
@@ -16,11 +16,25 @@ import com.datadog.android.sessionreplay.internal.recorder.GlobalBounds
...
+import com.google.gson.JsonParser
+import com.google.gson.JsonSyntaxException
 
 internal object MiscUtils {
 
     internal const val OPAQUE_ALPHA_VALUE = 255
 
+    @Suppress("SwallowedException")
+    internal fun safeDeserializeToJsonObject(data: ByteArray): JsonObject? {
+        val jsonString = String(data)
```
0xnm, line 29: if `data` is empty, we can do this check quickly and return `null`

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactoryTest.kt
```diff
@@ -0,0 +1,231 @@
...
+            data = fakeImageRepresentation.toByteArray(),
+            metadata = fakeMetaData.toString().toByteArray(Charsets.UTF_8)
+        )
+        val fakeListResources = listOf(fakeRawBatchEvent)
+
+        // When
+        assertThatThrownBy {
+            testedRequestBodyFactory.create(fakeListResources)
+        }.isInstanceOf(InvalidPayloadFormatException::class.java)
+            .hasMessage(UNABLE_GET_APPLICATION_ID_ERROR)
```
0xnm, line 142: formatting is here is a bit off

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactoryTest.kt
```diff
@@ -0,0 +1,231 @@
...
+            data = fakeImageRepresentation.toByteArray(),
+            metadata = fakeMetaData.toString().toByteArray(Charsets.UTF_8)
+        )
+        val fakeListResources = listOf(fakeRawBatchEvent)
+
+        // When
+        assertThatThrownBy {
+            testedRequestBodyFactory.create(fakeListResources)
+        }.isInstanceOf(InvalidPayloadFormatException::class.java)
+            .hasMessage(DESERIALIZE_METADATA_ERROR)
```
0xnm, line 223: formatting is off here as well

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactoryTest.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    @StringForgery
+    lateinit var fakeApplicationId: String
+
+    @Mock
+    lateinit var fakeDatadogContext: DatadogContext
+
+    @BeforeEach
+    fun `set up`(forge: Forge) {
+        val fakeRumFeature = HashMap<String, String>()
```
0xnm, line 78: minor suggestion. The same for other similar instances

```suggestion
        val fakeRumFeature = mapOf<String, String>(APPLICATION_ID to fakeApplicationId)
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    @Throws(InvalidPayloadFormatException::class)
+    internal fun create(
+        resources: List<RawBatchEvent>
+    ): RequestBody {
+        @Suppress("UnsafeThirdPartyFunctionCall") // Handled up in the caller chain
+        val builder = MultipartBody.Builder().setType(MultipartBody.FORM)
+        val applicationId = getApplicationId(resources)
+
+        resources.filter {
```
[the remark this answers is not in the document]
jonathanmos: good catch 🙏

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,164 @@
...
+            val groupedByApplicationId = resourcesMetaData.groupBy {
+                it.get(APPLICATION_ID_KEY)
+            }
+
+            groupedByApplicationId.forEach {
+                if (it.value.size > maxSize) {
+                    maxSize = it.value.size
+                    selectedApplicationId = it.key.asString
+                }
+            }
```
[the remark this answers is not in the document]
0xnm: Can `application_id` be changed during the SDK run? I doubt this can be the case, unless RUM feature is stopped and another one is registered, but we don't have such possibility right now.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,154 @@
...
+internal class ResourceRequestBodyFactory(
+    private val internalLogger: InternalLogger
+) {
+
+    @Throws(InvalidPayloadFormatException::class)
+    internal fun create(
+        resources: List<RawBatchEvent>
+    ): RequestBody {
+        val deserializedResources: List<ResourceEvent> = resources.mapNotNull {
+            val deserializedObject = MiscUtils.safeDeserializeToJsonObject(internalLogger, it.metadata)
```
0xnm, line 30:
```suggestion
        val deserializedResources = resources.mapNotNull {
            val resourceMetadata = MiscUtils.safeDeserializeToJsonObject(internalLogger, it.metadata)
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,154 @@
...
+            val groupedByApplicationId = resources.groupBy {
+                it.applicationId
+            }
+
+            groupedByApplicationId.forEach {
+                if (it.value.size > maxSize) {
+                    maxSize = it.value.size
+                    selectedApplicationId = it.key
+                }
+            }
```
0xnm, line 99: this can be simplified:

```suggestion
           selectedApplicationId = resources
                   .groupBy { it.applicationId }
                   .maxBy { it.value.size }
                   .key
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/ResourceEvent.kt
```diff
@@ -0,0 +1,31 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.base64
```
0xnm, line 7: should it be in the `recorder` or `base64` package? Seems it doesn't belong to any of them, it is just an intermediary object for the upload phase.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/ResourceEventForgeryFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+import java.util.UUID
+
+internal class ResourceEventForgeryFactory : ForgeryFactory<ResourceEvent> {
+    override fun getForgery(forge: Forge): ResourceEvent {
+        return ResourceEvent(
+            applicationId = forge.getForgery<UUID>().toString(),
+            identifier = forge.getForgery<UUID>().toString(),
+            resourceData = forge.getForgery<UUID>().toString().toByteArray()
```
0xnm, line 19: seems like copy-paste error here, it should be just a string, not necessarily UUID

```suggestion
            resourceData = forge.aString().toByteArray()
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestBodyFactoryTest.kt
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    @Test
+    fun `M return valid requestBody W create()`() {
+        // Given
+        val fakeRawBatchEvent = RawBatchEvent(
+            data = fakeImageRepresentation.toByteArray(),
+            metadata = fakeMetaData.toString().toByteArray(Charsets.UTF_8)
+        )
+        val fakeListResources = listOf(fakeRawBatchEvent)
```
0xnm, line 83: What prevents us from generating multiple items for this list? This will be real use-case scenario. If I understand it correctly, single batch may contain multiple images.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/domain/ResourceRequestFactory.kt
```diff
@@ -0,0 +1,85 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.domain
```
0xnm, line 7: I believe all upload related classes shouldn't be in the `domain` package, but in `net` package, this is at least what we follow for other features.

0xnm (commented):
nice work, I left some final comments to address. Nothing blocking, but I think it is better to handle them before the merge.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,147 @@
...
+                }
+            } else {
+                null
+            }
+        }
+
+        if (deserializedResources.isEmpty()) {
+            @Suppress("ThrowingInternalException")
+            throw InvalidPayloadFormatException(NO_RESOURCES_TO_SEND_ERROR)
+        }
```
xgouchet, line 60: > [!WARNING]
> Please let's not throw exceptions here. We're not implementing an interface so we can decide what the signature is, it's better to return a null here than throw an exception. 
> 
> First, creating an exception means that the VM needs to build the stack trace which is costly.
> Second, even with the `@Throws` annotation, there's no checked exceptions in Kotlin, meaning we don't have a way to enforce that the caller use a `try/catch` block, which could lead to us crashing the host app.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,147 @@
...
+
+        val applicationId = getApplicationId(deserializedResources)
+
+        @Suppress("UnsafeThirdPartyFunctionCall") // Handled up in the caller chain
```
xgouchet, line 64: > [!WARNING]
> It might be handled by the caller in the current implementation right now, but you can't know for sure it will be in the future. Again this could lead to future unforeseen crashes

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,147 @@
...
+
+    private fun getApplicationId(resources: List<ResourceEvent>): String {
+        val applicationIds = resources.map {
+            it.applicationId
+        }
+
+        var selectedApplicationId = applicationIds[0]
+
+        if (applicationIds.size > 1) {
+            // more than one applicationId, so take the one from the largest group
```
xgouchet, line 81: > [!NOTE]
> What's the reason we use this vs the first or last application Id?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
```diff
@@ -0,0 +1,147 @@
...
+
+        val applicationId = getApplicationId(deserializedResources)
+
+        @Suppress("UnsafeThirdPartyFunctionCall") // Handled up in the caller chain
+        val builder = MultipartBody.Builder().setType(MultipartBody.FORM)
+
+        addResourcesSection(builder, deserializedResources)
+        addApplicationIdSection(builder, applicationId)
+        @Suppress("UnsafeThirdPartyFunctionCall") // Handled up in the caller chain
+        return builder.build()
```
xgouchet, line 70: > [!WARNING]
> Same as above, each of these line can throw, better have a `try/catch` here

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/MiscUtils.kt
```diff
@@ -12,14 +12,62 @@ import android.graphics.Point
...
     internal const val OPAQUE_ALPHA_VALUE = 255
+    private const val GET_STRING_FROM_JSON_ERROR = "Error getting string property from json"
+    internal const val DESERIALIZE_JSON_ERROR = "Error deserializing json object"
+
+    internal fun safeDeserializeToJsonObject(internalLogger: InternalLogger, jsonByteArray: ByteArray): JsonObject? {
+        if (jsonByteArray.isEmpty()) return null
+
+        val jsonString = String(jsonByteArray)
+        return try {
+            @Suppress("UnsafeThirdPartyFunctionCall")
```
xgouchet, line 35: > [!WARNING]
> Why are you suppressing the check here? You should either catch everything possible, since it's a `safe…` function, or add a comment to tell why we're suppressing it

xgouchet (changes requested):
Please let's make sure we don't overuse the `@Supress` annotation. 
Those checks were put in place for a reason: making sure that we never crash the host application. We can use some of those but we need to make sure that it is always in a 100% safe way.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
jonathanmos: My concern is that if we use the first or last applicationId we potentially end up discarding a majority of resources in some cases - an extreme example would be 9 resources with one applicationId followed by one resource with another. We would lose 9 and only send 1.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
jonathanmos: In continuation of what we discussed, I've modified DataOkHttpUploader to accept a nullable request and giving an error upload status if null is received

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
jonathanmos: Changed to take the last applicationId in the list

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactoryTest.kt
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    @Test
+    fun `M return valid requestBody W create()`() {
+        // Given
+        val fakeRawBatchEvent = RawBatchEvent(
+            data = fakeImageRepresentation.toByteArray(),
+            metadata = fakeMetaData.toString().toByteArray(Charsets.UTF_8)
+        )
+        val fakeListResources = listOf(fakeRawBatchEvent)
```
0xnm, line 81: Still it is better to test the work with multiple `RawBatchEvent`, but it is not blocking for me. List of only 1 element is kind of edge-case and not generic.

0xnm (approved):
lgtm!

xgouchet (approved):
Nice work, just some nitpicking in the tests ;)

features/dd-sdk-android-logs/src/test/kotlin/com/datadog/android/log/internal/net/LogsRequestFactoryTest.kt
```diff
@@ -61,22 +61,22 @@ internal class LogsRequestFactoryTest {
         val request = testedFactory.create(fakeDatadogContext, batchData, batchMetadata)
 
         // Then
-        assertThat(request.url).isEqualTo(
+        assertThat(request?.url).isEqualTo(
```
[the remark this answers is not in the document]
jonathanmos: 🙏  I've changed all relevant places in the tests

0xnm (approved):

xgouchet (approved):

PR #1829, opened 2024-01-17
RUM-2875 Introduce the FeatureContextUpdateLitener API

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -54,6 +56,8 @@ internal class SdkFeature(
 ) : FeatureScope {
 
     internal val initialized = AtomicBoolean(false)
+    internal val contextUpdateListeners: ConcurrentHashMap<FeatureContextUpdateListener, Any?> =
```
mariusc83, line 59: not sure if I should use a ConcurrentHashMap here or just a Set and make a copy when notifying the listeners to make sure it is concurrent

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCore.kt
```diff
@@ -227,6 +236,38 @@ internal class DatadogCore(
         }
     }
 
+    override fun setContextUpdateListener(featureName: String, listener: FeatureContextUpdateListener) {
```
mariusc83, line 239: wanted to keep the same approach as for receiver but it makes more sense to move this logic into the `SdkFeature` itself.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
```diff
@@ -45,8 +45,8 @@ interface FeatureSdkCore : SdkCore {
      * with the given name registered, callback won't be called.
      */
     fun updateFeatureContext(
-        featureName: String,
-        updateCallback: (context: MutableMap<String, Any?>) -> Unit
+            featureName: String,
+            updateCallback: (context: MutableMap<String, Any?>) -> Unit
```
0xnm, line 49: better to revert, not part of the actual code change.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -54,6 +56,8 @@ internal class SdkFeature(
 ) : FeatureScope {
 
     internal val initialized = AtomicBoolean(false)
+    internal val contextUpdateListeners: ConcurrentHashMap<FeatureContextUpdateListener, String> =
+        ConcurrentHashMap()
```
0xnm, line 60: Seems like it is used like a `Set`, and value is actually never used. In this case it can be just `Collections.newSetFromMap(ConcurrentHashMap())` and placeholder for he value is not needed.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -166,6 +170,39 @@ internal class SdkFeature(
...
+    // region Context Update Listener
+
+    internal fun notifyContextUpdated(context: Map<String, Any?>) {
+        contextUpdateListeners.keys.forEach {
+            it.onContextUpdate(context)
+        }
+    }
+
+    internal fun setContextUpdateListener(listener: FeatureContextUpdateListener) {
+        synchronized(contextUpdateListeners) {
```
0xnm, line 182: Do we actually need `ConcurrentHashMap` then if each access that modifies the state of `contextUpdateListeners` is backed by `synchronized`?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -166,6 +170,39 @@ internal class SdkFeature(
...
+            // the argument is always non - null, so we can suppress the warning
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            if (contextUpdateListeners.contains(listener)) {
+                internalLogger.log(
+                    InternalLogger.Level.WARN,
+                    InternalLogger.Target.USER,
+                    { CONTEXT_UPDATE_LISTENER_ALREADY_EXISTS.format(Locale.US, wrappedFeature.name) }
```
0xnm, line 189: But we don't actually overwrite anything, do we? Both key and value stay the same.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCore.kt
```diff
@@ -197,6 +198,8 @@ internal class DatadogCore(
                 val mutableContext = featureContext.toMutableMap()
                 updateCallback(mutableContext)
                 it.setFeatureContext(featureName, mutableContext)
+                feature.notifyContextUpdated(mutableContext)
+                mutableContext
```
0xnm, line 202: why to we need this line?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureContextUpdateListener.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Listener for feature context updates.
+ */
+fun interface FeatureContextUpdateListener {
+
+    /**
+     * Called when the context for a feature is updated.
+     * @param event the updated context
+     */
+    @AnyThread
+    fun onContextUpdate(event: Map<String, Any?>)
```
0xnm, line 21: In this case, generally speaking, we cannot reuse the same listener object for context updates of the different features.

What if we add a feature name as a first argument? That way we could re-use the same object.

Not an ask or blocker for me, just some brainstorming.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
mariusc83: yes, wanted to use the same thing but not available for API 21 :(

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
mariusc83: yes we need it as I did not want to `synchronize` when we notify the listeners. When I travers them for notification another thread might remove or add a new one and it will get into `ConcurrentModificationException`

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
mariusc83: yes you are right...unless you are providing a different instance of the same listener. The `equals` method will say true and we are replacing the previous instance with the new one.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCore.kt
mariusc83: oh left over sorry

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
0xnm: It should be available https://developer.android.com/reference/java/util/Collections#newSetFromMap(java.util.Map%3CE,java.lang.Boolean%3E)

mariusc83: ah yes...that's not what I tested, I tested the `Collections.newKeySet` which is a new method, but indeed that one would work also. Tks !!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCore.kt
```diff
@@ -197,6 +198,11 @@ internal class DatadogCore(
                 val mutableContext = featureContext.toMutableMap()
                 updateCallback(mutableContext)
                 it.setFeatureContext(featureName, mutableContext)
+                // notify all the other features
+                features.filter { it.key != featureName }
+                    .forEach { (_, feature) ->
+                        feature.notifyContextUpdated(featureName, mutableContext)
```
0xnm, line 204: even though the type on the receiver side is `Map` here, maybe it is worth to make a copy as a safe-guard?

```suggestion
                        feature.notifyContextUpdated(featureName, mutableContext.toMap())
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -319,6 +359,8 @@ internal class SdkFeature(
     // endregion
 
     companion object {
+        internal const val CONTEXT_UPDATE_LISTENER_ALREADY_EXISTS =
+            "Feature \"%s\" already has this listener registered, overwriting it."
```
0xnm, line 363: Still there is no overwrite, it is already there. Maybe we can drop the part about `overwriting`?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -382,6 +396,56 @@ internal class DatadogCoreTest {
...
+        // Then
+        mockInternalLogger.verifyLog(
+            InternalLogger.Level.WARN,
+            InternalLogger.Target.USER,
+            DatadogCore.MISSING_FEATURE_FOR_CONTEXT_UPDATE_LISTENER.format(Locale.US, feature)
+        )
+    }
+
+    @Test
+    fun `𝕄 remove context update listener 𝕎 setContextUpdateListener()`(
```
0xnm, line 434:
```suggestion
    fun `𝕄 remove context update listener 𝕎 removeContextUpdateReceiver()`(
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -549,6 +551,133 @@ internal class SdkFeatureTest {
...
+            testedFeature.removeContextUpdateListener(it)
+        }
+
+        // Then
+        assertThat(testedFeature.contextUpdateListeners.toTypedArray())
+            .containsExactlyInAnyOrderElementsOf(remainingListeners)
+    }
+
+    @Test
+    fun `M update registered listeners W notifyContextUpdateListener()`(forge: Forge) {
```
0xnm, line 661:
```suggestion
    fun `M update registered listeners W notifyContextUpdated()`(forge: Forge) {
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -549,6 +551,133 @@ internal class SdkFeatureTest {
...
+        val countDownLatch = CountDownLatch(mockListeners.size)
+        mockListeners.forEach {
+            Thread {
+                testedFeature.setContextUpdateListener(it)
+                countDownLatch.countDown()
+            }.start()
+        }
+
+        // Then
+        countDownLatch.await(5, TimeUnit.SECONDS)
```
0xnm, line 571: Alternatively this can be written without `CountDownLatch`:

```suggestion
        mockListeners.map {
            Thread {
                testedFeature.setContextUpdateListener(it)
                countDownLatch.countDown()
            }.apply { start() }
        }.forEach { it.join(5, TimeUnit.SECONDS) }

        // Then
```

but not a big deal

0xnm (dismissed):
lgtm, I left some suggestions

jonathanmos (approved):

xgouchet (approved):

PR #1831, opened 2024-01-19
RUM-2485 Provide the parent container information for browser rum events

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
+++ b/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
@@ -28,7 +28,9 @@ import com.datadog.android.webview.internal.replay.WebViewReplayFeature
 import com.datadog.android.webview.internal.rum.TimestampOffsetProvider
 import com.datadog.android.webview.internal.rum.WebViewRumEventConsumer
 import com.datadog.android.webview.internal.rum.WebViewRumEventContextProvider
+import com.datadog.android.webview.internal.rum.WebViewRumEventMapper
 import com.datadog.android.webview.internal.rum.WebViewRumFeature
+import com.datadog.android.webview.internal.rum.domain.NoOpNativeRumViewsCache
 import com.datadog.android.webview.internal.storage.NoOpDataWriter
 
 /**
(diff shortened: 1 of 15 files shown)


features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
```diff
@@ -137,6 +141,12 @@ object WebViewTracking {
     }
 
     private fun resolveRumFeature(sdkCore: FeatureSdkCore): WebViewRumFeature? {
+        (
+        sdkCore.getFeature(WebViewRumFeature.WEB_RUM_FEATURE_NAME)
+                ?.unwrap<StorageBackedFeature>() as? WebViewRumFeature
```
0xnm, line 146: Is there a particular reason for doing `?.unwrap<StorageBackedFeature>() as? WebViewRumFeature` instead of the `?.unwrap<WebViewRumFeature>()`? Same for the similar calls below.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/domain/WebViewNativeRumViewsHandler.kt
```diff
@@ -0,0 +1,93 @@
...
+            eventTimestamp != null
+        ) {
+            val newEntry = Triple(activeViewId, eventTimestamp, hasReplay)
+            addToCache(newEntry)
+        }
+        purgeHistory()
+    }
+
+    private fun addToCache(
+        entry: Triple<String, Long, Boolean>
```
0xnm, line 54: I would suggest to create a dedicated class maybe for that.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
mariusc83: oh...nope I think I did that just because I've seen it that way in another place. Better how you suggested

mariusc83: I found the explanation why, is because in the core we are force casting it, we are not using `as?` :
```
    override fun <T : Feature> unwrap(): T = wrappedFeature as T
```

xgouchet (approved):

xgouchet (approved):

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/WebViewTracking.kt
0xnm: but it is us who is owning implementation anyway, so we if got the scope for the feature, we can safely unwrap with using `as` under the hood.

PR #1832, opened 2024-01-22
RUM-1702: Start session when RUM is initialized

0xnm:
### What does this PR do?

This change starts RUM session during the `Rum.enable` call. The general rule is like that:

* If application process has `foreground` type, then session will be started (with interaction time set at the moment of initialization), and `ApplicationLaunch` view with `application_started` event will be created.
* Otherwise, if application process has something different from `foreground` as type, session will be initialized only if background tracking is enabled, no `ApplicationLaunch` view will be created.

This PR is using the notion of `SdkInit` class, even though it is possible to use only `ApplicationStarted` event in order to be on-par with the similar logic in iOS SDK. Also it is still using event system for the session initialization in order to minimize the footprint of the change (it will be easier to rollback it if needed).

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/Rum.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/Rum.kt
@@ -11,6 +11,7 @@ import android.os.Looper
 import com.datadog.android.Datadog
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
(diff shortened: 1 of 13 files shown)


fuzzybinary (approved):
This LGTM. Agree we should have @ncreated take a look and make sure the logic is the same.

ncreated (approved):
It looks okay 👍. Good idea to map the new "sdk init" command into existing "app start" to minimise the change impact 👌.

xgouchet (approved):

PR #1837, opened 2024-01-23
RUM-2600: Add traversal flag to snapshot items

jonathanmos:
### What does this PR do?
Adds a flag to snapshot items that indicates that they have finished traversing the view tree. Also, increases the interval that snapshots are considered stale from 200ms to 1 second

### Motivation
The RecordedDataQueueHandler in some situations could attempt to process an item that was still performing a traversal. Because items are inserted into the queue at the beginning of their traversals without nodes, the handler could regard the item as invalid and drop it from the queue. With this change, items that haven't finished traversals will only be dropped if stale. 

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/SnapshotRecordedDataQueueItem.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/SnapshotRecordedDataQueueItem.kt
@@ -15,15 +15,22 @@ internal class SnapshotRecordedDataQueueItem(
     recordedQueuedItemContext: RecordedQueuedItemContext,
     internal val systemInformation: SystemInformation
 ) : RecordedDataQueueItem(recordedQueuedItemContext) {
-    internal var nodes = emptyList<Node>()
+    @Volatile internal var nodes = emptyList<Node>()
+
+    @Volatile internal var isFinishedTraversal = false
     internal var pendingJobs = AtomicInteger(0)
 
     override fun isValid(): Boolean {
+        if (!isFinishedTraversal) {
+            // item is always valid unless traversal has finished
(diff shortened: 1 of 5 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/SnapshotRecordedDataQueueItem.kt
```diff
@@ -16,14 +16,20 @@ internal class SnapshotRecordedDataQueueItem(
     internal val systemInformation: SystemInformation
 ) : RecordedDataQueueItem(recordedQueuedItemContext) {
     internal var nodes = emptyList<Node>()
+    internal var isFinishedTraversal = false
```
0xnm, line 19: if this one can be modified from the different threads, it should be either `@Volatile` or `AtomicBoolean`

jonathanmos: It's modified only on the main thread - created as false and modified to true after the traversal

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/SnapshotRecordedDataQueueItem.kt
0xnm: Same may be applied to the `nodes` property as well btw, maybe we should add `@Volatile`/`AtomicReference` there.

Anyway, passing mutable structures between the threads is already a dangerous idea.

0xnm (approved):

0xnm (approved):

PR #1843, opened 2024-01-26
Use shared dependencies license plugin

xgouchet:
### What does this PR do?

Use the shared dependency license plugin replacing the one embedded in `buildSrc`


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 # Commons
-kotlin = "1.8.10"
-kotlinSP = "1.8.10-1.0.9"
+kotlin = "1.9.10"
```
0xnm, line 3: why do we need this change?

More over, we cannot use Kotlin 1.8.20 and above because of the following issue https://youtrack.jetbrains.com/issue/KT-60748 - we either need to apply workaround provided, or rename tests.

ambushwork:
I have same doubt as Nikita, otherwise LTGM!

mariusc83 (approved):
LGTM

xgouchet:
@ambushwork 
> I have same doubt as Nikita, otherwise LTGM!
All our tests have been renamed since, so we're good to go :)

ambushwork (approved):

PR #1850, opened 2024-02-08
RUM-1633 Drop batch telemetry where duration or age have negative values

mariusc83:
### What does this PR do?

In order to prevent the graphs skewing in the telemetry dashboards we are dropping the metrics for which the batch age and batch duration values are negatives due to the device time manual modification or time zone change. 

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
@@ -84,12 +84,18 @@ internal class BatchMetricsDispatcher(
 
     // region Internal
 
+    @SuppressWarnings("ReturnCount")
     private fun resolveBatchDeletedMetricAttributes(
         file: File,
         deletionReason: RemovalReason
     ): Map<String, Any?>? {
         val fileCreationTimestamp = file.nameAsTimestampSafe(internalLogger) ?: return null
         val fileAgeInMillis = dateTimeProvider.getDeviceTimestamp() - fileCreationTimestamp
+        if (fileAgeInMillis < 0) {
+            // the device time was manually modified or the time zone changed
+            // we are dropping this metric to not skew our charts
+            return null
+        }
         return mapOf(
             TRACK_KEY to trackName,
             TYPE_KEY to BATCH_DELETED_TYPE_VALUE,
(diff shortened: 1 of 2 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -89,7 +90,7 @@ internal class BatchMetricsDispatcher(
         deletionReason: RemovalReason
     ): Map<String, Any?>? {
         val fileCreationTimestamp = file.nameAsTimestampSafe(internalLogger) ?: return null
-        val fileAgeInMillis = dateTimeProvider.getDeviceTimestamp() - fileCreationTimestamp
+        val fileAgeInMillis = max(0, (dateTimeProvider.getDeviceTimestamp() - fileCreationTimestamp))
```
0xnm, line 93: Then it would still skew, but towards the zero. maybe it is better to ignore than this metric if it is negative?
Or maybe we don't need to use `dateTimeProvider` here? Because file is created using `newFileName = System.currentTimeMillis().toString()`, which is not affected at all by NTP synchronization. Meaning that to calculate file age property, we still need to use `System.currentTimeMillis()` directly, this will be reliable, because we want duration, and not instant.

wdyt?

mariusc83: iOS has the same and we wanted to align Android also to make sure we have the same charts. I am ready to discuss this into the parking lot today during our daily.

0xnm: The thing is that file created using `System.currentTimeMillis()`, without NTP-adjustment (well, maybe there is adjustment done by the device). It means that for file age at least we need also to use `System.currentTimeMillis()` without NTP-adjustment, this __should__ guarantee that time behind these calls is monotonic.

mariusc83: exactly why we are using that here and not the NTP adjustment one. But no one can guarantee that by the time we send the metric the System time was not skewed or the time zone was not changed. This will give a negative age. We've already noticed this behaviour in our data.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcherTest.kt
```diff
@@ -340,6 +366,38 @@ internal class BatchMetricsDispatcherTest {
...
+        @Forgery fakeMetadata: BatchClosedMetadata,
+        forge: Forge
+    ) {
+        // Given
+        val fakeFile: File = forge.forgeValidClosedFile()
+        whenever(fakeFile.name).thenReturn(
+            (
+            fakeMetadata.lastTimeWasUsedInMs +
+                    forge.aLong(min = 100, max = 1000)
+            ).toString()
```
0xnm, line 380: maybe just extract this into a variable? formatting looks strange here.

0xnm (approved):

jonathanmos (approved):

xgouchet (approved):

Issue #1836, opened 2024-01-23
User property from future set on View Loads


geranzo:
**Problem description**
We have a custom user property that we update with the list of screens in navigation back stack. This is done every time the current screen changes, whether navigating forward or backward, just before we call `RumMonitor::startView(...)` to report a new `View Load`. This is working fine for our custom actions, the user property seen in their attributes is being updated correctly. However we noticed this is not the case for the View Loads. Check out below the part of session, the View Load of `Home` screen at 4.36s has already a `user_nav_back_stack` user property updated with the data that was set over a minute later, at 1 min 21 s when the `Profile` View Load happened.

![image](https://github.com/DataDog/dd-sdk-android/assets/4363310/36672c98-d270-4ad0-b730-29b8ccd50241)

**Describe what you expected:**
We'd expect custom user properties to be up to date for View Loads in the same manner as they are for the custom actions.

**Additional context**
 - Datadog SDK version: v2.5.0
 - Datadog Gradle version: v1.12.0

0xnm:
Hello @geranzo!

Can you please share the way you set these user properties?

Is it done like that:

```
val propertiesA = mapOf("a" to "b")
GlobalRumMonitor.get().addAction(..., propertiesA)
val propertiesB = mapOf("c" to "d")
GlobalRumMonitor.get().startView(..., propertiesB)
```
or you are utilizing `Datadog.setUserInfo` to set them?

Basically code snippets will be useful to understand the setup you have. By looking on the description, it seems that you also start/stop views manually, not using our built-in tracking strategies?

geranzo:
> Can you please share the way you set these user properties?

This is our method to set a user property:

```
    override fun setUserProperty(name: String, value: Any?) {
        if (value != null) {
            globalRum.get().addAttribute(name, value)
        } else {
            removeUserProperty(name)
        }
    }
```

> Basically code snippets will be useful to understand the setup you have. By looking on the description, it seems that you also start/stop views manually, not using our built-in tracking strategies?

Yes, we call `globalRum.get().startView(...)` ourselves on a callback from Compose Navigation library.

0xnm:
Thanks for a quick response. Can you please expand `removeUserProperty` method? Does it call `globalRum.get().removeAttribute`? And at which point do you call `setUserProperty` - is it only navigation events?

Also is it possible that calls to send custom action are coming from non-UI thread (shouldn't be an issue, but just for the understanding)?

[...]

PR #1851, opened 2024-02-09
RUM-3012 Do not update RUM View global properties after the view is stopped

mariusc83:
### What does this PR do?

Opened from [User property from future set on View Loads · Issue #1836 · DataDog/dd-sdk-android](https://github.com/DataDog/dd-sdk-android/issues/1836) 

As of now view has 3 states: active, stopped and complete. Stopped - view was stopped, because another view started or it was stopped explicitly, but it still exists in the scope hierarchy and will send view updates until all the child scopes are complete.

Imagine the following flow:

View A |--- Active ---|--- Stopped---|Complete
View B                |--- Active -------------....
Here we have View A which moves to the stopped state once View B starts.

Customer is using global RUM attributes, which are updated with the current screen name (or any attribute specific to the current active screen) on each screen transition.

Say on View B activation, global attributes receive foo:bar. The expectation of the customer is not to see foo:bar in View A  attributes (this screen is not active anymore), but it will be there, because View A updates will be sent even when View B is already active with the current state of the global attributes.

View updates for View A should stop reading global attributes if it is stopped, but view reduction process on the backend should keep attributes which were received for the view updates during the active state.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -58,7 +58,8 @@ internal open class RumViewScope(
 
     internal val url = key.url.replace('.', '/')
 
-    internal val attributes: MutableMap<String, Any?> = initialAttributes.toMutableMap()
+    internal val eventAttributes: MutableMap<String, Any?> = initialAttributes.toMutableMap()
+    private var globalAttributes: MutableMap<String, Any?> = resolveGlobalAttributes(sdkCore)
 
     private var sessionId: String = parentScope.getRumContext().sessionId
     internal var viewId: String = UUID.randomUUID().toString()
@@ -142,7 +143,6 @@ internal open class RumViewScope(
             it.putAll(getRumContext().toMap())
             it[RumFeature.VIEW_TIMESTAMP_OFFSET_IN_MS_KEY] = serverTimeOffsetInMs
         }
-        attributes.putAll(GlobalRumMonitor.get(sdkCore).getAttributes())
         cpuVitalMonitor.register(cpuVitalListener)
         memoryVitalMonitor.register(memoryVitalListener)
         frameRateVitalMonitor.register(frameRateVitalListener)
@@ -161,6 +161,7 @@ internal open class RumViewScope(
         event: RumRawEvent,
         writer: DataWriter<Any>
     ): RumScope? {
+        updateGlobalAttributes(sdkCore, event)
         when (event) {
             is RumRawEvent.ResourceSent -> onResourceSent(event, writer)
             is RumRawEvent.ActionSent -> onActionSent(event, writer)
@@ -279,7 +280,7 @@ internal open class RumViewScope(
                     )
                 }
             }
-            attributes.putAll(event.attributes)
+            eventAttributes.putAll(event.attributes)
             stopped = true
             sendViewUpdate(event, writer)
             sendViewChanged()
@@ -702,7 +703,6 @@ internal open class RumViewScope(
     @Suppress("LongMethod", "ComplexMethod")
     private fun sendViewUpdate(event: RumRawEvent, writer: DataWriter<Any>) {
         val viewComplete = isViewComplete()
-        attributes.putAll(GlobalRumMonitor.get(sdkCore).getAttributes())
         version++
 
         // make a local copy, so that closure captures the state as of now
@@ -735,7 +735,7 @@ internal open class RumViewScope(
         val isSlowRendered = resolveRefreshRateInfo(refreshRateInfo) ?: false
         // make a copy - by the time we iterate over it on another thread, it may already be changed
         val eventFeatureFlags = featureFlags.toMutableMap()
-        val eventAdditionalAttributes = attributes.toMutableMap()
+        val eventAdditionalAttributes = (eventAttributes + globalAttributes).toMutableMap()
 
         sdkCore.newRumEventWriteOperation(writer) { datadogContext ->
             val currentViewId = rumContext.viewId.orEmpty()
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -161,6 +161,7 @@ internal open class RumViewScope(
         event: RumRawEvent,
         writer: DataWriter<Any>
     ): RumScope? {
+        updateGlobalAttributes(sdkCore, event)
```
mariusc83, line 164: With this new logic we are not taking into account the current value of global attributes when we send the view update on `stop` view. Before we were always taking the last value. This might change the way the sdk works in our clients apps so we need to debate more on this one. @0xnm @plousada @xgouchet

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -896,8 +906,6 @@ internal open class RumViewScope(
         pendingActionCount++
         val rumContext = getRumContext()
 
-        val globalAttributes = GlobalRumMonitor.get(sdkCore).getAttributes().toMutableMap()
```
0xnm, line 899: should we still keep a local copy? like `val globalAttributes = globalAttributes.toMap()`.

Event will be created on another thread, and by the time it will be created `globalAttributes` may have some other value already, so it is better to have a local copy.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -850,6 +850,17 @@ internal open class RumViewScope(
             .submit()
     }
 
+    private fun updateGlobalAttributes(sdkCore: InternalSdkCore, event: RumRawEvent): Map<String, Any?> {
+        if (!stopped && event !is RumRawEvent.StopView) {
```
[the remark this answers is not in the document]
mariusc83: because as you explained in the description of the ticket this will propagate the global attribute in case the user started a new view. A new view start will stopped the previous one automatically sending a `stopView` if I remember correctly. If the user added a global attribute to be propagated with the new view at `start` the old view will get this new attribute also if the `stop` was not called before and it is triggered by the `start` of the new view.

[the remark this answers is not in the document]
0xnm: No, `StopView` event is created only from `RumMonitor#stopView`. If new view starts, it won't create `StopView` event under the hood of `RumMonitor#startView` (instead, view will be stopped if it receives `StartView`, so maybe you want to address this case?).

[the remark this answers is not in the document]
mariusc83: oh...damn...this is so complicated, yup in that case I will update the ticket, keep the stopView update and do not update it in case of `StartView` instead.

xgouchet (approved):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -1388,6 +1389,22 @@ internal class RumViewScopeTest {
...
+            mockViewChangedListener,
+            mockResolver,
+            mockCpuVitalMonitor,
+            mockMemoryVitalMonitor,
+            mockFrameRateVitalMonitor,
+            featuresContextResolver = mockFeaturesContextResolver,
+            trackFrustrations = fakeTrackFrustrations,
+            sampleRate = fakeSampleRate
+        )
+        mockSessionReplayContext(testedScope)
```
0xnm, line 1407: why session replay comes into play here?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7839,6 +7858,433 @@ internal class RumViewScopeTest {
...
```
0xnm, line 7899: same, not clear why it is needed here (same for the tests below)

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7839,6 +7858,433 @@ internal class RumViewScopeTest {
...
+
+        // When
+        val result = testedScope.handleEvent(
+            RumRawEvent.StopView(fakeKey, emptyMap()),
```
0xnm, line 7903: should we add attributes here to make sure that they are taken into account as well at the assertion stage?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7839,6 +7858,433 @@ internal class RumViewScopeTest {
...
+            mockFrameRateVitalMonitor,
+            featuresContextResolver = mockFeaturesContextResolver,
+            trackFrustrations = fakeTrackFrustrations,
+            sampleRate = fakeSampleRate
+        )
+        mockSessionReplayContext(testedScope)
+
+        // When
+        val result = testedScope.handleEvent(
+            RumRawEvent.StartView(fakeKey, emptyMap()),
```
0xnm, line 7956: in theory we cannot receive `StartView` event with the same key as the active scope already has (it should be the key inherited from another view)

0xnm (approved):
lgtm, I've added some comments, but they are not blocking

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
mariusc83: good point, I think it was a request left over from other tests but as I do not assess the session replay status here I think I can remove it.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
mariusc83: maybe let me think about it 🤔

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
mariusc83: yes that's correct, let me change that. For me was more important the attributes update trigger but yes to be 100% correct a different key here would be needed.

jonathanmos (approved):

PR #1854, opened 2024-02-14
RUM-2709: Implement resource capture during traversal

jonathanmos:
### What does this PR do?
Captures the bitmaps during traversal and writes them to batches

### Motivation
Third PR of the Resource endpoint 

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
@@ -19,6 +19,7 @@ import android.widget.TextView
 import android.widget.Toolbar
 import androidx.appcompat.widget.SwitchCompat
 import com.datadog.android.api.InternalLogger
+import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueHandler
 import com.datadog.android.sessionreplay.internal.recorder.base64.Base64LRUCache
 import com.datadog.android.sessionreplay.internal.recorder.base64.Base64Serializer
 import com.datadog.android.sessionreplay.internal.recorder.base64.BitmapPool
@@ -77,8 +78,12 @@ enum class SessionReplayPrivacy {
     MASK_USER_INPUT;
 
     @Suppress("LongMethod")
-    internal fun mappers(internalLogger: InternalLogger): List<MapperTypeWrapper> {
-        val base64Serializer = buildBase64Serializer()
+    internal fun mappers(
+        internalLogger: InternalLogger,
+        applicationId: String,
+        recordedDataQueueHandler: RecordedDataQueueHandler
+    ): List<MapperTypeWrapper> {
+        val base64Serializer = buildBase64Serializer(applicationId, recordedDataQueueHandler)
         val imageWireframeHelper = ImageWireframeHelper(logger = internalLogger, base64Serializer = base64Serializer)
         val uniqueIdentifierGenerator = UniqueIdentifierGenerator
 
@@ -167,11 +172,17 @@ enum class SessionReplayPrivacy {
         return mappersList
     }
 
-    private fun buildBase64Serializer(): Base64Serializer {
+    private fun buildBase64Serializer(
+        applicationId: String,
+        recordedDataQueueHandler:
+        RecordedDataQueueHandler
+    ): Base64Serializer {
         val bitmapPool = BitmapPool()
         val base64LRUCache = Base64LRUCache()
 
         val builder = Base64Serializer.Builder(
+            applicationId = applicationId,
+            recordedDataQueueHandler = recordedDataQueueHandler,
             bitmapPool = bitmapPool,
             base64LRUCache = base64LRUCache
         )
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/DataQueueHandler.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/DataQueueHandler.kt
@@ -1,0 +1,24 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.async
+
+import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal interface DataQueueHandler {
+    fun addResourceItem(
+        identifier: String,
+        applicationId: String,
+        resourceData: ByteArray
+    ): ResourceRecordedDataQueueItem?
+    fun addTouchEventItem(
+        pointerInteractions: List<MobileSegment.MobileRecord>
+    ): TouchEventRecordedDataQueueItem?
+    fun addSnapshotItem(systemInformation: SystemInformation): SnapshotRecordedDataQueueItem?
+    fun tryToConsumeItems()
+    fun clearAndStopProcessingQueue()
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/NoopDataQueueHandler.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/NoopDataQueueHandler.kt
@@ -1,0 +1,34 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.async
+
+import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal class NoopDataQueueHandler : DataQueueHandler {
+    override fun addResourceItem(
+        identifier: String,
+        applicationId: String,
+        resourceData: ByteArray
+    ): ResourceRecordedDataQueueItem? = null
+
+    override fun addTouchEventItem(
+    pointerInteractions: List<MobileSegment.MobileRecord>
+    ): TouchEventRecordedDataQueueItem? = null
+
+    override fun addSnapshotItem(
+    systemInformation: SystemInformation
+    ): SnapshotRecordedDataQueueItem? = null
+
+    override fun tryToConsumeItems() {
+        // noop
+    }
+
+    override fun clearAndStopProcessingQueue() {
+        // noop
+    }
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/NoopDataQueueHandler.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/NoopDataQueueHandler.kt
@@ -17,11 +17,11 @@ internal class NoopDataQueueHandler : DataQueueHandler {
     ): ResourceRecordedDataQueueItem? = null
 
     override fun addTouchEventItem(
-    pointerInteractions: List<MobileSegment.MobileRecord>
+        pointerInteractions: List<MobileSegment.MobileRecord>
     ): TouchEventRecordedDataQueueItem? = null
 
     override fun addSnapshotItem(
-    systemInformation: SystemInformation
+        systemInformation: SystemInformation
     ): SnapshotRecordedDataQueueItem? = null
 
     override fun tryToConsumeItems() {
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
@@ -9,10 +9,10 @@ package com.datadog.android.sessionreplay.internal.net
 import androidx.annotation.VisibleForTesting
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.storage.RawBatchEvent
-import com.datadog.android.sessionreplay.internal.processor.EnrichedResource.Companion.APPLICATION_ID_INTERNAL_KEY
-import com.datadog.android.sessionreplay.internal.processor.EnrichedResource.Companion.APPLICATION_ID_OUTER_KEY
 import com.datadog.android.sessionreplay.internal.processor.EnrichedResource.Companion.APPLICATION_ID_RESOURCE_KEY
+import com.datadog.android.sessionreplay.internal.processor.EnrichedResource.Companion.APPLICATION_KEY
 import com.datadog.android.sessionreplay.internal.processor.EnrichedResource.Companion.FILENAME_KEY
+import com.datadog.android.sessionreplay.internal.processor.EnrichedResource.Companion.ID_KEY
 import com.datadog.android.sessionreplay.internal.utils.MiscUtils
 import com.google.gson.JsonObject
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
@@ -126,8 +126,8 @@ internal class ResourceRequestBodyFactory(
     private fun addApplicationIdSection(builder: MultipartBody.Builder, applicationId: String) {
         val applicationIdOuter = JsonObject()
         val applicationIdInner = JsonObject()
-        applicationIdInner.addProperty(APPLICATION_ID_INTERNAL_KEY, applicationId)
-        applicationIdOuter.add(APPLICATION_ID_OUTER_KEY, applicationIdInner)
+        applicationIdInner.addProperty(ID_KEY, applicationId)
+        applicationIdOuter.add(APPLICATION_KEY, applicationIdInner)
         applicationIdOuter.addProperty(TYPE_KEY, TYPE_RESOURCE)
 
         @Suppress("TooGenericExceptionCaught")
@@ -154,7 +154,7 @@ internal class ResourceRequestBodyFactory(
 
         if (body != null) {
             builder.addFormDataPart(
-                name = NAME_RESOURCE,
+                name = NAME_EVENT,
                 filename = FILENAME_BLOB,
                 body = body
             )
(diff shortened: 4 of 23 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -35,11 +39,18 @@ internal class Base64Serializer private constructor(
...
+    private val md5HashGenerator: MD5HashGenerator,
+    private val recordedDataQueueHandler: DataQueueHandler,
+    private val applicationId: String
 ) {
     private var isBase64CacheRegisteredForCallbacks: Boolean = false
     private var isBitmapPoolRegisteredForCallbacks: Boolean = false
 
+    // resources previously sent in this session -
+    // optimization to avoid sending the same resource multiple times
+    private val previouslySentResources: MutableSet<String> =
```
0xnm, line 51: we can call this one simply `resourcesSent` (or maybe `resourcesSeen`)

```suggestion
    private val resourcesSent: MutableSet<String> =
```

can we also do the napkin math for this one? if I understand it correctly, value will be MD5 hash, which is 32 characters. If Android VM encoding used is UTF-16, than 2 bytes per character, meaning 64 bytes roughly (not counting service fields of the String object) per item. How many items do we expect here?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -127,13 +139,35 @@ internal class Base64Serializer private constructor(
...
 
-        val base64String = convertBitmapToBase64(
-            byteArray = byteArray,
-            bitmap = bitmap,
-            shouldCacheBitmap = shouldCacheBitmap
-        ).toByteArray(Charsets.UTF_8)
-        val resourceId = md5HashGenerator.generate(byteArray)?.toByteArray(Charsets.UTF_8)
-        val cacheData = CacheData(base64String, resourceId)
+        val resourceId = md5HashGenerator.generate(byteArray)
+        var base64String = "".toByteArray(Charsets.UTF_8)
```
0xnm, line 143: something is wrong here, why do we use empty string here and not a `byteArray`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -127,13 +139,35 @@ internal class Base64Serializer private constructor(
...
+                return
+            }
+
+            if (!previouslySentResources.contains(resourceId)) {
+                previouslySentResources.add(resourceId)
+                recordedDataQueueHandler.addResourceItem(
+                    identifier = resourceId,
+                    resourceData = byteArray,
+                    applicationId = applicationId
+                )
```
0xnm, line 163: this seems like a design flaw: imo the only job of serializer is to serialize, why do we need the queue handler here? Also by having this we are injecting `applicationId` into the serializer constructor, which looks strange (same reason: serializer should just serialize and return result), and to support this we need to pass `applicationId` through the mappers chain.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/EnrichedResource.kt
```diff
@@ -32,7 +32,9 @@ internal data class EnrichedResource(
     }
 
     internal companion object {
-        internal const val APPLICATION_ID_KEY = "application_id"
+        internal const val APPLICATION_ID_RESOURCE_KEY = "applicationId"
```
0xnm, line 35: is there any impact from `application_id` to `applicationId` change?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/EnrichedResource.kt
```diff
@@ -32,7 +32,9 @@ internal data class EnrichedResource(
...
+        internal const val APPLICATION_ID_OUTER_KEY = "application"
+        internal const val APPLICATION_ID_INTERNAL_KEY = "id"
```
0xnm, line 37: maybe we need some explanation what outer vs internal key means, not quite clear

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
```diff
@@ -196,7 +200,7 @@ internal class ResourceRequestBodyFactory(
         internal const val TYPE_KEY = "type"
         internal const val TYPE_RESOURCE = "resource"
         internal const val NAME_IMAGE = "image"
-        internal const val NAME_RESOURCE = "resource"
+        internal const val NAME_RESOURCE = "event"
```
0xnm, line 203: should it be renamed to `NAME_EVENT`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/NoopDataQueueHandler.kt
```diff
@@ -0,0 +1,34 @@
...
+        resourceData: ByteArray
+    ): ResourceRecordedDataQueueItem? = null
+
+    override fun addTouchEventItem(
+    pointerInteractions: List<MobileSegment.MobileRecord>
+    ): TouchEventRecordedDataQueueItem? = null
+
+    override fun addSnapshotItem(
+    systemInformation: SystemInformation
+    ): SnapshotRecordedDataQueueItem? = null
```
0xnm, line 25: formatting

```suggestion
    override fun addTouchEventItem(
        pointerInteractions: List<MobileSegment.MobileRecord>
    ): TouchEventRecordedDataQueueItem? = null

    override fun addSnapshotItem(
        systemInformation: SystemInformation
    ): SnapshotRecordedDataQueueItem? = null
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeatureTest.kt
```diff
@@ -184,13 +183,10 @@ internal class SessionReplayFeatureTest {
...
         verify(lambda).invoke(captor.capture(), any(), any())
-        if (ResourcesFeature.RESOURCES_ENDPOINT_ENABLED) {
-            assertThat(captor.firstValue)
-                .isInstanceOf(SessionReplayResourcesWriter::class.java)
-        } else {
-            assertThat(captor.firstValue)
-                .isInstanceOf(NoOpResourcesWriter::class.java)
-        }
+        // always a noop, since even when the resource feature flag is enabled
+        // it only changes to the implementation datawriter after onInitialise
```
0xnm, line 187: but there is `onInitialize` call above, which contradicts this comment

0xnm (commented):
I went though the production files and left some questions, didn't go through the tests yet.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: Probably we can remove this set and use the existing lrucache - if an item is in the lrucache we know we've seen it before. The disadvantage only being that an item may be evicted from the lru and we would send it again

jonathanmos: As many images as there are in the session. Perhaps we could take advantage of the existing lrucache to make the memory usage bounded, although since multiple drawables could have the same resourceId it would need to be something like a map of resourceIds to refcount, and only remove when we see no more references for the resourceId

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
```diff
@@ -166,11 +170,17 @@ enum class SessionReplayPrivacy {
         return mappersList
     }
 
-    private fun buildBase64Serializer(): Base64Serializer {
+    private fun buildBase64Serializer(
```
mariusc83, line 173: I think I asked before but shouldn't we get rid of the Base64 serializer already ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
0xnm: It is fine to keep this as well if we expect the total size to be small, that is why I'm curious about the estimations. Let's say we sent 1000 resources, each ID string takes, for example, 80 bytes in the memory. This is 80 Kb, which is okay-ish.

mariusc83 (commented):
I partially reviewed it but would need first some answers to figure it out the rest.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
jonathanmos: In the next pr I'm going to rename this class and remove the base64 logic. We'll still need the majority of the logic there - the fallback mechanism to resolve the bitmaps, and the caches.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/DataQueueHandler.kt
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.async
+
+import com.datadog.android.sessionreplay.internal.recorder.SystemInformation
+import com.datadog.android.sessionreplay.model.MobileSegment
+
+internal interface DataQueueHandler {
+    fun addResourceItem(
```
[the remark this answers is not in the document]
jonathanmos: Resolving the bitmap is on the mainthread, as we did previously. However, adding the resource item to the queue should be in the background since serializeBitmap is on a background thread

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/net/ResourceRequestBodyFactory.kt
jonathanmos: yes, makes more sense 👍 I'll rename

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/EnrichedResource.kt
jonathanmos: The api is for a nested object -> something like { application: { id: value }}. I can add a comment there, or perhaps rename to `APPLICATION_KEY` and `ID_KEY`?

0xnm: yes, this makes much more sense - these are just string literal which are supposed to be used as keys, no matter where.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: We need to refactor this class and to split up the responsibilities. In the next pr I'll in any case rename it and also see what can be extracted out into separate components (since most of the logic in the class concerns getting the bitmaps). Regarding the applicationId let's discuss and see if there's a better way, because the `addResourceItem` is happening asynchronously

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/EnrichedResource.kt
jonathanmos: there's no impact, because this key is only for the EnrichedResource and not sent. Originally I used the same key as in the api, but the applicationId part of the api is no longer a single key

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeatureTest.kt
jonathanmos: This was a bit of an issue for me in the test, because the relevant `onInitialise` is part of the `ResourcesFeature`, and here we're testing the `SessionReplayFeature`. The RF is also created inside SRF, so I don't hold a reference to call `onInitialise` on

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
jonathanmos: I'm passing an empty string in the wireframe, so holding an empty string in the cache (as a bytearray). This is only necessary in this pr, as the next one will deprecate base64 and we won't need to hold a value for it at all in the cache. Even so, the name of the variable is problematic - I'll change to make this clearer

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -103,6 +108,19 @@ internal class DrawableUtils(
         fun onFailure()
     }
 
+    @VisibleForTesting
+    internal fun getBitmapSizeLimit(requestedSizeInBytes: Int?): Int {
```
mariusc83, line 112: this is more like `resolveBitmapSizeLimit`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
mariusc83: Let's do this properly in the next PR, I do agree with @0xnm that we need to split and provide single responsibilities here.

mariusc83 (commented):
Once @0xnm comments will be addressed I am ok to approve it. For me the biggest blocker was the `BaS

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
jonathanmos: This method decides between one of two constants (or returns the input if the input wasn't null) - is it not a getter? To me, 'resolve' implies that it might not return a result or that work needs to be done to reach the result.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/EnrichedResource.kt
0xnm, line 35:
```suggestion
        internal const val APPLICATION_ID_KEY = "applicationId"
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/base64/Base64Serializer.kt
```diff
@@ -35,11 +39,19 @@ internal class Base64Serializer private constructor(
...
+    private val recordedDataQueueHandler: DataQueueHandler,
+    private val applicationId: String
 ) {
     private var isBase64CacheRegisteredForCallbacks: Boolean = false
     private var isBitmapPoolRegisteredForCallbacks: Boolean = false
 
+    // resources previously sent in this session -
+    // optimization to avoid sending the same resource multiple times
+    // atm this set is unbounded but expected to use relatively little space (~80kb per 1k items)
+    private val resourcesSeen: MutableSet<String> =
```
0xnm, line 52: minor: after a bit of more thinking, let's emphasise that we store IDs here, and not some values, by renaming to `resourceIdsSeen`. wdyt?

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandlerTest.kt
```diff
@@ -393,6 +413,38 @@ internal class RecordedDataQueueHandlerTest {
         verifyNoMoreInteractions(mockProcessor)
     }
 
+    @Test
+    fun `M remove item from queue W tryToConsumeItems() { invalid resource item }`() {
+        // Given
+        val spy = spy(fakeResourceItem)
+
+        doReturn(false).whenever(spy).isValid()
```
0xnm, line 421: can we just create `fakeResourceItem` which will have `isValid` evaluated to `false` instead of involving `spy`? not a big deal, but usually `spy` is kind of hack.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/processor/EnrichedResourceTest.kt
```diff
@@ -39,4 +52,39 @@ internal class EnrichedResourceTest : ObjectTest<EnrichedResource>() {
     override fun createUnequalInstance(source: EnrichedResource, forge: Forge): EnrichedResource {
         return forge.getForgery()
     }
+
+    @Test
+    fun `M return valid binary metadata W asBinaryMetadat`(
```
0xnm, line 57:
```suggestion
    fun `M return valid binary metadata W asBinaryMetadata`(
```

0xnm (approved):
lgtm! I left some comments, but nothing blocking.

mariusc83 (approved):

PR #1858, opened 2024-02-18
RUM-2708: Enable Resource Endpoint by default

jonathanmos:
### What does this PR do?
• Renames `base64` packages to `resources`
• Renames resource classes
• Removes base64 from the lrucache
• Removes the Resource Endpoint feature flag. 

### Motivation
Final PR for the Resource Endpoint

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesLRUCache.kt
```diff
@@ -125,26 +121,18 @@ internal class Base64LRUCache(
     }
 }
 
-internal data class CacheData(val base64Encoding: ByteArray, var resourceId: ByteArray?) {
+internal data class CacheData(val resourceId: ByteArray) {
```
0xnm, line 124: do we need a class with only single property? seems like we can get rid of it.

0xnm (approved):
great work!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesLRUCache.kt
jonathanmos: Good point. Removing the base64 indeed removes our need for this container class

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializer.kt
```diff
@@ -90,14 +99,15 @@ internal class Base64Serializer private constructor(
 
     // region private
 
+    @Suppress("ReturnCount")
     @WorkerThread
     private fun serializeBitmap(
```
mariusc83, line 104: why is this called serializeBitmap ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializer.kt
```diff
@@ -123,33 +133,48 @@ internal class Base64Serializer private constructor(
...
+            bitmapPool?.put(bitmap)
         }
 
+        if (resourceId == null) {
+            // resourceId is mandatory for resource endpoint
+            resourcesSerializerCallback.onReady()
+            return
+        }
+
+        if (!resourceIdsSeen.contains(resourceId)) {
```
mariusc83, line 152: shouldn't we do the check before ? Are we computing the `byteArray` for this resource for nothing if the resource was already seen ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializer.kt
```diff
@@ -17,29 +17,38 @@ import androidx.annotation.VisibleForTesting
...
 import com.datadog.android.sessionreplay.model.MobileSegment
+import java.util.Collections
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.LinkedBlockingDeque
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 
 @Suppress("TooManyFunctions")
-internal class Base64Serializer private constructor(
+internal class ResourcesSerializer private constructor(
```
mariusc83, line 32: I think this class should not be called a serializer ? I think the scope change into something different. For me this class has too many responsibilities now and it is so hard to track the logic inside. We are using it to check if the resource already exists in cache was already solved before and also in the same time to update the queue and mark the wireframe as resolved. These are 2 different responsibilities and should be extracted in separated components. Also there is nothing that suggests that this is a serializer for me here.

mariusc83: I do not mind if we do this in another PR but I really think we need to do it now and simplify this logic a bit more.

mariusc83 (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializer.kt
jonathanmos: It's renamed in the next pr to ResourceResolver and some responsibilities are extracted out

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializer.kt
jonathanmos: We're calculating the md5 hash from the bytearray, so I'm not sure we can move this earlier in the flow

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializer.kt
jonathanmos: This is renamed in the next PR to resolveResourceHash

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesLRUCacheTest.kt
```diff
@@ -38,32 +38,23 @@ import org.mockito.quality.Strictness
...
     @BeforeEach
-    fun setup(forge: Forge) {
-        val fakeResourceId = forge.aNullable { aString() }
-        val fakeBase64ByteArray = fakeBase64Encoding.toByteArray(Charsets.UTF_8)
-        val fakeResourceIdByteArray = fakeResourceId?.toByteArray(Charsets.UTF_8)
-        fakeCacheData = CacheData(fakeBase64ByteArray, fakeResourceIdByteArray)
-        internalCache = LruCache<String, CacheData>(MAX_CACHE_MEMORY_SIZE_BYTES)
-        testedCache = Base64LRUCache(
+    fun setup() {
+        internalCache = LruCache<String, ByteArray>(MAX_CACHE_MEMORY_SIZE_BYTES)
```
0xnm, line 56: it seems this one can be simply a local variable rather than a property, it is used only in `setup` method

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/NoopDataQueueHandler.kt
```diff
@@ -17,11 +17,11 @@ internal class NoopDataQueueHandler : DataQueueHandler {
     ): ResourceRecordedDataQueueItem? = null
 
     override fun addTouchEventItem(
-        pointerInteractions: List<MobileSegment.MobileRecord>
+    pointerInteractions: List<MobileSegment.MobileRecord>
     ): TouchEventRecordedDataQueueItem? = null
 
     override fun addSnapshotItem(
-        systemInformation: SystemInformation
+    systemInformation: SystemInformation
```
0xnm, line 24: this change needs to be reverted

0xnm (approved):

0xnm (approved):

PR #1859, opened 2024-02-19
RUM-3246 Extract the dd-trace-core module from the java agent

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


features/dd-trace-core/build.gradle.kts
0xnm, line 1: I'm not sure if `dd-trace-core` should be in `features` folder, because it is not a feature directly, it is a module which is used by the feature module `dd-sdk-android-trace`.

But I don't have strong preference here anyway, I'm fine also with such location.

features/dd-trace-core/build.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+import com.datadog.gradle.config.javadocConfig
+import com.datadog.gradle.config.junitConfig
+import com.datadog.gradle.config.kotlinConfig
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+
+plugins {
+    // Build
+    id("com.android.library")
+    kotlin("android")
+    id("com.google.devtools.ksp")
```
0xnm, line 12: There is no Kotlin in this module, why this plugin is needed? Same for other Kotlin-related plugins and configurations in this file.

features/dd-trace-core/build.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+import com.datadog.gradle.config.dependencyUpdateConfig
+import com.datadog.gradle.config.javadocConfig
+import com.datadog.gradle.config.junitConfig
+import com.datadog.gradle.config.kotlinConfig
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+
+plugins {
+    // Build
+    id("com.android.library")
+    kotlin("android")
```
0xnm, line 11: why do we need Kotlin plugin if code is only in Java?

features/dd-trace-core/build.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+android {
+    defaultConfig {
+        consumerProguardFiles("consumer-rules.pro")
+    }
+    compileOptions {
+        isCoreLibraryDesugaringEnabled = true
+    }
+    namespace = "datadog"
```
0xnm, line 38: should it be `datadog.trace` at least?

features/dd-trace-core/build.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+    implementation(libs.androidXAnnotation)
+    implementation(libs.datadogSketchesJava)
+    implementation(libs.re2j)
+    compileOnly(libs.spotbugs)
+
+    // Generate NoOp implementations
+    ksp(project(":tools:noopfactory"))
+
+    // Lint rules
+    lintPublish(project(":tools:lint"))
```
0xnm, line 58: do we need this one?

features/dd-trace-core/lint.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lint>
+    <issue id="all">
+        <ignore path="*"/>
```
xgouchet, line 4: > [!WARNING]
> Is this really a good idea to ignore all lints? I get why we might want this while making the module but we should not commit this

features/dd-trace-core/src/main/java/datadog/appsec/api/blocking/Blocking.java
0xnm, line 1: I don't think we need files from `appsec` and `container` folders. Are they used?

xgouchet (commented):
This whole PR is huge, and hard to read, can we maybe split it in two. 
Also you didn't bring in any test, and I'm sure there are some tests written for this, rigth?

0xnm (commented):
I gave a quick look on the module setup in general and left some comments.

Also I think we can cleanup module further, because I see some properties/classes which are referencing different server-related technologies, which we don't need to have.

One more thing to do is to check initialization flow/time - probably this code tries to scan environment during the setup, and many things are absent on mobile, we can save some time removing redundant work.

features/dd-trace-core/src/test/resources/dd-trace-core.version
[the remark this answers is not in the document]
mariusc83: Yes I do agree but I plan to do this at the end once I have set up all the process and add the use case into the sample app. This way I can at least test after I remove some code. I am going to open a task in JIRA to just not forget about it.

gradle/libs.versions.toml
```diff
@@ -83,6 +83,17 @@ coroutines = "1.4.2"
 # Local Server
 ktor = "1.6.0"
 
+# Otel
+slf4j = "1.7.30"
+moshi = "1.11.0"
```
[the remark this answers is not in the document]
mariusc83: moshi is being use under the hood. This will mean re - writing all these classes and the only reason we have chosen this path was to actually not have to do this. I am going to try to see how I can remove it but I am not putting my hopes here as I do not have any unit test in place to be able to refactor this code.

gradle/libs.versions.toml
```diff
@@ -83,6 +83,17 @@ coroutines = "1.4.2"
 # Local Server
 ktor = "1.6.0"
 
+# Otel
+slf4j = "1.7.30"
```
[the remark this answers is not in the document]
mariusc83: this is just the `API` which under the hood it checks for the implementation and if it can't find any it defaults to the NoOp one. This package is used all over the Java code so it will be huge to start removing all its usage.

features/dd-trace-core/build.gradle.kts
mariusc83: Just a bad copy paste from others. Tks for noticing it.

features/dd-trace-core/build.gradle.kts
mariusc83: yeah I am working on that now...I did not want to change the package in the beginning to be able to update this code easily from the source one. Once I change the package will be a bit harder but I think we need to.

features/dd-trace-core/src/main/java/datadog/appsec/api/blocking/Blocking.java
mariusc83: Again like mentioned before I already removed lots of code from these module blindly so I would rather wait until the end once I maybe have also the unit tests in place or the Sample app use case to be able to test at least what I am doing.

mariusc83:
> This whole PR is huge, and hard to read, can we maybe split it in two. Also you didn't bring in any test, and I'm sure there are some tests written for this, rigth?

I know but honestly I would rather have it all in one shot, at least this PR which is the starting point and is mostly copy pasting and adapting code. Regarding the unit tests, yes they are all written in groovy so I will have to adapt them. They are not forgotten is just that this will be done later in a separated task ( I suspect this will be a trick task).

gradle/libs.versions.toml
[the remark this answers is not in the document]
0xnm: Is it used for some network transport? We have our own anyway, so we should replace it.

features/dd-trace-core/lint.xml
xgouchet: Since we're pulling the code into our repo, we become responsible for it, and must be sure that it's going to work and meet our reliability criterias. 

Ignoring **all** issues is a disaster waiting to happen. I agree we might not want to fix all the issues right now, but we need to make sure we didn't miss anything important. 

I agree that it will be tedious but let's have more granularity on which issues we suppress (with a JIRA ticket to fix it later) and which need to be fixed right now.

gradle/libs.versions.toml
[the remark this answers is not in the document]
mariusc83: yup will look into in future tasks once I will have more context here and more safety nets.

features/dd-trace-core/lint.xml
0xnm: I agree, let's see what we can address and what we cannot.

features/dd-trace-core/build.gradle.kts
0xnm: Well, it can be here or next to the `dd-sdk-android-core` (which is not in the `features`). Both locations may have their own reasoning.

features/dd-trace-core/build.gradle.kts
0xnm: I don't think it is related to the package, it is just a namespace afaik, and not necessarily should be the same as package path.

features/dd-sdk-android-trace/transitiveDependencies
```diff
@@ -1,13 +1,28 @@
 Dependencies List
 
 androidx.annotation:annotation:1.3.0                            :   30 Kb
+com.github.spotbugs:spotbugs-annotations:4.2.0                  :   14 Kb
+com.github.spotbugs:spotbugs:4.2.0                              :    3 Mb
```
0xnm, line 5: Is it a runtime or compile time dependency? If it exists in the final artifact, we need to remove it.

features/dd-sdk-android-trace/transitiveDependencies
```diff
@@ -1,13 +1,28 @@
...
+com.google.code.findbugs:jsr305:3.0.2                           :   19 Kb
 com.google.code.gson:gson:2.10.1                                :  276 Kb
 io.opentracing:opentracing-api:0.32.0                           :   18 Kb
 io.opentracing:opentracing-noop:0.32.0                          :   10 Kb
 io.opentracing:opentracing-util:0.32.0                          :   10 Kb
+net.jcip:jcip-annotations:1.0                                   :    2 Kb
+org.apache.bcel:bcel:6.5.0                                      :  678 Kb
+org.apache.commons:commons-lang3:3.11                           :  564 Kb
+org.apache.commons:commons-text:1.9                             :  211 Kb
```
0xnm, line 14: wondering if we can get rid of this, seems too heavy

features/dd-trace-core/api/apiSurface
0xnm, line 1: We may remove `apiSurface` plugin, there is no Kotlin code in this module.

features/dd-trace-core/build.gradle.kts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+dependencies {
+    coreLibraryDesugaring(libs.desugarJdk)
+    implementation(libs.slf4j)
+    implementation(libs.moshi)
+    implementation(libs.jctools)
+    implementation(libs.kotlin)
+    implementation(libs.okHttp)
+    implementation(libs.kotlin)
```
0xnm, line 45: `libs.kotlin` dependency is included twice. Also we need it only for tests, so it should be `testImplementation`.

features/dd-trace-core/build.gradle.kts
0xnm: I see Kotlin is used in tests, but still, there is nothing here which needs KSP support, this line can be removed.

features/dd-sdk-android-trace/transitiveDependencies
mariusc83: it is compile time

features/dd-sdk-android-trace/transitiveDependencies
mariusc83: yeah I have a task for that (the one that is required to fix that lint rule). I suspect is either this dep or moshi so I plan to remove both.

features/dd-trace-core/lint.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lint>
+    <issue id="DuplicatePlatformClasses">
+        <ignore path="*"/>
+    </issue>
+</lint>
```
0xnm, line 6: nit: missing new line at the EOF

0xnm (approved):
nice! I didn't run into source file content too much, but let's address other issues/try to cut more stuff in the next PRs.

xgouchet (approved):

PR #1870, opened 2024-02-21
RUM-2708: Refactor and split classes (do not merge this one)

jonathanmos:
This pr has unresolvable git conflicts. Maintaining it only to keep the comment history temporarily (until comments are resolved)

### What does this PR do?
• Extract cache logic into `BitmapCacheManager`
• Rename `ResourcesSerializer` to `ResourceResolver`
• Move `ResourceDataQueueItem` insertion into the `DataQueueHandler` to a new `ResourceItemCreationHandler`
• Cleanup the callbacks so that all exit points from the `ResourceResolver` come from the same method
• Add telemetry/maintainer logging to failures in MD5 hash generation and webp compression

### Motivation
Separate the refactor pr from the pr that removes the feature flag, to make it easier to review

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
```diff
@@ -79,11 +81,20 @@ enum class SessionReplayPrivacy {
...
+        val resourcesSerializer = buildResourcesSerializer(
+            applicationId,
+            recordedDataQueueHandler,
+            webPImageCompression
+        )
         val uniqueIdentifierGenerator = UniqueIdentifierGenerator
+        val imageWireframeHelper = ImageWireframeHelper(
+            resourceResolver = resourcesSerializer,
+            imageCompression = webPImageCompression
+        )
```
0xnm, line 97: Why do we include `webPImageCompression` into serializer and also into `ImageWireframeHelper`, which already has serializer included. Seems like some flaw in the design/responsibility leak (I guess it is needed only at the serialization step?).

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapCachesManager.kt
```diff
@@ -0,0 +1,93 @@
...
+        return bitmapPool.getBitmapByProperties(width, height, config)
+    }
+
+    internal class Builder(
+        private var resourcesLRUCache: Cache<Drawable, CacheData>,
+        private var bitmapPool: BitmapPool,
+        private var logger: InternalLogger
+    ) {
+
+        fun build(): BitmapCachesManager {
```
0xnm, line 85: do we need a builder here? It seems we can simply use constructor directly

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
```diff
@@ -0,0 +1,328 @@
...
+            // if we got here it means we saw the bitmap before,
+            // so we don't need to send the resource again
+            resourcesSerializerCallback.onReady(resourceId)
+            return
+        }
+
+        val bitmapFromDrawable = if (drawable is BitmapDrawable && shouldUseDrawableBitmap(drawable)) {
+            drawable.bitmap // cannot be null - we already checked in shouldUseDrawableBitmap
+        } else {
+        null
```
0xnm, line 68:
```suggestion
            null
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
```diff
@@ -0,0 +1,328 @@
...
+                    }
+
+                    override fun onFailed() {
+                        resourcesSerializerCallback.onReady(null)
+                    }
+                }
+            )
+        }.let {
+            threadPoolExecutor.executeSafe("resolveResourceId", logger, it)
+        }
```
0xnm, line 93: minor: since lambda comes last in the `executeSafe`, we can simply do the following syntax:

```
threadPoolExecutor.executeSafe("resolveResourceId", logger) {
      @Suppress("ThreadSafety") // this runs inside an executor
      createBitmapAsync(
          drawable = drawable,
          drawableWidth = drawableWidth,
          drawableHeight = drawableHeight,
          displayMetrics = displayMetrics,
          bitmapFromDrawable = bitmapFromDrawable,
          resolveResourceCallback = object : ResolveResourceCallback {
              override fun onResolved(resourceId: String, byteArray: ByteArray) {
                  resourceItemCreationHandler.queueItemIfNotPreviouslySeen(resourceId, byteArray)
                  resourcesSerializerCallback.onReady(resourceId)
              }

              override fun onFailed() {
                  resourcesSerializerCallback.onReady(null)
              }
          }
      )
}
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    // endregion
+
+    // region testing
+
+    @VisibleForTesting
+    internal fun getThreadPoolExecutor(): ExecutorService = threadPoolExecutor
```
0xnm, line 101: you can simply do `internal val` property, no need for extra method.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
```diff
@@ -0,0 +1,328 @@
...
+            BitmapCachesManager.Builder(
+                resourcesLRUCache,
+                bitmapPool,
+                logger
+            ).build(),
+        private var drawableUtils: DrawableUtils = DrawableUtils(
+            logger = logger,
+            bitmapCachesManager = bitmapCachesManager
+        ),
+        private var md5HashGenerator: MD5HashGenerator = MD5HashGenerator(logger)
```
0xnm, line 287: why `var` everywhere? cannot they be `val`? Also it seems we can simply use constructor directly instead of Builder?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializerCallback.kt
```diff
@@ -7,5 +7,5 @@
 package com.datadog.android.sessionreplay.internal.recorder.resources
 
 internal interface ResourcesSerializerCallback {
-    fun onReady()
+    fun onReady(resourceId: String?)
```
0xnm, line 10: in which case it can be null?

0xnm (commented):
I covered production files so far and left some questions

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayPrivacy.kt
jonathanmos: In the imageWireframeHelper we're using it to get the mimetype for the wireframe, and in the serializer to compress the image

0xnm: Shouldn't then serializer return MIME type with the piece serialized? Not a big deal though, just thinking.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesSerializerCallback.kt
jonathanmos: If we failed to resolve the resourceId. In that case we still want to decrement the pending images, but not to update the wireframe

0xnm: I would say if the flow is `If we failed to resolve the resourceId` then we maybe should have `onFailed` instead of `onReady` with `null` value, but not a blocker to me.

PR #1889, opened 2024-02-26
RUM-3338 Allow disabling ANR tracking

xgouchet:
### What does this PR do?

Add a configuration option to disable ANR tracking in the RUM feature configuration.

### Motivation

Our in-house ANR tracking provides information that are an estimation and not exactly the same as what the Google Play Store console reports. This makes it an optional feature to let customer disable it.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
@@ -243,6 +243,17 @@ data class RumConfiguration internal constructor(
         }
 
         /**
+         * Enable or disable ANR tracking.
+         * This tracking uses a custom heuristic to estimate ANRs, but might give different result than the
+         * Google Play Store.
+         * @param enabled whether ANR tracking should be enabled (default: true)
+         */
+        fun trackApplicationNotResponding(enabled: Boolean): Builder {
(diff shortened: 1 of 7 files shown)


0xnm:
This goes in parallel with the work about ANR changes (different way to detect them for Android 11 and above, introduction of fatal vs non-fatal ANRs which will bring new configuration method), so I would propose to wait a bit so that all these changes are aligned.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -242,6 +242,17 @@ data class RumConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Enable or disable ANR tracking.
+         * This tracking uses a custom heuristic to estimate ANRs, but might give different result than the
+         * Google Play Store.
+         * @param enabled whether ANR tracking should be enabled (default: true)
+         */
+        fun trackApplicationNotResponding(enabled: Boolean): Builder {
```
mariusc83, line 251: don't you need to add this also in the `TelemetryConfiguration` event ?

0xnm:
@xgouchet probably this one can be closed in favour of https://github.com/DataDog/dd-sdk-android/pull/1914

xgouchet:
Yep Agreed

PR #1903, opened 2024-03-05
RUM-3394 Provide the OtelDataWriter to be used in the OtelTracer

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
@@ -13,9 +13,12 @@ import com.datadog.android.api.feature.StorageBackedFeature
 import com.datadog.android.api.net.RequestFactory
 import com.datadog.android.api.storage.FeatureStorageConfiguration
 import com.datadog.android.trace.event.SpanEventMapper
+import com.datadog.android.trace.internal.data.NoOpOtelWriter
 import com.datadog.android.trace.internal.data.NoOpWriter
+import com.datadog.android.trace.internal.data.OtelTraceWriter
 import com.datadog.android.trace.internal.data.TraceWriter
 import com.datadog.android.trace.internal.domain.event.DdSpanToSpanEventMapper
+import com.datadog.android.trace.internal.domain.event.OtelDdSpanToSpanEventMapper
 import com.datadog.android.trace.internal.domain.event.SpanEventMapperWrapper
 import com.datadog.android.trace.internal.domain.event.SpanEventSerializer
 import com.datadog.android.trace.internal.net.TracesRequestFactory
@@ -33,6 +36,7 @@ internal class TracingFeature constructor(
 ) : StorageBackedFeature {
 
     internal var dataWriter: Writer = NoOpWriter()
+    internal var otelDataWriter: com.datadog.trace.common.writer.Writer = NoOpOtelWriter()
     internal val initialized = AtomicBoolean(false)
 
     // region Feature
@@ -41,6 +45,7 @@ internal class TracingFeature constructor(
 
     override fun onInitialize(appContext: Context) {
         dataWriter = createDataWriter(sdkCore)
+        otelDataWriter = createOtelDataWriter(sdkCore)
         initialized.set(true)
     }
 
@@ -67,7 +72,20 @@ internal class TracingFeature constructor(
         val internalLogger = sdkCore.internalLogger
         return TraceWriter(
             sdkCore,
-            legacyMapper = DdSpanToSpanEventMapper(networkInfoEnabled),
+            ddSpanToSpanEventMapper = DdSpanToSpanEventMapper(networkInfoEnabled),
+            eventMapper = SpanEventMapperWrapper(spanEventMapper, internalLogger),
+            serializer = SpanEventSerializer(internalLogger),
+            internalLogger = internalLogger
+        )
+    }
+
+    private fun createOtelDataWriter(
+        sdkCore: FeatureSdkCore
+    ): com.datadog.trace.common.writer.Writer {
+        val internalLogger = sdkCore.internalLogger
+        return OtelTraceWriter(
+            sdkCore,
+            ddSpanToSpanEventMapper = OtelDdSpanToSpanEventMapper(networkInfoEnabled),
             eventMapper = SpanEventMapperWrapper(spanEventMapper, internalLogger),
             serializer = SpanEventSerializer(internalLogger),
             internalLogger = internalLogger
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
@@ -14,17 +14,18 @@ import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.api.storage.EventBatchWriter
 import com.datadog.android.api.storage.RawBatchEvent
 import com.datadog.android.event.EventMapper
+import com.datadog.android.event.NoOpEventMapper
 import com.datadog.android.trace.internal.domain.event.ContextAwareMapper
 import com.datadog.android.trace.internal.storage.ContextAwareSerializer
 import com.datadog.android.trace.model.SpanEvent
-import com.datadog.legacy.trace.common.writer.Writer
-import com.datadog.opentracing.DDSpan
+import com.datadog.trace.common.writer.Writer
+import com.datadog.trace.core.DDSpan
 import java.util.Locale
 
-internal class TraceWriter(
+internal class OtelTraceWriter(
     private val sdkCore: FeatureSdkCore,
     internal val ddSpanToSpanEventMapper: ContextAwareMapper<DDSpan, SpanEvent>,
-    internal val eventMapper: EventMapper<SpanEvent>,
+    internal val eventMapper: EventMapper<SpanEvent> = NoOpEventMapper(),
     private val serializer: ContextAwareSerializer<SpanEvent>,
     private val internalLogger: InternalLogger
 ) : Writer {
@@ -37,19 +38,24 @@ internal class TraceWriter(
     override fun write(trace: MutableList<DDSpan>?) {
         if (trace == null) return
         sdkCore.getFeature(Feature.TRACING_FEATURE_NAME)
-            ?.withWriteContext { datadogContext, eventBatchWriter ->
-                trace.forEach { span ->
-                    @Suppress("ThreadSafety") // called in the worker context
-                    writeSpan(datadogContext, eventBatchWriter, span)
+                ?.withWriteContext { datadogContext, eventBatchWriter ->
+                    trace.forEach { span ->
+                        @Suppress("ThreadSafety") // called in the worker context
+                        writeSpan(datadogContext, eventBatchWriter, span)
+                    }
                 }
-            }
     }
 
-    override fun close() {
+    override fun flush(): Boolean {
+        // NO - OP
+        return true
+    }
+
+    override fun incrementDropCounts(p0: Int) {
         // NO - OP
     }
 
-    override fun incrementTraceCount() {
+    override fun close() {
         // NO - OP
     }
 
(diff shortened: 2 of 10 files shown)


features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
```diff
@@ -33,6 +36,7 @@ internal class TracingFeature constructor(
 ) : StorageBackedFeature {
 
     internal var dataWriter: Writer = NoOpWriter()
+    internal var otelDataWriter: com.datadog.trace.common.writer.Writer = NoOpOtelWriter()
```
0xnm, line 39: alternatively you can use `as` in the import to avoid clashing and using full name.

Also now we have `com.datadog.legacy.trace.common.writer.Writer` and `com.datadog.trace.common.writer.Writer` - what is the difference? I see that in `develop` branch `TraceWriter` implements `com.datadog.trace.common.writer.Writer`.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
```diff
@@ -73,4 +78,17 @@ internal class TracingFeature constructor(
...
         )
     }
+
+    private fun createOtelDataWriter(
+        sdkCore: FeatureSdkCore
+    ): com.datadog.trace.common.writer.Writer {
+        val internalLogger = sdkCore.internalLogger
+        return OtelTraceWriter(
+            sdkCore,
+            legacyMapper = OtelDdSpanToSpanEventMapper(networkInfoEnabled),
```
0xnm, line 88: If we are creating a new writer, why do we have the notion of `legacy` here?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
```diff
@@ -0,0 +1,92 @@
...
+    private val internalLogger: InternalLogger
+) : Writer {
+
+    // region Writer
+    override fun start() {
+        // NO - OP
+    }
+
+    override fun write(trace: MutableList<DDSpan>?) {
+        if (trace == null) return
```
0xnm, line 39: can it be null or just Kotlin inferred argument type as nullable by default?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    override fun close() {
+        // NO - OP
+    }
+
+    // endregion
+
+    @WorkerThread
+    private fun writeSpan(
```
0xnm, line 65: isn't it exactly the same code as in `TraceWriter`? Maybe we can re-use the common part?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    // endregion
+
+    // region internal
+
+    private fun resolveMetrics(event: DDSpan): SpanEvent.Metrics {
+        val metrics = resolveMetricsFromSpanContext(event).apply {
+            this["_dd.agent_psr"] = 1.0f
+            this["_sampling_priority_v1"] = 1
```
0xnm, line 62: why do we have it here, but not in `DdSpanToSpanEventMapper`?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
```diff
@@ -0,0 +1,136 @@
...
+            service = model.serviceName,
+            duration = model.durationNano,
+            start = model.startTime + serverOffset,
+            error = model.error.toLong(),
+            meta = metadata,
+            metrics = metrics
+        )
+    }
+
+    private fun resolveTraceId(model: DDSpan): String {
```
0xnm, line 43: this should be also in the `internal` region

0xnm, line 43: I think the whole method can be replace with `model.traceId.toHexString().takeLast(RUM_ENDPOINT_REQUIRED_TRACE_ID_LENGTH)`

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/OtelTracerBuilderProviderTest.kt
```diff
@@ -91,16 +97,44 @@ internal class OtelTracerBuilderProviderTest {
...
+        assertThat(tracer).isNotNull
         mockInternalLogger.verifyLog(
             InternalLogger.Level.ERROR,
             InternalLogger.Target.USER,
             OtelTracerProvider.TRACING_NOT_ENABLED_ERROR_MESSAGE
         )
     }
 
+    @Test
+    fun `M use a NoOpWriter W build { TracingFeature not enabled }`() {
```
0xnm, line 112:
```suggestion
    fun `M use a NoOpOtelWriter W build { TracingFeature not enabled }`() {
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapperTest.kt
```diff
@@ -0,0 +1,202 @@
...
+        val expectedMeta = fakeSpan.baggage + fakeSpan.tags.map {
+            it.key to it.value.toString()
+        }
+        val expectedMetrics = fakeSpan.expectedMetrics()
+
+        // When
+        val event = testedMapper.map(fakeDatadogContext, fakeSpan)
+
+        // Then
+        assertThat(event).hasSpanId(fakeSpan.spanId.toHexString())
```
0xnm, line 67: it seems we need to wrap lines of this assertion block, each `hasXXX` should be on the new line for readability. Same applies for other tests in this file.

0xnm (commented):
Looks good! I left some questions/suggestions.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
mariusc83: I moved the old classes last week (the ones in develop) in the legacy package in order to keep better track on what I have to remove later when we will only default on the `CoreTracer`

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
mariusc83: not sure why I added this back then I guess it is more `DdSpanToSpanEvent` mapper

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
mariusc83: I need to check that

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
mariusc83: I am trying to keep this separated as it is going to be better once I delete the old one later when we switch to the `CoreTracer`

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
mariusc83: because there is being added by us directly in the Java code the way I've seen

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
mariusc83: well it is not marked as @NonNull in the interface so I wanted just to make sure. Probably it will never be send as null but don't want to touch the Java code for now.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
0xnm: thanks for the explanation! clear enough. I missed this part about moving the classes.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
```diff
@@ -0,0 +1,130 @@
...
+        )
+    }
+
+    // endregion
+
+    // region internal
+
+    private fun resolveTraceId(model: DDSpan): String {
+        // the argument will never be negative
+        @Suppress("UnsafeThirdPartyFunctionCall")
```
0xnm, line 49: it is better to add the signature to the detekt file

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
```diff
@@ -0,0 +1,130 @@
...
+        return model.traceId.toHexString().takeLast(RUM_ENDPOINT_REQUIRED_TRACE_ID_LENGTH)
```
0xnm, line 50: is it safe to do that? I can think about other parts of the code which may read trace ID (for example some customer code, or the values we inject into RUM resources) as full (not truncated one), and send it to Datadog backend as well, in this case it will be 2 different trace IDs. Maybe for now we should generate 64 bit instead?

0xnm (approved):
lgtm. I left some questions, but they are not blocking.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
mariusc83: it was not there ? I thought it is already, let me check.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
mariusc83: yes, this is only for being able to comply with our endpoint in RUM which doesn't support the 128 bits ids. This will probably change once I will work on that ticket to align with what @ganeshnj was doing in iOS the other days. The first 64 bits are `0`s anyway and our backend doesn't support them as that.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/OtelDdSpanToSpanEventMapper.kt
mariusc83: it is already in the rules, here I am just discarding it because that argument will never be negative.

mariusc83: https://github.com/DataDog/dd-sdk-android/blob/develop/detekt_custom.yml#L224

PR #1909, opened 2024-03-12
RUM-2878: Report fatal ANRs

0xnm:
### What does this PR do?

This PR adds support of reporting fatal ANRs (ANR which resulted in the `Application Not Responding` dialog where user decided to close the application).

This functionality relies on the [ApplicationExitInfo](https://developer.android.com/reference/android/app/ApplicationExitInfo) API (available since Android API 30), which can be read on the next application run. So in that sense it is similar to the approach we have to report NDK crashes: on the next application run we need to run this information and match it with the last RUM view event we know.

Since we cannot mark a particular `ApplicationExitInfo` as read in the system, we have to store the timestamp of the last `ApplicationExitInfo` we processed to avoid the possible delivery on the next application run.

When processing `ApplicationExitInfo`, we will read trace information attached to it and try to parse it (it is non-structured log).

This PR doesn't add the possibility to disable reporting of non-fatal (or unknown) ANRs, and it doesn't add `error.category` info, this will be done in next PRs.


xgouchet (approved):
Nice job

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCore.kt
```diff
@@ -487,8 +502,8 @@ internal class DatadogCore(
         internal const val EVENT_RECEIVER_ALREADY_EXISTS =
             "Feature \"%s\" already has event receiver registered, overwriting it."
 
-        const val LAST_VIEW_EVENT_DIR_MISSING_MESSAGE = "Directory structure %s for writing" +
-            " last view event doesn't exist."
+        const val NO_NEED_TO_WRITE_LAST_VIEW_EVENT = "No need to write last RUM view event: NDK" +
```
jonathanmos, line 505: minor: should this be public? It doesn't look like it's being accessed outside the internal scope at the moment

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/anr/AndroidTraceParserTest.kt
```diff
@@ -0,0 +1,298 @@
+package com.datadog.android.rum.internal.anr
```
jonathanmos, line 1: minor: missing copyright

jonathanmos (approved):

PR #1932, opened 2024-03-21
RUM-3669 avoid feature flag spamming events

xgouchet:
### What does this PR do?

This PR attempts to improve the performance linked to setting FeatureFlag evaluations on a RUM View. Namely: 

- create an `addFeatureFlagBatchEvaluation()` method allowing to send a batch of feature flags instead of sending them once at a time. This limits the number of event the Rum View Scope has to process, and the number of view updates written on disk. 
- add a check in the `RumViewScope` handling the feature flags updates: instead of writing an event update for every call to the `addFeatureFlagEvaluation()` and `addFeatureFlagBatchEvaluation()` methods, it only writes an event if at least one feature flag value was changed. 

### Motivation

For some customer updating many feature flags at once, that would result in a lot of RawRumEvents being processed, and a lot of RUM View updates being written on disk. In some extreme cases it could lead to a lot of backpressure on the executor queues.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
@@ -284,6 +284,13 @@ interface RumMonitor {
     )
 
     /**
+     * Adds result of evaluating a set of feature flag to the view.
+     * Feature flag evaluations are local to the active view and are cleared when the view is stopped.
+     * @param featureFlags the map of feature flags
+     */
+    fun addFeatureFlagBatchEvaluation(featureFlags: Map<String, Any>)
+
+    /**
      * Adds a global attribute to all future RUM events.
      * @param key the attribute key (non null)
      * @param value the attribute value (or null)
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
@@ -7504,6 +7504,34 @@ internal class RumViewScopeTest {
     }
 
     @Test
+    fun `M send event only once W handleEvent(AddFeatureFlagEvaluation) on active view {same value}`(
+        @StringForgery flagName: String,
+        @StringForgery flagValue: String
+    ) {
+        // WHEN
+        testedScope.handleEvent(
+            RumRawEvent.AddFeatureFlagEvaluation(
+                name = flagName,
+                value = flagValue
+            ),
+            mockWriter
+        )
+        testedScope.handleEvent(
+            RumRawEvent.AddFeatureFlagEvaluation(
+                name = flagName,
+                value = flagValue
+            ),
+            mockWriter
+        )
+
+        // THEN
+        argumentCaptor<ViewEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture())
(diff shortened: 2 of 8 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -283,6 +283,13 @@ interface RumMonitor {
         value: Any
     )
 
+    /**
+     * Adds result of evaluating a set of feature flag to the view.
+     * Feature flag evaluations are local to the active view and are cleared when the view is stopped.
+     * @param featureFlags the map of feature flags
+     */
+    fun addFeatureFlagBatchEvaluation(featureFlags: Map<String, Any>)
```
0xnm, line 291: maybe we should deprecate `addFeatureFlagEvaluation` method in favour of this one? And call this one also `addFeatureFlagEvaluation`?

0xnm (approved):
I'm ok with the change, we just need to define its shape in terms of the pubilc API and align with iOS, I left an idea regarding that.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
ncreated: Both make sense IMO, they strike a balance between convenience (single flag) and performance (bulked flags). However, we could enhance the API comments in both to mention the performance benefit in batching FF evaluations.

My suggestion would be tho to name it `addFeatureFlagEvaluations` (plural). Keep the old one singular.

ncreated (approved):
Looks good 👍. Minor suggestion on API naming, non-blocking.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7594,6 +7622,165 @@ internal class RumViewScopeTest {
 
     // endregion
 
+    // region Feature Flags Batch
+
+    @Test
+    fun `M send event W handleEvent(AddFeatureFlagBatchEvaluation) on active view`(
```
0xnm, line 7628: class name should be updated here and in the tests below

```suggestion
    fun `M send event W handleEvent(AddFeatureFlagEvaluations) on active view`(
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitorTest.kt
```diff
@@ -1348,6 +1348,42 @@ internal class DatadogRumMonitorTest {
...
+
+            val event = firstValue as RumRawEvent.AddFeatureFlagEvaluation
+            assertThat(event.name).isEqualTo(name)
+            assertThat(event.value).isEqualTo(value)
+        }
+        verifyNoMoreInteractions(mockScope, mockWriter)
+    }
+
+    @Test
+    fun `M delegate event to rootScope W addFeatureFlagBatchEvaluation`(
```
0xnm, line 1370:
```suggestion
    fun `M delegate event to rootScope W addFeatureFlagEvaluations`(
```

0xnm (dismissed):

0xnm (approved):
lgtm. regarding the release, I think we will need to release 2.8.0 with `buildId` support soon (maybe in 2 weeks?), so maybe this change can be shipped there instead of hotfix.

PR #1938, opened 2024-03-25
RUM-3670 backpressure strategy

xgouchet:
### What does this PR do?

Create a BackPressure setting and BackPressure compatible BlockingQueue (will be used in a follow up PR)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
```diff
@@ -0,0 +1,91 @@
...
+            @Suppress("UnsafeThirdPartyFunctionCall") // can't have NPE here
+            super.offer(it)
+        }
+    }
+
+    override fun offer(e: E, timeout: Long, unit: TimeUnit?): Boolean {
+        @Suppress("UnsafeThirdPartyFunctionCall") // can't have NPE here
+        val accepted = super.offer(e, timeout, unit)
+        if (!accepted) {
+            return offer(e)
```
mariusc83, line 32: why do you still force it with the direct method if it is not accepted ?

mariusc83: is it only to get the telemetry ?

mariusc83 (dismissed):
Nicely done !!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -245,6 +247,19 @@ internal constructor(
...
 
+        /**
+         * Sets the strategy to handle scalability issues.
+         * Many operations (data processing, event I/O, …) are queued in background threads.
+         * This configuration lets one decide how to handle the edge case when the queue starts growing, which can lead
+         * to a lot of memory usage, delayed processing, and possibly OOM or ANR.
+         * @param backpressureStrategy the backpressure strategy (default strategy ignores new tasks if a queue reaches
+         * 1024 items)
+         */
+        fun setBackpressureStrategy(backpressureStrategy: BackPressureStrategy): Builder {
```
0xnm, line 258: do we want to add this in the core or only for RUM for now? Except of RUM no other feature is using a dedicated executor service for the event processing (they all use I/O executor services for writing events, though), and only RUM has a single thread to pull work items from the queue, while other services have multiple threads (so, in theory, they should consume the queue faster).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressureExecutorService.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.datadog.android.core.thread.FlushableExecutorService
+import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.TimeUnit
+
+internal class BackPressureExecutorService(
+    val logger: InternalLogger,
+    backpressureStrategy: BackPressureStrategy
+) : ThreadPoolExecutor(
+    CORE_POOL_SIZE,
+    CORE_POOL_SIZE,
```
0xnm, line 20: we need to indicate somehow that it is a single thread executor service - it is possible to understand that right now only by looking on the source code of this class.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/thread/FlushableExecutorService.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.configuration.BackPressureStrategy
+import java.util.concurrent.ExecutorService
+
+/**
+ * An [ExecutorService] which backing queue can be drained to a collection.
+ *
+ */
+interface FlushableExecutorService : ExecutorService {
```
0xnm, line 17: we can add `InternalApi` annotation here just in case

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
```diff
@@ -0,0 +1,91 @@
...
+        } else {
+            if (remainingCapacity() == 0) {
```
0xnm, line 34: what is the proposed usage for the `onThresholdReached` callback? it is to have some logs on the customer side?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
```diff
@@ -0,0 +1,91 @@
...
+                }
+
+                BackPressureMitigation.IGNORE_NEWEST -> {
+                    onItemDropped(e)
+                    true
+                }
+            }
+        } else {
+            if (remainingCapacity == 1) {
+                onThresholdReached()
```
0xnm, line 61: should we trigger this one? in theory, we may reach threshold, but there is no backpressure - by the time next element arrives it can fit into the queue.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
```diff
@@ -0,0 +1,91 @@
...
+            operation(e)
+        }
+    }
+
+    private fun onThresholdReached() {
+        backPressureStrategy.onThresholdReached()
+        logger.log(
+            level = InternalLogger.Level.WARN,
+            targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
+            messageBuilder = { "BackPressuredBlockingQueue reached capacity:${backPressureStrategy.capacity}" },
```
0xnm, line 72:
```suggestion
            messageBuilder = { "BackPressuredBlockingQueue reached capacity: ${backPressureStrategy.capacity}" },
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/thread/DropOldestBackPressuredBlockingQueueTest.kt
```diff
@@ -0,0 +1,420 @@
...
+        @StringForgery fakeNewItem: String
+    ) {
+        // Given
+        for (i in 0 until fakeBackPressureThreshold) {
+            testedQueue.put(fakeItemList[i % fakeItemList.size])
+        }
+
+        // When
+        Thread {
+            // PutInserts the specified element into this queue, waiting if necessary for space to become available.
```
0xnm, line 406:
```suggestion
            // Put inserts the specified element into this queue, waiting if necessary for space to become available.
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/thread/IgnoreNewestBackPressuredBlockingQueueTest.kt
```diff
@@ -0,0 +1,420 @@
...
+        @StringForgery fakeNewItem: String
+    ) {
+        // Given
+        for (i in 0 until fakeBackPressureThreshold) {
+            testedQueue.put(fakeItemList[i % fakeItemList.size])
+        }
+
+        // When
+        Thread {
+            // PutInserts the specified element into this queue, waiting if necessary for space to become available.
```
0xnm, line 406:
```suggestion
            // Put inserts the specified element into this queue, waiting if necessary for space to become available.
```

0xnm (commented):
great work! I added some questions for the better understanding.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/BackPressureMitigation.kt
```diff
@@ -0,0 +1,19 @@
...
+
+package com.datadog.android.core.configuration
+
+/**
+ * Defines the mitigation to use when a queue hits the maximum back pressure capacity.
+ */
+enum class BackPressureMitigation {
+
+    /**  Drop the oldest items already in the queue to make room for new ones. */
+    DROP_OLDEST,
```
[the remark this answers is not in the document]
xgouchet: It'll be the latter as each addition is independent. We don't have a way to "add n items" to the queue, it will always be 5⨉"add one item", which mean each one will be resolved independently.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/BackPressureStrategy.kt
```diff
@@ -0,0 +1,20 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core.configuration
+
+/**
```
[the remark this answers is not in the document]
xgouchet: The default value is described in the `setBackpressureStrategy` of the Configuration builder

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
xgouchet: IMHO it makes more sense to make it a global configuration. A huge backlog in one executor can propagate to the persistence executor (which is what was experienced by our customer)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
xgouchet: It's something that our customer asked for: being able to disable some of their custom tracking if they reach this threshold for instance.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
xgouchet: Yes, that's the point here to say that we reached the threshold, and that adding one more item _could_ lead to a loss of event. The idea is to have a callback before getting to that loss.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
0xnm: ok, such use-case makes sense indeed. I guess they are not going to enable it back when it comes again below the threshold (we don't have anyway callback for that, when it comes)?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/thread/DropOldestBackPressuredBlockingQueueTest.kt
```diff
@@ -0,0 +1,420 @@
...
+
+    // endregion
+
+    // region offer(e)
+
+    @Test
+    fun `M accept item W offer() {empty}`(
+        @StringForgery fakeNewItem: String
+    ) {
+        // Given
```
jonathanmos, line 151: minor: Empty `Given` blocks in this class

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/thread/IgnoreNewestBackPressuredBlockingQueueTest.kt
```diff
@@ -0,0 +1,420 @@
...
+        )
+    }
+
+    // region add(e)
+
+    @Test
+    fun `M accept item W add() {empty}`(
+        @StringForgery fakeNewItem: String
+    ) {
+        // Given
```
jonathanmos, line 69: minor: Empty `Given` blocks in this class

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
```diff
@@ -0,0 +1,91 @@
...
+            }
+            operation(e)
+        }
+    }
+
+    private fun onThresholdReached() {
+        backPressureStrategy.onThresholdReached()
+        logger.log(
+            level = InternalLogger.Level.WARN,
+            targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
0xnm, line 71: hope it won't be very spammy

0xnm (approved):

PR #1939, opened 2024-03-26
RUM-3670 use backpressured executors

xgouchet:
### What does this PR do?

Apply the Backpressure strategy and blocking queue to all `ExecutorService`s.

### Additional Notes

> [!NOTE]
> We are still using our legacy `LoggingScheduledThreadPoolExecutor` for scheduled tasks (e.g.: uploader, vitals tracking, …). Those will only have one our two tasks in their queue, so it's not as urgent to use the backpressure strategy there. A task (`RUM-3704`) has been created to keep track of this.


dd-sdk-android-core/BackgroundThreads.md
```diff
@@ -0,0 +1,103 @@
...
+- (internal) used in the ConsentAwareStorage to process the `writeCurrentBatch()` calls
+- (external) used in the AbstractStorage to process the `writeCurrentBatch()` calls
+- (internal) used in the ConsentAwareFileOrchestrator to migrate files when tracking consent changes
+- (internal) used in the DatadogExceptionHandler to ensure the crash is written to disk before letting the process exit
+
+##### Load and scalability
+
+##### Implementation details
+
+Except for our unit tests the actual implementation is our own `BackPressureExecutorService`
```
0xnm, line 23:
```suggestion
Except for our unit tests the actual implementation is our own `BackPressureExecutorService`.
```

dd-sdk-android-core/BackgroundThreads.md
```diff
@@ -0,0 +1,103 @@
...
+---
+
+### CoreFeature.uploadExecutorService: `ScheduledThreadPoolExecutor`
+
+##### Purpose and Usage
+
+This instance (one per CoreFeature) is used to process the data upload process.
+It is specifically used for:
+
+- (internal) creates the initial Configuration Telemetry on DatadogCore startup (delayed to allow CP SDKs to update the
```
0xnm, line 34:
```suggestion
- (internal) creates the initial Configuration Telemetry on DatadogCore startup (delayed to allow cross-platform SDKs to update the
```

dd-sdk-android-core/BackgroundThreads.md
```diff
@@ -0,0 +1,103 @@
+# Background Threads
+
+This document lists all background threads created by the Datadog SDK (and its features and integrations),
+with their purposes, where they are created, and the stress testing done on each of them.
+
+### CoreFeature.persistenceExecutorService: `FlushableExecutorService`
+
+##### Purpose and Usage
```
0xnm, line 8: do you think it is useful to tell in this doc how many threads the particular service is managing?

dd-sdk-android-core/BackgroundThreads.md
```diff
@@ -0,0 +1,103 @@
...
+
+- (internal) creates the initial Configuration Telemetry on DatadogCore startup (delayed to allow CP SDKs to update the
+  configuration)
+- (internal) used in the DataUploadScheduler to schedule regular upload cycles
+
+##### Load and scalability
+
+##### Implementation details
+
+Except for our unit tests, the actual implementation is our own `LoggingScheduledThreadPoolExecutor`
```
0xnm, line 42:
```suggestion
Except for our unit tests, the actual implementation is our own `LoggingScheduledThreadPoolExecutor`.
```

similar change is needed for other text like this one.

dd-sdk-android-core/BackgroundThreads.md
```diff
@@ -0,0 +1,103 @@
...
+
+---
+
+### RumFeature.anrDetectorExecutorService: `ExecutorService`
+
+##### Purpose and Usage
+
+This instance (one per RumFeature) is used for our own pre API 30 ANR detection mechanism.
```
0xnm, line 81:
```suggestion
This instance (one per RumFeature) is used for our own pre-API 30 ANR detection mechanism.
```

dd-sdk-android-core/BackgroundThreads.md
```diff
@@ -0,0 +1,103 @@
...
+Except for our unit tests, the actual implementation is our own `BackPressureExecutorService`
+
+---
+
+### DatadogRumMonitor.executorService: `ExecutorService`
+
+##### Purpose and Usage
+
+This instance (one per DatadogRumMonitor) is used to defer the process of raw events (usually captured from the
+main thread) and eventually generate actual Json event that need to be written on disk.
```
0xnm, line 96:
```suggestion
main thread) and eventually generate actual Json event that needs to be written on disk.
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
```diff
@@ -71,4 +73,14 @@ interface FeatureSdkCore : SdkCore {
...
+
+    /**
+     * Returns a new [ExecutorService], set up with backpressure and internal monitoring.
+     */
+    fun createExecutorService(): ExecutorService
+
+    /**
+     * Returns a new [ScheduledExecutorService], set up with internal monitoring.
+     */
+    fun createScheduledExecutorService(): ScheduledExecutorService
```
0xnm, line 85: I think it is a bit dangerous to not know how many threads will be managed under the hood by the service created.

We may have some change in the implementation of those methods in the future, for example `createExecutorService` won't return single-threaded service anymore, and we will face bugs which are hard to catch (because, for example, in unit tests we mock executors, so unlikely we will catch them there).

Maybe we can adopt lingvo from `Executors` class, which have methods like `newSingleThreadExecutor` (we may name `createExecutorService` exactly like this btw). And for the `createScheduledExecutorService` write in the javadoc that it will create an executor with max number of threads matching number of cores?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -560,10 +566,12 @@ internal class CoreFeature(
     }
 
     private fun setupExecutors() {
-        @Suppress("UnsafeThirdPartyFunctionCall") // pool size can't be <= 0
-        uploadExecutorService = LoggingScheduledThreadPoolExecutor(CORE_DEFAULT_POOL_SIZE, internalLogger)
-        @Suppress("UnsafeThirdPartyFunctionCall") // workQueue can't be null
-        persistenceExecutorService = persistenceExecutorServiceFactory(internalLogger)
+        uploadExecutorService = LoggingScheduledThreadPoolExecutor(
```
0xnm, line 569: should we use `scheduledExecutorServiceFactory` for this one?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/thread/ScheduledExecutorServiceFactory.kt
```diff
@@ -0,0 +1,28 @@
...
+package com.datadog.android.core.thread
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.configuration.BackPressureStrategy
+import java.util.concurrent.ScheduledExecutorService
+
+/**
+ * A factory for [ScheduledExecutorService].
+ */
+fun interface ScheduledExecutorServiceFactory {
```
0xnm, line 16: it seems this one can be `internal`, if I'm not wrong. It is used only in the `CoreFeature` constructor signature, which is internal.

```suggestion
internal fun interface ScheduledExecutorServiceFactory {
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -560,10 +566,12 @@ internal class CoreFeature(
...
+            CORE_DEFAULT_POOL_SIZE,
+            internalLogger,
+            backpressureStrategy
+        )
+        persistenceExecutorService = executorServiceFactory.create(internalLogger, backpressureStrategy)
```
0xnm, line 574: It seems here it is quite severe change: currently `persistenceExecutorService` is using an executor where max number of threads is `Runtime.getRuntime().availableProcessors()`. But with this change it will use `BackPressureExecutorService` which is providing a single threaded executor. Is this intended change?

0xnm (commented):
I covered so far only production files in a quick look

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
mariusc83: I actually wonder why we used an executor with more than 1 thread there ? The way I know is that the producer should be serial and the consumer (uploader) can use more threads. There is no reason to use more threads to write as the file will be locked anyway.

dd-sdk-android-core/BackgroundThreads.md
xgouchet: For now, I don't think it's relevant, this doc can evolve as our needs grow

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
xgouchet: Unfortunately because we rely on it being a specific implementation to drain it (when we call the flush method), we can't use the generic factory which returns a `ScheduledExecutorService`. And I don't want to polute the factory by coupling tightly to a specific class.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
xgouchet: It is indeed on purpose. The persistence executor would not gain much by being multi threaded since we have locks to avoid writing the same file concurrently. Since most of the issues happen by spamming the same track, we would essentially have the same bottleneck, with the additional context switching cost of juggling more threads.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
mariusc83: In that case why do we use the `scheduledExecutorServiceFactory` ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -393,7 +391,7 @@ internal class RumFeature(
 
     private fun initializeVitalReaders(periodInMs: Long) {
         @Suppress("UnsafeThirdPartyFunctionCall") // pool size can't be <= 0
-        vitalExecutorService = LoggingScheduledThreadPoolExecutor(1, sdkCore.internalLogger)
+        vitalExecutorService = sdkCore.createScheduledExecutorService()
```
0xnm, line 394: maybe the possible improvement for the future: `LoggingScheduledThreadPoolExecutor` is calling `ScheduledThreadPoolExecutor` constructor which allow unlimited growth of the number of threads in the pool (it is limited by `Int.MAX_VALUE`).

Unlikely we may have a big number of threads there, but if by some reason operations scheduled on this executor are very slow, in order to satisfy the execution schedule we may see uncontrolled growth of the number of threads in the pool. Probably we can limit max value.

mariusc83, line 394: Oh I see why do we need this now ;)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
mariusc83: Nevermind, I got my answer below ;)

0xnm (approved):
lgtm. I left some comments/suggestions, but they are not blocking.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -144,7 +137,9 @@ internal class CoreFeature(
     internal var site: DatadogSite = DatadogSite.US1
 
     internal lateinit var uploadExecutorService: ScheduledThreadPoolExecutor
-    internal lateinit var persistenceExecutorService: ExecutorService
+    internal lateinit var persistenceExecutorService: FlushableExecutorService
```
mariusc83, line 140: Now that I see this problem I wonder if we should not use the factories for the features persistence + uploader executors. This way each feature will have its own. We will exchange more threads with less backpressure but I think this will be more performant as each feature has its own persistence folder so everything will be faster.

mariusc83 (approved):
LGTM But I would really like us to think about what I added there as a opened question.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
xgouchet: right now the persistence is using the factory, as for the uploader, so far it is blocked because we are tightly coupled to the implementation class but we can take another go at this in the future.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
xgouchet: For now all our scheduledExecutor instances should only have  one queued task at a time, which mean we shouldn't need more than one thread each.

0xnm: That is true for most of our use-cases, except vitals: there we are scheduling 2 different tasks - CPU and Memory read. But anyway, my comment was rather for the generic use-case.

PR #1940, opened 2024-03-26
RM-1844: Add Method Call Telemetry

jonathanmos:
### What does this PR do?
Adds telemetry for view tree traversal time

### Motivation

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/MethodCalledTelemetry.kt
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.telemetry
+
+import com.datadog.android.Datadog
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.InternalSdkCore
+
+internal class MethodCalledTelemetry(
+    private val operationName: String,
```
jonathanmos, line 14: no need to have this parameter since the value never changes

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/TelemetryWrapper.kt
```diff
@@ -0,0 +1,47 @@
...
+        // The name of the class that calls the method.
+        callerClass: String,
+
+        // The sampling rate of the method call. Value between `0.0` and `100.0`,
+        // where `0.0` means NO event will be processed and `100.0` means ALL events will be processed.
+        // Note that this value is multiplicated by telemetry sampling (by default 20%) and
+        // metric events sampling (hardcoded to 15%). Making it effectively 3% sampling rate
+        // for sending events, when this value is set to `100`.
+        samplingRate: Float = 100.0f
+    ): MethodCalledTelemetry? {
```
jonathanmos, line 30: need to change this to a base class for metrics

maciejburda (approved):
Compared to [iOS implementation](https://github.com/DataDog/dd-sdk-ios/pull/1723) and [documentation (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3528197707/Method+Called+Telemetry+-+Spec).

LGTM

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/TelemetryWrapper.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.telemetry
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.sampling.RateBasedSampler
+import com.datadog.android.core.sampling.Sampler
+import com.datadog.android.sessionreplay.internal.recorder.telemetry.MethodCalledTelemetry.Companion.METHOD_CALL_OPERATION_NAME
+
+internal class TelemetryWrapper(
```
xgouchet, line 14: > [!CAUTION]
> This class and is a bit confusing. 
> This class is _technically_ a wrapper, but the name here doesn't give us any insights on what it does. It feels like its main purpose is to create a base metric entity, and IMO doesn't justify creating a class just for that. Besides, using a string as parameter to know what instance to create can lead to issues. If we do use this at all, we should create at least an enum to prevent bad values and get compiler warnings to handle every case).
>  I don't think it's that necessary. Instead I would feel it more relevant to use a factory method in the  `MetricBase`'s companion object.
> E.g.: 

```kotlin
internal interface MetricBase {

    companion object {
        fun startMethodCallMetric(logger: InternalLogger, callerClass: String, samplingRate: Float = 100.0f) : MetricBase? {
        }
    }
}
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/TelemetryWrapper.kt
```diff
@@ -0,0 +1,47 @@
...
+        // The name of the class that calls the method.
+        callerClass: String,
+
+        // The sampling rate of the method call. Value between `0.0` and `100.0`,
+        // where `0.0` means NO event will be processed and `100.0` means ALL events will be processed.
+        // Note that this value is multiplicated by telemetry sampling (by default 20%) and
+        // metric events sampling (hardcoded to 15%). Making it effectively 3% sampling rate
+        // for sending events, when this value is set to `100`.
+        samplingRate: Float = 100.0f
+    ): MetricBase? {
```
xgouchet, line 30: > [!NOTE]
> If you want to document the arguments of a method, please use KDoc styling instead.
> E.g.: 

```suggestion
    /*
     * @param operationName the platform agnostic name of the operation.
     * @param callerClass the name of the class that calls the method.
     * @param samplingRate the sampling rate of the method call. Value between `0.0` and `100.0`,
     * where `0.0` means NO event will be processed and `100.0` means ALL events will be processed.
     * Note that this value is multiplicated by telemetry sampling (by default 20%) and
     * metric events sampling (hardcoded to 15%). Making it effectively 3% sampling rate
     * for sending events, when this value is set to `100`.
     */
    internal fun startMetric(
        operationName: String,
        callerClass: String,
        samplingRate: Float = 100.0f
    ): MetricBase? {
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/MethodCalledTelemetry.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.datadog.android.core.InternalSdkCore
+import com.datadog.android.sessionreplay.internal.recorder.telemetry.MetricBase.Companion.METRIC_TYPE
+
+internal class MethodCalledTelemetry(
+    private val callerClass: String,
+    private val logger: InternalLogger,
+    private val startTime: Long = System.nanoTime(),
+    private val internalSdkCore: InternalSdkCore? = Datadog.getInstance() as? InternalSdkCore
+) : MetricBase {
+    internal fun stopMethodCalled(isSuccessful: Boolean) {
```
xgouchet, line 20: > [!CAUTION]
> Instead of making this method specific to this class and internal, you should make it generic in the `MetricBase` interfaces (e.g.: `MetricBase.stopMetric(isSuccessful)`). 
> Otherwise you're creating a tight coupling and bad practice as shown above: callers that get a MetricBase instance need to cast if `as? MethodCalledTelemetry` to be able to stop it. This means that any change to what instance is returned when calling `startMetric` will impact a lot of calling sites.

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/MetricBase.kt
```diff
@@ -0,0 +1,44 @@
...
+internal interface MetricBase {
+    fun sendMetric(isSuccessful: Boolean)
+
+    companion object {
+        // Basic Metric type key.
+        internal const val METRIC_TYPE = "metric_type"
+
+        internal enum class TelemetryMetrics {
+            MethodCalled
+        }
```
xgouchet, line 23: > [!WARNING]
> This enum should not be declared in the companion object, it should instead be in its own file. Also it should be called `TelemetryMetricType`.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/MethodCalledTelemetry.kt
```diff
@@ -0,0 +1,114 @@
...
+            osMap[OS_NAME] = it
+        }
+
+        deviceInfo?.osVersion?.let {
+            osMap[OS_VERSION] = it
+        }
+
+        deviceInfo?.deviceBuildId?.let {
+            osMap[OS_BUILD] = it
+        }
```
xgouchet, line 57: > [!NOTE]
> I'm just realising this now, but why do we need to attach those manually here? Aren't they automatically part of every telemetry log we send anyway ?

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/MethodCalledTelemetry.kt
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder.telemetry
+
+import com.datadog.android.Datadog
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.InternalSdkCore
+import com.datadog.android.sessionreplay.internal.recorder.telemetry.MetricBase.Companion.METRIC_TYPE
+
+internal class MethodCalledTelemetry(
```
mariusc83, line 14: I wonder why we do not follow the same architecture we have for the `BatchMetrics`. This way we can keep things aligned. See: https://github.com/DataDog/dd-sdk-android/blob/develop/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/telemetry/MetricBase.kt
```diff
@@ -0,0 +1,40 @@
...
+
+// Base class for metric events
+internal interface MetricBase {
+    fun sendMetric(isSuccessful: Boolean)
+
+    companion object {
+        // Basic Metric type key.
+        internal const val METRIC_TYPE = "metric_type"
+
+        fun startMetric(
```
mariusc83, line 21: this method looks quite unusual here bundled in this interface as a companion object. Instead we could have these methods added in the `MetricsDispatcher` interface and then have a global `DatadogMetricsDispatcher(batchMetricsDispatcher, methodCalledMetricsDispatcher)` which delegates under the hood to the bundled components. What do you guys think ? @xgouchet @jonathanmos @0xnm

xgouchet (approved):

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/TypeDefinition.kt
```diff
@@ -113,10 +113,20 @@ sealed class TypeDefinition {
                 }
             }
 
+            val mergedAdditionalProperties =
```
xgouchet, line 116: > [!IMPORTANT]
> Can you please add unit tests to make sure we have all cases covered here

xgouchet (approved):

xgouchet (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -33,15 +36,20 @@ internal class TelemetryEventHandler(
     internal val eventSampler: Sampler,
     internal val configurationExtraSampler: Sampler =
         RateBasedSampler(DEFAULT_CONFIGURATION_SAMPLE_RATE),
-    private val maxEventCountPerSession: Int = MAX_EVENTS_PER_SESSION
+    private val maxEventCountPerSession: Int = MAX_EVENTS_PER_SESSION,
+    internalSdkCore: InternalSdkCore? = Datadog.getInstance() as? InternalSdkCore,
+    private val deviceInfo: DeviceInfo? = internalSdkCore?.getDatadogContext()?.deviceInfo
```
0xnm, line 41: why this is needed, if inside `handleEvent` method there is already `DatadogContext` provided as a part of `withWriteContext` lambda?

PR #1945, opened 2024-03-27
RUM-2207 Call RUM error mapper even for crashes

fuzzybinary:
### What does this PR do?

This changes the RUM error event mapper to call the user supplied mapper even for crashes, which it previously excluded.

The error event mapper still does not support discarding fatal errors, and a warning is sent if the user returns null for a fatal error in the mapper.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMapper.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMapper.kt
@@ -40,10 +40,21 @@ internal data class RumEventMapper(
             is ViewEvent -> viewEventMapper.map(event)
             is ActionEvent -> actionEventMapper.map(event)
             is ErrorEvent -> {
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMapper.kt
```diff
@@ -40,10 +40,21 @@ internal data class RumEventMapper(
...
+                    if (mappedEvent == null) {
+                        internalLogger.log(
+                            InternalLogger.Level.WARN,
+                            InternalLogger.Target.USER,
+                            { NO_DROPPING_FATAL_ERRORS_WARNING_MESSAGE }
+                        )
+                        event
+                    } else {
+                        mappedEvent
+                    }
```
xgouchet, line 55: > [!IMPORTANT]
> Could you add the tests for this use case ?

xgouchet (changes requested):

0xnm (approved):

xgouchet (approved):

PR #1949, opened 2024-03-28
RUM-3532 keep webview wireframe hidden

xgouchet:
### What does this PR do?

In order to keep webviews consistent in session replays, instead of removing them from the segments when computing a diff, we keep them and mark them as hidden.


sample/kotlin/src/main/res/menu/sr_webview.xml
```diff
@@ -0,0 +1,19 @@
...
+    <item
+            android:title="@string/webview_show"
+            android:id="@+id/webview_show"
+    />
+
+    <item
+            android:title="@string/webview_hide"
+            android:id="@+id/webview_hide"
+    />
+</menu>
```
0xnm, line 19:
```suggestion
</menu>

```

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
```diff
@@ -193,11 +193,12 @@ class SampleApplication : Application() {
...
-                    R.id.nav_host_fragment,
-                    true,
-                    SampleNavigationPredicate()
-                )
+                ActivityViewTrackingStrategy(true)
+//                NavigationViewTrackingStrategy(
+//                    R.id.nav_host_fragment,
+//                    true,
+//                    SampleNavigationPredicate()
+//                )
```
0xnm, line 201: it is for the test or permanent change?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/viewpager/FragmentC.kt
```diff
@@ -6,27 +6,24 @@
...
 
     companion object {
-        const val NAME = "Fragment C"
+        fun newInstance(): FragmentC {
+            return FragmentC().apply {
+                arguments = bundleOf(
+                    ARG_PAGE_NAME to "Fragment C",
+                    ARG_WEB_VIEW_URL to
+                        "https://datadoghq.dev/browser-sdk-test-playground/" +
+                        "?client_token=${BuildConfig.DD_CLIENT_TOKEN}" +
```
0xnm, line 21: probably we don't want the client token to be recorded here, application ID as well. Should we use the URL from `FragmentA`/`FragmentB`?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/webview/WebViewModel.kt
```diff
@@ -16,7 +17,10 @@ internal class WebViewModel(
 
     fun onResume() {
         localServer.start(
-            "https://datadoghq.dev/browser-sdk-test-playground/webview-support/"
+            "https://datadoghq.dev/browser-sdk-test-playground/" +
+                "?client_token=${BuildConfig.DD_CLIENT_TOKEN}" +
+                "&application_id=${BuildConfig.DD_RUM_APPLICATION_ID}" +
+                "&site=datadoghq.com"
```
0xnm, line 23: should we have here a site used by the app/SDK? the one from `BuildConfig.DD_SITE_NAME`

sample/kotlin/src/main/kotlin/com/datadog/android/sample/viewpager/FragmentC.kt
xgouchet: It's the same code as in the webview fragment, and the client token is necessary for this page. 
Also the goal is to have different content to ensure the replay works (if all fragments load the same page it's hard to make sure each webview is handled independently and consistently)

sample/kotlin/src/main/kotlin/com/datadog/android/sample/webview/DatadogSiteExt.kt
```diff
@@ -0,0 +1,35 @@
...
+            Timber.e("Error setting site to ${BuildConfig.DD_SITE_NAME}")
+            null
+        }.browserSite()
+    }
+
+private fun DatadogSite?.browserSite(): String {
+    return when (this) {
+        DatadogSite.US1,
+        DatadogSite.STAGING,
+        null -> "datadoghq.com"
```
0xnm, line 27: is there a reason why staging is pointing to us1?

0xnm (approved):

PR #1951, opened 2024-03-28
RUM-3761 prevent crashing the host app in the ViewOnDrawInterceptor

xgouchet:
### What does this PR do?

Prevent a crash previously unnoticed

### Motivation

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewOnDrawInterceptor.kt
```diff
@@ -44,16 +45,33 @@ internal class ViewOnDrawInterceptor(
     }
 
     fun stopIntercepting() {
-        decorOnDrawListeners.entries.forEach {
-            it.key.viewTreeObserver.removeOnDrawListener(it.value)
+        decorOnDrawListeners.entries.forEach { (decorView, listener) ->
+            if (decorView.viewTreeObserver.isAlive) {
```
0xnm, line 49: seems we have this check already in the `stopInterceptingSafe` anyway, it can be removed here

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewOnDrawInterceptor.kt
```diff
@@ -44,16 +45,33 @@ internal class ViewOnDrawInterceptor(
...
-                window.viewTreeObserver.removeOnDrawListener(it)
+        decorViews.forEach { decorView ->
+            decorOnDrawListeners.remove(decorView)?.let { listener ->
+                stopInterceptingSafe(decorView, listener)
+            }
+        }
+    }
+
+    private fun stopInterceptingSafe(decorView: View, listener: OnDrawListener) {
+        if (decorView.viewTreeObserver.isAlive) {
```
0xnm, line 65: I'm wondering if we can create memory leak or some issue here, because the storage of listeners in `ViewTreeObserver` is of type `ArrayList<OnDrawListener> mOnDrawListeners`.

It means if we added listener, but didn't manage to remove it, next time we add it we will have 2 items backed by the same instance inside `ArrayList<OnDrawListener> mOnDrawListeners`, meaning we will be called 2 times.

But anyway, there is no other way to remove the listener other than calling `removeOnDrawListener`.

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewOnDrawInterceptor.kt
xgouchet: Good catch

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/ViewOnDrawInterceptor.kt
xgouchet: Based on the source code, once a ViewTreeObserver is killed, it can't get alive anymore, meaning that we wouldn't add a listener to the same tree. 
That being said the same crash could happen in addOnDrawListener, I updated my PR

0xnm (approved):

PR #1960, opened 2024-04-02
RUM-3249 Provide the bundleWithRum capability for OtelTracer

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/detekt_custom.yml
+++ b/detekt_custom.yml
@@ -1257,6 +1257,7 @@ datadog:
       # endregion
       # region Opentelemetry
       - "io.opentelemetry.api.trace.TracerBuilder.build()"
+      - "io.opentelemetry.api.trace.SpanBuilder.setAttribute(kotlin.String?, kotlin.String?)"
       - "io.opentelemetry.api.trace.TracerBuilder.setInstrumentationVersion(kotlin.String?)"
       # endregion
       # region RxJava
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
@@ -12,6 +12,7 @@ import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
 import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.log.LogAttributes
 import com.datadog.android.trace.internal.TracingFeature
 import com.datadog.android.trace.internal.data.NoOpOtelWriter
 import com.datadog.opentelemetry.trace.OtelTracerBuilder
@@ -19,6 +20,7 @@ import com.datadog.trace.api.IdGenerationStrategy
 import com.datadog.trace.api.config.TracerConfig
 import com.datadog.trace.bootstrap.instrumentation.api.AgentTracer
 import com.datadog.trace.core.CoreTracer
+import io.opentelemetry.api.trace.SpanBuilder
 import io.opentelemetry.api.trace.Tracer
 import io.opentelemetry.api.trace.TracerBuilder
 import io.opentelemetry.api.trace.TracerProvider
@@ -34,8 +36,10 @@ import java.util.Properties
  *
  */
 class OtelTracerProvider(
+    private val sdkCore: FeatureSdkCore,
     private val coreTracer: AgentTracer.TracerAPI,
-    private val internalLogger: InternalLogger
+    private val internalLogger: InternalLogger,
+    private val bundleWithRumEnabled: Boolean
 ) : TracerProvider {
 
     private val tracers: MutableMap<String, Tracer> = mutableMapOf()
@@ -76,15 +80,12 @@ class OtelTracerProvider(
     /** @inheritDoc */
     override fun tracerBuilder(instrumentationScopeName: String): TracerBuilder {
         val resolvedInstrumentationScopeName = resolveInstrumentationScopeName(instrumentationScopeName)
-        return OtelTracerBuilder(resolvedInstrumentationScopeName, coreTracer, internalLogger)
-    }
-
-    private fun resolveInstrumentationScopeName(instrumentationScopeName: String): String {
-        return if (instrumentationScopeName.trim { it <= ' ' }.isEmpty()) {
-            DEFAULT_TRACER_NAME
-        } else {
-            instrumentationScopeName
-        }
+        return OtelTracerBuilder(
+            resolvedInstrumentationScopeName,
+            coreTracer,
+            internalLogger,
+            resolveSpanBuilderDecorator()
+        )
     }
 
     /**
@@ -112,6 +113,7 @@ class OtelTracerProvider(
             }
         private var partialFlushThreshold = DEFAULT_PARTIAL_MIN_FLUSH
         private val globalTags: MutableMap<String, String> = mutableMapOf()
+        private var bundleWithRumEnabled: Boolean = true
 
         /**
          * @param sdkCore SDK instance to bind to. If not provided, default instance will be used.
@@ -141,7 +143,7 @@ class OtelTracerProvider(
                 .partialFlushMinSpans(partialFlushThreshold)
                 .idGenerationStrategy(IdGenerationStrategy.fromName("SECURE_RANDOM", false))
                 .build()
-            return OtelTracerProvider(coreTracer, sdkCore.internalLogger)
+            return OtelTracerProvider(sdkCore, coreTracer, sdkCore.internalLogger, bundleWithRumEnabled)
         }
 
         /**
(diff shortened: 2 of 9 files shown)


detekt_custom.yml
```diff
@@ -1258,6 +1258,7 @@ datadog:
       # region Opentelemetry
       - "io.opentelemetry.api.trace.TracerBuilder.build()"
       - "io.opentelemetry.api.trace.TracerBuilder.setInstrumentationVersion(kotlin.String?)"
+      - "io.opentelemetry.api.trace.SpanBuilder.setAttribute(kotlin.String?, kotlin.String?)"
```
0xnm, line 1261: minor: just ordering

```suggestion
      # region OpenTelemetry
      - "io.opentelemetry.api.trace.SpanBuilder.setAttribute(kotlin.String?, kotlin.String?)"
      - "io.opentelemetry.api.trace.TracerBuilder.build()"
      - "io.opentelemetry.api.trace.TracerBuilder.setInstrumentationVersion(kotlin.String?)"
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
```diff
@@ -134,6 +136,15 @@ class OtelTracerProvider(
...
             }
+            val rumFeature = sdkCore.getFeature(Feature.RUM_FEATURE_NAME)
+            if (bundleWithRumEnabled && rumFeature == null) {
+                sdkCore.internalLogger.log(
+                    InternalLogger.Level.WARN,
+                    InternalLogger.Target.USER,
+                    { RUM_NOT_ENABLED_ERROR_MESSAGE }
+                )
+                bundleWithRumEnabled = false
+            }
```
0xnm, line 147: IMO we need to revisit this logic (I know it is already exists for `AndroidTracer`): in theory with v2 we can have things enabled independently, so nothing in the setup prevents enabling RUM after Tracer is enabled. IMO this shouldn't prevent us to do bundling with RUM. Is such logic exist on iOS as well?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
```diff
@@ -195,6 +206,17 @@ class OtelTracerProvider(
             return this
         }
 
+        /**
+         * Enables the trace bundling with the current active View. If this feature is enabled all
+         * the spans from this moment on will be bundled with the current view information and you
+         * will be able to see all the traces sent during a specific view in the Rum Explorer.
```
0xnm, line 212:
```suggestion
         * will be able to see all the traces sent during a specific view in the RUM Explorer.
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
```diff
@@ -229,11 +251,59 @@ class OtelTracerProvider(
...
+            spanBuilder.setAttribute(LogAttributes.RUM_APPLICATION_ID, applicationId)
+            spanBuilder.setAttribute(LogAttributes.RUM_SESSION_ID, sessionId)
+            spanBuilder.setAttribute(LogAttributes.RUM_VIEW_ID, viewId)
+            if (actionId != null) {
+                spanBuilder.setAttribute(LogAttributes.RUM_ACTION_ID, actionId)
+            }
+        } else {
+            internalLogger.log(
+                InternalLogger.Level.WARN,
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
0xnm, line 288: is it useful in the telemetry?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/OtelTracerBuilderProviderTest.kt
```diff
@@ -577,4 +607,225 @@ internal class OtelTracerBuilderProviderTest {
     }
 
     // endregion
+
+    // region bundle with RUM
+
+    @Test
+    fun `M build a Span with RUM context W buildSpan`() {
```
0xnm, line 614: same applies to the tests below

```suggestion
    fun `M build a Span with RUM context W startSpan`() {
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
mariusc83: Hmmm...good point. I just wanted to follow what was already in the `AndroidTracer` but it makes sense maybe to remove it. Waiting for @xgouchet opinion here also.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
mariusc83: Good question. I was thinking that maybe it is useful for us to see how many times we are missing the link with RUM because RUM context is corrupted or session is not there. But I am opened to discussions here.

0xnm: My concern is just to avoid spamming telemetry / avoid having many telemetry events to process in the SDK itself, given that it is tracing, so we may have a lot of spans.

mariusc83: I'll remove it in that case and we will see if we need it in case someone reports missing links.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
xgouchet: I agree we need to evaluat the bundling at span creation as the RUM feature can be enabled/disabled independently

mariusc83: yes agree,  we are already checking and warning the user at the spanbuilder decoration level so there is no need for other log for now. I will just remove this from here and add a ticket to remove it also for the `AndroidTracer`.

xgouchet (approved):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
```diff
@@ -245,6 +306,14 @@ class OtelTracerProvider(
         internal const val DEFAULT_SERVICE_NAME_IS_MISSING_ERROR_MESSAGE =
             "Default service name is missing during" +
                 " OtelTracerProvider creation, did you initialize SDK?"
+        internal const val RUM_NOT_ENABLED_ERROR_MESSAGE =
+            "You're trying to bundle the traces with a RUM context, " +
+                "but the RUM feature was disabled in your Configuration. " +
```
0xnm, line 311: we need to update this message (and similar message in `AndroidTracer`) - RUM is not enabled in core `Configuration` class anymore in v2.

0xnm (approved):
Is change to remove RUM feature check in `AndroidTracer` will be in another PR?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
```diff
@@ -228,11 +241,59 @@ class OtelTracerProvider(
...
+
+    private fun resolveSpanBuilderDecorator(): (SpanBuilder) -> SpanBuilder {
+        return if (bundleWithRumEnabled) {
+            resolveSpanBuilderDecoratorFromContext()
+        } else {
+            NO_OP_SPAN_BUILDER_DECORATOR
+        }
+    }
+
+    private fun resolveSpanBuilderDecoratorFromContext(): (SpanBuilder) -> SpanBuilder = { spanBuilder ->
```
0xnm, line 262: minor: I think the name can be improved here (can be done in another PR) - nothing in the name points that it will be RUM content applied. Maybe something like `rumAwareSpanBuilderDecorator` for this method will be better (also this can be just a property, not a method).

0xnm (approved):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/OtelTracerProvider.kt
mariusc83: yes but that is the idea, this time is for RUM but it should be feature agnostic that's the reason why it is not having any RUM mentioned.

0xnm: then still the name doesn't resolve anything `from` context, it decorates `with` context.

PR #1961, opened 2024-04-02
RUM-2910: Report time since the application start for crashes in RUM

0xnm:
### What does this PR do?

This PR adds the following:

* Report time since application start for JVM crashes to RUM
* Report time since application start for NDK crashes to RUM

This is achieved by introducing `error.time_since_app_start` attribute (in milliseconds). Application start is defined as the start of the application process (the same definition as for the `ApplicationLaunch` view).

Also this PR moves application start time provider to the core module, because:

* It is needed for NDK crashes (there is no way to get reliably application start time in NDK; we could use static initialization during the library load, but the library may be loaded later at the application lifecycle)
* In case of Android API below N we can rely on the `DatadogCore` class loading, which happens before `RumFeature` class loading and thus should give more precise timestamp.

This PR doesn't add support of reporting the similar time for the fatal ANRs, because additional development is needed there to keep somehow application start time between application launches.


xgouchet (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -436,7 +436,8 @@ internal open class RumViewScope(
                             stack = it.stack,
                             state = it.state
                         )
-                    }.ifEmpty { null }
+                    }.ifEmpty { null },
+                    timeSinceAppStart = event.timeSinceAppStartNs?.let { TimeUnit.NANOSECONDS.toMillis(it) }
```
mariusc83, line 440: maybe you should extract this in a function ?

mariusc83 (approved):

PR #1962, opened 2024-04-03
Fix flaky `TodoWithoutTask` tests

0xnm:
### What does this PR do?

Regex `[A-Z]+-\\d{1,8}` may generate something like `ABC-0` because of the `d{1,8}`, and this will fail some tests, because 0 as a task number is not allowed.

This PR fixes that.

Note: In theory it is possible to have  something like `ABC-012`, but generating `012` and avoiding to have all zeros at the same time requires using matching groups/negative lookaheads, which is not supported by `Forge`.


xgouchet (approved):
Nice catch, thanks

PR #1963, opened 2024-04-03
Fix message when writer is NoOp

xgouchet:
### What does this PR do?

This avoids misleading logs when a RUM session is expired and the writer becomes no op (it returns false on write).

### Motivation

Avoid noise and misleading noise in logcat for customer, and provide a more relevant explanation


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/utils/SdkCoreExt.kt
```diff
@@ -44,20 +45,28 @@ internal class WriteOperation(
...
         sdkCore.getFeature(Feature.RUM_FEATURE_NAME)
             ?.withWriteContext { datadogContext, eventBatchWriter ->
-                try {
-                    val event = eventSource(datadogContext)
+                if (rumDataWriter is NoOpDataWriter) {
+                    sdkCore.internalLogger.log(
+                        level = InternalLogger.Level.INFO,
+                        target = InternalLogger.Target.USER,
+                        messageBuilder = { WRITE_OPERATION_IGNORED }
+                    )
```
0xnm, line 53: suggestion: we probably should also call `onSuccess/onError` here (don't know which one makes more sense), because at least for the `RumViewScope` until `sum(pending) == 0` it won't be marked as complete and thus garbage collected. We may have the case when there is `NoOpDataWriter` here and RUM pipeline working if session is not tracked, I believe.

0xnm (commented):
nice catch! I added a suggestion.

0xnm (approved):

PR #1967, opened 2024-04-04
Fix flaky tests

xgouchet:
### What does this PR do?

This commit attemps to fix several tests that are regularly flaky in CI


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -756,7 +756,7 @@ internal class DatadogCoreTest {
 
     @Test
     fun `𝕄 stop all features 𝕎 stop()`(
-        @StringForgery fakeFeatureNames: List<String>
+        @StringForgery fakeFeatureNames: Set<String>
```
0xnm, line 759: already here https://github.com/DataDog/dd-sdk-android/pull/1965 :)

0xnm (approved):

PR #1968, opened 2024-04-05
RUM-3796: Optimise `BatchFileOrchestator` performance

0xnm:
### What does this PR do?

This PR optimise `BatchFileOrchestator` performance by doing the following changes:

* Remove `File.isFile` check - this is expensive syscall and it should be only files in the particular folder anyway (it is only `BatchFileOrchestrator` who works with that folder) and we check file names anyway. Also `File.isFile` doc says: `Any non-directory file created by a Java application is guaranteed to be a normal file.`
* Remove regex for the batch file name check in favor of conversion to `Long`.
* Don't sort files where it is not needed
* Reuse files collection where it is possible to minimise the number of syscalls.
* Remove file cache (it is not really needed after `File.isFile` check is removed as benchmarks show below)
* Increase cleanup check frequency from 1s to 5s.

#### Benchmarks

The testing approach was the following: generate 500 files of size ranging from 1Kb to 32Kb, evenly distributed in the last 24h by their names as timestamps. Then do 100 calls to `getWritableFile` passing `forceNewFile = true` at every even call, and `forceNewFile = false` at every odd call.

Tests:

1. no cache, no cleanup frequency threshold (meaning cleanup can be executed at every call to `getWritableFile`)
2. no cache, cleanup frequency threshold = 1s
3. cache for 400 files, cleanup frequency threshold = 1s
4.  cache for 800 files, cleanup frequency threshold = 1s

Table below shows the average duration of each `getWritableFile` call in milliseconds.

|   	|   1	|   2	|   3	|   4	|
|---	|---	|---	|---	|---	|
|   Baseline	|   136	|   40	|   23	|   23	|
|   `File.isFile` removed	|   95	|   19	|   22	|   22	|
|   Remove regex	|   88	|   20	|   21	|   21	|
|   Remove sorting if not needed	|   86	|   21	|   22	|   22	|
|   Reuse files collection	|   65	|   20	|   21	|   21	|

NB: it shouldn't be considered as a proper benchmarking methodology (JMH is not compatible with AGP, there is some variability, etc.), but can give an idea of the impact for the improvements.

Last 2 changes (cache removal and change of the cleanup frequency) are not listed, because they directly impact tests.

All changes are in separate commits.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -313,40 +305,39 @@ internal class BatchFileOrchestrator(
...
-        return rootDir.listFilesSafe(fileFilter, internalLogger).orEmpty().sorted()
+    private fun listBatchFiles(): List<File> {
+        return rootDir.listFilesSafe(fileFilter, internalLogger).orEmpty().toList()
     }
 
-    private val File.metadata: File
-        get() = File("${this.path}_metadata")
+    private fun listSortedBatchFiles(): List<File> {
+        // note: since it is using File#compareTo, lexicographical sorting will be used, meaning "10" comes before "9".
+        // but for our needs it is fine, because the moment when Unix timestamp adds one more digit will be in 2286.
```
xgouchet, line 314: Few, I guess it'll be for others to deal with this then…

xgouchet (approved):
Really nice job there !

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
0xnm: Unless SDK is installed on time machine device

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -313,40 +305,39 @@ internal class BatchFileOrchestrator(
...
     // region FileFilter
 
-    internal inner class BatchFileFilter(private val internalLogger: InternalLogger) : FileFilter {
+    internal inner class BatchFileFilter : FileFilter {
         @Suppress("ReturnCount")
         override fun accept(file: File?): Boolean {
             if (file == null) return false
 
-            @Suppress("UnsafeThirdPartyFunctionCall") // value is not null
-            if (knownBatchFiles.get(file) != null) {
```
mariusc83, line 337: isn't this dangerous ? I know the uploaded checks this set when trying to pick a file for uploading ?

0xnm: It is a cache, not a list of files locked for upload. Cache is not needed now as the performance without the cache is the same as with it (check the difference between tests 2, 3 and 4 once first optimization was applied).

mariusc83 (approved):

PR #1969, opened 2024-04-05
RUM-3868: Don't traverse non-visible ViewGroups for searching user interaction targets

0xnm:
### What does this PR do?

This a fix for the following bug: let's imagine we have a container view and a child view in this container, both are visible. If we set `View.GONE` or `View.INVISIBLE` for the container view, both views won't be visible, but if we click on the place where child view would be rendered, the tap will be reported.

This happens because when we hide container view, child view still have a visibility `View.VISIBLE` despite not participating in rendering (if parent group has `View.INVISIBLE`) or even in the layout process (if parent group has `View.GONE`).

This means we need to take into account the visibility of the view groups as well when we are searching for the user interaction target.

Note: child cannot be shown without parent shown, meaning even if we have `View.INVISIBLE` for parent container switching visibility for child view has no effect (at least for built-in view groups as I tested).


xgouchet (approved):
Nice catch ! 
And I think you won the longest branch name contest too :D

PR #1986, opened 2024-04-15
RUM-3861 Update testing ci steps to limit OOM and memory usage

xgouchet:
### What does this PR do?

Updates our CI to 

- spread the load of the runner during our most intensive jobs
- add an OOM dump artifact if this reappears in the future


.codecov.yml
0xnm, line 1: why do we remove `codecov`? I find it useful, I rely on the data I receive for each PR to see the coverage change and if any files touched in the PR are not covered enough.

.github/workflows/codeql-analysis.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
     # Manually build the java bytecode
     - name: Execute Gradle build
-      run: ./gradlew assembleLibraries
+      run: ./gradlew assembleLibrariesDebug
```
0xnm, line 56: ideally we still need to build both build types (because it is a security scan, it needs to cover everything) or at least `release` build type instead of `debug`, because `release` is what will be used by the customer later.

CONTRIBUTING.md
```diff
@@ -61,7 +61,7 @@ This project hosts the following modules:
 You can build the SDK using the following Gradle command:
 
 ```shell script
-./gradlew assembleLibraries
+./gradlew assembleLibrariesDebug assembleLibrariesRelease
```
0xnm, line 64: let's maybe still keep `assembleLibraries`? it is for the local run, without the agent, so we can run everything at once.

xgouchet: 👍

.codecov.yml
xgouchet: What I'll do is replace the test:release with test:kover then

.gitlab-ci.yml
```diff
@@ -104,18 +104,23 @@ test:debug:
...
       junit: "**/build/test-results/testDebugUnitTest/*.xml"
+    paths:
+      - "*.hprof"
 
 test:release:
   tags: [ "arch:amd64" ]
   image: $CI_IMAGE_DOCKER
   stage: test
   timeout: 1h
+  when: manual
```
0xnm, line 123: should this change be reverted? in which case we are supposed to run this task manually?

.codecov.yml
0xnm, line 1: since we are keeping `codecov.sh`, we should keep this file as well

0xnm (approved):

PR #1991, opened 2024-04-17
RUM-3597: Start sending batches immediately after feature is initialized

0xnm:
### What does this PR do?

This change removes initial delay for start sending batches, now they the first upload will be triggered right after the feature is initialized instead of waiting for some delay.

This allows to submit more data in case if application is crashing early.

The downside is that there is a chance that first batch uploaded won't be sent to the batch upload metrics telemetry, because RUM monitor may be not ready yet, but this is acceptable.


xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadScheduler.kt
```diff
@@ -39,10 +40,8 @@ internal class DataUploadScheduler(
     )
 
     override fun startScheduling() {
-        scheduledThreadPoolExecutor.scheduleSafe(
-            "Data upload",
-            runnable.currentDelayIntervalMs,
-            TimeUnit.MILLISECONDS,
+        scheduledThreadPoolExecutor.executeSafe(
```
mariusc83, line 43: Nice catch

mariusc83 (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -96,14 +96,16 @@ internal class SdkFeature(
...
+            setupUploader(wrappedFeature, dataUploadConfiguration)
         }
 
         if (wrappedFeature is TrackingConsentProviderCallback) {
             coreFeature.trackingConsentProvider.registerCallback(wrappedFeature)
         }
 
         initialized.set(true)
+
+        uploadScheduler.startScheduling()
```
mariusc83, line 108: this function name is misleading here now, maybe we should have a different interface here, this is not scheduling anything here now.

0xnm: I would keep it for consistency with `stopScheduling` and class name - `UploadScheduler`

PR #2006, opened 2024-04-23
RUM-4055 Report Opentelemetry data in the configuration Telemetry

mariusc83:
### What does this PR do?

Following the iOS [Pr](https://github.com/DataDog/dd-sdk-ios/pull/1704) we are going to report the usage of the `OpenTelemetry` API in our configuration telemetry as follows:
`hasTrace:true | false`, `tracerApi: OpenTracing | OpenTelemetry | null`, `traceApiVersion: [openTelemetryApiVersion] | null`

### Motivation

### Additional Notes


xgouchet (changes requested):
> [!IMPORTANT]
> It feels like we need to review the way we send the configuration telemetry information for other features than RUM. 
> Right now we detect the OT by reflection, and this PR tries to add the OpenTelemetry version in the RUM build config. This doesn't make any sense and adds noise to a feature that should not have those kind of information. 
> Maybe we should consider a different way for features to communicate this information and RUM would be agnostic of the keys and values.

mariusc83:
> Important
> 
> It feels like we need to review the way we send the configuration telemetry information for other features than RUM. Right now we detect the OT by reflection, and this PR tries to add the OpenTelemetry version in the RUM build config. This doesn't make any sense and adds noise to a feature that should not have those kind of information. Maybe we should consider a different way for features to communicate this information and RUM would be agnostic of the keys and values.

@xgouchet I understand your concern, we could send it in the `Trace` feature context as we do for the `is_opentelemetry_enabled` property. But we still need to keep the build config entry as the other option is to send it as a hardcoded value that we will definitely forget to update. What do you think ?

P.S. I think opening a RFC only for this topic only would be too much IMO. @plousada what do you think ?

xgouchet:
> @xgouchet I understand your concern, we could send it in the `Trace` feature context as we do for the `is_opentelemetry_enabled` property. But we still need to keep the build config entry as the other option is to send it as a hardcoded value that we will definitely forget to update. What do you think ?
> 
> P.S. I think opening a RFC only for this topic only would be too much IMO. @plousada what do you think ?

I don't have an issue having the info set by build config, it just would make more sense to do it on the otel/tracing module, instead of the RUM one

mariusc83:
> > @xgouchet I understand your concern, we could send it in the `Trace` feature context as we do for the `is_opentelemetry_enabled` property. But we still need to keep the build config entry as the other option is to send it as a hardcoded value that we will definitely forget to update. What do you think ?
> > P.S. I think opening a RFC only for this topic only would be too much IMO. @plousada what do you think ?
> 
> I don't have an issue having the info set by build config, it just would make more sense to do it on the otel/tracing module, instead of the RUM one



> > @xgouchet I understand your concern, we could send it in the `Trace` feature context as we do for the `is_opentelemetry_enabled` property. But we still need to keep the build config entry as the other option is to send it as a hardcoded value that we will definitely forget to update. What do you think ?
> > P.S. I think opening a RFC only for this topic only would be too much IMO. @plousada what do you think ?
> 
> I don't have an issue having the info set by build config, it just would make more sense to do it on the otel/tracing module, instead of the RUM one

Yes this is what I am changing now, makes more sense to have it there. I am pushing a commit in few seconds.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandlerTest.kt
```diff
@@ -353,19 +354,28 @@ internal class TelemetryEventHandlerTest {
...
-    fun `𝕄 create config event 𝕎 handleEvent(SendTelemetry) { configuration with tracing settings }`(
-        @Forgery fakeConfiguration: TelemetryCoreConfiguration,
-        @BoolForgery useTracing: Boolean,
-        forge: Forge
+    @ParameterizedTest
+    @MethodSource("tracingConfigurationParameters")
+    fun `𝕄 create config event 𝕎 handleEvent(SendTelemetry) { tracing configuration with tracing settings }`(
+        configuration: Triple<Boolean, TelemetryEventHandler.TracerApi?, String?>,
+        @Forgery fakeConfiguration: TelemetryCoreConfiguration
     ) {
```
0xnm, line 362: instead of using `Triple` without clear names it is possible to be more explicit:

```suggestion
    @ParameterizedTest
    @MethodSource("tracingConfigurationParameters")
    fun `𝕄 create config event 𝕎 handleEvent(SendTelemetry) { tracing configuration with tracing settings }`(
        useTracer: Boolean,
        tracerApi: TelemetryEventHandler.TracerApi?,
        tracerApiVersion: String?,
        @Forgery fakeConfiguration: TelemetryCoreConfiguration
    ) {
```

and then in the method provider instead of `Triple` you can use `Arguments.of`

features/dd-sdk-android-trace/build.gradle.kts
```diff
@@ -39,6 +39,14 @@ plugins {
 
 android {
     defaultConfig {
+        buildFeatures {
+            buildConfig = true
+        }
+        buildConfigField(
+            "String",
+            "OPENTELEMETRY_API_VERSION_NAME",
+            "\"${libs.versions.openTelemetry.get()}\""
```
0xnm, line 48: maybe there is something from OpenTelemetry in the `META-INF` folder of the resulting package? this would help to avoid `BuildConfig` usage.

Also the value here may be wrong, because it is written for the version which was used to compile our library, but once final app is compiled, there is a chance that there is a newer OpenTelemetry version in the classpath, so it means that our telemetry will report the wrong version.

Additionally, in theory since this property is written at the our library compile time, we can know the version used just by looking on the SDK version. This means that we can set up logs processing pipeline, which will transform SDK version to the OpenTelemetry version used (but that would require to maintain this rule).

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/OtelTracerBuilderProviderTest.kt
```diff
@@ -852,14 +852,51 @@ internal class OtelTracerBuilderProviderTest {
...
             assertThat(traceContext).doesNotContainKey(expectedActiveTraceContextName)
         }
     }
 
     // endregion
 
+    // region Tracing context
+
+    @Test
+    fun `M set opentelemetry as enabled in Context W build { TracingFeature enabled }`() {
```
0xnm, line 866:
```suggestion
    fun `M set OpenTelemetry as enabled in Context W build { TracingFeature enabled }`() {
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/OtelTracerBuilderProviderTest.kt
```diff
@@ -852,14 +852,51 @@ internal class OtelTracerBuilderProviderTest {
...
+            verify(mockSdkCore, times(1)).updateFeatureContext(eq(Feature.TRACING_FEATURE_NAME), capture())
+            lastValue.invoke(traceContext)
+            assertThat(traceContext[TracingFeature.IS_OPENTELEMETRY_ENABLED_CONFIG_KEY]).isEqualTo(true)
+            assertThat(traceContext[TracingFeature.OPENTELEMETRY_API_VERSION_CONFIG_KEY])
+                .isEqualTo(BuildConfig.OPENTELEMETRY_API_VERSION_NAME)
+        }
+    }
+
+    @Test
+    fun `M set opentelemetry as enabled in Context W build { TracingFeature not enabled }`() {
```
0xnm, line 882:
```suggestion
    fun `M set OpenTelemetry as enabled in Context W build { TracingFeature not enabled }`() {
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/OtelTracerBuilderProviderTest.kt
```diff
@@ -852,14 +852,51 @@ internal class OtelTracerBuilderProviderTest {
...
+
+        // THEN
+        argumentCaptor<(MutableMap<String, Any?>) -> Unit> {
+            val traceContext: MutableMap<String, Any?> = mutableMapOf()
+            verify(mockSdkCore, times(1)).updateFeatureContext(eq(Feature.TRACING_FEATURE_NAME), capture())
+            lastValue.invoke(traceContext)
+            assertThat(traceContext[TracingFeature.IS_OPENTELEMETRY_ENABLED_CONFIG_KEY]).isEqualTo(true)
+            assertThat(traceContext[TracingFeature.OPENTELEMETRY_API_VERSION_CONFIG_KEY])
+                .isEqualTo(BuildConfig.OPENTELEMETRY_API_VERSION_NAME)
+        }
```
0xnm, line 895: the behavior shouldn't be like that, no context should be updated if feature is disabled, check the implementation https://github.com/DataDog/dd-sdk-android/blob/cff8b54d458f7cbf6f935deb7ff9b18aec23d9e3/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/DatadogCore.kt#L193

mariusc83: Yes I am aware of that but in that case all the other places where we actually call that method should be under `if(featureExists)` flag to be consistent. The only reason I am calling this method even if the feature is not registered is to be consistent with:
https://github.com/DataDog/dd-sdk-android/blob/cff8b54d458f7cbf6f935deb7ff9b18aec23d9e3/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/AndroidTracer.kt#L57
and 
https://github.com/DataDog/dd-sdk-android/blob/cff8b54d458f7cbf6f935deb7ff9b18aec23d9e3/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/AndroidTracer.kt#L68
and same 
thing for `OtelTracerProvider`.
The real question here is : in case the Trace feature is not registered should we return a workable AndroidTracer or OtelTracerProvider in the `build` method or just a NoOp ? Let's discuss this in the daily.

features/dd-sdk-android-trace/build.gradle.kts
0xnm: Not necessarily. Customer may just have some 3rd party dependency which references newer OpenTelemetry API artifact (or maybe they have their own OpenTelemetry tooling as well). In this case overall dependency graph will be resolved to the newer version, it will be used in the application, but our telemetry will still report the older version (because it will be the one which was used to compiler our SDK, not the one which will be used to compile the application).

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/OtelTracerBuilderProviderTest.kt
0xnm: I don't mean the need to check if feature registered or not before calling `updateFeatureContext`, I mean that this test name is wrong. The name is `M set opentelemetry as enabled in Context W build { TracingFeature not enabled }`, but no, if tracing feature is not registered, then nothing should be set, because the real implementation of `updateFeatureContext` won't do anything. We may assert only up to the point checking that `updateFeatureContext` was called, but not that callback to update values was executed (and test name should be updated).

But probably it makes no sense to even call `updateFeatureContext` if `tracingFeature = null`. By adding this check (and it makes sense, because `tracingFeature` is already in the method anyway) this test can be updated to the proper behavior.

mariusc83: IMO it makes no sense to call it at all but in the same time I still argue that we should probably talk about returning NoOps in case trace feature is not initialized.

0xnm (approved):

xgouchet (approved):

PR #2009, opened 2024-04-24
RUM-4821 unregister vital listeners when view is stopped

### LINKED ISSUE #2008: Memory Leak : AggregatingVitalMonitor listeners are never unregistered
### Describe the bug

in RumViewScope.init, https://github.com/DataDog/dd-sdk-android/blob/cff8b54d458f7cbf6f935deb7ff9b18aec23d9e3/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L145
```
cpuVitalMonitor.register(cpuVitalListener)
memoryVitalMonitor.register(memoryVitalListener)
frameRateVitalMonitor.register(frameRateVitalListener)
```
is called, but the corresponding AggregatingVitalMonitor.unregister calls never happen.
The session-scoped AggregatingVitalMonitors hold a map of listeners, and the https://github.com/DataDog/dd-sdk-android/blob/cff8b54d458f7cbf6f935deb7ff9b18aec23d9e3/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/AggregatingVitalMonitor.kt#L16 map continues to grow, since listeners are never unregistered when the view is stopped, causing a memory leak. 

### Reproduction steps

- break on VitalMonitor.register and VitalMonitor.unregister while starting / stopping views. observe the map of AggregatingVitalMonitor.listeners continues to grow. 

### Logcat logs

_No response_

### Expected behavior

Vital monitor listeners are unregistered from the listeners map when a view is stopped.

### Affected SDK versions

2.8.0

NA

Yes

### Kotlin / Java version

_No response_

### Gradle / AGP version

_No response_

_No response_

### Device Information

_No response_

_No response_

xgouchet:
### What does this PR do?

This fixes a bug where we forgot to unregister the vital listeners. This lead to two issues: 

- we created a memory leak, with listeners references being kept in the monitor and never cleaned up
- when a view would wait for event to finish and send its last update, vital value could be skewed because of what happened afterwards. 

Fixes #2008

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -246,6 +246,10 @@ internal open class RumViewScope(
             sendViewUpdate(event, writer)
             delegateEventToChildren(event, writer)
             sendViewChanged()
+
+            cpuVitalMonitor.unregister(cpuVitalListener)
+            memoryVitalMonitor.unregister(memoryVitalListener)
+            frameRateVitalMonitor.unregister(frameRateVitalListener)
         }
     }
 
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -540,7 +540,6 @@ internal open class RumViewScope(
     @WorkerThread
     private fun onStopSession(event: RumRawEvent.StopSession, writer: DataWriter<Any>) {
         stopScope(event, writer)
-
     }
 
     @WorkerThread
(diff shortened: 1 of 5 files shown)


mariusc83 (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -291,6 +293,10 @@ internal open class RumViewScope(
             stopped = true
             sendViewUpdate(event, writer)
             sendViewChanged()
+
+            cpuVitalMonitor.unregister(cpuVitalListener)
+            memoryVitalMonitor.unregister(memoryVitalListener)
+            frameRateVitalMonitor.unregister(frameRateVitalListener)
```
0xnm, line 299: We should do the same also in `onStartView`, there is view stop logic there as well. And probably in `onStopSession`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -244,6 +246,10 @@ internal open class RumViewScope(
             sendViewUpdate(event, writer)
             delegateEventToChildren(event, writer)
             sendViewChanged()
+
+            cpuVitalMonitor.unregister(cpuVitalListener)
+            memoryVitalMonitor.unregister(memoryVitalListener)
+            frameRateVitalMonitor.unregister(frameRateVitalListener)
```
0xnm, line 252: minor: maybe it is worth to extract these calls into a dedicated method

0xnm (approved):

PR #2010, opened 2024-04-25
Make constructors of `DatadogSite` private

0xnm:
### What does this PR do?

Constructors of `DatadogSite` shouldn't be public.

This may be see as a breaking change (removing items from public API), but this was never intended to be public.


fuzzybinary (approved):

xgouchet (approved):
That works, but it isn't an issue, as enum have a private `init`. 
If someone tried to write the following, it wouldn't compile anyway : 

```kotlin
val customSite = DatadogSite("custom", "custom-intake.example.com")
```

This line triggers two compilation errors: 

> - Cannot access '\<init>': it is private in 'DatadogSite'
> - Enum types cannot be instantiated

0xnm:
@xgouchet yes, my concern was mostly the look of the API surface file, these custom constructors are declared there.

PR #2011, opened 2024-04-29
RUM-4320 improve masking arch

xgouchet:
### What does this PR do?

This PR improves the overall way we deal with privacy settings in the SR modules. 

### Motivation

The way we used to do thing made things difficult as it implied: 

- creating different classes of the same mapper to deal with privacy 
- complicated implementation in SR extensions (you had to have a `Map<Privacy, Map<Class<View>, Mapper<*>>>` which was difficult to follow and implement)
- it made the code and tests complex to aprehend, and required lots of boilerplate to support en

### Additional Notes

This PR uses the existing `MappingContext` that is passed to mappers during the snapshot creation to pass in the current privacy setting. This means each mapper deals with the privacy at snapshot time, and have all the logic in one place, instead of split in two or three classes. 
Also it means that the privacy could be changed dynamically without effort (e.g. with remote config).


0xnm (approved):
nice refactoring!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/TextViewMapper.kt
```diff
@@ -69,10 +58,17 @@ open class TextViewMapper internal constructor(
             density
         )
 
+        val testValueObfuscationRule = when (mappingContext.privacy) {
```
mariusc83, line 61: shouldn't be `textValueObfuscationRule` ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/NumberPickerMapper.kt
```diff
@@ -153,13 +156,24 @@ internal open class NumberPickerMapper(
...
         )
-        return listOf(
-            prevValueLabelWireframe,
-            topDividerWireframe,
-            selectedValueLabelWireframe,
-            bottomDividerWireframe,
-            nextValueLabelWireframe
-        )
+
+        return if (privacy == SessionReplayPrivacy.ALLOW) {
```
mariusc83, line 160: exactly what I had also in mind ;). Great job here. Later on when we will introduce the more granular privacy control (per element) we can introduce here a privacy resolver which resolves the actual privacy by taking the default one + the first one in the view tree hierarchy for this element, with the latter taken precedency.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/TextViewMapper.kt
mariusc83: Also the one thing we need to consider here: based on my previous metrics, the text value obfuscation part takes quite a considerable time on the UI Thread (especially if the text is big enough). I would add here a follow - up task where we could try to move this operation in the Worker Thread following the same approach we have in place for resolving the Bitmaps from the actual elements.

mariusc83 (approved):
Great job here. Exactly what I also had in mind. In the first place I wanted to stay away from introducing a state in those Mappers but in the end the difference in the `if/else` branches is not that high so we can work with it. It looks more scalable now and will be easier to add the `Fine Grained Masking` later.

PR #2021, opened 2024-05-02
RUM-4371 Extract OpenTelemetry support sdk into a dedicated module

mariusc83:
### What does this PR do?

Following previous discussions and plans in this PR we are performing the following:

1. We are extracting the OpenTelemetry support SDK code into a separated module: `dd-sdk-android-trace-otel` 
2. The ported code from `dd-java-agent` project containing the `CoreTracer` code will remain as part of the `dd-sdk-android-trace` module as this is considered the core tracing module. The idea behind this decision is to have later a better split of tracing API implementations support into dedicated module and all to delegate to the `CoreTracer` into the `dd-sdk-android-trace` module.
3. The newly created `dd-sdk-android-trace-otel` module will not require desugaring in its .aar metadata anymore and we added a safety mechanism in place in case our clients will omit the docs and use this module without desugaring

### Motivation

### Additional Notes

1. Please have in mind that the `dd-java-agent` code containing the `CoreTracer` logic is still using the `java.util.function` package and therefor will require desugaring for Android 23 and below. By discussing with the team we decided that it is fine for this code to stay in the `dd-sdk-android-trace` module as for now it will only be used by the `dd-sdk-android-trace-otel` module which will require desugaring in the release docs. In any case this code will be stripped out by R8 in the clients apps if it is not used.
2. Once this branch will be merged in develop and dogfooding will start the plan is to clean more the `CoreTracer` code and see if we can actually make it compatible for Android 23 and below by removing all the `java.util.function` logic. This is highly doubtable as I already performed some research and this turns out to be quite complicated.
3. I could not add the unit tests for the `ContextStorage.addWrapper` functionality as this is weirdly initialized in the OpenTelemetry code (when the class is loaded) and these tests are too flaky as there can be already the class loaded before the test started due to some weird Junit behaviour.


features/dd-sdk-android-trace-otel/README.md
```diff
@@ -0,0 +1,3 @@
+# Datadog Implementation SDK for OpenTelemetry
```
0xnm, line 1:
```suggestion
# Datadog Trace SDK for Android with OpenTelemetry support
```

or something similar

features/dd-sdk-android-trace-otel/build.gradle.kts
```diff
@@ -0,0 +1,100 @@
...
+            buildConfig = true
+        }
+        buildConfigField(
+            "String",
+            "OPENTELEMETRY_API_VERSION_NAME",
+            "\"${libs.versions.openTelemetry.get()}\""
+        )
+        consumerProguardFiles("consumer-rules.pro")
+    }
+    namespace = "com.datadog.android.opentelemetry"
```
0xnm, line 52: both namespace and root package should be `com.datadog.android.trace.opentelemetry`, because the common definitions are in `com.datadog.android.trace` . Thus, since it is a subset with a specific implementation belonging to the Trace feature, it should be `com.datadog.android.trace.opentelemetry` imo. This keeps the consistency.

features/dd-sdk-android-trace-otel/build.gradle.kts
```diff
@@ -0,0 +1,100 @@
...
+    sourceSets.named("test") {
+        // Required because AGP doesn't support kotlin test fixtures :/
+        java.srcDir("${project.rootDir.path}/dd-sdk-android-core/src/testFixtures/kotlin")
+    }
+}
+
+dependencies {
+    api(project(":dd-sdk-android-core"))
+    api(project(":features:dd-sdk-android-trace"))
+    api(libs.openTelemetryApi)
```
0xnm, line 62: this, probably, should be `implementation`? or are OpenTelemetry-specific methods are exposed directly as a part of Public API?

features/dd-sdk-android-trace-otel/build.gradle.kts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+kotlinConfig(jvmBytecodeTarget = JvmTarget.JVM_11)
+androidLibraryConfig()
+junitConfig()
+javadocConfig()
+dependencyUpdateConfig()
+publishingConfig(
+    "The Tracing feature to use with the Datadog monitoring " +
+        "library for Android applications."
```
0xnm, line 99: need to be updated

features/dd-sdk-android-trace/build.gradle.kts
```diff
@@ -60,7 +52,6 @@ android {
 }
 
 dependencies {
-    coreLibraryDesugaring(libs.desugarJdk)
     api(project(":dd-sdk-android-core"))
     api(libs.openTelemetryApi)
```
0xnm, line 56: if we are extracting `OpenTelemetry` into a dedicated module, why do we keep this one here?

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/CoreTracer.java
```diff
@@ -261,8 +260,25 @@ public ScopeState newScopeState() {
...
+                    Arrays.asList(InternalLogger.Target.USER, InternalLogger.Target.MAINTAINER),
+                    () -> "Trying to CoreTracer implementation on Android 23 and below. " +
+                            "In order for this to properly work you will need to enable coreDesugaring " +
+                            "in your compileOptions",
+                    e,
+                    false,
+                    null);
+            return new NoOpCoreTracerBuilder();
+        }
+    }
```
0xnm, line 279: why we are doing this check? as I understand, Otel is moved to the dedicated module and it was Otel which was using API 24+ classes. So, to my understanding, after this move `dd-sdk-android-trace` should be completely compatible with API 21?

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/CoreTracerBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+import com.datadog.trace.common.sampling.Sampler;
+import com.datadog.trace.common.writer.Writer;
+import com.datadog.trace.core.propagation.HttpCodec;
+import com.datadog.trace.core.taginterceptor.TagInterceptor;
+
+import java.util.Map;
+import java.util.Properties;
+
+public interface CoreTracerBuilder {
+    CoreTracerBuilder serviceName(String serviceName);
```
0xnm, line 24:
```suggestion
public interface CoreTracerBuilder {
   CoreTracerBuilder serviceName(String serviceName);
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/_InternalCoreWriterProvider.kt
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.trace
+
+import com.datadog.android.api.feature.StorageBackedFeature
+
+interface _InternalCoreWriterProvider : StorageBackedFeature {
+    fun getCoreTracerWriter(): com.datadog.trace.common.writer.Writer
+}
```
0xnm, line 13: Can you please describe this one? How it is used, why does it start with `_` and extends `StorageBackedFeature`? We are mixing here feature and provider concept. This one, probably, can be a standalone interface, without extending `StorageBackedFeature`.

features/dd-sdk-android-trace-otel/build.gradle.kts
mariusc83: nope, they are part of the Public API so the clients will use the API through us so we need to expose is as API

features/dd-sdk-android-trace/build.gradle.kts
mariusc83: just a bad copy / paste sorry

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/CoreTracer.java
mariusc83: so I like you asked. Not quite, as I explained the copied code from APM also uses `java.util.function` packages so this is my extra safety in case someone internally in our team uses this code wrongly. There was another approach I had in mind and that was about adding the `java9.util.function` package inside to replicate the `java.util.function` and replace all its usage in the `CoreTracer` code. It worked when I tested this morning with the code from this repo: https://github.com/retrostreams/android-retrostreams/tree/master/src/main/java/java9/util/function

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/_InternalCoreWriterProvider.kt
mariusc83: This is being used here: https://github.com/DataDog/dd-sdk-android/pull/2021/files#diff-1db0ecd609920bfa147ebcdc3403f1534b1d66e6506d8a055297399552783c5fR140 

As you can see it is outside of the module and `unwrap` function only accepts a `<T:Feature>` , therefor the need for that interface to extend `StorageBackedFeature`.
The reason why I added `_` as a prefix is to follow the `_InternalProxy` approach for RN and Flutter to expose that class but mark it as Internal.
Please let me know if you have a better idea to export that `coreTracerWriter` from there in the `dd-sdk-android-trace-otel` module.

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/CoreTracer.java
0xnm: I think once we remove the code we don't need, majority of such usages will go away, and for the rest we can do something `java9.util.function` indeed.

The problem with this check is that most probably during the development devs use emulators with the most recent API. And even if there are instrumented tests on pre-API devices, these are just tests, nobody is checking logcat there. So I don't think application developers will ever see this error before shipping app to prod.

mariusc83: But at least if they ship it to prod by mistake it is not going to crash...that's the reason why I wanted at least to have this safety net. Without it will crash. I agree that once we have the artifact from APM team the `CoreTracer` here will be replaced by that and problem will no longer be there. We will do code cleanup for sure but I already did some tests and some parts cannot be removed so in the end we will have 3 options:
- keep it like before, no safety net, no NoOpAgentTracer and make sure we do not use it wrongly
- keep it like it is proposed in this POC
- use `java9.util.function` substitution. This I tested and works but might have some limitations in some conditions (who knows what we might miss in those substitutions as that package was not maintained in the last 7 years). I this this is a bit hacky in a way, @xgouchet already had an opinion on this when I discussed this morning. Can you add your thoughts here @xgouchet please, tks.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/_InternalCoreWriterProvider.kt
0xnm: I still think it is better to keep them separate and don't use `_` prefix. `Feature` interface is indented for communication with SDK core, and here we are adding something completely not related.

IMO Instead, it is better to have just `CoreWriterProvider` interface, not extending any interface, annotated with `@InternalApi`, which has only one method `getCoreTracerWriter`.

And then in the code:

```
val tracingFeature = sdkCore.getFeature(Feature.TRACING_FEATURE_NAME).unwrap()
...
val coreTracerWriter = (tracingFeature as? `CoreWriterProvider`).getCoreTracerWriter()
```

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/CoreTracer.java
0xnm: My point is that we can come to the complete API 21 support without need for any check/desugaring if we remove redundant code and replace the sites using API 23+ symbols with API 21 compatible equivalents.

mariusc83: Duly noted, and this can be done in 2 steps, keep it like this for now and then later cleaning and replacing those `java.util.function` with compatible equivalents. This will allow us to go ahead with dogfooding and even release, I don't see this a blocker here. I am waiting for other comments here and then we take a final decision.

0xnm: Also this reflection check is not needed for API 24+.

mariusc83: Yes, aware of that will fix it in the real PR now, tks !!

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/android/trace/opentelemetry/intenal/DatadogContextStorageTest.kt
0xnm, line 1: there is a typo in the folder name and in the package name respectively: `intenal` -> `internal`. This applies to other similar test files as well.

features/dd-sdk-android-trace/build.gradle.kts
```diff
@@ -105,5 +95,10 @@ javadocConfig()
...
 publishingConfig(
     "The Tracing feature to use with the Datadog monitoring " +
-        "library for Android applications."
+            "library for Android applications."
 )
+android {
+    lint {
+        lintConfig = file("lint-baseline.xml")
+    }
+}
```
0xnm, line 104:
```suggestion
}

```

features/dd-sdk-android-trace/lint-baseline.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lint>
+    <issue id="NewApi">
+        <ignore path="src/main/java/com/datadog/trace"/> />
```
0xnm, line 4: I guess this is temporary? I thought the idea is to have `dd-sdk-android-trace` completely compatible with API 21 as it was before?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
```diff
@@ -63,12 +61,19 @@ internal class TracingFeature constructor(
...
 
     override fun onStop() {
-        dataWriter = NoOpWriter()
+        legacyTracerWriter = NoOpWriter()
         initialized.set(false)
     }
 
     // endregion
 
+    // region _InternalCoreWriterProvider
```
0xnm, line 70:
```suggestion
    // region InternalCoreWriterProvider
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/InternalCoreWriterProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.trace
+
+/**
+ * Internal interface to provide the core writer to the tracer.
+ */
+interface InternalCoreWriterProvider {
```
0xnm, line 12:
```suggestion
@InternalApi
interface InternalCoreWriterProvider {
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/InternalCoreWriterProvider.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.trace
+
+/**
+ * Internal interface to provide the core writer to the tracer.
```
0xnm, line 10: Should we explain what is the meaning of the `core` here? The type of the returned writer is `com.datadog.trace.common.writer.Writer`, so it has no mention of the core.

sample/kotlin/build.gradle.kts
```diff
@@ -143,11 +143,16 @@ android {
     }
 }
 
+repositories {
+    mavenLocal()
+}
+
```
0xnm, line 149: shouldn't be needed, we consume from the project, not from the published artifact

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+    internalLogger: InternalLogger? = null
+): T {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+        return action()
+    } else {
+        return try {
+            // we are forcing here the checkup as it could be triggered only at runtime and will be too late
+            Class.forName("java.util.function.Function")
+            action()
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
```
0xnm, line 29: if this code block will be executed in the hot path, we need an optimization to use reflection only once

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.os.Build
+import com.datadog.android.api.InternalLogger
+
+internal const val NEEDS_DESUGARING_ERROR_MESSAGE =
+    "Trying to use OpenTelemetry SDK support for Android 23 and below. " +
+        "In order for this to properly work you will need to enable coreDesugaring " +
+        "in your compileOptions"
+
+internal fun <T : Any?> executeSafelyOnAndroid23AndBelow(
+    action: () -> T,
```
0xnm, line 18: you may want to place `action` as last argument to be able to use syntax like `foo(args) { lambda }` for the invocation

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt
```diff
@@ -127,43 +134,51 @@ class OtelTracerProvider(
...
-                .build()
-            coreTracer.addScopeListener(object : ScopeListener {
-                override fun afterScopeActivated() {
-                    val activeSpanContext = coreTracer.activeSpan()?.context()
-                    if (activeSpanContext != null) {
-                        val activeSpanId = activeSpanContext.spanId.toString()
-                        val activeTraceId = activeSpanContext.traceId.toString()
-                        sdkCore.addActiveTraceToContext(activeTraceId, activeSpanId)
+        fun build(): TracerProvider {
+            return executeSafelyOnAndroid23AndBelow(
```
0xnm, line 138: nit: current method name gives an impression that code will be executed only on API 23 and below, which is not true. maybe it is worth to rename it to `executeIfAboveAndroid24`?

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt
```diff
@@ -309,6 +324,25 @@ class OtelTracerProvider(
 // endregion
 
     companion object {
+        init {
+            // We need to add the DatadogContextStorageWrapper and this should be executed before
+            // before io.opentelemetry.context.LazyStorage is loaded in the class loader to take effect.
+            // For now we should assume that our users are using `OtelTracerProvider` first in their code.
+            // Later on maybe we should consider a method DatdogOpenTelemetry.initialize() to be called
```
0xnm, line 331:
```suggestion
            // Later on maybe we should consider a method DatadogOpenTelemetry.initialize() to be called
```

features/dd-sdk-android-trace/lint-baseline.xml
mariusc83: nope it is not temporary and it cannot be compatible as long as it uses the `java.util.function` package. The `CoreTracer` code needs to stay in the `dd-sdk-android-trace` and for now it will be used only by the `dd-sdk-android-trace-otel` module which is going to require desugaring in the docs and also which is protected in case someone omits the docs.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/InternalCoreWriterProvider.kt
mariusc83: yup I will add an explanation there

sample/kotlin/build.gradle.kts
mariusc83: yes was only for local testing

features/dd-sdk-android-trace/lint-baseline.xml
mariusc83: The idea is not to block dogfooding too much. Maybe while dogfooding I can clean that code a bit more and find a way to make it compatible. I doubt it though as I tried at some point and it seems impossible without replacing that package with a `java.util.function.adapter` package, they are being used in too many places.

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt
mariusc83: hmm, but that name suggestion will mean that it will only be executed on `Android24` and above for me :). What if I just call it `executeSafely` ?

0xnm: Normally the method name should reflect when lambda will be executed, not when it won't.

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
mariusc83: yes, I thought about it myself, this code will only be called in the `OtelTracerProvider` builder method which normally should be happening few times in the clients code when an `OtelTracerProvider` is created. I don't think we should not anything else here.

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt
mariusc83: yes definitely and actually the name your proposed it's fine, my mind is a bit blurry after all this nordic weather :)

features/dd-sdk-android-trace-otel/api/apiSurface
```diff
@@ -0,0 +1,15 @@
...
+    constructor(com.datadog.android.api.SdkCore = Datadog.getInstance())
+    fun build(): io.opentelemetry.api.trace.TracerProvider
+    fun setTracingHeaderTypes(Set<com.datadog.android.trace.TracingHeaderType>): Builder
+    fun setService(String): Builder
+    fun setPartialFlushThreshold(Int): Builder
+    fun addTag(String, String): Builder
+    fun setSampleRate(Double): Builder
+    fun setBundleWithRumEnabled(Boolean): Builder
+  override fun toString(): String
+  companion object
```
0xnm, line 15: minor: need to make `companion` object `private` or `internal`, because it doesn't expose any public members in the current state

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
```diff
@@ -0,0 +1,41 @@
...
+    defaultActionReturnValue: T,
+    action: () -> T
+): T {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+        return action()
+    } else {
+        return try {
+            // we are forcing here the checkup as it could be triggered only at runtime and will be too late
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            Class.forName("java.util.function.Function")
```
0xnm, line 28: are we sure that in case if desugaring applied there will be actually such class in the classpath (btw, if we use reflection, we need to add a rule to consumer Proguard rules) and desugaring won't instead inline function calls?

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerBuilderProviderTest.kt
```diff
@@ -151,7 +153,24 @@ internal class OtelTracerBuilderProviderTest {
...
+
+        // WHEN
+        val tracer = testedOtelTracerProviderBuilder.build()
+
+        // THEN
+        assertThat(tracer).isNotNull
+        mockInternalLogger.verifyLog(
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.USER,
+            OtelTracerProvider.TRACING_NOT_ENABLED_ERROR_MESSAGE
```
0xnm, line 168: how then we are going distinguish between the cases when Tracing is not enabled vs there is a mistake from our side by not implementing the necessary interface? Shouldn't we introduce different errors for these cases?

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtilsTest.kt
```diff
@@ -0,0 +1,89 @@
...
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class MiscUtilsTest {
+
+    @Mock
+    lateinit var mockInternalLogger: InternalLogger
+
+    @Test
+    fun `M return defaultValue and log exception W executeSafelyOnAndroid23AndBelow{action throws exception, below 24}`(
```
0xnm, line 39:
```suggestion
    fun `M return defaultValue and log exception W executeIfJavaFunctionPackageExists {action throws exception, below 24}`(
```

features/dd-sdk-android-trace-otel/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+    }
+    namespace = "com.datadog.android.trace.opentelemetry"
+    sourceSets.named("test") {
+        // Required because AGP doesn't support kotlin test fixtures :/
+        java.srcDir("${project.rootDir.path}/dd-sdk-android-core/src/testFixtures/kotlin")
+    }
+}
+
+dependencies {
+    api(project(":dd-sdk-android-core"))
```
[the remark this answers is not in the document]
mariusc83: yes but I mentioned this usage on other modules also, I think for visibility

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
mariusc83: I checked a desugared application and there is a `java.util.function` package added in the `.dex` package at compile time. I am not really sure we need proguard rule here. Everything we are doing here is quite experimental and we rely on the fact that the desugaring will always work like that.

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtilsTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        @StringForgery defaultValue: String
+    ) {
+        // Given
+        val exception = Exception()
+        val action: () -> String = { throw exception }
```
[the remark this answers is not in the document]
mariusc83: Let's have a chat about this as I am not sure I understand exactly what you mean. One note from my part is that we will not be able to check the whole package existence through reflection unless we will go one by one. The more we discuss about this the more I think that maybe this safety mechanism is not the best idea and we should just go with desugaring required.

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
0xnm: It is for obfuscation. Production app will have `java.util.function.Function` renamed.

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtilsTest.kt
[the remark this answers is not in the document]
0xnm: Let's say Otel code is updated with some class which exists on say API 25. `Class.forName("function")` will be fine, but when `action()` is executed it will be `NoClassDefFound**Error**` for this new class, not exception. But imo with current state of the code it is fine.

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerBuilderProviderTest.kt
mariusc83: good point, it makes sense in this case I guess to have different messages. Maybe we should also change the `target` in this case to `MAINTAINER` ?

0xnm: for the case when checking internal interface - yes, for the case when checking Feature - it should stay with User target.

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
0xnm, line 28: btw, `Class.forName` may throw not only exceptions, but also Errors https://developer.android.com/reference/java/lang/Class#forName(java.lang.String), they won't be called by catch block in the current state.

brett0000FF (commented):
Docs were flagged to review due to `features/dd-sdk-android-trace-otel/README.md`. Do you intend to add more content here? Let me know, and we can take a look!

mariusc83:
> Docs were flagged to review due to `features/dd-sdk-android-trace-otel/README.md`. Do you intend to add more content here? Let me know, and we can take a look!

@brett0000FF yes we have a special task for docs in the README.md + docs in the public page.

features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/internal/MiscUtils.kt
mariusc83: yup maybe I will change it for `Throwable` instead to make sure.

0xnm (approved):
Approving this PR, but there are things which should be addressed in the later PRs:

* we shouldn't have `NewApi` errors suppressed
* all `NewApi` should be addressed

PR #2026, opened 2024-05-09
RUM-3589: Add UpdateView and event calls monitoring

jonathanmos:
### What does this PR do?
• Adds a new UpdateView event to update the view immediately with attribute changes when performing addAttribute/removeAttribute
• Adds the new apis addAttributes/removeAttributes. AddAttributes receives a map of KV, and removeAttributes receives a set of keys
• Adds RumEventCallMonitor to warn the user that more than X calls were made in a time period and that this may impact performance

### Motivation
Solve the issue where modifying attributes was not updating for the view until after the next event was sent

### Additional Notes


xgouchet (changes requested):
> [!CAUTION] 
> You're adding a lot of changes in a single PR (and apparently a single commit), which makes it harder to read, and might add more friction in the comments. In addition, all of those being a single commit means that when we look in our git history, it'll be harder to understand as well. Could you please split this in three different commits and PRs please?

PR #2027, opened 2024-05-10
RUM-4413 Add telemetry on SR resources track

xgouchet:
### What does this PR do?

In order to better read the performance of the Image Resource upload for Session Replay, we add the telemetry on batch deleted/closed/expired with a dedicated track name.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/Feature.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/Feature.kt
@@ -55,6 +55,11 @@ interface Feature {
         const val SESSION_REPLAY_FEATURE_NAME: String = "session-replay"
 
         /**
+         * Session Replay Resources sub-feature name.
+         */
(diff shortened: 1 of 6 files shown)


jonathanmos (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/Feature.kt
```diff
@@ -54,6 +54,11 @@ interface Feature {
          */
         const val SESSION_REPLAY_FEATURE_NAME: String = "session-replay"
 
+        /**
+         * Session Replay Resources sub-feature name.
+         */
+        const val SESSION_REPLAY_RESOURCES_FEATURE_NAME: String = "session-replay-resources"
```
jonathanmos, line 60: By adding this here, the feature name for Resources is now defined in two places - in Feature and also in the feature itself. I notice that this is also the case for Session Replay. Do we want to consolidate both of them to use only the constants from Feature and have a SST?

0xnm (approved):

PR #2028, opened 2024-05-10
RUM-4103 add telemetry to detect uncovered View/Drawable in Session Replay

xgouchet:
### What does this PR do?

Adds some telemetry when a View or Drawable cannot be converted into Session Replay data.

### Motivation

We need to better understand the UI features used by our customer that we don't have support of yet.

### DIFF
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversalTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversalTest.kt
@@ -16,6 +16,7 @@ import android.widget.LinearLayout
 import android.widget.RadioButton
 import android.widget.TextView
 import androidx.drawerlayout.widget.DrawerLayout
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.MapperTypeWrapper
 import com.datadog.android.sessionreplay.forge.ForgeConfigurator
 import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueRefs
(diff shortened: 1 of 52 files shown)


features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversalTest.kt
```diff
@@ -131,28 +138,55 @@ internal class TreeViewTraversalTest {
     @Test
     fun `M default to view mapper and continue W traverse { mapper for view type not provided }`(
```
jonathanmos, line 139: I think the description for this test may be wrong, since this test is traversing a single view

jonathanmos (approved):

features/dd-sdk-android-session-replay-material/src/main/kotlin/com/datadog/android/sessionreplay/material/MaterialExtensionSupport.kt
```diff
@@ -29,7 +30,7 @@ class MaterialExtensionSupport : ExtensionSupport {
     private val viewIdentifierResolver: ViewIdentifierResolver = DefaultViewIdentifierResolver
     private val colorStringFormatter: ColorStringFormatter = DefaultColorStringFormatter
     private val viewBoundsResolver: ViewBoundsResolver = DefaultViewBoundsResolver
-    private val drawableToColorMapper: DrawableToColorMapper = DrawableToColorMapper.getDefault()
+    private val drawableToColorMapper: DrawableToColorMapper = DrawableToColorMapper.getDefault(InternalLogger.UNBOUND)
```
0xnm, line 33: It means no telemetry will be sent for this one, because `InternalLogger.UNBOUND` is not bound to any SDK instance (and hence to the RUM feature). Is it supposed to be like that?

0xnm (approved):

features/dd-sdk-android-session-replay-material/src/main/kotlin/com/datadog/android/sessionreplay/material/MaterialExtensionSupport.kt
xgouchet: Good point thanks

PR #2029, opened 2024-05-12
RUM-3589: Update RUM View when attributes are modified - 1

jonathanmos:
### What does this PR do?
• Adds a new UpdateView event to update the view immediately with attribute changes when performing addAttribute/removeAttribute

### Motivation

### Additional Notes


0xnm (commented):
I have a concern regarding this change (sending view updates on the global attributes update).

PR says:

> Fixes an issue where performing addAttribute/removeAttribute is not reflected in the RUM dashboard until after the next event

Is it really an issue? Normally, when developing, it is not an issue to send next event. And when application is in production, people don't track individual sessions real-time.

I can think of about the crash, but even in case of the crash we send view update, so. changes to the global attributes should be reflected there.

Probably we are trying to solve the issue when global attribute is set **after view is started** and there are no subsequent events which would cause view update?

Also I have a concern about backpressure. The change we are making may add quite a lot of pressure to the RUM executor. Adding `addAttributes` is fine to have batching, but for the customers who already integrated our SDK it will be still `addAttribute` call, without any batching.

PR #2032, opened 2024-05-14
Create package name consistency rule

xgouchet:
### What does this PR do?

This PR include a new Detekt rule enforcing our package naming policy: 

- A top level named declaration (ie: `(enum|data|sealed)? class`, `object`, `interface`, `fun`, `val|var`) which is marked as `internal` must be in a package matching `/**.internal.**/` 
- A top level named declaration (ie: `(enum|data|sealed)? class`, `object`, `interface`, `fun`, `val|var`) which is not marked as `internal` must not be in a package matching `/**.internal.**/`
- this rule ignores issues on annotated declarations


integrations/dd-sdk-android-compose/src/test/kotlin/com/datadog/android/compose/ComposeNavigationObserverTest.kt
0xnm, line 1: this file should be moved to the `internal` package, to match class under the test

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/utils/RequestUniqueIdentifierTest.kt
```diff
@@ -16,7 +16,7 @@ import okhttp3.Request
 import okhttp3.RequestBody
 import okhttp3.RequestBody.Companion.toRequestBody
 import okio.BufferedSink
-import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions
```
0xnm, line 19: to be reverted? doesn't seem to be related with the change

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/PackageNameVisibility.kt
```diff
@@ -0,0 +1,101 @@
...
+    ruleSetConfig: Config
+) : AbstractTypedRule(ruleSetConfig) {
+
+    private val withBreakingChanges: Boolean by config(defaultValue = true)
+    private val ignoredAnnotations: List<String> by config(defaultValue = emptyList())
+
+    override val issue: Issue = Issue(
+        javaClass.simpleName,
+        Severity.Warning,
+        "This rule reports when a top level declaration's visibility doesn't match the package naming policy.",
```
0xnm, line 43: semantically we then have a problem with classes generated from `@NoOpImplementation` declaration - they will be `internal` by default, but in the same package (which may be non-internal) as the original interface.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/MapUtils.kt
```diff
@@ -12,7 +12,7 @@ package com.datadog.android.core.internal.utils
  * Special value for missing attribute.
  */
 @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
-val NULL_MAP_VALUE: Object = Object()
+internal val NULL_MAP_VALUE: Object = Object()
```
0xnm, line 15: I think it is a blocking change. This one should be unique, because usage site in the core is `when (item) NULL_MAP_VALUE -> ...`, which is translated to `Intrinsics.areEqual(item, MapUtilsKt.getNULL_MAP_VALUE())`, which under the hood is doing `first == null ? second == null : first.equals(second)`, so objects are compared by reference.

In this PR `NULL_MAP_VALUE` is redeclared in Logs and Traces, and while for Trace it is fine, since the usage there doesn't go to the core, in case of Logs supplying another object will break the flow.

0xnm (commented):
Thanks for the cleanup! I left some comments. My concern is about the usage of `@InternalApi` on the interfaces/non-final classes directly, because it is a limitation of the check, but if we are sure that implementation of the interface/class extension won't land in the legit public API, this is fine.

0xnm (approved):

PR #2034, opened 2024-05-15
Improve the TextViewMapper unit tests

xgouchet:
### What does this PR do?

Split the logic of the TextViewMapper in two classes: TextViewMapper and EditTextViewMapper, to simplify some of the logic and make testing easier.


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/DefaultRecorderProvider.kt
```diff
@@ -98,7 +100,7 @@ internal class DefaultRecorderProvider(
             MapperTypeWrapper(
                 SwitchCompat::class.java,
                 SwitchCompatMapper(
-                    textViewMapper,
+                    textViewMapper as TextViewMapper<SwitchCompat>,
```
0xnm, line 103: this part with explicit casting looks dangerous and not very obvious (partially because it is not obvious that `SwitchCompat`, etc. are sub-classes of `TextView`). Is there a way to avoid explicit cast?

0xnm (approved):

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/DefaultRecorderProvider.kt
xgouchet: Those cast are totally safe, but I agree they are not obvious. 
They are temporary though, the next PRs will be about letting `SwitchCompatMapper`(and others) extend `TextViewMapper` instead.

PR #2038, opened 2024-05-16
RUM-4098: Add TLVFormat DataStore persistence

jonathanmos:
### What does this PR do?
Adds the capability to write/read files in TLVformat, and defines a DataStore type that features can use to persist data between sessions. 

### Motivation
Part of the effort to persist bitmap hashes between sessions. Integrating the datastore with SR will be in a second PR. 

### Additional Notes


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
```diff
@@ -0,0 +1,126 @@
...
+        val tlvHeader = TLVBlockType.fromValue(shortValue.toUShort())
+
+        if (tlvHeader == null) {
+            logTypeCorruptionError(shortValue)
+        }
+
+        return tlvHeader
+    }
+
+    private fun readData(stream: InputStream): ByteArray {
```
mariusc83, line 71: if we read everything in one buffer isn't it going to blow up the RAM memory ? I am only asking because usually with buffers we should work with smaller sizes in order to maintain the performance. Not sure if we can avoid having all the data in memory at once here though.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+    private val dataStoreFileHelper: DataStoreFileHelper
+) : DataStoreHandler {
+    @WorkerThread
+    override fun <T : Any> write(
+        dataStoreFileName: String,
+        featureName: String,
+        serializer: Serializer<T>,
+        data: T
+    ) {
+        val dataStoreDirectory = createDataStoreDirectoryIfNecessary(featureName)
```
0xnm, line 44: minor: this call can be probably made inside `createDataStoreFileIfNecessary`

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+    private val storageDir: File,
+    private val internalLogger: InternalLogger,
+    private val fileReaderWriter: FileReaderWriter,
+    private val fileTLVBlockReader: FileTLVBlockReader,
+    private val dataStoreFileHelper: DataStoreFileHelper
+) : DataStoreHandler {
+    @WorkerThread
+    override fun <T : Any> write(
+        dataStoreFileName: String,
+        featureName: String,
```
0xnm, line 40: should we simply host this handler in `SdkFeature` instead of `CoreFeature`? this will eliminate the need to pass `featureName` in the method and instead we could pass it in the constructor. This makes more sense.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+
+        val lastUpdateBlock = getLastUpdateDateBlock()
+        val versionCodeBlock = getVersionCodeBlock()
+        val dataBlock = getDataBlock(data, serializer)
+
+        if (lastUpdateBlock == null || versionCodeBlock == null || dataBlock == null) return
+
+        writeToFile(
+            dataStoreFile,
+            lastUpdateBlock + versionCodeBlock + dataBlock
```
0xnm, line 55: performance nit: pretty sure that this operation will be like:

* create a byte array in the memory to host result of `lastUpdateBlock + versionCodeBlock`
* create a byte array in the memory to host result of above and ` + dataBlock`.

So to have a final result we will have 2 arrays/allocations (unless some very smart compiler optimization kicks in). Instead you can use this method https://github.com/DataDog/dd-sdk-android/blob/develop/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/utils/ByteArrayExt.kt#L48 - it will make a single allocation.

```suggestion
            listOf(lastUpdateBlock, versionCodeBlock, dataBlock).join(separator = emptyByteArray())
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+        dataStoreFileName: String
+    ): File {
+        val datastoreFile = dataStoreFileHelper.getDataStoreFile(
+            dataStoreDirectory = dataStoreDirectory,
+            dataStoreFileName = dataStoreFileName
+        )
+
+        if (!datastoreFile.existsSafe(internalLogger)) {
+            datastoreFile.createNewFileSafe(internalLogger)
+        }
```
0xnm, line 270: this is not needed, I think. Once you have a `File` object, underlying file will be created automatically once write operation is executed.

This leads to the removal of this method, because then the only thing it does is calling `getDataStoreFile`

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+        )
+
+        if (!datastoreFile.existsSafe(internalLogger)) {
+            return null // no datastore file found
+        }
+
+        return readFromDataStoreFile(datastoreFile, deserializer, fileTLVBlockReader, version)
+    }
+
+    @WorkerThread
```
0xnm, line 85: do we need annotation on all these private methods? all the public entry points in this class already have `@WorkerThread` annotations.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+    @Suppress("ReturnCount")
+    private fun <T : Any> readFromDataStoreFile(
+        datastoreFile: File,
+        deserializer: Deserializer<String, T>,
+        fileTLVBlockReader: FileTLVBlockReader,
+        requestedVersion: Int
+    ): T? {
+        val tlvBlocks = fileTLVBlockReader.all(datastoreFile)
```
0xnm, line 145: maybe `read` is a better name here? Kotlin is using `all` method to check if all items satisfy criteria.

```suggestion
        val tlvBlocks = fileTLVBlockReader.read(datastoreFile)
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+
+        // there should be as many blocks read as there are block types
+        if (tlvBlocks.size != TLVBlockType.values().size) {
```
0xnm, line 148: is it a blocker? what if in SDK version b we added new block type which didn't exist in SDK version a. This shouldn't prevent us from reading block that we know, isn't it?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+            logInvalidNumberOfBlocksError(tlvBlocks.size)
+            return null
+        }
+
+        val dataStoreContents = tryToMapToDataStoreContents(deserializer, tlvBlocks)
+            ?: return null
+
+        val fileVersionIsWrong = dataStoreContents.versionCode != requestedVersion
+        val fileIsTooOld = isDataStoreTooOld(dataStoreContents.lastUpdateDate)
```
0xnm, line 157: why this limitation comes? what prevents us from reading data that is older than 30 days?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+
+        // the version check is a double redundancy. Since we store each file in a folder
+        // whose name contains the version it should be impossible to read a file
+        // that contains the wrong version.
+        if (fileVersionIsWrong) {
+            logInvalidVersionError()
+        }
```
0xnm, line 164: why do we need it then?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
```diff
@@ -0,0 +1,126 @@
...
+import com.datadog.android.core.internal.persistence.file.FileReaderWriter
+import com.datadog.android.core.internal.persistence.file.existsSafe
+import com.datadog.android.core.internal.persistence.file.lengthSafe
+import java.io.File
+import java.io.IOException
+import java.io.InputStream
+import java.util.Locale
+import java.util.concurrent.atomic.AtomicBoolean
+
+internal class FileTLVBlockReader(
```
0xnm, line 22:
```suggestion
internal class TLVBlockFileReader(
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
```diff
@@ -0,0 +1,126 @@
...
+    val internalLogger: InternalLogger,
+    val fileReaderWriter: FileReaderWriter
+) {
+    private var endOfStream = AtomicBoolean(false)
+
+    @WorkerThread
+    internal fun all(file: File): List<TLVBlock> {
+        if (!file.existsSafe(internalLogger) || file.lengthSafe(internalLogger) == 0L) {
```
0xnm, line 30: aren't we are doing `existsSafe` check in the caller? Also `lengthSafe` check seems redundant, we can read empty file below it seems and return empty list.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
```diff
@@ -0,0 +1,126 @@
...
+import java.io.IOException
+import java.io.InputStream
+import java.util.Locale
+import java.util.concurrent.atomic.AtomicBoolean
+
+internal class FileTLVBlockReader(
+    val internalLogger: InternalLogger,
+    val fileReaderWriter: FileReaderWriter
+) {
+    private var endOfStream = AtomicBoolean(false)
```
0xnm, line 26: this is the source of bugs I believe - by having it a shared property you may have a situation where different `write` requests use/update it concurrently. It should be a local variable at least.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
```diff
@@ -0,0 +1,126 @@
...
+            return arrayListOf()
+        }
+
+        val blocks = mutableListOf<TLVBlock>()
+        val stream = fileReaderWriter.readData(file).inputStream()
+
+        while (!endOfStream.get()) {
+            val nextBlock = readBlock(stream) ?: break
+            blocks.add(nextBlock)
+        }
```
0xnm, line 40: `endOfStream` is not needed here actually. If you want to keep the look here, do can do `do-while` and use `nextBlock` as a marker.

```
var nextBlock: TLVBlock? = null
do {
  nextBlock = readBlock(stream)
  if (nextBlock != null) block.add(nextBlock)
} while (nextBlock != null)
```

there are other ways to re-write it as well.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
```diff
@@ -0,0 +1,126 @@
...
+    private var endOfStream = AtomicBoolean(false)
+
+    @WorkerThread
+    internal fun all(file: File): List<TLVBlock> {
+        if (!file.existsSafe(internalLogger) || file.lengthSafe(internalLogger) == 0L) {
+            return arrayListOf()
+        }
+
+        val blocks = mutableListOf<TLVBlock>()
+        val stream = fileReaderWriter.readData(file).inputStream()
```
0xnm, line 35: if you open stream, you need also to close it. Also you probably want to have it buffered in order to reduce the number of syscalls. You can check some code here https://github.com/DataDog/dd-sdk-android/blob/a7837387c6a137beb317d49c8cdc66835c62aed4/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt#L126-L141

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -156,6 +159,36 @@ internal class SdkFeature(
...
+        serializer: Serializer<T>,
+        data: T
+    ) {
+        coreFeature.dataStoreHandler.write(
+            dataStoreFileName = dataStoreFileName,
+            featureName = featureName,
+            serializer = serializer,
+            data = data
+        )
+    }
```
0xnm, line 174: it means the caller should care about calling it on the worker thread? should we do an explicit dispatch to the persistence thread here? don't know the usage patterns though

0xnm (changes requested):
good start! I reviewed production files so far, didn't go through the tests yet. I left some questions/suggestions.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
mariusc83: I agree with @0xnm here. I would rather use the persistence thread here and make this operations serial. This will also get rid of all the possible race conditions you could experience.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -156,6 +159,36 @@ internal class SdkFeature(
...
+
+    override fun <T : Any> readFromDataStore(
```
mariusc83, line 161: if we use the persistence thread I think this operation could contain a callback as a lambda which will return the `T?` value

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
```diff
@@ -0,0 +1,289 @@
...
+        if (lastUpdateBlock == null || versionCodeBlock == null || dataBlock == null) return
+
+        writeToFile(
+            dataStoreFile,
+            lastUpdateBlock + versionCodeBlock + dataBlock
+        )
+    }
+
+    @WorkerThread
+    override fun <T : Any> read(
```
mariusc83, line 60: are these operations Thread safe ? I am not able to see where do we make sure that we do not read a corrupted file (meaning that we are waiting for reading as long as the writing is in progress). A good approach here would be to use this: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html In case you want to keep the same logic. The other approach (recommended one) is to use the persistence executor service which makes sure all the access to these files is serial.

mariusc83 (changes requested):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -29,6 +31,43 @@ interface FeatureScope {
         callback: (DatadogContext, EventBatchWriter) -> Unit
     )
 
+    /**
+     * Write data to the datastore.
+     *
+     * @param T datatype of the data to write to the datastore.
+     * @param dataStoreFileName name of the datastore file as there could be multiple such files per feature.
```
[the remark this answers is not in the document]
jonathanmos: I think the rationale on ios related to how difficult it would be to handle corruption in a large file vs in multiple smaller ones, and avoiding the complexity of seeking/updating records

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
jonathanmos: yes, agreed - I've modified this to remove the variable as you suggested

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/FileTLVBlockReader.kt
jonathanmos: I've changed the inputStream to be buffered. It's possible that in an image heavy application we could have a very large number of resource hashes being written every time, and we're in any case keeping them always in memory in order to avoid sending them to the backend. If we see that this causes a performance hit we may need to rethink this mechanism (also on ios). If I'm not mistaken every md5 hash is 16 bytes, so we would have around 65000 Strings per mb, and we can try to optimize the encoding to use less space.

jonathanmos: On ios I see that they've put in a maximum length of 10mb per entry and if it attempts to exceed this size then it throws an exception - I'll implement a similar mechanism to prevent extremely large edge cases.

0xnm: Actually you don't need to have an input stream at all. Input stream is designed to back potentially infinite data with a cost of handling a bunch of possible `IOException` types.

Here you have everything as a `ByteArray` directly if I'm not wrong, so everything is already loaded in the memory and it finite, you can work with an array directly, without coming to the streams.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
jonathanmos: I'm moving the version validation logic to the calling feature. To align with ios, version will be different for the directory and for the entries - if read is called without specifying the version then data will always be returned if an entry with that key is found. If version is specified then if the entry has a different version no data will be returned. When writing, specifying a version will write with that version, not specifying a version will write the entry with version 0 (even if it downgrades from a previous higher version number on the entry)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -29,6 +31,43 @@ interface FeatureScope {
...
+     * @param featureName of the calling feature, to determine the path to the datastore file.
+     * @param serializer to use to serialize the data.
+     * @param data to write.
+     */
+    fun <T : Any> writeToDataStore(
+        dataStoreFileName: String,
```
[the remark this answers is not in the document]
jonathanmos: This is a good point and there's a question if we should replace the slashes in the path or log an exception. I'm in favor of the latter, since otherwise we're causing the user to believe that there's a key in the datastore that doesn't actually exist as they provided it

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/FileDataStoreHandler.kt
jonathanmos: On ios they have the same exception that they call "unexpectedBlocks", so it's a question of whether we should be aligned on the behavior

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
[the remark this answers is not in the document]
jonathanmos: I've removed the validation now that this is called "key"

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/persistence/datastore/DataStoreCallback.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core.persistence.datastore
```
0xnm, line 7: should be in `com.datadog.android.api.storage.datastore`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/persistence/datastore/DataStoreCallback.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Called when an exception occurred getting data from the datastore.
+     */
+    fun onFailure()
+
+    /**
+     * Called when no data is found for the given key.
+     */
+    fun onNoData()
```
0xnm, line 29: side note: maybe we can reduce the number of methods in this interface by doing argument of `onSuccess` nullable and maybe renaming it to `onResult` or something like that, then we won't need this method, because no data is also successful result in some sense.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/persistence/datastore/DataStoreContent.kt
```diff
@@ -0,0 +1,19 @@
...
+package com.datadog.android.core.persistence.datastore
+
+/**
+ * Datastore entry contents and metadata.
+ *
+ * @param T type of data used by this entry in the datastore.
+ * @property versionCode version used by the entry.
+ * @property data content of the entry.
+ */
+data class DataStoreContent<T : Any>(
```
0xnm, line 16: if we make `DataStoreCallback` to return just `T`, then this class could be internal.

detekt_custom.yml
```diff
@@ -997,6 +1002,7 @@ datadog:
       - "kotlin.Boolean.hashCode()"
       - "kotlin.Byte.toInt()"
       - "kotlin.ByteArray.constructor(kotlin.Int)"
+      - "kotlin.ByteArray.inputStream()"
```
0xnm, line 1005: can be removed, not streams usage in this PR

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -424,6 +427,33 @@ internal class SdkFeatureTest {
 
     // region FeatureScope
 
+    @Test
+    fun `M unregister datastore W stop()`() {
+        // Given
+        testedFeature.initialize(appContext.mockInstance, fakeInstanceId)
+
+        assertThat(testedFeature.dataStore)
+            .isInstanceOf(DataStoreHandler::class.java)
```
0xnm, line 436: this line is useless, because `testedFeature.dataStore` has `DataStoreHandler` type anyway by declaration.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/SdkFeatureTest.kt
```diff
@@ -424,6 +427,33 @@ internal class SdkFeatureTest {
...
+    }
+
+    @Test
+    fun `M register datastore W initialize()`() {
+        // When
+        testedFeature.initialize(appContext.mockInstance, fakeInstanceId)
+
+        // Then
+        assertThat(testedFeature.dataStore)
+            .isInstanceOf(DataStoreHandler::class.java)
```
0xnm, line 453: same as above

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlock.kt
```diff
@@ -0,0 +1,60 @@
...
+        val dataFieldSize = data.size
+
+        val entrySize = typeFieldSize + dataLengthFieldSize + dataFieldSize
+
+        if (entrySize > maxEntrySize) {
+            logEntrySizeExceededError(maxEntrySize)
+            return null
+        }
+
+        val tlvTypeAsShort = type.rawValue.toShort()
```
0xnm, line 33: minor nit: this is potentially lossy conversion because of the `UShort` vs `Short`. But ofc with our usage we will never hit it.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlock.kt
```diff
@@ -0,0 +1,60 @@
...
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        return ByteBuffer
+            .allocate(entrySize)
+            .putShort(tlvTypeAsShort)
+            .putInt(dataFieldSize)
+            .put(data)
+            .array()
+    }
+
+    private fun logEntrySizeExceededError(maxEntrySize: Int) {
```
0xnm, line 46: maybe it is worth to log the actual size as well?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DatastoreFileReader.kt
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        val dataStoreContent = tryToMapToDataStoreContents(deserializer, tlvBlocks)
+
+        if (dataStoreContent == null) {
+            callback.onFailure()
+            return
+        }
+
+        if (requestedVersion != 0 && dataStoreContent.versionCode != requestedVersion) {
```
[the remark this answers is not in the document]
jonathanmos: The versioning is separate. Datastore version (for the folder structure) only changes if we want to make a breaking change to the TLV storage of the datastore files. Entry version is per key, and is there to allow the caller to specify that they only want the entry if it matches a specific version.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/NoOpDataStoreHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.datadog.android.core.internal.persistence.datastore
+
+import com.datadog.android.core.internal.persistence.Deserializer
+import com.datadog.android.core.persistence.Serializer
+import com.datadog.android.core.persistence.datastore.DataStoreCallback
+import com.datadog.android.core.persistence.datastore.DataStoreHandler
+
+internal class NoOpDataStoreHandler : DataStoreHandler {
```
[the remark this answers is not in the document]
jonathanmos: I added this because the NoOpFactorySymbolProcessor doesn't seem to correctly support generics

mariusc83 (approved):
LGTM !!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -9,12 +9,18 @@ package com.datadog.android.api.feature
...
 
 /**
  * Represents a Datadog feature.
  */
 interface FeatureScope {
 
+    /**
+     * Property to enable interaction with the DataStore.
+     */
+    var dataStore: DataStoreHandler
```
0xnm, line 22: blocking: It should be immutable property here, we don't want to have a possibility to modify it at the call site.

```suggestion
    val dataStore: DataStoreHandler
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -9,12 +9,18 @@ package com.datadog.android.api.feature
...
 import com.datadog.android.api.storage.EventBatchWriter
+import com.datadog.android.api.storage.datastore.DataStoreHandler
 
 /**
  * Represents a Datadog feature.
  */
 interface FeatureScope {
 
+    /**
+     * Property to enable interaction with the DataStore.
```
0xnm, line 20: nit: there is no `DataStore` class/reference per se. We need maybe to use `DataStoreHandler` reference or write `data store`

```suggestion
     * Property to enable interaction with the data store.
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/datastore/DataStoreCallback.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Callback for asynchronous operations on the datastore.
+ */
+interface DataStoreCallback<T : Any> {
+
+    /**
+     * Triggered on successfully fetching data from the datastore.
+     *
+     * @param dataStoreContent contains the datastore content if there was data to fetch.
```
0xnm, line 19: missing description in which case we can get `null` here, given that we also have `onFailure` callback.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/datastore/DataStoreCallback.kt
```diff
@@ -0,0 +1,27 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.api.storage.datastore
+
+import com.datadog.android.core.persistence.datastore.DataStoreContent
+
+/**
+ * Callback for asynchronous operations on the datastore.
```
0xnm, line 12: nit: missing `<T>` type parameter description

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/datastore/DataStoreCallback.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Triggered on successfully fetching data from the datastore.
+     *
+     * @param dataStoreContent contains the datastore content if there was data to fetch.
+     */
+    fun onSuccess(dataStoreContent: DataStoreContent<T>?)
+
+    /**
+     * Triggered when an exception occurred getting data from the datastore.
```
0xnm, line 24: nit: maybe we should say that it is triggered when we cannot get data? we say it is triggered on exception, but there is no exception in the argument.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/datastore/DataStoreHandler.kt
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+    /**
+     * Read data from the datastore.
+     *
+     * @param T datatype of the data to read from the datastore.
+     * @param key name of the datastore entry.
+     * @param version optional version to use when reading from the datastore.
+     * If specified, will only return data if the persistent entry exactly matches this version number.
+     * @param callback to return result asynchronously.
```
0xnm, line 41: worth mentioning that is will be called on the worker thread and not on the caller thread, maybe?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DataStoreFileHandler.kt
```diff
@@ -0,0 +1,51 @@
...
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.storage.datastore.DataStoreCallback
+import com.datadog.android.api.storage.datastore.DataStoreHandler
+import com.datadog.android.core.internal.persistence.Deserializer
+import com.datadog.android.core.internal.utils.submitSafe
+import com.datadog.android.core.persistence.Serializer
+import java.util.concurrent.ExecutorService
+
+internal class DataStoreFileHandler(
```
0xnm, line 17: should we have any internal method (not exposed via public API) to remove all the content of data store? Given that we have `Datadog.clearAllData` method, should it clear also the data store?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DatastoreFileReader.kt
```diff
@@ -0,0 +1,130 @@
...
+        internalLogger.log(
+            level = InternalLogger.Level.ERROR,
+            target = InternalLogger.Target.MAINTAINER,
+            messageBuilder = { UNEXPECTED_BLOCKS_ORDER_ERROR }
+        )
+    }
+
+    internal companion object {
+        internal const val INVALID_NUMBER_OF_BLOCKS_ERROR =
+            "Read error - datastore entry has invalid number of blocks. Was: %s expected: %s"
```
0xnm, line 126:
```suggestion
            "Read error - datastore entry has invalid number of blocks. Was: %s, expected: %s"
```

Also shouldn't it be `%d` here instead of `%s`, given that arguments are numbers? But probably for integers it doesn't matter.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DataStoreFileHandler.kt
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    override fun <T : Any> value(
+        key: String,
+        version: Int?,
+        callback: DataStoreCallback<T>,
+        deserializer: Deserializer<String, T>
+    ) {
+        executorService.submitSafe("dataStoreRead", internalLogger) {
```
0xnm, line 47: actually you don't need to call `submitSafe`, because you are not interested in the returned `Future` value. You can simply use `executeSafe`.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/datastore/DataStoreFileHandlerTest.kt
```diff
@@ -0,0 +1,179 @@
...
+    lateinit var fakeDataString: String
+
+    private lateinit var fileCallback: DataStoreCallback<ByteArray>
+
+    @BeforeEach
+    fun setup() {
+        whenever(mockExecutorService.submit(any())) doAnswer {
+            it.getArgument<Runnable>(0).run()
+            StubFuture()
+        }
```
0xnm, line 82: If you use `executeSafe` in the `DataStoreFileHandler`, `StubFuture` is not needed.

```suggestion
        whenever(mockExecutorService.execute(any())) doAnswer {
            it.getArgument<Runnable>(0).run()
        }
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/datastore/DataStoreFileReaderTest.kt
```diff
@@ -0,0 +1,289 @@
...
+                override fun onSuccess(dataStoreContent: DataStoreContent<ByteArray>?) {
+                    assertThat(dataStoreContent).isNull()
+                }
+
+                override fun onFailure() {
+                    // should not get here
+                    assertThat(1).isEqualTo(2)
+                }
+            }
+        )
```
0xnm, line 136: there is a better way to deal with such assertion (`assertThat(1).isEqualTo(2)` looks very strange, tbh). You need to use a mock and `argumentCaptor`

```suggestion
        // Given
        whenever(mockDataStoreFile.existsSafe(mockInternalLogger)).thenReturn(false)
        val mockCallback = mock<DataStoreCallback<ByteArray>>()

        // When
        testedDatastoreFileReader.read(
            key = fakeKey,
            deserializer = mockDeserializer,
            callback = mockCallback
        )

        // Then
        argumentCaptor<DataStoreContent<ByteArray>?> {
          verify(mockCallback).onSuccess(capture())
          assertThat(lastValue).isNull
          verifyNoMoreInteractions(mockCallback)
        }
```

Same approach can be applied to the tests below as well.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -0,0 +1,188 @@
...
+    private lateinit var mockFile: File
+
+    @Mock
+    private lateinit var mockFileReaderWriter: FileReaderWriter
+
+    @Mock
+    private lateinit var mockInternalLogger: InternalLogger
+
+    @StringForgery(regex = "^(\\w{3})\$") // a minimal number of chars to avoid flakiness
+    private lateinit var fakeDataString: String
```
0xnm, line 54: what flakiness can we get an why it is 3 characters?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    private lateinit var fakeVersionBytes: ByteArray
+    private lateinit var fakeDataBytes: ByteArray
+    private lateinit var fakeBufferBytes: ByteArray
+
+    @BeforeEach
+    fun setup(@IntForgery(min = 0, max = 10) fakeVersion: Int) {
```
0xnm, line 61: Should it be `Short.MAX_VALUE` here?

```suggestion
    fun setup(@IntForgery(min = 0, max = Short.MAX_VALUE) fakeVersion: Int) {
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockTest.kt
```diff
@@ -0,0 +1,123 @@
...
+        // Then
+        assertThat(block?.size).isEqualTo(fakeString.length + 6)
+        val type = block?.copyOfRange(0, 2)
+        val length = block?.copyOfRange(2, 6)
+        val data = block?.copyOfRange(6, block.size)
+        val typeAsShort = type?.let { ByteBuffer.wrap(it).getShort() }
+        val lengthAsInt = length?.let { ByteBuffer.wrap(it).getInt() }
+
+        assertThat(typeAsShort).isEqualTo(fakeTLVType)
+        assertThat(lengthAsInt).isEqualTo(data?.size)
```
0xnm, line 89: you can use `checkNotNull(block)` here, this will also do a smart cast for the `block` references below, removing the need to use `?` everywhere

```suggestion
        // Then
        checkNotNull(block)
        assertThat(block.size).isEqualTo(fakeString.length + 6)
        val type = block.copyOfRange(0, 2)
        val length = block.copyOfRange(2, 6)
        val data = block.copyOfRange(6, block.size)
        val typeAsShort = ByteBuffer.wrap(type).getShort()
        val lengthAsInt = ByteBuffer.wrap(length).getInt()

        assertThat(typeAsShort).isEqualTo(fakeTLVType)
        assertThat(lengthAsInt).isEqualTo(data.size)
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockTypeTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
```
0xnm, line 25: minor: nothing of this is really used here

```suggestion
```

detekt_custom.yml
```diff
@@ -747,6 +748,7 @@ datadog:
       # endregion
       # region Kotlin Collections
       - "kotlin.Array.all(kotlin.Function1)"
+      - "kotlin.Array.associateBy(kotlin.Function1)"
```
0xnm, line 751: is it used anywhere in this PR?

0xnm (changes requested):

detekt_custom.yml
jonathanmos: It's used in the TLVBlockType in order to associate the short values with the field names

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
jonathanmos: I think it should be `Int.MAX_VALUE`. The `Short` values are for the TLV field identifiers

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
jonathanmos: I can't remember what the reason was and flakiness is not reproducing for me 😕  I'll remove this for now

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DataStoreFileHandler.kt
jonathanmos: On ios there is no such mechanism, but we should have one if there is a usecase. Is clearAllData supposed to clear only data that we would send to the backend, or all sdk data? From the documentation in the interface it implies that it clears only data that would be sent, and at the moment the DS data is not sent

0xnm: It depends on the usecase indeed. I think it is rather not about preventing data to be sent, but just to clear all the data written if customer wants to do it because of the different privacy/legal concerns.

jonathanmos: I've opened a task to do this in a separate pr

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/datastore/DataStoreFileReaderTest.kt
```diff
@@ -0,0 +1,284 @@
...
+
+        // When
+        testedDatastoreFileReader.read(
+            key = fakeKey,
+            deserializer = mockDeserializer,
+            callback = mockCallback
+        )
+
+        // Then
+        argumentCaptor<DataStoreContent<ByteArray>> {
```
0xnm, line 135: shouldn't we specify nullable type here or use `nullableArgumentCaptor`? a bit strange that it compiles/doesn't crash when we specify non-nullable type, but get/check `null`. 

```suggestion
        argumentCaptor<DataStoreContent<ByteArray>?> {
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    @StringForgery
+    private lateinit var fakeDataString: String
+
+    private lateinit var fakeVersionBytes: ByteArray
+    private lateinit var fakeDataBytes: ByteArray
+    private lateinit var fakeBufferBytes: ByteArray
+
+    @BeforeEach
+    fun setup(@IntForgery(min = 0, max = Int.MAX_VALUE) fakeVersion: Int) {
```
0xnm, line 61: I think `max` is already `Int.MAX_VALUE` by default in the `IntForgery`

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -0,0 +1,188 @@
...
+            .startsWith("TLV header corrupt. Invalid type")
+    }
+
+    @Test
+    fun `M return valid object W read() { valid TLV format }`() {
+        // When
+        val tlvArray = testedReader.read(file = mockFile)
+
+        // Then
+        assertThat(tlvArray.size).isEqualTo(2)
```
0xnm, line 118:
```suggestion
        assertThat(tlvArray).hasSize(2)
```

0xnm (approved):

PR #2040, opened 2024-05-20
Update the Method Call metric usage

xgouchet:
### What does this PR do?

Update the MethodCall metric code to allow multiple usage with different operation names + cleanup some SR code

### DIFF
--- a/dd-sdk-android-core/api/apiSurface
+++ b/dd-sdk-android-core/api/apiSurface
@@ -43,6 +43,7 @@ interface com.datadog.android.api.InternalLogger
   fun log(Level, Target, () -> String, Throwable? = null, Boolean = false, Map<String, Any?>? = null)
   fun log(Level, List<Target>, () -> String, Throwable? = null, Boolean = false, Map<String, Any?>? = null)
   fun logMetric(() -> String, Map<String, Any?>)
+  fun startPerformanceMeasure(String, com.datadog.android.core.metrics.TelemetryMetricType, Float, String): com.datadog.android.core.metrics.PerformanceMetric?
   companion object 
     val UNBOUND: InternalLogger
 interface com.datadog.android.api.SdkCore
@@ -104,7 +105,7 @@ interface com.datadog.android.api.feature.FeatureScope
   fun withWriteContext(Boolean = false, (com.datadog.android.api.context.DatadogContext, com.datadog.android.api.storage.EventBatchWriter) -> Unit)
   fun sendEvent(Any)
   fun <T: Feature> unwrap(): T
-  fun startPerformanceMeasure(String, com.datadog.android.core.metrics.TelemetryMetricType, Float): com.datadog.android.core.metrics.PerformanceMetric?
+fun <R: Any?> com.datadog.android.api.InternalLogger.measureMethodCallPerf(Class<*>, String, Float = 100f, () -> R): R
 interface com.datadog.android.api.feature.FeatureSdkCore : com.datadog.android.api.SdkCore
   val internalLogger: com.datadog.android.api.InternalLogger
   fun registerFeature(Feature)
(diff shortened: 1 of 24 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -47,11 +47,13 @@ interface FeatureScope {
      * @param callerClass  name of the class calling the performance measurement.
      * @param metric name of the metric that we want to measure.
      * @param samplingRate value between 0-100 for sampling the event.
+     * @param operationName the name of the operation being measured
      * @return a PerformanceMetric object that can later be used to send telemetry, or null if sampled out
      */
     fun startPerformanceMeasure(
```
0xnm, line 53: side note: I added some comments to the original PR https://github.com/DataDog/dd-sdk-android/pull/1940#pullrequestreview-2037499101, but they were too late. I think in the future we should refactor this a bit a remove this method from the `FeatureScope`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScopeExt.kt
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Measures the execution time for the given block and report it as a MethodCall telemetry metric.
+ * @param R the type of the result of the operation
+ * @param callerClass the class calling the measured method
+ * @param operationName the operationName to report in the metric
+ * @param samplingRate the sampling rate for the metric
+ * @param operation the operation to report
+ */
+fun <R : Any?> FeatureScope?.measureMethodCallPerf(
```
0xnm, line 19:
```suggestion
@InternalApi
fun <R : Any?> FeatureScope?.measureMethodCallPerf(
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScopeExt.kt
```diff
@@ -0,0 +1,42 @@
...
+        val metric = startPerformanceMeasure(
+            callerClass = callerClass.name,
+            metric = TelemetryMetricType.MethodCalled,
+            samplingRate = samplingRate,
+            operationName = operationName
+        )
+
+        val result = operation()
+
+        val isSuccessful = (result != null) && ((result !is Collection<*>) || result.isNotEmpty())
```
0xnm, line 37: why generally speaking empty collection shouldn't be considered as a successful result? but probably for the methods we are going to cover with this call it makes sense.

features/dd-sdk-android-session-replay/build.gradle.kts
```diff
@@ -52,6 +52,7 @@ android {
 
 dependencies {
     api(project(":dd-sdk-android-core"))
+    api(project(":features:dd-sdk-android-trace"))
```
0xnm, line 55: blocking: seems like unnecessary change? I don't see any `com.datadog.android.trace` or `com.datadog.trace` usages in this PR

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/listener/WindowsOnDrawListener.kt
```diff
@@ -31,59 +27,44 @@ internal class WindowsOnDrawListener(
...
-        val performanceMetric = sessionReplayFeature?.startPerformanceMeasure(
-            callerClass = this.javaClass.name,
-            metric = TelemetryMetricType.MethodCalled,
-            samplingRate = methodCallTelemetrySamplingRate
-        )
-
-        val nodes = views
-            .mapNotNull {
+        val nodes = sessionReplayFeature.measureMethodCallPerf(
+            javaClass,
```
0xnm, line 58: should this be in the consumer rules to prevent name obfuscation?

dd-sdk-android-core/api/apiSurface
```diff
@@ -43,6 +43,7 @@ interface com.datadog.android.api.InternalLogger
   fun log(Level, Target, () -> String, Throwable? = null, Boolean = false, Map<String, Any?>? = null)
   fun log(Level, List<Target>, () -> String, Throwable? = null, Boolean = false, Map<String, Any?>? = null)
   fun logMetric(() -> String, Map<String, Any?>)
+  fun startPerformanceMeasure(String, com.datadog.android.core.metrics.TelemetryMetricType, Float, String): com.datadog.android.core.metrics.PerformanceMetric?
```
0xnm, line 46: nice change!

0xnm (approved):

dd-sdk-android-core/api/apiSurface
xgouchet: Yeah It makes more sense, and makes it easier to use too

PR #2041, opened 2024-05-20
RUM-4098: Use the datastore for Session Replay resources

jonathanmos:
### What does this PR do?
Use the datastore from Session Replay, so that resource hashes are persisted between sessions.

### Motivation
The second part of the effort to persist bitmap hashes between sessions. 

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    private fun writeResources() {
+        featureScope?.dataStore?.setValue(
+            data = knownResources,
+            key = DATASTORE_HASHES_CONTENT_FILENAME,
+            serializer = resourceHashesSerializer
+        )
+    }
+
+    private fun createOrOverwriteLastUpdateDateFile(featureScope: FeatureScope?) {
```
xgouchet, line 65: > [!TIP]
> the name of this method seems a bit verbose, but more importantly makes an assumption that the feature data store is stored in a file.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,164 @@
...
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.api.storage.datastore.DataStoreCallback
+import com.datadog.android.core.internal.persistence.Deserializer
+import com.datadog.android.core.persistence.Serializer
+import com.datadog.android.core.persistence.datastore.DataStoreContent
+import com.datadog.android.sessionreplay.internal.ResourcesFeature.Companion.SESSION_REPLAY_RESOURCES_FEATURE_NAME
+import java.util.Collections
+
+@Suppress("TooManyFunctions")
+internal class ResourcesDataStoreManager(
```
xgouchet, line 20: > [!CAUTION] 
> Here we use two independent storages for the actual content and the lastModified information. 
> It could be the source of discrepancies and bugs, and IMO it would be better if the "lastModified" information was handled directly as part of the data store itself, or as part of the data being stored. Case in point, the "last modified" date is only written if : 
> - we fail to read the last modified date (`onFetchFailure` in the `initializeManager`)
> - there is no previous modified date (`onFetchLastUpdateDateSuccess`)
> - when the data is expired (`handleDataStoreExpired`)
> This would also prevent duplicating the read/write calls. and having to synchronize both data in all possible cases.

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
jonathanmos: 👍  you're right. I'll change to more generic names that don't assume file

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
jonathanmos: I originally went with this because that was the ios implementation, but a single file is cleaner - I've modified accordingly

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceHashesEntryDeserializer.kt
```diff
@@ -0,0 +1,32 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.resources
+
+import com.datadog.android.core.internal.persistence.Deserializer
+
+internal class ResourceHashesEntryDeserializer : Deserializer<String, ResourceHashesEntry> {
+    override fun deserialize(model: String): ResourceHashesEntry? {
```
xgouchet, line 12: > [!CAUTION]
> Such a custom made serialization format seems like it won't allow for any future improvements. I think it'd be better to store data as jsons (we can easily create a local json schema and generate the kotlin implementation like we do for all other events).

jonathanmos: I've changed to use JSON serialization

xgouchet: Why did you make the Json serialization by hand instead of using a Json Schema?

jonathanmos: I've fixed it now

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -90,9 +93,20 @@ internal class SessionReplayFeature(
 
         val resourcesFeature = registerResourceFeature(sdkCore)
 
+        val resourcesDataStoreManager = ResourcesDataStoreManager(
+            featureSdkCore = sdkCore,
+            resourceHashesSerializer = ResourceHashesEntrySerializer(),
+            resourcesHashesDeserializer = ResourceHashesEntryDeserializer(internalLogger = sdkCore.internalLogger)
```
0xnm, line 99: for the naming consistency

```suggestion
            resourceHashesDeserializer = ResourceHashesEntryDeserializer(internalLogger = sdkCore.internalLogger)
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/RecordedDataProcessor.kt
```diff
@@ -36,13 +38,20 @@ internal class RecordedDataProcessor(
     override fun processResources(
         item: ResourceRecordedDataQueueItem
     ) {
-        val enrichedResource = EnrichedResource(
-            resource = item.resourceData,
-            applicationId = item.applicationId,
-            filename = item.identifier
-        )
+        val resourceHash = item.identifier
+        val isKnownResource = resourcesDataStoreManager.wasResourcePreviouslySent(resourceHash)
```
0xnm, line 42: it is better to use present tense, here is a proposed naming change:

```suggestion
        val isKnownResource = resourcesDataStoreManager.isPreviouslySentResource(resourceHash)
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceHashesEntrySerializer.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.resources
+
+import com.datadog.android.core.persistence.Serializer
+import com.datadog.android.sessionreplay.model.ResourceHashesEntry
+
+internal class ResourceHashesEntrySerializer : Serializer<ResourceHashesEntry> {
+    override fun serialize(model: ResourceHashesEntry): String {
+        return model.toJson().asJsonObject.toString()
```
0xnm, line 14: why do we need a call to `asJsonObject` here?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.datadog.android.core.persistence.Serializer
+import com.datadog.android.core.persistence.datastore.DataStoreContent
+import com.datadog.android.sessionreplay.internal.ResourcesFeature.Companion.SESSION_REPLAY_RESOURCES_FEATURE_NAME
+import com.datadog.android.sessionreplay.model.ResourceHashesEntry
+import java.util.Collections
+
+internal class ResourcesDataStoreManager(
+    private val featureSdkCore: FeatureSdkCore,
+    private val resourceHashesSerializer: Serializer<ResourceHashesEntry>,
+    private val resourcesHashesDeserializer: Deserializer<String, ResourceHashesEntry>,
```
0xnm, line 23:
```suggestion
    private val resourceHashesDeserializer: Deserializer<String, ResourceHashesEntry>,
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+    private val featureScope: FeatureScope? = featureSdkCore.getFeature(
```
0xnm, line 24: you shouldn't keep it as a property, because feature has its lifecycle.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+        SESSION_REPLAY_RESOURCES_FEATURE_NAME
+    )
+) {
+    private val knownResources = Collections.synchronizedSet(mutableSetOf<String>())
+    private var storedLastUpdateDateNs = System.nanoTime()
+
+    init {
+        initializeManager()
```
0xnm, line 32: why not putting the content of `initializeManager` directly here, in this block?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    internal fun wasResourcePreviouslySent(resourceHash: String): Boolean =
+        knownResources.contains(resourceHash)
+
+    internal fun store(resourceHash: String) {
+        knownResources.add(resourceHash)
+        writeResourcesToStore()
```
0xnm, line 40: I think we are missing atomicity property here. What if hash was added to the `knownResources`, but it wasn't written to the store by some reason? Does it have any impact?

Maybe we can first write to the store, and then, if we know that write operation succeeded, update `knownResources`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    // region internal
+
+    private fun initializeManager() {
+        fetchStoredResourceHashes(
+            onFetchSuccessful = { storedEntry ->
+                val lastUpdateDateNs = storedEntry?.data?.lastUpdateDateNs
```
0xnm, line 48: early null-check will allow you to remove many `.?` in the code below

```suggestion
            onFetchSuccessful = { storedEntry ->
                if (storedEntry == null) return@lambda
                val lastUpdateDateNs = storedEntry.data.lastUpdateDateNs
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+        featureScope?.dataStore?.setValue(
+            data = data,
+            key = DATASTORE_HASHES_ENTRY_NAME,
+            serializer = resourceHashesSerializer
+        )
+    }
+
+    private fun fetchStoredResourceHashes(
+        onFetchSuccessful: (dataStoreContent: DataStoreContent<ResourceHashesEntry>?) -> Unit = {},
+        onFetchFailure: () -> Unit = {}
```
0xnm, line 84: why do we need default values here?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+                override fun onFailure() {
+                    onFetchFailure()
+                }
+            }
+        )
+    }
+
+    private fun handleDataStoreExpired(storedHashes: Set<String>?) {
+        storedHashes?.let { knownResources.removeAll(it) }
+        storedLastUpdateDateNs = System.nanoTime()
```
0xnm, line 103: this property can be modified and read from the different threads, so you need to make it `@Volatile` or use atomic reference.

0xnm (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceHashesEntrySerializer.kt
mariusc83, line 14: shouldn't `toJson().toString()` method be enough ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
```diff
@@ -0,0 +1,120 @@
...
+internal class ResourcesDataStoreManager(
+    private val featureSdkCore: FeatureSdkCore,
+    private val resourceHashesSerializer: Serializer<ResourceHashesEntry>,
+    private val resourcesHashesDeserializer: Deserializer<String, ResourceHashesEntry>,
+    private val featureScope: FeatureScope? = featureSdkCore.getFeature(
+        SESSION_REPLAY_RESOURCES_FEATURE_NAME
+    )
+) {
+    private val knownResources = Collections.synchronizedSet(mutableSetOf<String>())
+    private var storedLastUpdateDateNs = System.nanoTime()
```
mariusc83, line 29: An `AtomicLong` should be used here.

mariusc83 (changes requested):
I think we need to address first the remaining concurrency issue in the `ResourcesDataStoreManager` class

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
jonathanmos: 🤔  do we want to make them dependent? The datastore copy is the persistent copy of the known hashes between sessions, while the knownResources is for the current session. Even if we failed to save to persistent storage, we've still seen this hash for the current session and should avoid sending it

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
jonathanmos: I wasn't sure about this because it's only modified in the initialization of the class and otherwise read, and we're using a single instance

0xnm: it still will be accessed probably by the different threads, so if you want to have thread-safe design of this class, it should be compliant

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourcesDataStoreManager.kt
0xnm: Don't know, I'm just raising the question if we need atomicity here or not.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceHashesEntryDeserializerTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        // Then
+        val messageCaptor = argumentCaptor<() -> String>()
+        verify(mockLogger).log(
+            eq(InternalLogger.Level.ERROR),
+            eq(listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY)),
+            messageCaptor.capture(),
+            isNull(),
+            eq(false),
+            eq(null)
```
0xnm, line 84:
```suggestion
            isNull(),
            eq(false),
            isNull()
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,158 @@
...
+                        callback = onDeleteDatastoreCallback
+                    )
+                } else {
+                    knownResources.addAll(storedHashes)
+                    finishedInitializingManager()
+                }
+            },
+            onFetchFailure = {
+                deleteStoredHashesEntry(onDeleteDatastoreCallback)
+            }
```
0xnm, line 68: I think there is unnecessary complexity here: probably we don't need to do anything for both `onFetchFailure` and `didDataStoreExpire` cases. That is because once new hashes come, we can write them to the store and this entry will override the previous one (which cannot be fetched or outdated). The only case when it won't happen - version difference (by path, I believe?), but I'm not sure it is critical.

This complexity forced to add callbacks for the `removeValue`/`setValue` APIs, which are not normally necessary (and also provide no-op value in `writeResourcesToStore`).

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    private fun finishedInitializingManager() {
+        isInitialized.set(true)
+    }
+
+    private fun writeResourcesToStore() {
+        val data = ResourceHashesEntry(
+            lastUpdateDateNs = storedLastUpdateDateNs,
+            resourceHashes = knownResources.toList()
+        )
```
0xnm, line 93: probably not the most efficient design. Imagine you have 1000 hashes, each is 36 characters (UUID length). All chars are ASCII, so each char is 1 byte. It means that once you add 1001th hash, you need to write (and also serialize) all 1000 hashes you wrote before, which is 36KB. So with each write operation you have data overhead of 36KB, instead of storing only the hash you need to.

mariusc83 (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/datastore/DataStoreHandler.kt
```diff
@@ -23,12 +23,14 @@ interface DataStoreHandler {
...
      * @param version optional version for the entry.
      * If not specified will give the entry version 0 - even if that would be a downgrade from the previous version.
+     * @param callback to indicate whether the operation succeeded or not.
      * @param serializer to use to serialize the data.
      */
     fun <T : Any> setValue(
         key: String,
         data: T,
         version: Int = 0,
+        callback: DataStoreWriteCallback,
```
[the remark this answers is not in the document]
jonathanmos: For when we remove the expired datastore, because this is an async operation and therefore we need to know when it's finished in order to be sure that the initialization of the datastoremanager is complete.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/datastore/DataStoreHandler.kt
```diff
@@ -55,9 +57,11 @@ interface DataStoreHandler {
      * This executes on a worker thread and not on the caller thread.
      *
      * @param key name of the datastore entry
+     * @param callback to indicate whether the operation succeeded or not.
      */
     fun removeValue(
-        key: String
+        key: String,
+        callback: DataStoreWriteCallback
```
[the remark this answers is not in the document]
jonathanmos: For when we remove the expired datastore, because this is an async operation and therefore we need to know when it's finished in order to be sure that the initialization of the datastoremanager is complete.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/datastore/DataStoreHandler.kt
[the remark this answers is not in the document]
jonathanmos: At the moment for `setValue` there is no specific use-case, but I felt that it makes sense to also have a callback here, since we have one for `removeValue`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
jonathanmos: What I'm concerned about is that without having this callback we need to return that the manager has finished initialization when we make the deletion call - but this call is async. So we could have a situation where a call is made to store a resourceHash while this removal of the entry is still in progress

jonathanmos: I didn't notice that you wrote this about onFailure. Yes, you're right - even for a corrupt entry the next write will overwrite it so this is redundant 👍  I'll remove it

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
jonathanmos: as discussed, we'll leave this for now and perhaps improve in a future pr

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,159 @@
...
+                    // no need to do anything here
+                }
+            },
+            serializer = resourceHashesSerializer
+        )
+    }
+
+    private fun fetchStoredResourceHashes(
+        onFetchSuccessful: (dataStoreContent: DataStoreContent<ResourceHashesEntry>?) -> Unit,
+        onFetchFailure: () -> Unit
```
0xnm, line 116: minor: there is no need to have this argument, because the only call site of this method gives empty lambda.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,159 @@
...
+            key = DATASTORE_HASHES_ENTRY_NAME,
+            callback = object : DataStoreWriteCallback {
+                override fun onSuccess() {
+                    // no need to do anything here
+                }
+
+                override fun onFailure() {
+                    // no need to do anything here
+                }
+            },
```
0xnm, line 109: let's maybe make the callback in `setValue` signature as optional or give a no-op value by default? Majority of use-cases won't require providing a callback, same for `removeValue`.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,159 @@
...
+
+                override fun onFailure() {
+                    onFetchFailure()
+                }
+            }
+        )
+    }
+
+    private fun handleDataStoreExpired(storedHashes: Set<String>, callback: DataStoreWriteCallback) {
+        knownResources.removeAll(storedHashes)
```
0xnm, line 136: do you need this line? `handleDataStoreExpired` is called always when `isReady` is `false`, meaning `knownResources` is always empty at this point.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManagerTest.kt
```diff
@@ -0,0 +1,411 @@
...
+            resourceHashesDeserializer = mockResourceHashesEntryDeserializer
+        )
+        val writeThread = Runnable {
+            testedDataStoreManager.cacheResourceHash(fakeHash)
+        }
+
+        val readThread = Runnable {
+            assertThat(testedDataStoreManager.isPreviouslySentResource(fakeHash))
+                .isTrue()
+        }
```
0xnm, line 169: these are not 2 different threads, this is still the same thread. `Runnable` is a simple interface representing unit of work.

I'm not sure that it is easy to write a proper test for the assertion of the thread-safety here, so let's delete this test, in the current state it doesn't correspond to what it wants to do anyway.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Collections
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicLong
+
+internal class ResourceDataStoreManager(
+    private val featureSdkCore: FeatureSdkCore,
+    private val resourceHashesSerializer: Serializer<ResourceHashesEntry>,
+    private val resourceHashesDeserializer: Deserializer<String, ResourceHashesEntry>
+) {
+    private val knownResources = Collections.synchronizedSet(mutableSetOf<String>())
```
0xnm, line 27: note: if performance is critical here, it is better to use `Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>)`. Overall the performance of `ConcurrentHashMap` is better that synchronised map (but there are some differences in behavior).

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManagerTest.kt
```diff
@@ -0,0 +1,411 @@
...
+        whenever(
+            mockDataStoreHandler.value(
+                key = eq(DATASTORE_HASHES_ENTRY_NAME),
+                version = anyOrNull(),
+                callback = any<DataStoreReadCallback<ResourceHashesEntry>>(),
+                deserializer = any()
+            )
+        ) doAnswer {
+            @Suppress("UNCHECKED_CAST")
+            val callback = it.arguments[2] as DataStoreReadCallback<ResourceHashesEntry>
```
0xnm, line 223: variant without `Suppress` (same can be applied to the similar lines below)

```suggestion
            val callback = it.getArgument<DataStoreReadCallback<ResourceHashesEntry>>(2)
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
jonathanmos: This lambda should actually not be empty but should have a call to `finishedInitializingManager`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt
jonathanmos: Good point - I've modified this

0xnm (approved):
lgtm. but my concern about writing all the hashes each time we just need to add one more to storage stays, not sure how much it can become a bottleneck in the app with many resources.

xgouchet (approved):

PR #2044, opened 2024-05-21
RUM-4516 flag critical events in custom persistence

xgouchet:
### What does this PR do?

Adds an additional flag when writing event to let storage handle those differently. 

### Motivation

Because some users with custom PersistenceStrategy use mainly a memory based storage, they need to know when the app is about to crash in order to persist the data in a way that can be retrieved after an app reboot. 

### Additional Notes

An additional type was created for telemetry, just to be able to distinguish it.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/DataWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/DataWriter.kt
@@ -18,8 +18,12 @@ interface DataWriter<T> {
     /**
      * Writes the element with a given [EventBatchWriter].
      *
+     * @param writer the writer to use
+     * @param element the event to write
+     * @param eventType additional info about the event
+     *
      * @return true if element was written, false otherwise.
      */
     @WorkerThread
-    fun write(writer: EventBatchWriter, element: T): Boolean
+    fun write(writer: EventBatchWriter, element: T, eventType: EventType): Boolean
 }
--- a/dd-sdk-android-core/api/apiSurface
+++ b/dd-sdk-android-core/api/apiSurface
@@ -135,10 +135,14 @@ interface com.datadog.android.api.net.RequestFactory
     const val QUERY_PARAM_SOURCE: String
     const val QUERY_PARAM_TAGS: String
 interface com.datadog.android.api.storage.DataWriter<T>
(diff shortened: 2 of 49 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/DataWriter.kt
```diff
@@ -18,8 +18,12 @@ interface DataWriter<T> {
...
      *
+     * @param writer the writer to use
+     * @param element the event to write
+     * @param eventType additional info about the event
+     *
      * @return true if element was written, false otherwise.
      */
     @WorkerThread
-    fun write(writer: EventBatchWriter, element: T): Boolean
+    fun write(writer: EventBatchWriter, element: T, eventType: EventType): Boolean
```
0xnm, line 28: what do you think about adding a default value here `eventType: EventType = EventType.DEFAULT` in order to reduce the change layout and make this change backward compatible (just in case)?

0xnm (commented):
Overall the change looks clear to me, I left one question.

Although maybe alternative to this would be just adding something like `PersistenceStrategy#flush` method which could be called when crash happens, this could reduce change layout.

Or, maybe, we could say that if custom persistence strategy is used, then user is responsible for it, meaning they should register their own unhandled exception handler (but, probably, this is not a great integration experience).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/DataWriter.kt
xgouchet: Well that's a thought but I'm trying to limit the use of default values when it's not necessary. 
IMHO, using default values makes sense for public APIs our customer might use, but for our internals, it's a source of errors since we could get lazy and not set a value we're supposed to set.

xgouchet:
@0xnm regarding your idea to provide a `flush` method, it does make a sense, but IMO makes an assumption on what the implementation looks like (in this case that the custom implementation will always be an in memory). 
If a customer wanted to use a local DB instead of batch files, the notion of flushing data would make no sense. Flagging the events themselves as Crash means they can add their own logic on top of it, e.g.  send them first when we request data to be sent in the next intake request.

0xnm (approved):

jonathanmos (approved):

PR #2048, opened 2024-05-22
RUM-4561 Delegate Drawable copy to background thread

xgouchet:
### What does this PR do?

Our previous mechanism of copying a drawable's content was performed on the main thread. 
When starting an app, this would create a lot of main thread task to run. And any new image would also take some of the main thread time, creating a lot of jank frames in any non hello world app. 

This will fix the performance issue some of our customers started to notice when enabling SR.

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
@@ -9,6 +9,7 @@ package com.datadog.android.sessionreplay.internal.recorder
 import android.view.View
 import android.view.ViewGroup
 import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.feature.measureMethodCallPerf
 import com.datadog.android.sessionreplay.MapperTypeWrapper
 import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueRefs
 import com.datadog.android.sessionreplay.internal.recorder.mapper.QueueStatusCallback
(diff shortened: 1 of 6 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
```diff
@@ -76,6 +81,14 @@ internal class TreeViewTraversal(
             )
         }
 
+        val resolvedWireframes = internalLogger.measureMethodCallPerf(
+            javaClass,
```
jonathanmos, line 85: Similar to what was done with WindowsOnDrawListener, do we need to also add this class to proguard rules to prevent classname obfuscation?

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
xgouchet: Makes sense indeeed

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
```diff
@@ -59,8 +58,8 @@ internal class DrawableUtils(
             resizeBitmapCallback = object :
                 ResizeBitmapCallback {
                 override fun onSuccess(bitmap: Bitmap) {
-                    mainThreadHandler.post {
-                        @Suppress("ThreadSafety") // this runs on the main thread
+                    executorService.submitSafe("drawOnCanvas", internalLogger) {
```
0xnm, line 61: is it safe to operate with bitmap/canvas from the worker thread on all Android versions?

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/DrawableUtils.kt
xgouchet: From the tests I've done, it seems so. I'll do more test to ensure it works on all levels.

xgouchet: It does work on all versions 👍

PR #2051, opened 2024-05-23
RUM-4260 Make CoreTracer code Java 7 compatible

mariusc83:
### What does this PR do?

In this PR we are cleaning up the imported Java code for the `CoreTracer` and make it Java 7 compatible in order to not require desugaring on Android 23 and below.

The are several steps in achieving this:

- we are adding back the Injectors into the `CoreTracer` as I realized that we will be needing this later for Context propagation
- we are replacing the logic where `java.util.Base64`, `java.util.Map`, `java.util.Stream` is being used in order to make it Java 7 compatible
- for the places where we cannot delete or replace would be too complex we are introducing our own package with substitutes for this lambda functions
- we are performing more cleaning after and re - enabling the NewApi rule for linting

Have in mind that maybe some more cleaning could be performed but I don't find this idea wise. What I could clean so far was code that I was really certain that it cannot affect the current functionality, it will not have any impact on the APM metrics and will not have repercussions later. We already went into trouble by removing the `Injector` which seemed not to be used for the moment and I just realized during this PR work that this will have a negative impact later.

### Motivation

### Additional Notes

I performed a Profiling by analyzing the CPU consumption and Memory consumption while using the new Otel code and I could not see any potential issue (spike).


features/dd-sdk-android-trace/src/main/java/java8/util/function/BiConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+/*
+ * Copyright (c) 2012, 2013, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
```
0xnm, line 11: important: I think we may hit an issue here since it is GPLv2 license. So if we copy and modify source code from Oracle JDK that is under GPLv2, we must release that modified code under GPLv2. But to be checked.

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/metrics/SpanMetricRegistryImpl.java
```diff
@@ -1,48 +1,52 @@
...
+        return INSTANCE;
+    }
+
+    private SpanMetricRegistryImpl() {
+        this.spanMetrics = new ConcurrentHashMap<>();
+    }
+
+    @Override
+    public SpanMetrics get(String instrumentationName) {
+        return MapUtils.computeIfAbsent(instrumentationName, spanMetrics, SpanMetricsImpl::new);
```
0xnm, line 26: such form of constructor invocation `SpanMetricsImpl::new` is still compatible with Java 7?

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/propagation/XRayHttpCodec.java
```diff
@@ -22,246 +22,251 @@
...
-      }
-    }
-  }
+            for (Map.Entry<String, String> entry : context.baggageItems()) {
+                final String header;
+                if (invertedBaggageMapping.containsKey(entry.getKey())) {
+                    header = invertedBaggageMapping.get(entry.getKey());
+                } else {
+                    header = entry.getKey();
+                }
```
0xnm, line 112: can be probably a bit simpler

```suggestion
                String header = invertedBaggageMapping.get(entry.getKey());
                if (header == null) {
                    header = entry.getKey();
                }
```

0xnm (commented):
lgtm, but my only concern is licensing after we copy code labeled as GPLv2, let's check this point.

mariusc83:
> lgtm, but my only concern is licensing after we copy code labeled as GPLv2, let's check this point.

Yes I wanted to open the subject in the daily as I am not 100% sure also. I think we are safe but we need to double check.

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/metrics/SpanMetricRegistryImpl.java
mariusc83: lint doesn't say anything...I will double check

features/dd-sdk-android-trace/src/main/java/java8/util/function/BiConsumer.java
mariusc83: Oh...I copy pasted these files so I assumed that it will not add our licence automatically, need to remove that part. Also I wanted to double check in the daily if for these files we are good if we copy paste them together with the License header also. @xgouchet any hints ?

0xnm: if it is fine to **copy** these files, this copyright should be kept

mariusc83: yes but I don't think we should keep ours that was automatically added in the header of the file. I think we should only keep theirs as this code does not belong us. I removed our header but will still need to discuss it further to make sure we are 100% safe.

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/metrics/SpanMetricRegistryImpl.java
mariusc83: it seems so...compiler and lint is not complaining at all it is actually suggesting me to use that in case I use the function implementation itself.

xgouchet (approved):

PR #2056, opened 2024-05-28
RUM-1462 Fix thread safety warnings

xgouchet:
### What does this PR do?

A rule we used for a while now is using the Thread annotation (e.g.: `@UIThread`, `@WorkerThread`, …) to ensure operations that should happen in a specific context didn't cross thread expectations. 

Unfortunately we had a large number of `@Suppress` annotations and thread  boundaries crossings in our code  base. 

This PR provides two things :
- enhance the rule to detekt some known thread switching (e.g. using executors) to avoid the need of legitimate `@Suppress` annotations
- fix places where needed

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
@@ -6,6 +6,7 @@
 
 package com.datadog.android.api.feature
 
+import androidx.annotation.AnyThread
 import com.datadog.android.api.context.DatadogContext
 import com.datadog.android.api.storage.EventBatchWriter
 
@@ -13,6 +14,7 @@ import com.datadog.android.api.storage.EventBatchWriter
  * Represents a Datadog feature.
  */
 interface FeatureScope {
+
     /**
      * Utility to write an event, asynchronously.
      * @param forceNewBatch if `true` forces the [EventBatchWriter] to write in a new file and
@@ -22,6 +24,7 @@ interface FeatureScope {
      * [DatadogContext] will have a state created at the moment this method is called, before the
      * thread switch for the callback invocation.
      */
+    @AnyThread
     fun withWriteContext(
         forceNewBatch: Boolean = false,
         callback: (DatadogContext, EventBatchWriter) -> Unit
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
@@ -11,6 +11,7 @@ import android.content.Context
 import android.content.pm.ApplicationInfo
 import android.os.Build
 import android.util.Log
+import androidx.annotation.AnyThread
 import androidx.annotation.WorkerThread
 import com.datadog.android.Datadog
 import com.datadog.android.api.InternalLogger
@@ -35,6 +36,7 @@ import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
 import com.datadog.android.core.internal.time.DefaultAppStartTimeProvider
 import com.datadog.android.core.internal.utils.scheduleSafe
+import com.datadog.android.core.internal.utils.submitSafe
 import com.datadog.android.core.thread.FlushableExecutorService
 import com.datadog.android.error.internal.CrashReportsFeature
 import com.datadog.android.ndk.internal.NdkCrashHandler
@@ -151,11 +153,13 @@ internal class DatadogCore(
     }
 
     /** @inheritDoc */
+    @AnyThread
     override fun setTrackingConsent(consent: TrackingConsent) {
         coreFeature.trackingConsentProvider.setConsent(consent)
     }
 
     /** @inheritDoc */
+    @AnyThread
     override fun setUserInfo(
         id: String?,
         name: String?,
@@ -173,20 +177,21 @@ internal class DatadogCore(
     }
 
     /** @inheritDoc */
+    @AnyThread
     override fun addUserProperties(extraInfo: Map<String, Any?>) {
         coreFeature.userInfoProvider.addUserProperties(extraInfo)
     }
 
     /** @inheritDoc */
+    @AnyThread
     override fun clearAllData() {
         features.values.forEach {
             it.clearAllData()
         }
-        // TODO RUM-1462 address Thread safety
-        @Suppress("ThreadSafety") // removal of the data is done in synchronous manner
-        coreFeature.deleteLastViewEvent()
-        @Suppress("ThreadSafety") // removal of the data is done in synchronous manner
-        coreFeature.deleteLastFatalAnrSent()
+        getPersistenceExecutorService().submitSafe("Clear all data", internalLogger) {
+            coreFeature.deleteLastViewEvent()
+            coreFeature.deleteLastFatalAnrSent()
+        }
     }
 
     /** @inheritDoc */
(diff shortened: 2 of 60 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -173,20 +177,21 @@ internal class DatadogCore(
...
-        // TODO RUM-1462 address Thread safety
-        @Suppress("ThreadSafety") // removal of the data is done in synchronous manner
-        coreFeature.deleteLastViewEvent()
-        @Suppress("ThreadSafety") // removal of the data is done in synchronous manner
-        coreFeature.deleteLastFatalAnrSent()
+        getPersistenceExecutorService().submitSafe("Clear all data", internalLogger) {
+            coreFeature.deleteLastViewEvent()
+            coreFeature.deleteLastFatalAnrSent()
+        }
     }
```
0xnm, line 195: I think it is done in sync way by design, so that the call on the caller side is blocking (even though it may happen on the main thread) and the same ideology as with `features.values.forEach { it.clearAllData() }` above. Is it blocking in iOS SDK?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -173,20 +177,21 @@ internal class DatadogCore(
...
     override fun clearAllData() {
         features.values.forEach {
             it.clearAllData()
         }
-        // TODO RUM-1462 address Thread safety
-        @Suppress("ThreadSafety") // removal of the data is done in synchronous manner
-        coreFeature.deleteLastViewEvent()
-        @Suppress("ThreadSafety") // removal of the data is done in synchronous manner
-        coreFeature.deleteLastFatalAnrSent()
+        getPersistenceExecutorService().submitSafe("Clear all data", internalLogger) {
```
mariusc83, line 191: isn't this going to create issues if we expect this method to return in the same stack immediately ?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
```diff
@@ -124,19 +125,20 @@ internal class ConsentAwareStorage(
     }
 
     /** @inheritdoc */
-    @WorkerThread
+    @AnyThread
     override fun dropAll() {
-        synchronized(lockedBatches) {
-            lockedBatches.forEach {
-                deleteBatch(it, RemovalReason.Flushed)
+        executorService.submitSafe("ConsentAwareStorage.dropAll", internalLogger) {
```
mariusc83, line 130: same question as above ? What if we call this from an unit test `tearDown` or `setup` method ?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -728,6 +729,13 @@ internal class DatadogCoreTest {
         )
         val mockCoreFeature = mock<CoreFeature>()
         testedCore.coreFeature = mockCoreFeature
+        val mockExecutorService: FlushableExecutorService = mock()
+        whenever(mockCoreFeature.persistenceExecutorService) doReturn mockExecutorService
+        whenever(mockExecutorService.submit(any())) doAnswer {
```
mariusc83, line 734: I see so this is because we've made those methods async ?

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/ThreadSafety.kt
```diff
@@ -59,7 +69,7 @@ class ThreadSafety : Rule() {
         }
     }
 
-    private var parentFunGroup: ThreadGroup = ThreadGroup.UNKNOWN
+    private val parentFunGroupStack: MutableList<ThreadGroup> = mutableListOf(ThreadGroup.UNKNOWN)
```
0xnm, line 72: minor: since we are treating it like a stack, we can use `Stack` type here (or `Deque` -> `LinkedList`) which will allows us simply use `push`/`pop`/`peek` methods instead of `add(index, ...)` and `removeAt(index, ...)`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -637,14 +637,16 @@ internal class DatadogRumMonitor(
                 rootScope.handleEvent(event, writer)
             }
         } else if (event is RumRawEvent.SendTelemetry) {
-            @Suppress("ThreadSafety") // TODO RUM-3756 delegate to another thread
-            telemetryEventHandler.handleEvent(event, writer)
+            // If an issue arise, use the unbound logger to prevent a cycle that would lead
+            // to a stack overflow
+            executorService.submitSafe("Telemetry event handling", InternalLogger.UNBOUND) {
```
0xnm, line 642: does it make sense to create a dedicated thread for the telemetry processing? RUM monitor thread may already be busy and we don't need processing thread to be especially this one.

on another hand, even though `TelemetryEventHandler.handleEvent` is annotated with `@WorkerThread`, in reality this annotation may be redundant, because the amount of work done there before the switch to `withWriteContext` is very small.

mariusc83 (approved):

0xnm (commented):
lgtm, but I'm curious about the behaviour of `Datadog.clearAllData` in iOS SDK.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -22,6 +24,7 @@ interface FeatureScope {
      * [DatadogContext] will have a state created at the moment this method is called, before the
      * thread switch for the callback invocation.
      */
+    @AnyThread
```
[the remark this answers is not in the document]
xgouchet: We could, I mostly added it to make it explicit while checking all thread safety warnings

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
xgouchet: There's no such expectations, and it would mostly create issues if called from the main thread on a running app.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
xgouchet: Just checked, on iOS the call is async

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
xgouchet: > [!IMPORTANT]
> Honestly, if we need to call this from a test, we can adapt the test to this call being async. We should not make things synchronous at runtime because of testing constraints.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
xgouchet: Yep

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
xgouchet: That's a good point, I'll change that.

0xnm (approved):

PR #2062, opened 2024-05-29
Remove the dd-sketches dependency and related logic

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


features/dd-sdk-android-trace-otel/consumer-rules.pro
```diff
@@ -1 +1,2 @@
--keepnames class java.util.function.Function { *; }
\ No newline at end of file
+-keepnames class java.util.function.Function { *; }
+-dontwarn com.google.auto.**
```
mariusc83, line 2: This was added in this PR just for more speed. It is required by the `OpenTelemetry` API dependency which complains in R8

0xnm: > PR just for more speed

I'm not clear how does it speed up the things. I don't think we should add this line anyway, `com.google.auto.*` may exist in customer dependencies.

mariusc83: I wonder if we should do anything about it here or just leave this on the customer side and mention in the docs ? In any case without any action from our side R8 fails when compiling the APK

0xnm (approved):

PR #2075, opened 2024-06-07
RUM-3458: Fix time drift in RecordedDataQueueHandler

jonathanmos:
### What does this PR do?
Fixes an issue in the RecordedDataQueueHandler where the "current" time was being passed in to the loop, but never updated inside it. This could lead to the time drifting when the queue was very long, leading to items being considered out of date.

### Motivation
We have telemetry indicating numerous DataQueueItems are being dropped for being out of date. 

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
@@ -13,74 +13,24 @@ import com.datadog.android.api.InternalLogger
 import com.datadog.android.core.internal.utils.executeSafe
 import com.datadog.android.sessionreplay.internal.processor.RecordedDataProcessor
 import com.datadog.android.sessionreplay.internal.processor.RumContextDataHandler
-import com.datadog.android.sessionreplay.internal.utils.TimeProvider
 import com.datadog.android.sessionreplay.model.MobileSegment
 import com.datadog.android.sessionreplay.recorder.SystemInformation
 import java.util.Locale
 import java.util.Queue
-import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.ExecutorService
-import java.util.concurrent.LinkedBlockingDeque
-import java.util.concurrent.ThreadPoolExecutor
-import java.util.concurrent.TimeUnit
 
 /**
  * Responsible for storing the Snapshot and Interaction events in a queue.
  * Allows for asynchronous enrichment, which still preserving the event order.
  * The items are added to the queue from the main thread and processed on a background thread.
  */
-@Suppress("TooManyFunctions")
-internal class RecordedDataQueueHandler : DataQueueHandler {
-    private var executorService: ExecutorService
-    private var processor: RecordedDataProcessor
-    private var rumContextDataHandler: RumContextDataHandler
-    private var timeProvider: TimeProvider
-    private val internalLogger: InternalLogger
+internal class RecordedDataQueueHandler(
+    private var processor: RecordedDataProcessor,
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueHandler.kt
```diff
@@ -170,8 +124,14 @@ internal class RecordedDataQueueHandler : DataQueueHandler {
...
     @Synchronized
-    private fun triggerProcessingLoop(currentTime: Long) {
+    private fun triggerProcessingLoop(initialTime: Long) {
+        var currentTime = initialTime
+
         while (recordedDataQueue.isNotEmpty()) {
+            mainThreadHandler.post {
+                currentTime = timeProvider.getDeviceTimestamp()
+            }
+
```
0xnm, line 134: I'm not sure it is a right solution. Maybe it can reduce the problem somehow, but the issue here is concurrency: `mainThreadHandler.post` is non-blocking call, so by the time `currentTime = timeProvider.getDeviceTimestamp()` is executed, you can be already advance in the `while` loop past `post` line, or even be on the next turn.

jonathanmos: I've changed the implementation to have a separate property to store `System.nanoTime` for each item and query that, so it will be monotonic and we're no longer dependent upon calling on a certain thread

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
```diff
@@ -104,8 +109,23 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
         this.recordedDataQueueHandler = RecordedDataQueueHandler(
             processor = processor,
             rumContextDataHandler = rumContextDataHandler,
-            timeProvider = timeProvider,
-            internalLogger = internalLogger
+            internalLogger = internalLogger,
+
+            /**
+             * TODO RUMM-0000 consider change to LoggingThreadPoolExecutor once V2 is merged.
```
0xnm, line 115: I think it is time to do it: it is a simple change and can be made in another PR

0xnm (approved):
lgtm. Since we care only about item age at this point, using 2 `System.nanoTime()` instants is fine and result of this call is _monotonically_ increasing.

However, there is a second timestamp involved, which is attached to the queue item and then used later to build wireframes (`nextItem.recordedQueuedItemContext.timestamp`), so it is better to have another pair of eyes for the review here, because I'm don't have a deep knowledge of the whole processing pipeline.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
xgouchet, line 115: > [!CAUTION]
> Do not use a "RUM-000" ticket, actually create one and reference it here.

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
jonathanmos: 👍  I've opened a ticket and modified this to reference

xgouchet (approved):

PR #2081, opened 2024-06-12
RUM-4894 Fix the CoreTracer flaky tests

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
+++ b/reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
@@ -35,6 +35,7 @@ import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.data.Offset
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.RepeatedTest
+import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
 import org.mockito.junit.jupiter.MockitoExtension
@@ -768,7 +769,7 @@ internal class OtelTracerProviderTest {
             }
     }
 
-    @RepeatedTest(10)
+    @Test
     fun `M use user-keep or user-drop priority W buildSpan { tracer was provided a sample rate }`(
         @StringForgery fakeInstrumentationName: String,
         @DoubleForgery(min = 0.0, max = 100.0) sampleRate: Double,
@@ -778,14 +779,18 @@ internal class OtelTracerProviderTest {
         val testedProvider = OtelTracerProvider.Builder(stubSdkCore).setSampleRate(sampleRate).build()
         val tracer = testedProvider.tracerBuilder(fakeInstrumentationName).build()
         val blockingWriterWrapper = tracer.useBlockingWriter()
-        val numberOfSpans = 100
+        val numberOfSpans = 300
         val normalizedSampleRate = sampleRate / 100.0
         val expectedKeptSpans = (numberOfSpans * normalizedSampleRate).toInt()
         val expectedDroppedSpans = numberOfSpans - expectedKeptSpans
 
         // When
         repeat(numberOfSpans) {
-            tracer.spanBuilder(forge.anAlphabeticalString()).startSpan().end()
+            val span = tracer.spanBuilder(forge.anAlphabeticalString()).startSpan()
+            // there is a throttle on the sampler which drops all the spans over the 100 limit in 1 second
+            // so we need to sleep a bit to make sure the spans are not dropped because of throttling
+            Thread.sleep(10)
+            span.end()
         }
 
         // Then
@@ -803,7 +808,10 @@ internal class OtelTracerProviderTest {
         val keptSpans = spansWritten.filter {
             it.getInt(SAMPLING_PRIORITY_KEY) == PrioritySampling.USER_KEEP.toInt()
         }
-        val offset = 10
+        // Because of the way sampling works the deviation can be quite high so we will have to use an offset of 15%
+        // here to make sure this test will never be flaky
+        val offset = numberOfSpans * 15 / 100
+        assertThat(droppedSpans.size + keptSpans.size).isEqualTo(numberOfSpans)
         assertThat(droppedSpans.size).isCloseTo(expectedDroppedSpans, Offset.offset(offset))
         assertThat(keptSpans.size).isCloseTo(expectedKeptSpans, Offset.offset(offset))
     }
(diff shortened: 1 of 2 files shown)


reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
```diff
@@ -803,7 +803,7 @@ internal class OtelTracerProviderTest {
         val keptSpans = spansWritten.filter {
             it.getInt(SAMPLING_PRIORITY_KEY) == PrioritySampling.USER_KEEP.toInt()
         }
-        val offset = 10
+        val offset = 20
```
0xnm, line 806: I think the test will be still flaky, even with this change. And 20% is quite big offset, maybe it makes sense only a small absolute values. And if you have actual say 1, and expected as 2, this won't help (because diff will be 50% for such numbers).

0xnm (approved):

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
mariusc83: I tested with 1000 repeated tests and never failed. I think it is fine but also my question is regarding the offset as it is too high. Should we still keep this test ?

0xnm: I don't have a definitive answer here. I think we have similar tests checking sampling rate in other modules, and they probably don't have a high offset? Maybe we can modify the test somehow?

0xnm (approved):

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
mariusc83: yes...will try to have a look what else I can do

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
```diff
@@ -803,7 +805,10 @@ internal class OtelTracerProviderTest {
         val keptSpans = spansWritten.filter {
             it.getInt(SAMPLING_PRIORITY_KEY) == PrioritySampling.USER_KEEP.toInt()
         }
-        val offset = 10
+        // Because of the way sampling works the deviation can be quite high so we will have to use an offset of 20
+        // here to make sure this test will never be flaky
+        val offset = 20
```
xgouchet, line 810: > [!NOTE]
> The offset should be based off of the `numberOfSpans` and not hardcoded, e.g.: 
```kotlin
    val offset = (numberOfSpans * 15) / 100
```

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
xgouchet: > [!NOTE]
> One way to limit flakyness is to increase the `numberOfSpans`

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
```diff
@@ -768,14 +769,14 @@ internal class OtelTracerProviderTest {
             }
     }
 
-    @RepeatedTest(10)
+    @Test
     fun `M use user-keep or user-drop priority W buildSpan { tracer was provided a sample rate }`(
         @StringForgery fakeInstrumentationName: String,
-        @DoubleForgery(min = 0.0, max = 100.0) sampleRate: Double,
+        @IntForgery(min = 0, max = 100) sampleRate: Int,
```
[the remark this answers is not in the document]
mariusc83: yes good catch was a left over during my tests to see why it is flaky

0xnm (changes requested):

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
```diff
@@ -778,14 +779,18 @@ internal class OtelTracerProviderTest {
...
         val expectedKeptSpans = (numberOfSpans * normalizedSampleRate).toInt()
         val expectedDroppedSpans = numberOfSpans - expectedKeptSpans
 
         // When
         repeat(numberOfSpans) {
-            tracer.spanBuilder(forge.anAlphabeticalString()).startSpan().end()
+            val span = tracer.spanBuilder(forge.anAlphabeticalString()).startSpan()
+            // there is a throttle on the sampler which drops all the spans over the 100 limit in 1 second
+            // so we need to sleep a bit to make sure the spans are not dropped because of throttling
+            Thread.sleep(100)
```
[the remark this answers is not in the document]
mariusc83: yes I wanted to make sure but indeed we can sleep for 10ms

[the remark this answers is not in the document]
0xnm: still ideally we won't have to use `sleep`, but 3s is much better than 30s.

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
```diff
@@ -803,7 +808,10 @@ internal class OtelTracerProviderTest {
         val keptSpans = spansWritten.filter {
             it.getInt(SAMPLING_PRIORITY_KEY) == PrioritySampling.USER_KEEP.toInt()
         }
-        val offset = 10
+        // Because of the way sampling works the deviation can be quite high so we will have to use an offset of 15%
+        // here to make sure this test will never be flaky
+        val offset = numberOfSpans * 15 / 100
```
0xnm, line 813: nit: I think here we should use float types, otherwise we are loosing a lot by doing an integer division.

```suggestion
        val offset = numberOfSpans * 15f / 100
```

0xnm (approved):

reliability/single-fit/trace/src/test/kotlin/com/datadog/android/trace/integration/otel/OtelTracerProviderTest.kt
mariusc83: the offset needs to be an int in the end so I will keep it like that.

PR #2083, opened 2024-06-13
RUM-1662: Create SessionEndedMetric and SessionEndedMetricDispatcher with unit test

ambushwork:
### What does this PR do?

* [ADD] Add `SessionEndedMetric` class to record the 
* [ADD] SessionEndedMetricDispatcher to coordinates multiple `SessionEndedMetric` and send telemetry when session is ended
  
### Motivation

Ticket: https://datadoghq.atlassian.net/browse/RUM-1662

### Additional Notes

This is the first step of this task, only implementation of two classes with unit test assuring that the data structure and json format matches the specification. In next PR, these two classes will be used inside `RumSessionScope` or `RumApplicatioScope` to track the session.

### DIFF
--- a/detekt_custom.yml
+++ b/detekt_custom.yml
@@ -201,7 +201,7 @@ datadog:
       - "java.util.concurrent.Callable.call():java.lang.Exception"
       - "java.util.concurrent.ConcurrentHashMap.contains(kotlin.Any?):java.lang.NullPointerException"
       - "java.util.concurrent.ConcurrentHashMap.remove(com.datadog.android.api.feature.FeatureContextUpdateReceiver):java.lang.NullPointerException"
-      - "java.util.concurrent.ConcurrentHashMap.remove(kotlin.String):java.lang.NullPointerException"
+      - "java.util.concurrent.ConcurrentHashMap.remove(kotlin.String?):java.lang.NullPointerException"
       - "java.util.concurrent.ConcurrentLinkedQueue.offer(com.datadog.android.sessionreplay.internal.async.RecordedDataQueueItem):java.lang.NullPointerException"
       - "java.util.concurrent.CopyOnWriteArraySet.removeAll(kotlin.collections.Collection):java.lang.NullPointerException,java.lang.ClassCastException"
       - "java.util.concurrent.CountDownLatch.await(kotlin.Long, java.util.concurrent.TimeUnit?):java.lang.InterruptedException"
@@ -262,7 +262,6 @@ datadog:
       - "kotlin.ByteArray.copyOf(kotlin.Int):java.lang.NegativeArraySizeException"
       - "kotlin.ByteArray.copyOfRange(kotlin.Int, kotlin.Int):java.lang.IndexOutOfBoundsException,java.lang.IllegalArgumentException"
       - "kotlin.ByteArray.get(kotlin.Int):java.lang.IndexOutOfBoundsException"
-      - "kotlin.collections.MutableSet.first():java.util.NoSuchElementException"
       - "kotlin.Double.roundToLong():java.lang.IllegalArgumentException"
       # endregion
       # region Kotlin Collections
@@ -274,6 +273,8 @@ datadog:
       - "kotlin.collections.List.get(kotlin.Int):java.util.NoSuchElementException"
       - "kotlin.collections.MutableIterator.next():java.util.NoSuchElementException"
       - "kotlin.collections.MutableIterator.remove():java.lang.UnsupportedOperationException,java.lang.IllegalStateException"
+      - "kotlin.collections.MutableSet.first():java.util.NoSuchElementException"
+      - "kotlin.collections.MutableSet.last():java.util.NoSuchElementException"
       # endregion
       # region Kotlin Coroutines
       - "kotlinx.coroutines.Deferred.await():java.util.concurrent.CancellationException"
@@ -950,6 +951,8 @@ datadog:
       - "kotlin.collections.MutableSet.remove(com.datadog.android.core.internal.persistence.ConsentAwareStorage.Batch)"
       - "kotlin.collections.MutableSet.remove(java.io.File)"
       - "kotlin.collections.MutableSet.toList()"
+      - "kotlin.collections.MutableSet.lastOrNull()"
+      - "kotlin.collections.MutableSet.sortedByDescending(kotlin.Function1)"
       - "kotlin.collections.Set.any(kotlin.Function1)"
       - "kotlin.collections.Set.associate(kotlin.Function1)"
       - "kotlin.collections.Set.contains(com.datadog.android.trace.TracingHeaderType)"
@@ -991,6 +994,11 @@ datadog:
       - "kotlin.collections.setOf(kotlin.Array)"
       - "kotlin.collections.setOf(kotlin.Int)"
       - "kotlin.collections.setOf(kotlin.String)"
+      - "kotlin.collections.linkedMapOf()"
+      - "kotlin.collections.MutableCollection.count(kotlin.Function1)"
+      - "kotlin.collections.MutableCollection.sumOf(kotlin.Function1)"
+      - "kotlin.collections.MutableCollection.sum()"
+      - "kotlin.collections.List.sortedByDescending(kotlin.Function1)"
       - "kotlin.emptyArray()"
       - "kotlin.sequences.Sequence.groupBy(kotlin.Function1)"
       - "kotlin.sequences.Sequence.filter(kotlin.Function1)"
@@ -1006,6 +1014,7 @@ datadog:
       - "kotlin.Any.hashCode()"
       - "kotlin.Any.toString()"
       - "kotlin.Boolean.hashCode()"
+      - "kotlin.Boolean.not()"
       - "kotlin.Byte.toInt()"
       - "kotlin.ByteArray.constructor(kotlin.Int)"
       - "kotlin.Char.isLowerCase()"
@@ -1028,6 +1037,7 @@ datadog:
       - "kotlin.Int.toDouble()"
       - "kotlin.Int.toFloat()"
       - "kotlin.Int.toLong()"
+      - "kotlin.Int.coerceAtMost(kotlin.Int)"
       - "kotlin.IntArray.constructor(kotlin.Int)"
       - "kotlin.IntArray.joinToString(kotlin.CharSequence, kotlin.CharSequence, kotlin.CharSequence, kotlin.Int, kotlin.CharSequence, kotlin.Function1?)"
       - "kotlin.Long.coerceIn(kotlin.Long, kotlin.Long)"
@@ -1048,6 +1058,7 @@ datadog:
       - "kotlin.String.trim(kotlin.Function1)"
       # endregion
       # region Kotlin Tuples
+      - "kotlin.Pair.constructor(kotlin.String, kotlin.Int)"
       - "kotlin.Pair.constructor(com.datadog.android.sessionreplay.internal.utils.SessionReplayRumContext, com.google.gson.JsonArray)"
       - "kotlin.Pair.constructor(com.datadog.android.sessionreplay.model.MobileSegment, com.google.gson.JsonObject)"
       - "kotlin.Pair.constructor(com.google.gson.JsonObject, kotlin.Long)"
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
@@ -1,0 +1,215 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric
+
+import com.datadog.android.core.metrics.PerformanceMetric
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.internal.domain.scope.RumViewManagerScope
+import com.datadog.android.rum.model.ViewEvent
+import java.util.concurrent.TimeUnit
+
+/**
+ * Metric for rum session ended event.
+ */
+@Suppress("TooManyFunctions")
+internal class SessionEndedMetric(
+    private val sessionId: String,
+    private val startReason: RumSessionScope.StartReason
+) {
+
+    private val trackedViewsById = mutableMapOf<String, TrackedView>()
+
+    private val errorKindFrequencies = mutableMapOf<String, Int>()
+
+    private var firstTrackedView: TrackedView? = null
+    private var lastTrackedView: TrackedView? = null
+    private var wasStopped: Boolean = false
+
+    /**
+     * Called on view tracked event, return true if the view is recorded by the metric, false otherwise.
+     */
+    fun onViewTracked(rumViewEvent: ViewEvent): Boolean {
+        if (rumViewEvent.session.id != sessionId) {
+            return false
+        }
+        val trackedView = TrackedView(
+            viewUrl = trackedViewsById[rumViewEvent.view.id]?.viewUrl ?: rumViewEvent.view.url,
+            startMs = trackedViewsById[rumViewEvent.view.id]?.startMs ?: rumViewEvent.date,
+            durationNs = rumViewEvent.view.timeSpent
+        )
+
+        trackedViewsById[rumViewEvent.view.id] = trackedView
+        if (firstTrackedView == null) {
+            firstTrackedView = trackedView
+        }
+        lastTrackedView = trackedView
+        return true
+    }
+
+    fun onErrorTracked(sdkErrorKind: String) {
+        errorKindFrequencies[sdkErrorKind] = (errorKindFrequencies[sdkErrorKind] ?: 0) + 1
+    }
+
+    fun onSessionStopped() {
+        wasStopped = true
+    }
+
+    fun toMetricAttributes(): Map<String, Any?> {
+        return mapOf(
+            METRIC_TYPE_KEY to METRIC_TYPE_VALUE,
+            RSE_KEY to resolveRseAttributes()
+        )
+    }
+
+    private fun calculateDuration(): Long {
+        return lastTrackedView?.let { last ->
+            firstTrackedView?.let { first ->
+                TimeUnit.MILLISECONDS.toNanos(last.startMs - first.startMs) + last.durationNs
+            }
+        } ?: 0L
+    }
+
+    private fun resolveRseAttributes(): Map<String, Any?> {
+        return mapOf(
+            PROCESS_TYPE_KEY to PROCESS_TYPE_VALUE,
+            PRECONDITION_KEY to startReason.asString,
+            DURATION_KEY to calculateDuration(),
+            WAS_STOPPED_KEY to wasStopped,
+            VIEW_COUNTS_KEY to resolveViewCountsAttributes(),
+            SDK_ERRORS_COUNT_KEY to resolveSDKErrorsCountAttributes()
+        )
+    }
+
+    private fun resolveViewCountsAttributes(): Map<String, Any?> {
+        return mapOf(
+            VIEW_COUNTS_TOTAL_KEY
+                to trackedViewsById.size,
+            VIEW_COUNTS_BG_KEY
+                to trackedViewsById.values.count { it.viewUrl == RumViewManagerScope.RUM_BACKGROUND_VIEW_URL },
+            VIEW_COUNTS_APP_LAUNCH_KEY
+                to trackedViewsById.values.count { it.viewUrl == RumViewManagerScope.RUM_APP_LAUNCH_VIEW_URL }
+        )
+    }
+
+    private fun resolveSDKErrorsCountAttributes(): Map<String, Any?> {
+        return mapOf(
+            SDK_ERRORS_COUNT_TOTAL_KEY to errorKindFrequencies.values.sum(),
+            SDK_ERRORS_COUNT_BY_KIND_KEY to resolveTop5ErrorsByKind()
+        )
+    }
+
+    private fun resolveTop5ErrorsByKind(): Map<String, Int> {
+        val limit = TOP_ERROR_LIMIT.coerceAtMost(errorKindFrequencies.size)
+        return errorKindFrequencies.entries.sortedByDescending {
+            it.value
+        }.subList(0, limit).associate { entry ->
+            escapeNonAlphanumericCharacters(entry.key) to entry.value
+        }
+    }
+
+    private fun escapeNonAlphanumericCharacters(key: String): String {
+        val regex = Regex("[^\\w']+")
+        return key.replace(regex, "_")
+    }
+
+    companion object {
+
+        /**
+         * The metric holds the map of TOP 5 error kinds to the number of their occurrences in the session.
+         */
+        private const val TOP_ERROR_LIMIT = 5
+
+        /**
+         * Title of the metric to be sent.
+         */
+        internal const val RUM_SESSION_ENDED_METRIC_NAME: String = "[Mobile Metric] RUM Session Ended"
+
+        /**
+         * Basic Metric type key.
+         */
+        internal const val METRIC_TYPE_KEY: String = PerformanceMetric.METRIC_TYPE
+
+        /**
+         * Metric type value.
+         */
+        internal const val METRIC_TYPE_VALUE: String = "rum session ended"
+
+        /**
+         * Name space of bundling.
+         */
+        internal const val RSE_KEY = "rse"
+
+        /**
+         * Key of process type.
+         */
+        internal const val PROCESS_TYPE_KEY = "process_type"
+
+        /**
+         * The type of OS component where the session was tracked. Android platform has only "app" type.
+         */
+        internal const val PROCESS_TYPE_VALUE = "app"
+
+        /**
+         * Key of the reason that led to the creation of this session.
+         */
+        internal const val PRECONDITION_KEY = "precondition"
+
+        /**
+         * Key of the session's duration in nanoseconds.
+         */
+        internal const val DURATION_KEY = "duration"
+
+        /**
+         * Key of boolean value indicating if the session was stopped through stopSession() API.
+         */
+        internal const val WAS_STOPPED_KEY = "was_stopped"
+
+        /**
+         * Key of the attributes of view counts.
+         */
+        internal const val VIEW_COUNTS_KEY = "views_count"
+
+        /**
+         * Key of the total view counts.
+         */
+        internal const val VIEW_COUNTS_TOTAL_KEY = "total"
+
+        /**
+         * Key of the number of standard "Background" views tracked during this session.
+         */
+        internal const val VIEW_COUNTS_BG_KEY = "background"
+
+        /**
+         * Key of the number of standard "ApplicationLaunch" views tracked during this session.
+         */
+        internal const val VIEW_COUNTS_APP_LAUNCH_KEY = "app_launch"
+
+        /**
+         * Key of the sdk errors attribute.
+         */
+        internal const val SDK_ERRORS_COUNT_KEY = "sdk_errors_count"
+
+        /**
+         * Key of the counts the total number of SDK errors that occurred during the session.
+         */
+        internal const val SDK_ERRORS_COUNT_TOTAL_KEY = "total"
+
+        /**
+         * Key of TOP [TOP_ERROR_LIMIT] error kinds to the number of their occurrences in the session.
+         */
+        internal const val SDK_ERRORS_COUNT_BY_KIND_KEY = "by_kind"
+    }
+
+    /**
+     * Class to hold the view information tracked by the metric.
+     */
+    internal data class TrackedView(
+        val viewUrl: String,
+        val startMs: Long,
+        val durationNs: Long
+    )
+}
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
@@ -1,0 +1,68 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.internal.domain.scope.RumViewManagerScope
+import com.datadog.android.rum.model.ViewEvent
+import java.util.concurrent.ConcurrentHashMap
+
+internal class SessionEndedMetricDispatcher(private val internalLogger: InternalLogger) :
+    SessionMetricDispatcher {
+
+    private val metricsBySessionId = ConcurrentHashMap<String, SessionEndedMetric>()
+
+    override fun startMetric(
+        sessionId: String,
+        startReason: RumSessionScope.StartReason
+    ) {
+        metricsBySessionId[sessionId] = SessionEndedMetric(sessionId, startReason)
+    }
+
+    override fun endMetric(sessionId: String) {
+        // the argument is always non - null, so we can suppress the warning
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        val metric = metricsBySessionId.remove(sessionId)
+        metric?.let {
+            internalLogger.logMetric(
+                messageBuilder = { SessionEndedMetric.RUM_SESSION_ENDED_METRIC_NAME },
+                additionalProperties = it.toMetricAttributes()
+            )
+        }
+    }
+
+    override fun onSessionStopped(sessionId: String) {
+        metricsBySessionId[sessionId]?.onSessionStopped()
+    }
+
+    override fun onViewTracked(sessionId: String, viewEvent: ViewEvent) {
+        val result = metricsBySessionId[sessionId]?.onViewTracked(viewEvent) ?: false
+        if (result.not()) {
+            internalLogger.log(
+                InternalLogger.Level.INFO,
+                InternalLogger.Target.MAINTAINER,
+                { buildViewTrackError(sessionId, viewEvent) }
+            )
+        }
+    }
+
+    override fun onSdkErrorTracked(sessionId: String, errorKind: String?) {
+        errorKind?.let {
+            metricsBySessionId[sessionId]?.onErrorTracked(it)
+        }
+    }
+
+    private fun buildViewTrackError(sessionId: String, viewEvent: ViewEvent): String {
+        val viewType = when (viewEvent.view.url) {
+            RumViewManagerScope.RUM_APP_LAUNCH_VIEW_URL -> "AppLaunch"
+            RumViewManagerScope.RUM_BACKGROUND_VIEW_URL -> "Background"
+            else -> "Custom"
+        }
+        return "Failed to track $viewType view in session with different UUID $sessionId"
+    }
+}
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionMetricDispatcher.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionMetricDispatcher.kt
@@ -1,0 +1,43 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric
+
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.model.ViewEvent
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface to dispatch the session metric.
+ */
+@NoOpImplementation
+internal interface SessionMetricDispatcher {
+
+    /**
+     * Starts a session metric with given session id and start reason of this session.
+     */
+    fun startMetric(sessionId: String, startReason: RumSessionScope.StartReason)
+
+    /**
+     * Ends the session metric with given session id.
+     */
+    fun endMetric(sessionId: String)
+
+    /**
+     * Called when the session is stopped.
+     */
+    fun onSessionStopped(sessionId: String)
+
+    /**
+     * Called when a view is tracked by this session metric.
+     */
+    fun onViewTracked(sessionId: String, viewEvent: ViewEvent)
+
+    /**
+     * Called when a sdk error is tracked by this session metric.
+     */
+    fun onSdkErrorTracked(sessionId: String, errorKind: String?)
+}
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/FakeInternalLogger.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/FakeInternalLogger.kt
@@ -1,0 +1,57 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.metrics.PerformanceMetric
+import com.datadog.android.core.metrics.TelemetryMetricType
+
+class FakeInternalLogger : InternalLogger {
+
+    var lastMetric: Pair<String, Map<String, Any?>>? = null
+
+    var errorLog: String? = null
+
+    override fun log(
+        level: InternalLogger.Level,
+        target: InternalLogger.Target,
+        messageBuilder: () -> String,
+        throwable: Throwable?,
+        onlyOnce: Boolean,
+        additionalProperties: Map<String, Any?>?
+    ) {
+        errorLog = messageBuilder()
+    }
+
+    override fun log(
+        level: InternalLogger.Level,
+        targets: List<InternalLogger.Target>,
+        messageBuilder: () -> String,
+        throwable: Throwable?,
+        onlyOnce: Boolean,
+        additionalProperties: Map<String, Any?>?
+    ) {
+        // do nothing
+    }
+
+    override fun logMetric(
+        messageBuilder: () -> String,
+        additionalProperties: Map<String, Any?>
+    ) {
+        lastMetric = Pair(messageBuilder(), additionalProperties)
+    }
+
+    override fun startPerformanceMeasure(
+        callerClass: String,
+        metric: TelemetryMetricType,
+        samplingRate: Float,
+        operationName: String
+    ): PerformanceMetric? {
+        // do nothing
+        return null
+    }
+}
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
@@ -1,0 +1,431 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.internal.domain.scope.RumViewManagerScope
+import com.datadog.android.rum.model.ViewEvent
+import com.datadog.android.rum.utils.forge.Configurator
+import com.datadog.tools.unit.extensions.TestConfigurationExtension
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.annotation.Forgery
+import fr.xgouchet.elmyr.annotation.StringForgery
+import fr.xgouchet.elmyr.junit5.ForgeConfiguration
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.AssertionsForClassTypes.assertThat
+import org.json.JSONObject
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.Mock
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+import java.util.concurrent.TimeUnit
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class),
+    ExtendWith(TestConfigurationExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+class RumSessionEndedMetricTest {
+
+    @StringForgery
+    private lateinit var fakeSessionId: String
+
+    @Forgery
+    private lateinit var fakeStartReason: RumSessionScope.StartReason
+
+    @Mock
+    lateinit var mockInternalLogger: InternalLogger
+
+    @Test
+    fun `M create attributes with session ended metric type W toMetricAttributes`() {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        assertThat(attributes[SessionEndedMetric.METRIC_TYPE_KEY]).isEqualTo(SessionEndedMetric.METRIC_TYPE_VALUE)
+    }
+
+    @Test
+    fun `M create attributes with session ended process type W toMetricAttributes`() {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val processType = rse[SessionEndedMetric.PROCESS_TYPE_KEY] as String
+        assertThat(processType).isEqualTo(SessionEndedMetric.PROCESS_TYPE_VALUE)
+    }
+
+    @Test
+    fun `M create attributes with session ended precondition W toMetricAttributes`() {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val precondition = rse[SessionEndedMetric.PRECONDITION_KEY] as String
+        assertThat(precondition).isEqualTo(fakeStartReason.asString)
+    }
+
+    @Test
+    fun `M init attributes W toMetricAttributes { empty metric }`() {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+
+        assertThat(rse[SessionEndedMetric.DURATION_KEY] as Long).isEqualTo(0L)
+        assertThat(rse[SessionEndedMetric.WAS_STOPPED_KEY] as Boolean).isFalse()
+
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY] as Int).isEqualTo(0)
+        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_BG_KEY] as Int).isEqualTo(0)
+        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_APP_LAUNCH_KEY] as Int).isEqualTo(0)
+
+        val errorCounts = rse[SessionEndedMetric.SDK_ERRORS_COUNT_KEY] as Map<*, *>
+        assertThat(errorCounts[SessionEndedMetric.SDK_ERRORS_COUNT_TOTAL_KEY] as Int).isEqualTo(0)
+        assertTrue((errorCounts[SessionEndedMetric.SDK_ERRORS_COUNT_BY_KIND_KEY] as Map<*, *>).isEmpty())
+    }
+
+    @Test
+    fun `M calculate wasStopped attribute W onSessionStopped`() {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        sessionEndedMetric.onSessionStopped()
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val wasStopped = rse[SessionEndedMetric.WAS_STOPPED_KEY] as Boolean
+        assertThat(wasStopped).isTrue()
+    }
+
+    @Test
+    fun `M calculate duration W onViewTracked { track single view }`(@Forgery event: ViewEvent) {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(event.session.id, fakeStartReason)
+
+        // When
+        sessionEndedMetric.onViewTracked(event)
+
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val duration = rse[SessionEndedMetric.DURATION_KEY] as Long
+        assertThat(duration).isEqualTo(event.view.timeSpent)
+    }
+
+    @Test
+    fun `M calculate duration W onViewTracked {  multiple views tracked }`(
+        @Forgery event1: ViewEvent,
+        @Forgery event2: ViewEvent,
+        @Forgery event3: ViewEvent
+    ) {
+        // Given
+        val duration1 = TimeUnit.SECONDS.toNanos(15)
+        val duration2 = TimeUnit.SECONDS.toNanos(10)
+        val duration3 = TimeUnit.SECONDS.toNanos(5)
+        val date1 = TimeUnit.SECONDS.toMillis(20)
+        val date2 = TimeUnit.SECONDS.toMillis(30)
+        val date3 = TimeUnit.SECONDS.toMillis(40)
+        val sessionEvent1 = event1.copy(
+            session = event1.session.copy(id = fakeSessionId),
+            date = date1,
+            view = event1.view.copy(timeSpent = duration1)
+        )
+        val sessionEvent2 = event2.copy(
+            session = event2.session.copy(id = fakeSessionId),
+            date = date2,
+            view = event2.view.copy(timeSpent = duration2)
+        )
+        val sessionEvent3 = event3.copy(
+            session = event3.session.copy(id = fakeSessionId),
+            date = date3,
+            view = event3.view.copy(timeSpent = duration3)
+        )
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        sessionEndedMetric.onViewTracked(sessionEvent1)
+        sessionEndedMetric.onViewTracked(sessionEvent2)
+        sessionEndedMetric.onViewTracked(sessionEvent3)
+
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val duration = rse[SessionEndedMetric.DURATION_KEY] as Long
+        val expected = TimeUnit.MILLISECONDS.toNanos(date3 - date1) + duration3
+        assertThat(duration).isEqualTo(expected)
+    }
+
+    @Test
+    fun `M ignore unexpected view W onViewTracked { track different session id view event }`(
+        @Forgery event1: ViewEvent,
+        @Forgery event2: ViewEvent
+    ) {
+        // Given
+        val duration1 = TimeUnit.SECONDS.toNanos(15)
+        val duration2 = TimeUnit.SECONDS.toNanos(10)
+        val date1 = TimeUnit.SECONDS.toMillis(20)
+        val date2 = TimeUnit.SECONDS.toMillis(30)
+
+        val currentSessionViewEvent = event1.copy(
+            session = event1.session.copy(id = fakeSessionId),
+            date = date1,
+            view = event1.view.copy(timeSpent = duration1)
+        )
+        val differentSessionViewEvent = event2.copy(
+            date = date2,
+            view = event2.view.copy(timeSpent = duration2)
+        )
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        sessionEndedMetric.onViewTracked(currentSessionViewEvent)
+        sessionEndedMetric.onViewTracked(differentSessionViewEvent)
+
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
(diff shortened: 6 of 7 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,221 @@
...
+        val limit = TOP_ERROR_LIMIT.coerceAtMost(trackedErrorMap.size)
+        trackedErrorMap.map { Pair(it.key, it.value) }.sortedByDescending {
+            it.second
+        }.subList(0, limit).forEach {
+            map[it.first.escapeNonAlphanumericCharacters()] = it.second
+        }
+        return map
+    }
+
+    private fun String.escapeNonAlphanumericCharacters(): String {
```
mariusc83, line 120: nit: usually we are adding extension functions when it's being shared between multiple classes. Using a simple private function would be better here and more optimised.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,221 @@
...
+        return mapOf(
+            METRIC_TYPE_KEY to METRIC_TYPE_VALUE,
+            RSE_KEY to resolveRseAttributes()
+        )
+    }
+
+    private fun calculateDuration(): Long {
+        return lastTrackedView?.let { last ->
+            firstTrackedView?.let { first ->
+                last.startMs + last.durationNs - first.startMs
```
mariusc83, line 73: are you adding millisecond and nanoseconds here ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Metric for rum session ended event.
+ */
+@Suppress("TooManyFunctions")
+internal class SessionEndedMetric(
+    private val sessionId: String,
+    private val precondition: String
+) {
+
+    private val trackedViewMap by lazy { mutableMapOf<String, TrackedView>() }
```
mariusc83, line 22: when do we purge this map ? I am a bit worried about memory allocation.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,221 @@
...
+        if (rumViewEvent.session.id != sessionId) {
+            return false
+        }
+        val trackedView = trackedViewMap[rumViewEvent.view.id] ?: TrackedView(
+            viewUrl = rumViewEvent.view.url,
+            startMs = rumViewEvent.date,
+            durationNs = rumViewEvent.view.timeSpent
+        )
+
+        trackedViewMap[rumViewEvent.view.id] = trackedView.copy(durationNs = rumViewEvent.view.timeSpent)
```
mariusc83, line 43: nit: in case the view is tracked for the first time we are creating 2 objects here instead of one. This maybe could be optimised but is not that bad.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,221 @@
...
+                to trackedViewMap.values.count { it.viewUrl == RumViewManagerScope.RUM_APP_LAUNCH_VIEW_URL }
+        )
+    }
+
+    private fun resolveSDKErrorsCountAttributes(): Map<String, Any?> {
+        return mapOf(
+            SDK_ERRORS_COUNT_TOTAL_KEY to trackedErrorMap.values.sumOf {
+                it
+            },
+            SDK_ERRORS_COUNT_BY_KIND_KEY to resolveTop5ErrorsByKind()
```
mariusc83, line 105: is it really count here ? the way I see it it is a `Map<String,Int>` that we attach here, am I wrong ?

mariusc83 (changes requested):

mariusc83 (changes requested):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,71 @@
...
+        if (result.not()) {
+            internalLogger.log(
+                InternalLogger.Level.INFO,
+                InternalLogger.Target.MAINTAINER,
+                { buildViewTrackError(sessionId, viewEvent) }
+            )
+        }
+    }
+
+    override fun onSdkErrorTracked(errorKind: String?) {
```
0xnm, line 56: what is the supposed call site for this? If we can supply `sessionId` as an argument here, then we won't need to have `LinkedHashMap` and do `last` (not a big deal though).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,71 @@
...
+
+package com.datadog.android.rum.internal.metric
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.internal.domain.scope.RumViewManagerScope
+import com.datadog.android.rum.model.ViewEvent
+
+/**
+ * Implementation of [SessionMetricDispatcher].
+ */
```
0xnm, line 15: nit: can be removed, signature is already self-explanatory.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,221 @@
...
+ */
+@Suppress("TooManyFunctions")
+internal class SessionEndedMetric(
+    private val sessionId: String,
+    private val precondition: String
+) {
+
+    private val trackedViewMap by lazy { mutableMapOf<String, TrackedView>() }
+
+    private val trackedErrorMap by lazy { mutableMapOf<String, Int>() }
```
0xnm, line 25: any benefit of `lazy`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,221 @@
...
+    private fun resolveSDKErrorsCountAttributes(): Map<String, Any?> {
+        return mapOf(
+            SDK_ERRORS_COUNT_TOTAL_KEY to trackedErrorMap.values.sumOf {
+                it
+            },
+            SDK_ERRORS_COUNT_BY_KIND_KEY to resolveTop5ErrorsByKind()
+        )
+    }
+
+    private fun resolveTop5ErrorsByKind(): Map<String, Int> {
```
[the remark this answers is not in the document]
ambushwork: Nice one, updated!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: according to the spec: the `total` is the count, `by_kind` is a map, with the error name as key, the error count of each kind:
```
  “sdk_errors_count”: { “total”: <int>, “by_kind”: { “<error.kind>”: <int> } },
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,222 @@
...
+    fun onErrorTracked(sdkErrorKind: String) {
+        trackedErrorMap[sdkErrorKind] = (trackedErrorMap[sdkErrorKind] ?: 0) + 1
+    }
+
+    fun onSessionStopped() {
+        wasStopped = true
+    }
+
+    fun toMetricAttributes(): Map<String, Any?> {
+        return resolveAttributes()
```
0xnm, line 61: since this method just calls another method, you can simply inline it:

```suggestion
        return mapOf(
            METRIC_TYPE_KEY to METRIC_TYPE_VALUE,
            RSE_KEY to resolveRseAttributes()
        )
```
```suggestion
        return resolveAttributes()
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,222 @@
...
+            VIEW_COUNTS_APP_LAUNCH_KEY
+                to trackedViewMap.values.count { it.viewUrl == RumViewManagerScope.RUM_APP_LAUNCH_VIEW_URL }
+        )
+    }
+
+    private fun resolveSDKErrorsCountAttributes(): Map<String, Any?> {
+        return mapOf(
+            SDK_ERRORS_COUNT_TOTAL_KEY to trackedErrorMap.values.sumOf {
+                it
+            },
```
0xnm, line 105: nit: can be also like that, if I'm not wrong

```suggestion
            SDK_ERRORS_COUNT_TOTAL_KEY to trackedErrorMap.values.sum(),
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,222 @@
...
+ */
+@Suppress("TooManyFunctions")
+internal class SessionEndedMetric(
+    private val sessionId: String,
+    private val precondition: String
+) {
+
+    private val trackedViewMap by lazy { mutableMapOf<String, TrackedView>() }
+
+    private val trackedErrorMap by lazy { mutableMapOf<String, Int>() }
```
0xnm, line 25: minor: maybe it makes sense to change the naming

```suggestion
    private val trackedViewsById by lazy { mutableMapOf<String, TrackedView>() }

    private val errorKindFrequencies by lazy { mutableMapOf<String, Int>() }
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,222 @@
...
+            SDK_ERRORS_COUNT_BY_KIND_KEY to resolveTop5ErrorsByKind()
+        )
+    }
+
+    private fun resolveTop5ErrorsByKind(): Map<String, Int> {
+        val map = mutableMapOf<String, Int>()
+        val limit = TOP_ERROR_LIMIT.coerceAtMost(trackedErrorMap.size)
+        trackedErrorMap.entries.sortedByDescending {
+            it.value
+        }.subList(0, limit).associate { entry ->
```
0xnm, line 115:
```suggestion
        }.subList(0, limit).associate { entry ->
```
```suggestion
        }.take(limit).associate { entry ->
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,409 @@
...
+    @Test
+    fun `M metric type W creating metric`() {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakePrecondition)
+
+        // When
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        Assertions.assertEquals(attributes[SessionEndedMetric.METRIC_TYPE_KEY], SessionEndedMetric.METRIC_TYPE_VALUE)
```
0xnm, line 57: minor: we usually use static imports for `Assertions.assertEquals`, etc. So that it is not needed to write `Assertions` each time.

0xnm, line 57: arguments should be swapped: `assertEquals` has `expected,actual` order, not `actual,expected`. Same applies to other tests.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,409 @@
...
+    private lateinit var fakeSessionId: String
+
+    @StringForgery
+    private lateinit var fakePrecondition: String
+
+    @Mock
+    lateinit var mockInternalLogger: InternalLogger
+
+    @Test
+    fun `M metric type W creating metric`() {
```
0xnm, line 49: I think we need to update the test naming here to be more descriptive. Here is a proposed example:

```suggestion
    fun `M create attributes with session ended metric type W toMetricAttributes`() {
```

Same format may be applied to other tests.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,409 @@
...
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakePrecondition)
+
+        // When
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+
+        // Assertions.assertTrue() // todo with duration
```
0xnm, line 99: Don't leave dangling `TODO` - it is either should be addressed to should be backed by the JIRA ticket reference. Here I think it should be handled as a part of this PR.

0xnm (changes requested):
nice feature! I went through the production files and gave a brief look on the tests so far.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: updated.

detekt_custom.yml
```diff
@@ -1000,6 +1001,13 @@ datadog:
       - "kotlin.sequences.Sequence.mapNotNull(kotlin.Function1)"
       - "kotlin.sequences.Sequence.reduceOrNull(kotlin.Function2)"
       - "kotlin.sequences.Sequence.sortedBy(kotlin.Function1)"
+      - "kotlin.collections.linkedMapOf()"
+      - "kotlin.Boolean.not()"
+      - "kotlin.collections.MutableCollection.count(kotlin.Function1)"
+      - "kotlin.collections.MutableCollection.sumOf(kotlin.Function1)"
+      - "kotlin.Int.coerceAtMost(kotlin.Int)"
+      - "kotlin.Pair.constructor(kotlin.String, kotlin.Int)"
+      - "kotlin.collections.List.sortedByDescending(kotlin.Function1)"
```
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionMetricDispatcher.kt
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Interface to dispatch the session metric.
+ */
+@NoOpImplementation
+internal interface SessionMetricDispatcher {
+
+    /**
+     * Starts a session metric with given session id and start reason of this session.
+     */
+    fun startMetric(sessionId: String, precondition: String)
```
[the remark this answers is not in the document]
ambushwork: updated with `RumSessionScope.StartReason`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionMetricDispatcher.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Called when a view is tracked by this session metric.
+     */
+    fun onViewTracked(sessionId: String, viewEvent: ViewEvent)
+
+    /**
+     * Called when a sdk error is tracked by this session metric.
+     */
+    fun onSdkErrorTracked(errorKind: String?)
```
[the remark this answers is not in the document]
ambushwork: I plan to use `SendTelemetry.kind` as the input of the function, so it's not a enum

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,71 @@
...
+internal class SessionEndedMetricDispatcher(private val internalLogger: InternalLogger) :
+    SessionMetricDispatcher {
+
+    private val metricsMap by lazy { linkedMapOf<String, SessionEndedMetric>() }
```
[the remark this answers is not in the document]
ambushwork: updated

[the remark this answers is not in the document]
ambushwork: I assume that it can have less memory allocation until the metricsMap is initialised, is it justified to you?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,71 @@
...
+        val metric = metricsMap[sessionId]
+
+        metric?.let {
+            internalLogger.logMetric(
+                messageBuilder = { SessionEndedMetric.RUM_SESSION_ENDED_METRIC_NAME },
+                additionalProperties = metric.toMetricAttributes()
+            )
+        }
+
+        metricsMap.remove(sessionId)
```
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,71 @@
...
+                { buildViewTrackError(sessionId, viewEvent) }
+            )
+        }
+    }
+
+    override fun onSdkErrorTracked(errorKind: String?) {
+        errorKind?.let {
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            metricsMap.entries.last().value.onErrorTracked(it)
+        }
```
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
ambushwork: I plan to call it inside `TelemetryEventHandler` when handling a error, I am not sure I can have the session id there

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Updated all the similar functions names

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: it's a forgetten todo, now fixed now

mariusc83 (commented):
Good job here, once @0xnm will be addressed will have a last review.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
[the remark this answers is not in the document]
0xnm: It will be at least the same, because `by lazy` is just a sugar around `SynchronizedLazyImpl` class which will be created to wrap a property. This class has its own fields and also involves some synchronization (for the initialization path).

In fact, in many languages containers (like map, list, etc.) are just very thin objects around the objects they store. By itself container takes just few bytes of the memory (excluding the size of the objects they keep).

So `lazy` here doesn't help. Usually it should be used for things which are heavy to initialize and they may be not needed during the class lifecycle, or for the things which cannot be initialized at the object creation time.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
0xnm: You can get it there in a few ways (depends on the exact place you want to use):

* if it is inside `sdkCore.getFeature(Feature.RUM_FEATURE_NAME)?.withWriteContext` you can get it from `DatadogContext`
* you also have `onSessionStarted` there, so you can just keep the current session ID in the property

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: update all the similar functions

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: removed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
ambushwork: OK, then I added `sessionId` as parameter inside this function

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    override fun startMetric(
+        sessionId: String,
+        precondition: String
+    ) {
+        metricsMap[sessionId] = SessionEndedMetric(sessionId, precondition)
+    }
+
+    override fun endMetric(sessionId: String) {
+        val metric = metricsMap[sessionId]
```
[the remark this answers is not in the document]
ambushwork: Do you think ConcurrentHashMap will solve the issue?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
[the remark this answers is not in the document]
ambushwork: Good to know, I remove all the lazy now

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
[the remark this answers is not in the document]
0xnm: sounds like a reasonable choice if this collection can be modified from the different threads

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/FakeInternalLogger.kt
```diff
@@ -0,0 +1,57 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.metrics.PerformanceMetric
+import com.datadog.android.core.metrics.TelemetryMetricType
+
+class FakeInternalLogger : InternalLogger {
```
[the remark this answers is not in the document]
ambushwork: I think with a "fake" internal logger, I can easier intercept the telemetry event in unit test

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/FakeInternalLogger.kt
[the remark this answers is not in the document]
0xnm: you can also intercept them by using mock + argument captors, like (here is an example if only single argument needs to be captured; for multiple arguments syntax will be slightly different):

```
val mockLogger = mock<InternalLogger>()

argumentCaptor<FooBar>() {
  
  verify(mockLogger).sendMetric(capture(),...)
  assertThat(lastValue).isEqualTo(...)
 
}

```

We have multiple sites where we use mocks + captors in our tests. But it is not blocking, you can keep `FakeInternalLogger` if you want.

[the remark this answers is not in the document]
ambushwork: Yes, actually I hesitated between mock and fake, I think fake here is simpler and it can meet what I need in the unit test, so let's keep it for the moment

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -0,0 +1,215 @@
...
+    private val startReason: RumSessionScope.StartReason
+) {
+
+    private val trackedViewsById = mutableMapOf<String, TrackedView>()
+
+    private val errorKindFrequencies = mutableMapOf<String, Int>()
+
+    private var firstTrackedView: TrackedView? = null
+    private var lastTrackedView: TrackedView? = null
+    private var wasStopped: Boolean = false
```
0xnm, line 30: minor: it is better to use present tense here

```suggestion
    private var isStopped: Boolean = false
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
0xnm: seems it is still `subList` usage here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.internal.domain.scope.RumSessionScope
+import com.datadog.android.rum.internal.domain.scope.RumViewManagerScope
+import com.datadog.android.rum.model.ViewEvent
+import java.util.concurrent.ConcurrentHashMap
+
+internal class SessionEndedMetricDispatcher(private val internalLogger: InternalLogger) :
+    SessionMetricDispatcher {
+
+    private val metricsBySessionIdMap: ConcurrentHashMap<String, SessionEndedMetric> = ConcurrentHashMap()
```
0xnm, line 18: minor: can drop `Map` suffix, naming of the format `xByY` already implies that it is association here. Also can drop explicit type declaration here.

```suggestion
    private val metricsBySessionId = ConcurrentHashMap<String, SessionEndedMetric>()
```

mariusc83 (approved):
Nice job !!

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+    @Test
+    fun `M create attributes with session ended metric type W toMetricAttributes`() {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        assertEquals(SessionEndedMetric.METRIC_TYPE_VALUE, attributes[SessionEndedMetric.METRIC_TYPE_KEY])
```
0xnm, line 60: very minor: in most cases we are using `assertThat(...).isEqualTo(...)` instead of `assertEquals`. but this is really a minor thing. Overall, `assertThat` usage can be applied to all the call sites like `assertEquals`, `assertTrue`, etc. call sites below. It gives a set of more specific assertions depending on the argument type.

```suggestion
        assertEquals(SessionEndedMetric.METRIC_TYPE_VALUE, attributes[SessionEndedMetric.METRIC_TYPE_KEY])
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        assertFalse(rse[SessionEndedMetric.WAS_STOPPED_KEY] as Boolean)
+
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertEquals(viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY] as Int, 0)
+        assertEquals(viewCounts[SessionEndedMetric.VIEW_COUNTS_BG_KEY] as Int, 0)
+        assertEquals(viewCounts[SessionEndedMetric.VIEW_COUNTS_APP_LAUNCH_KEY] as Int, 0)
+
+        val errorCounts = rse[SessionEndedMetric.SDK_ERRORS_COUNT_KEY] as Map<*, *>
+        assertEquals(errorCounts[SessionEndedMetric.SDK_ERRORS_COUNT_TOTAL_KEY] as Int, 0)
+        Assertions.assertTrue((errorCounts[SessionEndedMetric.SDK_ERRORS_COUNT_BY_KIND_KEY] as Map<*, *>).isEmpty())
```
0xnm, line 113: static import to be used

```suggestion
        assertThat(errorCounts[SessionEndedMetric.SDK_ERRORS_COUNT_BY_KIND_KEY] as Map<*, *>).isEmpty()
```

bonus point for using `assertThat(...).isEmpty(...)` is more detailed message in case of failure, rather that something like "condition was false, but expected to to be true".

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val wasStopped = rse[SessionEndedMetric.WAS_STOPPED_KEY] as Boolean
+        Assertions.assertTrue(wasStopped)
+    }
+
+    @Test
+    fun `M duration W track single view`(@Forgery event: ViewEvent) {
```
0xnm, line 132:
```suggestion
    fun `M calculate duration W onViewTracked { track single view }`(@Forgery event: ViewEvent) {
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        assertEquals(viewCounts[SessionEndedMetric.VIEW_COUNTS_BG_KEY] as Int, 0)
+        assertEquals(viewCounts[SessionEndedMetric.VIEW_COUNTS_APP_LAUNCH_KEY] as Int, 0)
+
+        val errorCounts = rse[SessionEndedMetric.SDK_ERRORS_COUNT_KEY] as Map<*, *>
+        assertEquals(errorCounts[SessionEndedMetric.SDK_ERRORS_COUNT_TOTAL_KEY] as Int, 0)
+        Assertions.assertTrue((errorCounts[SessionEndedMetric.SDK_ERRORS_COUNT_BY_KIND_KEY] as Map<*, *>).isEmpty())
+    }
+
+    @Test
+    fun `M wasStopped attribute W onSessionStopped called`() {
```
0xnm, line 117:
```suggestion
    fun `M calculate wasStopped attribute W onSessionStopped`() {
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val precondition = rse[SessionEndedMetric.PRECONDITION_KEY] as String
+        assertEquals(fakeStartReason.asString, precondition)
+    }
+
+    @Test
+    fun `M init attributes W creating empty metric`() {
```
0xnm, line 92:
```suggestion
    fun `M init attributes W toMetricAttributes { empty metric }`() {
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        val attributes = sessionEndedMetric.toMetricAttributes()
+
+        // Then
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val duration = rse[SessionEndedMetric.DURATION_KEY] as Long
+        assertEquals(event.view.timeSpent, duration)
+    }
+
+    @Test
+    fun `M duration W track multi view`(
```
0xnm, line 148:
```suggestion
    fun `M calculate duration W onViewTracked { multiple views tracked }`(
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val duration = rse[SessionEndedMetric.DURATION_KEY] as Long
+        val expected = TimeUnit.MILLISECONDS.toNanos(date3 - date1) + duration3
+        assertEquals(expected, duration)
+    }
+
+    @Test
+    fun `M ignore W track different session id view event`(
```
0xnm, line 191:
```suggestion
    fun `M ignore unexpected view W onViewTracked { track different session id view event }`(
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+
+        // When
+        sessionEndedMetric.onViewTracked(currentSessionViewEvent)
+        sessionEndedMetric.onViewTracked(differentSessionViewEvent)
+
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertEquals(1, viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY])
```
0xnm, line 220:
```suggestion
        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY]).hasSize(1)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    @Test
+    fun `M view counts total correct W track views`(@Forgery events: List<ViewEvent>) {
```
0xnm, line 224: `M` stands for must, so the phrase after it should have a verb. Overall, other tests names in this file should be updated.

```suggestion
    fun `M calculate total view count W onViewTracked`(@Forgery events: List<ViewEvent>) {
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+
+        // When
+        events.map {
+            it.copy(session = it.session.copy(id = fakeSessionId))
+        }.forEach {
+            sessionEndedMetric.onViewTracked(it)
+        }
+        // Then
```
0xnm, line 234:
```suggestion
        }

        // Then
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertEquals(events.count(), viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY])
```
0xnm, line 238:
```suggestion
        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY]).hasSize(events.size)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+                viewEvent.view.url
+            }
+            viewEvent.copy(
+                session = viewEvent.session.copy(id = fakeSessionId),
+                view = viewEvent.view.copy(url = url)
+            )
+        }.forEach {
+            sessionEndedMetric.onViewTracked(it)
+        }
+        // Then
```
0xnm, line 261:
```suggestion
        }

        // Then
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertEquals(backgroundViewCount, viewCounts[SessionEndedMetric.VIEW_COUNTS_BG_KEY])
```
0xnm, line 265:
```suggestion
        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_BG_KEY]).hasSize(backgroundViewCount)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+                viewEvent.view.url
+            }
+            viewEvent.copy(
+                session = viewEvent.session.copy(id = fakeSessionId),
+                view = viewEvent.view.copy(url = url)
+            )
+        }.forEach {
+            sessionEndedMetric.onViewTracked(it)
+        }
+        // Then
```
0xnm, line 288:
```suggestion
        }

        // Then
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertEquals(backgroundViewCount, viewCounts[SessionEndedMetric.VIEW_COUNTS_BG_KEY])
+    }
+
+    @Test
+    fun `M view counts app launch correct W track views`(@Forgery events: List<ViewEvent>) {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+        val appLaunchViewCount = Random.nextInt(10)
```
0xnm, line 272: bonus point: in case of failure you will get a seed printed, which can be used to reproduce the failure.

```suggestion
    fun `M view counts app launch correct W track views`(
      @Forgery events: List<ViewEvent>,
      @IntForgery(min = 1, max = 10) appLaunchViewCount
    ) {
        // Given
        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertEquals(events.count(), viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY])
+    }
+
+    @Test
+    fun `M view counts background correct W track views`(@Forgery events: List<ViewEvent>) {
+        // Given
+        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
+        val backgroundViewCount = Random.nextInt(events.size)
```
0xnm, line 245:
```suggestion
    fun `M view counts background correct W track views`(
      @Forgery events: List<ViewEvent>,
      forge: Forge
    ) {
        // Given
        val sessionEndedMetric = SessionEndedMetric(fakeSessionId, fakeStartReason)
        val backgroundViewCount = forge.anInt(min = 1, max = events.size + 1)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+                view = viewEvent.view.copy(url = url)
+            )
+        }.forEach {
+            sessionEndedMetric.onViewTracked(it)
+        }
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertEquals(appLaunchViewCount, viewCounts[SessionEndedMetric.VIEW_COUNTS_APP_LAUNCH_KEY])
```
0xnm, line 292:
```suggestion
        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_APP_LAUNCH_KEY]).hasSize(appLaunchViewCount)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        // When
+        errorKinds.forEach {
+            sessionEndedMetric.onErrorTracked(it)
+        }
+
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val sdkErrorCount = rse[SessionEndedMetric.SDK_ERRORS_COUNT_KEY] as Map<*, *>
+        assertEquals(errorKinds.size, sdkErrorCount[SessionEndedMetric.SDK_ERRORS_COUNT_TOTAL_KEY])
```
0xnm, line 309:
```suggestion
        assertThat(sdkErrorCount[SessionEndedMetric.SDK_ERRORS_COUNT_TOTAL_KEY]).hasSize(errorKinds.size)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,412 @@
...
+        Assertions.assertNotNull(rseObject.get(SessionEndedMetric.WAS_STOPPED_KEY) as Boolean)
+
+        val viewCountsObject = rseObject.get(SessionEndedMetric.VIEW_COUNTS_KEY) as JSONObject
+        Assertions.assertNotNull(viewCountsObject.get(SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY) as Int)
+        Assertions.assertNotNull(viewCountsObject.get(SessionEndedMetric.VIEW_COUNTS_BG_KEY) as Int)
+        Assertions.assertNotNull(viewCountsObject.get(SessionEndedMetric.VIEW_COUNTS_APP_LAUNCH_KEY) as Int)
+
+        val sdkErrorsCount = rseObject.get(SessionEndedMetric.SDK_ERRORS_COUNT_KEY) as JSONObject
+        Assertions.assertNotNull(sdkErrorsCount.get(SessionEndedMetric.SDK_ERRORS_COUNT_TOTAL_KEY) as Int)
+        Assertions.assertNotNull(sdkErrorsCount.get(SessionEndedMetric.SDK_ERRORS_COUNT_BY_KIND_KEY) as JSONObject)
```
0xnm, line 410: it cannot be null, because you already cast to the non-nullable type. If value is null, it would fail earlier, with `ClassCastException`, so I guess you need to modify these lines.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,136 @@
...
+@ForgeConfiguration(Configurator::class)
+class SessionEndedMetricDispatcherTest {
+
+    private val fakeInternalLogger = FakeInternalLogger()
+
+    @Forgery
+    private lateinit var fakeStartReason: RumSessionScope.StartReason
+
+    @Test
+    fun `M session stopped W call onSessionStop()`(@StringForgery fakeSessionId1: String) {
```
0xnm, line 41:
```suggestion
    fun `M session stopped W call onSessionStopped()`(@StringForgery fakeSessionId1: String) {
```

0xnm (changes requested):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Do you think change `as` to `as?` will do the trick? If the value is not mapped to the desired type, it will become null, then we can check its nullability.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: The attribute name in the json is "was_stopped", is it better to align the property name with this?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
0xnm: yes, it will. but you can also use `assertThat(...).isInstanceOf(...)` (or maybe even more descriptive assertion exists), which will be even more descriptive.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
0xnm: I think we can keep class property name different from the attribute name. But not a big deal tbh.

ambushwork: OK, since that I prefer keeping it, so that it makes property refer the attribute.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
ambushwork: updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Good point, updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: `viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY]` returns actually an `Int` not a collection, so I don't think it has `hasSize` API

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Same, viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY] returns actually an Int not a collection, so I don't think it has hasSize API, I updated with .isEqualTo()

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Same, viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY] returns actually an Int not a collection, so I don't think it has hasSize API, I updated with .isEqualTo()

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Good point, updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Same, viewCounts[SessionEndedMetric. VIEW_COUNTS_APP_LAUNCH_KEY] returns actually an Int not a collection, so I don't think it has hasSize API, I updated with .isEqualTo()

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: Good point, updated

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
0xnm: ah, I missed that

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,136 @@
...
+@ForgeConfiguration(Configurator::class)
+class SessionEndedMetricDispatcherTest {
+
+    private val fakeInternalLogger = FakeInternalLogger()
+
+    @Forgery
+    private lateinit var fakeStartReason: RumSessionScope.StartReason
+
+    @Test
+    fun `M session stopped W call onSessionStopped()`(@StringForgery fakeSessionId1: String) {
```
0xnm, line 41:
```suggestion
    fun `M register session stop W call onSessionStopped()`(@StringForgery fakeSessionId: String) {
```

0xnm (approved):
added a couple of comments/suggestion, but nothing blocking, all of them are regarding tests

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,431 @@
...
+        ).isInstanceOf(java.lang.Integer::class.java)
+        assertThat(
+            viewCountsObject.get(SessionEndedMetric.VIEW_COUNTS_APP_LAUNCH_KEY)
+        ).isInstanceOf(java.lang.Integer::class.java)
+
+        val sdkErrorsCount = rseObject.get(SessionEndedMetric.SDK_ERRORS_COUNT_KEY) as JSONObject
+
+        assertThat(
+            sdkErrorsCount.get(SessionEndedMetric.SDK_ERRORS_COUNT_TOTAL_KEY)
+        ).isInstanceOf(java.lang.Integer::class.java)
```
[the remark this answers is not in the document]
0xnm: probably boxed vs unboxed value then

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,68 @@
...
+        sessionId: String,
+        startReason: RumSessionScope.StartReason
+    ) {
+        metricsBySessionId[sessionId] = SessionEndedMetric(sessionId, startReason)
+    }
+
+    override fun endMetric(sessionId: String) {
+        // the argument is always non - null, so we can suppress the warning
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        val metric = metricsBySessionId.remove(sessionId)
```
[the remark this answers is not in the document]
ambushwork: Actually it's already declared here:
https://github.com/DataDog/dd-sdk-android/blob/develop/detekt_custom.yml#L204
Since it's a java function, it can throw NPE if the parameter is null, we just need to suppress it since in kotlin we are sure that the param is not null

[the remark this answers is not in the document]
0xnm: This declaration is a bit strange, because what should through NPE is the following one (notice the nullable type as an argument):

```
"java.util.concurrent.ConcurrentHashMap.remove(kotlin.String?):java.lang.NullPointerException"
```

while `"java.util.concurrent.ConcurrentHashMap.remove(kotlin.String)` cannot throw (because of non-nullable type).

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,136 @@
...
+
+        dispatcher.endMetric(fakeSessionId1)
+        dispatcher.endMetric(fakeSessionId2)
+
+        // Then
+        Assertions.assertNotNull(fakeInternalLogger.errorLog)
+    }
+
+    @Test
+    fun `M view tracked on given session W track view with session id`(
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        // When
+        dispatcher.onSessionStopped(fakeSessionId1)
+        dispatcher.endMetric(fakeSessionId1)
+
+        // Then
+        assertEquals(true, fakeInternalLogger.isSessionStopped())
+    }
+
+    @Test
+    fun `M log error W track view in different session`(
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        // Given
+        val dispatcher = SessionEndedMetricDispatcher(fakeInternalLogger)
+        dispatcher.startMetric(fakeSessionId1, fakeStartReason)
+
+        // When
+        dispatcher.onSessionStopped(fakeSessionId1)
+        dispatcher.endMetric(fakeSessionId1)
+
+        // Then
+        assertEquals(true, fakeInternalLogger.isSessionStopped())
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcherTest.kt
ambushwork: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,431 @@
...
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_BG_KEY]).isEqualTo(backgroundViewCount)
+    }
+
+    @Test
+    fun `M view counts app launch correct W track views`(
+        @Forgery events: List<ViewEvent>,
+        @IntForgery(min = 1, max = 10) appLaunchViewCount: Int
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -0,0 +1,431 @@
...
+
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes()
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val viewCounts = rse[SessionEndedMetric.VIEW_COUNTS_KEY] as Map<*, *>
+        assertThat(viewCounts[SessionEndedMetric.VIEW_COUNTS_TOTAL_KEY]).isEqualTo(events.count())
+    }
+
+    @Test
+    fun `M view counts background correct W track views`(
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        dispatcher.endMetric(fakeSessionId1)
+        assertEquals(0, fakeInternalLogger.getViewCounts())
+    }
+
+    private fun FakeInternalLogger.isSessionStopped(): Boolean {
+        return lastMetric?.second?.let { attributes ->
+            val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+            rse[SessionEndedMetric.WAS_STOPPED_KEY] as Boolean
+        } ?: false
+    }
```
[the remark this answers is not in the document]
ambushwork: I updated with the `Boolean?` return type, so if the attribute is absent, the assertions will fail

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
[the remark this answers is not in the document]
ambushwork: I fixed in detekt_custom.yml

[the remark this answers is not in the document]
0xnm: I don't see any change here or in `detekt_custom.yml` regarding that. Probably you didn't push a commit?

0xnm (dismissed):

ambushwork:
@0xnm indeed I forgot to push the commit, now it should be good

0xnm (approved):

PR #2085, opened 2024-06-14
Update OtelTraceProvider.Builder sample rate docs

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt
+++ b/features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt
@@ -226,7 +226,8 @@ class OtelTracerProvider internal constructor(
         }
 
         /**
-         * Sets the sample rate of spans.
+         * Sets the sample rate of spans. Please have in mind that there is an automatic limit of 100
+         * spans per second.
          * @param sampleRate the sample rate as a percentage between 0 and 100 (default is 100%)
          */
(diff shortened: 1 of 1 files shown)


features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt
```diff
@@ -226,7 +226,8 @@ class OtelTracerProvider internal constructor(
         }
 
         /**
-         * Sets the sample rate of spans.
+         * Sets the sample rate of spans. Please have in mind that there is an automatic limit of 100
```
0xnm, line 229: so is it a change in the behavior compared to the previous tracing code version. If yes, shouldn't we disable this behavior by default to align with what we current have?

mariusc83: actually I need to re - do this...we have a new config option (DD_TRACE_RATE_LIMIT) that is modifying that behaviour. I need to investigate on that.

mariusc83: In case we disable it what will happen when we adopt the new `artifact` from them ? Can we still disable it ? Maybe we should keep it and align with that, I need to still see what is this `DD_TRACE_RATE_LIMIT` which is different than the `DD_TRACE_SAMPLE_RATE`

PR #2087, opened 2024-06-17
RUM-4879: Granular telemetry sampling for mappers

jonathanmos:
### What does this PR do?
Have more granular telemetry sampling for mappers, using three levels - default, reduced and rare, with each level reducing the number of events by a factor of 10.

### Motivation
At the moment the default sampling rate is producing far too many events for some mappers.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
@@ -13,9 +13,14 @@ import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.feature.measureMethodCallPerf
 import com.datadog.android.sessionreplay.MapperTypeWrapper
 import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueRefs
+import com.datadog.android.sessionreplay.internal.recorder.mapper.ButtonMapper
+import com.datadog.android.sessionreplay.internal.recorder.mapper.DecorViewMapper
+import com.datadog.android.sessionreplay.internal.recorder.mapper.ImageViewMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.QueueStatusCallback
+import com.datadog.android.sessionreplay.internal.recorder.mapper.ViewWireframeMapper
 import com.datadog.android.sessionreplay.model.MobileSegment
 import com.datadog.android.sessionreplay.recorder.MappingContext
+import com.datadog.android.sessionreplay.recorder.mapper.TextViewMapper
 import com.datadog.android.sessionreplay.recorder.mapper.TraverseAllChildrenMapper
 import com.datadog.android.sessionreplay.recorder.mapper.WireframeMapper
 import com.datadog.android.sessionreplay.utils.AsyncJobStatusCallback
@@ -86,7 +91,7 @@ internal class TreeViewTraversal(
         val resolvedWireframes = internalLogger.measureMethodCallPerf(
             javaClass,
             "$METHOD_CALL_MAP_PREFIX ${mapper.javaClass.simpleName}",
-            METHOD_CALL_SAMPLING_RATE
+            getSamplingRateForMapper(mapper.javaClass.simpleName)
         ) {
             mapper.map(view, mappingContext, jobStatusCallback, internalLogger)
         }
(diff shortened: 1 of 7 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
```diff
@@ -104,13 +109,29 @@ internal class TreeViewTraversal(
...
 
-    companion object {
-        const val METHOD_CALL_SAMPLING_RATE = 0.1f
-        private const val METHOD_CALL_MAP_PREFIX: String = "Map with"
+    internal companion object {
+        internal enum class MethodCallSamplingRate(val rate: Float) {
+            DEFAULT(rate = 0.1f),
+            REDUCED(rate = 0.01f),
+            RARE(rate = 0.001f)
+        }
```
0xnm, line 133: you can move this class one level up, it can be nested inside `TreeViewTraversal`, not inside `TreeViewTraversal.Companion`

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
```diff
@@ -104,13 +109,29 @@ internal class TreeViewTraversal(
...
     )
 
-    companion object {
-        const val METHOD_CALL_SAMPLING_RATE = 0.1f
-        private const val METHOD_CALL_MAP_PREFIX: String = "Map with"
+    internal enum class MethodCallSamplingRate(val rate: Float) {
+        DEFAULT(rate = 0.1f),
+        REDUCED(rate = 0.01f),
+        RARE(rate = 0.001f)
+    }
```
xgouchet, line 132: > [!NOTE]
> Woudln't it make sense to keep those name and values shared in the `MethodCalledTelemetry` class itself, so that it could be reused in other places as well?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
```diff
@@ -86,7 +91,7 @@ internal class TreeViewTraversal(
         val resolvedWireframes = internalLogger.measureMethodCallPerf(
             javaClass,
             "$METHOD_CALL_MAP_PREFIX ${mapper.javaClass.simpleName}",
-            METHOD_CALL_SAMPLING_RATE
+            getSamplingRateForMapper(mapper.javaClass.simpleName)
```
xgouchet, line 94: > [!WARNING]
> Using a different sampling rate per mapper means that it'll skew our estimation of which ones are more or less used. Given that those mappers will be called a lot anyway, I'd be in favor of using the lowest sampling rate for all of them, to be able to track also which ones are more used throughout our customers' apps.

xgouchet:
> [!IMPORTANT]
> Shouldn't we target this to the latest release branch instead of develop, to publish an hotfix? Or are we planning a new release soon?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
jonathanmos: If we don't need different sample rates per mapper then the whole pr becomes much simpler. We need only to change the sample rate for all of "Map With" telemetry to 0.001f

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
jonathanmos: MethodCalledTelemetry is internal to the core and I don't want to modify the visibility for this, but I've extracted the enum to its own file

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
jonathanmos: 👍  I've changed to have all mappers use the lowest rate

xgouchet (approved):

PR #2090, opened 2024-06-19
RUM-1662: Integration of SessionEndedMetric into sdk core

ambushwork:
### What does this PR do?

* Integration of `SessionEndedMetricDispatcher` into core of SDK
* Add integration tests to test "was_stopped" and view counts can be calculate correctly

### Motivation

https://datadoghq.atlassian.net/browse/RUM-1662

### Additional Notes

* the way of tracking sdk error is modified, the session Id retrieved inside TelmetryEventHandler is always `NULL_UUID`, so now it relies on the last session id in `SessionEndedMetricDispatcher`
* `StubInternalLogger` is updated to send the event when `logMetric` is called, so that in integration test the event can be intercepted.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
@@ -14,6 +14,7 @@ import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.rum.RumSessionListener
 import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.metric.SessionMetricDispatcher
 import com.datadog.android.rum.internal.utils.percent
 import com.datadog.android.rum.internal.vitals.VitalMonitor
 import java.security.SecureRandom
@@ -25,6 +26,7 @@ import java.util.concurrent.atomic.AtomicLong
 internal class RumSessionScope(
     private val parentScope: RumScope,
     private val sdkCore: InternalSdkCore,
+    private val sessionEndedMetricDispatcher: SessionMetricDispatcher,
     internal val sampleRate: Float,
     internal val backgroundTrackingEnabled: Boolean,
     internal val trackFrustrations: Boolean,
@@ -55,6 +57,7 @@ internal class RumSessionScope(
     internal var childScope: RumScope? = RumViewManagerScope(
         this,
         sdkCore,
+        sessionEndedMetricDispatcher,
         backgroundTrackingEnabled,
         trackFrustrations,
         viewChangedListener,
@@ -149,6 +152,7 @@ internal class RumSessionScope(
 
     private fun stopSession() {
         isActive = false
+        sessionEndedMetricDispatcher.onSessionStopped(sessionId)
     }
 
     private fun isSessionComplete(): Boolean {
@@ -169,6 +173,11 @@ internal class RumSessionScope(
         val isBackgroundEvent = event.javaClass in RumViewManagerScope.validBackgroundEventTypes
         val isSdkInitInForeground = event is RumRawEvent.SdkInit && event.isAppInForeground
         val isSdkInitInBackground = event is RumRawEvent.SdkInit && !event.isAppInForeground
+
+        // When the session is expired, time-out or stopSession API is called, session ended metric should be sent
+        if (isExpired || isTimedOut || isActive.not()) {
+            sessionEndedMetricDispatcher.endMetric(sessionId)
+        }
 
         if (isInteraction || isSdkInitInForeground) {
             if (isNewSession || isExpired || isTimedOut) {
@@ -203,6 +212,7 @@ internal class RumSessionScope(
         sessionId = UUID.randomUUID().toString()
         sessionStartNs.set(nanoTime)
         sessionListener?.onSessionStarted(sessionId, !keepSession)
+        sessionEndedMetricDispatcher.startMetric(sessionId, reason)
     }
 
     private fun updateSessionStateForSessionReplay(state: State, sessionId: String) {
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
@@ -50,8 +50,9 @@ internal class SessionEndedMetric(
         return true
     }
 
-    fun onErrorTracked(sdkErrorKind: String) {
-        errorKindFrequencies[sdkErrorKind] = (errorKindFrequencies[sdkErrorKind] ?: 0) + 1
+    fun onErrorTracked(sdkErrorKind: String?) {
+        val errorKindKey = sdkErrorKind ?: SDK_ERROR_DEFAULT_KIND
+        errorKindFrequencies[errorKindKey] = (errorKindFrequencies[errorKindKey] ?: 0) + 1
     }
 
     fun onSessionStopped() {
@@ -202,6 +203,11 @@ internal class SessionEndedMetric(
          * Key of TOP [TOP_ERROR_LIMIT] error kinds to the number of their occurrences in the session.
          */
         internal const val SDK_ERRORS_COUNT_BY_KIND_KEY = "by_kind"
+
+        /**
+         * Placeholder of error kind if the attribute is absent.
+         */
+        internal const val SDK_ERROR_DEFAULT_KIND = "Empty error kind"
     }
 
     /**
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
@@ -52,9 +52,15 @@ internal class SessionEndedMetricDispatcher(private val internalLogger: Internal
     }
 
     override fun onSdkErrorTracked(sessionId: String, errorKind: String?) {
-        errorKind?.let {
-            metricsBySessionId[sessionId]?.onErrorTracked(it)
-        }
+        metricsBySessionId[sessionId]?.onErrorTracked(errorKind) ?: internalLogger.log(
+            level = InternalLogger.Level.INFO,
+            target = InternalLogger.Target.MAINTAINER,
+            { buildSdkErrorTrackError(sessionId, errorKind) }
+        )
+    }
+
+    private fun buildSdkErrorTrackError(sessionId: String, errorKind: String?): String {
+        return "Failed to track $errorKind error, session $sessionId has ended"
     }
 
     private fun buildViewTrackError(sessionId: String, viewEvent: ViewEvent): String {
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
@@ -32,6 +32,7 @@ import com.datadog.android.rum.internal.RumErrorSourceType
 import com.datadog.android.rum.internal.anr.ANRException
 import com.datadog.android.rum.internal.domain.RumContext
 import com.datadog.android.rum.internal.domain.Time
+import com.datadog.android.rum.internal.metric.SessionMetricDispatcher
 import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
 import com.datadog.android.rum.internal.monitor.StorageEvent
 import com.datadog.android.rum.internal.vitals.VitalInfo
@@ -187,6 +188,9 @@ internal class RumViewScopeTest {
     @Mock
     lateinit var mockViewChangedListener: RumViewChangedListener
 
+    @Mock
+    private lateinit var mockSessionEndedMetricDispatcher: SessionMetricDispatcher
+
     @BoolForgery
     var fakeTrackFrustrations: Boolean = true
 
@@ -268,6 +272,7 @@ internal class RumViewScopeTest {
         testedScope = RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             fakeKey,
             fakeEventTime,
             fakeAttributes,
@@ -368,6 +373,7 @@ internal class RumViewScopeTest {
         testedScope = RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             fakeKey,
             fakeEventTime,
             fakeAttributes,
@@ -566,6 +572,7 @@ internal class RumViewScopeTest {
         val anotherScope = RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             fakeKey,
             fakeEventTime,
             fakeAttributes,
@@ -751,6 +758,7 @@ internal class RumViewScopeTest {
         RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             key,
             fakeEventTime,
             fakeAttributes,
@@ -805,6 +813,7 @@ internal class RumViewScopeTest {
         RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             key,
             fakeEventTime,
             fakeAttributes,
@@ -1385,6 +1394,7 @@ internal class RumViewScopeTest {
         testedScope = RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             fakeKey,
             fakeEventTime,
             fakeAttributes,
@@ -1478,6 +1488,7 @@ internal class RumViewScopeTest {
         testedScope = RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             fakeKey,
             fakeEventTime,
             fakeAttributes,
@@ -1568,6 +1579,7 @@ internal class RumViewScopeTest {
         testedScope = RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             fakeKey,
             fakeEventTime,
             fakeAttributes,
@@ -1664,6 +1676,7 @@ internal class RumViewScopeTest {
         testedScope = RumViewScope(
             mockParentScope,
             rumMonitor.mockSdkCore,
+            mockSessionEndedMetricDispatcher,
             fakeKey,
             fakeEventTime,
             fakeAttributes,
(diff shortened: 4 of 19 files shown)


integrations/dd-sdk-android-okhttp-otel/transitiveDependencies
0xnm, line 1: change in this file needs to be reverted, OpenTelemetry is presented in the runtime configuration

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -50,8 +50,9 @@ internal class SessionEndedMetric(
         return true
     }
 
-    fun onErrorTracked(sdkErrorKind: String) {
-        errorKindFrequencies[sdkErrorKind] = (errorKindFrequencies[sdkErrorKind] ?: 0) + 1
+    fun onErrorTracked(sdkErrorKind: String?) {
+        val errorKindKey = sdkErrorKind ?: SDK_ERROR_DEFAULT_KIND
+        errorKindFrequencies[errorKindKey] = (errorKindFrequencies[sdkErrorKind] ?: 0) + 1
```
0xnm, line 55: tiny mistake here, I believe

```suggestion
        errorKindFrequencies[errorKindKey] = (errorKindFrequencies[errorKindKey] ?: 0) + 1
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -202,6 +203,11 @@ internal class SessionEndedMetric(
          * Key of TOP [TOP_ERROR_LIMIT] error kinds to the number of their occurrences in the session.
          */
         internal const val SDK_ERRORS_COUNT_BY_KIND_KEY = "by_kind"
+
+        /**
+         * Place holder of error kind if the attribute is absent.
```
0xnm, line 208:
```suggestion
         * Placeholder of error kind if the attribute is absent.
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
0xnm, line 1: should we have some verification added here to check that `onViewTracked` is called (same for `RumSessionScopeTest`)? Or you want to rely solely on the integration tests?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -51,12 +53,32 @@ internal class SessionEndedMetricDispatcher(private val internalLogger: Internal
         }
     }
 
-    override fun onSdkErrorTracked(sessionId: String, errorKind: String?) {
```
0xnm, line 54: cannot we have a situation when we stopped session A and started session B, but due to async nature of different pipelines in the SDK, we have this method called with error coming from session A by the time `startMetric` was called with Session B? not sure what is the probability of that though.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -17,11 +17,13 @@ internal class SessionEndedMetricDispatcher(private val internalLogger: Internal
 
     private val metricsBySessionId = ConcurrentHashMap<String, SessionEndedMetric>()
 
+    private var lastSessionId: String? = null
```
0xnm, line 20: this needs to be `AtomicReference` then, following thread-safe design of this class.

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
```diff
@@ -0,0 +1,125 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+
+@Suppress("DANGEROUS_CHARACTERS")
```
0xnm, line 33: It is better to avoid such characters rather than try to suppress it. This warning exists on purpose.

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    fun `set up`(forge: Forge) {
+        stubSdkCore = StubSDKCore(forge)
+        fakeRumConfiguration = RumConfiguration.Builder(fakeApplicationId)
+            .trackNonFatalAnrs(false)
+            .setTelemetrySampleRate(100f)
+            .build()
+    }
+
+    @Test
+    fun `M receive an event with "was_stopped" is true W stopSession() is called`(
```
0xnm, line 60: justification: `W methodCall()` already implies we are doing method call

```suggestion
    fun `M receive an event with "was_stopped" is true W stopSession()`(
```

0xnm (changes requested):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
ambushwork: I am not sure the probability of this neither, but with this pull request, the error coming from Session A will belongs to Session B since `startMetric` has called. I am not sure if we have anyway to improve this, since `endMetric` of Session A has been called, the telemetry is sent out, the error will not be able to catch the train which has left.

0xnm: In theory we can have a skew in the metric data we receive in such case, so maybe it is better to rollback this change and still pass `sessionId` to this method from the `DatadogContext` we have in the `TelemetryEventHandler`?

ambushwork: I have two questions over this:
1. even if we use session Id to track the error, the error count in session B can be correct, but in the session A will always be less than the correct one since the metric can be sent out, is that something we can tolerate?
2. I tried to use `sdkCore.getFeature(Feature.RUM_FEATURE_NAME)?.withWriteContext` that you suggested to retrieve the session id for the error, I got always NULL_UUID, is that normal?

0xnm: 1. Don't know honestly, depends on the further metric usage and if we can have falsy data because of that.
2. You don't need to use it directly, you can hook inside the lambda in the `TelemetryEventHandler` or you can use  something like `RumContext.fromFeatureContext(sdkCore.getDatadogContext().featuresContext[Feature.RUM_FEATURE_NAME])`. `NULL_UUID` means that session is not yet created.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
ambushwork: imo, `M receive an event with correct view counts W track multiple views` in the integration test can make sure that `onViewTracked` is called, otherwise it won't have the correct view count

integrations/dd-sdk-android-okhttp-otel/transitiveDependencies
ambushwork: reverted

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
ambushwork: This is removed, we don't need to record the last session id

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetricDispatcher.kt
```diff
@@ -51,12 +53,32 @@ internal class SessionEndedMetricDispatcher(private val internalLogger: Internal
...
+                    { buildSdkErrorTrackError(it, errorKind) }
+                )
+            }
+        } ?: run {
+            internalLogger.log(
+                level = InternalLogger.Level.INFO,
+                target = InternalLogger.Target.MAINTAINER,
+                { buildSdkErrorLastSessionNullError(errorKind) }
+            )
         }
```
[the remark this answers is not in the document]
ambushwork: fixed

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
ambushwork: fixed

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
ambushwork: fixed

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    fun `M receive an event with correct view counts W tracks multiple views`(
+        @IntForgery(min = 1, max = 5) repeatCount: Int
+    ) {
+        // Given
+        Rum.enable(fakeRumConfiguration, stubSdkCore)
+        val rumMonitor = GlobalRumMonitor.get(stubSdkCore)
+
+        repeat(repeatCount) {
+            rumMonitor.startView(key = "key1", name = "View1")
+            rumMonitor.stopView(key = "key1")
```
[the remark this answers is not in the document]
ambushwork: good point, fixed

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
```diff
@@ -0,0 +1,125 @@
...
+        map.entries.forEach {
+            val entryValue = it.value
+            if (entryValue is Map<*, *>) {
+                return checkAttributeWithValue(entryValue, actualKey, actualValue)
+            } else if (it.key == actualKey) {
+                return it.value == actualValue
+            }
+        }
+        return false
+    }
```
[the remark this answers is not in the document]
ambushwork: you are right, it's fixed

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
```diff
@@ -44,6 +46,16 @@ internal class StubInternalLogger : InternalLogger {
...
+
+        val rumFeature = sdkCore.getFeature(RUM_FEATURE_NAME) ?: return
+        val message = messageBuilder()
+        val telemetryEvent =
+            mapOf(
+                "type" to "mobile_metric",
+                "message" to message,
+                "additionalProperties" to additionalProperties
+            )
+        rumFeature.sendEvent(telemetryEvent)
```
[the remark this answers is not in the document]
ambushwork: now I keep a `lastMetric` inside `StubInternalLogger` to store the telemetry event

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
ambushwork: Updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -149,6 +152,7 @@ internal class RumSessionScope(
 
     private fun stopSession() {
         isActive = false
+        sessionEndedMetricDispatcher.onSessionStopped(sessionId)
```
0xnm, line 155: so we are going to call `onSessionStopped` only if `RumMonitor.stopSession` was called and we are not going to call it when session is stopped naturally (because of the expiration)?

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        Rum.enable(fakeRumConfiguration, stubSdkCore)
+        val rumMonitor = GlobalRumMonitor.get(stubSdkCore)
+
+        rumMonitor.startView(key = viewKey, name = viewName)
+
+        // When
+        rumMonitor.stopSession()
+
+        // Then
+        RumEventAssert.assertThat(stubSdkCore.lastMetric())
```
0xnm, line 72: suggestion: you can also use static import for `RumEventAssert.assertThat` here.

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/tests/assertj/RumEventAssert.kt
```diff
@@ -121,11 +122,25 @@ class RumEventAssert(actual: JsonObject) :
...
     // endregion
 
     companion object {
         fun assertThat(actual: JsonObject): RumEventAssert {
             return RumEventAssert(actual)
         }
+
+        fun assertThat(actual: Any): RumEventAssert {
+            return assertThat(Gson().toJsonTree(actual).asJsonObject)
+        }
```
0xnm, line 144: IMO this looks like a hack.

Ideally, I don't think it should be `Any` argument type here, and probably this shouldn't be inside `RumEventAssert`. `RumEventAssert` is used for the RUM events only (the ones which are going to be sent to the intake), which are RUM View, RUM Error, etc. events.

In your case you have argument which is not a RUM event (and obviously none of the RUM events can have "wasStopped" or "hasViewCount" attribute). This approach works, but it is mixing the things. Maybe it is better rather to create another assertion class (say for `Map<String, Any>` or even `Map<*,*>`) and since you are not operating with JSON you can write something like that there (let's assume `actual` is of type `Map<String, Any>)`:

```
fun hasWasStopped(wasStopped: Boolean): TelemetryMetricAssert {
        assertThat(actual).containsPathEntry(path, value)
        return this
    }
```

and then you just need to implement `containsPathEntry` which will take the path like `additionalProperties.rse.views_count.total` in a similar manner how it was done for json (just split in the fragments and iterate over).

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
```diff
@@ -12,6 +12,8 @@ import com.datadog.android.core.metrics.TelemetryMetricType
 
 @Suppress("UnsafeThirdPartyFunctionCall")
 internal class StubInternalLogger : InternalLogger {
+
+    var lastMetric: Map<String, Any> = hashMapOf()
     override fun log(
```
0xnm, line 17: following approach of capturing all the data in the test, it is probably more useful to record all the events (and then access the necessary using `last()` or `findLast`, etc.), also making this property immutable

```suggestion
    val telemetryEventsWritten = mutableListOf<Map<String, Any>>()

    override fun log(
```

0xnm (changes requested):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
ambushwork: the purpose here is to register `was_stopped` in the metric when `stopSession` API is called by user, the metric will not be ended due to this. The metric is still ended when session has expired, timeout or stopped.

reliability/single-fit/rum/src/test/kotlin/com/datadog/android/rum/integration/RumSessionEndedIntegrationTelemetryTest.kt
ambushwork: fixed

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
ambushwork: updated

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubSDKCore.kt
```diff
@@ -72,6 +72,13 @@ class StubSDKCore(
         datadogContext = datadogContext.copy(networkInfo = networkInfo)
     }
 
+    /**
+     * Returns the last metric is sent by [StubInternalLogger].
+     */
+    fun lastMetric(): Map<String, Any> {
+        return (internalLogger as StubInternalLogger).telemetryEventsWritten.last()
```
0xnm, line 79: to be precise, you need to use `filter { it["type"] == "mobile_metric" }` here, because property itself is called `telemetryEventsWritten`, so if someone else adds there non-metric telemetry, `telemetryEventsWritten.last()` will give a wrong result and things won't be correct.

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
```diff
@@ -12,6 +12,8 @@ import com.datadog.android.core.metrics.TelemetryMetricType
 
 @Suppress("UnsafeThirdPartyFunctionCall")
 internal class StubInternalLogger : InternalLogger {
+
+    var telemetryEventsWritten = mutableListOf<Map<String, Any>>()
```
0xnm, line 16: no need to be mutable reference.

```suggestion
    val telemetryEventsWritten = mutableListOf<Map<String, Any>>()
```

0xnm (approved):

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
ambushwork: fixed

reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubSDKCore.kt
ambushwork: Make sense, I use `lastOrNull{ it["type"] == "mobile_metric"}` to return the last telemetry metric recorded.

0xnm (approved):

PR #2096, opened 2024-06-20
RUM-4098: Add api to clear all datastore data

jonathanmos:
### What does this PR do?
Adds an api to delete all saved datastore entries. 

### Motivation
The core's clearAllData api should return the sdk to a fresh state and therefore requires removal of datastore information.

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DatastoreFileWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DatastoreFileWriter.kt
@@ -88,13 +88,14 @@ internal class DatastoreFileWriter(
 
     @WorkerThread
     internal fun clearAllData() {
-        val dataStoreDirectory = dataStoreFileHelper.createDataStoreDirectoryIfNecessary(
+        val dataStoreDirectory = dataStoreFileHelper.getDataStoreDirectory(
             featureName = featureName,
-            storageDir = storageDir,
-            internalLogger = internalLogger
+            storageDir = storageDir
         )
 
-        dataStoreDirectory.deleteDirectoryContentsSafe(internalLogger)
+        if (dataStoreDirectory.existsSafe(internalLogger)) {
+            dataStoreDirectory.deleteDirectoryContentsSafe(internalLogger)
+        }
     }
 
     private fun <T : Any> getDataBlock(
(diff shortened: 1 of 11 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DatastoreFileWriter.kt
```diff
@@ -71,6 +72,17 @@ internal class DatastoreFileWriter(
         }
     }
 
+    @WorkerThread
+    internal fun clearAllData() {
+        val dataStoreDirectory = dataStoreFileHelper.createDataStoreDirectoryIfNecessary(
```
0xnm, line 77: this is a bit confusing: why do we need to create a directory (if it is not present) just to make it empty? Also `deleteDirectoryContentsSafe` will make `number_of_files` I/O calls, while we can just do 1 I/O call - delete the directory itself.

Maybe it is better just to get a `File` object pointing to the directory (just an object, without creating directory), check if it exists and then just call `delete` if it is the case?

jonathanmos: Are you sure that it's possible in one call? The documentation for `delete` indicates that the directory [must be empty](https://developer.android.com/reference/java/io/File#delete()) to be deleted

mariusc83: if it doesn't work I would just create a method there that checks if a dir exists by path and it deletes all the files inside recursively. As @0xnm mentions we should not create a dir just to be able to call that method.

mariusc83 (changes requested):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DatastoreFileWriter.kt
jonathanmos: I was just reusing an existing method that does both. I'll split it, so that one method just returns the path and another also creates if necessary and then we won't have an issue with this.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/datastore/DatastoreFileWriter.kt
```diff
@@ -85,6 +86,18 @@ internal class DatastoreFileWriter(
...
 
+    @WorkerThread
+    internal fun clearAllData() {
+        val dataStoreDirectory = dataStoreFileHelper.getDataStoreDirectory(
+            featureName = featureName,
+            storageDir = storageDir
+        )
+
+        if (dataStoreDirectory.existsSafe(internalLogger)) {
+            dataStoreDirectory.deleteDirectoryContentsSafe(internalLogger)
```
0xnm, line 97: there is no need to introduce `deleteDirectoryContentsSafe` method, you can simply do `dataStoreDirectory.deleteSafe` here (one call to delete the whole directory instead of many calls to delete each file).

```suggestion
            dataStoreDirectory.deleteSafe(internalLogger)
```

Is there any reason to delete individual files instead of the whole directory?

0xnm (changes requested):

0xnm (approved):

xgouchet (approved):

mariusc83 (approved):

PR #2098, opened 2024-06-21
RUM-4822 Allow adding Interceptor in internal OkHttpClient

xgouchet:
### What does this PR do?

Let customers add Interceptor (e.g.: `DatadogInterceptor`) to monitor the internal OkHttp Client used to upload batches to Datadog.

> [!NOTE]
> This will be part of a debug version of the SDK to help our customer regarding Incident 27794, and won't be merged into develop in the current state.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -42,7 +43,8 @@ internal constructor(
         val site: DatadogSite,
         val batchProcessingLevel: BatchProcessingLevel,
         val persistenceStrategyFactory: PersistenceStrategy.Factory?,
-        val backpressureStrategy: BackPressureStrategy
+        val backpressureStrategy: BackPressureStrategy,
+        val additionalInterceptor: List<Interceptor>
```
0xnm, line 47:
```suggestion
        val additionalInterceptors: List<Interceptor>
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
```diff
@@ -259,6 +261,18 @@ internal constructor(
             return this
         }
 
+        /**
+         * Add a network interceptor to the internal OkHttpClient used to upload tracked events to Datadog's servers.
+         * @param interceptor the interceptor to add
+         * Note you can call this method more than once if you need to add several interceptors
+         */
+        fun addNetworkInterceptor(interceptor: Interceptor): Builder {
```
mariusc83, line 269: why not instead allowing a list of interceptors here ?

0xnm (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt
xgouchet: Mostly because it mimicks the API used by OkHttp, where you add interceptors one at a time. 
Plus this is just for one customer, and we only expect them to submit a single interceptor, so it doesn't make sense to force them to create a list to do so.

0xnm (approved):

mariusc83 (approved):

PR #2100, opened 2024-06-21
Fix the Okhttp Otel parent span feature when not using RUM

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/integrations/dd-sdk-android-okhttp-otel/src/main/kotlin/com/datadog/android/okhttp/otel/OkHttpExt.kt
+++ b/integrations/dd-sdk-android-okhttp-otel/src/main/kotlin/com/datadog/android/okhttp/otel/OkHttpExt.kt
@@ -7,7 +7,7 @@
 package com.datadog.android.okhttp.otel
 
 import com.datadog.android.okhttp.TraceContext
-import com.datadog.trace.api.sampling.PrioritySampling
+import com.datadog.legacy.trace.api.sampling.PrioritySampling
 import io.opentelemetry.api.trace.Span
 import okhttp3.Request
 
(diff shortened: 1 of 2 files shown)


integrations/dd-sdk-android-okhttp-otel/src/main/kotlin/com/datadog/android/okhttp/otel/OkHttpExt.kt
```diff
@@ -18,7 +18,9 @@ import okhttp3.Request
  */
 fun Request.Builder.addParentSpan(span: Span): Request.Builder {
     val context = span.spanContext
-    val prioritySampling = if (context.isSampled) PrioritySampling.USER_KEEP.toInt() else PrioritySampling.UNSET.toInt()
+    // because the DatadogTracer is using the old legacy DDTracer for handling the spans around request
```
0xnm, line 21: what is the `DatadogTracer`? I cannot find such symbol in the code

mariusc83: yes because it should be `DatadogInterceptor` there ;)

integrations/dd-sdk-android-okhttp-otel/src/main/kotlin/com/datadog/android/okhttp/otel/OkHttpExt.kt
```diff
@@ -18,7 +18,9 @@ import okhttp3.Request
  */
 fun Request.Builder.addParentSpan(span: Span): Request.Builder {
     val context = span.spanContext
-    val prioritySampling = if (context.isSampled) PrioritySampling.USER_KEEP.toInt() else PrioritySampling.UNSET.toInt()
+    // because the DatadogInterceptor is using the old legacy DDTracer for handling the spans around request
+    // we need to provide the legacy sampling codes here.
+    val prioritySampling = if (context.isSampled) PrioritySampling.USER_KEEP else PrioritySampling.UNSET
```
0xnm, line 23: so the difference is only in the `UNSET` value? `0x80` vs `Integer.MIN_VALUE`. Do we have a possibility of such mismatch in other parts of the code?

0xnm (approved):

integrations/dd-sdk-android-okhttp-otel/src/main/kotlin/com/datadog/android/okhttp/otel/OkHttpExt.kt
mariusc83: yes that was the problem, we do not have any other places this is just because we are using Otel here and on the other side we are using OpenTracing. We really need to get rid of that ASAP and switch everything to CoreTracer

xgouchet (approved):

PR #2104, opened 2024-06-21
RUM-4822 Prepare debug build

xgouchet:
### What does this PR do?

Prepare a "debug" build specifically for this incident 

> [!NOTE]
> This will be part of a debug version of the SDK to help our customer regarding Incident 27794, and won't be merged into develop in the current state.

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
@@ -49,7 +49,7 @@ fun Project.publishingConfig(projectDescription: String) {
 
                 groupId = MavenConfig.GROUP_ID
                 artifactId = projectName
-                version = AndroidConfig.VERSION.name
+                version = AndroidConfig.VERSION.name + "-debug"
 
                 pom {
                     name.set(projectName)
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/com/datadog/gradle/config/MavenConfig.kt
```diff
@@ -49,7 +49,7 @@ fun Project.publishingConfig(projectDescription: String) {
 
                 groupId = MavenConfig.GROUP_ID
                 artifactId = projectName
-                version = AndroidConfig.VERSION.name
+                version = AndroidConfig.VERSION.name + "-debug"
```
0xnm, line 52: I'm afraid that if we publish it, it will be picked by the Gradle still following the ordering rule https://docs.gradle.org/current/userguide/single_versions.html#version_ordering in case of the dynamic versioning use. And given that it has public API change in the `startResource` method, this can be breaking.

How we are going to distribute/publish this version?

xgouchet: Actually, according to that documentation 
> A version with an extra non-numeric part is considered lower than a version without: 1.1.a < 1.1

Following that logic, a version `2.11.0-debug` will be considered lower that the existing `2.11.0`, so for the (I'm guessing) few customers who might use dynamic versioning, this still won't be an issue.

0xnm: ok, we are safe on that side then.

0xnm (approved):

jonathanmos (approved):

0xnm (approved):

PR #2108, opened 2024-06-25
Add sampling rate to internal metrics

xgouchet:
### What does this PR do?

- [x] Add sampling rate option for internal metrics
- [x] Reduce the amount of "Gap between view" telemetry, and discarding gaps above 3 seconds (which most probably denote an app being sent to background instead of an actual gap while navigating inside the tracked app).

> [!NOTE]
> This metric will be applied in addition to the telemetry sampling rate, so if a logMetric is created with a sampling rate of 10%, with the global telemetry sampling rate being sampled at 20%, it has a 2% chance of being kept.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
@@ -105,9 +105,15 @@ interface InternalLogger {
      * [com.datadog.android.telemetry.model.TelemetryDebugEvent.Telemetry] event.
      * @param messageBuilder the lambda building the metric message
      * @param additionalProperties additional properties to add to the metric
+     * @param samplingRate value between 0-100 for sampling the event. Note that the sampling rate applied to this
+     * metric will be applied in addition to the global telemetry sampling rate.
      */
     @InternalApi
-    fun logMetric(messageBuilder: () -> String, additionalProperties: Map<String, Any?>)
+    fun logMetric(
+        messageBuilder: () -> String,
+        additionalProperties: Map<String, Any?>,
+        samplingRate: Float
+    )
 
     /**
      * Start measuring a performance metric.
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/metrics/MethodCallSamplingRate.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/metrics/MethodCallSamplingRate.kt
@@ -8,9 +8,12 @@ package com.datadog.android.core.metrics
 
 /**
  * Sampling rates for Method Call telemetry.
- * @param rate the rate to sample at.
+ * @param rate the rate to sample at (between 0 and 100).
  */
 enum class MethodCallSamplingRate(val rate: Float) {
+    ALL(rate = 100.0f),
+    HIGH(rate = 10.0f),
+    MEDIUM(rate = 1.0f),
     DEFAULT(rate = 0.1f),
     REDUCED(rate = 0.01f),
     RARE(rate = 0.001f)
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
@@ -12,6 +12,7 @@ import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.storage.DataWriter
 import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
+import com.datadog.android.core.metrics.MethodCallSamplingRate
 import com.datadog.android.rum.DdRumContentProvider
 import com.datadog.android.rum.internal.anr.ANRException
 import com.datadog.android.rum.internal.domain.RumContext
@@ -21,6 +22,7 @@ import com.datadog.android.rum.internal.metric.SessionMetricDispatcher
 import com.datadog.android.rum.internal.vitals.NoOpVitalMonitor
 import com.datadog.android.rum.internal.vitals.VitalMonitor
 import java.util.Locale
+import java.util.concurrent.TimeUnit
 
 @Suppress("LongParameterList")
 internal class RumViewManagerScope(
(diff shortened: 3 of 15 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -60,11 +62,13 @@ internal class RumViewManagerScope(
...
                 val gap = event.eventTime.nanoTime - it.nanoTime
-                sdkCore.internalLogger.log(
-                    InternalLogger.Level.INFO,
-                    listOf(InternalLogger.Target.TELEMETRY, InternalLogger.Target.MAINTAINER),
-                    { MESSAGE_GAP_BETWEEN_VIEWS.format(Locale.US, gap) }
-                )
+                if (gap in 1 until THREE_SECONDS_GAP) {
+                    sdkCore.internalLogger.logMetric(
+                        messageBuilder = { MESSAGE_GAP_BETWEEN_VIEWS.format(Locale.US, gap) },
+                        additionalProperties = mapOf("view_gap" to gap),
```
ambushwork, line 68: Should we avoid hardcoded string in the code? maybe create an attribute in `companion object` for it.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
```diff
@@ -105,9 +105,15 @@ interface InternalLogger {
...
      * @param additionalProperties additional properties to add to the metric
+     * @param samplingRate value between 0-100 for sampling the event. Note that the sampling rate applied to this
+     * metric will be applied in addition to the global telemetry sampling rate.
      */
     @InternalApi
-    fun logMetric(messageBuilder: () -> String, additionalProperties: Map<String, Any?>)
+    fun logMetric(
+        messageBuilder: () -> String,
+        additionalProperties: Map<String, Any?>,
+        samplingRate: Float
```
ambushwork, line 115: Will it be a good idea to give a default (also recommended) value for the sample rate? so in the place we cannot decide which sample rate we should use, we can have at least a good fallback.

xgouchet: IMO, adding a default might lead into us not actively chosing a sampling rate. 
I'm all for providing sensible defaults on the public APIs used by our customers, but for our own internal code, I prefer forcing us to think about every param we use.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
xgouchet: Good point

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
ambushwork: Make sense to me

ambushwork (approved):
LGTM!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/metrics/MethodCallSamplingRate.kt
```diff
@@ -8,9 +8,12 @@ package com.datadog.android.core.metrics
...
 /**
  * Sampling rates for Method Call telemetry.
- * @param rate the rate to sample at.
+ * @param rate the rate to sample at (between 0 and 100).
  */
 enum class MethodCallSamplingRate(val rate: Float) {
+    ALL(rate = 100.0f),
+    HIGH(rate = 10.0f),
+    MEDIUM(rate = 1.0f),
     DEFAULT(rate = 0.1f),
```
0xnm, line 17: minor: I think there is a bit of naming clash between `MEDIUM` and `DEFAULT`. one can think they are synonyms.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -302,5 +308,7 @@ internal class RumViewManagerScope(
 
         internal const val MESSAGE_UNKNOWN_MISSED_TYPE = "An RUM event was detected, but no view is active, " +
             "its missed type is unknown"
+
+        internal val THREE_SECONDS_GAP = TimeUnit.SECONDS.toNanos(3)
```
0xnm, line 312:
```suggestion
        internal val THREE_SECONDS_GAP_NS = TimeUnit.SECONDS.toNanos(3)
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -60,11 +62,13 @@ internal class RumViewManagerScope(
             startForegroundView(event, writer)
             lastStoppedViewTime?.let {
                 val gap = event.eventTime.nanoTime - it.nanoTime
-                sdkCore.internalLogger.log(
-                    InternalLogger.Level.INFO,
-                    listOf(InternalLogger.Target.TELEMETRY, InternalLogger.Target.MAINTAINER),
-                    { MESSAGE_GAP_BETWEEN_VIEWS.format(Locale.US, gap) }
-                )
+                if (gap in 1 until THREE_SECONDS_GAP) {
```
0xnm, line 65: according to the telemetry, we can also have a negative gap by some reason. Do we want to log negative gaps as well? This is something abnormal imo.

jonathanmos (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
xgouchet: True, I'll add a specific metric for that case

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/metrics/MethodCallSamplingRate.kt
0xnm: seems like it will be in another PR? this PR was merged without this update

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
0xnm: seems like it will be in another PR? this PR was merged without this update

PR #2111, opened 2024-06-27
RUM-3536 Introduce the TraceContextInjection to handle sampling in distributed traces

mariusc83:
### What does this PR do?

Following the requirements specified in this [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3529509193/RUM+SDKs+Trace+Context+Injection+control) we are adding this new API and fixing the traces propagation in our `TracingInterceptor`.

This also fixes the issue: #2124

### Motivation

### Additional Notes

### DIFF
--- a/integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt
+++ b/integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt
@@ -88,13 +88,10 @@ constructor(
         val builder = OkHttpClient.Builder()
 
         builder.eventListenerFactory(DatadogEventListener.Factory(sdkInstanceName))
-        builder.addInterceptor(
-            DatadogInterceptor(
-                sdkInstanceName,
-                firstPartyHosts,
-                traceSampler = RateBasedSampler(sampleRate)
-            )
-        )
+        val interceptorBuilder = DatadogInterceptor.Builder(firstPartyHosts)
+            .setTraceSampler(RateBasedSampler(sampleRate))
+        sdkInstanceName?.let { interceptorBuilder.setSdkInstanceName(it) }
+        builder.addInterceptor(interceptorBuilder.build())
         return builder
     }
 
--- a/integrations/dd-sdk-android-okhttp/api/apiSurface
+++ b/integrations/dd-sdk-android-okhttp/api/apiSurface
@@ -16,9 +16,9 @@ class com.datadog.android.okhttp.DatadogEventListener : okhttp3.EventListener
     constructor(String? = null)
     override fun create(okhttp3.Call): okhttp3.EventListener
 open class com.datadog.android.okhttp.DatadogInterceptor : com.datadog.android.okhttp.trace.TracingInterceptor
-  constructor(String? = null, Map<String, Set<com.datadog.android.trace.TracingHeaderType>>, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
-  constructor(String? = null, List<String>, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
-  constructor(String? = null, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
+  DEPRECATED constructor(String? = null, Map<String, Set<com.datadog.android.trace.TracingHeaderType>>, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
+  DEPRECATED constructor(String? = null, List<String>, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
+  DEPRECATED constructor(String? = null, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
   override fun intercept(okhttp3.Interceptor.Chain): okhttp3.Response
   override fun onRequestIntercepted(com.datadog.android.api.feature.FeatureSdkCore, okhttp3.Request, io.opentracing.Span?, okhttp3.Response?, Throwable?)
   override fun canSendSpan(): Boolean
@@ -26,7 +26,6 @@ open class com.datadog.android.okhttp.DatadogInterceptor : com.datadog.android.o
   class Builder : BaseBuilder<DatadogInterceptor, Builder>
     constructor(Map<String, Set<com.datadog.android.trace.TracingHeaderType>>)
     constructor(List<String>)
-    override fun getThis(): Builder
     override fun build(): DatadogInterceptor
     fun setRumResourceAttributesProvider(com.datadog.android.rum.RumResourceAttributesProvider): Builder
 data class com.datadog.android.okhttp.TraceContext
@@ -46,7 +45,6 @@ open class com.datadog.android.okhttp.trace.TracingInterceptor : okhttp3.Interce
   class Builder : BaseBuilder<TracingInterceptor, Builder>
     constructor(Map<String, Set<com.datadog.android.trace.TracingHeaderType>>)
     constructor(List<String>)
-    override fun getThis(): Builder
     fun setTraceOrigin(String): Builder
     override fun build(): TracingInterceptor
   abstract class BaseBuilder<T: TracingInterceptor, R: BaseBuilder<T, R>>
@@ -55,5 +53,4 @@ open class com.datadog.android.okhttp.trace.TracingInterceptor : okhttp3.Interce
     fun setTracedRequestListener(TracedRequestListener): R
     fun setTraceSampler(com.datadog.android.core.sampling.Sampler): R
     fun setTraceContextInjection(com.datadog.android.okhttp.TraceContextInjection): R
-    abstract fun getThis(): R
     abstract fun build(): T
--- a/integrations/dd-sdk-android-okhttp/api/apiSurface
+++ b/integrations/dd-sdk-android-okhttp/api/apiSurface
@@ -23,14 +23,37 @@ open class com.datadog.android.okhttp.DatadogInterceptor : com.datadog.android.o
   override fun onRequestIntercepted(com.datadog.android.api.feature.FeatureSdkCore, okhttp3.Request, io.opentracing.Span?, okhttp3.Response?, Throwable?)
   override fun canSendSpan(): Boolean
   override fun onSdkInstanceReady(com.datadog.android.core.InternalSdkCore)
+  class Builder : BaseBuilder<DatadogInterceptor, Builder>
+    constructor(Map<String, Set<com.datadog.android.trace.TracingHeaderType>>)
+    constructor(List<String>)
+    override fun getThis(): Builder
+    override fun build(): DatadogInterceptor
+    fun setRumResourceAttributesProvider(com.datadog.android.rum.RumResourceAttributesProvider): Builder
 data class com.datadog.android.okhttp.TraceContext
   constructor(String, String, Int)
+enum com.datadog.android.okhttp.TraceContextInjection
+  - All
+  - Sampled
 fun okhttp3.Request.Builder.parentSpan(io.opentracing.Span): okhttp3.Request.Builder
 interface com.datadog.android.okhttp.trace.TracedRequestListener
   fun onRequestIntercepted(okhttp3.Request, io.opentracing.Span, okhttp3.Response?, Throwable?)
 open class com.datadog.android.okhttp.trace.TracingInterceptor : okhttp3.Interceptor
-  constructor(String? = null, List<String>, TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
-  constructor(String? = null, Map<String, Set<com.datadog.android.trace.TracingHeaderType>>, TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
-  constructor(String? = null, TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
+  DEPRECATED constructor(String? = null, List<String>, TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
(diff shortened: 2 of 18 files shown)


integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt
```diff
@@ -49,6 +50,10 @@ constructor(
     private val sampleRate: Float = DEFAULT_SAMPLE_RATE
 ) : AppGlideModule() {
 
+    private val firstPartyHostsWithHeaderType = firstPartyHosts.associateWith {
```
mariusc83, line 53: the only problem I see with this is that some users will not know what to use for their tracing header types values so we might want to provide a Public Constant: `DEFAULT_TYPES` here ?

integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt
```diff
@@ -88,13 +93,10 @@ constructor(
...
-        builder.addInterceptor(
-            DatadogInterceptor(
-                sdkInstanceName,
-                firstPartyHosts,
-                traceSampler = RateBasedSampler(sampleRate)
-            )
-        )
+        val interceptorBuilder = DatadogInterceptor.Builder(firstPartyHostsWithHeaderType)
+            .setTraceSampler(RateBasedSampler(sampleRate))
+        sdkInstanceName?.let { interceptorBuilder.setSdkInstanceName(it) }
```
0xnm, line 98: I would pull `sdkInstanceName` to the `DatadogInterceptor.Builder` constructor - SDK core association is a major part of the API, not something optional.

integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt
0xnm: why then we need to provide association if `setOf(TracingHeaderType.DATADOG, TracingHeaderType.TRACECONTEXT)` is what already set in the SDK by default? Shouldn't we have a public API for both `firstPartyHosts` and `firstPartyHostsWithHeaderType` as we have now?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -397,6 +417,48 @@ internal constructor(
 
     // endregion
 
+    // region Builder
+
+    /**
+     * A Builder for the [DatadogInterceptor].
+     */
+    open class Builder(tracedHosts: Map<String, Set<TracingHeaderType>>) :
```
0xnm, line 425: We probably need to remove `open` modifier from here and `DatadogInterceptor` as well. I think it was originally added because we have `RumInterceptor : DatadogInterceptor`.

But removing it involves a risk - some customers may have already sub-classed `DatadogInterceptor`.

0xnm, line 425: IMO having a `map` here is unnecessary complication and it should be just `tracedHosts: List<String>`. Asking to provide tracing header types may be too advanced and increases setup curve. Probably default header types will work for 90%+ installations.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -397,6 +417,48 @@ internal constructor(
...
+        BaseBuilder<DatadogInterceptor, Builder>(tracedHosts) {
+
+        private var rumResourceAttributesProvider: RumResourceAttributesProvider = NoOpRumResourceAttributesProvider()
+
+        override fun getThis(): Builder {
```
0xnm, line 430: `getThis`? why is it needed?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -397,6 +417,48 @@ internal constructor(
...
+                rumResourceAttributesProvider,
+                traceSampler,
+                traceContextInjection,
+                localTracerFactory
+            )
+        }
+
+        /**
+         * Sets the [RumResourceAttributesProvider] to use to provide custom attributes to the RUM.
+         * By default, it is a NoOp.
```
0xnm, line 451:
```suggestion
         * By default, it won't attach any custom attributes.
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/TraceContextInjection.kt
```diff
@@ -0,0 +1,22 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.okhttp
+
+/**
+ * Defines whether the trace context should be injected into all requests or only sampled ones.
```
0xnm, line 10: should we explain what is the trace context and why it might be useful if request is not sampled (I, personally, have no idea, tbh).

integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt
mariusc83: in the same time we have it optional now, I would only do that if we do not allow it as optional anymore otherwise it doesn't make sense to have it in the constructor if it can be NULL

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
mariusc83: I don't want to do that, at least not now.

mariusc83: it will but then we will have to provide the `setTracedHostWithHeaderTypes` method in the `Builder` and that will create some confusion. It will look like this:
```
val builder = Builder(tracedHosts).setFirstPartyHostsWithHeaderTypes(firstPartyHostsWithHeaderTypes)
```
It will look ugly and users will wonder what is the need for the `tracedHosts` in the Builder.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
mariusc83: `Generics` check the `Builder` and `BaseBuilder` class. It is just a patter for having a `BaseBuilder` using Java Generics. In any case this is `internal`

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/TraceContextInjection.kt
mariusc83: I just follow the same docs as in `iOS`. I am going to spend some time with Mael maybe we can define something better.

integrations/dd-sdk-android-glide/src/main/kotlin/com/datadog/android/glide/DatadogGlideModule.kt
mariusc83: we could but the way I see it `firstPartyHosts` is mandatory (not optional) so it will have to be in the `Builder` constructor. If we want to introduce both APIs and if a user wants to provide custom headerTypes for each host it will look like this:
```
val builder = Builder(tracedHosts).setFirstPartyHostsWithHeaderTypes(firstPartyHostsWithHeaderTypes)
```
I do not like that.

mariusc83: otherwise we could have:
```
val builder = Builder.()setFirstPartyHostsWithHeaderTypes(firstPartyHostsWithHeaderTypes)
val builder = Builder().setFirstPartyHosts()
```
But this will mean that some users will forget to set this up in the builder and will create support issues.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
0xnm: > we will have to provide the setTracedHostWithHeaderTypes method in the Builder

It won't be needed if constructor itself will have another overload. APM setup is already quite complicated, we need to have it simpler with reasonable defaults.

mariusc83: so we provide overloads to this `Builder` constructor ? I would rather not and just provide a public api `DEFAULT_HEADER_TYPES = setOf(TracingHeaderType.DATADOG, TracingHeaderType.TRACECONTEXT)`
We are trying to run away from overloads. I am opened to any of this options but I would prefer the one without overloads.
@xgouchet @plousada , what do you think ?

0xnm: > We are trying to run away from overloads
It is just because there were too much of them in the original constructor. Basically we need to put what is mandatory in the constructor and cannot be replaced by some default value.

For now it is only the list of the traced hosts, and the rest of the pieces needed to make interceptor work can have a default value, so they go into the methods of the `Builder` - this separation was missing in the original design of the `DatadogInterceptor`. So having an overload just because of the single **mandatory** parameter looks fine to me.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
xgouchet: > [!TIP]
> Maybe this method could be `protected`, otherwise it might be confusing to our customers

xgouchet (approved):

integrations/dd-sdk-android-okhttp/api/apiSurface
```diff
@@ -16,21 +16,43 @@ class com.datadog.android.okhttp.DatadogEventListener : okhttp3.EventListener
...
+  DEPRECATED constructor(String? = null, List<String>, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
+  DEPRECATED constructor(String? = null, com.datadog.android.okhttp.trace.TracedRequestListener = NoOpTracedRequestListener(), com.datadog.android.rum.RumResourceAttributesProvider = NoOpRumResourceAttributesProvider(), com.datadog.android.core.sampling.Sampler = RateBasedSampler(DEFAULT_TRACE_SAMPLE_RATE))
   override fun intercept(okhttp3.Interceptor.Chain): okhttp3.Response
   override fun onRequestIntercepted(com.datadog.android.api.feature.FeatureSdkCore, okhttp3.Request, io.opentracing.Span?, okhttp3.Response?, Throwable?)
   override fun canSendSpan(): Boolean
   override fun onSdkInstanceReady(com.datadog.android.core.InternalSdkCore)
+  class Builder : BaseBuilder<DatadogInterceptor, Builder>
+    constructor(Map<String, Set<com.datadog.android.trace.TracingHeaderType>>)
+    constructor(List<String>)
+    override fun getThis(): Builder
```
0xnm, line 29: blocking: we shouldn't have this as a part of the public API. Is there a way to avoid that?

0xnm (changes requested):

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorContextInjectionSampledTest.kt
```diff
@@ -0,0 +1,1437 @@
...
+            .setTracedRequestListener(mockRequestListener)
+            .setTraceOrigin(fakeOrigin)
+            .setTraceSampler(mockTraceSampler)
+            .setTraceContextInjection(TraceContextInjection.Sampled)
+            .setLocalTracerFactory(factory)
+            .build()
+    }
+
+    fun getExpectedOrigin(): String? {
+        return fakeOrigin
```
jonathanmos, line 208: Minor: Why do we need this method and not compare directly to fakeOrigin? Also, can this return a null value? fakeOrigin is defined as an alphabetical string so should be non-null

jonathanmos (approved):

integrations/dd-sdk-android-okhttp/api/apiSurface
mariusc83: hmm I really do not know why is added here as it is marked as `internal`. I need to check again. Thanks for spotting this. Do you have any idea why it still appears here ?

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorContextInjectionSampledTest.kt
mariusc83: just legacy code that I did not change

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -397,6 +422,64 @@ internal constructor(
 
     // endregion
 
+    // region Builder
+
+    /**
+     * A Builder for the [DatadogInterceptor].
+     * @param tracedHostsWithHeaderType a list of all the hosts and header types that you want to
+     * be automatically tracked by this interceptor. If registering a GlobalTracer, the tracer must be
```
0xnm, line 430:
```suggestion
     * be automatically tracked by this interceptor. If registering a [GlobalTracer], the tracer must be
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -626,6 +701,136 @@ internal constructor(
 
     // endregion
 
+    // region Builder
+
+    /**
+     * A Builder class for the [TracingInterceptor].
+     * @param tracedHostsWithHeaderType a list of all the hosts and header types that you want to
+     * be automatically tracked by this interceptor. If registering a GlobalTracer, the tracer must be
```
0xnm, line 709:
```suggestion
     * be automatically tracked by this interceptor. If registering a [GlobalTracer], the tracer must be
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -626,6 +701,136 @@ internal constructor(
...
+            this.traceContextInjection = traceContextInjection
+            return getThis()
+        }
+
+        internal fun setLocalTracerFactory(factory: (SdkCore, Set<TracingHeaderType>) -> Tracer): R {
+            this.localTracerFactory = factory
+            return getThis()
+        }
+
+        @Suppress("UndocumentedPublicFunction")
```
0xnm, line 823: can be removed, it is not a public function anymore

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -626,6 +701,136 @@ internal constructor(
...
+        internal abstract fun getThis(): R
+
+        /**
+         * Build the [TracingInterceptor].
```
0xnm, line 827:
```suggestion
         * Build the interceptor.
```

integrations/dd-sdk-android-okhttp/api/dd-sdk-android-okhttp.api
```diff
@@ -44,6 +44,15 @@ public class com/datadog/android/okhttp/DatadogInterceptor : com/datadog/android
 	protected fun onRequestIntercepted (Lcom/datadog/android/api/feature/FeatureSdkCore;Lokhttp3/Request;Lio/opentracing/Span;Lokhttp3/Response;Ljava/lang/Throwable;)V
 }
 
+public final class com/datadog/android/okhttp/DatadogInterceptor$Builder : com/datadog/android/okhttp/trace/TracingInterceptor$BaseBuilder {
+	public fun <init> (Ljava/util/List;)V
+	public fun <init> (Ljava/util/Map;)V
+	public fun build ()Lcom/datadog/android/okhttp/DatadogInterceptor;
+	public synthetic fun build ()Lcom/datadog/android/okhttp/trace/TracingInterceptor;
+	public synthetic fun getThis$dd_sdk_android_okhttp_release ()Lcom/datadog/android/okhttp/trace/TracingInterceptor$BaseBuilder;
```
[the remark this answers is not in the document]
mariusc83: will try but honestly if there's not any I will not fix this tool in this PR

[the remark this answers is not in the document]
mariusc83: was actually fine is just that the file was not re - generated, now it should be fine

PR #2113, opened 2024-06-27
Add sampling rate to internal metrics

xgouchet:
### What does this PR do?

- [x] Add sampling rate option for internal metrics
- [x] Reduce the amount of "Gap between view" telemetry, and discarding gaps above 3 seconds (which most probably denote an app being sent to background instead of an actual gap while navigating inside the tracked app).

> [!NOTE]
> This metric will be applied in addition to the telemetry sampling rate, so if a logMetric is created with a sampling rate of 10%, with the global telemetry sampling rate being sampled at 20%, it has a 2% chance of being kept.

### DIFF
--- a/reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
+++ b/reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
@@ -40,11 +40,8 @@ internal class StubInternalLogger : InternalLogger {
         throwable?.printStackTrace()
     }
 
-    override fun logMetric(
-        messageBuilder: () -> String,
-        additionalProperties: Map<String, Any?>
-    ) {
-        println("M [T]: ${messageBuilder()}")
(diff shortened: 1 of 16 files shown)


reliability/stub-core/src/main/kotlin/com/datadog/android/core/stub/StubInternalLogger.kt
```diff
@@ -40,11 +40,8 @@ internal class StubInternalLogger : InternalLogger {
         throwable?.printStackTrace()
     }
 
-    override fun logMetric(
-        messageBuilder: () -> String,
-        additionalProperties: Map<String, Any?>
-    ) {
-        println("M [T]: ${messageBuilder()}")
+    override fun logMetric(messageBuilder: () -> String, additionalProperties: Map<String, Any?>, samplingRate: Float) {
+        println("M [T]: ${messageBuilder()} | $samplingRate%")
```
mariusc83, line 44: you forgot this ?

mariusc83 (changes requested):

mariusc83 (approved):

0xnm (approved):

PR #2114, opened 2024-06-28
Fix RumSessionEnded metric flaky test

ambushwork:
### What does this PR do?

Fix the min boundary of generating the int for unit test.

### DIFF
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
@@ -290,7 +290,7 @@ class RumSessionEndedMetricTest {
         // Given
         val sessionEndedMetric =
             SessionEndedMetric(fakeSessionId, fakeStartReason, fakeNtpOffsetAtStart, backgroundEventTracking)
-        val backgroundViewCount = forge.anInt(min = 1, max = events.size)
+        val backgroundViewCount = forge.anInt(min = 1, max = events.size + 1)
 
         // When
         events.mapIndexed { index, viewEvent ->
@@ -322,7 +322,7 @@ class RumSessionEndedMetricTest {
         // Given
         val sessionEndedMetric =
             SessionEndedMetric(fakeSessionId, fakeStartReason, fakeNtpOffsetAtStart, backgroundEventTracking)
-        val appLaunchViewCount = forge.anInt(min = 1, max = events.size)
+        val appLaunchViewCount = forge.anInt(min = 1, max = events.size + 1)
 
         // When
         events.mapIndexed { index, viewEvent ->


features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -290,7 +290,7 @@ class RumSessionEndedMetricTest {
         // Given
         val sessionEndedMetric =
             SessionEndedMetric(fakeSessionId, fakeStartReason, fakeNtpOffsetAtStart, backgroundEventTracking)
-        val backgroundViewCount = forge.anInt(min = 1, max = events.size)
+        val backgroundViewCount = forge.anInt(min = 1.coerceAtMost(events.size - 1), max = events.size)
```
0xnm, line 293: I'm not quite clear with this change: given that `events.size` can be 1 at minimum, it means `min = 0`. Why then simply not specifying `min = 0` directly if we allow this value? 🤔 

But anyway, by looking on this test name we probably want the following: at least 1 background view event among all the view events we receive. It means, given that `max` is exclusive and not inclusive, that the logical change will be simply the following one:

```suggestion
        val backgroundViewCount = forge.anInt(min = 1, max = events.size + 1)
```

If `events.size` is 1, then number generated will be 1 (this is range `[1,2)`). If `events.size` is 2, then generation range will be `[1,3)` = `1,2`.

ambushwork: yes, the reason why min=1 is because we have a discussion in the previous PR:
https://github.com/DataDog/dd-sdk-android/pull/2083#discussion_r1642860197

And Indeed I didn't read the doc of  `anInt` carefully,  `max` is actually an exclusive boudary, fixed now.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -290,7 +290,7 @@ class RumSessionEndedMetricTest {
         // Given
         val sessionEndedMetric =
             SessionEndedMetric(fakeSessionId, fakeStartReason, fakeNtpOffsetAtStart, backgroundEventTracking)
-        val backgroundViewCount = forge.anInt(min = 1, max = events.size)
+        val backgroundViewCount = forge.anInt(min = 1, max = events.size + 1)
```
mariusc83, line 293: is there any possibility that `events.size ==0` In that case that function will throw a RuntimeException.

0xnm (approved):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
ambushwork: as @0xnm explained, `@Forgery events: List<ViewEvent>` will not generate empty list, and I made some pressure test with 500 repeats, I didn't see any exception on that.

xgouchet (approved):

PR #2115, opened 2024-06-28
RUM:4717: Improve CheckableTextViewMapper

ambushwork:
#### Summary
This PR improves the session replay `CheckableTextViewMapper`, tries to generate the wireframes which are much closer to the real UI elements.
Since `CheckableTextViewMapper` is a base class of the other mappers, following UI components are affected in session replay:
* CheckedTextView
* RadioButton
* Checkbox

#### Main changes in the PR:
* In `CheckableWireFrameMapper`, merge `resolveCheckedCheckable()` and `resolveNotCheckedCheckable()` functions into one function `resolveCheckable`, because we don't generate different `ShapeWireframe` for the "checked" or "not checked" states, instead, we retrieve the drawable to generate `ImageWireframe`.
* In `CheckableTextViewMapper`,  Inside the implementation of `resolveCheckable` we try to retrieve the drawable of "checked" and "not checked", and then set the `ColorStateList` of the view on the drawable to have the correct tint.
* In `CheckableCompoundButton`, the bounds is adjusted for the drawable rendering, `getCheckableDrawable` is implemented to retrieve the drawable for any `CompoundButton`
* In `CheckedTextView` `getCheckableDrawable` is implemented to retrieve the drawable for `CheckedTextView`.
* Based on the facts above, no specific unit test is needed for each component to verify if the `ShapeWireframe` is correctly generated, we only need to test that `ImageWireFrameHelper.createImageWireframe` is called.
* Add more UI demo in the sample application.

### Demo
| Sample Application | Session Replay |
|------|-------|
| [Screen_recording_20240628_100222.webm](https://github.com/DataDog/dd-sdk-android/assets/15606345/9c05afc0-a1c4-4ef1-ad34-b3479dac0e1e)|https://mobile-integration.datadoghq.com/rum/replay/sessions/f187f5ef-23af-492e-abf6-69f4a248a441?applicationId=38030dde-f9f9-4e52-9443-b9804a030080&seed=141f3702-f9b4-47e9-876f-61219f0cf05c&ts=1719561697402 |

### Motivation
* RUM-4718 - Improve CheckedTextViewMapper
* RUM:4717 - Improve CheckBoxMapper
* RUM:4716 - Improve RadioButton

#### Remain issue
* Masked wireframe to decide
* Resource 401 issue for `CompoundButtonMapper` for the version under 23


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
```diff
@@ -36,27 +40,95 @@ internal abstract class CheckableCompoundButtonMapper<T : CompoundButton>(
...
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
             view.buttonDrawable?.let {
                 checkBoxHeight = it.intrinsicHeight.toLong()
             }
         }
-        // minus the padding
-        checkBoxHeight -= MIN_PADDING_IN_PX * 2
         checkBoxHeight = checkBoxHeight.densityNormalized(pixelsDensity)
         return GlobalBounds(
-            x = viewGlobalBounds.x + MIN_PADDING_IN_PX.densityNormalized(pixelsDensity),
```
0xnm, line 49: why we don't account for padding anymore?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
```diff
@@ -36,27 +40,95 @@ internal abstract class CheckableCompoundButtonMapper<T : CompoundButton>(
...
+                level = InternalLogger.Level.ERROR,
+                targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
+                messageBuilder = { GET_DRAWABLE_FAIL_MESSAGE }
+            )
+            null
+        }
+    }
+
+    private fun cloneCheckableDrawable(view: T, drawable: Drawable): Drawable? {
+        return drawable.constantState?.newDrawable(view.resources)?.apply {
```
0xnm, line 103: is `newDrawable` expensive call CPU and memory-wise?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
```diff
@@ -36,27 +40,95 @@ internal abstract class CheckableCompoundButtonMapper<T : CompoundButton>(
...
+            CompoundButton::class.java.getDeclaredField("mButtonDrawable").apply {
+                isAccessible = true
+            }
+        } catch (e: NoSuchFieldException) {
+            null
+        } catch (e: SecurityException) {
+            null
+        } catch (e: NullPointerException) {
+            null
+        }
```
0xnm, line 132: Don't suppress `SwallowedException`. In fact you can simply create a method which will get a value with 2 calls inside:

```
val drawableField = CompoundButton::class.java.getDeclaredField("mButtonDrawable").apply {
    isAccessible = true
}
return drawableField?.get(view) as? Drawable
```

this will allow you to have an isolated reflection usage site with a common handler for the exceptions coming from both calls and inject `internalLogger` as method argument to log exceptions.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseCheckableTextViewMapperTest.kt
```diff
@@ -0,0 +1,287 @@
...
+    var mockCloneDrawableIntrinsicWidth: Int = 0
+
+    @BeforeEach
+    fun `set up`() {
+        mockClonedDrawable = mock {
+            whenever(it.intrinsicHeight) doReturn mockCloneDrawableIntrinsicHeight
+            whenever(it.intrinsicWidth) doReturn mockCloneDrawableIntrinsicWidth
+        }
+        mockCheckedConstantState = mock {
+            whenever(it.newDrawable(anyVararg(Resources::class))) doReturn mockClonedDrawable
```
0xnm, line 122: why `anyVarag` is used here instead of just `any<Resources>()`? I don't see any `newDrawable` method with `vararg` arguments.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseCheckableTextViewMapperTest.kt
```diff
@@ -0,0 +1,287 @@
...
+                opacity = mockCheckableTextView.alpha
+            )
+        } else {
+            null
+        }
+    }
+
+    // region Unit Tests
+
+    @TestTargetApi(Build.VERSION_CODES.M)
```
0xnm, line 187: Beware about the pitfalls of `TestTargetApi` usage: it will set `Build.VERSION.SDK_INT`, but since it is `static final` field, at some point JIT may inline it, and reflection changes won't have any effect (usually happens after few dozens of calls to this field).

sample/kotlin/src/main/res/color/checkbox_state_tint.xml
```diff
@@ -0,0 +1,12 @@
...
+  ~ This product includes software developed at Datadog (https://www.datadoghq.com/).
+  ~ Copyright 2016-Present Datadog, Inc.
+  -->
+
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <!-- Disabled state -->
+    <item android:color="#60FF0000" android:state_enabled="false" />
+    <item android:color="@android:color/holo_green_dark" android:state_checked="false" />
+    <item android:color="@android:color/holo_red_dark" android:state_checked="true" />
+</selector>
```
0xnm, line 12:
```suggestion
</selector>

```

0xnm (approved):
lgtm. Although I don't have a deep knowledge of SR, so a review from someone else who is closer to the Session Replay topic than I am is preferable.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
ambushwork: Although different `Drawable` has different implementation of this function, most of them (including `AnimatedStateListDrawable` that `CompoundButton` is using) are just Instantiate a new instance of drawable, so I would say there is a memory allocation, but not expensive on CPU.
```
@Override
public Drawable newDrawable() {
    return new AnimatedStateListDrawableCompat(this, null);
}
```

ambushwork:
> lgtm. Although I don't have a deep knowledge of SR, so a review from someone else who is closer to the Session Replay topic than I am is preferable.

Agree, @xgouchet @jonathanmos  do you want to take a look at this?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
ambushwork: Actually this is a result of discussion with @xgouchet we prefer using field instead of a method because the reflection may can have some performance issue if we call it for every wireframe creation, so we just do it once during the initialisation of the class if we use a static field.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
```diff
@@ -36,27 +40,95 @@ internal abstract class CheckableCompoundButtonMapper<T : CompoundButton>(
...
+                    level = InternalLogger.Level.ERROR,
+                    targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
+                    messageBuilder = { GET_DRAWABLE_FAIL_MESSAGE },
+                    throwable = e
+                )
+                null
+            } catch (e: IllegalArgumentException) {
+                internalLogger.log(
+                    level = InternalLogger.Level.ERROR,
+                    targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 85: are we sure we want to send these errors with the buttonDrawable as telemetry - would it be actionable?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
ambushwork: Before we drew our own shape to display the square of the checkbox, so the padding should be used. but now we directly use the drawable inside `CompoundButton`, which already has a inset padding, we don't need to add an additional one.
image

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseCheckableTextViewMapperTest.kt
ambushwork: fixed, now I use a `mockResources` on this point.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseCheckableTextViewMapperTest.kt
ambushwork: Good to know, if there is any modification that I should apply on this test?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
0xnm: makes sense indeed then

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/BaseCheckableTextViewMapperTest.kt
0xnm: ideally we could inject `BuildSdkVersionProvider` wrapper which we can read an mock, but we can leave it for the later refactoring.

sample/kotlin/src/main/res/color/checkbox_state_tint.xml
ambushwork: fixed

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
ambushwork: Good question, I assume it will happen when clients have some very customised view from `CompoundButton`, the action that we can take is to investigate the proportion of the case and then take some implementation coverage

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
```diff
@@ -36,27 +40,95 @@ internal abstract class CheckableCompoundButtonMapper<T : CompoundButton>(
...
+            val checkableDrawableIndex = if (view.isChecked) {
+                CHECK_BOX_CHECKED_DRAWABLE_INDEX
+            } else {
+                CHECK_BOX_NOT_CHECKED_DRAWABLE_INDEX
+            }
+            (view.buttonDrawable?.constantState as? DrawableContainer.DrawableContainerState)?.getChild(
+                checkableDrawableIndex
+            )
+        } else {
+            // view.buttonDrawable is not available below API 23, so reflection is used to retrieve it.
```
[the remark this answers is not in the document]
ambushwork: Actually a good idea, the thing is that, as i discussed with Nikita in the thread above, here the reflection logic is better to be kept as a class field for the performance issue, because calling a method of reflection for every wireframe creation can be time consuming. Based on that, I wonder if you still prefer to have a dedicated class with only static fields inside?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
```diff
@@ -36,27 +40,95 @@ internal abstract class CheckableCompoundButtonMapper<T : CompoundButton>(
     @UiThread
     override fun resolveCheckableBounds(view: T, pixelsDensity: Float): GlobalBounds {
         val viewGlobalBounds = viewBoundsResolver.resolveViewGlobalBounds(view, pixelsDensity)
-        var checkBoxHeight = DEFAULT_CHECKABLE_HEIGHT_IN_PX
+        var checkBoxHeight = DEFAULT_CHECKABLE_HEIGHT_IN_DP * pixelsDensity.toLong()
```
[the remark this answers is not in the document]
ambushwork: you are right, i updated it with more readable logic

jonathanmos (approved):

xgouchet (approved):

PR #2123, opened 2024-07-08
RUM-4731: Add MethodCall telemetry for compose mapper

ambushwork:
### What does this PR do?

Adds MethodCall telemetry for compose mapper, so that we can know the duration of each compose mapper execution.

### Motivation

RUM-4731

### DIFF
--- a/features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/ComposeWireframeMapper.kt
+++ b/features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/ComposeWireframeMapper.kt
@@ -10,9 +10,8 @@ import android.os.SystemClock
 import androidx.compose.runtime.Composer
 import androidx.compose.runtime.tooling.CompositionGroup
 import androidx.compose.ui.platform.ComposeView
-import com.datadog.android.Datadog
 import com.datadog.android.api.InternalLogger
-import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.api.feature.measureMethodCallPerf
 import com.datadog.android.sessionreplay.compose.internal.data.ComposeContext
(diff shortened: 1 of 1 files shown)


features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/ComposeWireframeMapper.kt
```diff
@@ -121,50 +124,62 @@ internal class ComposeWireframeMapper(
...
+            mapper.map(compositionGroup, composeContext, parentUiContext)
+        }
     }
 
     // endregion
 
     companion object {
         private const val HUNDRED: Double = 100.0
+        private const val METHOD_CALL_MAP_PREFIX: String = "[Compose] Map with"
+        const val METHOD_CALL_SAMPLING_RATE = 1f
```
jonathanmos, line 183: Is there a reason to set this so high? It will send telemetry for every call to the mapper. You can use the existing [MethodCallSamplingRate](https://github.com/DataDog/dd-sdk-android/blob/develop/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/metrics/MethodCallSamplingRate.kt) enum for the sampling rate

ambushwork: Currently it's on feature branch of compose session replay, so it will be only me who can generate data when testing, and if I understand correctly `1f` means 1/100 of call instead of every call. 

And it seems `MethodCallSamplingRate` is not available yet in this feature branch. probably we will do a merge from master after Xavier comes back

jonathanmos: "1f means 1/100 of call instead of every call." - you're absolutely correct, my mistake.

jonathanmos (approved):

PR #2126, opened 2024-07-17
RUM-648 Add request id in okhttp request

xgouchet:
### What does this PR do?

This PR adds a better way to track OkHttp requests as resources avoiding key collisions. Our keys used to only take into account the url, now we generate a UUID attached to the request to have a more accurate resource representation.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
@@ -126,7 +126,7 @@ interface RumMonitor {
             " Use `startResource` method which takes `RumHttpMethod` as `method` parameter instead."
     )
     fun startResource(
-        key: String,
+        key: Any,
         method: String,
         url: String,
         attributes: Map<String, Any?> = emptyMap()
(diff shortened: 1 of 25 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/resource/ResourceId.kt
```diff
@@ -0,0 +1,41 @@
...
+ * Describes a Resource being tracked in RUM.
+ *
+ * A resource will first try to be matched with the UUID if possible, and if missing will use the key property.
+ * The UUID should be unique (based off of [UUID] class). If missing, because key could have duplicates,
+ * this can lead to unpredictable behaviors and erroneous Resource timings.
+ *
+ * @param key a key to identify the resource
+ * @param uuid a UUID based unique identifier (optional)
+ */
+class ResourceId(
```
ambushwork, line 21: is it a public class for client?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -126,7 +126,7 @@ interface RumMonitor {
             " Use `startResource` method which takes `RumHttpMethod` as `method` parameter instead."
     )
     fun startResource(
-        key: String,
+        key: Any,
```
ambushwork, line 129: I wonder why we need to keep `Any` type here? can we just use the new class `ResourceId` instead?

0xnm (approved):
nicely done!

PR #2127, opened 2024-07-18
RUM-5447 Fix the racing condition in the RotatingDnsResolver logic

mariusc83:
### What does this PR do?

In this PR we are fixing the race condition problem that was caused by a concurrent access to an unprotected list in the `RotatingDnsResolver` class. This was raised by the Support team in this ticket: https://datadog.zendesk.com/agent/tickets/1761693

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolver.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolver.kt
@@ -28,9 +28,11 @@ internal class RotatingDnsResolver(
         }
 
         fun rotate() {
-            val first = addresses.removeFirstOrNull()
-            if (first != null) {
-                addresses.add(first)
+            synchronized(addresses) {
+                val first = addresses.removeFirstOrNull()
+                if (first != null) {
+                    addresses.add(first)
+                }
             }
         }
     }
@@ -44,12 +46,18 @@ internal class RotatingDnsResolver(
 
         return if (knownHost != null && isValid(knownHost)) {
             knownHost.rotate()
-            knownHost.addresses
+            safeCopy(knownHost.addresses)
         } else {
             @Suppress("UnsafeThirdPartyFunctionCall") // handled by caller
             val result = delegate.lookup(hostname)
             knownHosts[hostname] = ResolvedHost(hostname, result.toMutableList())
-            result
+            safeCopy(result)
+        }
+    }
+
+    private fun safeCopy(list: List<InetAddress>): List<InetAddress> {
+        return synchronized(list) {
+            list.toList()
         }
     }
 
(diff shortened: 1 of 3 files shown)


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
```diff
@@ -111,6 +114,30 @@ internal class RotatingDnsResolverTest {
         assertThat(result2).containsExactlyElementsOf(fakeInetAddresses)
     }
 
+    @RepeatedTest(30)
```
0xnm, line 117: do we need to repeat it 30 times? won't it be too long? we are already spinning 100 threads

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
```diff
@@ -111,6 +114,30 @@ internal class RotatingDnsResolverTest {
...
+
+        // When
+        repeat(100) {
+            Thread {
+                assertThatNoException().isThrownBy {
+                    Thread.sleep(forge.aLong(min = 0, max = 100))
+                    testedDns.lookup(fakeHostname)
+                }
+                countDownLatch.countDown()
+            }.start()
```
0xnm, line 136: if I'm not wrong, this will never throw in the end, the thread spinned will be killed, but error won't be propagated further and won't stop the test.

Moreover, since `countDownLatch.countDown()` won't happen in case of the error, `countDownLatch.await()` will be waiting forever.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
mariusc83: yes but it's randomly re-produceable. I am afraid that if we lower it it might be a false positive

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
mariusc83: hmmm not sure based on my tests, theoretically `assertThatNoException` should stop the test with a fail. In that case the `countDownLatch` will not wait anymore. I will double check this though.

xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolver.kt
```diff
@@ -44,12 +46,18 @@ internal class RotatingDnsResolver(
...
         } else {
             @Suppress("UnsafeThirdPartyFunctionCall") // handled by caller
             val result = delegate.lookup(hostname)
             knownHosts[hostname] = ResolvedHost(hostname, result.toMutableList())
-            result
+            safeCopy(result)
+        }
+    }
+
+    private fun safeCopy(list: List<InetAddress>): List<InetAddress> {
```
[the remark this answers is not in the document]
mariusc83: I asked myself the same question but for now I could not see any other place. My rule of thumb is that an extension function should have at least 2 different classes where is being used in order to be extracted so that's the reason why I decided to keep it private, non extension here.

PR #2130, opened 2024-07-22
RUM-3826 Add telemetry on batch writing operation

xgouchet:
### What does this PR do?

Adds a telemetry to measure the delay between the moment data is planned to be written, and the moment it's actually written.

### Motivation

This delay is expected, and will drive the backpressure experienced by the executors on the persistence layer.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
@@ -20,6 +20,8 @@ import com.datadog.android.core.internal.persistence.file.FileReaderWriter
 import com.datadog.android.core.internal.persistence.file.batch.BatchFileReaderWriter
 import com.datadog.android.core.internal.persistence.file.existsSafe
 import com.datadog.android.core.internal.utils.submitSafe
+import com.datadog.android.core.metrics.MethodCallSamplingRate
+import com.datadog.android.core.metrics.TelemetryMetricType
 import com.datadog.android.privacy.TrackingConsent
 import java.io.File
 import java.util.Locale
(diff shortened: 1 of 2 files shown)


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorageTest.kt
```diff
@@ -98,8 +101,32 @@ internal class ConsentAwareStorageTest {
...
     @BeforeEach
     fun `set up`() {
+        whenever(mockPendingOrchestrator.getRootDir()) doReturn File(mockPendingRootParentFile, fakeRootDirName)
+        whenever(mockGrantedOrchestrator.getRootDir()) doReturn File(mockGrantedRootParentFile, fakeRootDirName)
+
+        whenever(
+            mockInternalLogger.startPerformanceMeasure(
+                "com.datadog.android.core.internal.persistence.ConsentAwareStorage",
+                TelemetryMetricType.MethodCalled,
+                0.001f,
```
jonathanmos, line 125: minor: could use the `MethodCallSamplingRate` enum

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
```diff
@@ -57,6 +59,13 @@ internal class ConsentAwareStorage(
             TrackingConsent.NOT_GRANTED -> null
         }
 
+        val metric = internalLogger.startPerformanceMeasure(
```
jonathanmos, line 62: why create this outside the lock and not inside the synchronized block?

xgouchet: > [!NOTE]
> The goal here is not to measure just the write operation, but also the delay between the moment we want to write, and the moment the data is actually written. Because write operations are queued in a single thread, we could create a bottleneck, and measure this delay here will help us find ways to improve it.

jonathanmos (approved):

mariusc83 (approved):

PR #2133, opened 2024-07-23
[RUM-5566] Added handleSyntheticsAttribute in RumInternalProxy

marco-saia-datadog:
### What does this PR do?

Adds `handleSyntheticsAttribute` in `_RumInternalProxy` for internal usage in cross-platform frameworks.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
```diff
@@ -31,6 +32,7 @@ import com.datadog.android.telemetry.model.TelemetryConfigurationEvent
     "VariableNaming"
 )
 class _RumInternalProxy internal constructor(private val rumMonitor: AdvancedRumMonitor) {
+    private var handledSyntheticsAttribute = false
```
mariusc83, line 35: is this variable going to be handle from different threads ? In that case because you check for its value below the way I remember (hope I am not wrong) booleans are not volatile by default in JVM so you either use `@Volatile` annotation or use an `AtomicBoolean` in there. Also if this is the case you might need some unit tests that access these methods from multiple threads.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
```diff
@@ -40,6 +42,28 @@ class _RumInternalProxy internal constructor(private val rumMonitor: AdvancedRum
         rumMonitor.updatePerformanceMetric(metric, value)
     }
 
+    @Suppress("MemberVisibilityCanBePrivate")
+    fun handleSyntheticsAttribute(testId: String?, resultId: String?) {
```
xgouchet, line 46: > [!IMPORTANT]
> Let's name it `setSyntheticsAttribute` instead, to keep consistent with the RumMonitor interface

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
```diff
@@ -40,6 +42,28 @@ class _RumInternalProxy internal constructor(private val rumMonitor: AdvancedRum
...
+
+        this.handledSyntheticsAttribute = true
+        if (testId.isNullOrBlank() || resultId.isNullOrBlank()) {
+            return
+        }
+
+        rumMonitor.setSyntheticsAttribute(testId, resultId)
+    }
+
+    internal fun handleSyntheticsAttribute(intent: Intent) {
```
xgouchet, line 59: > [!IMPORTANT]
> Let's rename it `setSyntheticsAttributeFromIntent`

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/config/GlobalRumMonitorTestConfiguration.kt
```diff
@@ -9,19 +9,33 @@ package com.datadog.android.rum.utils.config
...
 import fr.xgouchet.elmyr.Forge
 import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+
+@Suppress("TestFunctionName")
+internal abstract class InternalAdvancedRumMonitor : AdvancedRumMonitor {
+    override fun _getInternal(): _RumInternalProxy? {
+        return null
+    }
+}
```
xgouchet, line 24: > [!NOTE]
> Why is this part required ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
marco-saia-datadog: Good observation. The new API should only be called internally, and only once. The variable prevents SDK consumers from calling it again and override the synthetics attribute. Since there is no guarantee that it won't be called from different threads, I believe it still makes sense to make it volatile or use an `AtomicBoolean`.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/config/GlobalRumMonitorTestConfiguration.kt
marco-saia-datadog: Without explicitly overriding the `_getInternal()` method, a runtime error is thrown when executing the unit tests, saying that `AdvancedRumMonitor` does not implement the abstract method `_getInternal()`. It might be because it is marked with `@JvmSynthetic`.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/config/GlobalRumMonitorTestConfiguration.kt
```diff
@@ -9,19 +9,33 @@ package com.datadog.android.rum.utils.config
...
 import com.datadog.android.rum.RumMonitor
+import com.datadog.android.rum._RumInternalProxy
 import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
 import com.datadog.tools.unit.extensions.config.MockTestConfiguration
 import fr.xgouchet.elmyr.Forge
 import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+
+@Suppress("TestFunctionName")
+internal abstract class InternalAdvancedRumMonitor : AdvancedRumMonitor {
```
mariusc83, line 20: Like @xgouchet I do not understand why do you need this class : `InternalAdvancedRumMonitor` ?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/config/GlobalRumMonitorTestConfiguration.kt
```diff
@@ -9,19 +9,33 @@ package com.datadog.android.rum.utils.config
...
-    MockTestConfiguration<RumMonitor>(AdvancedRumMonitor::class.java) {
+    MockTestConfiguration<RumMonitor>(InternalAdvancedRumMonitor::class.java) {
 
     lateinit var mockSdkCore: InternalSdkCore
 
     override fun setUp(forge: Forge) {
         super.setUp(forge)
         mockSdkCore = mock()
+
+        (mockInstance as? InternalAdvancedRumMonitor)?.let {
```
mariusc83, line 35: can't you just mock this on the `AdvancedMonitor` interface itself ?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/config/GlobalRumMonitorTestConfiguration.kt
marco-saia-datadog: It seems like the `_getInternal` method is not visible in unit testing, unless explicitly implemented.

https://github.com/DataDog/dd-sdk-android/pull/2133#discussion_r1689429377

mariusc83: yes because mockito is using Java under the hood and that method is not visible in Java if you annotate it with @JvmSynthetic :(. I don't like this but I guess we can't go around it.

mariusc83 (approved):

xgouchet (approved):

marco-saia-datadog:
I have added [a check](https://github.com/DataDog/dd-sdk-android/pull/2133/files#diff-55794e90e3a4711c7805c0b60b027fdac6c46b882bff7040d6959de41dac3d99R94-R98) to verify that the lateinit variable `sdkCore` has been initialized in the `ActivityLifecycleTrackingStrategy`:

```kotlin
if (::sdkCore.isInitialized) {
/// ...
}
```

A test was failing because of this missing check, as the previous implementation would only use the `sdkCore` if the activity contained the synthetics attributes intent extras, that are not passed for that particular test.

@mariusc83 @xgouchet let me know if you prefer to modify the test instead, or if you think this approach is safer.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityLifecycleTrackingStrategy.kt
```diff
@@ -92,16 +91,11 @@ abstract class ActivityLifecycleTrackingStrategy :
...
-        val extras = intent.safeExtras
-        val testId = extras?.getString("_dd.synthetics.test_id")
-        val resultId = extras?.getString("_dd.synthetics.result_id")
-        if (!testId.isNullOrBlank() && !resultId.isNullOrBlank()) {
-            (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor)
-                ?.setSyntheticsAttribute(
-                    testId,
-                    resultId
-                )
+        if (::sdkCore.isInitialized) {
```
mariusc83, line 94: While there is nothing wrong with this condition the way I see it now, I wonder which test was failing without adding it ? What happened before if the `sdkCore` was not initialized ? Was it defaulting to a `NoOp` ?

marco-saia-datadog: Previously we checked if the intent contained the synthetics attributes in the extras, and only then we would use the `sdkCore`. 

With this PR we delegate the responsibility of checking the intent extras to the new RUMMonitor method, so we have to access the core before checking the intent extras. 

The problem occurs because - in that specific test - the register method is never called.

This is the test: [when activity created will not delegate to the bundled strategies() ](https://github.com/DataDog/dd-sdk-android/blob/6319509dfae068ade55741b6b57173670f0f22cf/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/MixedViewTrackingStrategyTest.kt#L84)

Issue #2131, opened 2024-07-23
Repeated network error due to "browser-intake-datadoghq.com"


lzanita09:
### Describe the bug

We have been seeing a ton of repeated errors like these from our release variant:
```
20:24:17.556  W  [_dd.sdk_core.default]: Batch c4e38407-3402-401c-83e2-171154a9180f [751 bytes] (Traces Request) failed because of a network error; we will retry later.
20:24:17.556  E  [_dd.sdk_core.default]: Unable to execute the request; we will retry later.
20:24:17.556  E  
20:24:17.556  W  [_dd.sdk_core.default]: Batch 4cc740c2-0824-4fff-9084-1c6d2f20c865 [1748 bytes] (RUM Request) failed because of a network error; we will retry later.
20:24:17.559  E  [_dd.sdk_core.default]: Unable to execute the request; we will retry later.
20:24:17.559  E  
20:24:17.559  W  [_dd.sdk_core.default]: Batch 02dd0b1f-5d59-4f0f-a115-04f14d3e1d54 [751 bytes] (Traces Request) failed because of a network error; we will retry later.
20:24:17.559  E  [_dd.sdk_core.default]: Unable to execute the request; we will retry later.
20:24:17.559  E  
20:24:17.559  W  [_dd.sdk_core.default]: Batch 251b08fa-cd95-4aac-9a2b-65114a3ea1ea [7033 bytes] (RUM Request) failed because of a network error; we will retry later.
20:24:17.561  E  [_dd.sdk_core.default]: Unable to execute the request; we will retry later.
20:24:17.561  E  
20:24:17.561  W  [_dd.sdk_core.default]: Batch af1811b7-b41d-4689-88d1-2838fd607dda [751 bytes] (Traces Request) failed because of a network error; we will retry later.
20:24:17.561  E  [_dd.sdk_core.default]: Unable to execute the request; we will retry later.
```

After putting a breakpoint on `DataOkHttpUploader#upload` method, we were able to see the exception from the try/catch block around `executeUploadRequest`:
```
Unable to resolve host "browser-intake-datadoghq.com": No address associated with hostname
```

Because this seems to happen repeatedly, it caused the app to drain the battery very quickly.

### Reproduction steps

Unsure how to reproduce this issue consistently. 

### Affected SDK versions

2.11.0

### Latest working SDK version

Unsure, tried 2.7.0, 2.8.0, 2.9.0, and 2.11.0, all had the same issue

[...]

PR #2135, opened 2024-07-23
RUM-5564 increase retry delay on DNS error

xgouchet:
### What does this PR do?

When a DNS issue prevents the SDK from reaching our intake hosts, we increase the delay to avoid an infinite loop. 

### Motivation

By default our retry loop can be quite frequent, but DNS issues often take some time to be fixed. Although an actual issue in our DNS registration is unlikely, there seems to also be some issues on some networks preventing the devices from resolving the intake host address. 

### Additional Notes

Relates to #2131

### DIFF
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/DNSErrorStatusForgeryFactory.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/DNSErrorStatusForgeryFactory.kt
@@ -1,0 +1,18 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.utils.forge
+
+import com.datadog.android.core.internal.data.upload.UploadStatus
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+internal class DNSErrorStatusForgeryFactory : ForgeryFactory<UploadStatus.DNSError> {
+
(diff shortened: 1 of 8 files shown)


dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/DNSErrorStatusForgeryFactory.kt
```diff
@@ -0,0 +1,18 @@
...
+package com.datadog.android.utils.forge
+
+import com.datadog.android.core.internal.data.upload.UploadStatus
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+internal class DNSErrorStatusForgeryFactory : ForgeryFactory<UploadStatus.DNSError> {
+
+    override fun getForgery(forge: Forge): UploadStatus.DNSError {
+        return UploadStatus.DNSError
```
mariusc83, line 16: why we need a factory here which only returns one single type of enum ?

xgouchet: It's not an enum but an object as part of a `sealed class`

jonathanmos (approved):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/DNSErrorStatusForgeryFactory.kt
mariusc83: even so in this case this factory only returns one value of that `UploadStatus` sealed class. Wouldn't it be more simple to just add this value in your test data without using forge ?

mariusc83 (approved):
Except for that Forge factory which I don't think is needed this LGTM

PR #2140, opened 2024-07-24
Fix flaky test in TLVBlockFileReader

xgouchet:
### What does this PR do?
We had some flakiness in the TLV tests


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -88,6 +89,7 @@ internal class TLVBlockFileReaderTest {
     @Test
     fun `M log error W read() { invalid TLV type }`() {
         // Given
+        assumeTrue(fakeDataString.length > 2)
```
ambushwork, line 92: Minor:  this assumption can fix flaky test, but it seems this case worth being covered as well.  we can have a if else condition to test if the length is smaller or equal than 2, `internalLogger` should output error which starts with "Failed to deserialize TLV data length".

ambushwork (approved):

mariusc83 (approved):

PR #2142, opened 2024-07-25
RUM-5553: Implement MetricHttpClient and RequestBodyBuilder

ambushwork:
### What does this PR do?

Implement MetricHttpClient and RequestBodyBuilder,
The request corresponds to the spec: 
https://docs.datadoghq.com/api/latest/metrics/#submit-metrics

### Motivation

RUM-5553

### DIFF
--- a/tools/benchmark/src/main/java/com/datadog/benchmark/DatadogExporterConfiguration.kt
+++ b/tools/benchmark/src/main/java/com/datadog/benchmark/DatadogExporterConfiguration.kt
@@ -1,0 +1,148 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.benchmark
+
+/**
+ * Describes the configuration to be used in Datadog open telemetry exporter.
+ * @param serviceName the name of the service.
+ * @param resource the name of resource.
+ * @param applicationName the name of host application.
+ * @param applicationVersion the version of host application.
+ * @param applicationId the id of the host application.
+ * @param apiKey the api key for submitting metrics to datadog end points.
+ * @param run the run description of benchmark test.
+ * @param endPoint the endpoint to submit the metrics.
+ * @param intervalInSeconds the interval of seconds of sampling and uploading vital data.
+ */
+data class DatadogExporterConfiguration internal constructor(
+    val serviceName: String? = null,
+    val resource: String? = null,
+    val applicationName: String? = null,
+    val applicationVersion: String? = null,
+    val applicationId: String? = null,
+    val apiKey: String,
+    val run: String? = null,
+    val endPoint: EndPoint = EndPoint.US1,
+    val intervalInSeconds: Long = DEFAULT_INTERVAL_IN_SECONDS
+) {
+    /**
+     * A Builder class for a [DatadogExporterConfiguration].
+     *
+     * @param apiKey api key to submit metrics to Datadog endpoint
+     */
+    class Builder(private val apiKey: String) {
+        private var serviceName: String? = null
+        private var resource: String? = null
+        private var applicationName: String? = null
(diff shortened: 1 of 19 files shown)


tools/benchmark/src/main/java/com/datadog/benchmark/internal/RequestBodyBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.sdk.metrics.data.DoublePointData
+import io.opentelemetry.sdk.metrics.data.LongPointData
+import io.opentelemetry.sdk.metrics.data.MetricData
+import io.opentelemetry.sdk.metrics.data.MetricDataType
+import io.opentelemetry.sdk.metrics.data.PointData
+import io.opentelemetry.sdk.resources.Resource
+import java.util.concurrent.TimeUnit
+
+internal class RequestBodyBuilder(private val metricContext: MetricContext) {
```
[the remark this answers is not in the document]
ambushwork: Good point, `RequestBodyBuilder` interface is created for it.

tools/benchmark/src/main/java/com/datadog/benchmark/internal/MetricHttpClient.kt
```diff
@@ -0,0 +1,163 @@
...
+            requestBodyBuilder.buildJsonElement(metrics)
+        )
+        if (response.isSuccess) {
+            Log.i(BENCHMARK_LOGCAT_TAG, "${metrics.size} Metrics data uploaded.")
+        } else {
+            Log.e(BENCHMARK_LOGCAT_TAG, "Metrics data failed to upload. ${response.exceptionOrNull()}")
+        }
+    }
+
+    private fun postRequest(
```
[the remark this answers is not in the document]
ambushwork: In my opinion `postRequest` will also work with traces because it takes generic parameters which are `DatadogExporterConfiguration` and `JsonElement`.

Maybe the name of the class is misleading, it shouldn't be dedicated to metrics, so I rename if `DatadogHttpClient` if it seems better to you

xgouchet (approved):

tools/benchmark/src/main/java/com/datadog/benchmark/DatadogExporterConfiguration.kt
```diff
@@ -0,0 +1,148 @@
...
+            this.applicationVersion = applicationVersion
+            return this
+        }
+
+        /**
+         * Sets the run environment description.
+         *
+         * @param run run environment description
+         */
+        fun setRun(run: String): Builder {
```
0xnm, line 112: why not call it `setEnv(env: String)` then?

tools/benchmark/src/main/java/com/datadog/benchmark/EndPoint.kt
```diff
@@ -0,0 +1,47 @@
...
+        traces = "https://trace.browser-intake-us5-datadoghq.com/"
+    ),
+    EU1(
+        metrics = "https://api.datadoghq.eu/",
+        traces = "https://public-trace-http-intake.logs.datadoghq.eu/"
+    ),
+    US1_FED(
+        metrics = "https://api.ddog-gov.com/",
+        traces = "https://trace.browser-intake-ddog-gov.com/"
+    );
```
0xnm, line 36: if it is only for us, do we need this one? Also I don't see `AP1`.

tools/benchmark/src/main/java/com/datadog/benchmark/DatadogExporterConfiguration.kt
ambushwork: Because in the RFC the tag name should be `run`, so here I want to align with it:
``` 
[
  "device_model:iphone1.2",
  "os:ios",
  "os_version:17.4",
  "run:baseline|instrumented|profiling",
  "scenario:sr|rum|trace|logs",
  "application_id:com.datadoghq.foo",
  "sdk_version:2.0.0",
  "branch:develop"
]

```

PR #2144, opened 2024-07-25
RUM-5553: Add DatadogMeter to read vital data for benchmark

ambushwork:
### What does this PR do?

Add DatadogMeter to read vital data for benchmark.

With this commit, the sample application can activate the `DatadogMeter` by adding this:

```
val datadogExporterConfiguration = com.datadog.benchmark.DatadogExporterConfiguration
            .Builder(apiKey = "xxxxxx")
            .setApplicationName("Sample Application")
            .setApplicationId(BuildConfig.APPLICATION_ID)
            .setApplicationVersion(BuildConfig.VERSION_NAME)
            .setIntervalInSeconds(20L)
            .setRun("baseline")
            .build()
 val meter = DatadogMeter.create(datadogExporterConfiguration)

meter.startGauges()

meter.stopGauges()
```

### Motivation
RUM-5553

### Additional Notes

Some vital readers classes and unit tests are copied from feature modules for following reasons:
* This module should depend or use as less as possible the functionalities from SDK to get better result.
* We might have different requirements of metrics apart from the vitals for client ( for example for the cpu vital we wish to have a cpu ticks timeseries instead of a cumulative value in production) , so isolate them from SDK is easier to maintain, and will have a risk of breaking production.
* In the future we might make PR of this module to public `opentelemetry-java`  repository, all the classes need to be independent from our SDK.

### DIFF
--- a/tools/benchmark/src/main/java/com/datadog/benchmark/DatadogMeter.kt
+++ b/tools/benchmark/src/main/java/com/datadog/benchmark/DatadogMeter.kt
@@ -1,0 +1,105 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.benchmark
+
+import com.datadog.benchmark.exporter.DatadogMetricExporter
+import com.datadog.benchmark.internal.reader.CPUVitalReader
+import com.datadog.benchmark.internal.reader.FpsVitalReader
+import com.datadog.benchmark.internal.reader.MemoryVitalReader
+import com.datadog.benchmark.internal.reader.VitalReader
+import com.datadog.benchmark.noop.NoOpObservableDoubleGauge
+import io.opentelemetry.api.OpenTelemetry
+import io.opentelemetry.api.metrics.Meter
+import io.opentelemetry.api.metrics.ObservableDoubleGauge
+import io.opentelemetry.sdk.OpenTelemetrySdk
+import io.opentelemetry.sdk.metrics.SdkMeterProvider
+import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader
+import java.util.concurrent.TimeUnit
+
+/**
+ * This class is responsible for managing the performance gauges related to CPU, FPS, and memory usage.
+ * It provides functionalities to start and stop monitoring these metrics, which will be uploaded to
+ * Datadog metric API.
+ */
+class DatadogMeter private constructor(private val meter: Meter) {
+
+    private val cpuVitalReader: CPUVitalReader = CPUVitalReader()
+    private val memoryVitalReader: MemoryVitalReader = MemoryVitalReader()
+    private val fpsVitalReader: FpsVitalReader = FpsVitalReader()
+
+    private val gaugesByMetricName: MutableMap<String, ObservableDoubleGauge> = mutableMapOf()
+
+    /**
+     * Starts cpu, memory and fps gauges.
+     */
+    fun startGauges() {
+        startGauge(cpuVitalReader)
+        startGauge(memoryVitalReader)
+        startGauge(fpsVitalReader)
+    }
+
+    /**
+     * Stops cpu, memory and fps gauges.
+     */
+    fun stopAllGauges() {
+        stopGauge(cpuVitalReader)
+        stopGauge(memoryVitalReader)
+        stopGauge(fpsVitalReader)
+    }
+
+    private fun startGauge(reader: VitalReader) {
+        synchronized(reader) {
+            // Close the gauge if it exists before.
+            val metricName = reader.metricName()
+            gaugesByMetricName[metricName]?.close()
+            reader.start()
+            meter.gaugeBuilder(metricName).apply {
+                reader.unit()?.let { unit ->
+                    setUnit(unit)
+                }
+            }.buildWithCallback { observableDoubleMeasurement ->
+                reader.readVitalData()?.let { data ->
+                    observableDoubleMeasurement.record(data)
+                }
+            }.also { observableDoubleGauge ->
+                gaugesByMetricName[metricName] = observableDoubleGauge
+            }
+        }
+    }
+
+    private fun stopGauge(reader: VitalReader) {
+        synchronized(reader) {
+            reader.stop()
+            gaugesByMetricName[reader.metricName()]?.close()
+            gaugesByMetricName[reader.metricName()] = NoOpObservableDoubleGauge()
+        }
+    }
+
+    companion object {
+
+        /**
+         * Creates an instance of [DatadogMeter] with given configuration.
+         */
+        fun create(datadogExporterConfiguration: DatadogExporterConfiguration): DatadogMeter {
+            val sdkMeterProvider = SdkMeterProvider.builder()
+                .registerMetricReader(
+                    PeriodicMetricReader.builder(DatadogMetricExporter(datadogExporterConfiguration))
+                        .setInterval(datadogExporterConfiguration.intervalInSeconds, TimeUnit.SECONDS)
+                        .build()
+                )
+                .build()
+
+            val openTelemetry: OpenTelemetry = OpenTelemetrySdk.builder()
+                .setMeterProvider(sdkMeterProvider)
+                .build()
+            val meter = openTelemetry.getMeter(METER_INSTRUMENTATION_SCOPE_NAME)
+            return DatadogMeter(meter)
+        }
+
+        private const val METER_INSTRUMENTATION_SCOPE_NAME = "datadog.open-telemetry"
+    }
+}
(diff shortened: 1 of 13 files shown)


tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/CPUVitalReader.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.datadog.benchmark.ext.canReadSafe
+import com.datadog.benchmark.ext.existsSafe
+import com.datadog.benchmark.ext.readTextSafe
+import java.io.File
+
+/**
+ * Reads the CPU `utime` based on the `/proc/self/stat` file.
+ * cf. documentation https://man7.org/linux/man-pages/man5/procfs.5.html
+ */
+internal class CPUVitalReader(
```
xgouchet, line 18: > [!IMPORTANT]
> Regarding the CPU usage benchmark, right now we're reading the total number of ticks used by the host process. The main issue it's a value that is  always growing throughout the host app lifespan. 
> If we report it as a metric, it needs to be a value that can be looked at outside of its run context. 
> A way to do it is reporting the "ticks per second". In the RUM vitals, we do it on a RUM View by dividing the total ticks used by the duration of the view. In this case, because we don't have the notion of views, and we can't be sure that the read operations will be performed at exact regular intervals, you can simply divide the diff since the last read by the time spent since last read.

xgouchet (changes requested):

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/CPUVitalReader.kt
ambushwork: Actually this class is a copy version of SDK `CPUVitalReader` with a small modification which makes CPU ticks from  cumulative value to a timeseries statistic as I wrote in the description in the PR.
I save a `lastCpuTicks` just as you said so every output is the difference from current and last ticks during a given interval.

So on web it looks like this:
image


Please check if this logic seems good to you?

tools/benchmark/src/main/java/com/datadog/benchmark/DatadogMeter.kt
```diff
@@ -0,0 +1,103 @@
...
+    private fun startGauge(reader: VitalReader) {
+        synchronized(reader) {
+            // Close the gauge if it exists before.
+            val metricName = reader.metricName()
+            gaugesByMetricName[metricName]?.close()
+            reader.start()
+            meter.gaugeBuilder(metricName).apply {
+                reader.unit()?.let {
+                    this.setUnit(it)
+                }
```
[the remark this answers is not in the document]
ambushwork: I change implicit parameter of `it` to explicit one with name "unit", so it becomes:
```
 meter.gaugeBuilder(metricName).apply {
                reader.unit()?.let { unit ->
                    setUnit(unit)
                }
```

Does it seem better to you?

xgouchet (approved):

mariusc83 (dismissed):

ambushwork:
@jonathanmos I just made the fix on the gauge as we discussed, please review it

jonathanmos (dismissed):

jonathanmos (approved):

tools/benchmark/src/main/java/com/datadog/benchmark/exporter/LogsMetricExporter.kt
```diff
@@ -0,0 +1,59 @@
...
+        applicationId = datadogExporterConfiguration.applicationId ?: DEFAULT_APPLICATION_ID,
+        intervalInSeconds = datadogExporterConfiguration.intervalInSeconds
+    )
+
+    override fun getAggregationTemporality(instrumentType: InstrumentType): AggregationTemporality {
+        return AggregationTemporality.DELTA
+    }
+
+    override fun export(metrics: Collection<MetricData>): CompletableResultCode {
+        print(metrics)
```
0xnm, line 35: should this class be called `LogcatLogsMetricExporter` or this is some debugging leftover? If not, prefer using `Log.xxx` call.

tools/benchmark/src/main/java/com/datadog/benchmark/ext/FileExt.kt
```diff
@@ -0,0 +1,91 @@
...
+ * - isFile()
+ * - isDir()
+ * - listFiles(…)
+ * - length()
+ * The following set of extension make sure that every call to those methods
+ * is safeguarded to avoid crashing the customer's app.
+ */
+
+@Suppress("TooGenericExceptionCaught", "SwallowedException")
+private fun <T> File.safeCall(
```
0xnm, line 27: do we need these methods? they are originally were developed for the SDK usage, so that exception is not propagated to the client code. But if I get it right, this benchmark tooling is just internal tooling? So we probably don't care?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/DatadogMetricProfiler.kt
```diff
@@ -0,0 +1,106 @@
...
+            val currentFrameTime = System.nanoTime()
+            val elapsedTime: Long = currentFrameTime - lastFrameTime
+
+            if (elapsedTime >= TimeUnit.MILLISECONDS.toNanos(intervalMs)) {
+                val fps: Double = frameCount / (elapsedTime / NANO_IN_SECOND)
+                currentFps = fps.toInt()
+                lastFrameTime = currentFrameTime
+                frameCount = 0
+            }
+            Choreographer.getInstance().postFrameCallback(this)
```
0xnm, line 45: why Choreographer is chosen over what `androidx.metrics` gives?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/FpsVitalReader.kt
```diff
@@ -0,0 +1,62 @@
...
+import java.util.concurrent.TimeUnit
+
+internal class FpsVitalReader : VitalReader {
+
+    private var currentFps: Double = 0.0
+    private var frameCount = 0
+    private var lastFrameTime: Long = 0
+    private val intervalMs = FPS_SAMPLE_INTERVAL_IN_MS
+
+    private val frameCallback = object : Choreographer.FrameCallback {
```
0xnm, line 19: same question: why `Choreographer` over `androidx.metrics`?

tools/benchmark/src/main/java/com/datadog/benchmark/exporter/LogsMetricExporter.kt
ambushwork: yes, the purpose is to print the metrics to logcat, and in`print` function it's using `Log.i`

tools/benchmark/src/main/java/com/datadog/benchmark/ext/FileExt.kt
ambushwork: Indeed, the whole file will be removed

tools/benchmark/src/main/java/com/datadog/benchmark/internal/DatadogMetricProfiler.kt
ambushwork: I didn't know `androidx.metrics` can do the same thing, is there any advantage of `android.metrics` comparing to  `Choreographer`?

tools/benchmark/src/main/java/com/datadog/benchmark/exporter/LogsMetricExporter.kt
0xnm: let's, maybe, make it explicit by using `Log` directly? it will allow to filter by tag and also we will know log level.

tools/benchmark/src/main/java/com/datadog/benchmark/internal/DatadogMetricProfiler.kt
0xnm: yes, a lot of advantages: it a wrapper around different new APIs allowing collecting more granular metrics. Also by registering `Choreographer` callback and not un-registering it when app goes into background/losing focus, you make CPU running more (this point probably doesn't matter a lot for internal tooling).

Anyway, FPS is not the good metric to judge about the performance and we were discussion to ditch it and replace it with something else (especially considering dynamic refresh rate). Much more meaningful metric will be something like the % of skipped/junky frames, for example.

ambushwork: Good to know, I will first make a pull request to clean up the code on all the places where you mentioned, and another dedicated commit to replace the `Choreographer`

0xnm: Overall maybe for this work we shouldn't use FPS as a metric at all. But this is just my opinion.

ambushwork: Maybe you are right, with current synthetics test and metrics result, we can hardly see any difference between baseline and instrumented test, I will try to see if `androidx.metrics` can provide some more valuable result.

PR #2158, opened 2024-08-01
Make sure user properties are immutable when setUserInfo

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
@@ -11,7 +11,6 @@ import android.os.Build
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.context.NetworkInfo
 import com.datadog.android.api.context.TimeInfo
-import com.datadog.android.api.context.UserInfo
 import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.feature.FeatureContextUpdateReceiver
 import com.datadog.android.api.feature.FeatureEventReceiver
@@ -225,14 +224,7 @@ internal class DatadogCoreTest {
         testedCore.setUserInfo(id, name, email, fakeUserProperties)
 
         // Then
-        verify(mockUserInfoProvider).setUserInfo(
-            UserInfo(
-                id = id,
-                name = name,
-                email = email,
-                additionalProperties = fakeUserProperties
-            )
-        )
+        verify(mockUserInfoProvider).setUserInfo(id, name, email, fakeUserProperties)
     }
 
     @Test
@@ -258,11 +250,10 @@ internal class DatadogCoreTest {
 
         // Then
         verify(mockUserInfoProvider).setUserInfo(
-            UserInfo(
-                id,
-                name,
-                email
-            )
+            id,
+            name,
+            email,
+            emptyMap()
         )
         verify(mockUserInfoProvider).addUserProperties(
             properties = mapOf(
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
@@ -11,8 +11,11 @@ import com.datadog.android.core.internal.persistence.DataWriter
 import com.datadog.android.utils.forge.Configurator
 import com.datadog.tools.unit.forge.exhaustiveAttributes
 import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.annotation.AdvancedForgery
 import fr.xgouchet.elmyr.annotation.Forgery
+import fr.xgouchet.elmyr.annotation.MapForgery
 import fr.xgouchet.elmyr.annotation.StringForgery
+import fr.xgouchet.elmyr.annotation.StringForgeryType
 import fr.xgouchet.elmyr.junit5.ForgeConfiguration
 import fr.xgouchet.elmyr.junit5.ForgeExtension
 import org.assertj.core.api.Assertions.assertThat
@@ -29,7 +32,7 @@ import org.mockito.quality.Strictness
 @MockitoSettings(strictness = Strictness.LENIENT)
 internal class DatadogUserInfoProviderTest {
 
-    lateinit var testedProvider: DatadogUserInfoProvider
+    private lateinit var testedProvider: DatadogUserInfoProvider
 
     @Mock
     lateinit var mockWriter: DataWriter<UserInfo>
@@ -53,7 +56,7 @@ internal class DatadogUserInfoProviderTest {
         @Forgery userInfo: UserInfo
     ) {
         // When
-        testedProvider.setUserInfo(userInfo)
+        testedProvider.setUserInfo(userInfo.id, userInfo.name, userInfo.email, userInfo.additionalProperties)
         val result = testedProvider.getUserInfo()
 
         // Then
@@ -63,24 +66,20 @@ internal class DatadogUserInfoProviderTest {
     @Test
     fun `M delegate to persister W setUserInfo`(@Forgery userInfo: UserInfo) {
         // When
-        testedProvider.setUserInfo(userInfo)
+        testedProvider.setUserInfo(userInfo.id, userInfo.name, userInfo.email, userInfo.additionalProperties)
 
         // Then
         verify(mockWriter).write(userInfo)
     }
 
     @Test
-    fun `M keep existing properties W setExtraProperties() is called`(
+    fun `M keep existing properties W addUserProperties() is called`(
         @Forgery userInfo: UserInfo,
         @StringForgery forge: Forge
     ) {
         // Given
         val customProperties = forge.exhaustiveAttributes()
-        testedProvider.setUserInfo(
-            userInfo.copy(
-                additionalProperties = customProperties
-            )
-        )
+        testedProvider.setUserInfo(userInfo.id, userInfo.name, userInfo.email, customProperties)
 
         // When
         testedProvider.addUserProperties(customProperties)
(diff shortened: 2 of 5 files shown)


ambushwork (approved):
LGTM!

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -235,6 +235,36 @@ internal class DatadogCoreTest {
...
+        @StringForgery name: String,
+        @StringForgery(regex = "\\w+@\\w+") email: String,
+        @MapForgery(
+            key = AdvancedForgery(string = [StringForgery(StringForgeryType.ALPHA_NUMERICAL)]),
+            value = AdvancedForgery(string = [StringForgery(StringForgeryType.ALPHA_NUMERICAL)])
+        ) fakeUserProperties: Map<String, String>
+    ) {
+        // Given
+        val fakeMutableUserProperties = fakeUserProperties.toMutableMap()
+        val fakeExpectedUserProperties = fakeMutableUserProperties.toMap()
```
xgouchet, line 251: > [!NOTE]
> Probably nitpick, but I'd use the same source of truth here. That way we don't do the same implementation as in the production code. 

```suggestion
        val fakeExpectedUserProperties = fakeUserProperties.toMap()
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -235,6 +235,36 @@ internal class DatadogCoreTest {
...
+        val mockUserInfoProvider = mock<MutableUserInfoProvider>()
+        testedCore.coreFeature.userInfoProvider = mockUserInfoProvider
+        testedCore.setUserInfo(id, name, email, fakeMutableUserProperties)
+
+        // When
+        fakeMutableUserProperties.keys.forEach { key ->
+            fakeMutableUserProperties[key] = forge.anAlphaNumericalString()
+        }
```
xgouchet, line 259: > [!IMPORTANT]
> We're only testing mutating values for existing keys, but not removal / additions to the mutable map. We need to make sure that any kind of mutation to the original map will not impact the user properties.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -235,6 +235,36 @@ internal class DatadogCoreTest {
         )
     }
 
+    @Test
+    fun `M use immutable values W setUserInfo()`(
```
xgouchet, line 239: > [!IMPORTANT]
> We don't have the bug right now but it might be relevant to also test that we don't allow mutation when calling `addUserProperties()`

xgouchet (changes requested):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -235,6 +235,67 @@ internal class DatadogCoreTest {
...
+    ) {
+        // Given
+        val fakeMutableUserProperties = fakeUserProperties.toMutableMap()
+        val fakeExpectedUserProperties = fakeUserProperties.toMap()
+        val mockUserInfoProvider = mock<MutableUserInfoProvider>()
+        testedCore.coreFeature.userInfoProvider = mockUserInfoProvider
+        testedCore.setUserInfo(id, name, email, fakeMutableUserProperties)
+
+        // When
+        fakeMutableUserProperties.values.drop(forge.anInt(0, fakeMutableUserProperties.size))
```
xgouchet, line 287: > [!CAUTION]
> Calling the `values.drop()` doesn't modify the map, only the list of values. You need to call `m.remove(forge.anElementFrom(m.keys))`

> [!NOTE]
> Even if the drop worked, here we would only drop the 'first' n elements. Meaning that you would not remove stuff in the 'middle' of the map. It's not that relevant as Maps aren't ordered, but in some cases it could have some importance.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
```diff
@@ -90,6 +90,49 @@ internal class DatadogUserInfoProviderTest {
         ).isEqualTo(customProperties)
     }
 
+    @Test
+    fun `M use immutable properties W setExtraProperties() is called { changing properties values }`(
```
xgouchet, line 94:
```suggestion
    fun `M use immutable properties W addUserProperties() is called { changing properties values }`(
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
```diff
@@ -90,6 +90,49 @@ internal class DatadogUserInfoProviderTest {
...
+        }
+
+        // Then
+        assertThat(
+            testedProvider.getUserInfo().additionalProperties
+        ).isEqualTo(fakeExpectedProperties)
+    }
+
+    @Test
+    fun `M use immutable properties W setExtraProperties() is called { adding or removing properties }`(
```
xgouchet, line 115:
```suggestion
    fun `M use immutable properties W addUserProperties() is called { adding or removing properties }`(
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
```diff
@@ -90,6 +90,49 @@ internal class DatadogUserInfoProviderTest {
...
+        @StringForgery forge: Forge
+    ) {
+        // Given
+        val fakeProperties = forge.exhaustiveAttributes()
+        val fakeExpectedProperties = fakeProperties.toMap()
+        val fakeMutableProperties = fakeProperties.toMutableMap()
+        testedProvider.addUserProperties(fakeMutableProperties)
+
+        // When
+        fakeMutableProperties.values.drop(forge.anInt(0, fakeMutableProperties.size))
```
xgouchet, line 125: > [!CAUTION]
> Same remark as above

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
xgouchet, line 94: > [!WARNING]
> Why are you testing the `setUserInfo()` on the `DatadogCore` class, but the  `addUserProperties()` on the `DatadogUserInfoProvider` class? Both methods exist on both classes, and it might be relevant to test the same behavior on both classes.

xgouchet (changes requested):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
mariusc83: because the immutability is handled on different places. The `addUserProperties` is handled inside the `DatadogUserInfoProvider` and that's a mock in the `DatadogCoreTesting`. There are 2 different logics that we are testing here.

xgouchet: IMO We need to test both. The DatadogCore should make sure that we don't transfer the immutability as it doesn't know how the UserInfoProvider works, and same from the DatadogUserInfoProvider point of view

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -235,6 +235,94 @@ internal class DatadogCoreTest {
...
```
xgouchet, line 317: > [!CAUTION]
> Calling the `values.drop()` doesn't modify the map, only the list of values. You need to call `m.remove(forge.anElementFrom(m.keys))`

> [!NOTE]
> Even if the drop worked, here we would only drop the 'first' n elements. Meaning that you would not remove stuff in the 'middle' of the map. It's not that relevant as Maps aren't ordered, but in some cases it could have some importance.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
```diff
@@ -90,6 +90,67 @@ internal class DatadogUserInfoProviderTest {
...
```
xgouchet, line 146: > [!CAUTION]
> Calling the `values.drop()` doesn't modify the map, only the list of values. You need to call `m.remove(forge.anElementFrom(m.keys))`

> [!NOTE]
> Even if the drop worked, here we would only drop the 'first' n elements. Meaning that you would not remove stuff in the 'middle' of the map. It's not that relevant as Maps aren't ordered, but in some cases it could have some importance.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
mariusc83: I don't agree with handling this in 2 places. Testing this in the DatadogCore is not possible as the DatadogCore only delegates to `DatadodUserInfoProvider` which already creates a copy from `original` + `new properties`. If we choose to handle this here also it will create an extra copy. I don't want to exchange tests for less optimised code.

xgouchet: In this case let's put all the check in the userInfoProvider. It doesn't make sense to have the checks spread in two locations

xgouchet (approved):

PR #2160, opened 2024-08-02
Json schema update

xgouchet:
### What does this PR do?

This updates our Kotlin data class generation from JsonSchema
- Handle nested `one_of` descriptions
- Better handling of `constant` properties: we now check the constant has the expected value

### DIFF
--- a/buildSrc/src/test/kotlin/com/example/model/WeirdCombo.kt
+++ b/buildSrc/src/test/kotlin/com/example/model/WeirdCombo.kt
@@ -17,19 +17,13 @@ import kotlin.collections.List
 import kotlin.jvm.JvmStatic
 import kotlin.jvm.Throws
 
-public data class Household(
-    public val pets: List<Animal>? = null,
-    public val situation: Situation? = null,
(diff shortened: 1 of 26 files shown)


buildSrc/src/test/kotlin/com/example/model/WeirdCombo.kt
```diff
@@ -0,0 +1,335 @@
...
+                        throw JsonParseException(
+                            "Unable to parse json into type Bird",
+                            e
+                        )
+                    }
+                }
+            }
+        }
+
+        public data class Paper(
```
mariusc83, line 184: shouldn't we generate `sealed` classes here instead ?

xgouchet: The `Anything` class is already sealed

mariusc83: my bad didn't spot that. that's perfect. Nice job here !!

mariusc83 (approved):

PR #2168, opened 2024-08-05
Fix Benchmark baseline run bug and add unit test

ambushwork:
### What does this PR do?

The session replay was wrongly enabled when the "run" is "baseline",
The PR fixes it by checking the "run" condition.

Unit test is added to protect it.

### DIFF
--- a/tools/benchmark/src/main/java/com/datadog/benchmark/exporter/DatadogMetricExporter.kt
+++ b/tools/benchmark/src/main/java/com/datadog/benchmark/exporter/DatadogMetricExporter.kt
@@ -13,11 +13,11 @@ import io.opentelemetry.sdk.metrics.export.MetricExporter
 internal class DatadogMetricExporter(datadogExporterConfiguration: DatadogExporterConfiguration) : MetricExporter {
 
     private val metricContext: MetricContext = MetricContext(
-        deviceModel = Build.MODEL,
-        osVersion = Build.VERSION.RELEASE,
-        run = datadogExporterConfiguration.run ?: DEFAULT_RUN_NAME,
+        deviceModel = Build.MODEL ?: UNKNOWN_TAG_VALUE,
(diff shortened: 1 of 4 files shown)


tools/benchmark/src/main/java/com/datadog/benchmark/exporter/DatadogMetricExporter.kt
```diff
@@ -46,6 +46,8 @@ internal class DatadogMetricExporter(datadogExporterConfiguration: DatadogExport
     }
 
     companion object {
+        private const val DEFAULT_MODEL_NAME = "unknown model"
```
mariusc83, line 49: maybe we should call these constants `UNKNOWN_...`

mariusc83 (approved):

tools/benchmark/src/main/java/com/datadog/benchmark/exporter/DatadogMetricExporter.kt
ambushwork: Actually the text here should be displayed on the web, as the tag value such as "device: unknown model", 
that's why I use plain text, do you prefer we use upper case with underscore on this?

image

mariusc83: nope I was referring to the variable name UNKNOWN instead of DEFAULT as prefix

mariusc83 (approved):

tools/benchmark/src/main/java/com/datadog/benchmark/exporter/DatadogMetricExporter.kt
ambushwork: OK,  then you are right, I updated the variable name, and now all attributes share the same unknown value.

mariusc83 (approved):

PR #2170, opened 2024-08-07
RUM-5750 Use NO_EXPORT_FLAG for BroadcastReceiver on API above 26

mariusc83:
### What does this PR do?

This issue was reported by a customer. Their security check tool detected this issue in our SDK where we do not register a receiver with NO_EXPORT_FLAG in API 33 to 26. This method actually exists also below API 33 up to API 26 but the flag Context.NO_EXPORT_FLAG was not available so that's the reason we skipped it. 

I chose not to use the `ContextCompat.NO_EXPORT_FLAG` just to not have to introduce this new dependency just for this case and added a `0x4` local flag myself (same value as the one in Context and ContextCompat). The worse thing it could happen is that this flag to change later but I checked and the SDK is not crashing it is just ignoring it. On the other side if the flag value (logic will change under the hood) we might end up with a different access rights for this receiver exposing again a security issue. I think this is highly unlikely but reviewers please have this in mind.
### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/receiver/ThreadSafeReceiver.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/receiver/ThreadSafeReceiver.kt
@@ -18,13 +18,16 @@ internal abstract class ThreadSafeReceiver : BroadcastReceiver() {
 
     val isRegistered = AtomicBoolean(false)
 
-    @SuppressLint("UnspecifiedRegisterReceiverFlag")
+    // We suppress the warning here as this method is not available on all Android versions
+    @SuppressLint("WrongConstant", "UnspecifiedRegisterReceiverFlag")
     fun registerReceiver(
         context: Context,
         filter: IntentFilter
     ): Intent? {
         val intent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
             context.registerReceiver(this, filter, Context.RECEIVER_NOT_EXPORTED)
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            context.registerReceiver(this, filter, RECEIVER_NOT_EXPORTED_COMPAT)
         } else {
             context.registerReceiver(this, filter)
         }
@@ -36,5 +39,9 @@ internal abstract class ThreadSafeReceiver : BroadcastReceiver() {
         if (isRegistered.compareAndSet(true, false)) {
             context.unregisterReceiver(this)
         }
+    }
+
+    companion object {
+        internal const val RECEIVER_NOT_EXPORTED_COMPAT: Int = 0x4
     }
 }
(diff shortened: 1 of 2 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/receiver/ThreadSafeReceiver.kt
```diff
@@ -18,13 +18,15 @@ internal abstract class ThreadSafeReceiver : BroadcastReceiver() {
...
 
-    @SuppressLint("UnspecifiedRegisterReceiverFlag")
+    @SuppressLint("WrongConstant", "UnspecifiedRegisterReceiverFlag")
     fun registerReceiver(
         context: Context,
         filter: IntentFilter
     ): Intent? {
         val intent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
             context.registerReceiver(this, filter, Context.RECEIVER_NOT_EXPORTED)
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
mariusc83, line 28: Normally we should use ContextCompat here:
https://developer.android.com/reference/androidx/core/content/ContextCompat#RECEIVER_NOT_EXPORTED() 
but not sure if we should bring the ContextCompat dependency only for this reason. The drawback on this approach would be in case that Flag value will be changed in the future (quite unlikely) . I am not sure we should use ContextCompat  only for this reason. What do you think ?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/receiver/ThreadSafeReceiverTest.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    @Mock
+    lateinit var mockIntentFilter: IntentFilter
+
+    @BeforeEach
+    fun `set up`() {
+        testedReceiver = TestableThreadSafeReceiver()
+    }
+
+    // r
```
jonathanmos, line 53: this comment looks like it got redacted

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/receiver/ThreadSafeReceiverTest.kt
```diff
@@ -0,0 +1,133 @@
...
+        // When
+        testedReceiver.registerReceiver(mockContext, mockIntentFilter)
+
+        // Then
+        verify(mockContext).registerReceiver(
+            testedReceiver,
+            mockIntentFilter,
+            ThreadSafeReceiver.RECEIVER_NOT_EXPORTED_COMPAT
+        )
+        assertThat(this.testedReceiver.isRegistered.get()).isTrue()
```
0xnm, line 67: minor: `this` can be removed. Same for other similar lines in this file.
```suggestion
        assertThat(testedReceiver.isRegistered.get()).isTrue()
```

0xnm (approved):

PR #2173, opened 2024-08-08
Fix ButtonCompositionGroupMapper crash when calculating the corner radius

ambushwork:
### What does this PR do?

Fix the crash caused by ButtonCompositionGroupMapper when calculating the corner radius

### Motivation

A crash is found when I test with an Android open source application,

here is the trace:

```
 java.lang.IllegalStateException: Size is unspecified
       at androidx.compose.ui.geometry.Size.getWidth-impl(Size.kt:48)
       at androidx.compose.ui.geometry.Size.getMinDimension-impl(Size.kt:128)
       at androidx.compose.foundation.shape.PercentCornerSize.toPx-TmRCtEA(CornerSize.kt:100)
       at com.datadog.android.sessionreplay.compose.internal.mappers.ButtonCompositionGroupMapper.map(ButtonCompositionGroupMapper.kt:42)
       at com.datadog.android.sessionreplay.compose.internal.mappers.AbstractCompositionGroupMapper.map(AbstractCompositionGroupMapper.kt:37)
```

### Demo

In the demo we can see the corner radius of the button is correct

| Application screen | Session Replay Screen|
| ---- | ---- |
|image| image|


features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/ButtonCompositionGroupMapper.kt
```diff
@@ -33,13 +33,19 @@ internal class ButtonCompositionGroupMapper(
         var cornerRadius: Number? = null
         var colors: ButtonColors? = null
 
+        // Size of the button must be specified in order to calculate the corner radius,
+        // when [CornerSize] is a percentage.
```
mariedm, line 37: In which cases can a `CornerSize` be a percentage (or not)?

ambushwork: It totally depends on  the choice of our clients (app developers) how they want to implement UI, either they can use a corner with specific dimension size, so no matter what the size of the button is, the corner keeps same. Either they can use percentage like I show in the screenshot, so the corner radius adapt with button size.

In this case, we must support both of them.

mariusc83 (approved):
Good catch

PR #2192, opened 2024-08-13
Update Gradle to 8.9 and AGP to 8.5.2

0xnm:
### What does this PR do?

Since we are now using Kotlin 1.9, we can also update Gradle and AGP to the newer versions (which were bringing Kotlin 1.9 to the classpath).


features/dd-sdk-android-trace/src/main/java/com/datadog/exec/CommonTaskExecutor.java
```diff
@@ -66,8 +66,14 @@ public <T> ScheduledFuture<?> scheduleAtFixedRate(
...
         final PeriodicTask<T> periodicTask = new PeriodicTask<>(task, target);
+
+        // was scheduleAtFixedRate before, but Android Lint gives this:
+        // Error: Use of scheduleAtFixedRate is strongly discouraged because it can lead to
+        // unexpected behavior when Android processes become cached (tasks may unexpectedly
+        // execute hundreds or thousands of times in quick succession when a process
+        // changes from cached to uncached); prefer using scheduleWithFixedDelay [DiscouragedApi]
         final ScheduledFuture<?> future =
-            executorService.scheduleAtFixedRate(
+            executorService.scheduleWithFixedDelay(
```
ambushwork, line 76: In this case should we change the function name at L58 since it's wrapper of it?

.gitlab-ci.yml
```diff
@@ -4,7 +4,7 @@ include:
 # SETUP
 
 variables:
-  CURRENT_CI_IMAGE: "9"
+  CURRENT_CI_IMAGE: "10"
```
ambushwork, line 7: By curiosity, what does it stand for?

ambushwork (approved):

features/dd-sdk-android-trace/src/main/java/com/datadog/exec/CommonTaskExecutor.java
0xnm: I would keep it the same to avoid more code changes. Ideally they have the same behavior - some task executed periodically, just the tiny details are different - in the former case it will be scheduled even if previous task is not yet completed (I believe), while in the latter case it won't.

.gitlab-ci.yml
0xnm: It is just the tag for the Docker image. We usually bump it, so that we don't overwrite existing one. If we overwrite, there is a chance that builds from older branches won't work with a newer Docker image (because of some env incompatibility), so it is better to stay on the safer side.

jonathanmos (approved):

PR #2194, opened 2024-08-14
RUM-5754: Add profiling interface to internal module

ambushwork:
### What does this PR do?
Adds following interface into SDK core:

* BenchmarkCounter
* BenchmarkGauge
* BenchmarkProfiler
* BenchmarkSpan
* BenchmarkTracer

Adds an object class to for BenchmarkProfiler
* GlobalBenchmark

### Motivation
RUM-5754

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkProfiler.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkProfiler.kt
@@ -1,0 +1,22 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiler
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface of benchmark profiler to be implemented to provide [BenchmarkTracer].
+ * This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+interface BenchmarkProfiler {
+
+    /**
+     * Returns a [BenchmarkTracer].
+     */
+    fun getTracer(operation: String): BenchmarkTracer
+}
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpan.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpan.kt
@@ -1,0 +1,21 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiler
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface of benchmark span to be implemented. This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+interface BenchmarkSpan {
+
+    /**
+     * Marks the end of [BenchmarkSpan] execution.
+     */
+    fun stop()
+}
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkTracer.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkTracer.kt
@@ -1,0 +1,25 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiler
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface of benchmark tracer to be implemented to provide [BenchmarkSpan].
+ * This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+interface BenchmarkTracer {
+
+    /**
+     * Returns a new [BenchmarkSpan].
+     *
+     * @param spanName The name of the returned span.
+     * @return a new [BenchmarkSpan].
+     */
+    fun startSpan(spanName: String): BenchmarkSpan
+}
--- a/dd-sdk-android-internal/api/apiSurface
+++ b/dd-sdk-android-internal/api/apiSurface
@@ -1,2 +1,11 @@
+interface com.datadog.android.internal.profiler.BenchmarkProfiler
+  fun getTracer(String): BenchmarkTracer
+interface com.datadog.android.internal.profiler.BenchmarkSpan
+  fun stop()
+interface com.datadog.android.internal.profiler.BenchmarkTracer
+  fun startSpan(String): BenchmarkSpan
+object com.datadog.android.internal.profiler.GlobalBenchmark
+  fun register(BenchmarkProfiler)
+  fun get(): BenchmarkProfiler
 annotation com.datadog.tools.annotation.NoOpImplementation
   constructor(Boolean = false)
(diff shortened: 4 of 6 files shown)


0xnm:
I'm not sure the approach we are taking here: why this is added to the SDK core module? Where implementations are supposed to be?

The main issue is that SDK core is what is distributed to the clients, while we are working just on the internal tooling to benchmark the SDK (it is not supposed to be published). So we are adding something for the internal needs and this is being added to the binary used by the clients - they don't need it.

Also once it is published, it may give a wrong impression on the client side that we support metrics in the SDK, which is not true.

ambushwork:
@0xnm I am not supposed to share the direct link of confluence here, so please search for the "[RFC] Session Replay Performance Metrics" for more detailed context.

Briefly speaking, the reason why interfaces are defined inside the `core` is because we need to use directly in our sdk code (core and features) to measure the function span. And the real implementations will be in `tools/benchmark` module.  

So on client side, they are not supposed to import `tools/benchmark` module, so all the code we plant for the profiling will execute the No-op implementation. And all the interfaces are marked as `@InternalApi` in case of external usages.

On our side,  the benchmark application depends on `tools/benchmark`, with `GlobalBenchmark` object, it is able to register the real implementation.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/profiler/BenchmarkCounter.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import com.datadog.android.lint.InternalApi
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface of benchmark counter to be implemented to perform count metrics.
+ * This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+@InternalApi
```
jonathanmos, line 17: >Warning
Documentation for InternalApi notes: "Don't use this annotation on interfaces". Is this intentional? Same question for the other interfaces in the PR.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/profiler/GlobalBenchmark.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+@InternalApi
+object GlobalBenchmark {
+
+    private var benchmarkProfiler: BenchmarkProfiler = NoOpBenchmarkProfiler()
+
+    /**
+     * Registers the implementation of [BenchmarkProfiler].
+     */
+    @InternalApi
```
jonathanmos, line 23: InternalApi is both on the singleton and on the method. Doesn't the annotation at the class level preclude adding at the method level?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/profiler/BenchmarkCounter.kt
ambushwork: The reason given by the documentation why it shouldn't be used on interface is that :
> because implementation will be flagged as internal as a whole
Actually what I want is exactly that all the implementations from this interface should be internal.

But in case of confusion of the documentation, I removed all the `@InternalApi` on interface, I will put them manually on all the implementations, nice catch here!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/profiler/GlobalBenchmark.kt
ambushwork: You are right, removed.

mariusc83 (approved):

dd-sdk-android-internal/build.gradle.kts
```diff
@@ -38,6 +39,10 @@ android {
 
 dependencies {
     implementation(libs.kotlin)
+
+    // Generate NoOp implementations
+    ksp(project(":tools:noopfactory"))
+    implementation(project(":tools:noopfactory"))
```
0xnm, line 45: we shouldn't have this as a runtime dependency

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkCounter.kt
```diff
@@ -0,0 +1,16 @@
...
+package com.datadog.android.internal.profiler
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface of benchmark counter to be implemented to perform count metrics.
+ * This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+interface BenchmarkCounter
```
0xnm, line 16: why do we have an interface without any methods? Is it useful to have such at the current point?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkGauge.kt
```diff
@@ -0,0 +1,16 @@
...
+package com.datadog.android.internal.profiler
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface of benchmark gauge to be implemented to perform gauge metrics.
+ * This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+interface BenchmarkGauge
```
0xnm, line 16: same question as above

0xnm (changes requested):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkCounter.kt
ambushwork: It's reserved for the future development according to the RFC, but you are right we can remove it since we don't have idea what are the methods inside.

ambushwork:
PR is updated, now the interfaces are added into `internal` module.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkProfiler.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiler
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Interface of benchmark profiler to be implemented to provide [BenchmarkTracer], [BenchmarkCounter]
+ * and [BenchmarkGauge]. This should only used by internal benchmarking.
```
0xnm, line 13: there is no `BenchmarkCounter` and `BenchmarkGauge`

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpan.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Interface of benchmark span to be implemented.
+ * This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+interface BenchmarkSpan {
+
+    /**
+     * Marks the end of {@code Span} execution.
```
0xnm, line 19: updated to match KDoc formatting for class references

```suggestion
     * Marks the end of [Span] execution.
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkTracer.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Interface of benchmark tracer to be implemented to provide [BenchmarkSpan].
+ * This should only used by internal benchmarking.
+ */
+@NoOpImplementation
+interface BenchmarkTracer {
+
+    /**
+     * Returns a new [BenchmarkSpan].
+     *
+     * @param spanName The name of the returned Span.
```
0xnm, line 21:
```suggestion
     * @param spanName The name of the returned span.
```

0xnm (dismissed):
I've added few suggestions, but nothing blocking.

xgouchet (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkProfiler.kt
ambushwork: updated

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpan.kt
ambushwork: updated

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkTracer.kt
ambushwork: updated

PR #2196, opened 2024-08-15
RUM-5763: Add TouchPrivacy to Session Replay

jonathanmos:
### What does this PR do?
Adds a public TouchPrivacy api

Possible values for touch:
`SHOW` - record all touch interactions
`HIDE` - do not record touch interactions

Default value will be HIDE.

Usage example:
```
SessionReplayConfiguration.Builder() 
  .setTouchPrivacy(TouchPrivacy.SHOW)
```

Note: At the moment this will not hide the change in a button background that occurs as a result of a click (or other view hierarchy side effects). 

### Motivation
Part of the move to fine grained privacy options.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
@@ -20,7 +20,8 @@ data class SessionReplayConfiguration internal constructor(
     internal val customOptionSelectorDetectors: List<OptionSelectorDetector>,
     internal val sampleRate: Float,
     internal val imagePrivacy: ImagePrivacy,
-    internal val startRecordingImmediately: Boolean
+    internal val startRecordingImmediately: Boolean,
+    internal val touchPrivacy: TouchPrivacy
 ) {
 
     /**
@@ -33,6 +34,7 @@ data class SessionReplayConfiguration internal constructor(
         private var privacy = SessionReplayPrivacy.MASK
         private var imagePrivacy = ImagePrivacy.MASK_ALL
         private var startRecordingImmediately = true
+        private var touchPrivacy = TouchPrivacy.HIDE
         private var extensionSupport: ExtensionSupport = NoOpExtensionSupport()
 
         /**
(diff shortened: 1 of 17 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -76,6 +78,11 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        internal fun setTouchPrivacy(level: TouchPrivacy): Builder {
```
mariusc83, line 81: why internal ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
```diff
@@ -51,13 +53,15 @@ internal class RecorderWindowCallback(
...
     override fun dispatchTouchEvent(event: MotionEvent?): Boolean {
         if (event != null) {
-            // we copy it and delegate it to the gesture detector for analysis
-            @Suppress("UnsafeThirdPartyFunctionCall") // internal safe call
-            val copy = copyEvent(event)
-            try {
-                handleEvent(copy)
-            } finally {
-                copy.recycle()
+            if (touchPrivacy == TouchPrivacy.SHOW) {
```
mariusc83, line 56: this API is only set in the configuration so it is immutable. I think instead we should make sure we never register a touch recorder in case this is `HIDE`. There's no interest in keeping it and is better optimised this way.

mariusc83 (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
jonathanmos: We don't want to make these apis public until all of fgm is completed, however I've decided that it's better that I open a feature branch and merge these prs into that, and then this could be public in that branch

mariusc83 (approved):
LGTM

0xnm (approved):

ambushwork (approved):
LGTM!

PR #2197, opened 2024-08-16
Fix integration tests pipeline for API 21

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


.gitlab-ci.yml
```diff
@@ -49,9 +49,13 @@ stages:
...
     - yes | ~/android_sdk/cmdline-tools/latest/bin/sdkmanager --licenses || true
     - echo "no" | ~/android_sdk/cmdline-tools/latest/bin/avdmanager --verbose create avd --force --name "$EMULATOR_NAME" --package "$ANDROID_EMULATOR_IMAGE"
-  run-instrumented-integration:
+  compile_instrumented_integration:
     - $ANDROID_HOME/emulator/emulator -avd "$EMULATOR_NAME" -grpc-use-jwt -no-snapstorage -no-audio -no-window -no-boot-anim -verbose -qemu -machine virt &
     - GRADLE_OPTS="-Xmx3072m" ./gradlew :instrumented:integration:assembleDebug :instrumented:integration:assembleDebugAndroidTest --stacktrace --no-daemon
+  compile_instrumented_integration_for_api_21:
+    - $ANDROID_HOME/emulator/emulator -avd "$EMULATOR_NAME" -grpc-use-jwt -no-snapstorage -no-audio -no-window -no-boot-anim -verbose -qemu -machine virt &
+    - GRADLE_OPTS="-Xmx3072m" ./gradlew :instrumented:integration:assembleDebug :instrumented:integration:assembleDebugAndroidTest -Puse-api21-java-backport -Puse-desugaring --stacktrace --no-daemon
+  run-instrumented-integration:
```
0xnm, line 58: you can still keep `run-instrumented-integration` step without splitting and use the same technique to append `-Puse-api21-java-backport -Puse-desugaring` flags as I used in https://github.com/DataDog/dd-sdk-android/pull/2193

also do we need `use-desugarding` flag? seems we can simply use `use-api21-java-backport`

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/sessionreplay/chekboxandradio/SrCheckBoxAndRadioFieldsAllowTest.kt
```diff
@@ -27,6 +28,7 @@ internal class SrCheckBoxAndRadioFieldsAllowTest :
     )
 
     @Test
+    @Ignore("These tests were not maintained anymore and they need to be fixed")
```
0xnm, line 31: we need to add TODO reference everywhere we add this annotation

tools/javabackport/src/main/kotlin/java/nio/file/Path.kt
0xnm, line 1: do we need to add this if we add desugaring?

tools/javabackport/src/main/kotlin/java/util/concurrent/atomic/LongAdder.kt
0xnm, line 1: do we need to add this if we add desugaring?

mariusc83: yes because desugaring doesn't handle everything  (this and the Path class are not handled :( )

.gitlab-ci.yml
mariusc83: ok will try that technique instead. But regarding desugaring flag...I would rather keep them separated (one is for desugaring, the other is for java backport)

0xnm (approved):

PR #2199, opened 2024-08-19
Create internal module

ambushwork:
### What does this PR do?

Creates a new module "internal" in SDK

### Motivation

This module is created to contain some interfaces which should not be exposed to the clients, but need to be shared between each modules in SDK.


internal/.gitignore
0xnm, line 1: we should follow existing naming notation for the public modules (this one has to be published to Maven as well), where all of them start with `dd-sdk-android-xxx`. This includes both module name and also folder name.

Also it should be job in the Gitlab CI definition file to publish this module.

internal/build.gradle.kts
0xnm, line 1: we need to add plugin/configuration call similar to other public modules.

0xnm (changes requested):

internal/.gitignore
ambushwork: renamed it to `dd-sdk-android-internal`.

0xnm (approved):

PR #2205, opened 2024-08-20
RUM-5877 run core-it integration tests in Gitlab

mariusc83:
### What does this PR do?

This PR add the Gitlab steps for running the core-it instrumented tests in Gitlab using the Sonoma Mac OS runners.
We are also fixing the legacy instrumented tests Gitlab execution scripts which were not failing the job if the tests were failing.

Mainly the issue with the Gitlab scripts was that Gitlab runs them through `bash` and Sonoma forces to `zsh` therefore it runs each line in isolation meaning that the `exit code` values is being reset from one line to another.

### Motivation

### Additional Notes

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -49,14 +49,25 @@ stages:
     - echo y | ~/android_sdk/cmdline-tools/latest/bin/sdkmanager --install "$ANDROID_EMULATOR_IMAGE"
     - yes | ~/android_sdk/cmdline-tools/latest/bin/sdkmanager --licenses || true
     - echo "no" | ~/android_sdk/cmdline-tools/latest/bin/avdmanager --verbose create avd --force --name "$EMULATOR_NAME" --package "$ANDROID_EMULATOR_IMAGE"
-  run-instrumented-integration:
+  run-legacy-integration-instrumented:
+    - set +e
+    - exit_code=0
     - $ANDROID_HOME/emulator/emulator -avd "$EMULATOR_NAME" -grpc-use-jwt -no-snapstorage -no-audio -no-window -no-boot-anim -verbose -qemu -machine virt &
     - GRADLE_OPTS="-Xmx3072m" ./gradlew :instrumented:integration:assembleDebug :instrumented:integration:assembleDebugAndroidTest --stacktrace --no-daemon $( (( $ANDROID_API <= 23 )) && echo "-Puse-api21-java-backport -Puse-desugaring" )
-    - set +e
     - $ANDROID_HOME/platform-tools/adb install -t -d $( (( $ANDROID_API >= 23 )) && echo "-g" ) -r instrumented/integration/build/outputs/apk/androidTest/debug/integration-debug-androidTest.apk
     - $ANDROID_HOME/platform-tools/adb install -t -d $( (( $ANDROID_API >= 23 )) && echo "-g" ) -r instrumented/integration/build/outputs/apk/debug/integration-debug.apk
-    - $ANDROID_HOME/platform-tools/adb shell am instrument -w com.datadog.android.sdk.integration.test/androidx.test.runner.AndroidJUnitRunner
+    - $ANDROID_HOME/platform-tools/adb shell am instrument -w com.datadog.android.sdk.integration.test/androidx.test.runner.AndroidJUnitRunner || exit_code=$?
     - $ANDROID_HOME/platform-tools/adb emu kill
+    - if [[ "$exit_code" -ne 0 ]]; then exit 1; fi
+    - exit 0
+  run-core-it-instrumented:
+    - set +e
+    - exit_code=0
+    - $ANDROID_HOME/emulator/emulator -avd "$EMULATOR_NAME" -grpc-use-jwt -no-snapstorage -no-audio -no-window -no-boot-anim -verbose -qemu -machine virt &
+    - GRADLE_OPTS="-Xmx3072m" ./gradlew :reliability:core-it:connectedDebugAndroidTest --stacktrace --no-daemon $( (( $ANDROID_API <= 23 )) && echo "-Puse-api21-java-backport -Puse-desugaring" ) || exit_code=$?
+    - $ANDROID_HOME/platform-tools/adb emu kill
+    - if [[ "$exit_code" -ne 0 ]]; then exit 1; fi
+    - exit 0
   set-publishing-credentials:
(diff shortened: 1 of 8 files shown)


.gitlab-ci.yml
```diff
@@ -50,13 +50,21 @@ stages:
...
   run-instrumented-integration:
+    - set +e
+    - exit_code=0
     - $ANDROID_HOME/emulator/emulator -avd "$EMULATOR_NAME" -grpc-use-jwt -no-snapstorage -no-audio -no-window -no-boot-anim -verbose -qemu -machine virt &
-    - GRADLE_OPTS="-Xmx3072m" ./gradlew :instrumented:integration:assembleDebug :instrumented:integration:assembleDebugAndroidTest --stacktrace --no-daemon $( (( $ANDROID_API <= 23 )) && echo "-Puse-api21-java-backport -Puse-desugaring" )
+    - GRADLE_OPTS="-Xmx3072m" ./gradlew :instrumented:integration:assembleDebug :instrumented:integration:connectedDebugAndroidTest --stacktrace --no-daemon $( (( $ANDROID_API <= 23 )) && echo "-Puse-api21-java-backport -Puse-desugaring" ) || exit_code=$?
+    - $ANDROID_HOME/platform-tools/adb emu kill
+    - if [[ "$exit_code" -ne 0 ]]; then exit 1; fi
+    - exit 0
+  run-core-it-instrumented:
```
0xnm, line 60: I am confused by what we have (mainly because of the naming/split) and have a few questions:

* what `it` in the `core-it` means? Is it `integration`? If yes, then we kind of have `run-instrumented-integration` snippet and `run-core-it[integration]-instrumented` snippet. Both of them are for integration + instrumented. Maybe we should replace `it` with full form for the clarity?
* As a consequence from the point above: what is the difference between the two? If we need to add a test which is both instrumented + integration, where should it go?

reliability/core-it/build.gradle.kts
```diff
@@ -13,21 +14,30 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 plugins {
     // Build
-    id("com.android.library")
+    id("com.android.application")
```
0xnm, line 17: in this case you need to remove `androidLibraryConfig` call below, because there is no `LibraryExtension` registered.

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/InternalSdkCoreContextTest.kt
```diff
@@ -134,7 +132,7 @@ class InternalSdkCoreContextTest : MockServerTest() {
         assertThat(context.appBuildId).isEqualTo(BUILD_ID)
         assertThat(context.source).isEqualTo(ANDROID_SOURCE)
         assertThat(context.processInfo.isMainProcess).isTrue()
-        assertThat(context.networkInfo.connectivity).isEqualTo(NetworkInfo.Connectivity.NETWORK_WIFI)
+        assertThat(context.networkInfo.connectivity).isNotNull()
```
0xnm, line 135: why this change is needed? isn't emulator on WiFi by default?

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/MockServerTest.kt
```diff
@@ -19,21 +20,28 @@ abstract class MockServerTest : BaseTest() {
...
-        protected val mockServerWrapper: MockWebServerWrapper = MockWebServerWrapper()
+        private val MOCK_SERVER_LOCK = Any()
+        protected var mockServerWrapper: MockWebServerWrapper = MockWebServerWrapper()
 
         @BeforeClass
         @JvmStatic
         fun setupTestSuite() {
             if (mockServerStarted.compareAndSet(false, true)) {
-                mockServerWrapper.start()
+                synchronized(MOCK_SERVER_LOCK) {
```
0xnm, line 30: Isn't test runner single-threaded?

reliability/core-it/build.gradle.kts
mariusc83: yup missed that

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/MockServerTest.kt
mariusc83: I am not 100% sure, my memory tells me that it is but just wanted to make sure to avoid any possible flackyness here. Should we remove the lock ?

.gitlab-ci.yml
mariusc83: the `instrumented-integration` are the old integration tests which we can find in the `instrumented/integration` package. They use to be running in Bitrise. I think from now on any integration test will be added in the `reliability` folder in one of the modules there. If you are asking about the difference between `instrumented` and `integration` we should not confuse these terms. `integration` is any test that tests a feature in its whole (different than `unit test`) whereas `instrumented` tells about the nature of the runner that executes that test, in our case that being the AndroidJUnitRunner which executes the integration tests on the Emulator itself.

mariusc83: regarding your comment I think I can re - name a bit the scripts there to make more sense.

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/InternalSdkCoreContextTest.kt
mariusc83: you would be surprised but sometimes, especially on lower APIs it returns 3G there :).

.gitlab-ci.yml
0xnm: My question is not about `instrumented` vs `integration`, my question is solely about 2 different modules (+ related jobs) which are both `instrumented` + `integration`.

I would also suggest renaming `core-it` module to expand the `it` term used there.

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/MockServerTest.kt
mariusc83: On the other hand I think it is safe to keep the lock. The instrumented tests are being executed in a different thread and the mockWebServer can be interacted with from the SDK uploader threads. I think it is better to have this here, it will not harm us anyway.

.gitlab-ci.yml
mariusc83: the `core-it` is more to be in sync with the previous `single-fit` (single feature integration test). Do you mean to have something like `core-it-instrumented` ?

0xnm (approved):

PR #2213, opened 2024-08-27
Migrate NoOpImplementation from core to internal module

ambushwork:
### What does this PR do?

Migrate NoOpImplementation from core to internal module

### Motivation

* `NoOpImplementation` is an internal annotation, it make senses to have it inside `internal` module instead of `core`.
* Move it into `internal`, so that the interfaces in `internal` can use it as well


dd-sdk-android-core/build.gradle.kts
```diff
@@ -104,6 +104,8 @@ dependencies {
     implementation(libs.androidXCollection)
     implementation(libs.androidXWorkManager)
 
+    implementation(project(":dd-sdk-android-internal"))
```
mariusc83, line 107: if we have the core depending on it why do we need it also in the other modules depending on core ? If you need the API in the other modules maybe we could use `api` instead of `implementation` here

mariusc83 (approved):

0xnm (approved):

PR #2214, opened 2024-08-27
Add SpanEvent serializer in benchmark module

ambushwork:
### What does this PR do?

Add SpanEvent serializer in benchmark module

### Additional Notes

* `DatadogDataConstraints` constructor is assigned with default value of `internalLogger`, so that in benchmark module, it doesn't have to provide an implementation of `InternalLogger`

### DIFF
--- a/tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
+++ b/tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
@@ -1,0 +1,36 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.benchmark.internal
+
+import com.datadog.benchmark.internal.model.SpanEvent
+import com.google.gson.JsonArray
+import com.google.gson.JsonObject
+
+internal class SpanEventSerializer {
+    // region Serializer
+
+    fun serialize(env: String, spanEvents: List<SpanEvent>): String {
+        val spans = JsonArray(spanEvents.size)
+        spanEvents.forEach {
+            spans.add(it.toJson())
+        }
+        val jsonObject = JsonObject()
+        jsonObject.add(TAG_SPANS, spans)
+        jsonObject.addProperty(TAG_ENV, env)
+
+        return jsonObject.toString()
+    }
+
+    // end region
+
+    companion object {
+
+        // PAYLOAD TAGS
+        internal const val TAG_SPANS = "spans"
+        internal const val TAG_ENV = "env"
+    }
+}


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/constraints/DatadogDataConstraints.kt
```diff
@@ -16,7 +17,9 @@ import java.util.Locale
  *
  * @param internalLogger Internal logger.
  */
-class DatadogDataConstraints(private val internalLogger: InternalLogger) : DataConstraints {
+class DatadogDataConstraints(
+    private val internalLogger: InternalLogger = NoOpInternalLogger()
```
0xnm, line 21: maybe it is better to revert this change and just explicitly provide `NoOpInternalLogger` at the call site in the `benchmark` module?

dd-sdk-android-internal/api/dd-sdk-android-internal.api
0xnm, line 1: empty file in the commit?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/constraints/DatadogDataConstraints.kt
ambushwork: `InernalLogger` and `NoOpInternalLogger` are now inside `core`, `internal` doesn't have dependency on it, we need to migrate them first if we want to do that, what do you prefer?

ambushwork: Yes, sorry, I was confused by myself, ignore what I said above. it's actually the problem of `internal` modifier on `NoOpInternalLogger`

tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Date
+
+internal class SpanEventSerializer(
+    private val dataConstraints: DataConstraints = DatadogDataConstraints()
+) {
+
+    // region Serializer
+
+    fun serialize(env: String, spanEvents: List<SpanEvent>): String {
+        val spans = JsonArray(1)
```
mariusc83, line 25: why `JsonArray(1)` here ?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
```diff
@@ -0,0 +1,87 @@
...
+        jsonObject.addProperty(TAG_ENV, env)
+
+        return jsonObject.toString()
+    }
+
+    // endregion
+
+    // region Internal
+
+    private fun sanitizeKeys(model: SpanEvent): SpanEvent {
```
mariusc83, line 40: let's name this `span` instead of `model`

tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
```diff
@@ -0,0 +1,87 @@
...
+        val newUserObject = sanitizeUserAttributes(model.meta.usr)
+        val newMetricsObject = sanitizeMetrics(model.metrics)
+        return model.copy(meta = model.meta.copy(usr = newUserObject), metrics = newMetricsObject)
+    }
+
+    private fun sanitizeUserAttributes(usr: SpanEvent.Usr): SpanEvent.Usr {
```
mariusc83, line 46: do you really need to sanitize here ? These are internal spans so I am not sure we need this. WDYT ?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
ambushwork: It seems we only send one span in each `spans` attribute, refer to :
https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/SpanEventSerializer.kt

tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
ambushwork: I am not sure, I copy them from `feature-traces` in case that we might need it, I can remove it since it's not useful

tools/benchmark/src/main/java/com/datadog/benchmark/internal/SpanEventSerializer.kt
ambushwork: you are right, now changed to `spanEvents.size`

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/constraints/DatadogDataConstraints.kt
ambushwork: this is removed since Marius said sanitize is not necessary here.

0xnm (approved):

xgouchet (approved):

mariusc83 (approved):

PR #2217, opened 2024-08-28
Add SpanEvent request build and upload process in benchmark module

ambushwork:
### What does this PR do?

* Rename `MetricsContext` to `BenchmarkContext` so that it can be used for span as well
* Create `SpanRequestBodyBuilder` to build span request
* Create `BenchmarkSpanToSpanEventMapper` to map from `BenchmarkSpan` to `SpanEvent`
* Update `DatadogHttpClient` to be able to upload span request
* Add some unit tests.


dd-sdk-android-core/build.gradle.kts
```diff
@@ -109,6 +109,7 @@ dependencies {
     // Generate NoOp implementations
     ksp(project(":tools:noopfactory"))
 
+    implementation(project(":dd-sdk-android-internal"))
```
0xnm, line 112: there is already such dependency declared few lines above

tools/benchmark/src/main/java/com/datadog/benchmark/internal/BenchmarkSpanToSpanEventMapper.kt
```diff
@@ -0,0 +1,60 @@
...
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.sdk.trace.data.SpanData
+
+internal class BenchmarkSpanToSpanEventMapper {
+
+    fun map(
+        context: BenchmarkContext,
+        spanData: SpanData
+    ): SpanEvent {
+        val durationNano = spanData.endEpochNanos - spanData.startEpochNanos
```
0xnm, line 20:
```suggestion
        val durationNanos = spanData.endEpochNanos - spanData.startEpochNanos
```

to be consistent with naming

tools/benchmark/src/main/java/com/datadog/benchmark/internal/DatadogHttpClient.kt
```diff
@@ -22,36 +24,43 @@ import java.util.UUID
...
+        .build(),
+    private val metricRequestBodyBuilder: MetricRequestBodyBuilder = MetricRequestBodyBuilder(context),
+    private val spanRequestBuilder: SpanRequestBodyBuilder = SpanRequestBodyBuilder(context)
+) {
 
-    private val metricRequestBodyBuilder = MetricRequestBodyBuilder(context)
+    internal fun uploadMetric(metrics: List<MetricData>) {
+        postRequest(
+            operationName = OPERATION_NAME_METRICS,
+            url = EndPoint.AP1.metricUrl(),
```
0xnm, line 43: why do we uploading to `AP1` and not to `US1`?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/DatadogHttpClient.kt
```diff
@@ -22,36 +24,43 @@ import java.util.UUID
...
+            metricRequestBodyBuilder.buildJson(metrics)
+        )
+    }
 
-    fun uploadMetric(metrics: List<MetricData>) {
-        val response = postRequest(
+    internal fun uploadSpanEvent(spanEvents: List<SpanEvent>) {
+        postRequest(
+            operationName = OPERATION_NAME_TRACES,
+            url = EndPoint.AP1.tracesUrl(),
```
0xnm, line 52: same as above

tools/benchmark/src/main/java/com/datadog/benchmark/internal/RequestBodyBuilder.kt
```diff
@@ -6,8 +6,7 @@
 
 package com.datadog.benchmark.internal
 
-import com.google.gson.JsonElement
-
 internal interface RequestBodyBuilder<T> {
-    fun buildJsonElement(data: List<T>): JsonElement
+
+    fun buildJson(data: List<T>): String
```
0xnm, line 11: to be precise, not return type is not `JsonXXX`, but simply `String`, so I guess method should be renamed to just `build`, because `String` return type doesn't imply that it should be serialized json there

tools/benchmark/src/main/java/com/datadog/benchmark/internal/DatadogHttpClient.kt
ambushwork: Nice catch, actually it should be neither `AP1` nor `US1`, it should be the end point defined in the configuration.

dd-sdk-android-core/build.gradle.kts
ambushwork: removed

0xnm (approved):

PR #2220, opened 2024-08-29
Add implementations of benchmark profiling api

ambushwork:
### What does this PR do?

* Update `BenchmarkTracer` API
* Add implementations for following interfaces:
   * BenchmarkSpanBuilder
   * BenchmarkProfiler
   * BenchmarkSpan
   * BenchmarkTracer
* Add convenient static function to measure the span.

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpanExt.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpanExt.kt
@@ -1,0 +1,32 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiler
+
+/**
+ * Wraps the provided lambda within a [BenchmarkSpan].
+ * @param T the type returned by the lambda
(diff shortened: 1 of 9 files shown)


dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpanExt.kt
```diff
@@ -0,0 +1,32 @@
...
+ * @param operationName the name of the [BenchmarkSpan] created around the lambda
+ * (default is `true`)
+ * @param block the lambda function traced by this newly created [BenchmarkSpan]
+ *
+ */
+inline fun <T : Any?> withinBenchmarkSpan(
+    operationName: String,
+    block: BenchmarkSpan.() -> T
+): T {
+    val tracer = GlobalBenchmark.get().getTracer("dd.sdk.android")
```
0xnm, line 21: is it `dd.sdk.android` and not `dd-sdk-android` because of the some format requirements?

0xnm (approved):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiler/BenchmarkSpanExt.kt
ambushwork: No, it's just used as a unique identifier for the open-telemetry tracer, since we only want one tracer for benchmarking, it can be any string here.
But in case of confusion, I change it to "dd-sdk-android".

0xnm (approved):

PR #2223, opened 2024-08-30
Integrate benchmark span measurements in session replay

ambushwork:
### What does this PR do?

Benchmark spans measurements are put in two places of session replay:
* Span with name "SnapshotProducer" when `snapshotProducer.produce` is called
* Span with name view class name when mapping the view.

### Demo

image

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt
@@ -9,6 +9,7 @@ package com.datadog.android.sessionreplay.internal.recorder
 import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.UiThread
+import com.datadog.android.internal.profiler.withinBenchmarkSpan
 import com.datadog.android.sessionreplay.ImagePrivacy
 import com.datadog.android.sessionreplay.SessionReplayPrivacy
 import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueRefs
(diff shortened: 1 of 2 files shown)


xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt
```diff
@@ -54,35 +55,37 @@ internal class SnapshotProducer(
...
-        val traversedTreeView = treeViewTraversal.traverse(view, mappingContext, recordedDataQueueRefs)
-        val nextTraversalStrategy = traversedTreeView.nextActionStrategy
-        val resolvedWireframes = traversedTreeView.mappedWireframes
-        if (nextTraversalStrategy == TraversalStrategy.STOP_AND_DROP_NODE) {
-            return null
-        }
-        if (nextTraversalStrategy == TraversalStrategy.STOP_AND_RETURN_NODE) {
-            return Node(wireframes = resolvedWireframes, parents = parents)
-        }
+        return withinBenchmarkSpan(view::class.java.simpleName) {
```
0xnm, line 58: can you please explain the approach? does it mean that effectively we will measure it only in the internal runs and for runs on the customer side it will be no-op and hence no performance penalty?

ambushwork: Yes exactly, by default `GlobalBenchmark` has only ` NoOpBenchmarkProfiler` as the implementation, unless `GlobalBenchmark.register` is called to register a real implementation inside.

Currently the real implementation is in `:tools/benchmark` module, so only `:sample/benchmark` application who depends on it will have the real effect.

0xnm (approved):

PR #2228, opened 2024-09-02
Update benchmark metrics memory reader probe interval

ambushwork:
### What does this PR do?

Update benchmark metrics memory reader probe interval

Logic changes:
* Before: Memory reader reads the stats file directly for the memory data when `readVitalData()` is called every 10 seconds.
* After: Memory reader keeps reading the stats file every 500ms internally and keeps the max memory data until `readVitalData` is called to output this result.

### DIFF
--- a/tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
+++ b/tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
@@ -7,16 +7,44 @@
 package com.datadog.benchmark.internal.reader
 
 import java.io.File
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.TimeUnit
+import kotlin.math.max
 
 /**
  * Reads the device's `VmRSS` based on the `/proc/self/status` file.
  * cf. documentation https://man7.org/linux/man-pages/man5/procfs.5.html
  */
 internal class MemoryVitalReader(
-    internal val statusFile: File = STATUS_FILE
+    internal val statusFile: File = STATUS_FILE,
+    private val readerDelay: Long = STATUS_FILE_READ_INTERVAL_MS
 ) : VitalReader {
 
-    override fun readVitalData(): Double? {
+    @Volatile
+    private var maxMemory: Double = 0.0
+    private val executorService: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+
+    private val readRunnable: Runnable = Runnable {
+        maxMemory = maxMemory.coerceAtLeast(readStatusFile() ?: 0.0)
+        maxMemory = max(readStatusFile() ?: 0.0, maxMemory)
+    }
+
+    override fun readVitalData(): Double {
+        val current = maxMemory
+        maxMemory = 0.0
+        return current
+    }
+
+    override fun start() {
+        executorService.scheduleWithFixedDelay(readRunnable, 0L, readerDelay, TimeUnit.MILLISECONDS)
+    }
+
+    override fun stop() {
+        executorService.shutdownNow()
+    }
+
+    private fun readStatusFile(): Double? {
         if (!(statusFile.exists() && statusFile.canRead())) {
             return null
         }
@@ -39,6 +67,8 @@ internal class MemoryVitalReader(
     companion object {
 
         private const val BYTES_IN_KB = 1000
+
+        private const val STATUS_FILE_READ_INTERVAL_MS = 500L
 
         private const val STATUS_PATH = "/proc/self/status"
         internal val STATUS_FILE = File(STATUS_PATH)
--- a/tools/benchmark/src/test/java/internal/MemoryVitalReaderTest.kt
+++ b/tools/benchmark/src/test/java/internal/MemoryVitalReaderTest.kt
@@ -63,7 +63,8 @@ internal class MemoryVitalReaderTest {
     fun `set up`() {
         fakeFile = File(tempDir, "stat")
         fakeStatusContent = generateStatusContent(fakeVmRss)
-        testedReader = MemoryVitalReader(fakeFile)
+        testedReader = MemoryVitalReader(fakeFile, 50)
+        testedReader.start()
     }
 
     @Test
@@ -81,6 +82,7 @@ internal class MemoryVitalReaderTest {
         fakeFile.writeText(fakeStatusContent)
 
         // When
+        Thread.sleep(100L)
         val result = testedReader.readVitalData()
 
         // Then
@@ -88,20 +90,22 @@ internal class MemoryVitalReaderTest {
     }
 
     @Test
-    fun `M read correct data W readVitalData() {multiple times}`(
+    fun `M read maximum data W readVitalData() during memory file updates`(
         @IntForgery(1) vmRssValuesKb: List<Int>
     ) {
-        // Given
-        val results = mutableListOf<Double>()
+        // Given: Set a memory update interval longer than file reader interval to make sure every update data is read.
+        val memoryUpdateIntervalMs = 100L
+
         // When
         vmRssValuesKb.forEach { vmRss ->
+            // Set a Longer
+            Thread.sleep(memoryUpdateIntervalMs)
             fakeFile.writeText(generateStatusContent(vmRss))
-            val result = testedReader.readVitalData()
-            results.add(result!!)
         }
+        val result = testedReader.readVitalData()
 
         // Then
-        assertThat(results).isEqualTo(vmRssValuesKb.map { vmRssKb -> vmRssKb.toDouble() * 1000 })
+        assertThat(result).isEqualTo(vmRssValuesKb.max().toDouble() * 1000)
     }
 
     @Test
@@ -110,7 +114,7 @@ internal class MemoryVitalReaderTest {
         val result = testedReader.readVitalData()
 
         // Then
-        assertThat(result).isNull()
+        assertThat(result).isZero()
     }
 
     @Test
@@ -124,7 +128,7 @@ internal class MemoryVitalReaderTest {
         val result = testedReader.readVitalData()
 
         // Then
-        assertThat(result).isNull()
+        assertThat(result).isZero()
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
```diff
@@ -7,16 +7,41 @@
...
+    private val executorService: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+
+    private val readRunnable: Runnable = Runnable {
+        maxMemory = (readStatusFile() ?: 0.0).coerceAtLeast(maxMemory)
+    }
+
+    override fun readVitalData(): Double {
+        return maxMemory.apply {
+            maxMemory = 0.0
+        }
```
0xnm, line 33: this is kind of confusing syntax: `apply` has `this` as implicit receiver and the name means that we are doing something with receiver and returning it. But here nothing is done with receiver, instead modification is done to the property of `MemoryVitalReader`.

I think it will be better to write it without any sugar:

```
val current = maxMemory
maxMemory = 0.0
return current
```

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
```diff
@@ -7,16 +7,41 @@
...
  * cf. documentation https://man7.org/linux/man-pages/man5/procfs.5.html
  */
 internal class MemoryVitalReader(
-    internal val statusFile: File = STATUS_FILE
+    internal val statusFile: File = STATUS_FILE,
+    private val readerDelay: Long = STATUS_FILE_READ_INTERVAL_MS
 ) : VitalReader {
 
-    override fun readVitalData(): Double? {
+    private var maxMemory: Double = 0.0
```
0xnm, line 23: you need to make this property thread-safe: write and read is performed from different threads

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
```diff
@@ -7,16 +7,41 @@
...
+    private val executorService: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+
+    private val readRunnable: Runnable = Runnable {
+        maxMemory = (readStatusFile() ?: 0.0).coerceAtLeast(maxMemory)
+    }
+
+    override fun readVitalData(): Double {
```
0xnm, line 30: can you please explain this change? shouldn't it be external caller which is making the call to `readVitalData` control the frequency of the reads?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
```diff
@@ -7,16 +7,41 @@
...
+    internal val statusFile: File = STATUS_FILE,
+    private val readerDelay: Long = STATUS_FILE_READ_INTERVAL_MS
 ) : VitalReader {
 
-    override fun readVitalData(): Double? {
+    private var maxMemory: Double = 0.0
+    private val executorService: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+
+    private val readRunnable: Runnable = Runnable {
+        maxMemory = (readStatusFile() ?: 0.0).coerceAtLeast(maxMemory)
```
mariusc83, line 27: it is a bit weird the way we use `coerceAtLeast` here as there should be `minMemory` instead of `max`

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
```diff
@@ -7,16 +7,41 @@
...
+    }
+
+    override fun readVitalData(): Double {
+        return maxMemory.apply {
```
mariusc83, line 31: this value is accessible/modifiable from different threads so you will need an AtomicDouble here

mariusc83 (changes requested):

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
ambushwork: Actually the external caller does have a settled frequency (10s by default), but to keep this external caller simple and agnostic about all types of reader, it just collects data from all the readers and export the data immediately every 10s. So every reader has to prepare their own data to be read if they have its own logic (like here memory needs to read more frequently and only expose the maximum data during every read, but CPU reader doesn't have to).

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
ambushwork: Agree, I will try to avoid confusing syntax sugar.

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
ambushwork: actually this is the function proposed by Kotlin compiler to convert if I type `Math.max`

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
ambushwork: Good point, since `AtomicDouble` doesn't exist, I use `AtomicLong` to store the data

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
ambushwork: Good point, I use AtomicLong to store the `maxMemory` data here

0xnm: probably simple `@Volatile` annotation will be enough, because you only get and set value

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
mariusc83: yes but I would expect to be `coerceAtLeast(minMemory)`

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
0xnm: > external caller does have a settled frequency (10s by default)

This was an approach designed for the SDK, where all the readers were queried at the same frequency. Now, if we are designing the solution for our internal needs and can allow ourselves more frequent querying, maybe it is worth to think of just of independent frequency for all the readers instead?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
ambushwork: `coerceAtLeast` ensures the this value is not less than the current `maxMemory`, if I reverse the caller and param, will it looks better to you?
```
 maxMemory = maxMemory.coerceAtLeast(readStatusFile()?:0.0)
```

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
ambushwork: You mean that instead of having an internal frequency of every reader, it should be external caller who decides the different frequency of each reader?
I think with the solution of internal frequency of every reader, the external caller can be agnostic, and easier to test each reader independently.

0xnm: I mean right now with such solution you have 2 frequencies - of external caller and internal of the reader, which requires you to have different threads and introduce thread-safety guards. It is not quite clear why should we have 2 different frequencies while we can have only one - external, which still can be defined individually for each reader while having a single thread.

ambushwork: Here the benefits that I see to separate internal frequency and external frequency is that:
*  the external caller frequency is the one that can be configured in the benchmark app, it can control both query interval of all readers, and it is also the frequency of upload. It depends on how frequently we want to sample the data.
* Having an internal frequency which can't be changed from outside can make sure that we have always the accurate vital data from each reader. external caller or app doesn't have to care about the frequency of detailed vital and doesn't have to store a cache like `maxMemory` during the time for example.

What do you think about it?

tools/benchmark/src/main/java/com/datadog/benchmark/internal/reader/MemoryVitalReader.kt
xgouchet: I agree with @mariusc83 , the doc of `coerceAtLeast` states 
> Ensures that `this` value is not less than the specified minimumValue.
> Returns: `this` value if it's greater than or equal to the minimumValue or the minimumValue otherwise.
So by definition you should call it with the `minMemory`

ambushwork: Yes, I understand, but here the min value should be the current `maxMemory`, it ensures the next `maxMerory` is not less than the current one. I found it's a bit weird to rename it to `minMemory` because it doesn't stand for what it actually means. Do you have any suggestion on that, or maybe I just use the `Math.max` to avoid the confusion?
```
maxMemory = Math.max(readStatusFile()?:0.0, maxMemory)
```

xgouchet: Understood, indeed it makes more sense to use Max rather than coerceAtLeast. 

Why don't you use the static `kotlin.math.max` fun then instead of the Java `Math.max`?

ambushwork: you are right, static kotlin max is more understandable and it doesn't throw compiler warning, updated!

mariusc83 (approved):
Tks for the updates !!

PR #2237, opened 2024-09-04
Merge 2.13.0 on develop

xgouchet:


buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
```diff
@@ -20,7 +20,7 @@ object AndroidConfig {
     const val MIN_SDK_FOR_WEAR = 23
     const val BUILD_TOOLS_VERSION = "34.0.0"
 
-    val VERSION = Version(2, 13, 0, Version.Type.Snapshot)
+    val VERSION = Version(2, 13, 0, Version.Type.Release)
```
0xnm, line 23: this change should be reverted before the merge (and version should be bumped in the `develop` branch)

PR #2243, opened 2024-09-05
RUM-5843 Introduce the RumMonitor#addViewLoadingTime API

mariusc83:
### What does this PR do?

### Motivation

### Additional Notes


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -323,6 +323,15 @@ interface RumMonitor {
      */
     fun stopSession()
 
+    /**
+     * Adds view loading time to current RUM view based on the time elapsed since the view was started.
+     * This method should be called only once per view.
```
0xnm, line 328: Why do we enforce it requirement on the client side? Cannot we just check in the implementation that if it was called already for the view, then ignore subsequent calls?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -6960,6 +6960,165 @@ internal class RumViewScopeTest {
...
+        viewLoadingTimeEvents.forEach {
+            testedScope.handleEvent(
+                it,
+                mockWriter
+            )
+        }
+
+        // Then
+        argumentCaptor<ViewEvent> {
+            verify(mockWriter, times(viewLoadingTimeEvents.size))
```
0xnm, line 7050: why should we allow multiple calls?

sample/benchmark/src/main/java/com/datadog/benchmark/sample/MainActivity.kt
```diff
@@ -38,6 +39,7 @@ class MainActivity : AppCompatActivity() {
         datadogBenchmark.stop()
     }
 
+    @OptIn(ExperimentalRumApi::class)
```
0xnm, line 42: It seems it is not needed here? I don't see any `addViewLoadingTime` call here

sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/NavigationSample.kt
```diff
@@ -129,3 +149,12 @@ internal sealed class Screen(
 internal fun <T> oneOf(vararg items: T): T {
     return items[Random.nextInt(items.size)]
 }
+
+private class DatadogLifecycleObserver : LifecycleEventObserver {
+    @OptIn(ExperimentalRumApi::class)
+    override fun onStateChanged(source: LifecycleOwner, event: Lifecycle.Event) {
+        if (event == Lifecycle.Event.ON_RESUME) {
+            GlobalRumMonitor.get().addViewLoadingTime()
```
0xnm, line 157: won't we have multiple calls for the same view here when application goes info background and then back into foreground?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/gdpr/GdprDialogFragment.kt
```diff
@@ -47,6 +49,12 @@ internal class GdprDialogFragment : DialogFragment() {
         return rootView
     }
 
+    @OptIn(ExperimentalRumApi::class)
+    override fun onResume() {
+        super.onResume()
+        GlobalRumMonitor.get().addViewLoadingTime()
```
0xnm, line 55: same: I guess we can have multiple calls for the same view?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/gdpr/GdprDialogFragment.kt
```diff
@@ -55,4 +63,10 @@ internal class GdprDialogFragment : DialogFragment() {
             TrackingConsent.NOT_GRANTED -> R.id.not_granted
         }
     }
+
+    @OptIn(ExperimentalRumApi::class)
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        GlobalRumMonitor.get().addViewLoadingTime()
```
0xnm, line 70: `onResume` will be called right after `onViewCreated` effectively discarding this call

sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/ImageComponentsFragment.kt
```diff
@@ -61,42 +67,55 @@ internal class ImageComponentsFragment : Fragment() {
...
 
+    @OptIn(ExperimentalRumApi::class)
     private fun loadImageView() {
         viewModel.fetchRemoteImage(
             LARGE_IMAGE_URL,
             imageViewRemote,
             object : ImageLoadedCallback {
                 override fun onImageLoaded(resource: Drawable) {
                     imageViewRemote.setImageDrawable(resource)
+                    if (imageLoadedCounter.decrementAndGet() == 0) {
```
0xnm, line 78: should we extract this into a common private method? same snippet is used 4 times.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/WebViewRecordFragment.kt
```diff
@@ -29,11 +30,15 @@ internal class WebViewRecordFragment : Fragment() {
         "datadoghq.dev"
     )
 
+    @Volatile
+    private var pageWasLoaded: Boolean = false
```
0xnm, line 34:
```suggestion
    private var pageIsLoaded: Boolean = false
```

sample/kotlin/src/main/kotlin/com/datadog/android/sample/viewpager/PagerChildFragment.kt
```diff
@@ -31,6 +33,10 @@ internal open class PagerChildFragment : Fragment() {
     )
     private lateinit var webView: WebView
 
+    @Volatile
+    private var pageWasLoaded: Boolean = false
```
0xnm, line 37:
```suggestion
    private var pageIsLoaded: Boolean = false
```

sample/kotlin/src/main/kotlin/com/datadog/android/sample/webview/WebFragment.kt
```diff
@@ -27,8 +29,12 @@ internal class WebFragment : Fragment() {
         "datadoghq.dev"
     )
 
+    @Volatile
+    private var pageWasLoaded: Boolean = false
```
0xnm, line 33:
```suggestion
    private var pageIsLoaded: Boolean = false
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
mariusc83: we follow same approach as on iOS and for now on their end they allow it to be called multiple times. @ganeshnj maybe you want to add some thoughts here ?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
mariusc83: same answer as above.

sample/benchmark/src/main/java/com/datadog/benchmark/sample/MainActivity.kt
mariusc83: yup a left - over from previous tries :(

sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/NavigationSample.kt
mariusc83: when the app goes into background there will be a new view...we stop the view on the `onPause` the way I remember.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/gdpr/GdprDialogFragment.kt
mariusc83: same as above

sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/WebViewRecordFragment.kt
mariusc83: I know that the WebView callbacks are being executed on a different thread this is the reason why I want to keep it as @Volatile

sample/kotlin/src/main/kotlin/com/datadog/android/sample/webview/WebFragment.kt
mariusc83: same as above

sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/NavigationSample.kt
0xnm: no, here view tracking relies on navigation - I don't think navigation events are fired when app changes the state (but I may be wrong). While `addViewLoadingTime` call here relies not on the navigation, but on the lifecycle events.

mariusc83: let me double check this, you might be right, I was assuming we rely on the view lifecycle here

sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/WebViewRecordFragment.kt
0xnm: my suggestion is not about volatile, but about naming

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -323,6 +323,15 @@ interface RumMonitor {
...
+     * If the view is not started, this method does nothing.
+     * If the view is not active(stopped), this method does nothing.
```
xgouchet, line 330: > [!NOTE]
> Since the loading time is applied to the `current RUM view` (maybe should be the `current active RUM View`), those two lines are misleading IMO. It feels like we make the assumption that a view is necessarily an Activity/Fragment with a lifecycle (start, resume, …), which is not always the case.

```suggestion
     * If no view is started or active, this method does nothing.
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
xgouchet: > [!IMPORTANT]
> I agree with @0xnm , we can't ask the customer to know whether or not the method has already been called. They could have several code paths that would lead to calling this method, and it would be a hassle to ask them to track whether or not that method have been called for a given View. 
> We need to make our SDK resilient enough to handle "misuse".

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
xgouchet: > [!NOTE]
> Not sure what the RFC stated, but imo the first loading time is the one that should be kept, otherwise, if the same screen is updated/refreshed, we would report a longer loading time than what the end user experienced.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/NavigationSample.kt
```diff
@@ -129,3 +149,15 @@ internal sealed class Screen(
 internal fun <T> oneOf(vararg items: T): T {
     return items[Random.nextInt(items.size)]
 }
+
+private class DatadogLifecycleObserver : LifecycleEventObserver {
+    private var viewIsLoaded = false
+
+    @OptIn(ExperimentalRumApi::class)
+    override fun onStateChanged(source: LifecycleOwner, event: Lifecycle.Event) {
+        if (event == Lifecycle.Event.ON_RESUME && !viewIsLoaded) {
```
xgouchet, line 158: > [!NOTE]
> Is this sample the kind of use case we want to showcase ? Tracking the `onResume` time is not relevant most of the time. We should probably instead only do what you added in the `DataListViewModel` where we actually load something from the network, which would be a better example of how to use this new API.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
mariusc83: I am following the same docs as in iOS

sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/NavigationSample.kt
mariusc83: Regarding your last comment, that is a nice to have in the Sample app, my intention was to add the viewLoadingTime for most of the views there. In that case I will remove the call for this view and maybe add a task to enrich the sample app. This is not the scope for this PR. @0xnm what's your opinion on that ?

0xnm: It depends on the nature of the screen, some screens may not load anything from the network/disk. But maybe to highlight that this method shouldn't be blindly used in the `onResume` we may want indeed to have it called only in the data stream.

mariusc83: On the same time we could agree that there will be use cases on the customer's end where simple static views like these ones would have to be measured. In that case we will remove these use - cases from the sample app and only maintain the ones that imply an async data ?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/WebViewRecordFragment.kt
```diff
@@ -47,12 +49,18 @@ internal class WebViewRecordFragment : Fragment() {
...
+                super.onPageFinished(view, url)
+                GlobalRumMonitor.get().addViewLoadingTime(overwrite = false)
+            }
+        }
         webView.settings.javaScriptEnabled = true
         WebViewTracking.enable(webView, webViewTrackingHosts)
         startCustomRumViewButton.setOnClickListener {
             GlobalRumMonitor.get().startView(this, "Custom RUM View")
         }
+        // add webview on load listener
```
xgouchet, line 63: > [!NOTE]
> Is this comment needed here ? Feels like its a TODO that didn't get removed

xgouchet (approved):

PR #2244, opened 2024-09-05
RUM-5985: Add isContainer attribute to session replay span

ambushwork:
### What does this PR do?

Add a "isContainer" attribute into session replay span request to differentiate the components and containers view.

### Motivation
In the benchmark dashboard, we want to have a top list graph to see which components take the leads of the time consumption in SR. here is the current graph:
image

As we can see the top list is occupied by the containers, which doesn't what we expect. Because containers' spans contain the children view, so we want to have only real components here.

### Demo
After adding this attributes, we can differentiate them by adding filter `@attribute.container:false `

Add we got pure components span in the list
image

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/BenchmarkExt.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/BenchmarkExt.kt
@@ -1,0 +1,27 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import com.datadog.android.internal.profiler.BenchmarkSpan
+import com.datadog.android.internal.profiler.withinBenchmarkSpan
+
+private const val ATTRIBUTE_CONTAINER = "attribute.container"
+
+/**
+ * A wrap function of [withinSRBenchmarkSpan] dedicated to session replay span recording.
+ */
+internal inline fun <T : Any?> withinSRBenchmarkSpan(
+    spanName: String,
+    isContainer: Boolean = false,
+    block: BenchmarkSpan.() -> T
+): T {
+    return withinBenchmarkSpan(
+        spanName,
+        mapOf(ATTRIBUTE_CONTAINER to isContainer.toString()),
+        block
+    )
+}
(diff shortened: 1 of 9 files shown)


tools/benchmark/src/main/java/com/datadog/benchmark/internal/BenchmarkSpanToSpanEventMapper.kt
```diff
@@ -28,19 +28,25 @@ internal class BenchmarkSpanToSpanEventMapper {
...
             metrics = resolveMetrics()
         )
     }
 
-    private fun resolveMeta(): SpanEvent.Meta {
-        // TODO: RUM-5985 Fill SpanEvent with useful meta data
+    private fun resolveMeta(spanData: SpanData): SpanEvent.Meta {
+        val map = mutableMapOf<String, String>()
+
+        spanData.attributes.forEach { t, u ->
```
0xnm, line 39: what `t` and `u` stand for?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/BenchmarkExt.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import com.datadog.android.internal.profiler.BenchmarkSpan
+import com.datadog.android.internal.profiler.withinBenchmarkSpan
+
+private const val ATTRIBUTE_CONTAINER = "attribute.container"
+
+/**
+ * A wrap function of [withinBenchmarkSpan] dedicated to session replay span recording.
+ */
+internal inline fun <T : Any?> measureSpan(
```
0xnm, line 17: shouldn't function name be aligned with the name of the function it wraps? it essentially just adds one single attribute

ambushwork: updated

tools/benchmark/src/main/java/com/datadog/benchmark/internal/BenchmarkSpanToSpanEventMapper.kt
ambushwork: it's `AttributeKey` and `Any`(value), updated with meaningful names

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/BenchmarkExt.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import com.datadog.android.internal.profiler.BenchmarkSpan
+import com.datadog.android.internal.profiler.withinBenchmarkSpan
+
+private const val ATTRIBUTE_CONTAINER = "attribute.container"
+
+/**
+ * A wrap function of [withinBenchmarkSpan] dedicated to session replay span recording.
+ */
+internal inline fun <T : Any?> withinBenchmarkSpan(
```
xgouchet, line 17: > [!NOTE]
> As it's a version dedicated for SR, and to easily disambiguate between this helper and the base method, maybe we could name it `withinSRBenchmarkSpan` instead, WDYT?

0xnm (approved):

xgouchet (approved):

PR #2247, opened 2024-09-06
RUM-6093:Add TimeBank in session replay recorder for dynamic optimisation

ambushwork:
### What does this PR do?

Add TimeBank in session replay recorder for dynamic optimisation.

The main idea is to have a time balance in this `TimeBank`, every execution of session replay recording will consume this balance, and the balance will be recharged during the time passing until a maximum balance is reached.

Before every execution of recording, if the balance is negative, the execution will skipped until the balance is recharged to positive.

### Motivation

RUM-6093

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
@@ -8,11 +8,14 @@ package com.datadog.android.sessionreplay.internal.recorder
 
 import android.os.Handler
 import android.os.Looper
+import com.datadog.android.api.InternalLogger
 import java.util.concurrent.TimeUnit
 
 internal class Debouncer(
     private val handler: Handler = Handler(Looper.getMainLooper()),
-    private val maxRecordDelayInNs: Long = MAX_DELAY_THRESHOLD_NS
+    private val maxRecordDelayInNs: Long = MAX_DELAY_THRESHOLD_NS,
+    private val timeBank: TimeBank = RecordingTimeBank(),
+    private val internalLogger: InternalLogger
 ) {
 
     private var lastTimeRecordWasPerformed = 0L
@@ -37,8 +40,28 @@ internal class Debouncer(
     }
 
     private fun executeRunnable(runnable: Runnable) {
-        runnable.run()
-        lastTimeRecordWasPerformed = System.nanoTime()
+        runInTimeBalance {
+            runnable.run()
+            lastTimeRecordWasPerformed = System.nanoTime()
+        }
+    }
+
+    private fun runInTimeBalance(block: () -> Unit) {
+        if (timeBank.updateAndCheck(System.nanoTime())) {
+            val startTimeInNano = System.nanoTime()
+            block()
+            val endTimeInNano = System.nanoTime()
+            timeBank.consume(endTimeInNano - startTimeInNano)
+        } else {
+            // TODO RUM-5188 report with telemetry about the frames skipped
+            internalLogger.log(
+                InternalLogger.Level.INFO,
+                InternalLogger.Target.MAINTAINER,
+                messageBuilder = {
+                    MSG_FRAME_SKIP
+                }
+            )
+        }
     }
 
     companion object {
@@ -47,5 +70,7 @@ internal class Debouncer(
 
         // one frame time
         internal const val DEBOUNCE_TIME_IN_MS: Long = 64
+
+        private const val MSG_FRAME_SKIP = "Session Replay skipped this recording due to the time limit"
     }
 }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
@@ -1,0 +1,53 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import java.util.concurrent.TimeUnit
+import kotlin.math.min
+
+/**
+ * Time Bank is a concept representing an allocated execution quota per second. For example, if the quota is set to
+ * 100 milliseconds per second, it means that within any given second, no more than 100 milliseconds can be used for
+ * executing operations. If the full quota of 100 milliseconds has already been used within a second, further execution
+ * is not permitted until the next second begins and the quota is recharged. Conversely, if less than 100 milliseconds
+ * has been used and the second has not yet elapsed, execution may continue until the quota is reached.
+ */
+internal class RecordingTimeBank(
+    private val maxTimeBalancePerSecondInMs: Long = DEFAULT_MAX_TIME_BALANCE_PER_SEC_IN_MS
+) : TimeBank {
+
+    // The normalized factor of balance increasing by time. If increasing 100ms balance in the bank takes 1000ms,
+    // then the factor will be 100ms/1000ms = 0.1f
+    private val balanceFactor = maxTimeBalancePerSecondInMs.toDouble() / TimeUnit.SECONDS.toMillis(1)
+
+    @Volatile
+    private var recordingTimeBalanceInNano = TimeUnit.MILLISECONDS.toNanos(maxTimeBalancePerSecondInMs)
+
+    @Volatile
+    private var lastCheckTime: Long = 0
+
+    override fun consume(executionTime: Long) {
+        recordingTimeBalanceInNano -= executionTime
+    }
+
+    override fun updateAndCheck(timestamp: Long): Boolean {
+        increaseTimeBank(timestamp)
+        lastCheckTime = timestamp
+        return recordingTimeBalanceInNano >= 0
+    }
+
+    private fun increaseTimeBank(timestamp: Long) {
+        val timePassedSinceLastExecution = timestamp - lastCheckTime
+        recordingTimeBalanceInNano += (timePassedSinceLastExecution * balanceFactor).toLong()
+        recordingTimeBalanceInNano =
+            min(TimeUnit.MILLISECONDS.toNanos(maxTimeBalancePerSecondInMs), recordingTimeBalanceInNano)
+    }
+
+    companion object {
+        private const val DEFAULT_MAX_TIME_BALANCE_PER_SEC_IN_MS = 100L
+    }
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TimeBank.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TimeBank.kt
@@ -1,0 +1,26 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+@NoOpImplementation
+internal interface TimeBank {
+
+    /**
+     * Called to consume execution time from the bank.
+     */
+    fun consume(executionTime: Long)
+
+    /**
+     * Called to update time bank balance and check if the given timestamp
+     * is allowed according to the current time balance.
+     *
+     * @return true if the given timestamp is allowed by time bank to execute a task, false otherwise.
+     */
+    fun updateAndCheck(timestamp: Long): Boolean
+}
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
@@ -1,0 +1,135 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.recorder
+
+import com.datadog.android.sessionreplay.forge.ForgeConfigurator
+import com.datadog.android.sessionreplay.internal.recorder.RecordingTimeBank
+import com.datadog.android.sessionreplay.internal.recorder.TimeBank
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.junit5.ForgeConfiguration
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+import java.util.concurrent.TimeUnit
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(ForgeConfigurator::class)
+class RecordingTimeBankTest {
+
+    private lateinit var recordingTimeBank: TimeBank
+
+    @BeforeEach
+    fun `set up`() {
+        recordingTimeBank = RecordingTimeBank(TEST_MAX_BALANCE_IN_MS)
+    }
+
+    @Test
+    fun `M allow the first execution W check`(forge: Forge) {
+        // Given
+        val timestamp = forge.aLong(min = 0)
+
+        // When
+        val actual = recordingTimeBank.updateAndCheck(timestamp)
+
+        // Then
+        assertThat(actual).isTrue()
+    }
+
+    @Test
+    fun `M skip the next execution W previous consume out the balance`(forge: Forge) {
+        // Given
+        val firstTimestamp = forge.aLong(min = 0)
+        val firstExecutionTime = forge.aLong(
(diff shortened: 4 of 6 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TimeBank.kt
```diff
@@ -0,0 +1,17 @@
...
+package com.datadog.android.sessionreplay.internal.recorder
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+@NoOpImplementation
+internal interface TimeBank {
+
+    fun consume(executionTime: Long)
+
+    fun updateAndCheck(timestamp: Long): Boolean
```
0xnm, line 16: Maybe some description of that is needed. It is not clear what this class is supposed to do and what methods are doing by looking on the names.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
```diff
@@ -41,11 +46,31 @@ internal class Debouncer(
...
+        if (timeBank.updateAndCheck(System.nanoTime())) {
+            val startTimeInNano = System.nanoTime()
+            block()
+            val endTimeInNano = System.nanoTime()
+            timeBank.consume(endTimeInNano - startTimeInNano)
+        } else {
+            // TODO: RUM-5188 report with telemetry about the frames skipped
+            internalLogger.log(
+                InternalLogger.Level.INFO,
+                InternalLogger.Target.USER,
```
0xnm, line 59: Is it useful for the user? What are they supposed to do if they see such log entry? Shouldn't it be `MAINTAINER` here?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Everytime the session replay recording task tries to execute, [updateAndCheck] to increase the time balance
+ * according to the time passed since last call, and check if the balance is positive to execute the task.
+ */
+internal class RecordingTimeBank(
+    private val maxTimeBalancePerSecondInMs: Long = DEFAULT_MAX_TIME_BALANCE_PER_SEC_IN_MS
+) : TimeBank {
+
+    @Volatile
+    private var recordingTimeBalanceInNano = TimeUnit.MILLISECONDS.toNanos(maxTimeBalancePerSecondInMs)
+    private var lastStartTime: Long = 0
```
0xnm, line 23: Shouldn't this mutable property be `@Volatile` as well?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import java.util.concurrent.TimeUnit
+import kotlin.math.min
+
+/**
+ * Implementation of [TimeBank], it maintains a time balance which can be consumed for session replay recording.
+ * Everytime the session replay recording task tries to execute, [updateAndCheck] to increase the time balance
```
0xnm, line 14:
```suggestion
 * Every time the session replay recording task tries to execute, [updateAndCheck] is called to increase the time balance
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
```diff
@@ -0,0 +1,49 @@
...
+ */
+internal class RecordingTimeBank(
+    private val maxTimeBalancePerSecondInMs: Long = DEFAULT_MAX_TIME_BALANCE_PER_SEC_IN_MS
+) : TimeBank {
+
+    @Volatile
+    private var recordingTimeBalanceInNano = TimeUnit.MILLISECONDS.toNanos(maxTimeBalancePerSecondInMs)
+    private var lastStartTime: Long = 0
+
+    // The time to increase 1 nano second balance in the bank. For example the max balance per second is 100ms,
```
0xnm, line 25:
```suggestion
    // The time to increase 1 nanosecond balance in the bank. For example the max balance per second is 100ms,
```

Also I'm still not clear with the description overall. Why do we need this property?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
```diff
@@ -29,10 +32,12 @@ internal class Debouncer(
...
         val timePassedSinceLastExecution = System.nanoTime() - lastTimeRecordWasPerformed
-        if (timePassedSinceLastExecution >= maxRecordDelayInNs) {
-            executeRunnable(runnable)
-        } else {
-            handler.postDelayed({ executeRunnable(runnable) }, DEBOUNCE_TIME_IN_MS)
+        runInTimeBalance {
+            if (timePassedSinceLastExecution >= maxRecordDelayInNs) {
+                executeRunnable(runnable)
+            } else {
+                handler.postDelayed({ executeRunnable(runnable) }, DEBOUNCE_TIME_IN_MS)
```
xgouchet, line 39: > [!WARNING]
> Shouldn't you remove previously posted runnables, if there are multiple debounced executions ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
ambushwork: Here is a temporary handling of the skipped frames, and final solution will be report them with RUM session ended metrics. But you are right, in this case we'd better log to maintainer for now.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
ambushwork: This is the factor of balance, bigger it is, faster the balance will increase by time.
I removed it as the class property, and changed it to a local variable, and also update the comment to make it more clear, please check if it is more understandable to you?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TimeBank.kt
ambushwork: updated, please check it.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
```diff
@@ -29,10 +32,12 @@ internal class Debouncer(
         }
         handler.removeCallbacksAndMessages(null)
         val timePassedSinceLastExecution = System.nanoTime() - lastTimeRecordWasPerformed
-        if (timePassedSinceLastExecution >= maxRecordDelayInNs) {
-            executeRunnable(runnable)
-        } else {
-            handler.postDelayed({ executeRunnable(runnable) }, DEBOUNCE_TIME_IN_MS)
+        runInTimeBalance {
```
mariusc83, line 35: is this needed to handle the problem in Android 23 and below where the SR debouncer doesn't work ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
mariusc83: this is done above at line 33

ambushwork: You are right, so I changed the place where `runInTimeBalance` should wrap up. 
Now `runInTimeBalance` wrap only the function of `executeRunnable`, so we check first the condition of debounce frequency, then check if it meets the time balance.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
```diff
@@ -0,0 +1,50 @@
...
+    @Volatile
+    private var lastStartTime: Long = 0
+
+    override fun consume(executionTime: Long) {
+        recordingTimeBalanceInNano -= executionTime
+    }
+
+    override fun updateAndCheck(timestamp: Long): Boolean {
+        increaseTimeBank(timestamp)
+        lastStartTime = timestamp
```
mariusc83, line 33: shouldn't `lastStartTime` be updated in the `consume()`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/Debouncer.kt
ambushwork: The task is not aimed to handle the infinite calling of recorder problem under Android 23, it's an overall optimisation of all versions, but I believe it can somehow solve it by the way, because if under Android 23, the snapshot is recorded too frequently, it will very soon use out the time balance in the bank, and following recording will be refused due to this optimisation.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
```diff
@@ -0,0 +1,50 @@
...
+        return recordingTimeBalanceInNano >= 0
+    }
+
+    private fun increaseTimeBank(timestamp: Long) {
+        val timePassedSinceLastExecution = timestamp - lastStartTime
+        // The factor of balance increasing by time. if increasing 100ms balance in the bank takes 1000ms, then the
+        // factor will be 100ms/1000ms = 0.1f
+        val balanceFactor = maxTimeBalancePerSecondInMs.toFloat() / TimeUnit.SECONDS.toMillis(1)
```
mariusc83, line 41: I am not sure which one is the variable here ? it seems that this has always the same value as `maxTimeBalancePerSecondInMs` will never change ? Am I wrong ?

ambushwork: Indeed `balanceFactor` will never change, I add this variable as an intermediate argument to avoid a long formula in the code, trying to make it understandable, what do you think?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
ambushwork: Well for the reason why we need this variable, is that I try to avoid to have a super long mathematic formula in the code, so I add this intermediate variable with name to make it more understandable.


This `balanceFactor` is calculated from the parameter in the constructor `maxTimeBalancePerSecondInMs`, for example if `maxTimeBalancePerSecondInMs`=100ms, it means we should only allow 100ms per second to execute the task that we want. So in this time bank, it also means that we can have maximum 100ms in the balance, and to reach the max (100ms) from 0 we need 1000ms, which means the increasing factor is 0.1.


With this factor, we can easily calculate the balance increasing by the given time interval, for example last time when the balance update is at `t0`, current time is `t1`, then the balance we should increase should be `(t1 - t0)*factor`.

Don't hesitate to tell me if it is still confusing,  the goal is to make the code and the document understandable for every one.

0xnm: My question is not the formula, my question is why we need this. Why do we need the factor and why we simply not matching bank with execution time by using 1:1 ratio.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
mariusc83: in that case shouldn't we call it `lastCheckTime` ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
mariusc83: In that case isn't more readable if you declare only once in your class and keep it as a `val` ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
ambushwork: Document updated

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/RecordingTimeBank.kt
ambushwork: updated

mariusc83 (approved):

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
```diff
@@ -0,0 +1,133 @@
...
+    @Test
+    fun `M allow the first execution W check`(forge: Forge) {
+        // Given
+        val timestamp = forge.aLong(min = 0)
+
+        // When
+        val actual = recordingTimeBank.updateAndCheck(timestamp)
+
+        // Then
+        assertThat(actual).isEqualTo(true)
```
0xnm, line 49:
```suggestion
        assertThat(actual).isTrue
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
```diff
@@ -0,0 +1,133 @@
...
+            max = firstTimestamp + firstExecutionTime + interval
+        )
+
+        // When
+        recordingTimeBank.updateAndCheck(firstTimestamp)
+        recordingTimeBank.consume(firstExecutionTime)
+        val actual = recordingTimeBank.updateAndCheck(secondTimestamp)
+
+        // Then
+        assertThat(actual).isEqualTo(false)
```
0xnm, line 72:
```suggestion
        assertThat(actual).isFalse
```

similar change for the lines below

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
```diff
@@ -0,0 +1,133 @@
...
+            )
+        // When
+        recordingTimeBank.updateAndCheck(firstTimestamp)
+        recordingTimeBank.consume(firstExecutionTime)
+        recordingTimeBank.updateAndCheck(secondTimestamp)
+        val actual = recordingTimeBank.updateAndCheck(thirdTimestamp)
+
+        // Then
+        assertThat(actual).isEqualTo(true)
+        //
```
0xnm, line 102:
```suggestion
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
```diff
@@ -0,0 +1,133 @@
...
+
+        val thirdTimestamp =
+            forge.aLong(
+                min = secondTimestamp + firstExecutionTime /
+                    ((TimeUnit.SECONDS.toMillis(1) / TEST_MAX_BALANCE_IN_MS))
+            )
+        // When
+        recordingTimeBank.updateAndCheck(firstTimestamp)
+        recordingTimeBank.consume(firstExecutionTime)
+        recordingTimeBank.updateAndCheck(secondTimestamp)
```
0xnm, line 97: we need this one to be `false`, so probably it is better to make sure that this condition is satisfied by doing `check(recordingTimeBank.updateAndCheck(secondTimestamp) == false)`

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
```diff
@@ -0,0 +1,133 @@
...
+            min = firstTimestamp + firstExecutionTime,
+            max = firstTimestamp + firstExecutionTime + interval
+        )
+
+        val thirdTimestamp =
+            forge.aLong(
+                min = secondTimestamp + firstExecutionTime /
+                    ((TimeUnit.SECONDS.toMillis(1) / TEST_MAX_BALANCE_IN_MS))
+            )
+        // When
```
0xnm, line 94:
```suggestion
            )

        // When
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
```diff
@@ -0,0 +1,133 @@
...
+        val actual = recordingTimeBank.updateAndCheck(thirdTimestamp)
+
+        // Then
+        assertThat(actual).isEqualTo(true)
+        //
+    }
+
+    @Test
+    fun `M allow everything W set max balance to 1000ms per sec`(forge: Forge) {
+        recordingTimeBank = RecordingTimeBank(1000)
```
0xnm, line 107: what if it is more than 1s? shouldn't we cover this as well?

0xnm (approved):
lgtm, I added some suggestions

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/recorder/RecordingTimeBankTest.kt
ambushwork: Indeed, I updated the test to `M allow everything W set max balance more than 1000ms per sec` with a random value with 1000ms as min value.

0xnm (approved):

PR #2251, opened 2024-09-09
Merge 2.13.1

xgouchet:


buildSrc/src/main/kotlin/com/datadog/gradle/config/AndroidConfig.kt
```diff
@@ -64,6 +64,11 @@ fun Project.androidLibraryConfig() {
             unitTests.isReturnDefaultValues = true
         }
 
+        @Suppress("UnstableApiUsage")
+        testFixtures {
+            enable = true
```
0xnm, line 69: I think this change should be reverted

xgouchet: 👍

0xnm (approved):

jonathanmos (approved):

PR #2257, opened 2024-09-11
RUM-6104 ensure UploadWorker uses the sdk instance name

### LINKED ISSUE #2252: How can i populate DATADOG_INSTANCE_NAME in UploadWorker.kt?
### Question

When using datadog in a library, i initilize it with an instance name but the name does not get populated into `DATADOG_INSTANCE_NAME` in the `UploadWorker.kt`, this results in an exception thrown inside the worker:

```
SDK instance with name _dd.sdk_core.default is not found, returning no-op implementation. 
Please make sure to call Datadog.initialize([instanceName]) before getting the instance.
SDK instance was requested from: (Ask Gemini)

at com.datadog.android.Datadog.getInstance(Datadog.kt:139)
at com.datadog.android.core.internal.data.upload.UploadWorker.doWork(UploadWorker.kt:42)
at androidx.work.Worker$1.run(Worker.java:82)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
at java.lang.Thread.run(Thread.java:1012)
2024-09-09 11:32:40.574 19812-20100 Datadog 
co.onestep.android.demoapp
E  Datadog has not been initialized.
```

Since those lines
```
 val instanceName = inputData.getString(DATADOG_INSTANCE_NAME)
 val sdkCore = Datadog.getInstance(instanceName) as? InternalSdkCore
```

return a null instance name and the call to Datadog.getInstance(instanceName/*this is always null/*) throws an exception

xgouchet:
### What does this PR do?

When the UploadWorker is woken by the system, it reads the SDK Instance name from the WorkRequest input data, and tries to get the instance with that name. 

However, the instance name was never sent meaning that for applications specifying an instance name would never find a relevant SDK. 

### Motivation

Fixes #2252

### DIFF
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
@@ -19,6 +19,7 @@ import com.datadog.android.core.internal.ContextProvider
 import com.datadog.android.core.internal.CoreFeature
 import com.datadog.android.core.internal.DatadogCore
 import com.datadog.android.core.internal.SdkFeature
+import com.datadog.android.core.internal.lifecycle.ProcessLifecycleCallback
 import com.datadog.android.core.internal.lifecycle.ProcessLifecycleMonitor
 import com.datadog.android.core.internal.net.DefaultFirstPartyHostHeaderTypeResolver
 import com.datadog.android.core.internal.net.info.NetworkInfoProvider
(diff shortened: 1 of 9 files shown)


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -803,6 +805,20 @@ internal class DatadogCoreTest {
         assertThat(testedCore.isDeveloperModeEnabled).isFalse()
     }
 
+    @RepeatedTest(128)
```
0xnm, line 808: Shouldn't it be just `@Test` here? Is it needed to repeat it so many times?

0xnm (approved):

ambushwork:
Question: in the description of the PR it says before the fix, ".. meaning that for applications specifying an instance name would never find a relevant SDK." I wonder how `UploadWorker` can work correctly after a while of application running?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
xgouchet: You're right, i used it locally :D

ambushwork (approved):

mariusc83 (approved):

PR #2258, opened 2024-09-11
RUM-6129 Introduce the API usage telemetry event and API

mariusc83:
### What does this PR do?

In this PR we are introducing the new `InternalLogger#logApisUsage` method which will be used to report usage of different APIs as Telemetry from our sdk. In order to achieve this we had to refactor and simplify the whole `Telemetry` related logic in the SDK to make it more scalable. For this we introduced the new `InternalTelemetryEvent` which will be shared through out all the module from the `dd-sdk-android-internal` module.

### Motivation

### Additional Notes


xgouchet (changes requested):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -502,17 +503,20 @@ internal class DatadogCore(
...
-                "use_persistence_strategy_factory" to (configuration.coreConfig.persistenceStrategyFactory != null)
+            val event = InternalTelemetryEvent.Configuration(
+                trackErrors = configuration.crashReportsEnabled,
+                batchSize = configuration.coreConfig.batchSize.windowDurationMs,
+                useProxy = configuration.coreConfig.proxy != null,
+                useLocalEncryption = configuration.coreConfig.encryption != null,
+                batchUploadFrequency = configuration.coreConfig.uploadFrequency.baseStepMs,
+                batchProcessingLevel = configuration.coreConfig.batchProcessingLevel.maxBatchesPerUploadJob
+            )
+            rumFeature.sendEvent(
```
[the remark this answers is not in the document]
mariusc83: hmmm actually I see your point also regarding the first comment. I don't know why I kept using the Map...we could just send the `InternalTelemetryEvent` everywhere

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -298,6 +292,20 @@ internal class RumFeature(
 
     // region Internal
 
+    private fun handleTelemetryEvent(event: Map<*, *>) {
```
0xnm, line 295: probably with that change we can send class directly instead of using `Map`? It will require a change of the check right at the beginning of `onReceive`, but it is a minor thing.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumRawEvent.kt
```diff
@@ -212,16 +211,9 @@ internal sealed class RumRawEvent {
...
-        val type: TelemetryType,
-        val message: String,
-        val stack: String?,
-        val kind: String?,
-        val coreConfiguration: TelemetryCoreConfiguration?,
-        val additionalProperties: Map<String, Any?>?,
-        val onlyOnce: Boolean = false,
-        override val eventTime: Time = Time(),
-        val isMetric: Boolean = false
+    internal data class TelemetryEventWrapper(
```
0xnm, line 214: we probably can keep the same name for the class - `SendTelemetry`. but not a big deal

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -74,38 +87,35 @@ internal class TelemetryEventHandler(
...
+                    )
+                }
+
+                is InternalTelemetryEvent.InterceptorInstantiated -> {
                     trackNetworkRequests = true
                     null
                 }
-            }
 
+                else -> null
```
0xnm, line 117: shouldn't we log here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -322,12 +332,57 @@ internal class TelemetryEventHandler(
...
+                        ),
+                        usage = TelemetryUsageEvent.Usage.AddViewLoadingTime(
+                            overwritten = event.overwrite,
+                            noView = event.noView,
+                            noActiveView = event.noActiveView
+                        )
+                    )
+                )
+            }
+            else -> null
```
0xnm, line 382: we don't need `else` here I think, because we are using sealed class. So if we remove `else` here and add one more sub-type to `InternalTelemetryEvent.ApiUsage`, without added additional branch in `when`, we will get a compilation error, which is for good.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
```diff
@@ -132,6 +133,18 @@ interface InternalLogger {
...
 
+    /**
+     * Logs an API usage from the internal implementation.
+     * @param apiUsageEvent the API event being tracked
+     * @param samplingRate value between 0-100 for sampling the event. Note that the sampling rate applied to this
+     * event will be applied in addition to the global telemetry sampling rate.
+     */
+    @InternalApi
+    fun logApiUsage(
+        apiUsageEvent: InternalTelemetryEvent.ApiUsage,
```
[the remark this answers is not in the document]
mariusc83: I didn't see any but will try to test it from the sample app

features/dd-sdk-android-rum/api/dd-sdk-android-rum.api
```diff
@@ -5146,6 +5146,239 @@ public final class com/datadog/android/sqlite/DatadogDatabaseErrorHandler : andr
...
+	public synthetic fun <init> (Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$Dd;JLjava/lang/String;Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$Source;Ljava/lang/String;Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$Application;Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$Session;Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$View;Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$Action;Ljava/util/List;Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$Telemetry;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun component1 ()Lcom/datadog/android/telemetry/model/CommonTelemetryProperties$Dd;
```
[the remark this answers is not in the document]
mariusc83: I am going to have a look for something but not going to spend too much on this

features/dd-sdk-android-rum/generate_telemetry_models.gradle.kts
```diff
@@ -15,10 +15,6 @@ tasks.register(
 ) {
     inputDirPath = "src/main/json/telemetry"
     targetPackageName = "com.datadog.android.telemetry.model"
-    ignoredFiles = arrayOf(
-        "_common-schema.json",
-        "usage-schema.json"
```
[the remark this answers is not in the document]
mariusc83: oh ok...thought it will need also the `_common-schema` as `usage-schema` is referencing it

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
mariusc83: yup exactly what @xgouchet was pointing out also, I am going to apply the changes now.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
mariusc83: actually that's a sealed class so we don't really need the `else` here anyway. In case we do not treat a case we will be warned by the editor.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
mariusc83: yup it is exactly the case in line `117` also so you are right

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -251,45 +248,50 @@ internal class RumFeature(
...
                 sdkCore.internalLogger.log(
                     InternalLogger.Level.WARN,
                     InternalLogger.Target.USER,
-                    { UNKNOWN_EVENT_TYPE_PROPERTY_VALUE.format(Locale.US, event["type"]) }
+                    { UNSUPPORTED_EVENT_TYPE.format(Locale.US, event::class.java.canonicalName) }
                 )
+                return
+            }
+
+            else -> when (event["type"]) {
```
0xnm, line 271: I think it is better to swap these branches, and just do:

```
is Map<*, *> ->
else ->
```

reason: `!is` is harder to read and then in `else` it it takes some time to understand that type of the object there is `Map`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -656,8 +596,8 @@ internal class RumFeature(
         internal const val LOG_ERROR_WITH_STACKTRACE_EVENT_MISSING_MANDATORY_FIELDS =
             "RUM feature received a log event with stacktrace" +
                 " where mandatory message field is either missing or has a wrong type."
-        internal const val TELEMETRY_MISSING_MESSAGE_FIELD = "RUM feature received a telemetry" +
-            " event, but mandatory message field is either missing or has a wrong type."
+        internal const val TELEMETRY_MISSING_EVENT_FIELD_WARNING_MESSAGE = "RUM feature received a telemetry" +
```
0xnm, line 599: is it still needed?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -44,19 +47,20 @@ internal class TelemetryEventHandler(
...
 
         eventIDsSeenInCurrentSession.add(event.identity)
         totalEventsSeenInCurrentSession++
-
         sdkCore.getFeature(Feature.RUM_FEATURE_NAME)?.withWriteContext { datadogContext, eventBatchWriter ->
-            val timestamp = event.eventTime.timestamp + datadogContext.time.serverTimeOffsetMs
-            val telemetryEvent: Any? = when (event.type) {
-                TelemetryType.DEBUG -> {
+            val timestamp = wrappedEvent.eventTime.timestamp + datadogContext.time.serverTimeOffsetMs
+            val telemetryEvent: Any? = when (event) {
```
0xnm, line 62: does it become exhaustive `when`? The reason why I'm asking it because there are 2 sealed classes involved here, not only one. Will it be a compilation error if we add say `InternalTelemetry.Log.XXX` into classes definition?

0xnm (approved):
I added few comments/suggestions, but nothing blocking.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
mariusc83: I think the way it behaves is that whenever you add a new type of the sealed class it will warn you that is not handled but will double check it.

0xnm: yes, but you now have 2 sealed base classes in this hierarchy, not one, that is why I'm curious.

mariusc83: Yeah, I confirm whenever you add a new type to that sealed class it is only to crash at compile time saying that you need to be exhaustive.

0xnm (approved):

xgouchet (approved):

PR #2259, opened 2024-09-11
RUM-5184: Add dynamic optimization field in SessionReplayConfiguration

ambushwork:
### What does this PR do?

[Adding API changes] Add dynamic optimization field in SessionReplayConfiguration

### Motivation

RUM-5184

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
@@ -22,7 +22,8 @@ data class SessionReplayConfiguration internal constructor(
     internal val imagePrivacy: ImagePrivacy,
     internal val startRecordingImmediately: Boolean,
     internal val touchPrivacy: TouchPrivacy,
-    internal val textAndInputPrivacy: TextAndInputPrivacy
+    internal val textAndInputPrivacy: TextAndInputPrivacy,
+    internal val dynamicOptimizationEnabled: Boolean
 ) {
 
     /**
@@ -42,6 +43,7 @@ data class SessionReplayConfiguration internal constructor(
         private var touchPrivacy = TouchPrivacy.HIDE
         private var textAndInputPrivacy = TextAndInputPrivacy.MASK_ALL
         private var extensionSupport: ExtensionSupport = NoOpExtensionSupport()
+        private var dynamicOptimizationEnabled = true
 
         /**
          * Adds an extension support implementation. This is mostly used when you want to provide
@@ -150,6 +152,15 @@ data class SessionReplayConfiguration internal constructor(
         }
 
         /**
+         * This option controls whether optimization is enabled or disabled for recording Session Replay data.
+         * By default the value is true, meaning the dynamic optimization is enabled.
+         */
+        fun setDynamicOptimizationEnabled(dynamicOptimizationEnabled: Boolean): Builder {
+            this.dynamicOptimizationEnabled = dynamicOptimizationEnabled
+            return this
+        }
+
+        /**
          * Builds a [SessionReplayConfiguration] based on the current state of this Builder.
          */
         fun build(): SessionReplayConfiguration {
@@ -162,7 +173,8 @@ data class SessionReplayConfiguration internal constructor(
                 customMappers = customMappers(),
                 customOptionSelectorDetectors = extensionSupport.getOptionSelectorDetectors(),
                 sampleRate = sampleRate,
-                startRecordingImmediately = startRecordingImmediately
+                startRecordingImmediately = startRecordingImmediately,
+                dynamicOptimizationEnabled = dynamicOptimizationEnabled
             )
         }
 
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/DebouncerTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/DebouncerTest.kt
@@ -12,6 +12,7 @@ import com.datadog.android.api.feature.FeatureScope
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.sessionreplay.forge.ForgeConfigurator
 import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.annotation.BoolForgery
 import fr.xgouchet.elmyr.junit5.ForgeConfiguration
 import fr.xgouchet.elmyr.junit5.ForgeExtension
 import org.assertj.core.api.Assertions.assertThat
@@ -50,6 +51,9 @@ internal class DebouncerTest {
     @Mock
     lateinit var mockRumFeature: FeatureScope
 
+    @BoolForgery
+    var fakeDynamicOptimizationEnabled: Boolean = false
+
     lateinit var testedDebouncer: Debouncer
 
     @BeforeEach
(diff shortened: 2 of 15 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -76,6 +78,16 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Sets the dynamic optimisation for the Session Replay feature.
+         * This will enable the time limitation that Session Replay recording can use on the main thread, enabling it
+         * might reduce the session replay frames recorded by the SDK.
```
0xnm, line 84:
```suggestion
         * This will enable the time limitation that Session Replay recording can use on the main thread. Enabling it
         * might reduce the session replay frames recorded by the SDK.
```

but still, for the customers this explanation is somewhat cryptic. Maybe we should re-word it a bit simpler/more understandable? Something like:

```
This option enables/disables the optimization, which can limit amount of the Session Replay data recorded. If optimization is enabled, there is a time quota per second which Session Replay can utilize to record the data.
```

A bit simplified, but you see the idea.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -76,6 +78,16 @@ data class SessionReplayConfiguration internal constructor(
...
+         */
+        fun setDynamicOptimisation(dynamicOptimization: Boolean): Builder {
```
0xnm, line 86: we need to choose which spelling we use: `optimiSation` or `optimiZation`

ambushwork: Yes, that's the question I always wanted to ask, do we prefer usually British English or American English in the code?

mariusc83 (dismissed):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
0xnm: We never had the explicit rule for this one. Maybe you can search across the codebase and see which one is used the most. Anyway, my point here is that it should be consistent naming for this declaration.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
jonathanmos: Idk if we should specifically refer to specfic mechanisms in the comment for this parameter, because it might equally apply for disabling the sdk on weak devices on startup, or any other type of dynamic performance improvement that we want to do now or in the future.

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
ambushwork: I reverse the meaning of the parameter, now it changed from `dynamicOptimization` to `disableDynamicOptimization`, document updated as well, please check it again

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
ambushwork: So I choose `Optimization`, updated everywhere.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/DebouncerTest.kt
```diff
@@ -60,20 +64,50 @@ internal class DebouncerTest {
...
             sdkCore = mockSdkCore,
-            timeBank = mockTimeBank
+            timeBank = mockTimeBank,
+            disableDynamicOptimization = fakeDisableDynamicOptimization
         )
     }
 
     @Test
-    fun `M send telemetry W frame is skipped by time bank`() {
+    fun `M not optimise W dynamic optimization is false`() {
```
0xnm, line 73:
```suggestion
    fun `M not optimize W dynamic optimization is false`() {
```

0xnm (dismissed):
I left a few comments/suggestions, but overall lgtm. I'm not sure about the parameter description, it is better to check with @jonathanmos.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -149,6 +151,16 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        /**
+         * This option controls whether optimization is enabled or disabled for recording Session Replay data.
```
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -149,6 +151,16 @@ data class SessionReplayConfiguration internal constructor(
...
+         * When optimization is enabled, a time quota is set per second, which limits the amount of data that can
```
jonathanmos, line 156: Here we're referring to a specific optimization. In the future we'll disable devices that have low specs and potentially other improvements, so will we want to list them all here, or to be more ambiguous about what optimizations we are performing?

jonathanmos (approved):
LGTM except for minor comments

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
ambushwork: I think we can be more ambiguous instead of listing them all, because currently we haven't implemented all the improvements yet.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/DebouncerTest.kt
```diff
@@ -60,20 +64,50 @@ internal class DebouncerTest {
...
             sdkCore = mockSdkCore,
-            timeBank = mockTimeBank
+            timeBank = mockTimeBank,
+            disableDynamicOptimization = fakeDisableDynamicOptimization
         )
     }
 
     @Test
-    fun `M send telemetry W frame is skipped by time bank`() {
+    fun `M not optimize W dynamic optimization is false`() {
```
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
ambushwork: updated

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -149,6 +151,15 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        /**
+         * This option controls whether optimization is enabled or disabled for recording Session Replay data.
+         * By default the value is false, meaning the dynamic optimization is enabled.
+         */
+        fun disableDynamicOptimization(disableDynamicOptimization: Boolean): Builder {
```
xgouchet, line 158: > [!WARNING]
> Having a "negative" name with a boolean can become confusing for users. When I pass in false, do I want the feature to be disable (meaning off) or not disabled. Also if the method is called `disabledX`, one could try to find a matching `enableX`. A better way would be to have something like `setDynamicOptimizationEnabled(enabled: Boolean)` which is easier to understand.

ambushwork: After a discussion with @jonathanmos , we agree that to change the variable name to `dynamicOptimizationEnabled`

0xnm: I think it should be `setDynamicOptimizationEnabled` then, because we should be consistent with the lingvo used in the our API already. You can search for `*enable*` in `apiSurface` files to see how it should be consistent.

ambushwork: good point, updated

xgouchet (approved):

jonathanmos (approved):

PR #2264, opened 2024-09-17
RUM-5182: Add precheck conditions when registering session replay feature

ambushwork:
### What does this PR do?

* Add `SystemRequirementConfiguration` class to contain requirements to check
* Add `SystemRequirementChecker` to check if every condition can be met:
    * `MemoryRequirementChecker` implementation to check max RAM size
    * `CpuRequirementChecker` implementation to check cpu cores number
* Add `SessionReplayRequirement` class to use `SystemRequirementConfiguration` and all the checkers to determine if the task can be executed after checking the conditions.

### Motivation

RUN-5182


dd-sdk-android-core/api/apiSurface
```diff
@@ -265,6 +265,7 @@ interface com.datadog.android.core.internal.persistence.Deserializer<P: Any, R:
   fun deserialize(P): R?
 fun java.io.File.canReadSafe(com.datadog.android.api.InternalLogger): Boolean
 fun java.io.File.existsSafe(com.datadog.android.api.InternalLogger): Boolean
+fun java.io.File.listFilesSafe(com.datadog.android.api.InternalLogger, java.io.FilenameFilter): Array<java.io.File>?
```
0xnm, line 268: nit: arguments order should be flipped, `InternalLogger` should come last as for the other similar methods.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/FileExt.kt
```diff
@@ -118,6 +119,17 @@ internal fun File.listFilesSafe(filter: FileFilter, internalLogger: InternalLogg
     }
 }
 
+/**
+ * Non-throwing version of [File.listFiles]. If exception happens, false is returned.
```
0xnm, line 123:
```suggestion
 * Non-throwing version of [File.listFiles]. If exception happens, null is returned.
```

detekt_custom.yml
```diff
@@ -816,6 +817,7 @@ datadog:
       - "kotlin.collections.List.findFirstForType(java.lang.Class)"
       - "kotlin.collections.List.firstOrNull()"
       - "kotlin.collections.List.firstOrNull(kotlin.Function1)"
+      - "kotlin.collections.List.firstNotNullOfOrNull(kotlin.Function1)"
```
0xnm, line 820: should come above according to the ordering

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/CpuRequirementChecker.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun name(): String = CPU_CHECK_NAME
+
+    override fun checkedValue(): Any? = checkedValue
+
+    private fun readCpuCoreNumber(): Int {
+        val files = cpuDirFile.listFilesSafe(internalLogger) { _, name -> name.matches(REGEX_CPU_CORE_FILE) }
+        val value = files?.size ?: fallbackReadCpuCoreNumber()
+        checkedValue = value
+        return value
+    }
```
0xnm, line 33: this will trigger `StrictMode` violations because of the I/O on the main thread

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/CpuRequirementChecker.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.internal.persistence.file.listFilesSafe
+import java.io.File
+
+internal class CpuRequirementChecker(
+    private val cpuDirFile: File = File(DIR_PATH),
+    private val internalLogger: InternalLogger
+) : SystemRequirementChecker {
+
+    private var checkedValue: Int? = null
```
0xnm, line 18: shouldn't it be simply `lateinit`? I guess number of cores cannot change over the run of the application?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/MemoryRequirementChecker.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    override fun name(): String = MEMORY_CHECK_NAME
+
+    override fun checkedValue(): Any? = checkedValue
+
+    private fun getMaxHeapSize(): Long {
+        if (!(memInfoFile.existsSafe(internalLogger) && memInfoFile.canReadSafe(internalLogger))) {
+            return 0L
+        }
```
0xnm, line 32: same: it will trigger `StrictMode` violation

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/MemoryRequirementChecker.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        val memorySizeKb = memInfoFile.readLinesSafe(internalLogger = internalLogger)?.firstNotNullOfOrNull { line ->
```
0xnm, line 34: minor: maybe it is easier to use: `filterNotNull + firstOrNull` instead? I find `firstNotNullOfOrNull` hard to read/understand

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/MemoryRequirementChecker.kt
```diff
@@ -0,0 +1,77 @@
...
+            if (line.startsWith(MEM_TOTAL_REGEX)) {
+                val tokens: List<String> = line.split(Regex("\\s+"))
```
0xnm, line 36:
```suggestion
                val tokens = line.split(Regex("\\s+"))
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SessionReplayRequirement.kt
```diff
@@ -0,0 +1,74 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.presequite
+
+import com.datadog.android.api.InternalLogger
+import okhttp3.internal.toImmutableMap
```
0xnm, line 10: we shouldn't use `okhttp3.internal` package, it doesn't exist anymore in OkHttp 5.x, see https://github.com/DataDog/dd-sdk-android/issues/1282

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SessionReplayRequirement.kt
```diff
@@ -0,0 +1,74 @@
...
+            it.checkMinimumRequirement(systemRequirementConfiguration)
+        }
+    }
+
+    private fun getCheckerReport(): Map<String, Any?> {
+        val map = mutableMapOf<String, Any?>()
+        checkers.forEach { checker ->
+            map[checker.name()] = checker.checkedValue()
+        }
+        return map.toImmutableMap()
```
0xnm, line 52: you can simply call `toMap` here instead if you want non-mutable instance or just `return map`, because the return type specified is read only.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SessionReplayRequirement.kt
```diff
@@ -0,0 +1,74 @@
...
+        private var checkers = mutableListOf<SystemRequirementChecker>()
+        private var systemRequirementConfiguration: SystemRequirementConfiguration =
+            SystemRequirementConfiguration.DEFAULT
+
+        fun addChecker(checker: SystemRequirementChecker): Builder {
+            checkers.add(checker)
+            return this
+        }
+
+        fun setSystemRequirementConfiguration(systemRequirementConfiguration: SystemRequirementConfiguration): Builder {
```
0xnm, line 65: side thought: with such API if I, for example, add requirement configuration, but forget to add checker for the parameter in the configuration, then it will pass even if requirement is not satisfied?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SessionReplayRequirement.kt
ambushwork: It's imported automatically, I will change it to `toMap`

dd-sdk-android-core/api/apiSurface
0xnm: if it is lambda, we can do like that indeed

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SessionReplayRequirement.kt
0xnm: yes, my point is rather if it is expected behavior. Normally, if there is a requirement in the config, but no checker for it registered, imo we at least should log that.

ambushwork: indeed, I will add a log to `User` if you agree with it.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,42 @@
...
+                cpuCoreNumber = cpuCoreNumber,
+                memorySizeInMb = memorySize
+            )
+        }
+    }
+
+    companion object {
+        val DEFAULT = Builder()
+            .setMemorySize(256)
+            .setCpuCoreNumber(1)
```
jonathanmos, line 39: won't this never trigger for cpu because there is always at least 1 cpu core?

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/presequite/MemoryRequirementCheckerTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        fakeMemoryFile = File(tempDir, "meminfo")
+        testedMemoryRequirementChecker = MemoryRequirementChecker(
+            memInfoFile = fakeMemoryFile,
+            internalLogger = mockInternalLogger
+        )
+    }
+
+    @Test
+    fun `M have correct name W call name()`() {
+        assertThat(testedMemoryRequirementChecker.name()).isEqualTo("ram")
```
jonathanmos, line 67: missing`Then`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
```diff
@@ -11,6 +11,10 @@ import com.datadog.android.api.SdkCore
 import com.datadog.android.api.feature.Feature.Companion.SESSION_REPLAY_FEATURE_NAME
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.sessionreplay.internal.SessionReplayFeature
+import com.datadog.android.sessionreplay.internal.presequite.CpuRequirementChecker
+import com.datadog.android.sessionreplay.internal.presequite.MemoryRequirementChecker
+import com.datadog.android.sessionreplay.internal.presequite.SessionReplayRequirement
+import com.datadog.android.sessionreplay.internal.presequite.SystemRequirementConfiguration
```
jonathanmos, line 17: presequite should be prerequisite I think

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/MemoryRequirementChecker.kt
ambushwork: break down to `mapNotNull` + `.firstOrNull`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/CpuRequirementChecker.kt
ambushwork: Indeed, But we can't ensure from this class inside that `readCpuCoreNumber` will be correctly called before `checkedValue`, so to avoid crash as possible as we can, I prefer a nullable value here, even if it is null, it will just have a empty value of cpu cores in telemetry.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/MemoryRequirementChecker.kt
ambushwork: removed

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SessionReplayRequirement.kt
ambushwork: updated

detekt_custom.yml
ambushwork: removed due to the suggested change below

detekt_custom.yml
```diff
@@ -898,6 +900,7 @@ datadog:
       - "kotlin.collections.MutableList.add(kotlin.Int, com.datadog.android.sessionreplay.model.MobileSegment.Wireframe)"
       - "kotlin.collections.MutableList.add(kotlin.Int, com.datadog.android.sessionreplay.model.MobileSegment.Wireframe.ShapeWireframe)"
       - "kotlin.collections.MutableList.add(kotlin.String)"
+      - "kotlin.collections.MutableList.add(com.datadog.android.sessionreplay.internal.presequite.SystemRequirementChecker)"
```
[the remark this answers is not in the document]
ambushwork: updated

detekt_custom.yml
```diff
@@ -950,6 +953,7 @@ datadog:
       - "kotlin.collections.MutableMap.remove(kotlin.String)"
       - "kotlin.collections.MutableMap.toMap()"
       - "kotlin.collections.MutableMap.toMutableMap()"
+      - "kotlin.collections.MutableMap.toImmutableMap()"
```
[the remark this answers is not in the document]
ambushwork: removed due to the suggested change below

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/CpuRequirementChecker.kt
ambushwork: Good point, now I dispatch the `SessionReplayRequirement.=runIfMeetRequirements` function into another thread, and unit test is updated accordingly, please check it.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/presequite/SystemRequirementConfiguration.kt
ambushwork: The idea is to set a quite low requirement at first, to have less impact on current usage, then we will raise it depending on our benchmark

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
ambushwork: Sorry for my English, I put it into Google translate it didn't correct me for the spell. updated now

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.internal.utils.submitSafe
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+
+internal class SessionReplayRequirement private constructor(
+    private val checkers: List<SystemRequirementChecker>,
+    private val systemRequirementConfiguration: SystemRequirementConfiguration
+) {
+    private var executorService: ExecutorService = Executors.newSingleThreadExecutor()
```
0xnm, line 18: We shouldn't spin new thread if it is possible to re-use existing ones. Here we create a thread for a single operation and then it is not needed. It is possible to use a thread from some existing pool?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    fun runIfMeetRequirements(
+        internalLogger: InternalLogger,
+        runnable: () -> Unit
+    ) {
+        executorService.submitSafe(OPERATION_NAME, internalLogger) {
```
0xnm, line 24: are we ok that Session Replay is going to start with delay? And it is going to be initialized on non-main thread.

features/dd-sdk-android-session-replay/build.gradle.kts
```diff
@@ -74,6 +74,12 @@ unMock {
     keep("android.widget.ImageView\$ScaleType")
     keep("android.graphics.Rect")
     keep("android.graphics.drawable.GradientDrawable")
+    keep("android.os.Handler")
+    keep("android.os.IMessenger")
+    keep("android.os.Looper")
+    keep("android.os.Message")
+    keep("android.os.MessageQueue")
+    keep("android.os.SystemProperties")
```
0xnm, line 82: this is not needed if `Looper` is not used

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirementTest.kt
```diff
@@ -0,0 +1,261 @@
...
+
+    companion object {
+        private val mainLooper = MainLooperTestConfiguration()
+
+        @TestConfigurationsProvider
+        @JvmStatic
+        fun getTestConfigurations(): List<TestConfiguration> {
+            return listOf(mainLooper)
+        }
+    }
```
0xnm, line 260: We don't need a `Looper`. This is a concept from Android and this test is running on JVM, not on Android. You can simply use existing Thread API to spin a new thread and block the existing one until the execution is not completed for the spinned one.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
ambushwork: Based on the fact that we need to read some files asynchronously, it seems that we don't have a choice to make it start with a delay. And Me and Jonathan currently don't see a side-effect to register the feature on a non-main thread.

features/dd-sdk-android-session-replay/build.gradle.kts
ambushwork: Removed

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirementTest.kt
ambushwork: Updated, now the unit test just simply check if the thread where the runnable is executed is the same as the one of the call site

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
ambushwork: I don't see any shared thread pool across the modules in our sdk, do you have any suggestion?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
0xnm: > we need to read some files asynchronously before starting

Well, we can still do everything on the main thread and just suppress StrictMode violations (check `StrictModeExt#allowThreadDiskReads`).

These violations don't take into account the amount of penalty, obviously. If I/O on the main thread is caused by very small and fast operation which doesn't add any lag to the FPS, then it is not that critical and may be done.

So it may be violation, but it doesn't cause any significant impact.

But anyway, final decision comes from the profiling data and the architecture.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementCheckerTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        fakeMemoryFile = File(tempDir, "meminfo")
+        testedMemoryRequirementChecker = MemoryRequirementChecker(
+            memInfoFile = fakeMemoryFile,
+            internalLogger = mockInternalLogger
+        )
+    }
+
+    @Test
+    fun `M have correct name W call name()`() {
+        // Given
```
jonathanmos, line 67: Empty Given and When blocks

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
ambushwork: After the discussion with @jonathanmos , we prefer not violate Strict mode here, meaning it's dispatched to a different thread to check the requirements, after it's done, we switch back to main thread to register session replay feature. and `ExecutorService` will be shutdown.

ambushwork:
@0xnm @jonathanmos I just refactored the PR, please review it again if you want, main changes:
* Add public API to `SessionReplayConfiguration` to set `SystemRequirementConfigutation`
* Switch back to main thread to register feature after checking the requirements.
*

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -160,6 +162,16 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Defines the minimum system requirements for enabling the Session Replay feature.
+         * When [SessionReplay.enable] is invoked, the system configuration is verified against these requirements.
+         * If the system meets the specified criteria, Session Replay will be successfully enabled.
+         */
+        fun setSystemRequirement(systemRequirementConfiguration: SystemRequirementConfiguration): Builder {
```
0xnm, line 170: should it be `setSystemRequirements` (because there is at least two: CPU and memory)

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,113 @@
...
+    private var systemRequirement: SystemRequirement? = null
+) {
+
+    internal fun runIfMeetRequirements(
+        internalLogger: InternalLogger,
+        runnable: () -> Unit
+    ) {
+        if (systemRequirement == null) {
+            systemRequirement = DefaultSystemRequirement()
+        }
```
0xnm, line 33:
```suggestion
    private var systemRequirement: SystemRequirement = DefaultSystemRequirement()
) {

    internal fun runIfMeetRequirements(
        internalLogger: InternalLogger,
        runnable: () -> Unit
    ) {
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,113 @@
...
+        systemRequirement?.runIfMeetRequirements(
+            systemRequirementConfiguration = this,
```
0xnm, line 35: This API is confusing: `SystemRequirement` is a part of `SystemRequirementConfiguration`, but has a method which takes `systemRequirementConfiguration`. And both have the method with same name. Won't it be easier to detach `SystemRequirement` from configuration and have 2 independent objects, which can be called in `SessionReplay.enable`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,113 @@
...
+            .setMinCPUCoreNumber(2)
+            .build()
+
+        /**
+         * A special instance representing no system requirements for enabling the Session Replay feature.
+         *
+         * With this instance, Session Replay will be enabled regardless of system specifications.
+         * This is useful in cases where system requirements should not restrict functionality.
+         */
+        val NONE: SystemRequirementConfiguration = Builder().build()
```
0xnm, line 111: minor: I think you can simplify the things here a bit and just call `SystemRequirementConfiguration` constructor here directly.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CpuRequirementChecker.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    private fun readCpuCoreNumber(): Int {
+        val files = cpuDirFile.listFilesSafe(internalLogger) { _, name -> name.matches(REGEX_CPU_CORE_FILE) }
+        val value = files?.size ?: fallbackReadCpuCoreNumber()
+        checkedValue = value
+        return value
+    }
+
+    private fun fallbackReadCpuCoreNumber(): Int {
+        return Runtime.getRuntime().availableProcessors()
```
0xnm, line 37: why we cannot rely solely on this and need to read `/sys/devices/system/cpu/` as a primary option?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
0xnm: As a little trick you can run it on `InternalSdkCore#persistenceExecutor`. The only drawback here is that it may be already many items there to process, but I guess it shouldn't be the case for the app start lifecycle stage (if we assume that Session Replay is initialized there).

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,113 @@
...
+        /**
+         * Sets the minimum RAM in megabytes.
+         *
+         * @param minRAMSize The minimum RAM in megabytes.
+         * @return The current Builder instance for chaining.
+         */
+        fun setMinRAM(minRAMSize: Int): Builder {
+            this.minRAMSize = minRAMSize
+            return this
+        }
```
0xnm, line 74: I have a concern about this one. Here the naming can imply that we are relying on the min RAM size for the device (let's say we want to run SR only on high-end devices and set 6000 here), but the actual check is comparing with max heap size. And max heap size != max device RAM size. Heap size for the process is assigned by the Zygote and in any case lower that max device RAM size.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CpuRequirementChecker.kt
ambushwork: In its doc it says:
> Returns the number of processors available to the Java virtual machine.


so that might not be the physical number of cores  in some cases I think

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
ambushwork: Here I leave the constructor as `null` by default because I don't want `DefaultSystemRequirement` to be initialised directly because it uses `Handler(Looper.getMainLooper())` as default handler, and I have `NONE` and `DEFAULT` instance in `companion object` will use it anyway, which can cause trouble in unit test.

so I prefer assign it when really use it. what do you think?

0xnm: I'm not sure what is the problem with unit-test, can you elaborate?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
0xnm: it seems we are loosing only very small piece being non-testable due to such move towards `SessionReplay.enable` (basically a single method call verification, I guess?), in exchange getting more clear class structure.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
ambushwork: yes, so we have two companion objects `DEFAULT`, `NONE` in `SystemRequirementConfiguration` class, they will be initialised directly at the very beginning of the runtime.
they call the constructor to build an instance of `SystemRequirementConfiguration`, so if we have `DefaultSystemRequirement` in the constructor, they will initialise it also with ` Handler(Looper.getMainLooper())` as the default handler.

In unit test, we need some setup to make `Looper.getMainLooper` become non-null, and if I understand correctly,  the setup happens during `@BeforeEach`, which will be too late.  And unit test throws error due to the null result of `Looper.getMainLooper`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SessionReplayRequirement.kt
ambushwork: I passed `sdkCore` inside `runIfMeetRequirements` function to get the persistance executor service.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * This class defines the system requirements necessary to enable the Session Replay feature.
+ *
+ * @param minCPUCoreNumber The minimum CPU core number.
+ * @param minRAMSize The minimum RAM in megabytes.
+ */
+class SystemRequirementConfiguration internal constructor(
+    val minCPUCoreNumber: Int,
+    val minRAMSize: Int,
```
jonathanmos, line 26: We could rename to minRAMSizeMb to make it clearer. WDYT?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.Executors
+
+/**
+ * This class defines the system requirements necessary to enable the Session Replay feature.
+ *
+ * @param minCPUCoreNumber The minimum CPU core number.
+ * @param minRAMSize The minimum RAM in megabytes.
+ */
+class SystemRequirementConfiguration internal constructor(
+    val minCPUCoreNumber: Int,
```
jonathanmos, line 25: How about minCPUCores?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/DefaultSystemRequirement.kt
```diff
@@ -0,0 +1,73 @@
...
+            if (checkAllMinimumRequirements(systemRequirementConfiguration, checkers)) {
+                handler.post {
+                    runnable()
+                }
+            } else {
+                internalLogger.log(
+                    level = InternalLogger.Level.INFO,
+                    listOf(InternalLogger.Target.TELEMETRY, InternalLogger.Target.USER),
+                    messageBuilder = {
+                        "Session replay is disabled because the system doesn't meet the minimum SDK requirements"
```
jonathanmos, line 38: I would remove the word 'SDK' to prevent customer confusion. Here we are setting minimum requirements for Session Replay, but the rest of our SDK continues to work normally, but if we have the word 'SDK' then we may imply that this is also the minimum requirements for Logs, Traces, etc.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
```diff
@@ -30,21 +30,26 @@ object SessionReplay {
...
+                imagePrivacy = sessionReplayConfiguration.imagePrivacy,
+                touchPrivacy = sessionReplayConfiguration.touchPrivacy,
+                textAndInputPrivacy = sessionReplayConfiguration.textAndInputPrivacy,
+                customMappers = sessionReplayConfiguration.customMappers,
+                customOptionSelectorDetectors = sessionReplayConfiguration.customOptionSelectorDetectors,
+                sampleRate = sessionReplayConfiguration.sampleRate,
+                startRecordingImmediately = sessionReplayConfiguration.startRecordingImmediately,
+                dynamicOptimizationEnabled = sessionReplayConfiguration.dynamicOptimizationEnabled
+            )
+            sdkCore.registerFeature(sessionReplayFeature)
```
jonathanmos, line 51: What was the decision in the end regarding the thread that this will be called on?

ambushwork: The decision I take is that we register the feature in main thread in the end no matter where the user call `enable` function.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
ambushwork: updated

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
ambushwork: good idea, fixed

ambushwork:
I applied the @0xnm 's change in my branch, please check again, thanks!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementChecker.kt
```diff
@@ -0,0 +1,79 @@
...
+        if (!(memInfoFile.existsSafe(internalLogger) && memInfoFile.canReadSafe(internalLogger))) {
+            return 0L
+        }
+
+        val memorySizeKb = memInfoFile.readLinesSafe(internalLogger = internalLogger)?.mapNotNull { line ->
+            if (line.startsWith(MEM_TOTAL_REGEX)) {
+                val tokens = line.split(Regex("\\s+"))
+                if (tokens.size > 1) {
+                    try {
+                        tokens[1].toLong() // The memory value is in kB
```
0xnm, line 41: maybe `toLongOrNull` to avoid dealing with the exception? Or we need it in our telemetry?

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/SessionReplayConfigurationForgeryFactory.kt
```diff
@@ -27,7 +28,11 @@ class SessionReplayConfigurationForgeryFactory : ForgeryFactory<SessionReplayCon
             customOptionSelectorDetectors = forge.aList { mock() },
             startRecordingImmediately = forge.aBool(),
             sampleRate = forge.aFloat(min = 0f, max = 100f),
-            dynamicOptimizationEnabled = forge.aBool()
+            dynamicOptimizationEnabled = forge.aBool(),
+            systemRequirementConfiguration = SystemRequirementConfiguration.Builder()
```
0xnm, line 32: maybe create a dedicated factory for the `SystemRequirementConfiguration`? Not sure if it will be useful elsewhere though.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementCheckerTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    @Mock
+    private lateinit var mockCpuDir: File
+
+    @Mock
+    private lateinit var mockInternalLogger: InternalLogger
+
+    @IntForgery(min = 1, max = 10)
+    private var fakeCpuCores: Int = 0
+
+    private lateinit var fakeCpuFileList: ArrayList<File>
```
0xnm, line 49:
```suggestion
    private lateinit var fakeCpuFileList: List<File>
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementCheckerTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    @BeforeEach
+    fun `set up`() {
+        fakeCpuFileList = ArrayList()
+        repeat(fakeCpuCores) { index ->
+            mock<File> {
+                whenever(it.name).doReturn("cpu$index")
+            }.apply {
+                fakeCpuFileList.add(this)
+            }
+        }
```
0xnm, line 60:
```suggestion
        fakeCpuFileList = mockCpuFileList(fakeCpuCores)
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementCheckerTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        // When
+        whenever(mockCpuDir.listFiles(any<FilenameFilter>())).doReturn(fakeCpuFileList.toTypedArray())
+        testedCPURequirementChecker.checkMinimumRequirement()
+
+        // Then
+        assertThat(testedCPURequirementChecker.checkedValue()).isEqualTo(fakeCpuCores)
+    }
+
+    private fun mockCpuFileList(number: Int): List<File> {
+        val fakeCpuFileList = ArrayList<File>()
```
0xnm, line 114:
```suggestion
        val fakeCpuFileList = mutableListOf<File>()
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementCheckerTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        repeat(number) { index ->
+            mock<File> {
+                whenever(it.name).doReturn("cpu$index")
+            }.apply {
+                fakeCpuFileList.add(this)
+            }
```
0xnm, line 120:
```suggestion
            fakeCpuFileList += mock<File> {
                whenever(it.name).doReturn("cpu$index")
            }
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementCheckerTest.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        // Then
+        assertThat(testedMemoryRequirementChecker.checkedValue()).isEqualTo((fakeMemorySizeInMb).toLong())
+    }
+
+    private fun generateFakeMemoryFileContent(memorySizeInMb: Int): String {
+        return mapOf<String, Any>(
+            "MemTotal" to (memorySizeInMb.toLong() * 1000)
+        ).map { (key, value) -> "$key:\t$value" }
+            .joinToString("\n")
```
0xnm, line 119: I'm a bit confused here: it seems that map has only one key-value pair, so the whole method can be written as `return "MemTotal:\t${memorySizeInMb.toLong() * 1000}"` if I'm not wrong.

Or `return "MemTotal:\t${memorySizeInMb.toLong() * 1000}\n"` if newline at the EOF is necessary.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/SystemRequirementConfigurationTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    fun `M have correct default value W build()`() {
+        // Given
+        val builder = SystemRequirementConfiguration.Builder()
+
+        // When
+        val result = builder.build()
+
+        // Then
+        assertThat(result.minCPUCores).isEqualTo(0)
+        assertThat(result.minRAMSizeMb).isEqualTo(0)
```
0xnm, line 46: very minor suggestion, doesn't change anything anyway: 

```suggestion
        assertThat(result.minCPUCores).isZero()
        assertThat(result.minRAMSizeMb).isZero()
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/utils/config/MainLooperTestConfiguration.kt
0xnm, line 1: do we use it anywhere?

features/dd-sdk-android-session-replay/api/apiSurface
```diff
@@ -24,11 +24,20 @@ data class com.datadog.android.sessionreplay.SessionReplayConfiguration
...
   - MASK
   - MASK_USER_INPUT
+class com.datadog.android.sessionreplay.SystemRequirementConfiguration
+  class Builder
+    fun setMinCPUCoreNumber(Int): Builder
+    fun setMinRAM(Int): Builder
+    fun build(): SystemRequirementConfiguration
+  companion object 
+    val DEFAULT: SystemRequirementConfiguration
+    val NONE: SystemRequirementConfiguration
```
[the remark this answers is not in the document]
ambushwork: Yes, I think we need to provide an option to the client that session replay can be enabled in any case without checking any system requirements.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -44,6 +45,7 @@ data class SessionReplayConfiguration internal constructor(
         private var textAndInputPrivacy = TextAndInputPrivacy.MASK_ALL
         private var extensionSupport: ExtensionSupport = NoOpExtensionSupport()
         private var dynamicOptimizationEnabled = true
+        private var systemRequirementConfiguration = SystemRequirementConfiguration.NONE
```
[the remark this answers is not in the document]
ambushwork: Well, from my point of view of product design,  I assume that at the beginning, most of user will not be aware of the existence of this API, in this case should we add a default limiter to disable the session replay? with this concern I use NONE as the default configuration, for the client who is aware of this API, they can use DEFAULT one to setup manually.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,155 @@
...
+                    uiHandler.post {
+                        runnable()
+                    }
+                } else {
+                    internalLogger.log(
+                        level = InternalLogger.Level.INFO,
+                        listOf(InternalLogger.Target.TELEMETRY, InternalLogger.Target.USER),
+                        messageBuilder = {
+                            "Session replay is disabled because the system doesn't meet the minimum " +
+                                "Session Replay requirements"
```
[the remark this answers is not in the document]
ambushwork: Yes, this log actually already targets to telemetry, and the concrete requirements are already logged in the `addtionalProperties`, please check ` getCheckerReport(checkers)` function.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,155 @@
...
+     */
+    class Builder {
+        private var minCPUCoreNumber: Int = 0
+        private var minRAMSize: Int = 0
+
+        /**
+         * Sets the minimum CPU core number requirement.
+         *
+         * @param cpuCoreNumber The minimum CPU core number.
+         * @return The current Builder instance for chaining.
```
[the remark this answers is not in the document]
ambushwork: removed

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,155 @@
...
+            this.minCPUCoreNumber = cpuCoreNumber
+            return this
+        }
+
+        /**
+         * Sets the minimum RAM in megabytes.
+         *
+         * @param minRAMSize The minimum RAM in megabytes.
+         * @return The current Builder instance for chaining.
+         */
```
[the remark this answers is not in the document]
ambushwork: added.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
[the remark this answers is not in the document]
0xnm: But is it useful to have it in the telemetry? We may have a lot of entries potentially.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementChecker.kt
```diff
@@ -0,0 +1,49 @@
...
+
+package com.datadog.android.sessionreplay.internal.prerequisite
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.internal.persistence.file.listFilesSafe
+import java.io.File
+
+internal class CPURequirementChecker(
+    private val minCPUCores: Int,
+    private val cpuDirFile: File = File(DIR_PATH),
```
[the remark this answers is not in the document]
ambushwork: I don't think so, the construction doesn't check the file existance, read or wirte.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementChecker.kt
ambushwork: yes, if it can not be converted to Long, we probably want to know what happens in our telemetry.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/SessionReplayConfigurationForgeryFactory.kt
ambushwork: Since it has only two attributes, a dedicated factory may be a bit overkill?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementChecker.kt
0xnm: but what can be the action from our side in this case? not sure if this file can have a different format given that it is fixed https://man7.org/linux/man-pages/man5/proc_meminfo.5.html

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/SessionReplayConfigurationForgeryFactory.kt
0xnm: It is rather a question if we need to create it elsewhere in the tests (now or in the future)

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementCheckerTest.kt
ambushwork: Indeed, I intended to add some irrelevant lines to simulate the real file, updated.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/utils/config/MainLooperTestConfiguration.kt
ambushwork: deleted

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementCheckerTest.kt
0xnm: you can still do :)

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/forge/SessionReplayConfigurationForgeryFactory.kt
ambushwork: I don't use it elsewhere for the moment, and probably not in the future since the whole feature is developed and covered by tests

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementChecker.kt
ambushwork: I intended to have the information about what exception that we can have when parsing the Long value, but probably we can't have any action even if we detect some errors, so let's remove the telemetry and change it to `toLongOrNull`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementChecker.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    private var checkedValue: Int? = null
+
+    override fun checkMinimumRequirement(): Boolean {
+        return readCPUCoreNumber() >= minCPUCores
+    }
+
+    override fun name(): String = CPU_CHECK_NAME
+
+    override fun checkedValue(): Any? = checkedValue
```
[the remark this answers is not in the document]
ambushwork: opinionated: I think we don't need to require the implementation to declare explicitly its generic type just because of this function, here we just need an optional value, if it returns just null we are OK with that.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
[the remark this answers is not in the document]
ambushwork: I am not sure we can have an action upon this telemetry, but maybe it can give us a visibility how frequently this system requirements work and what's the devices are mostly filtered out?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementChecker.kt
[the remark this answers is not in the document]
0xnm: For the attributes you are sending - yes, it doesn't matter, because value type there is `Any?`, so no additional cast is needed.

But for the tests you may simplify life a bit by doing there. Also it gives a clear indicator what is the type of the metric recorded.

Anyway, not blocking to me.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementCheckerTest.kt
ambushwork: some irrelevant memory infos are added here

features/dd-sdk-android-session-replay/api/apiSurface
0xnm, line 1: this file shouldn't be deleted

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,145 @@
...
+                            "Session Replay requirements"
+                    },
+                    onlyOnce = true,
+                    additionalProperties = getCheckerReport(checkers)
+                )
+            }
+        }
+    }
+
+    private fun checkAllMinimumRequirements(
```
0xnm, line 56: tiny nit: at this point probably it makes no sense to have a dedicated method

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,145 @@
...
+        return checkers.all {
+            it.checkMinimumRequirement()
+        }
+    }
+
+    private fun getCheckerReport(checkers: List<SystemRequirementChecker>): Map<String, Any?> {
+        val map = mutableMapOf<String, Any?>()
+        checkers.forEach { checker ->
+            map[checker.name()] = checker.checkedValue()
+        }
```
0xnm, line 68: to be deleted?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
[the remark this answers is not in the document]
0xnm: my point is that default is not called `DEFAULT`, but `NONE`, but I see that it is for the backward compatibility

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,145 @@
...
+            this.minCPUCoreNumber = cpuCoreNumber
+            return this
+        }
+
+        /**
+         * Sets the minimum requirement of total RAM of the device in megabytes.
+         *
+         * @param minRAMSize The minimum RAM in megabytes.
+         */
+        fun setMinRAM(minRAMSize: Int): Builder {
```
0xnm, line 98: should it be `setMinRAMSize`? Anyway, these names should be aligned with iOS

0xnm (commented):
I think this one mostly lgtm, but I would like to have another pair of eyes having a fresh look here, especially on the public API changes.

@xgouchet or @mariusc83 can you please take a look?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
ambushwork: removed

features/dd-sdk-android-session-replay/api/apiSurface
ambushwork: restored

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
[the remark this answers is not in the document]
ambushwork: You are right, but I wonder what name we should give to them, let's check the opinion of the other reviewers.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementChecker.kt
```diff
@@ -0,0 +1,69 @@
...
+package com.datadog.android.sessionreplay.internal.prerequisite
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.internal.persistence.file.canReadSafe
+import com.datadog.android.core.internal.persistence.file.existsSafe
+import com.datadog.android.core.internal.persistence.file.readLinesSafe
+import java.io.File
+
+internal class MemoryRequirementChecker(
+    private val minRamSize: Int,
```
jonathanmos, line 16:
```suggestion
    private val minRamSizeMb: Int,
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Builder class for configuring and creating instances of the [SystemRequirementConfiguration] needed to
+     * enable the Session Replay feature.
+     *
+     */
+    class Builder {
+        private var minCPUCoreNumber: Int = 0
+        private var minRAMSize: Int = 0
```
jonathanmos, line 73:
```suggestion
        private var minRAMSizeMb: Int = 0
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,137 @@
...
+            this.minCPUCoreNumber = cpuCoreNumber
+            return this
+        }
+
+        /**
+         * Sets the minimum requirement of total RAM of the device in megabytes.
+         *
+         * @param minRAMSize The minimum RAM in megabytes.
+         */
+        fun setMinRAMSize(minRAMSize: Int): Builder {
```
jonathanmos, line 90:
```suggestion
        fun setMinRAMSize(minRAMSizeMb: Int): Builder {
```

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
```diff
@@ -0,0 +1,137 @@
...
+        runnable: () -> Unit
+    ) {
+        val checkers = listOf(
+            CPURequirementChecker(minCPUCores, internalLogger = internalLogger),
+            MemoryRequirementChecker(minRAMSizeMb, internalLogger = internalLogger)
+        )
+        (sdkCore as InternalSdkCore).getPersistenceExecutorService().submitSafe(OPERATION_NAME, internalLogger) {
+            val checkResult = checkers.all {
+                it.checkMinimumRequirement()
+            }
```
xgouchet, line 39: > [!WARNING]
> So I have a bit of a concern here. Right now it's not a problem, but it could become one eventually. 
> We create a new Checker instance and perform the checks every time we call the `runIfMeetRequirements()` method. The thing is that the cpu and ram size will never change on a given device, meaning we could memoize the result of the check for the current session and ensure we don't perform the check on every call. Not necessary for this PR but we should probably add a task for this.

xgouchet (dismissed):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -160,6 +162,16 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Defines the minimum system requirements for enabling the Session Replay feature.
+         * When [SessionReplay.enable] is invoked, the system configuration is verified against these requirements.
+         * If the system meets the specified criteria, Session Replay will be successfully enabled.
```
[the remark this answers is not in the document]
jonathanmos: @xgouchet Should our approach not be to optimize performance by default (including not running by default on weak devices), but give users a way to opt-out of our optimizations? If so, the default precheck would be applied for all users with the caveat that they can turn it off with [the flag](https://github.com/DataDog/dd-sdk-android/pull/2259) that we've provided in the configuration.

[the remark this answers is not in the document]
ambushwork: @jonathanmos as we discussed with @MaelNamNam  before, this precheck should not be covered by the flag of the dynamic optimisation, they should be different API to enable/disable. I think we can keep it disable by default, I will add doc to elaborate

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SystemRequirementConfiguration.kt
ambushwork: I see the point here, so the way of memorising it for me is to persist result in shared preference or a file, but I may wonder if this solution will have the same effect with the current one because current checkers actually just read files. So the difference of two solutions is just the file that they are reading if I am not wrong?

xgouchet (approved):

features/dd-sdk-android-session-replay/api/apiSurface
```diff
@@ -24,11 +24,20 @@ data class com.datadog.android.sessionreplay.SessionReplayConfiguration
     fun startRecordingImmediately(Boolean): Builder
     fun setTextAndInputPrivacy(TextAndInputPrivacy): Builder
     fun setDynamicOptimizationEnabled(Boolean): Builder
+    fun setSystemRequirements(SystemRequirementConfiguration): Builder
```
0xnm, line 27: minor: maybe rename class as well `SystemRequirementConfiguration` -> `SystemRequirementsConfiguration`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
```diff
@@ -30,21 +32,37 @@ object SessionReplay {
...
+    }
 
-        sdkCore.registerFeature(sessionReplayFeature)
+    internal fun enable(
+        sessionReplayConfiguration: SessionReplayConfiguration,
+        uiHandler: Handler,
+        sdkCore: SdkCore = Datadog.getInstance()
+    ) {
+        val featureSdkCore = sdkCore as FeatureSdkCore
+        sessionReplayConfiguration.systemRequirementConfiguration.let {
```
0xnm, line 44:
```suggestion
        sessionReplayConfiguration.systemRequirementsConfiguration.let {
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementChecker.kt
```diff
@@ -0,0 +1,49 @@
...
+internal class CPURequirementChecker(
+    private val minCPUCores: Int,
+    private val cpuDirFile: File = File(DIR_PATH),
+    private val internalLogger: InternalLogger
+) : SystemRequirementChecker {
+
+    private var checkedValue: Int? = null
+
+    override fun checkMinimumRequirement(): Boolean {
+        return readCPUCoreNumber() >= minCPUCores
```
0xnm, line 22: we can optimize a bit this (and for the memory checker as well) - if `minCpuCores` is 0 (same for the RAM size), it makes no sense to do the check and we simply can return `true`. Saves some time by avoiding redundant syscalls.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementChecker.kt
```diff
@@ -0,0 +1,69 @@
...
+            0L
+        } else {
+            memorySizeKb / KB_IN_MB
+        }
+        checkedValue = value
+        return value
+    }
+
+    companion object {
+        private const val KB_IN_MB = 1000
```
0xnm, line 59: maybe 1024? Also it will be an integer division, but I guess rounding here is fine for (1 MB precision).

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementCheckerTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        fakeCpuFileList = mockCpuFileList(fakeCpuCores)
+        testedCPURequirementChecker = CPURequirementChecker(
+            minCPUCores = fakeCpuCores,
+            cpuDirFile = mockCpuDir,
+            internalLogger = mockInternalLogger
+        )
+    }
+
+    @Test
+    fun `M have correct name W call name()`() {
```
0xnm, line 62:
```suggestion
    fun `M have correct name W name()`() {
```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementCheckerTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        assertThat(testedCPURequirementChecker.name()).isEqualTo("cpu")
+    }
+
+    @Test
+    fun `M return true W check cpu core number meet configuration`() {
```
0xnm, line 67:
```suggestion
    fun `M return true W checkMinimumRequirement  { cpu core number meets configuration }`() {
```

same suggestion can be applied to the other test names as well

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/CPURequirementCheckerTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        // When
+        whenever(mockCpuDir.listFiles(any<FilenameFilter>())).doReturn(fakeCpuFileList.toTypedArray())
+        testedCPURequirementChecker.checkMinimumRequirement()
+
+        // Then
+        assertThat(testedCPURequirementChecker.checkedValue()).isEqualTo(fakeCpuCores)
+    }
+
+    private fun mockCpuFileList(number: Int): List<File> {
+        val fakeCpuFileList = ArrayList<File>()
```
0xnm, line 106: nit:

```suggestion
        val fakeCpuFileList = mutableListOf<File>()
```

0xnm (approved):
Lgtm. I left a few suggestions, but nothing blocking.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/prerequisite/MemoryRequirementChecker.kt
ambushwork: Here I try to align with `MemoryVitalReader.BYTES_IN_KB` which is also 1000

0xnm (approved):

PR #2270, opened 2024-09-19
RUM-6235 Handle sse request

### LINKED ISSUE #2266: Okhttp with DatadogInterceptor is blocking SSE streams
### Describe the bug

Hi 

The app I'm working on has a DatadogInterceptor and it's working normally until I implemented the Server-Sent Events.
Even though the streams are coming in between certain interval, app receives all the streams at once.

But when I remove the `DatadogInterceptor`, the app receives the streams normally.

Need your help to fix it.

### Reproduction steps

Add datadog interceptor to okhttpclient:
```
List<String> tracedHosts = List.of(BuildConfig.BFF_BASE_URL);
okHttpClientBuilder.addInterceptor(new DatadogInterceptor.Builder(tracedHosts).build())
```
okHttpClientBuilder is passed to create httpclient engine.

Create an SSE endpoint that sends data
```
    sse(OrderStatusResource.PATH) {

        val sseFlow = flow {
            statuses.forEach {
                println(it)
                emit(SseEvent(data = it))
                delay(5.seconds)
            }
        }.shareIn(GlobalScope, SharingStarted.WhileSubscribed(5_000))

        sseFlow.collect {
            val json = Json { encodeDefaults = true }
            val data = json.encodeToString(it)
            send(ServerSentEvent(data = data))
        }
    }
```

now consume the data on the client
```
        httpClient.sse(host = "192.168.251.226", port = 8080, path = OrderStatusResource.PATH) {
            incoming.collect { event ->
                println("SSE::>> $event")
                _flow.emit(OrderStatusDto(data = event.data.toString()))
                if (event.data.equals("Done")) {
                    httpClient.close()
                }
            }
        }
```

you will get all the payload at once when the connection closes.

but when i comment the `DatadogInterceptor`, everything works normally.

### Logcat logs

_No response_

### Expected behavior

App should receive SSE events normally

### Affected SDK versions

datadog-android 2.13.1

n/a

Yes

### Kotlin / Java version

kotlin 2.0.0

### Gradle / AGP version

agp 8.6.0-rc1

com.squareup.okhttp3 = "4.12.0"
io.ktor:ktor-client-okhttp = "3.0.0-rc-1"

### Device Information

_No response_

xgouchet:
### What does this PR do?

This PR Handles properly SSE (Server Sent Event) network request instrumented with our OkHttp interceptor. 
In order to report the size of a downloaded resource, we use the `peekBody()` method which tries to read the whole content of the network request. In case of a SSE request, it would wait until the end of the stream to eventually complete the interception and forward the content to the receiver, blocking the server messages. 

ServerSent events should have the `text/server-event` Mime Type, which we can use to detect such use case. 

### Motivation

Fixes #2266

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -89,7 +89,8 @@ timber = "5.0.1"
 coroutines = "1.4.2"
 
 # Local Server
-ktor = "1.6.0"
+ktor = "1.6.8"
+ktorServer = "3.0.0-rc-1"
 
 # Otel
 jctools = "3.3.0"
@@ -233,8 +234,10 @@ coroutinesCore = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", ver
 
 # Local Server
 ktorCore = { module = "io.ktor:ktor", version.ref = "ktor" }
-ktorNetty = { module = "io.ktor:ktor-server-netty", version.ref = "ktor" }
 ktorGson = { module = "io.ktor:ktor-gson", version.ref = "ktor" }
+ktorServerCore = { module = "io.ktor:ktor-server-core", version.ref = "ktorServer" }
+ktorServerNetty = { module = "io.ktor:ktor-server-netty", version.ref = "ktorServer" }
+ktorServerSSE = { module = "io.ktor:ktor-server-sse", version.ref = "ktorServer" }
 
 # Otel
 jctools = { module = "org.jctools:jctools-core", version.ref = "jctools" }
@@ -306,8 +309,12 @@ glide = [
 
 ktor = [
     "ktorCore",
-    "ktorNetty",
-    "ktorGson"
+    "ktorGson",
+]
+ktorServer = [
+    "ktorServerCore",
+    "ktorServerNetty",
+    "ktorServerSSE"
 ]
 
 traceCore = [
--- a/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
+++ b/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
@@ -357,13 +357,19 @@ internal constructor(
 
     private fun getBodyLength(response: Response, internalLogger: InternalLogger): Long? {
         return try {
-            val body = response.body ?: return null
+            val body = response.body
+            val contentType = body?.contentType()?.let {
+                // manually rebuild the mimetype has `toString()` can also include the charsets
+                it.type + "/" + it.subtype
+            }
+            if (body == null || contentType in STREAM_CONTENT_TYPES) {
+                return null
+            }
             // if there is a Content-Length available, we can read it directly
             // however, OkHttp will drop Content-Length header if transparent compression is
             // used (since the value reported cannot be applied to decompressed body), so to be
             // able to still read it, we force decompression by calling peekBody
-            body.contentLengthOrNull() ?: response.peekBody(MAX_BODY_PEEK)
-                .contentLengthOrNull()
+            body.contentLengthOrNull() ?: response.peekBody(MAX_BODY_PEEK).contentLengthOrNull()
         } catch (e: IOException) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
@@ -480,6 +486,10 @@ internal constructor(
     // endregion
 
     internal companion object {
+
+        internal val STREAM_CONTENT_TYPES = arrayOf(
+            "text/event-stream"
+        )
 
         internal const val WARN_RUM_DISABLED =
             "You set up a DatadogInterceptor for %s, but RUM features are disabled. " +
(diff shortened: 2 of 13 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -357,13 +357,19 @@ internal constructor(
 
     private fun getBodyLength(response: Response, internalLogger: InternalLogger): Long? {
         return try {
-            val body = response.body ?: return null
+            val body = response.body
+            val contentType = body?.contentType()?.let {
+                // manually rebuild the mimetype has `toString()` can also include the charsets
```
0xnm, line 362:
```suggestion
                // manually rebuild the mimetype, `toString()` can also include the charsets
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -481,6 +487,10 @@ internal constructor(
 
     internal companion object {
 
+        internal val STREAM_CONTENT_TYPES = arrayOf(
+            "text/event-stream"
```
0xnm, line 491: should we maybe try to include other stream types here? like for websocket, gRPC

sample/kotlin/build.gradle.kts
```diff
@@ -195,6 +195,7 @@ dependencies {
     implementation("io.opentracing.contrib:opentracing-rxjava-3:0.1.4") {
         exclude(group = "io.opentracing")
     }
+    implementation("com.launchdarkly:okhttp-eventsource:2.5.0")
```
0xnm, line 198: it is already included in the `// Network` section below

sample/kotlin/src/main/res/values/strings.xml
```diff
@@ -60,6 +60,7 @@
     <string name="button_async_operation">Async Operation</string>
     <string name="button_network_request">Network Request</string>
     <string name="button_start">Start</string>
+    <string name="button_start_sse">Start SSE Request</string>
```
0xnm, line 63: minor: maybe use full form? `Start Server-Side Event Request`

0xnm (approved):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
xgouchet: 👍

sample/kotlin/build.gradle.kts
xgouchet: 👍

sample/kotlin/src/main/res/values/strings.xml
xgouchet: I could but it'll make the button multiline on most device 😅

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -481,6 +487,10 @@ internal constructor(
 
     internal companion object {
 
+        internal val STREAM_CONTENT_TYPES = arrayOf(
```
0xnm, line 490: should it be `setOf`?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
xgouchet: Good idea

0xnm (approved):
there is one comment about `arrayOf` -> `setOf` (performance-wise on such small collection it won't matter though), otherwise lgtm

gradle/libs.versions.toml
```diff
@@ -89,7 +89,8 @@ timber = "5.0.1"
 coroutines = "1.4.2"
 
 # Local Server
-ktor = "1.6.0"
+ktor = "1.6.8"
+ktorServer = "3.0.0-rc-1"
```
mariusc83, line 93: is there a concern if we are using a `rc` here ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
xgouchet: 👍

gradle/libs.versions.toml
xgouchet: This lib is only used for the sample app, so it's ok

mariusc83 (approved):

PR #2272, opened 2024-09-19
RUN integration tests on API 35 in testing pyramid

mariusc83:
### What does this PR do?

Updates the latest API version for running our integration tests to API 35.

### Motivation

### Additional Notes


0xnm (approved):
lgtm!

as a part of API 35 support, we also need to modify target SDK, build tools version, etc. in the Gradle config and Docker file as well.

xgouchet (approved):

xgouchet (approved):

0xnm (approved):

ambushwork (approved):

ambushwork (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
```diff
@@ -8,12 +8,13 @@ package com.datadog.android.sessionreplay.internal.recorder.resources
...
 import android.graphics.Bitmap
 import com.datadog.android.sessionreplay.internal.utils.InvocationUtils
+import com.datadog.android.sessionreplay.internal.utils.safeConfig
 
 internal class BitmapPoolHelper(
     private val invocationUtils: InvocationUtils = InvocationUtils()
 ) {
     internal fun generateKey(bitmap: Bitmap) =
-        generateKey(bitmap.width, bitmap.height, bitmap.config)
+        generateKey(bitmap.width, bitmap.height, bitmap.safeConfig())
```
0xnm, line 17: So how it was working in the past? we now know, that annotation wasn't correct and `Bitmap.getConfig` may return `null`. It should crash runtime, because `generateKey` doesn't expect `null` value for the config parameter.

Also do we need `safeConfig` method? Maybe in some code paths we can handle `null` returned case without fallback value.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/CacheUtils.kt
```diff
@@ -12,6 +12,10 @@ import androidx.collection.LruCache
 internal class CacheUtils<K : Any, V : Any>(
     private val invocationUtils: InvocationUtils = InvocationUtils()
 ) {
+
+    // some of this memory level are not being triggered after API 34. We still need to keep them for now
+    // for lower versions
+    @Suppress("DEPRECATION")
```
0xnm, line 18: what is the replacement then? should we use it for newer APIs?

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
mariusc83: yes...in case it will return null it will crash. I've never seen this bug reported yet so I suspect this happens quite rarely. I am not sure that this will require a hotfix. @xgouchet @jonathanmos ?

mariusc83: the reason why I introduce the `safeConfig` is just to be re - used also in some tests where I had to replace it and maybe for later...looks cleaner this way.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/CacheUtils.kt
mariusc83: I think we can leave it like it is and prepare a task to remove those later. The only think that will happen for newer APIs is that those level will never be triggered. I don't see any harm now, maybe to create a task to clean these levels ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
jonathanmos: The getConfig method is annotated NonNull for some reason. I haven't seen any issues with this returning null, but if it can return it then we do indeed need a safe method.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/datastore/DataStoreFileReaderTest.kt
```diff
@@ -142,7 +142,7 @@ internal class DataStoreFileReaderTest {
     @Test
     fun `M log error W read() { invalid number of blocks }`() {
         // Given
-        blocksReturned.removeLast()
+        blocksReturned.removeAt(blocksReturned.size - 1)
```
xgouchet, line 145:
```suggestion
        blocksReturned.removeAt(blocksReturned.lastIndex)
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/datastore/DataStoreFileReaderTest.kt
```diff
@@ -213,7 +213,7 @@ internal class DataStoreFileReaderTest {
     @Test
     fun `M return onFailure W read() { invalid number of blocks }`() {
         // Given
-        blocksReturned.removeLast()
+        blocksReturned.removeAt(blocksReturned.size - 1)
```
xgouchet, line 216:
```suggestion
        blocksReturned.removeAt(blocksReturned.lastIndex)
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
0xnm: @jonathanmos It was annotated as `NonNull` by mistake, see https://cs.android.com/android/_/android/platform/frameworks/base/+/a4c2f9badb21180282596ac9730d4777f65e4b4b

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/CacheUtils.kt
0xnm: yes, no harm. My point is just that if something is deprecated, then probably there is a replacement for the newer APIs?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
mariusc83: The question was more if we can leave it like this until the next release where the safe method will be in place. You said you did not see any issue so far so I guess this is very rare when it returns null. We could leave it like this without a hotfix.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/CacheUtils.kt
mariusc83: Nope. no replacement suggested in their docs, I think they simply don't support these levels anymore and only kept few important ones. I think they did not need this granularity anymore.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
xgouchet: > [!WARNING]
> I'm not at all convinced we need a `safeCongif` here as we lose the information. Any local log or telemetry that could be sent from the generated key will hide the fact that the bitmap configuration was `null`. 
> In the current case, I'd prefer to have a `generateKey` allowing a nullable config instead

xgouchet (changes requested):

jonathanmos (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
mariusc83: I didn't see any telemetry there, in any case the Config in this case is just a variable in the key generation so with our without it `null` will have more or less same changes of collisions for 2 elements with `Bitmap.config` null so I am ok with any approach. @jonathanmos , @0xnm what do you guys think ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolHelper.kt
```diff
@@ -15,10 +15,15 @@ internal class BitmapPoolHelper(
...
         generateKey(bitmap.width, bitmap.height, bitmap.config)
 
-    internal fun generateKey(width: Int, height: Int, config: Bitmap.Config) =
-        "$width-$height-$config"
+    internal fun generateKey(width: Int, height: Int, config: Bitmap.Config?): String {
+        return if (config != null) {
+            "$width-$height-$config"
+        } else {
+            "$width-$height-null"
+        }
```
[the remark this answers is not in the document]
mariusc83: yes just a weird autocomplete

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/BitmapPoolTest.kt
```diff
@@ -148,13 +148,14 @@ internal class BitmapPoolTest {
...
-        testedCache.getBitmapByProperties(mockBitmap.width, mockBitmap.height, mockBitmap.config)
+        testedCache.getBitmapByProperties(mockBitmap.width, mockBitmap.height, mockConfig)
 
         // When
         testedCache.put(mockBitmap)
 
         // Then
-        val actualBitmap = testedCache.getBitmapByProperties(mockBitmap.width, mockBitmap.height, mockBitmap.config)
+        val actualBitmap =
+            testedCache.getBitmapByProperties(mockBitmap.width, mockBitmap.height, mockConfig)
```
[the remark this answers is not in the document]
mariusc83: `mockBitmap.config` returns nullable now and that method expects nonnull so we are using directly the `mockConfig` with which this `mockBitmap` is mocked with.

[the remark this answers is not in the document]
xgouchet: Why can't we make the `getBitmapByProperties()` method accept a nullable `Config`? The only usage of the config is to call the `generateKey` which this PR modifies to allow a nullable config.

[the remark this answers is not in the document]
mariusc83: But why we should change this method if this code if this is not required ? All the methods calling this code are keeping the contract, if we would change this code we would change only for the sake of this test, this is not the same thing as the `BitmapPoolHelper#generateKey` method which accepts a `bitmap` and delegates to `generateKey` method. This method has a different scope and requires in the contract that the `config` should be NonNull when calling it. I really don't get why do you want to change this.

xgouchet (approved):

xgouchet (approved):

PR #2273, opened 2024-09-20
RUM-6256: Optimize MD5 byte array to hex string conversion

ambushwork:
### What does this PR do?

This PR changes the way to convert MD5 bytes to HEX string.

### Motivation

When we broke down the steps of generating image wireframe, we noticed that `joinToString` function takes a huge amount of time when generating the MD5 resource id:

Screenshot 2024-09-20 at 11 48 11

#### Suspicious reasons: 

* String Concatenation Overhead:
joinToString internally uses StringBuilder, but string concatenation inside the lambda (using `"%02x".format(it)`) could be inefficient. If format() is invoked for each byte, it involves repeated function calls, which adds to overhead.

* Repeated Memory Allocations:
`"%02x".format(it)` creates a new string for each byte, which leads to multiple temporary objects being created. This puts pressure on the garbage collector.

### Outcome

After using the new function to convert the byte arrays, we can see the duration optimisation rate is up to **99%**

#### Span list
| joinToString |  convertToHexString |
| --- | ---- |
| image | image|

#### Time series

Screenshot 2024-09-20 at 11 46 55

### DIFF
--- a/dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/ByteArrayExtTest.kt
+++ b/dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/ByteArrayExtTest.kt
@@ -1,0 +1,66 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.internal.utils
+
+import com.datadog.android.internal.utils.toHexString
+import fr.xgouchet.elmyr.annotation.StringForgery
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+import java.util.Locale
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+class ByteArrayExtTest {
+
+    @Test
+    fun `M return correct hex string W convert {0x00, 0x00}`() {
+        // Given
+        val byteArray = byteArrayOf(0x00, 0x00) // 0xA is 10 in decimal
+
+        // When
+        val result = byteArray.toHexString()
+
+        // Then
+        val expectedHex = "0000"
+        assertThat(result).isEqualTo(expectedHex)
+    }
+
+    @Test
+    fun `M return correct hex string W convert {0xFF, 0xFF}`() {
+        // Given
+        val byteArray = byteArrayOf(0xFF.toByte(), 0xFF.toByte()) // 0xA is 10 in decimal
+
+        // When
+        val result = byteArray.toHexString()
+
+        // Then
+        val expectedHex = "ffff"
+        assertThat(result).isEqualTo(expectedHex)
+    }
+
+    @Test
+    fun `M return correct hex string W call toHexString()`(@StringForgery fakeInput: String) {
+        // Given
+        val fakeByteArray = fakeInput.toByteArray()
+
+        // When
+        val result = fakeByteArray.toHexString()
+
+        // Then
+        val expected = fakeByteArray.joinToString(separator = "") { "%02x".format(Locale.US, it) }
+        assertThat(result).isEqualTo(expected)
+    }
+}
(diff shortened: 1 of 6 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/MD5HashGenerator.kt
```diff
@@ -33,7 +32,21 @@ internal class MD5HashGenerator(
...
+    private fun convertToHexString(byteArray: ByteArray): String {
+        @Suppress("UnsafeThirdPartyFunctionCall") // byte array size is always positive.
+        val result = StringBuilder(byteArray.size * 2)
+        for (byte in byteArray) {
+            val intVal = byte.toInt() and BYTE_MASK
+            result.append(HEX_CHARS[intVal ushr HEX_SHIFT]) // Append first half of byte
+            result.append(HEX_CHARS[intVal and LOWER_NIBBLE_MASK]) // Append second half of byte
+        }
+        return result.toString()
+    }
```
0xnm, line 44: maybe add a test? the output should match then with "%02x" format output

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/MD5HashGenerator.kt
ambushwork: this private function is already covered by `MD5HashGeneratorTest` before, but to make it reusable in the future, I move this to a ext file in `dd-sdk-android-internal` module, and created a dedicated test for it.

dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/ByteArrayExtTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    @Test
+    fun `M return correct hex string W call toHexString()`(@StringForgery fakeInput: String) {
+        val fakeByteArray = fakeInput.toByteArray()
+
+        val result = fakeByteArray.toHexString()
+
+        val expected = fakeByteArray.joinToString(separator = "") { "%02x".format(Locale.US, it) }
+
+        assertThat(result).isEqualTo(expected)
```
0xnm, line 36:
```suggestion
        // Given
        val fakeByteArray = fakeInput.toByteArray()

        // When
        val result = fakeByteArray.toHexString()

        // Then
        val expected = fakeByteArray.joinToString(separator = "") { "%02x".format(Locale.US, it) }
        assertThat(result).isEqualTo(expected)
```

0xnm (approved):

dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/ByteArrayExtTest.kt
```diff
@@ -0,0 +1,40 @@
...
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+class ByteArrayExtTest {
+
+    @Test
+    fun `M return correct hex string W call toHexString()`(@StringForgery fakeInput: String) {
```
mariusc83, line 29: can we also add a 2 separated tests with the extremes ? [0x00,0x00] and [0xff, 0xff] ?

mariusc83 (approved):
Nice job overall just a small comment regarding the tests.

dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/ByteArrayExtTest.kt
ambushwork: Good point, added

dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/ByteArrayExtTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+import java.util.Locale
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+class ByteArrayExtTest {
```
xgouchet, line 26: > [!NOTE]
> Since the goal of this is to optimize the performance of converting a bytearray to string, shouldn't we add a performance test ? 
> - Make a large byte array
> - convert it to hex string using this new function, and joinToString
> - compare the performance of both methods

ambushwork: Interesting point, However I am not sure the performance test should be part of the unit test, since unit test is supposed to verify **correctness**. Also I am afraid of performance test can be flaky, it depends also on the running environment, the test can fail even though we know in general `toHexString` is better than `joinToString`.

So do you have any suggestion how we should proceed this performance test?

ambushwork: I will merge it since I got enough approval, don't hesitate to add anything here, we can do it in another PR!

PR #2274, opened 2024-09-23
Fix DatadogInterceptor flaky test

ambushwork:
### What does this PR do?

Fix DatadogInterceptor flaky test


0xnm (changes requested):
nice catch! but the same should be applied to the test above as well.

xgouchet (approved):

ambushwork:
> nice catch! but the same should be applied to the test above as well.

@0xnm if I understand correctly, the test above specifically set the mimeType a stream content type, which will always be parsed as `Native` type, which will not have a flaky result.

0xnm (approved):

PR #2279, opened 2024-09-24
RUM-6314: Use layout text to display textview overflow correctly

ambushwork:
### What does this PR do?

Fix the Session Replay issue that text view overflow could not be displayed properly

### Motivation

RUM-6314

### Demo

| App screen | Before | After |
| --- | --- | --- |
|image |image|image|

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/TextViewMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/TextViewMapper.kt
@@ -27,6 +27,7 @@ import com.datadog.android.sessionreplay.utils.ViewIdentifierResolver
 /**
  * A [WireframeMapper] implementation to map a [TextView] component.
  */
+@Suppress("TooManyFunctions")
 open class TextViewMapper<in T : TextView>(
     viewIdentifierResolver: ViewIdentifierResolver,
     colorStringFormatter: ColorStringFormatter,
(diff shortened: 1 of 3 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/TextViewMapper.kt
```diff
@@ -90,7 +90,7 @@ open class TextViewMapper<in T : TextView>(
         textAndInputPrivacy: TextAndInputPrivacy,
         isOption: Boolean
     ): String {
-        val originalText = textView.text?.toString().orEmpty()
+        val originalText = textView.layout?.text?.toString().orEmpty()
```
0xnm, line 93: should we fallback just to `textView.text` if `textView.layout` is `null` or it is not necessary?

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/TextViewMapper.kt
0xnm: It can be `null` according to Javadoc: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/widget/TextView.java;l=2629?q=TextView

ambushwork: Nice catch, so it makes sense to fallback to the full text

PR #2284, opened 2024-09-25
RUM-6039 Fix StrictMode warning regarding I/O disk operation on main thread

mariusc83:
…thread

### What does this PR do?

### Motivation

### Additional Notes


features/dd-sdk-android-ndk/src/main/kotlin/com/datadog/android/ndk/internal/NdkCrashReportsFeature.kt
```diff
@@ -51,7 +52,7 @@ internal class NdkCrashReportsFeature(
             NDK_CRASH_REPORTS_FOLDER
         )
         try {
-            allowThreadDiskReads {
+            allowThreadDiskWrites {
```
mariusc83, line 55: I wonder why do we need to create the dirs on the main thread, maybe the owner of this code can give more info while reviewing this code. Tks !!.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/FileOrchestrator.kt
```diff
@@ -61,4 +61,9 @@ internal interface FileOrchestrator {
      */
     @WorkerThread
     fun getMetadataFile(file: File): File?
+
+    /**
+     * @return the name of the root directory of this orchestrator or null if the root directory does not exist.
+     */
+    fun getRooDirName(): String?
```
0xnm, line 68:
```suggestion
    fun getRootDirName(): String?
```

detekt_custom.yml
```diff
@@ -384,8 +384,9 @@ datadog:
       - "android.os.Process.getStartElapsedRealtime()"
       - "android.os.Process.myPid()"
       - "android.os.StrictMode.allowThreadDiskReads()"
+      - "android.os.StrictMode.allowThreadDiskWrites()"
       - "android.os.StrictMode.setThreadPolicy(android.os.StrictMode.ThreadPolicy?)"
-      - "android.os.SystemClock.elapsedRealtime()"
+      - "android.os.SystemClock.elapsedRealtime()"R
```
0xnm, line 389:
```suggestion
      - "android.os.SystemClock.elapsedRealtime()"
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
```diff
@@ -63,7 +63,7 @@ internal class ConsentAwareStorage(
             callerClass = ConsentAwareStorage::class.java.name,
             metric = TelemetryMetricType.MethodCalled,
             samplingRate = MethodCallSamplingRate.RARE.rate,
-            operationName = "writeCurrentBatch[${orchestrator?.getRootDir()?.nameWithoutExtension}]"
+            operationName = "writeCurrentBatch[${orchestrator?.getRooDirName()}]"
```
0xnm, line 66: Side note: I'm not even sure what is the point of having directory name here. With what we currently have it will be essentially (value of `operationName`):`writeCurrentBatchrum-pending-v2` or `writeCurrentBatchrum-v2`.

Assuming that tracking consent information is probably useless, we are interested mostly in the feature name, which is a part of `MetricsDispatcher` already, so can be removed from here and be an additional attribute of all the events sent. But this performance metric is not using `MetricsDispatcher`, but `InternalLogger`, which is confusing.

It would be much more sense to have operation name = `writeCurrentBatch` and feature name attached as an attribute.

0xnm (approved):

features/dd-sdk-android-ndk/src/main/kotlin/com/datadog/android/ndk/internal/NdkCrashReportsFeature.kt
mariusc83: yes and no, I didn't wrote that function `allowThreadDiskRead`, I think @xgouchet did ;)

mariusc83: Check line 54 and 56 ;)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
mariusc83: That's the thing I asked on the original thread in Slack but no - one answered so I said to go with this and discuss it here maybe.

mariusc83: maybe we could just use feature name here....it will remove a lot of boiler code, what do you all think ? The only thing is that we will miss the folder type ? Granted/Pending. For that we can also pass the tracking consent in the operationName.

0xnm: I would avoid growing constructor much. Also what is the point of having tracking consent here?

0xnm: Maybe add enrich `MetricsDispatcher`? Implementation already has everything needed: logger and feature name,

features/dd-sdk-android-ndk/src/main/kotlin/com/datadog/android/ndk/internal/NdkCrashReportsFeature.kt
mariusc83: yes totally right there, I am still trying to think of a reason why we need to keep it there, do you see any ?

mariusc83: yes that's the reason why I am afraid as when the signal is handle you are in a different state it is not the same as in the JVM crash state. You have so little time to do something then the process is killed so you need to be very efficient there, I think for this main reason I prefer in this case to have that folder created already.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/ConsentAwareStorage.kt
mariusc83: Waiting for more opinions here before getting into it but I think it is the right call instead of using the root dir name.

mariusc83: the only thing is to not affect the metric which is based on the `operationName` in the dashboards

0xnm: It is our internal metric, so we can break it at any time.

PR #2285, opened 2024-09-25
RUM-6324: Add View Groups Session Replay demo screen in sample app

ambushwork:
### What does this PR do?

When testing SR with some other applications, we noticed that the card view is not correctly display in the player, before fixing the issue, we need to add an showcase screen for this kind of view groups.

### Motivation

RUM-6324

### Additional Notes

| Sample App| Player |
| --- | --- |
| Screenshot_20240925_135133|image|

### DIFF
--- a/sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/ViewGroupComponentsFragment.kt
+++ b/sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/ViewGroupComponentsFragment.kt
@@ -1,0 +1,25 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sample.sessionreplay
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.Fragment
+import com.datadog.android.sample.R
+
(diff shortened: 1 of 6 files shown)


sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/ViewGroupComponentsFragment.kt
```diff
@@ -0,0 +1,25 @@
...
+internal class ViewGroupComponentsFragment : Fragment() {
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        return inflater.inflate(R.layout.fragment_view_group_components, container, false)
+    }
+}
```
0xnm, line 25: minor: can be pretty much simplified

```suggestion
internal class ViewGroupComponentsFragment : Fragment(R.layout.fragment_view_group_components)
```

0xnm (approved):

xgouchet (approved):

sample/kotlin/src/main/kotlin/com/datadog/android/sample/sessionreplay/ViewGroupComponentsFragment.kt
ambushwork: Should we keep the `onCreateView` function because probably in the future we need to setup some views programatically in this function?

0xnm: sure, you can keep it in this case, it is not a blocking comment anyway

PR #2291, opened 2024-09-26
RUM-6218: Add privacy override for hidden views

jonathanmos:
### What does this PR do?
Sets up the common mechanism for privacy overrides and implements overriding view privacy to `hidden` in the hierarchy. 

Views will be able to set hidden override privacy using the api:
```[view].setSessionReplayHidden(hide = true/false)```
where true will register the view to be replaced with a placeholder with the label "Hidden", and false will remove the hidden override from the view. The children of hidden views will not be traversed.

### Motivation
This is part of the third stage of Fine Grained Masking. The first stage was the start/stop api, and the second was global fine grained masking levels for the app.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/api/apiSurface
+++ b/features/dd-sdk-android-session-replay/api/apiSurface
@@ -9,6 +9,7 @@ data class com.datadog.android.sessionreplay.MapperTypeWrapper<T: android.view.V
   constructor(Class<T>, com.datadog.android.sessionreplay.recorder.mapper.WireframeMapper<T>)
   fun supportsView(android.view.View): Boolean
   fun getUnsafeMapper(): com.datadog.android.sessionreplay.recorder.mapper.WireframeMapper<android.view.View>
+fun android.view.View.setSessionReplayHidden(Boolean)
 object com.datadog.android.sessionreplay.SessionReplay
   fun enable(SessionReplayConfiguration, com.datadog.android.api.SdkCore = Datadog.getInstance())
   fun startRecording(com.datadog.android.api.SdkCore = Datadog.getInstance())
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
@@ -1,0 +1,24 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay
+
+import android.view.View
+
+/**
+ * Allows setting a view to be "hidden" in the hierarchy in Session Replay.
+ * When hidden the view will be replaced with a placeholder in the replay and
+ * no attempt will be made to record it's children.
+ *
+ * @param hide pass `true` to hide the view, or `false` to remove the override
+ */
+fun View.setSessionReplayHidden(hide: Boolean) {
+    if (hide) {
+        this.setTag(R.id.datadog_hidden, true)
+    } else {
+        this.setTag(R.id.datadog_hidden, null)
+    }
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
@@ -13,7 +13,9 @@ import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.feature.measureMethodCallPerf
 import com.datadog.android.core.metrics.MethodCallSamplingRate
 import com.datadog.android.sessionreplay.MapperTypeWrapper
+import com.datadog.android.sessionreplay.R
 import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueRefs
+import com.datadog.android.sessionreplay.internal.recorder.mapper.HiddenViewMapper
 import com.datadog.android.sessionreplay.internal.recorder.mapper.QueueStatusCallback
 import com.datadog.android.sessionreplay.model.MobileSegment
 import com.datadog.android.sessionreplay.recorder.MappingContext
@@ -25,6 +27,7 @@ import com.datadog.android.sessionreplay.utils.NoOpAsyncJobStatusCallback
 internal class TreeViewTraversal(
     private val mappers: List<MapperTypeWrapper<*>>,
     private val defaultViewMapper: WireframeMapper<View>,
+    private val hiddenViewMapper: HiddenViewMapper,
     private val decorViewMapper: WireframeMapper<View>,
     private val viewUtilsInternal: ViewUtilsInternal,
     private val internalLogger: InternalLogger
@@ -51,7 +54,11 @@ internal class TreeViewTraversal(
         // try to resolve from the exhaustive type mappers
         var mapper = findMapperForView(view)
 
-        if (mapper != null) {
+        if (isHidden(view)) {
+            traversalStrategy = TraversalStrategy.STOP_AND_RETURN_NODE
+            mapper = hiddenViewMapper
+            jobStatusCallback = noOpCallback
+        } else if (mapper != null) {
             jobStatusCallback = QueueStatusCallback(recordedDataQueueRefs)
             traversalStrategy = if (mapper is TraverseAllChildrenMapper) {
                 TraversalStrategy.TRAVERSE_ALL_CHILDREN
@@ -104,6 +111,9 @@ internal class TreeViewTraversal(
     private fun findMapperForView(view: View): WireframeMapper<View>? {
         return mappers.firstOrNull { it.supportsView(view) }?.getUnsafeMapper()
     }
+
+    private fun isHidden(view: View): Boolean =
+        view.getTag(R.id.datadog_hidden) == true
 
     data class TraversedTreeView(
         val mappedWireframes: List<MobileSegment.Wireframe>,
--- a/features/dd-sdk-android-session-replay/src/main/res/values/ids.xml
+++ b/features/dd-sdk-android-session-replay/src/main/res/values/ids.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+  ~ Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+  ~ This product includes software developed at Datadog (https://www.datadoghq.com/).
+  ~ Copyright 2016-Present Datadog, Inc.
+  -->
+
+<resources>
+    <item name="datadog_hidden" type="id"/>
+</resources>
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/HiddenViewMapperTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/HiddenViewMapperTest.kt
@@ -1,0 +1,116 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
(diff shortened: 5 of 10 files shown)

### CODE REVIEW

File: features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt [hunk 1]
```diff
@@ -51,37 +54,43 @@ internal class TreeViewTraversal(
...
-            jobStatusCallback = noOpCallback
-        } else {
+        if (PrivacyOverrideManager.isHidden(view)) {
             traversalStrategy = TraversalStrategy.STOP_AND_RETURN_NODE
-            mapper = defaultViewMapper
+            mapper = hiddenViewMapper
             jobStatusCallback = noOpCallback
-            val viewType = view.javaClass.canonicalName ?: view.javaClass.name
+        } else {
+            if (mapper != null) {
```
xgouchet, line 62: > [!NOTE]
> You should chain the following ifs for readability instead of nesting them.

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.datadog.android.sessionreplay.internal.PrivacyOverrideManager
+
+/**
+ * Allows setting a view to be "hidden" in the hierarchy in Session Replay.
+ * When hidden the view will be replaced with a placeholder in the replay and
+ * no attempt will be made to record it's children.
+ *
+ * @param hide pass `true` to hide the view, or `false` to remove the override
+ */
+fun View.ddSessionReplayOverrideHidden(hide: Boolean) {
```
ambushwork, line 19: I think in the RFC, the API should be able to pass the sdk core inside? 
```
view.ddSessionReplayOverrideHidden(sdkInstance, hide = true)
```
It can avoid to make `PrivacyOverrideManager` as an singleton object, and will be easier to make the test I think

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideManagerTest.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.mockito.kotlin.mock
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(ForgeConfigurator::class)
+internal class PrivacyOverrideManagerTest {
```
ambushwork, line 33: Since you are using WeakReference in `PrivacyOverrideManager`, should we have some unit tests to simulate the garbage collection to ensure we have the expected behaviour? 

For example `W gc happens M "isHidden()" result is reset` ..

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
jonathanmos: Yes, the source of this rfc change was a discussion that I had today with xav. I'm in the process of making changes to the code to support multiple instances

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideManagerTest.kt
jonathanmos: Indeed - I'll add some tests for this

features/dd-sdk-android-session-replay/api/apiSurface
```diff
@@ -1,14 +1,16 @@
...
+enum com.datadog.android.sessionreplay.ImagePrivacy : PrivacyLevel
   - MASK_NONE
   - MASK_LARGE_ONLY
   - MASK_ALL
 data class com.datadog.android.sessionreplay.MapperTypeWrapper<T: android.view.View>
   constructor(Class<T>, com.datadog.android.sessionreplay.recorder.mapper.WireframeMapper<T>)
   fun supportsView(android.view.View): Boolean
   fun getUnsafeMapper(): com.datadog.android.sessionreplay.recorder.mapper.WireframeMapper<android.view.View>
+interface com.datadog.android.sessionreplay.PrivacyLevel
+fun android.view.View.ddSessionReplayOverrideHidden(Boolean)
```
0xnm, line 13: Are we using `dd` instead of `datadog` in the naming on purpose?

Also if we mention `override` in the name, it is not clear what value we overriding. Is it to make what is hidden visible, or the opposite? Maybe just `datadogSessionReplayHide(Boolean)` or something similar?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/PrivacyOverride.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import com.datadog.android.sessionreplay.ImagePrivacy
+import com.datadog.android.sessionreplay.TextAndInputPrivacy
+import com.datadog.android.sessionreplay.TouchPrivacy
+
+internal data class PrivacyOverride(
+    val imagePrivacy: ImagePrivacy? = null,
+    val touchPrivacy: TouchPrivacy? = null,
+    val textAndInputPrivacy: TextAndInputPrivacy? = null,
+    val hiddenPrivacy: Boolean = false
```
0xnm, line 17: maybe just `hidden`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/PrivacyOverrideManager.kt
```diff
@@ -0,0 +1,107 @@
...
+        val key = WeakViewKey(view)
+        return overridesMap[key]
+    }
+
+    internal fun isHidden(view: View): Boolean {
+        val key = WeakViewKey(view)
+        return overridesMap[key]?.hiddenPrivacy == true
+    }
+
+    private class WeakViewKey(view: View) : WeakReference<View>(view) {
```
0xnm, line 97: blocking:

Not quite the best solution, because your View will still be in the memory even after the screen is destroyed until GC kicks in. Is there are other way to keep view identity? I think in the Session Replay we are relying solely on the hash code, which is enough, because once existing view is destroyed, the new one created will still have `ddSessionReplayOverride....` call made on it if needed.

features/dd-sdk-android-session-replay/api/apiSurface
jonathanmos: dd was to make the extension a bit shorter, but I can change this. 'Override' is to make it clear that this is one of the privacy overrides, and group with the others that will be in a future pr - image override and text and input override.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/PrivacyOverrideManager.kt
```diff
@@ -0,0 +1,107 @@
...
+import android.view.View
+import com.datadog.android.sessionreplay.ImagePrivacy
+import com.datadog.android.sessionreplay.PrivacyLevel
+import com.datadog.android.sessionreplay.TextAndInputPrivacy
+import com.datadog.android.sessionreplay.TouchPrivacy
+import java.lang.ref.WeakReference
+import java.util.concurrent.ConcurrentHashMap
+
+internal object PrivacyOverrideManager {
+    private val overridesMap = ConcurrentHashMap<WeakViewKey, PrivacyOverride>()
```
mariusc83, line 18: Why not using directly a `WeakHashMap` ? The key is already `weak` in there: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/WeakHashMap.html

features/dd-sdk-android-session-replay/api/apiSurface
0xnm: >'Override' is to make it clear that this is one of the privacy overrides

Why do we need to make it clear? I was thinking that this API is just for the manual control, simply for the user to decide to hide it or not.

0xnm: > dd was to make the extension a bit shorter, but I can change this

Yes, I would rather propose to use `datadog` here. `dd` is not used anywhere in the public API lingvo (apart of the property name for the internal needs of the event models).

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Allows setting a view to be "hidden" in the hierarchy in Session Replay.
+ * When hidden the view will be replaced with a placeholder in the replay and
+ * no attempt will be made to record it's children.
+ *
+ * @param hide pass `true` to hide the view, or `false` to remove the override
+ */
+fun View.datadogSessionReplayHidden(hide: Boolean) {
+    if (hide) {
```
jonathanmos, line 19: should this be a synchronized block?

xgouchet: > [!NOTE]
> Looking at the underlying implementation of the View tags, it's not concurrent safe, especially in the case of setting multiple tags. 
> IMO, I don't feel that it's necessary though, as we're setting a boolean. From the SDK we can only force synchronisation between the write and read, and I feel like the write will happen only once (I can't think of scenarios where we'd have to hide/reveal the same view many times), so it's good enough.

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
0xnm: Normally, the common rule of working with Views is to call any methods/properties from the main thread (because some methods may throw exception if call wasn't done from the main thread).

However, while customer will probably follow this rule and call `setTag`, `getTag` on our side may happen on the worker thread I suppose?

jonathanmos: We are calling get only on the main thread, but customer could still theoretically call set from other threads. IAC it seems synchronization will not be necessary here

features/dd-sdk-android-session-replay/src/main/res/values/ids.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+  ~ Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+  ~ This product includes software developed at Datadog (https://www.datadoghq.com/).
+  ~ Copyright 2016-Present Datadog, Inc.
+  -->
+
+<resources>
+    <item name="datadog_hidden" type="id"/>
+</resources>
```
0xnm, line 9:
```suggestion
</resources>

```

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/HiddenViewMapperTest.kt
```diff
@@ -0,0 +1,117 @@
...
+            view = mockView,
+            asyncJobStatusCallback = mockAsyncJobStatusCallback,
+            internalLogger = mockInternalLogger,
+            mappingContext = mockMappingContext
+        )
+
+        val wireframe = wireframesList[0] as? MobileSegment.Wireframe.PlaceholderWireframe
+
+        // Then
+        checkNotNull(wireframe)
```
0xnm, line 96: we probably want to fail as early as possible if something goes wrong, and with this logic if `wireframesList[0]` is not `MobileSegment.Wireframe.PlaceholderWireframe` then we will get an error about the value being `null`, which is misleading.

maybe do something like this?

```suggestion

        // Then
        val wireframe = wireframesList[0]
        check(wireframe is MobileSegment.Wireframe.PlaceholderWireframe)
```

0xnm (approved):

mariusc83 (approved):

ambushwork (approved):
LGTM, well done!

PR #2293, opened 2024-09-27
Use SDK source value in Session Replay `MobileSegment.source` property

0xnm:
### What does this PR do?

This is to align with what iOS SDK has https://github.com/DataDog/dd-sdk-ios/blob/4dbdf4a10991ba65a02b02bf1fe87e4aa2765b38/DatadogSessionReplay/Sources/Feature/RequestBuilders/SegmentRequestBuilder.swift#L48

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/MobileSegmentExt.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/MobileSegmentExt.kt
@@ -6,6 +6,7 @@
 
 package com.datadog.android.sessionreplay.internal.processor
 
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.model.MobileSegment
 
 internal fun MobileSegment.Wireframe.copy(clip: MobileSegment.WireframeClip?): MobileSegment.Wireframe {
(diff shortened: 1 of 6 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/MobileSegmentExt.kt
```diff
@@ -17,3 +18,23 @@ internal fun MobileSegment.Wireframe.copy(clip: MobileSegment.WireframeClip?): M
...
+internal fun MobileSegment.Source.Companion.tryFromSource(
+    source: String,
+    internalLogger: InternalLogger
+): MobileSegment.Source {
+    return try {
+        fromJson(source)
+    } catch (e: NoSuchElementException) {
+        internalLogger.log(
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.USER,
```
ambushwork, line 31: I am wondering if use can fill the property `DatadogContext.source` himself during the development? if not why we should log it to user?

ambushwork (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/MobileSegmentExt.kt
0xnm: This is just the same approach we are using to convert `source: String` to `source: Enum` in RUM, with logging error to the user. I would agree that it probably makes no sense for the user. In the past it could be fail because of the Proguard, but now we encode string value in the enum directly, so it won't impact), so probably a leftover from the past. 

I will update to the `DEV` target.

xgouchet (approved):

PR #2298, opened 2024-09-30
RUM-3889 Send retry information into rum data upload requests

mariusc83:
### What does this PR do?

Following the internal RFC specifications we are adding new information for the RUM upload requests (retry count, previous retry failure code and request SHA1 idempotency key) in order to help understanding the performances of or our retry mechanism.

### Motivation

### Additional Notes


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
```diff
@@ -73,40 +86,90 @@ internal class RumRequestFactory(
...
             elements.add("${RumAttributes.VARIANT}:$variant")
         }
+        if (executionContext.previousResponseCode != null) {
+            elements.add("${LAST_FAILURE_STATUS_KEY}:${executionContext.previousResponseCode}")
+        }
 
         return elements.joinToString(",")
     }
 
+    @Suppress("TooGenericExceptionCaught")
```
ambushwork, line 130: is this `@Supress` necessary? it seems you catch all the possible exceptions

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
```diff
@@ -73,40 +86,90 @@ internal class RumRequestFactory(
...
+        val headers = mutableMapOf(
+            RequestFactory.HEADER_API_KEY to context.clientToken,
+            RequestFactory.HEADER_EVP_ORIGIN to context.source,
+            RequestFactory.HEADER_EVP_ORIGIN_VERSION to context.sdkVersion,
             RequestFactory.HEADER_REQUEST_ID to requestId
         )
+        if (idempotencyKey != null) {
+            headers[RequestFactory.DD_IDEMPOTENCY_KEY] = idempotencyKey
+        }
+        return headers
```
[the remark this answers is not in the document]
mariusc83: yup I also had this in mind but I was thinking that might be a bit overkill...I could actually do it differently without having 2 create 2 maps. Let me change this.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
mariusc83: good questions...it was before I added this in the `detekt.xml` let me see it now.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestExecutionContext.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package com.datadog.android.api.net
+
+/**
+ * Provides information about the current request execution context.
+ * @param attempts the number of attempts made to execute the request in case of a retry.
+ * @param previousResponseCode the response code of the previous request failure code in case of a retry.
+ */
+data class RequestExecutionContext(
```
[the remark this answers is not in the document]
mariusc83: nope, this is actually exposed in our `RequestFactory` public API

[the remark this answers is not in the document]
ambushwork: Yes I see it, but when I check all the implementations of `RequestFactory` exhaustively,  they are all internal classes, so my question is that will user create their own implementation of `RequestFactory`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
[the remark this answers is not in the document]
mariusc83: Actually the reason why I did this was that I return a `immutable map`. If you call `toMap` under the hood it also creates an `LinkedHashMap` so still mutable. So I think I have an uglier way but more optimised.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
mariusc83: yes it is still complaining and due to the `NullPointerException` catch at the end of the block. I am not sure why though as this is specifically thrown by the code under the hood.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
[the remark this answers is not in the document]
mariusc83: So basically I will leave it like this because under the hood after a quick research `toMap` returns like I mentioned above a `LinkedHashMap`. Actually I would have expected to return an `ImmutableMap` but apparently the immutability is only at the type (surface level). So in a Kotlin code you could easily do this:
```
  val map = mapOf(...)
  val immutableMap = map.toMap()
  val hashMap = immutableMap as LinkedHashMap
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
ambushwork: It's a default set of detekt to include NPE in this check, because for Kotlin code, NPE is not very informative, but since we are calling a Java function, there is nothing we can do about, so it's correct to have `@Suppress` here I think.

https://detekt.dev/docs/rules/exceptions/#toogenericexceptioncaught

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestExecutionContext.kt
```diff
@@ -0,0 +1,23 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.api.net
+
+/**
+ * Provides information about the current request execution context.
+ * @param attempts the number of attempts made to execute the request in case of a retry.
+ * @param previousResponseCode the response code of the previous request failure code in case of a retry.
```
0xnm, line 12: Seems like this sentence should be corrected. It should add at least the comma and state, that this is a code of the previous attempt, otherwise one may think that it can be the code of the previous batch upload.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestExecutionContext.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+data class RequestExecutionContext(
+    /**
+     * The number of attempts made to execute the request in case of a retry.
+     */
+    val attempts: Int = 0,
+    /**
+     * The response code of the previous request failure code in case of a retry.
+     */
```
0xnm, line 21: why duplicate the description there if it is already in the Kdoc above? These ones won't be treated as documentation and are ordinary comments.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestFactory.kt
```diff
@@ -17,12 +17,14 @@ fun interface RequestFactory {
     /**
      * Creates a request for the given batch.
      * @param context Datadog SDK context.
+     * @param executionContext Gives information about the execution context of this and previous requests.
```
0xnm, line 20:
```suggestion
     * @param executionContext Information about the execution context of this and previous requests.
```
Also to avoid confusion we must say that it is for the current batch only (otherwise it is not clear if it is for the previous batch as well).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestFactory.kt
```diff
@@ -67,5 +69,10 @@ fun interface RequestFactory {
          * Datadog tags query parameter name.
          */
         const val QUERY_PARAM_TAGS: String = "ddtags"
+
+        /**
+         * Datadog Idempotency key header, used for deduplication purposes.
```
0xnm, line 74: will it be used for any deduplication actually? I was under the impression that this is mostly for collecting the statistics right now

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
```diff
@@ -125,10 +136,20 @@ internal sealed class UploadStatus(
                         "and you're targeting the relevant Datadog site."
                 )
             }
+            if (this@UploadStatus.shouldRetry) {
+                append(
```
0xnm, line 140: should add a space in the beginning in case if there is already some text, otherwise it will be like `...targeting the relevant Datadog site.Attempt count was...`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
```diff
@@ -73,40 +86,90 @@ internal class RumRequestFactory(
...
 
         return elements.joinToString(",")
     }
 
+    @Suppress("TooGenericExceptionCaught")
+    private fun idempotencyKey(byteArray: ByteArray): String? {
+        try {
+            val digest = MessageDigest.getInstance("SHA-1")
+            val hashBytes = digest.digest(byteArray)
+            return hashBytes.toHexString()
```
0xnm, line 135: So each feature has to provide the key on its own? I'm not sure if it is the solution we would like to have. Is it possible to lift it to the core? Or at least have something like `ByteArray.sha1()` declared in the core. How iOS SDK is doing that?

mariusc83: I went to the core way but then I've seen that in iOS implementation they did it at the feature level and changed it. I think it makes sense at the feature level as `Logs` and `Traces` will not need this. Otherwise in the core you will have to make sure you do it only for the right features (if/else) and not even sure if we have the feature information in that place to check.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestFactory.kt
mariusc83: Well yes maybe not for now, not sure should we change this definition ?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestExecutionContext.kt
mariusc83: I am going to try to be more explicit here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
0xnm: They have a function to generate the key in the core https://github.com/DataDog/dd-sdk-ios/blob/dec751b901e17e74bc7f96af56558882cacd9282/DatadogInternal/Sources/Extensions/Data%2BCrypto.swift#L11 and it is applied on the each feature level. We should follow the same approach.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestFactory.kt
0xnm: I'm not even sure it will ever will. And anyway, if it will be, it will be used only for RUM events.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt
mariusc83: The only problem I have with this is that for now we are doing this only for RUM and I am afraid it will stay like that. I have no problem with moving this into core but that's the reason I preferred not to do it. Should we do it now or later in case we really need to share it between features ?

0xnm: I mean for now we can just declare extension method to generate SHA1 in the core and call it from the feature-specific factory. We already have the header declared in the core, isn't it?

mariusc83: I get the part with the function to have the extension in the core but don't get the header part ? The header is added here so you only need `ByteArray.generateSha1Fingerprint()` function if we want to go this way. I am still thinking it is overkilled as no other code will ever use this function but will let it here for others to say their opinion. @ambushwork @xgouchet @jonathanmos ?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
```diff
@@ -35,10 +36,16 @@ internal class DataOkHttpUploader(
     override fun upload(
         context: DatadogContext,
         batch: List<RawBatchEvent>,
-        batchMeta: ByteArray?
+        batchMeta: ByteArray?,
+        previousUploadStatus: UploadStatus?
```
xgouchet, line 40: > [!WARNING]
> It feels like we're making a whole lot of change in unrelated classes to simply get the number of attempts for a given batch. 
> If in the end we just keep the info in a `previousUploadStatus` property, why don't we put that info in the `DataOkHttpUploader` class directly, instead of the `DataUploadRunnable`? 

> [!IMPORTANT]
> Also there's a key thing here that we don't track is which batch was the previous upload status related to. Technically, we could make 13 attempts with batch A. Then batch A is deleted for being too old, and we make one successful attempt with batch B, which would report 13 retries… Is this the behavior we want to track ?

xgouchet (changes requested):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
mariusc83: yup here is a corner case and having followed the exact implementation in iOS I think they have the same corner case. I think the best way in this case is to count the attempt count and `previousFailureCode` based on the requestId in the OkHttpUploader. I am going to open the discussion in the daily today to make sure everyone is aware.

mariusc83: We also can leave it here and use the `batchID` to match it.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
```diff
@@ -7,28 +7,39 @@
...
 
 import com.datadog.android.api.InternalLogger
+import java.util.Locale
 
 internal sealed class UploadStatus(
     val shouldRetry: Boolean = false,
     val code: Int = UNKNOWN_RESPONSE_CODE,
-    val throwable: Throwable? = null
+    val throwable: Throwable? = null,
+    val attempt: Int = 0
```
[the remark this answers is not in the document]
mariusc83: While I agree this could be done in different ways the initial plan for me was to align to `iOS` implementation where this was added to the `UploadStatus` information, I also kept their naming -> `attempt`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
mariusc83: We have 2 options here, either keep the information in the `UploadStatus` and handle the `previousUpload` status by its `batchID` or move everything into the `DataOkHttpUploader` but I would rather not do it there as we do not have any information about the batch id and we will have to compute the fingerprint from the `batchData` or pass the `batchID` as a extra parameter to that method.

ganeshnj: actually no - scratch above, we don't keep in memory. the deletion of batch for other than successful upload can be a problem here.

ganeshnj: I would suggest save the batch identifier in the upload status just like retry information and only propagate the information when the identifier is same - else reset it.

I don't see any reason to keep this per batch, only for the active batch is sufficient.

mariusc83: I agree with this solution, this will mean that we can keep same implementation and just pass the `batchID` in the previous `UploadStatus` to match it here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
```diff
@@ -125,10 +132,23 @@ internal sealed class UploadStatus(
                         "and you're targeting the relevant Datadog site."
                 )
             }
+            if (code != null && attempt != null) {
+                if (isNotEmpty()) {
+                    append(" ")
+                }
```
xgouchet, line 138: > [!NOTE]
> The builder will never be empty at this point, the space can be included in the `RETRY_REQUEST_LOG_MESSAGE_FORMAT`

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
```diff
@@ -125,10 +132,23 @@ internal sealed class UploadStatus(
                         "and you're targeting the relevant Datadog site."
                 )
             }
+            if (code != null && attempt != null) {
```
xgouchet, line 135: > [!NOTE]
> What if there was several attempts but the last status doesn't have a status code (e.g.: network failure)? I think only the attempt should be checked. Besides, the beginning of the message already gives information about the status and code, so it feels redundand to add it here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/net/RequestExecutionContext.kt
```diff
@@ -0,0 +1,18 @@
...
+package com.datadog.android.api.net
+
+/**
+ * Provides information about the request execution context such as the number of attempts made to
+ * execute the request in case of a retry or the response code of the previous request failure code.
+ * @param attempts the number of attempts made to execute this particular request in case of a retry.
+ * @param previousResponseCode the response code of the previous request failure code in case of a retry.
+ */
+data class RequestExecutionContext(
+    val attempts: Int = 0,
```
xgouchet, line 16: > [!WARNING]
> Is this the number of attempts including the current one being made, or excluding it? It would be better to make this info explicit in the doc and in the name.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/UploadStatusTest.kt
```diff
@@ -99,15 +97,19 @@ internal class UploadStatusTest {
...
         // Then
         mockLogger.verifyLog(
             InternalLogger.Level.WARN,
             listOf(InternalLogger.Target.USER),
             "Batch [$fakeByteSize bytes] ($fakeContext) failed " +
-                "because of a network error (${status.throwable!!.message}); we will retry later."
+                "because of a network error (${status.throwable!!.message}); we will retry later." + " " +
+                UploadStatus.RETRY_REQUEST_LOG_MESSAGE_FORMAT.format(
+                    Locale.US, fakeRequestAttempts, status.code
+                )
```
xgouchet, line 112: > [!WARNING]
> You should not do that as you're using the same algorithm to create the expected output as the production code.

xgouchet (changes requested):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/UploadStatusTest.kt
mariusc83: yup I realized that, I am fixing it now, I am afraid I will have to do this in function of the targe.

xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
```diff
@@ -29,16 +31,22 @@ internal class DataOkHttpUploader(
     val androidInfoProvider: AndroidInfoProvider
 ) : DataUploader {
 
+    private var attempts = 1
+    private var previousUploadStatus: UploadStatus? = null
+    private var previousUploadedBatchId: BatchId? = null
```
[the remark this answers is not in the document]
mariusc83: I particularly looked into this myself also. Initially my intention was to make them thread safe but then I didn't want to add this if it's not really needed. I checked the executors and seen the 1 pool thread size and assumed it is fine. I am thinking that if we really have this scenario that you are mentioning we might end up in different problems as the upload runnable assumes that everything is executed sequentially.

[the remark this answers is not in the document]
0xnm: > if we really have this scenario that you are mentioning we might end up in different problems as the upload runnable assumes that everything is executed sequentially

And it will be executed sequentially (in the boundaries of the particular feature) even if there are multiple threads in the pool, because there is only one single runnable at the time for the particular feature and the next runnable is scheduled **only** at the end of the execution of the previous feature. And underlying storage is thread-safe, it is designed for the concurrent usage.

Here my point is only about volatile nature of these references. Imagine there is an executor running and thread was killed because of some reason; then new thread will be spinned off which may not see the changes made to these references by the previous thread. Or in the case if executor has multiple threads running.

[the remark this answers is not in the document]
mariusc83: That might be a problem indeed. I think marking them as volatile might make sense to make sure they are being kept in the HEAP and not in the Thread local memory. In case of atomicity we don't need it in this case as we are already running sequential. I am going to open a PR.

PR #2304, opened 2024-10-03
Override process importance for Session Replay integration tests

ambushwork:
### What does this PR do?

In integration test, the process is always treated as foreground service instead of foreground.
So when initialising the feature, `DatadogRumMonitor` consider the app is not in foreground,
which causes different test result in Rum feature.

In this case we need to force the importance of the process to have the same result of user environment.

### Motivation

Fix flaky test

### DIFF
--- a/instrumented/integration/src/main/kotlin/com/datadog/android/sdk/integration/sessionreplay/ProcessImportanceExt.kt
+++ b/instrumented/integration/src/main/kotlin/com/datadog/android/sdk/integration/sessionreplay/ProcessImportanceExt.kt
@@ -1,0 +1,26 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sdk.integration.sessionreplay
+
+import android.app.ActivityManager
+import com.datadog.android.rum.DdRumContentProvider
+
+/**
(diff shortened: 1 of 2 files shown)


instrumented/integration/src/main/kotlin/com/datadog/android/sdk/integration/sessionreplay/ProcessImportanceExt.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.datadog.android.rum.DdRumContentProvider
+
+/**
+ * In instrumentation tests, the process is treated as [IMPORTANCE_FOREGROUND_SERVICE],
+ * which differs from the behavior in a typical user environment. To prevent this difference
+ * from affecting test results, this function can be called to override the relevant variable
+ * inside [DdRumContentProvider], allowing the SDK to correctly identify that the application
+ * is in the foreground.
+ */
+fun overrideProcessImportance() {
```
mariusc83, line 19: This is a nice finding. I wonder how many integration tests are affected by this :(. Good that we fixed this.

mariusc83 (dismissed):

instrumented/integration/src/main/kotlin/com/datadog/android/sdk/integration/sessionreplay/ProcessImportanceExt.kt
ambushwork: Actually a lot, it's just in Session Replay test rule we didn't override it, but in `KioskTrackingActivityTestRule` and `GesturesTrackingActivityTestRule` we have already made this override.

0xnm (approved):

PR #2305, opened 2024-10-03
Make the DataOkHttpUploader state volatile

mariusc83:
### What does this PR do?

Initially I thought that given the fact that the `DataOkHttpUploader` is being executed sequentially there is no need for thread safety in it. After a team discussion we stumbled upon this corner case where the executor could actually kill that thread and spawn a new one therefor we need to make sure these states are Volatile and kept in the Heap.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
@@ -31,8 +31,13 @@ internal class DataOkHttpUploader(
     val androidInfoProvider: AndroidInfoProvider
 ) : DataUploader {
 
-    private var attempts = 1
+    @Volatile
+    private var attempts: Int = 1
+
+    @Volatile
     private var previousUploadStatus: UploadStatus? = null
+
+    @Volatile
     private var previousUploadedBatchId: BatchId? = null
 
     // region DataUploader
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploaderTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploaderTest.kt
@@ -667,6 +667,45 @@ internal class DataOkHttpUploaderTest {
     }
 
     @Test
+    fun `M pass the ExecutionContext to requestFactory W upload { same batchId retried, new thread each time }`(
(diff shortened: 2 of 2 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
```diff
@@ -31,8 +31,12 @@ internal class DataOkHttpUploader(
     val androidInfoProvider: AndroidInfoProvider
 ) : DataUploader {
 
-    private var attempts = 1
+    private var attempts: Int = 1
```
0xnm, line 34: why not making this one as `@Volatile` or even `AtomicInteger`. There is no concurrency right now (so probably `AtomicInteger` is not quite useful), but for the changes at least shouldn't we declare volatile access?

mariusc83: I was almost convinced that primitives are volatile in JVM. The docs are saying either true or false so let's do this also Volatile. There is no need for Atomicity there.

0xnm: https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html

> Reads and writes are atomic for reference variables and for most primitive variables (all types except long and double).
Reads and writes are atomic for all variables declared volatile (including long and double variables).

I'm not sure about ART though, but maybe it should maintain this as well.

0xnm (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataOkHttpUploader.kt
0xnm: Also atomic read/write is different from cross-thread visibility though.

> When a field is declared volatile, the compiler and runtime are put on notice that this variable is shared and that operations on it should not be reordered with other memory operations. Volatile variables are not cached in registers or in caches where they are hidden from other processors, so a read of a volatile variable always returns the most recent write by any thread.

mariusc83: Yes there are 2 different things but I am pretty sure I've read also about volatility once, anyway I guess without going into the bytecode you will never now for sure.

PR #2307, opened 2024-10-04
Read SR system requirement synchronously with strict mode allowance

ambushwork:
### What does this PR do?

Asynchronous enabling the session replay feature after checking the system requirement causes the dysfunction of touch recorder.
This commit fixes it by reading the files on the original thread synchronously with allowance of strict mode.

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
@@ -6,8 +6,6 @@
 
 package com.datadog.android.sessionreplay
 
-import android.os.Handler
-import android.os.Looper
 import com.datadog.android.Datadog
 import com.datadog.android.api.SdkCore
 import com.datadog.android.api.feature.Feature.Companion.SESSION_REPLAY_FEATURE_NAME
@@ -21,6 +19,8 @@ object SessionReplay {
 
     /**
      * Enables a SessionReplay feature based on the configuration provided.
+     * It is recommended to invoke this function as early as possible in the app's lifecycle,
+     * ideally within the `Application#onCreate` callback, to ensure proper initialization.
      *
      * @param sessionReplayConfiguration Configuration to use for the feature.
      * @param sdkCore SDK instance to register feature in. If not provided, default SDK instance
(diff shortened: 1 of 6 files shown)


sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
```diff
@@ -217,7 +217,7 @@ class SampleApplication : Application() {
 
     private fun initializeSessionReplay() {
         val shouldUseFgm = SecureRandom().nextInt(100) < USE_FGM_PCT
-        val systemRequirementsConfiguration = SystemRequirementsConfiguration.BASIC
+        val systemRequirementsConfiguration = SystemRequirementsConfiguration.NONE
```
0xnm, line 220: why we are doing this change? shouldn't we keep basic to have a case where files are actually accessed in the sample app?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
```diff
@@ -31,22 +29,10 @@ object SessionReplay {
     fun enable(
```
0xnm, line 29: side question: while we mitigated the issue for the case when SR is initialized in the `Application#onCreate` **and** on the main thread, I believe we can still have a similar issue if SR is initialialized by the customer from the worker thread / later in the app lifecycle. Is there a way to mitigate completely?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
ambushwork: We are usually using sample app to make some test on emulator, and emulator by default uses 1 cpu cores, which will not bypass this `BASIC` check.

0xnm: Then maybe use `setMinCPUCores` API instead of pre-built templates?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
ambushwork: With current solution of `LifecycleCallback`, I don't see a solution, once user misses the `onResume` event, they need to wait until the next `onResume` to make the touch recorder work properly

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
ambushwork: Good idea, updated

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
0xnm: If there is no solution for such case, then maybe we should highlight in the docs indeed that SR should be initialized early.

0xnm (approved):

mariusc83 (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
ambushwork: updated

PR #2308, opened 2024-10-04
Fix SeekBarWireframeMapper flaky test

ambushwork:
### What does this PR do?

The main cause is that `fakeProgress` has random value such as `0.00001`, which loses some precision when convert to Int and convert back to float, so `fakeProgressInt` from 0~100 limit only two decimal .


mariusc83 (approved):

0xnm:
Thanks for the explanation @ambushwork, but I still feel I'm not getting it, especially this part:

> which loses some precision when convert to Int and convert back to float

Can you give a concrete example where it leads to the test failure?

jonathanmos (approved):

ambushwork:
@0xnm yes, so if the test runs under previous code, `fakeProgress` can generate a random value such as `0.0002096` as a percentage from 0f to 1f. 
and let's imagine the `fakeMinValue` is 0 and `fakeMaxValue` is `22631`, so with the mock condition:
```
 whenever(mockView.progress) doReturn ((fakeMaxValue - fakeMinValue) * fakeProgress).toInt() + fakeMinValue
```
`mockView.progress` will return 4. This is the abosulote value from fakeMinValue to fakeMaxValue.

In the function `map` that we want to test, it will normalise the absolute value back to percentage value, by using 
```
view.progress.toFloat()/(maxValue-minValue)
```
which returns `0.0001767`.

so we can see there is already a difference between    `0.0002096` and `0.0001767`. and finally we will use this value to calculate the active track width of the seek bar, which should multiplied by the width of the whole track. Due to this, the difference of the two value is amplified.

0xnm:
@ambushwork I have the feeling that even with this change test may be flaky, just the probability will be much lower, but let's see.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/SeekBarWireframeMapperTest.kt
```diff
@@ -60,7 +59,9 @@ internal class SeekBarWireframeMapperTest : AbstractWireframeMapperTest<SeekBar,
     @IntForgery(min = 512, max = 65536)
     var fakeMaxValue: Int = 0
 
-    @FloatForgery(min = 0f, max = 1f)
+    @IntForgery(min = 0, max = 100)
+    var fakeProgressInt: Int = 0
```
0xnm, line 63: minor: `fakeProgressInt` is not used anywhere, so maybe you can just remove property and `Forgery` annotation and inject `forge` in `set up` method.

0xnm (approved):

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/SeekBarWireframeMapperTest.kt
ambushwork: Yes I can, I just have a question if I do so, does the `forge` instance of `set up` method uses the same seed as the property forgery? if not, it's not convenient to reproduce the failure.

0xnm: yes, it is the same seed

0xnm: it is just `fun setUp(forge: Forge)` and you'll get the necessary instance.

ambushwork: updated

ambushwork:
> @ambushwork I have the feeling that even with this change test may be flaky, just the probability will be much lower, but let's see.

yes, that's correct, with original code the chance of failure is kind like 1 in 2000, with the fix I think it can be reduced about 100 times, which I think it's acceptable.

PR #2309, opened 2024-10-04
RUM-6307 Make sure ConsentAwareFileOrchestrator is thread safe

mariusc83:
### What does this PR do?

We had an issue in iOS where switching from `PENDING` to `GRANTED` in the tracking consent asynchronously causes the loss of some events that are about to be written from a different queue. In this PR we are making sure we are covering this use case with integration tests and also that we mark the `delegateOrchestrator` as `Volatile` in the `ConsentAwareFileOrchestrator` in case the Executor is re - spawning the worker thread.

### Motivation

### Additional Notes

### DIFF
--- a/reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
+++ b/reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
@@ -31,6 +31,8 @@ import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
 
 /**
  * Instrumentation tests for the feature scope.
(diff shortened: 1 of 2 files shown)


reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
```diff
@@ -156,6 +156,49 @@ class FeatureScopeTest : MockServerTest() {
         }.doWait(LONG_WAIT_MS)
     }
 
+    @Test
+    fun mustReceiveTheEvents_whenFeatureWrite_trackingConsentPendingToGranted_switched_asynchronously() {
```
jonathanmos, line 160: minor: I notice that the tests in this class do not follow our standard of M W?

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
```diff
@@ -156,6 +156,49 @@ class FeatureScopeTest : MockServerTest() {
...
+            thread.start()
+            thread.join()
+        }
+        val switchConsentThread = Thread {
+            Datadog.setTrackingConsent(TrackingConsent.GRANTED)
+        }
+        switchConsentThread.start()
+        switchConsentThread.join()
+
+        // When
```
jonathanmos, line 190: empty When

jonathanmos (approved):

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
```diff
@@ -156,6 +156,49 @@ class FeatureScopeTest : MockServerTest() {
...
+                featureScope.withWriteContext { _, eventBatchWriter ->
+                    eventBatchWriter.write(
+                        rawBatchEvent,
+                        fakeBatchMetadata,
+                        eventType
+                    )
+                }
+            }
+            thread.start()
+            thread.join()
```
0xnm, line 182: I'm not quite clear why do we need to spin a new thread for every write request we are making, why not put `forEach` call into a single dedicated thread? By using `.join` inside every `.forEach` call you are anyway making `.withWriteContext` invocations sequential.

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
mariusc83: These are integration tests which ran with Junit4 and therefor we can't use that format in the function description.

ambushwork, line 160: By curiosity I remove `@Volatile` that you've added and run the test on my end, the test still passes, so I am wondering what's the case are you exactly testing against? are you testing the thread-safety when `delegateOrchestrator` is assigned in `ConsentAwareFileOrchestrator`? is there any false positive in this unit test?

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
mariusc83: I wanted to increase the potential of a collision, basically the worse case scenario.

0xnm: But that doesn't increase the chance of collision, isn't it? There is no race condition, because of the `.join` - all write requests are strictly sequential.

ambushwork (approved):

PR #2312, opened 2024-10-08
RUM-6219: Add Image and TextAndInput privacy overrides

jonathanmos:
### What does this PR do?
Adds privacy overrides for images and textAndInput.
Api example:

`view.setSessionReplayImagePrivacy(privacy: ImagePrivacy?)`
`view.setSessionReplayTextAndInputPrivacy(privacy: TextAndInputPrivacy?)`

These privacy overrides work on nearest parent - so if you override the privacy on a viewgroup, it should be propagated to the children unless they have their own override.

### Motivation
Part of the fine grained masking effort. 

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
@@ -22,3 +22,31 @@ fun View.setSessionReplayHidden(hide: Boolean) {
         this.setTag(R.id.datadog_hidden, null)
     }
 }
+
+/**
+ * Allows overriding the image privacy for a view in Session Replay.
+ *
+ * @param privacy the new privacy level to use for the view
+ * or null to remove the override.
+ */
+fun View.setSessionReplayImagePrivacy(privacy: ImagePrivacy?) {
+    if (privacy == null) {
+        this.setTag(R.id.datadog_image_privacy, null)
+    } else {
+        this.setTag(R.id.datadog_image_privacy, privacy.toString())
+    }
+}
+
+/**
+ * Allows overriding the text and input privacy for a view in Session Replay.
+ *
+ * @param privacy the new privacy level to use for the view
+ * or null to remove the override.
+ */
+fun View.setSessionReplayTextAndInputPrivacy(privacy: TextAndInputPrivacy?) {
+    if (privacy == null) {
+        this.setTag(R.id.datadog_text_and_input_privacy, null)
+    } else {
+        this.setTag(R.id.datadog_text_and_input_privacy, privacy.toString())
+    }
+}
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt
@@ -9,7 +9,9 @@ package com.datadog.android.sessionreplay.internal.recorder
 import android.view.View
 import android.view.ViewGroup
 import androidx.annotation.UiThread
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.ImagePrivacy
+import com.datadog.android.sessionreplay.R
 import com.datadog.android.sessionreplay.TextAndInputPrivacy
 import com.datadog.android.sessionreplay.internal.async.RecordedDataQueueRefs
 import com.datadog.android.sessionreplay.model.MobileSegment
@@ -22,7 +24,8 @@ import java.util.LinkedList
 internal class SnapshotProducer(
     private val imageWireframeHelper: ImageWireframeHelper,
     private val treeViewTraversal: TreeViewTraversal,
-    private val optionSelectorDetector: OptionSelectorDetector
+    private val optionSelectorDetector: OptionSelectorDetector,
+    private val internalLogger: InternalLogger
 ) {
 
     @UiThread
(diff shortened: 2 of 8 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
```diff
@@ -22,3 +22,31 @@ fun View.datadogSessionReplayHidden(hide: Boolean) {
...
     }
 }
+
+/**
+ * Allows overriding the image privacy for a view in Session Replay.
+ *
+ * @param privacy the new privacy level to use for the view
+ * or null to remove the override.
+ */
+fun View.datadogSessionReplayImagePrivacy(privacy: ImagePrivacy?) {
```
xgouchet, line 32: > [!NOTE]
> Shouldn't we use a verb for function name, like `setDatadogSessionReplayImagePrivacy` (or just `setSessionReplayImagePrivacy` or `setDatadogImagePrivacy`) ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
```diff
@@ -22,3 +22,31 @@ fun View.datadogSessionReplayHidden(hide: Boolean) {
...
+    }
+}
+
+/**
+ * Allows overriding the text and input privacy for a view in Session Replay.
+ *
+ * @param privacy the new privacy level to use for the view
+ * or null to remove the override.
+ */
+fun View.datadogSessionReplayTextAndInputPrivacy(privacy: TextAndInputPrivacy?) {
```
xgouchet, line 46: > [!NOTE]
> Shouldn't we use a verb for function name, like `setDatadogSessionReplayTextAndInputPrivacy` (or just `setSessionReplayTextAndInputPrivacy` or `setDatadogTextAndInputPrivacy`) ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
ambushwork, line 32: opinionated:
* The API names seems too long to use, I think we can use some abbreviation for "datadog" or "SessionReplay" since user can easily get the context by clicking into the function.
* Although it is declared in doc, It is still not intuitive for me that passing a null value means removing the override, probably a dedicated api to clear the override?
* What is the consequence if user call this API on a `TextView`, should we specify the case in the doc?

features/dd-sdk-android-session-replay/src/main/res/values/ids.xml
```diff
@@ -6,4 +6,6 @@
 
 <resources>
     <item name="datadog_hidden" type="id"/>
+    <item name="datadog_image_privacy" type="id"/>
```
mariusc83, line 9: can be these used later when we're going to introduce the namespace for xml templates ?

mariusc83 (approved):

features/dd-sdk-android-session-replay/src/main/res/values/ids.xml
jonathanmos: I'm not sure that the namespace will be practical. It would seem to be using custom attributes, however custom attributes can [only be read](https://developer.android.com/develop/ui/views/layout/custom-views/create-view) when a view is created and not afterwards. Let's discuss

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/PrivacyOverrideExtensions.kt
jonathanmos: I think that for the naming issues in general we should have a discussion (also on whether to have a separate api for removing the overrides). 

Regarding the textview, there could be compound drawables. Also the overrides are propagated downwards by nearest parent so even for cases where there's ostensibly no effect on the direct view, there may be views further down the tree that it would change the masking for. We can cover this in the docs.

jonathanmos: I've renamed the apis to align with our discussion

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt
```diff
@@ -97,4 +101,50 @@ internal class SnapshotProducer(
...
+        return mappingContext.copy(
+            imagePrivacy = imagePrivacy,
+            textAndInputPrivacy = textAndInputPrivacy
+        )
+    }
+
+    private fun logInvalidPrivacyLevelError(e: Exception) {
+        internalLogger.log(
+            InternalLogger.Level.ERROR,
+            listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
ambushwork, line 141: Can user take any action when they see this error?

ambushwork (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt
jonathanmos: It's in order for them to be aware that they have passed an invalid value (this would probably happen  in the future by them writing the wrong value in the xml, though atm we don't provide this option)

PR #2323, opened 2024-10-15
RUM-6385: Default sample rate for session replay

jonathanmos:
### What does this PR do?
Add a default sample rate of 100% for Session Replay (sample rate is now optional when creating the configuration).

### Motivation
Align with the other features. 

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
@@ -31,8 +31,9 @@ data class SessionReplayConfiguration internal constructor(
      * A Builder class for a [SessionReplayConfiguration].
      * @param sampleRate must be a value between 0 and 100. A value of 0
      * means no session will be recorded, 100 means all sessions will be recorded.
+     * If this value is not provided then Session Replay will default to a 100 sample rate.
      */
-    class Builder(@FloatRange(from = 0.0, to = 100.0) private val sampleRate: Float) {
+    class Builder(@FloatRange(from = 0.0, to = 100.0) private val sampleRate: Float = SAMPLE_IN_ALL_SESSIONS) {
         private var customEndpointUrl: String? = null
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -31,10 +31,12 @@ data class SessionReplayConfiguration internal constructor(
      * A Builder class for a [SessionReplayConfiguration].
      * @param sampleRate must be a value between 0 and 100. A value of 0
      * means no session will be recorded, 100 means all sessions will be recorded.
+     * If this value is not provided then Session Replay will default to a 100 sample rate.
      */
-    class Builder(@FloatRange(from = 0.0, to = 100.0) private val sampleRate: Float) {
+    class Builder(@FloatRange(from = 0.0, to = 100.0) sampleRate: Float? = null) {
```
ambushwork, line 36: **Question:** why not just add default value in constructor like:
```
 class Builder(@FloatRange(from = 0.0, to = 100.0) sampleRate: Float? = SAMPLE_IN_ALL_SESSIONS)
```

jonathanmos: I would have done this, but I want to avoid having the property be visible externally as optional. By doing it this way, externally the property of sampleRate is seen as `Float` instead of `Float?`

xgouchet: > [!NOTE]
> You can still have the property be non null (although, as the property is private it doesn't really matter), and have the constructor parameter use the default value : 

```
    class Builder(
        @FloatRange(from = 0.0, to = 100.0) private val sampleRate: Float = SAMPLE_IN_ALL_SESSIONS
    ) {
       // …
    }
```

xgouchet (approved):

ambushwork (approved):

PR #2324, opened 2024-10-16
RUM-6375: Force single core for session replay

jonathanmos:
### What does this PR do?
Allow session replay to run on only one core at a time. Attempting to start sr on more cores will send telemetry.

### Motivation
It probably doesn't make sense for session replay to run on multiple cores.

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
@@ -266,6 +266,8 @@ internal class DatadogCore(
         return coreFeature.createScheduledExecutorService(executorContext)
     }
 
+    override fun isCoreActive(): Boolean = isActive
+
     // endregion
 
     // region InternalSdkCore
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
@@ -103,7 +103,7 @@ object SessionReplay {
     }
 
     private fun isAlreadyRegistered() =
-        currentRegisteredCore?.get() != null
+        currentRegisteredCore?.get()?.isCoreActive() == true
 
     private fun logAlreadyRegisteredWarning(internalLogger: InternalLogger) =
         internalLogger.log(
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
@@ -6,17 +6,26 @@
 
 package com.datadog.android.sessionreplay
 
+import androidx.annotation.VisibleForTesting
 import com.datadog.android.Datadog
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
 import com.datadog.android.api.feature.Feature.Companion.SESSION_REPLAY_FEATURE_NAME
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.sessionreplay.internal.SessionReplayFeature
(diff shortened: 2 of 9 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
```diff
@@ -48,7 +58,18 @@ object SessionReplay {
                     startRecordingImmediately = sessionReplayConfiguration.startRecordingImmediately,
                     dynamicOptimizationEnabled = sessionReplayConfiguration.dynamicOptimizationEnabled
                 )
-                sdkCore.registerFeature(sessionReplayFeature)
+
+                val previousCore = activeSdkCore?.get()
```
jonathanmos, line 62: Is it necessary to also check if the core is active?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
```diff
@@ -83,4 +98,14 @@ object SessionReplay {
 
         sessionReplayFeature?.manuallyStopRecording()
     }
+
+    private fun isAlreadyRegistered() =
+        currentRegisteredCore?.get() != null
```
xgouchet, line 103: > [!IMPORTANT]
> The fact that a core instance is still present doesn't mean it's still active. We need to surface an `isActive()` function on the SDKCore interface to check whether or not the last registered core is still listening and sending events.

xgouchet (changes requested):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
jonathanmos: I've added a check for whether the core is still active

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
```diff
@@ -6,17 +6,26 @@
...
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.sessionreplay.internal.SessionReplayFeature
+import java.lang.ref.WeakReference
 
 /**
  * An entry point to Datadog Session Replay feature.
  */
 object SessionReplay {
 
+    @VisibleForTesting internal var currentRegisteredCore: WeakReference<SdkCore>? = null
```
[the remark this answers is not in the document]
xgouchet: The `SessionReplay` being a singleton object, it will stay in VM memory indefinitely, so indeed, having a standard reference to the core could have some memory leak.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -266,6 +266,8 @@ internal class DatadogCore(
         return coreFeature.createScheduledExecutorService(executorContext)
     }
 
+    override fun isCoreActive(): Boolean = isActive
```
xgouchet, line 269: > [!NOTE]
> Even though the implementation is trivial, it'd be nice to have some Unit tests around it.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplay.kt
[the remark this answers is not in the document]
ambushwork: Actually I am just wondering should we rely on GC to release the reference here? shouldn't we have something to unregister it explicitly?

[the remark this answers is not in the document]
jonathanmos: We're querying the weakreference and checking if the core is active so we should be ok. If either of those checks fails we overwrite the variable

ambushwork (approved):

PR #2325, opened 2024-10-16
Fix the regression for the TelemetryErrorEvent with throwable

mariusc83:
### What does this PR do?

When we refactor the Telemetry handling logic we missed a case when the `Log.Error` is reported with a `throwable`. Main reason why we missed this is that our integration tests were not reporting the problem in the CI (they were failing but this was not visible in the CI due to an issue in the shell environment on MacOs runners). This issue was currently discovered by @ambushwork while manually running the integration tests on his end.

There were 2 main problems caused by this regression was that in case this method was used `fun error(message: String, throwable: Throwable)`:

1. The `stacktrace` and `kind` were not correctly resolved from the `Throwable`.
2. In case 2 consecutive telemetry error events would be sent: first with `kind=null` and second with `throwable` because of the way we were resolving the `identifier` they were both considered same event and second one would not be sent. 

The impact is quite high as it is affecting all our telemetry error logs that are reporting only the `Throwable`.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTelemetryEvent.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTelemetryEvent.kt
@@ -6,6 +6,8 @@
 
 package com.datadog.android.internal.telemetry
 
+import com.datadog.android.internal.utils.loggableStackTrace
+
 @Suppress("UndocumentedPublicClass", "UndocumentedPublicFunction", "UndocumentedPublicProperty")
 sealed class InternalTelemetryEvent {
 
@@ -18,7 +20,15 @@ sealed class InternalTelemetryEvent {
             val error: Throwable? = null,
             val stacktrace: String? = null,
             val kind: String? = null
-        ) : Log(message, additionalProperties)
+        ) : Log(message, additionalProperties) {
+            fun resolveKind(): String? {
+                return kind ?: error?.javaClass?.canonicalName ?: error?.javaClass?.simpleName
+            }
+
+            fun resolveStacktrace(): String? {
+                return stacktrace ?: error?.loggableStackTrace()
+            }
+        }
     }
 
     data class Configuration(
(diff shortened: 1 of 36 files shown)


dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/ByteArrayExt.kt
```diff
@@ -24,6 +26,7 @@ private const val HEX_CHARS = "0123456789abcdef"
  * @return A hexadecimal [String] representation of the byte array.
  *
  */
+@InternalApi
```
ambushwork, line 29: > [!Note]
> Why function in internal module needs to be marked as `@InternalApi`?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTelemetryEvent.kt
```diff
@@ -6,7 +6,11 @@
 
 package com.datadog.android.internal.telemetry
 
+import com.datadog.android.internal.utils.loggableStackTrace
+import com.datadog.android.lint.InternalApi
+
 @Suppress("UndocumentedPublicClass", "UndocumentedPublicFunction", "UndocumentedPublicProperty")
+@InternalApi
```
ambushwork, line 13: > [!Note]
> Why function in internal module needs to be marked as `@InternalApi`?

mariusc83: In case somehow this is exposed through some module using `api`. I think it is safer to mark all these methods and classes with `@InternalApi`.

mariusc83: That's also true, let's remove these then.

ambushwork (approved):

PR #2328, opened 2024-10-17
RUM-6293: Remove Batch metrics inner sampler to increase sample rate

ambushwork:
### What does this PR do?

The whole context please check the ticket: RUM-6293

TL,DR;

The sample rate should be decreased by 10 times but wrongly decreased by 1000 times,

Since we have add sample rate option on `logMetric` method, now it doesn't make sense to have another sampler inside `BatchMetricsDispatcher`. By removing it, the sample rate is simply controlled by the parameter passed into `logMetric` and the total sample rate will be increased by (1/0.015f) = 66.6 times.

### Motivation

RUM-6293

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
@@ -16,8 +16,6 @@ import com.datadog.android.core.internal.persistence.file.existsSafe
 import com.datadog.android.core.internal.persistence.file.lengthSafe
 import com.datadog.android.core.internal.time.TimeProvider
 import com.datadog.android.core.metrics.MethodCallSamplingRate
-import com.datadog.android.core.sampling.RateBasedSampler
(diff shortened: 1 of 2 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -28,8 +26,7 @@ internal class BatchMetricsDispatcher(
     private val uploadConfiguration: DataUploadConfiguration?,
     private val filePersistenceConfig: FilePersistenceConfig,
     private val internalLogger: InternalLogger,
-    private val dateTimeProvider: TimeProvider,
-    private val sampler: Sampler = RateBasedSampler(METRICS_DISPATCHER_DEFAULT_SAMPLING_RATE)
```
0xnm, line 32: I hope we will do RUM-6394 soon as well, because we then create metric for nothing in most of the cases.

0xnm (approved):

mariusc83 (approved):

PR #2329, opened 2024-10-17
RUM-6596 Make sure the legacy integration test fail the CI job if fai…

mariusc83:
…ling

### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -53,10 +53,11 @@ stages:
     - set +e
     - exit_code=0
     - $ANDROID_HOME/emulator/emulator -avd "$EMULATOR_NAME" -grpc-use-jwt -no-snapstorage -no-audio -no-window -no-boot-anim -verbose -qemu -machine virt &
-    - GRADLE_OPTS="-Xmx3072m" ./gradlew :instrumented:integration:assembleDebug :instrumented:integration:assembleDebugAndroidTest --stacktrace --no-daemon $( (( $ANDROID_API <= 23 )) && echo "-Puse-api21-java-backport -Puse-desugaring" )
-    - $ANDROID_HOME/platform-tools/adb install -t -d $( (( $ANDROID_API >= 23 )) && echo "-g" ) -r instrumented/integration/build/outputs/apk/androidTest/debug/integration-debug-androidTest.apk
-    - $ANDROID_HOME/platform-tools/adb install -t -d $( (( $ANDROID_API >= 23 )) && echo "-g" ) -r instrumented/integration/build/outputs/apk/debug/integration-debug.apk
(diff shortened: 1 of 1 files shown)


instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/FragmentTrackingTest.kt
```diff
@@ -73,8 +74,8 @@ internal abstract class FragmentTrackingTest :
         // swipe to close the view
         onView(ViewMatchers.withId(R.id.btn_last)).perform(ViewActions.click())
         instrumentation.waitForIdleSync()
-        Thread.sleep(200) // give time to the view id to update
-
+        Thread.sleep(200)
```
ambushwork, line 77: > Tiny remark 
> Should we remove this sleep since we have `waitForPendingRUMEvents` below?

ambushwork (approved):

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/FragmentTrackingTest.kt
mariusc83: I want to keep it just for weird low API devices. To make sure it has time to stabilise. For some reason sometimes `instrumentation.waitForIdleSync` doesn't work really well on those emulators.

0xnm: I think @ambushwork is asking here about a bit different method - `waitForPendingRUMEvents`.

But speaking about `waitForIdleSync` - I'm curious why it doesn't work. Description of `IdleHandler` [says](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/os/MessageQueue.java;l=1131-1139?q=MessageQueue):

```
/**
 * Called when the message queue has run out of messages and will now
 * wait for more.  Return true to keep your idle handler active, false
 * to have it removed.  This may be called if there are still messages
 * pending in the queue, but they are all scheduled to be dispatched
 * after the current time.
 */
```

So, if click event above was added to the queue before the `waitForIdleSync`, it should work just fine? Or we have something in the test structure that works in async way?

mariusc83: I mean I can't tell you why it is not really working. It is just based on my previous experience with the emulators. I don't really mind removing this and hope this will work on all the emulators. I am about do it now actually.

PR #2332, opened 2024-10-18
Add non-blocking ITR CI jobs that run in parallel with the regular ones

nikita-tkachenko-datadog:
### What does this PR do?

For every CI job that was running tests with DD Java agent adds a "mirror" jobs that runs the same tests with ITR-enabled.
All the additional jobs are non-blocking.

### Motivation

The idea is to have ITR running in parallel with the regular CI flow for some time to evaluate that everything works correctly.
Once it is clear that ITR works as expected, the original jobs will be deleted and the new ones will take their place.

### Additional notes

The new jobs use a different Test Visibility service name (`dd-android-sdk-itr`) to avoid mixing ITR data with regular tests data.
When the switchover from the old jobs to the new ones is done, the service name will be updated.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -217,6 +217,7 @@ test:tools:
       - cache/caches/
       - cache/notifications/
   script:
+    - wget -O $DD_TRACER_FOLDER/dd-java-agent.jar 'https://output.circle-artifacts.com/output/job/317e0cef-c4a6-4e15-b659-0520d386f3df/artifacts/0/libs/dd-java-agent-1.42.0-SNAPSHOT.jar'
     - pip3 install datadog
     - rm -rf ~/.gradle/daemon/
     - export DD_AGENT_HOST="$BUILDENV_HOST_IP"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -180,6 +180,16 @@ test:debug:
     reports:
       junit: "**/build/test-results/testDebugUnitTest/*.xml"
 
+test:debug:
+  <<: *test-debug-template
+
+test:debug:itr:
```
mariusc83, line 186: isn't this going to increase considerably the execution time in the CI ?

nikita-tkachenko-datadog: This will increase compute time but not the wall time, as the new jobs will be running in parallel with the existing ones.
The idea is to have it like this for some time to ensure that the ITR-enabled jobs work adequately, and only after that to remove the old jobs.

nikita-tkachenko-datadog:
Putting this on pause again, as in order to migrate from CodeCov to Datadog's coverage we will need to be able to display combined code coverage for multiple Gradle builds, which is not supported yet.

PR #2334, opened 2024-10-20
RUM-6566: Implement touch override

jonathanmos:
### What does this PR do?

Implements an api to override touch privacy on views.

Example of the api:

```
val view = root.findViewById...
view.setSessionReplayTouchPrivacy([TouchPrivacy])
```

Note:

• Touch privacy is calculated every time an `ACTION_DOWN` is performed. In this way we can apply the override to the entire motion.
• Views can be overridden to hide touches, but also to **show** touches (if the general default was hide).
• Webviews are not affected by overrides, aside from `hidden` and touch privacy (so they are affected by this change).
• Setting `hidden` on views only affects whether a view is replaced by a placeholder. Hiding the touches on the view requires a separate touch override.
• If a touch falls within two override areas, one that has been overridden to `HIDE` and another that has been overridden to `SHOW`, `HIDE` will take priority.

### Motivation

Part of the fine grained masking effort.

### Additional Notes


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
```diff
@@ -115,6 +123,35 @@ internal class TreeViewTraversal(
...
+            val locationOnScreen = IntArray(2)
+
+            // this will always have size >= 2
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            view.getLocationOnScreen(locationOnScreen)
+
+            val x = locationOnScreen[0]
+            val y = locationOnScreen[1]
+            val viewArea = Rect(
+                x - view.paddingLeft,
```
ambushwork, line 140: > Question

Why we need to remove the padding of view for the touch privacy? I think we can include them since it's also the scope of the view

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/TouchPrivacyManager.kt
```diff
@@ -0,0 +1,72 @@
...
+import android.graphics.Rect
+import androidx.annotation.UiThread
+import androidx.annotation.VisibleForTesting
+import com.datadog.android.sessionreplay.TouchPrivacy
+
+internal class TouchPrivacyManager(
+    private val globalTouchPrivacy: TouchPrivacy
+) {
+    // areas on screen where overrides are applied
+    private val currentSnapshot = HashMap<Rect, TouchPrivacy>()
```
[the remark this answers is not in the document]
jonathanmos: Yes, you're right that it's too close. Renamed to `currentOverrideAreas` and `nextOverrideAreas`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/TouchPrivacyManager.kt
```diff
@@ -0,0 +1,72 @@
...
+        return if (isOverriddenToShowTouch) true else globalTouchPrivacy == TouchPrivacy.SHOW
+    }
+
+    @VisibleForTesting
+    internal fun getCurrentOverrideSnapshot(): Map<Rect, TouchPrivacy> {
+        return currentOverrideAreas
+    }
+
+    @VisibleForTesting
+    internal fun getNextOverrideSnapshot(): Map<Rect, TouchPrivacy> {
```
mariusc83, line 65: **🎯 suggestion:** ‏ are you modifying this ? If we call it snapshot then we should probably return a `copy` of it.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/TouchPrivacyManager.kt
```diff
@@ -0,0 +1,72 @@
...
+        return nextOverrideAreas
+    }
+
+    private fun isWithinOverrideArea(touchPoint: Point, overrideArea: Rect): Boolean =
```
mariusc83, line 69: **🎯 suggestion:** ‏ I think here you can user `overrideArea.contains(x,y)` right ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
```diff
@@ -115,6 +123,35 @@ internal class TreeViewTraversal(
...
+    @UiThread
+    private fun updateTouchOverrideAreas(view: View) {
+        val touchPrivacy = view.getTag(R.id.datadog_touch_privacy)
+
+        if (touchPrivacy != null) {
+            val locationOnScreen = IntArray(2)
+
+            // this will always have size >= 2
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            view.getLocationOnScreen(locationOnScreen)
```
mariusc83, line 135: **🎯 suggestion:** ‏ I am afraid that we are already doing this operation when we want to detect the view position on screen. We should make sure we only do this once and re - use the resolved coordinates as this operation is costly.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/TouchPrivacyManager.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.datadog.android.sessionreplay.TouchPrivacy
+
+internal class TouchPrivacyManager(
+    private val globalTouchPrivacy: TouchPrivacy
+) {
+    // areas on screen where overrides are applied
+    private val currentOverrideAreas = HashMap<Rect, TouchPrivacy>()
+
+    // built during the view traversal and copied to currentOverrideSnapshot at the end
+    private val nextOverrideAreas = HashMap<Rect, TouchPrivacy>()
```
mariusc83, line 22: **❔ question:** ‏ I wonder why do you need to keep 2 HashMaps here. What's the purpose of this approach ? Can we add more explanations here please as it is not very obvious from the review. Tks !!

jonathanmos: I'll add some additional comments to clarify, but briefly it's for two main reasons. First, because the touch handling is happening in parallel to the traversal so we want to resolve privacy using a complete map of privacy areas. Second, in order not to have to track the lifecycle of the overridden views to know when to remove them from the map. Using this method, all items are replaced on every traversal with the current privacy state.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt
jonathanmos: While it is expensive, in this case it happens only when we encounter the touch override flag on a view so it should be extremely infrequent. We are calling `getLocationOnScreen` in multiple places in Session Replay so refactoring to consolidate these calls for every traversal would be beneficial but I think out of scope for this pr. I can open a ticket to look into this.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/TouchPrivacyManager.kt
jonathanmos: These getters are used only by the unit tests. The maps are modified only by the class itself.

mariusc83 (approved):

PR #2337, opened 2024-10-21
Fix Image reflection issue and proguard rules

ambushwork:
### What does this PR do?

* Changes some proguard rules in order to make the code work in release build.
* Changes some reflection logic for image mapper in order to make it work in higher version of `Coil`
* Change the bitmap copy config in order to make it work for Bitmap ALPHA config.
* Update the sample app image showcase screen.

### DIFF
--- a/features/dd-sdk-android-session-replay-compose/consumer-rules.pro
+++ b/features/dd-sdk-android-session-replay-compose/consumer-rules.pro
@@ -7,7 +7,7 @@
 -keepnames class androidx.compose.runtime.Composition
 -keepnames class androidx.compose.ui.platform.ComposeView
 -keepnames class androidx.compose.material.DefaultButtonColors
--keepclassmembers class androidx.compose.ui.platform.WrappedComposition {
+-keep class androidx.compose.ui.platform.WrappedComposition {
     <fields>;
 }
 -keepclassmembers class androidx.compose.ui.platform.AbstractComposeView {
@@ -22,8 +22,8 @@
 -keepclassmembers class androidx.compose.foundation.BackgroundElement {
      <fields>;
 }
--keepclassmembers class androidx.compose.ui.node.LayoutNode {
-     <fields>;
+-keep class androidx.compose.ui.node.LayoutNode {
+     *;
 }
 -keepclassmembers class androidx.compose.ui.draw.PainterElement {
      <fields>;
@@ -43,10 +43,16 @@
 -keepclassmembers class androidx.compose.ui.graphics.AndroidImageBitmap {
      <fields>;
 }
--keepclassmembers class coil.compose.ContentPainterModifier {
+-keep class coil.compose.ContentPainterModifier {
      <fields>;
 }
--keepclassmembers class coil.compose.AsyncImagePainter {
+-keep class coil.compose.AsyncImagePainter {
      <fields>;
+}
+-keepclassmembers class androidx.compose.foundation.layout.PaddingElement{
+    <fields>;
+}
+-keepclassmembers class "androidx.compose.ui.graphics.GraphicsLayerElement"{
+    <fields>;
 }
 
--- a/features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/semantics/ImageSemanticsNodeMapper.kt
+++ b/features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/semantics/ImageSemanticsNodeMapper.kt
@@ -72,6 +72,10 @@ internal class ImageSemanticsNodeMapper(
             }
         }
         // TODO RUM-6535: support more painters.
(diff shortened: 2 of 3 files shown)


features/dd-sdk-android-session-replay-compose/consumer-rules.pro
```diff
@@ -43,10 +43,16 @@
...
 }
--keepclassmembers class coil.compose.ContentPainterModifier {
+-keep class coil.compose.ContentPainterModifier {
      <fields>;
 }
--keepclassmembers class coil.compose.AsyncImagePainter {
+-keep class coil.compose.AsyncImagePainter {
      <fields>;
 }
+-keepclassmembers class androidx.compose.foundation.layout.PaddingElement{
```
0xnm, line 52: in the changes above we are _sometimes_ replacing usages of `keepclassmembers` to just `keep`, so is using it here won't cause any issues?

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/semantics/ImageSemanticsNodeMapper.kt
```diff
@@ -82,7 +86,7 @@ internal class ImageSemanticsNodeMapper(
 
         val newBitmap = bitmap?.let {
             @Suppress("UnsafeThirdPartyFunctionCall") // isMutable is always false
-            it.copy(it.config, false)
+            it.copy(Bitmap.Config.ARGB_8888, false)
```
0xnm, line 89: what is the reason for this change?

0xnm (approved):

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/semantics/ImageSemanticsNodeMapper.kt
ambushwork: When testing with some other applications, I found that due to the original config of bitmap, some bitmaps are empty after copying it, for example `ALPHA_8`, I need to tweak the config in order to make it work.

Here is a related question on SO:
https://stackoverflow.com/questions/9247369/alpha-8-bitmaps-and-getpixel

0xnm: From SO:

> Simplest workaround is to waste lots of memory

Not sure though if it is useful to try to record `ALPHA_8` bitmaps, because documentation says:

```
 * Each pixel is stored as a single translucency (alpha) channel.
 * This is very useful to efficiently store masks for instance.
 * No color information is stored.
 * With this configuration, each pixel requires 1 byte of memory.
```

So probably it is a bitmask layer, rather than something that is rendered to the user. Also memory usage will be much higher - 4x more (pixel is 1 byte vs pixel is 4 bytes in cases of `ARGB_8888`).

Maybe we should keep here a TODO item for the further investigation? Maybe it makes sense to skip recording `ALPHA_8` bitmaps (if it won't affect anything).

ambushwork: theoretically yes it should be a translucency channel, but when I debug I see a bitmap like this:
image

and its original config is ALPHA_8. With the previous code it will copy nothing, but when changing to ARGB_888, it has real data.

0xnm: okay, makes sense then if we assume that with such config pixel is always just black and some alpha is applied. Maybe then we should use `ARGB_8888` only when original config is `ALPHA_8`? And use the original config otherwise.

xgouchet (approved):

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/mappers/semantics/ImageSemanticsNodeMapper.kt
ambushwork: I haven't test exhaustively all the config, but I will create a ticket to investigate all of them

features/dd-sdk-android-session-replay-compose/consumer-rules.pro
ambushwork: actually it depends on the usage, if in the code we want to check the class instance like this:
```
 if (ComposeReflection.WrappedCompositionClass?.isInstance(composition) == true) {

}
```
then only keeping the class members is not enough, we need to keep the class itself as well, otherwise we can just stick with `keepclassmembers`

PR #2342, opened 2024-10-24
RUM-5873: Reactivate Session Replay instrumented test for API 21

ambushwork:
### What does this PR do?

Reactive Session Replay instrumented test for API 21, 
Fix a drawable resolving issue due to the different behavior on lower API.

### Motivation

RUM-5873


0xnm (approved):
lgtm, but I believe we will still have a wrong stacktrace in case if test fails on API 21, because some classes used by AssertJ don't exist there?

ambushwork:
@0xnm I am not sure what do you mean, can you elaborate?

0xnm:
I mean AssertJ is using some Java 8+ classes (ex. `java.time.*`) when printing failure message which don't exist on API 21.

0xnm:
The only quick workaround is desugaring, but I don't think we want to enable it. So let's keep it like that.

PR #2347, opened 2024-10-24
RUM-5814 Provide device and os info into local spans metadata

mariusc83:
### What does this PR do?

After a short analysis on our local traces vs the traces created by RUM backend based on our RUM resource events we realized that we are missing some content for the local ones: `device` and `os` informations.
In this PR we are aligning the information in the local traces vs RUM backend traces by adding the `device` and `os` information into the local traces `meta`. 

Please note that in order to keep schemas aligned I added these new properties also into the `benchmark` module `span-schema.json` but made them optional as right now the benchmark module is not using these (we might need it later).

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/BaseSpanEventMapper.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/BaseSpanEventMapper.kt
@@ -1,0 +1,74 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.trace.internal.domain.event
+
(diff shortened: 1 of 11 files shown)


features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/event/BaseSpanEventMapper.kt
```diff
@@ -0,0 +1,74 @@
...
+
+package com.datadog.android.trace.internal.domain.event
+
+import com.datadog.android.api.context.DeviceInfo
+import com.datadog.android.api.context.DeviceType
+import com.datadog.android.api.context.NetworkInfo
+import com.datadog.android.api.context.UserInfo
+import com.datadog.android.trace.model.SpanEvent
+
+internal abstract class BaseSpanEventMapper {
```
xgouchet, line 15: > [!NOTE]
> Maybe we could make this a `BaseSpanEventMapper<T>: ContextAwareMapper<T, SpanEvent>`, as it's only used on classes that extend this, and the purpose of this class is linked with creating SpanEvent content.

xgouchet (approved):

jonathanmos (approved):

PR #2352, opened 2024-10-28
Add DrawableCopier interface for drawable copying for image wireframe

ambushwork:
### What does this PR do?

In Session Replay, we need to copy the drawable in case of any GC or modifying original resource issue.
However each mappers have slightly different way of copying the drawable, for example in most of case it's enough to just copy the drawable like : 
```
originalDrawable.constantState?.newDrawable(resources)
```

But in some mappers like `SwtichMapper` we need to append the tint list and the drawable state.

So `DrawableCopier` interface is added to make it possible that mapper define its own method to make the copy.

And `DefaultDrawableCopier` is provided for most common case.


jonathanmos (approved):
LGTM!

PR #2357, opened 2024-10-29
RUM-6896: Add custom drawable mappers to resolve extension drawable

ambushwork:
### What does this PR do?

In Session Replay, we noticed that there are some components have transparent background in player, which make the content not visible.

After investigation, we see that it's due to the background drawable that we currently can not parse.

According to Telemetry, here is a top list of the drawable that we don't support:

image

On top of that, it's `MaterialShapeDrawable`, which belongs to material components, in this case we need to inject a custom mapper from `dd-sdk-android-session-replay-material` to `dd-sdk-android-session-replay`, so that when user add the extension of material, it can parse the drawable.

### Motivation

RUM-6896:

### Demo

| Before | After  |
| --- | --- |
|Screenshot 2024-10-29 at 13 44 00 | Screenshot 2024-10-29 at 13 45 48|

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/utils/LegacyDrawableToColorMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/utils/LegacyDrawableToColorMapper.kt
@@ -10,27 +10,42 @@ package com.datadog.android.sessionreplay.utils
 import android.annotation.SuppressLint
 import android.graphics.Paint
 import android.graphics.PorterDuffColorFilter
+import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
 import android.graphics.drawable.GradientDrawable
 import android.graphics.drawable.InsetDrawable
 import android.graphics.drawable.LayerDrawable
 import android.graphics.drawable.RippleDrawable
+import android.graphics.drawable.ShapeDrawable
+import android.graphics.drawable.VectorDrawable
 import com.datadog.android.api.InternalLogger
 
 /**
  * Drawable utility object needed in the Session Replay Wireframe Mappers.
  * This class is meant for internal usage so please use it carefully as it might change in time.
  */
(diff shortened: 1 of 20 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/utils/LegacyDrawableToColorMapper.kt
```diff
@@ -10,27 +10,42 @@ package com.datadog.android.sessionreplay.utils
...
+open class LegacyDrawableToColorMapper(
+    private val extensionMappers: List<DrawableToColorMapper> = emptyList()
+) : DrawableToColorMapper {
 
     override fun mapDrawableToColor(drawable: Drawable, internalLogger: InternalLogger): Int? {
+        // First check if extension mappers can resolve the drawable
+        extensionMappers.forEach {
+            val result = it.mapDrawableToColor(drawable, internalLogger)
+            if (result != null) {
+                return result
```
0xnm, line 37: minor: to avoid confusion with the return target (`forEach` vs `mapDrawableToColor`) maybe it is better to use `firstOrNull` and then do null check?

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/utils/LegacyDrawableToColorMapper.kt
ambushwork: `firstOrNull` returns the `extensionMapper` which meets the condition instead of produced value, then we need to call the function again?
The only operator seems work here is `firstNotNullOfOrNull` but you told me this API is complex to understand ?

ambushwork: `firstNotNullOf ` is not safe to call if all the produced value are null, it can throw `NoSuchElementException`, that's why there is  an API called `firstNotNullOfOrNull`

PR #2358, opened 2024-10-29
Use repository-specific AMI with macOS Sonoma runners

0xnm:
### What does this PR do?

This PR makes use of macOS Sonoma runners with the Android SDK repo dedicated AMI.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -32,16 +32,8 @@ stages:
   - notify
 
 .snippets:
-  install-android-sdk:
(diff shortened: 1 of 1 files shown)


ambushwork (approved):

jorgetomtz:
Given that the android sdk and friends are being installed on the AMI now, can we also test removing those parts from the CI script to save time?

0xnm:
@jorgetomtz I removed cmdline tools installation part, but kept `sdkmanager` calls, because we still need to install components not coming with AMI (in particular for the certain API version which will vary over the time). But anyway, if certain API already has the necessary component installed, such call to `sdkmanager` will be no-op (I hope).

jorgetomtz (approved):

PR #2365, opened 2024-10-31
Fix CompoundButton mapper drawable clone issue

ambushwork:
### What does this PR do?

This PR is a follow-up of this [PR](https://github.com/DataDog/dd-sdk-android/pull/2354).

There were still some mapper copying their drawable before retrieving the cache, especially in `CompoundButton` mappers,

So this PR cleans up the related issue.

### Motivation

RUM-6897

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableTextViewMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableTextViewMapper.kt
@@ -11,6 +11,7 @@ import android.widget.Checkable
 import android.widget.TextView
 import androidx.annotation.UiThread
 import com.datadog.android.api.InternalLogger
+import com.datadog.android.sessionreplay.internal.recorder.resources.DrawableCopier
 import com.datadog.android.sessionreplay.model.MobileSegment
 import com.datadog.android.sessionreplay.recorder.MappingContext
 import com.datadog.android.sessionreplay.recorder.mapper.TextViewMapper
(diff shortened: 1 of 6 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableTextViewMapper.kt
```diff
@@ -109,6 +111,11 @@ internal abstract class CheckableTextViewMapper<T>(
                 view,
                 mappingContext.systemInformation.screenDensity
             )
+            val drawableCopier = object : DrawableCopier {
```
0xnm, line 114: you can add `fun` to the definition of `DrawableCopier` making it a functional interface: `fun interface DrawableCopier`.

This will allow you to write simply `val drawableCopier = DrawableCopier { originalDrawable, resources -> cloneCheckableDrawable(view, it) }`.

But anyway, since it is a stateless object maybe it is not worth to allocate a new one on each call, so you can lift it to the class property.

0xnm (changes requested):

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableTextViewMapper.kt
ambushwork: Updated with `fun interface`, meanwhile it's not stateless object, for example this implementation relies on the view instance passed from the function.

0xnm: Yes, but there is no any property in this object, so all the changes are local to the function and don't leave that scope.

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableTextViewMapper.kt
ambushwork: Yes I agree that there is no property for the object, but it still seems not possible to lift it to the class property due to the usage of view instance.

PR #2351, opened 2024-10-28
RUM-6897: Fix scrolling issue

jonathanmos:
### What does this PR do?
Part of fixing the frozen scrolling when MASK_NONE is selected for image privacy. We were using the copy of the drawable as the key in the resource cache and this resulted in cache misses. This pr changes the behavior to use the original drawable to generate the key.

### Motivation

What inspired you to submit this pull request?

### Additional Notes

Anything else we should know when reviewing?

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](CONTRIBUTING.md) doc)

PR #2367, opened 2024-10-31
Fix resolveResourceId not correctly calling job finished when drawable cloning failed

ambushwork:
### What does this PR do?

Snapshot item is never processed due to the fix in this [PR](https://github.com/DataDog/dd-sdk-android/pull/2351), 
this is because we call `job.started` callback but never call `job.finished` , so the snapshot is always in pending state.

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
@@ -62,7 +62,11 @@ internal class ResourceResolver(
             return
         }
 
-        val copiedDrawable = drawableCopier.copy(originalDrawable, resources) ?: return
+        val copiedDrawable = drawableCopier.copy(originalDrawable, resources)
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
```diff
@@ -62,7 +62,11 @@ internal class ResourceResolver(
             return
         }
 
-        val copiedDrawable = drawableCopier.copy(originalDrawable, resources) ?: return
+        val copiedDrawable = drawableCopier.copy(originalDrawable, resources)
+        if (copiedDrawable == null) {
+            resourceResolverCallback.onFailure()
```
jonathanmos, line 67: We should add a unit test to cover this

ambushwork: Good point, updated

jonathanmos (approved):

xgouchet (approved):

PR #2372, opened 2024-11-04
RUM-3460: Fix padding and resizing issue for image view mapper

ambushwork:
### What does this PR do?

In app development, clients usually use paddings to resize the content drawable in a given size image view, prior to this PR we didn't taken into account these paddings to resize our image wireframe, which leads some images/icons bigger than real.

So the updates in this PR:
* Add paddings for a demo image view in sample app
* Add paddings calculation in `resolveParentRectAbsPosition` function
* Fix the issue that  `GradientDrawable` didn't take into account the width&height after resizing.

### Motivation

RUM-3460

### Demo
image

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/DefaultImageWireframeHelper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/DefaultImageWireframeHelper.kt
@@ -7,7 +7,6 @@
 package com.datadog.android.sessionreplay.internal.recorder.resources
 
 import android.graphics.drawable.Drawable
-import android.graphics.drawable.GradientDrawable
 import android.graphics.drawable.InsetDrawable
 import android.graphics.drawable.LayerDrawable
 import android.view.View
@@ -238,7 +237,6 @@ internal class DefaultImageWireframeHelper(
                 }
             }
 
-            is GradientDrawable -> DrawableProperties(drawable, view.width, view.height)
             else -> DrawableProperties(drawable, width, height)
         }
     }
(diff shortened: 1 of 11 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/DefaultImageWireframeHelper.kt
```diff
@@ -238,7 +237,6 @@ internal class DefaultImageWireframeHelper(
                 }
             }
 
-            is GradientDrawable -> DrawableProperties(drawable, view.width, view.height)
```
0xnm, line 241: why this PR has this removed?

0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/DefaultImageWireframeHelper.kt
ambushwork: I will elaborate what is written in description:
"Fix the issue that GradientDrawable didn't take into account the width&height after resizing."
So after fixing the padding issue, I found that `GradientDrawable` is not correct resized,  which makes the views (purple circles and squares) exceed their bounds, so I have to fix it by removing this line. By removing this line, `GradientDrawable` will not be handled differently, it will fallback to `else` condition.

jonathanmos (approved):

xgouchet (approved):

xgouchet (approved):
> [!NOTE]
> Shouldn't we have some unit tests around those functions ?

ambushwork:
> Note
> 
> Shouldn't we have some unit tests around those functions ?

@xgouchet there are unit tests around these, I just updated with the padding case to cover it better, and instrumented test payloads are regenerated for them.

PR #2376, opened 2024-11-05
RUM-6400 Add warning log when initializing the SDK outside of main process

mariusc83:
### What does this PR do?

During our escalation sessions with our clients we encountered several situations where the SDK was not initialised outside the main process. Unless the application is using 2 different processes and has 2 SDK instances (one in each process) this scenario will lead to a non - functional SDK as the data will no be uploaded as we are not initialising our schedulers in a secondary process. 

In this PR we are adding a warning log for our users when the SDK is initialised in a secondary process just to better debug this situations and avoid long escalation back and forth. 

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
@@ -626,6 +626,13 @@ internal class CoreFeature(
         } else {
             appContext.packageName == currentProcess.processName
         }
+        if (!isMainProcess) {
+            internalLogger.log(
+                InternalLogger.Level.WARN,
+                InternalLogger.Target.USER,
+                { SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE }
+            )
+        }
     }
 
     private fun shutDownExecutors() {
@@ -676,6 +683,10 @@ internal class CoreFeature(
     // endregion
 
     companion object {
+        internal const val SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE =
+            "Datadog SDK was initialized in a secondary process: although data will still be captured," +
+                " nothing will be uploaded from this process. Make sure to also initialize the SDK from the main" +
+                " process of your application."
 
         internal val DEFAULT_FLUSHABLE_EXECUTOR_SERVICE_FACTORY =
             FlushableExecutorService.Factory { logger, executorContext, backPressureStrategy ->
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
@@ -781,6 +781,11 @@ internal class CoreFeatureTest {
 
         // Then
         assertThat(testedFeature.isMainProcess).isFalse()
+        mockInternalLogger.verifyLog(
+            InternalLogger.Level.WARN,
+            InternalLogger.Target.USER,
+            CoreFeature.SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE
+        )
     }
 
     @Test


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -626,6 +626,13 @@ internal class CoreFeature(
         } else {
             appContext.packageName == currentProcess.processName
         }
+        if (!isMainProcess) {
+            internalLogger.log(
```
mariusc83, line 630: @reviewers should we also send a telemetry here ? I think it is a bit overkill, what do you think ?

0xnm: would it be actionable? how it would help us?

mariusc83: not sure it will be that actionable, that is the reason why I did not add it and just asked your opinion. I think we are better off.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -676,6 +683,11 @@ internal class CoreFeature(
     // endregion
 
     companion object {
+        internal const val SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE =
+            "Datadog SDK was initialized in a secondary process. We strongly recommend to initialize the SDK" +
+                "in the application main process otherwise the SDK would not function properly. The only scenario" +
+                "where this is acceptable is where there will be 2 instances of the SDKs running " +
+                "in the same time and one instance is guaranteed to be in the main process."
```
0xnm, line 690:
```suggestion
        internal const val SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE =
            "Datadog SDK was initialized in a secondary process. We strongly recommend to initialize the SDK" +
                " in the application main process otherwise the SDK would not function properly. The only scenario" +
                " where this is acceptable is where there will be 2 instances of the SDKs running" +
                " in the same time and one instance is guaranteed to be in the main process."
```

I feel though that this is very verbose. Maybe we can re-word it simpler? Like tell user that SDK _must_ have instance initialized in the main process and that is it. And speaking of `the SDK would not function properly`: maybe we can just say that no data will be recorded (well, to be precise - uploaded).

This warning though raises a question why such limitation exists and why cannot we solve it (in theory we could?).

mariusc83: Yes I wanted to present all the possibilities in the message but I guess we can lower the amount of information. Regarding your second point there are some strong reasons why we took this approach. Let's debate about this.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -676,6 +683,9 @@ internal class CoreFeature(
     // endregion
 
     companion object {
+        internal const val SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE =
+            "Datadog SDK was initialized in a secondary process. We strongly recommend to initialize the SDK" +
+                "in the application main process. The sdk will not upload any data in this case."
```
xgouchet, line 688: > [!NOTE]
> `We strongly recommend to initialize the SDK in the application main process`
> This part of the message makes it seem like initialising the SDK in a secondary process is wrong, when it's totally legitimate to want to capture events in a non main process. 
> Maybe we could phrase it differently, e.g.: 

```suggestion
            "Datadog SDK was initialized in a secondary process: although data will still be captured, nothing will be uploaded from this process." + 
            "Make sure to also initialize the SDK from the main process of your application."
```

xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
mariusc83: yes I agree this makes more sense. the first message was making more sense with my initial message where I explained the difference between running it in the main or main + secondary also.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
ambushwork: I think it's an interesting data to know at least, and if we notice that there are really amount of clients doing it in background (probably meaning intentionally), the action will be making it work even if the initialization happens in background.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -676,6 +683,10 @@ internal class CoreFeature(
     // endregion
 
     companion object {
+        internal const val SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE =
+            "Datadog SDK was initialized in a secondary process: although data will still be captured," +
+                "nothing will be uploaded from this process. Make sure to also initialize the SDK from the main" +
+                " process of your application."
```
ambushwork, line 689: Question: does "the main process" refers to "main thread" in Android?

xgouchet: In Android each process has its own JVM running code, each with several threads.

xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -676,6 +683,10 @@ internal class CoreFeature(
     // endregion
 
     companion object {
+        internal const val SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE =
+            "Datadog SDK was initialized in a secondary process: although data will still be captured," +
+                "nothing will be uploaded from this process. Make sure to also initialize the SDK from the main" +
```
0xnm, line 688:
```suggestion
            "Datadog SDK was initialized in a secondary process: although data will still be captured," +
                " nothing will be uploaded from this process. Make sure to also initialize the SDK from the main" +
```

PR #2379, opened 2024-11-06
RUM-7085:Add compose session replay scenario for benchmark sample application

ambushwork:
### What does this PR do?

Add session replay compose scenario, so that the application can switch to compose view when receive the bundle argument, and also in metrics we can filter with `scenario:sr_compose` for compose session replay.

### Motivation

RUM-7085

### DIFF
--- a/sample/benchmark/src/main/java/com/datadog/benchmark/sample/fragment/SessionReplayAppcompatFragment.kt
+++ b/sample/benchmark/src/main/java/com/datadog/benchmark/sample/fragment/SessionReplayAppcompatFragment.kt
@@ -20,7 +20,7 @@ import android.widget.ImageView
 import android.widget.NumberPicker
 import android.widget.Spinner
 import androidx.fragment.app.Fragment
-import androidx.navigation.fragment.NavHostFragment.findNavController
+import androidx.navigation.fragment.NavHostFragment.Companion.findNavController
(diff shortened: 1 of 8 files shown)


sample/benchmark/src/main/java/com/datadog/benchmark/sample/fragment/SessionReplayAppcompatFragment.kt
```diff
@@ -20,7 +20,7 @@ import android.widget.ImageView
 import android.widget.NumberPicker
 import android.widget.Spinner
 import androidx.fragment.app.Fragment
-import androidx.navigation.fragment.NavHostFragment.findNavController
+import androidx.navigation.fragment.NavHostFragment.Companion.findNavController
```
0xnm, line 23: is this change coming from navigation dependency update brought by the Compose BOM?

0xnm (approved):

sample/benchmark/src/main/java/com/datadog/benchmark/sample/fragment/SessionReplayAppcompatFragment.kt
ambushwork: Yes exactly, it seems there is an override version of navigation lib due to the compose bom import.

PR #2385, opened 2024-11-12
RUM-6866 Allow typed Sampler

xgouchet:
### What does this PR do?

Preparing for a new implementation of a sampler for the TracingInterceptor, add a type to the sampler to allow adapting the sampling decision to the event being sampled.

### DIFF
--- a/dd-sdk-android-core/api/apiSurface
+++ b/dd-sdk-android-core/api/apiSurface
@@ -323,14 +323,14 @@ interface com.datadog.android.core.persistence.Serializer<T: Any>
 fun <T: Any> Serializer<T>.serializeToByteArray(T, com.datadog.android.api.InternalLogger): ByteArray?
 data class com.datadog.android.core.persistence.datastore.DataStoreContent<T: Any>
   constructor(Int, T?)
-class com.datadog.android.core.sampling.RateBasedSampler : Sampler
+open class com.datadog.android.core.sampling.RateBasedSampler<T: Any> : Sampler<T>
   constructor(() -> Float)
   constructor(Float)
   constructor(Double)
-  override fun sample(): Boolean
+  override fun sample(T): Boolean
   override fun getSampleRate(): Float
-interface com.datadog.android.core.sampling.Sampler
-  fun sample(): Boolean
+interface com.datadog.android.core.sampling.Sampler<T: Any>
+  fun sample(T): Boolean
   fun getSampleRate(): Float?
 interface com.datadog.android.core.thread.FlushableExecutorService : java.util.concurrent.ExecutorService
   fun drainTo(MutableCollection<Runnable>)
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/sampling/RateBasedSampler.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/sampling/RateBasedSampler.kt
@@ -13,10 +13,11 @@ import java.security.SecureRandom
 /**
  * [Sampler] with the given sample rate which can be fixed or dynamic.
  *
+ * @param T the type of items to sample.
  * @param sampleRateProvider Provider for the sample rate value which will be called each time
  * the sampling decision needs to be made. All the values should be on the scale [0;100].
  */
-class RateBasedSampler(private val sampleRateProvider: () -> Float) : Sampler {
+open class RateBasedSampler<T : Any>(private val sampleRateProvider: () -> Float) : Sampler<T> {
 
     /**
      * Creates a new instance of [RateBasedSampler] with the given sample rate.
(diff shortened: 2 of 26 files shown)


dd-sdk-android-core/api/apiSurface
```diff
@@ -323,14 +323,14 @@ interface com.datadog.android.core.persistence.Serializer<T: Any>
...
   constructor(() -> Float)
   constructor(Float)
   constructor(Double)
-  override fun sample(): Boolean
+  override fun sample(T): Boolean
   override fun getSampleRate(): Float
-interface com.datadog.android.core.sampling.Sampler
-  fun sample(): Boolean
+interface com.datadog.android.core.sampling.Sampler<T: Any>
+  fun sample(T): Boolean
```
0xnm, line 333: this is potentially a binary-incompatible change, because public constructors of `DatadogInterceptor` and `TracingInterceptor` (and builder methods since recently) take `Sampler` instance which can be some customer-defined `Sampler` rather than default `RateBasedSampler`.

Is there a way to avoid breaking it?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/sampling/RateBasedSampler.kt
```diff
@@ -13,13 +13,14 @@ import java.security.SecureRandom
...
+ * @param T the type of items to sample.
  * @param sampleRateProvider Provider for the sample rate value which will be called each time
  * the sampling decision needs to be made. All the values should be on the scale [0;100].
  */
-class RateBasedSampler(private val sampleRateProvider: () -> Float) : Sampler {
+open class RateBasedSampler<T : Any>(private val sampleRateProvider: () -> Float) : Sampler<T> {
 
     /**
-     * Creates a new instance of [RateBasedSampler] with the given sample rate.
+     * Creates a new instance lof [RateBasedSampler] with the given sample rate.
```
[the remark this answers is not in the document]
xgouchet: 👍

dd-sdk-android-core/api/apiSurface
xgouchet: I hesitated to do it, as indeed it's technically a breaking change on a public interface. That being said, looking at our code base, the interface is used in: 
- the `internal class TelemetryEventHandler` constructor, but the sampler is built internally based on a sample rate provided in the configuration
- the `internal class SessionReplayFeature` constructor, but the sampler is built internally based on a sample rate provided in the configuration
- the `internal class DatadogLogHandler` constructor, but the sampler is built internally based on a sample rate provided in the configuration
- an internal property of the `internal class WebViewLogEventConsumer`, built internally based on a sample rate provided in the configuration
- in the `DatadogInterceptor` and `TracingInterceptor` **deprecated** constructors; they were replaced by a Builder. The builder only accepts a Sampler, which is the only place we could have ab issue for customer, which I guess is acceptable. I'll also add a builder method to set the sampling rate as I feel it's missing.

mariusc83: **🎯 suggestion:** ‏ What about a keeping the builder method that accepts the `Sampler` and wrapping this old `Sampler` that the customer is passing to your own `Sampler<Unit>` ? In this case it will not be a breaking change anymore. And we could deprecate that method for later.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -788,16 +789,23 @@ internal constructor(
...
          * auto-instrumented requests.
          * @param traceSampler the trace sampler controlling the sampling of APM traces.
-         * By default it is [RateBasedSampler], which either can accept
-         * fixed sample rate or can get it dynamically from the provider. Value between `0.0` and
-         * `100.0`. A value of `0.0` means no trace will be kept, `100.0` means all traces will
-         * be kept (default value is `20.0`).
+         * By default it is a sampler accepting 20% of the traces.
          */
-        fun setTraceSampler(traceSampler: Sampler): R {
+        fun setTraceSampler(traceSampler: Sampler<Span>): R {
```
mariusc83, line 808: **🎯 suggestion:** ‏ Like I suggested in my previous comment maybe we can keep also the previous method and wrap the customer old `Sampler` implementation in our own `Sampler<Span>` which delegates to the customer. Not sure but I think is doable. We can also `Deprecate` the old method.

mariusc83 (approved):

dd-sdk-android-core/api/apiSurface
ambushwork: will it make sense to make the API be like `fun sample(T? = null): Boolean` instead of `fun sample(T): Boolean`?

ambushwork (approved):

dd-sdk-android-core/api/apiSurface
0xnm: I believe we probably don't want to allow nullable types in the interface definition. But probably even if we do so, it would be still binary incompatible change, because then customer implementing custom `Sampler` still need to update the method signature.

My main concern is indeed in the interceptors usage (other places are pretty much internal), but if we ultimately cannot avoid this and we need to progress on this topic, I'm fine with this minor breaking change.

0xnm (approved):

PR #2392, opened 2024-11-14
RUM-7191 Implemented the basic logic for time-to-network-settle view metric

mariusc83:
### What does this PR do?

In this PR we are adding the logic for resolving the newly introduced `Time To Network Settled` out-of-the-box view loading metric.
**Start**: Begins with RUM ViewScope initialization.
**End**: Tied to the end of the last resource load that commenced in the initial 100ms window or defined by the given strategy in the configuration. In case the application went into background or a new view started by the time the initial resources finished loading we will not report this metric for that specific view. In general this could be phrased as:
in case the view was stopped before all the resources were loaded we will not report this metric.

```mermaid
gantt
    title Timeline for Time-to-Network-Settled
    dateFormat  HH:mm:ss
    section Time-to-Network-Settled
    View Becomes Visible: des1, 00:00:00, 00:00:00:100  
    Network request :done, des2, 00:00:01, 00:00:02
    Network request :done, des3, 00:00:01, 00:00:04
    Network request :done, des4, 00:00:03, 00:00:06
    100ms limit: after des4, 00:00:04
    Time-to-Network-Settled Interval :active, des5, 00:00:00, 00:00:06
```

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
@@ -11,7 +11,6 @@ import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.storage.DataWriter
 import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
-import com.datadog.android.core.internal.utils.loggableStackTrace
 import com.datadog.android.internal.utils.loggableStackTrace
 import com.datadog.android.rum.GlobalRumMonitor
 import com.datadog.android.rum.RumAttributes
@@ -22,6 +21,8 @@ import com.datadog.android.rum.internal.FeaturesContextResolver
 import com.datadog.android.rum.internal.domain.RumContext
 import com.datadog.android.rum.internal.domain.Time
 import com.datadog.android.rum.internal.domain.event.ResourceTiming
+import com.datadog.android.rum.internal.metric.networksettled.InternalResourceContext
+import com.datadog.android.rum.internal.metric.networksettled.NetworkSettledMetricResolver
 import com.datadog.android.rum.internal.monitor.StorageEvent
 import com.datadog.android.rum.internal.utils.hasUserData
 import com.datadog.android.rum.internal.utils.newRumEventWriteOperation
@@ -44,7 +45,8 @@ internal class RumResourceScope(
     serverTimeOffsetInMs: Long,
     internal val firstPartyHostHeaderTypeResolver: FirstPartyHostHeaderTypeResolver,
     private val featuresContextResolver: FeaturesContextResolver,
-    internal val sampleRate: Float
+    internal val sampleRate: Float,
+    internal val networkSettledMetricResolver: NetworkSettledMetricResolver
 ) : RumScope {
 
     internal val resourceId: String = UUID.randomUUID().toString()
@@ -64,6 +66,12 @@ internal class RumResourceScope(
     private var kind: RumResourceKind = RumResourceKind.UNKNOWN
     private var statusCode: Long? = null
     private var size: Long? = null
+
+    init {
+        networkSettledMetricResolver.resourceWasStarted(
+            InternalResourceContext(resourceId, eventTime.nanoTime)
+        )
+    }
 
     // region RumScope
 
@@ -101,7 +109,6 @@ internal class RumResourceScope(
         writer: DataWriter<Any>
     ) {
         if (key != event.key) return
-
         stopped = true
         attributes.putAll(event.attributes)
         kind = event.kind
@@ -119,7 +126,6 @@ internal class RumResourceScope(
         writer: DataWriter<Any>
     ) {
         if (key != event.key) return
-
         timing = event.timing
         if (stopped && !sent) {
             sendResource(kind, statusCode, size, event.eventTime, writer)
@@ -132,9 +138,8 @@ internal class RumResourceScope(
         writer: DataWriter<Any>
     ) {
         if (key != event.key) return
-
         attributes.putAll(event.attributes)
-
+        reportResourceStoppedMetric(event.eventTime.nanoTime)
         sendError(
             event.message,
             event.source,
@@ -152,7 +157,7 @@ internal class RumResourceScope(
         writer: DataWriter<Any>
     ) {
         if (key != event.key) return
-
+        reportResourceStoppedMetric(event.eventTime.nanoTime)
         attributes.putAll(event.attributes)
 
         val errorCategory =
@@ -169,6 +174,12 @@ internal class RumResourceScope(
         )
     }
 
+    private fun reportResourceStoppedMetric(eventTimeStamp: Long) {
+        networkSettledMetricResolver.resourceWasStopped(
+            InternalResourceContext(resourceId, eventTimeStamp)
+        )
+    }
+
     @Suppress("LongMethod")
     private fun sendResource(
         kind: RumResourceKind,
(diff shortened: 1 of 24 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/networksettled/ViewNetworkSettledMetric.kt
```diff
@@ -0,0 +1,130 @@
...
+
+internal class ViewNetworkSettledMetric(
+    requestIdentificationStrategy: NetworkSettledRequestIdentificationStrategy,
+    private val internalLogger: InternalLogger,
+    private val requestValidator: RequestValidator = resolveRequestValidator(requestIdentificationStrategy)
+) {
+
+    private val resourceStartedTimestamps = HashMap<String, HashMap<String, Long>>()
+    private val resourceStoppedTimestamps = HashMap<String, PriorityQueue<Long>>()
+    private val viewCreatedTimestampsByBViewId = HashMap<String, Long>()
```
ncreated, line 21: suggestion/ We need these maps because of using single instance of `VNSMetric` for the entire RUM application. If we instead create a distinct one for each RUM view scope there will be no need for using hash maps and likely the `purge()` routine will be gone.

Considering data flow and the domain of this problem (individual view metric), do we gain anything by sharing the same `VNSMetric` instance with all view scopes?

mariusc83: you are totally right...I think there's nothing we can win here at least in this case. For the other metric we will have to share it so I am wondering if I should do it differently for this one ? What do you think ? It makes sense to remove the complexity though.

mariusc83: One of the other reasons I wanted to share it was to not have to instantiate this all the time but I guess the memory penalty is not that high.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/networksettled/ViewNetworkSettledMetric.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+        purge()
+    }
+
+    fun resourceWasStopped(context: NetworkSettledResourceContext, eventTimestampInNanos: Long) {
+        // check if we have a start timestamp for this resource
+        val startTimestamp = resourceStartedTimestamps[context.viewId]?.get(context.resourceId)
+        if (startTimestamp != null) {
+            val resourceSettleDuration = eventTimestampInNanos - startTimestamp
+            resourceStoppedTimestamps.getOrPut(context.viewId) { PriorityQueue() }.add(resourceSettleDuration)
```
ncreated, line 48: question/ Do we need a `PriorityQueue()`? At the end, all we're looking for is the `eventTimestampInNanos` of the last stopped resource which can be stored on single property. As long as we can assume that events arrive in order, the same can be achieved without using any data structures:
```kotlin
private var numberOfPendingResources: Int = 0
private var timestampOfLastStoppedResourceNs: Long?

fun resourceWasStarted() {
   numberOfPendingResources += 1
}

fun resourceWasStopped(eventTimestampInNanos: Long) {
   timestampOfLastStoppedResourceNs = eventTimestampInNanos
   numberOfPendingResources -= 1
}
```

mariusc83: you need a `MaxHeap` there (in Java that's a PriorityQueue) in order to make sure you have everything ordered as you cannot guarantee the resources are going to be stopped in the same order. So you need to make sure that the last one is the max. In the end you do not need that if you prefer to do the `sort` when you compute and pick the highest one. For me I prefer this because it is a bit more efficient.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -65,6 +67,12 @@ internal class RumResourceScope(
     private var statusCode: Long? = null
     private var size: Long? = null
 
+    init {
+        networkSettledMetricResolver.resourceWasStarted(
+            NetworkSettledResourceContext(resourceId, eventTime.nanoTime)
+        )
+    }
```
xgouchet, line 74: > [!NOTE]
> Why are we splitting the computation in both the `RumViewScope` and `RumResourceScope`. All Resource related event do go through the `RumViewScope` first (see the `RumViewScope.onStartResource()` method), so it might be simpler to compute the network settled directly via the `RumViewScope`. It also might be simpler to compute it in the `RumViewScope` instead of a `NetworkSettledMetricResolver`, wdyt?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/networksettled/NetworkSettledRequestIdentificationStrategy.kt
```diff
@@ -0,0 +1,19 @@
...
+
+internal sealed class NetworkSettledRequestIdentificationStrategy {
+    internal class TimeIntervalBasedStrategy(
+        val timeThresholdNs: Long = TimeUnit.MILLISECONDS.toNanos(DEFAULT_TIME_THRESHOLD_MS)
+    ) : NetworkSettledRequestIdentificationStrategy() {
+        companion object {
+            internal const val DEFAULT_TIME_THRESHOLD_MS = 100L
+        }
+    }
+}
```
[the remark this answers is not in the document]
mariusc83: This is not the end of it, we can discuss a bit more about this if you want. There will be another strategy which will accept a configurable interface which will be the `RequestValidator` which is an interface. I did not expose yet this interface because this will be in another task. The reason why I like strategies is that they are exhaustive and also the client will be able to pick from a definite set. One strategy is based only on a time interval so the above definition doesn't match. The client will have to only define the `timeInterval` and that will be it he doesn't need to implement any interface. The default value is `100ms` as this is defined in the RFC. I guess you will see it better once we add the other tasks also.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -58,7 +59,11 @@ internal open class RumViewScope(
     private val featuresContextResolver: FeaturesContextResolver = FeaturesContextResolver(),
     internal val type: RumViewType = RumViewType.FOREGROUND,
     private val trackFrustrations: Boolean,
-    internal val sampleRate: Float
+    internal val sampleRate: Float,
+    private val networkSettledMetricResolver: NetworkSettledMetricResolver = NetworkSettledMetricResolver(
+        NetworkSettledRequestIdentificationStrategy.TimeIntervalBasedStrategy(),
+        sdkCore.internalLogger
+    )
```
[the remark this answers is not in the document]
mariusc83: Based on the RFC we need to provide a strategy with time interval and have default value to 100ms. We can either do this or just provide 2 strategies, one with a constant 100ms interval and the other with a configurable required interval in the constructor: `StaticTimeIntervalBasedStrategy`, `CustomTimeIntervalBasedStrategy` but I do think this kind of overkill. WDYT.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/networksettled/RequestValidator.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric.networksettled
+
+import com.datadog.tools.annotation.NoOpImplementation
+
+@NoOpImplementation
+internal interface RequestValidator {
+    fun validate(
```
xgouchet, line 13: > [!NOTE]
> Nitpick: maybe we should make the naming a bit more explicit about the purpose of this interface (and use consistent name for Resources: Request is not a word we use in RUM), e.g.: 

```suggestion
internal interface InitialResourceIdentifier {
    fun isInitialResource(
```

xgouchet (approved):

xgouchet (approved):

PR #2411, opened 2024-11-21
Improve the log message when exception happen during Upload

xgouchet:
### What does this PR do?

Improve the log message when exception happen during Upload. Right now we only print the error message but not the error type which makes it harder to investigate when an unknown error appears

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
@@ -117,6 +117,8 @@ internal sealed class UploadStatus(
 
             if (throwable != null) {
                 append(" (")
+                append(throwable.javaClass.name)
+                append(": ")
                 append(throwable.message)
                 append(")")
             }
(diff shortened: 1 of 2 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
```diff
@@ -117,6 +117,8 @@ internal sealed class UploadStatus(
 
             if (throwable != null) {
                 append(" (")
+                append(throwable.javaClass.name)
```
0xnm, line 120: the downside is that it will be obfuscated in the release builds, so it will come as obfuscated to our telemetry as well. But I guess the major use-case is still debugging the issue with the debug build.

0xnm (approved):

jonathanmos (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/UploadStatus.kt
xgouchet: It will only be obfuscated if it's a user or 3rd party lib exception. All the Java and Android framework excpetions will be in cleartext (the support case that sprung this was because of an ArrayIndexOutOfBounds issue)

PR #2417, opened 2024-11-28
RUM-7171 Implement the basic logic for interaction-to-next-view-metric 

mariusc83:
### What does this PR do?

In this PR we are adding the logic to compute and provide the `interaction-to-next-view` time. 

Captures the time interval in nano seconds from the last user interaction on the previous view to when the current view becomes visible. The last user interaction needs to match some criteria that will
be detailed below where we explain how this metric is computed. 

Is Directly associated with the lifecycle of the view: the end part of this interval is
equal with the moment when the new view scope creation but under the hood this is triggered by a system tied to the internal view lifecycle events. More details below in the computation explanation.

```mermaid
gantt
    title Timeline for Interaction-to-Next-View
    dateFormat  HH:mm:ss

    section Interaction-to-Next-View
    User Interaction: des1, 00:00:00, 00:00:00:100  
    View Is Created :done, des2, 00:00:01, 00:00:02
    View Becomes Visible :done, des3, after des2, 00:00:03
    Interaction-to-Next-View Interval :active, des3, 00:00:00, 00:00:03

```
### Motivation

### Additional Notes


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    @VisibleForTesting
+    internal fun lastViewCreatedTimestamps(): Map<String, Long> {
+        return lastViewCreatedTimestamps
+    }
+
+    companion object {
+        // we need to keep at least 10 entries to be able to calculate the metric for consecutive views that
+        // are going to be created almost in the same time and will rely on the previous view interaction
+        private const val MAX_ENTRIES = 10
```
ncreated, line 114: question/ What are the "views created almost in the same time"? Does it mean very short navigation through series of views OR a problem with instrumentation?

mariusc83: yes...also for views that are already stopped but still request the metric for some view updates. I also want to prepare it for 2 active views later, this logic will support that scenario also.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
```diff
@@ -0,0 +1,116 @@
...
+                    }
+                )
+                return null
+            }
+        }
+        internalLogger.log(
+            InternalLogger.Level.WARN,
+            InternalLogger.Target.MAINTAINER,
+            { "[ViewNetworkSettledMetric] No previous interaction found for this viewId:$viewId" }
+        )
```
ncreated, line 66: question/ Won't we always end up here for the initial view (`ApplicationLaunch`)? If so, this log won't be actionable.

mariusc83: totally true....I guess we can condition this log by checking if there's any interaction yet added in the map, if not this is the applaunch. Good point. Actually added a test about this but never thought to condition the log.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+        return null
+    }
+
+    private fun purgeOldEntries() {
+        while (lastInteractions.entries.size > MAX_ENTRIES) {
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            // we make sure the collection is never empty
+            lastInteractions.entries.remove(lastInteractions.entries.first())
```
ambushwork, line 92: what's the reason behind removing the first entry of a hashmap?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
ambushwork, line 114: I don't quite understand why "we need to keep at least 10 entries", can you elaborate? for example if we keep less than 10 entries, what's the consequence?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
mariusc83: we are purging the data basically making sure we are only keeping the last (newly added) values.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
mariusc83: Yeah like I answered to @ncreated question above is because we can always have `inactive` views that are still asking for the metric so we need to handle multiple views here. Also later we might need to handle `multiple active` views. I think this number is too big though so I will lower it to 3 entries for now.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
ambushwork: OK, I see it's a `LinkedHashMap` so it makes sense

PR #2423, opened 2024-11-29
RUM-7489 Handle readOnly additionalProperties

xgouchet:
### What does this PR do?

Make sure that the `readOnly` setting on additional properties is taken into account when converting Json Schema into a Kotlin data class.

### DIFF
--- a/features/dd-sdk-android-rum/api/apiSurface
+++ b/features/dd-sdk-android-rum/api/apiSurface
@@ -1490,7 +1490,7 @@ data class com.datadog.android.rum.model.ViewEvent
       fun fromJson(kotlin.String): Privacy
       fun fromJsonObject(com.google.gson.JsonObject): Privacy
   data class CustomTimings
-    constructor(kotlin.collections.MutableMap<kotlin.String, kotlin.Long> = mutableMapOf())
+    constructor(kotlin.collections.Map<kotlin.String, kotlin.Long> = mapOf())
     fun toJson(): com.google.gson.JsonElement
     companion object 
       fun fromJson(kotlin.String): CustomTimings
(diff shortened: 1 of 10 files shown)


features/dd-sdk-android-rum/api/apiSurface
0xnm, line 1: I was expected to see one more change in this file: for `ViewEvent.featureFlags`. It also has `additionalProperties: { readOnly: true }`, but it seems because it is re-using `ViewEvent.Context` class, this property is ignored.

In theory, we shouldn't have class re-use here, because `TypeDefinition.additionalProperties` type was updated and it is participating in `matches` method, but I guess we still have it?

xgouchet: Actually the `featureFlags` property is defined as follow, which means the featureFlag property is read only, but its `additionalProperties` is not by default. 

```json 
{
        "feature_flags": {
          "type": "object",
          "description": "Feature flags properties",
          "additionalProperties": true,
          "readOnly": true
        }
}
```

Compare that to the `customTiming` property, which is defined as follow, where indeed the `additionalProperties` itself is required to be `readOnly`:

``` 
{
            "custom_timings": {
              "type": "object",
              "description": "User custom timings of the view. As timing name is used as facet path, it must contain only letters, digits, or the characters - _ . @ $",
              "additionalProperties": {
                "type": "integer",
                "minimum": 0,
                "readOnly": true
              },
              "readOnly": true
            }
}
```

0xnm: yes, I misread type definition, sorry.

0xnm (approved):

mariusc83 (approved):

PR #2429, opened 2024-12-03
Notify correctly the itnv metric resolver for last action in RumViewScope

mariusc83:
### What does this PR do?

The interval we consider for the `interaction-to-next-view` metric is the moment the last action finished on the previous view until the moment the new view was created. Based on the last discussion we decided to completely rely on the RUM events timestamps and mappers and only report an action as last action once it was successfully persisted. We also change the way we resolve the timestamps relying on the RUM View and Action events corrected timestamps using the NTP server.

### Motivation

### Additional Notes


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -382,13 +398,15 @@ internal open class RumViewScope(
 
         if (stopped) return
 
-        interactionToNextViewMetricResolver.onActionSent(
-            InternalInteractionContext(
-                viewId,
-                event.type,
-                event.eventTime.nanoTime
+        if (!event.waitForStop) {
+            interactionToNextViewMetricResolver.onActionSent(
```
ncreated, line 402: What about the `ACTION_DROPPED_WARNING` that may occur next in this function?

> RUM Action (%s on %s) was dropped because another action is still active for the same view.

According to RUM rules, if a new action occurs within 100ms after the previous one, the new action is ignored.

However, the way we update the ITNV here seems to disregard this rule. The metric is being updated for every new action, even if the action is skipped due to debouncing. This is problematic, as ITNV could eventually point to a RUM action that does not exist.

To illustrate, if actions A2 and A3 follow A1 immediately, both A2 and A3 will be skipped (not sent). With the current implementation, the next view’s ITNV will be calculated from A3 (`ITNV2(B)`), but A3 will not exist in the RUM. What we expect is `ITNV1(B)`, which should span from the existing action (A1):
```
[===== view A ========][======== view B =========]

  [-A1-]✅             |
    [-A2-]❌           |
      [-A3-]❌         |

-------------------------------------------------->

   ^---- ITNV1(B) -----^
       ^-- ITNV2(B) ---^ 
```

mariusc83: You are right but in the same time it might be a problem if we are following these rules. Normally should be the last action that follows to the next view and we should not care about our internal rules to send RUM actions. What do you think ? I was thinking to not add the RUM action rules in this equation as these are only to make sure we dispatch action events reliably.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -372,6 +373,21 @@ internal open class RumViewScope(
...
+    @WorkerThread
+    private fun onStopAction(event: RumRawEvent.StopAction, writer: DataWriter<Any>) {
+        delegateEventToChildren(event, writer)
+        if (stopped) return
+        if (event.type != null) {
+            interactionToNextViewMetricResolver.onActionSent(
+                InternalInteractionContext(
+                    viewId,
+                    event.type,
+                    event.eventTime.nanoTime
```
ncreated, line 385: clarification/ The `interactionToNextViewMetricResolver` is updated from both `onStopAction()` and `onStartAction()`, and we’re using event.eventTime in both cases. This gives the impression that we're treating "action start" and "action end" interchangeably, depending on the action type 🤔.

What we need is consistency in ITNV computation, specifically in determining the T1 moment in the following diagram. It should always represent the end of the action, regardless of the action type (tap, swipe, back, click).

Screenshot 2024-12-04 at 09 26 02

mariusc83: yes the confusion comes from the fact that we are using `StartAction` with `waitForStop:false` to send `TAP` events.

ncreated: Got it. So this is equal to "stop action" 👍.

mariusc83: yes for `Tap` events this is how we report it directly through the `StartAction`. I admit it is confusing.

PR #2430, opened 2024-12-04
[RUM-3845] Trying to get from Thread.sleep(SHORT_SLEEP_MS) in unit tests at DatadogEventListenerTest

satween:
### What does this PR do?

In this PR we are trying to get rid of `Thread.sleep `instructions inside `DatadogEventListenerTest.kt`.
To do so we use `sdkCore.time.deviceTimeNs` at DatadogEventListener.kt instead of `System.nanoTime()` method which let us mock time values 

### Motivation

This approach makes this test independent from CI machine time and should increase test stability

### Additional Notes

This is a WIP PR so changes could be non-final. 

### DIFF
--- a/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogEventListener.kt
+++ b/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogEventListener.kt
@@ -74,27 +74,27 @@ internal constructor(
     override fun callStart(call: Call) {
         super.callStart(call)
         sendWaitForResourceTimingEvent()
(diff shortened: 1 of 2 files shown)


0xnm (approved):
nice change!

side note: while the ticket itself mentions only a single class, I guess its point is to do similar change in other classes as well, mainly in the RUM scopes related-classes and their tests.

satween:
> nice change!
> 
> side note: while the ticket itself mentions only a single class, I guess its point is to do similar change in other classes as well, mainly in the RUM scopes related-classes and their tests; so the title of this PR can be a bit misleading.

Thanks, yep, planning to make similar changes in other places as well. 
PR title adjusted.

PR #2433, opened 2024-12-05
RUM-6394: Changing logApiUsage method signature - making event parameter computation lazy

satween:
### What does this PR do?

Changing signature of `logApiUsage` method of `InternalLogger` a bit - replacing `InternalTelemetryEvent.ApiUsage` event parameter with lambda function that will called only if sampling happens. This should reduce useless amount of work made by sdk and reduce impact to the customer's application resources consumption

### Motivation

Some times running `logApiUsage` is useless because actual logging wont happened because it not sampled. But in such cases parameter computation still happened so we will waste device's resources for the useless work. To avoid that we use lambda instead

### Additional Notes


ambushwork (approved):
Pipeline to fix, otherwise LGTM!

PR #2438, opened 2024-12-06
RUM-7527: Remove unused Datadog GlobalTracer class

ambushwork:
### What does this PR do?

Remove unused Datadog `GlobalTracer` class to avoid confusion with opentracing `GlobalTracer`

### Motivation

RUM-7527

### DIFF
--- a/features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/GlobalTracer.java
+++ b/features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/GlobalTracer.java
@@ -5,12 +5,21 @@ import com.datadog.trace.api.internal.InternalTracer;
 import java.util.ArrayList;
 import java.util.Collection;
 
+import kotlin.Deprecated;
+import kotlin.DeprecationLevel;
+import kotlin.ReplaceWith;
+
 /**
  * A global reference to the registered Datadog tracer.
  *
  * <p>OpenTracing's GlobalTracer cannot be cast to its DDTracer implementation, so this class exists
  * to provide a global window to datadog-specific features.
  */
+@Deprecated(
+        message = "Use [io.opentracing.util.GlobalTracer] instead, this class will be removed soon",
+        replaceWith = @ReplaceWith(expression = "GlobalTracer", imports = "io.opentracing.util"),
+        level = DeprecationLevel.ERROR
+)
 public class GlobalTracer {
   private static final Tracer NO_OP =
       new Tracer() {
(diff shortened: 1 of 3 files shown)


mariusc83 (approved):

0xnm (commented):
I'm not sure we can remove it, because even though it is not used for the means of the Android SDK, some users _may_ rely on this as an exported symbol for their usage with `com.datadog.trace.api.Tracer` (this would be strange though).

Maybe it is better to deprecate it instead?

mariusc83:
> I'm not sure we can remove it, because even though it is not used for the means of the Android SDK, some users _may_ rely on this as an exported symbol for their usage with `com.datadog.trace.api.Tracer` (this would be strange though).
> 
> Maybe it is better to deprecate it instead?

I really want to avoid the confusion here, I would say removing is better as there should not be any usage of the `CoreTracer` directly. This is our internal tracer and it is the only one that implements the `com.datadog.trace.api.Tracer`. If someone uses the `CoreTracer` directly then it will be a mistake.

0xnm:
Yes, for sure, its usage is a misuse, but at the same time it is binary-incompatible change and we can do it only for the next major release for the public-facing API (we cannot say that there is no anyone using this symbol for some strange needs).

At the same time `java.lang.Deprecated` doesn't allow to supply the message, but `kotlin.Deprecated` allows to do it (maybe we can use it in the Java code?).

But anyway, it is your call.

xgouchet:
I would tend to agree with @0xnm here: removing it might break using for some of our customers. 
Adding a Deprecated annotation (and adding javadoc to explain why) seems like a better first step. 
Awe could additionally add some telemetry to track the usage, and if we are sure no one uses it, we can then delete it.

jonathanmos (approved):

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/GlobalTracer.java
```diff
@@ -5,12 +5,19 @@
...
 /**
  * A global reference to the registered Datadog tracer.
  *
  * <p>OpenTracing's GlobalTracer cannot be cast to its DDTracer implementation, so this class exists
  * to provide a global window to datadog-specific features.
  */
+@Deprecated(
+        message = "Use [io.opentracing.util.GlobalTracer] instead, this class will be removed soon",
+        level = DeprecationLevel.ERROR
+)
```
0xnm, line 20: you can also utilize `replaceWith` argument.

0xnm (approved):
Lgtm. I added one suggestion and probably PR title should be updated (we are not removing anything yet).

features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/GlobalTracer.java
ambushwork: updated

PR #2439, opened 2024-12-06
RUM-7590 Correct the way we register the initial resources for TTNS metric

mariusc83:
…etric

### What does this PR do?

### Motivation

### Additional Notes


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -303,10 +302,10 @@ internal class RumResourceScope(
             )
         }
             .onError {
```
mariusc83, line 304: I need to do the same thing also for when the resource was stopped due to an error :(

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/networksettled/NetworkSettledMetricResolverTest.kt
```diff
@@ -252,6 +252,33 @@ internal class NetworkSettledMetricResolverTest {
         assertThat(afterViewStoppedMetricValue).isEqualTo(expectedMetricValue)
     }
 
+    @Test
+    fun `M skip last value W resolveMetric(){ view was stopped, last value was drop }`(forge: Forge) {
```
ambushwork, line 256: "last value was drop" -> "last value was dropped". And the test function seems related with "resource stopped", not "view stopped" if I am not mistaken

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/networksettled/NetworkSettledMetricResolverTest.kt
```diff
@@ -298,6 +325,45 @@ internal class NetworkSettledMetricResolverTest {
...
+        val startTimestamps = forge.forgeStartTimestamps()
+        val fakeResourcesIds = forge.aList(size = startTimestamps.size) {
+            forge.getForgery<UUID>().toString()
+        }
+        fakeResourcesIds.forEachIndexed { index, id ->
+            testedMetric.resourceWasStarted(
+                InternalResourceContext(id, startTimestamps[index])
+            )
+        }
+
```
ambushwork, line 359: is it better to assert that `testedMetric.getResourceStartedCacheSize()` is equal to the size of the list before the view is stopped to avoid false positive of next assertion? or is there already another test case covers it?

ambushwork (approved):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/networksettled/NetworkSettledMetricResolverTest.kt
mariusc83: not sure here...when the view is stopped all the resources should be cleaned as we no longer need them in computation. The default value of that cache should always be `0` when cleared.

ambushwork: I mean **before** view is stopped , what's the expected value of `getResourceStartedCacheSize`? is it also zero?

mariusc83: nope...that is not 0...we could assert that also...let me add it.

ambushwork (approved):

PR #2453, opened 2024-12-12
[Rum-6852]  Add "Effective Sample Rate" to telemetry events

satween:
### What does this PR do?

* Adds support for effective sample rate field
* Supporting unit tests for it 
* Supporting long-living events that could have two sample rates 
* Small improvements with annotations 

### Motivation

Main goal is to be able to see what exact value was applied through different sampling layer


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/logger/SdkInternalLogger.kt
```diff
@@ -253,6 +264,22 @@ internal class SdkInternalLogger(
         }
     }
 
+    private fun Map<String, Any?>?.putIfNotExistOrMapIsNull(key: String, value: Float?): MutableMap<String, Any?> {
```
mariusc83, line 267: **🎯 suggestion:** ‏ Usually we prefer not to use extension functions if they're only needed in one class. It is more performant as extension function tend to create more object at the bytecode level.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
```diff
@@ -477,6 +482,171 @@ internal class SdkInternalLoggerTest {
         }
     }
 
+    @ParameterizedTest
+    @ValueSource(floats = [0.0f, 0.3f])
```
ambushwork, line 486: you can use `forge` to generate random float

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/logger/SdkInternalLogger.kt
```diff
@@ -253,6 +264,22 @@ internal class SdkInternalLogger(
...
+        val mutableMap = this?.toMutableMap() ?: mutableMapOf()
+        return mutableMap.apply { putIfNotExist(key, value ?: return@apply) }
+    }
+
+    // This method same as putIfAbsent but putIfAbsent requires SDK version >= 24
+    @Suppress("UnsafeThirdPartyFunctionCall")
```
mariusc83, line 273: **🎯 suggestion:** ‏ I think we could move this extension method in the `dd-sdk-internal` module in a `MapExt.kt` file if you really think we will need it throughout the project otherwise I really don't recommend to have an extension function only for this case.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/MethodCalledTelemetryTest.kt
```diff
@@ -82,6 +82,7 @@ internal class MethodCalledTelemetryTest {
     lateinit var fakeOperationName: String
 
     private var fakeStartTime: Long = 0
+    private var fakeCreationSampleRate: Float = 20f
```
ambushwork, line 85: shouldn't we use a random value for this?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTelemetryEvent.kt
```diff
@@ -56,4 +59,21 @@ sealed class InternalTelemetryEvent {
...
+
+    companion object {
+
+        /* The sampling rate is used when creating metrics.
+         * Creating(head) sampling rate exist only for long-lived metrics like method performance.
+         * Creating(head) sampling rate is responsible for metric creation.
+         * Created metric still could not be sent.
+         * Sending depends on [REPORTING_SAMPLING_RATE_KEY] sample
+         */
+        const val CREATION_SAMPLING_RATE_KEY: String = "HEAD_SAMPLING_RATE_KEY"
```
mariusc83, line 71: **❔ question:** ‏ I really do not understand the need of this sampling rate. I think I will need to read more about it, let's talk in private.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
```diff
@@ -107,13 +108,16 @@ interface InternalLogger {
      * @param messageBuilder the lambda building the metric message
      * @param additionalProperties additional properties to add to the metric
      * @param samplingRate value between 0-100 for sampling the event. Note that the sampling rate applied to this
+     * @param creationSampleRate value between 0-100 that was applied for long-lived metrics. This value will not be
```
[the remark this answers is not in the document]
mariusc83: I subscribe here, I am not sure why you need this and what this should be

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
```diff
@@ -439,6 +466,37 @@ internal class TelemetryEventHandler(
...
+            ?.getFloat(InternalTelemetryEvent.CREATION_SAMPLING_RATE_KEY)
+            ?.percent() ?: 1.0
+
+        val reportingSamplingRate = properties
+            ?.getFloat(InternalTelemetryEvent.REPORTING_SAMPLING_RATE_KEY)
+            ?.percent() ?: 1.0
+
+        val eventSamplingRate = eventSpecificSamplingRate?.percent() ?: 1.0
+
+        val sessionSampleRate = rumConfig?.sampleRate?.percent() ?: 1.0
```
mariusc83, line 485: **❔ question:** ‏ why do we report this here ? Is this being used to sample telemetry events ?

mariusc83: **🔨 issue:** ‏ I just double checked in the sdk so `sessionSampleRate` it is not and it should not affect the telemetry event sampling. We need to fix this.

mariusc83 (changes requested):
We need to remove the `sessionSampleRate` from `effectiveSamplingRate` equation.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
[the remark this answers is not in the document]
satween: Make sense, updated, hope it's more clear now

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/logger/SdkInternalLogger.kt
```diff
@@ -253,6 +264,22 @@ internal class SdkInternalLogger(
         }
     }
 
+    private fun Map<String, Any?>?.putIfNotExistOrMapIsNull(key: String, value: Float?): MutableMap<String, Any?> {
+        val mutableMap = this?.toMutableMap() ?: mutableMapOf()
+        return mutableMap.apply { putIfNotExist(key, value ?: return@apply) }
```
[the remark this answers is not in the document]
satween: Made it even simpler, now there is only one method used. `toMap` is not quite required here IMO, because all other parameter types requires `Map` which implemented by MutableMap as well

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
satween: The goal here is to check null and non null parameters. (I pass null in case if creationSampleRate == 0.0) and do not copy-paste whole test for specific value, not sure `forge` is applicable here

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt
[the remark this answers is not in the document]
satween: Yeah, adjusted 👍

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/logger/SdkInternalLogger.kt
satween: Rewritten to classic fun, thanks 🙏

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/logger/SdkInternalLogger.kt
satween: It's not worth it, refactored to simpler one

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTelemetryEvent.kt
satween: Adusted doc a bit.

Overall, the reason why we need two sample rate is: 

> We have some metrics which sampling happens not from the place we call `logMetric`  but when we create such metric. For instance - `MethodCalledTelemetry`. It will be sampled at `SdkInternalLogger.startPerformanceMeasure()` but sent with `MethodCalledTelemetry.stopAndSend`. So the sampling rate for it will equal to `samplingRate` of `startPerformanceMeasure` method. 

Nevertheless there is a possibility to still sample metric at the `MethodCalledTelemetry.stopAndSend` so to compute effectiveSampleRate we need both of this values composed together

satween:
> We need to remove the `sessionSampleRate` from `effectiveSamplingRate` equation.

done

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt
satween: yep, removed from computation

ambushwork (approved):

PR #2454, opened 2024-12-13
fix: Don't warn about missing view on PerformanceMetric events

fuzzybinary:
### What does this PR do?

PerformanceMetric events are triggered automatically by some of the cross platform frameworks, and can result in log spam as views transition or if no view is active.  Fix it so that the performance metric event is considered a silent orphaned event type.

refs: RUM-6138


0xnm (commented):
What is the case when it is sent while there is no active view?

xgouchet (approved):

fuzzybinary:
> What is the case when it is sent while there is no active view?

There are a few:
* Transitions between views. Our router plugin stops views when they're pop'd and I've seen instances where there can be a frame or two sometimes when a new view isn't started yet so Flutter sends the rendering information even though there's no view.
* Web views have the same problem as above, but worse. We kill the current Flutter view while loading the browser, but don't get a new view until the Browser SDK tells us a new view is loaded.
* A bug or intentionally not tracked view in user code. But because these events are sent with almost every frame render and because they're sent automatically, this just becomes log spam.

PR #2456, opened 2024-12-16
RUM-6286: replacing `joinToString` when it possible, refactor a bit

satween:
### What does this PR do?

Here we're trying to reduce usages of joinToString method inside the project, replacing it with more performant string builder 

### Motivation

Using `joinToString`  may cause a lot of memory and cpu consumption which we trying to avoid, minimizing impact of sdk to user's code 

### Additional Notes


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,61 @@
...
+        val initialData = "something inside buffer already"
+        val buffer = StringBuilder(initialData)
+
+        // When
+        buffer.appendIfNotEmpty(' ')
+
+        // Then
+
+        assertThat(buffer.toString()).isEqualTo("$initialData ")
+    }
```
xgouchet, line 20: > [!WARNING]
> When you test this, even though the implementation is trivial, you're only ever testing that it works for `"something inside buffer already"`. 
> Instead you might want to use a property based approach : create a random non-empty string (we use the regex feature to ensure that there's at least one character, of any kind) and use that as input.

```suggestion
        @Test
    fun `M add char W addIfNotEmpty {buffer is not empty}`(
        @StringForgery(regex = ".+") initialContent: String
    ) {
        // Given
        val buffer = StringBuilder(initialContent)

        // When
        buffer.appendIfNotEmpty(' ')

        // Then
        assertThat(buffer.toString()).isEqualTo("$initialContent ")
    }
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    fun `M add str W addIfNotEmpty {buffer is not empty}`() {
```
xgouchet, line 23: > [!NOTE]
> Same as above applies here

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/PerformanceTest.kt
```diff
@@ -0,0 +1,64 @@
...
+                        itemsForJoin.forEachIndexed { i, item ->
+                            if (i > 0) append(" ")
+                            append(item)
+                        }
+                    }
+                }
+            )
+        }
+
+        assertThat(builderResult).isEqualTo(jointToStringResult) // same result
```
xgouchet, line 61: > [!WARNING]
> You're only making that assertion on the last iteration of your test. You could move this inside the loop to ensure every iteration behaves correctly.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/PerformanceTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    ExtendWith(ForgeExtension::class)
+)
+@ForgeConfiguration(Configurator::class)
+internal class PerformanceTest {
+
+    private fun List<Long>.mean() = sum().toDouble() / size
+
+    @Test
+    fun `M be faster than joinToString W buildString`(forge: Forge) {
+        val itemsForJoin = forge.aList(ITEMS_TO_JOINT) { forge.aString() }
```
xgouchet, line 31: > [!NOTE]
> You're running all the iterations on the same dataset. On the one hand, it gives a more relevant mean, but I'm curious whether we could get more insights if we moved the `itemsForJoin` inside the loop and have a different data set each time. 
> Also we're looking only at the mean, what about other insights (p90, stdev, etc…), can we learn info about edge cases from those?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
satween: good suggestion, applied

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
satween: 👍

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/PerformanceTest.kt
satween: done, thanks

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/PerformanceTest.kt
satween: Moving `itemsForJoin` creation inside the loop will significantly increase test execution time, so I left it outside for now. 

Regarding the p95 - it's a good point. After some research I've noticed that the STD value is relatively big, so mean is not the best metric for this measurement. I've changed  metric to p95 as it seems good compromise between accuracy and sustainability to flakiness. Still test is highly depends on the machine state, so if we will notice that it flaky will reduce metric to p90

satween: Upd: reduced to p90 for more stability

xgouchet (approved):

dd-sdk-android-core/api/apiSurface
```diff
@@ -190,6 +190,8 @@ class com.datadog.android.core.SdkReference
   fun get(): com.datadog.android.api.SdkCore?
 fun <T> allowThreadDiskReads(() -> T): T
 fun <T> allowThreadDiskWrites(() -> T): T
+fun StringBuilder.appendIfNotEmpty(String)
+fun StringBuilder.appendIfNotEmpty(Char)
```
0xnm, line 194: These functions should be in the `dd-sdk-android-internal` module instead. The problem here is that `core` module is re-exported to the client, so all the symbols here are visible to the client, but we don't want to pollute (make these symbols to appear in the suggestions dropdown) namespace of methods for classes from stdlib, this is only for our usage.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
```diff
@@ -0,0 +1,27 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.core
+
+/**
+ * This utility function helps replace joinToString calls with more efficient StringBuilder's methods calls.
+ * In case when content should be separated with some separator it's handy to add it in front of new string only
+ * if buffer already contain some data
```
0xnm, line 12:
```suggestion
 * In case when content should be separated with some separator, it's handy to add it in front of a new string only
 * if buffer already contains some data.
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
```diff
@@ -0,0 +1,27 @@
...
+ * @param str - string that should be added into buffer only if it already contains some data
```
0xnm, line 13:
```suggestion
 * @param str string that should be added to the buffer only if it already contains some data.
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
```diff
@@ -0,0 +1,27 @@
...
+fun StringBuilder.appendIfNotEmpty(str: String) = apply {
+    if (isNotEmpty()) append(str)
+}
+
+/**
+ * This utility function helps replace joinToString calls with more efficient StringBuilder's methods calls.
+ * In case when content should be separated with some separator it's handy to add it in front of new string only
+ * if buffer already contain some data
+ * @param char - char that should be added into buffer only if it already contains some data
+ */
```
0xnm, line 24:
```suggestion
/**
 * This utility function helps to replace [joinToString] calls with more efficient [StringBuilder] methods calls.
 * In case when content should be separated with some separator, it's handy to add it in front of a new string only
 * if buffer already contains some data.
 * @param char char that should be added to the buffer only if it already contains some data.
 */
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/error/internal/DatadogExceptionHandler.kt
```diff
@@ -133,12 +133,11 @@ internal class DatadogExceptionHandler(
...
             .filterKeys { it != crashedThread }
             .filterValues { it.isNotEmpty() }
-            .map {
-                val thread = it.key
+            .map { (thread, stackTrace) ->
                 ThreadDump(
                     name = thread.name,
                     state = thread.state.asString(),
-                    stack = it.value.loggableStackTrace(),
+                    stack = stackTrace.loggableStackTrace(),
```
0xnm, line 140: seems not related with the change done in the PR? just a small side improvement?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.jupiter.api.extension.Extensions
+
+@Extensions(
+    ExtendWith(ForgeExtension::class)
+)
+internal class StringBuilderExtKtTest {
+
+    @Test
+    fun `M add char W addIfNotEmpty {buffer is not empty}`(
+        @StringForgery(regex = ".+") initialContent: String
```
0xnm, line 17: probably regex is not needed here and any string is good? In this case you can leave only `@StringForgery`, without explicit argument? it will always generate non-empty string

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package com.datadog.android.core
+
+/**
+ * This utility function helps replace joinToString calls with more efficient StringBuilder's methods calls.
+ * In case when content should be separated with some separator it's handy to add it in front of new string only
+ * if buffer already contain some data
+ * @param str - string that should be added into buffer only if it already contains some data
+ */
+fun StringBuilder.appendIfNotEmpty(str: String) = apply {
```
0xnm, line 15: side thought: naming can be a bit confusing, because without looking at the method description I initially thought that it checks argument for being empty, not receiver.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    ) {
+        // Given
+        val buffer = StringBuilder(initialContent)
+
+        // When
+        buffer.appendIfNotEmpty(' ')
+
+        // Then
+
+        assertThat(buffer.toString()).isEqualTo("$initialContent ")
```
0xnm, line 27:
```suggestion
        // Then
        assertThat(buffer.toString()).isEqualTo("$initialContent ")
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    fun `M add str W addIfNotEmpty {buffer is not empty}`(
```
0xnm, line 31:
```suggestion
    fun `M add string W addIfNotEmpty {buffer is not empty}`(
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    ) {
+        // Given
+        val buffer = StringBuilder(initialContent)
+
+        // When
+        buffer.appendIfNotEmpty(" ")
+
+        // Then
+
+        assertThat(buffer.toString()).isEqualTo("$initialContent ")
```
0xnm, line 42:
```suggestion
        // Then
        assertThat(buffer.toString()).isEqualTo("$initialContent ")
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    fun `M add char W addIfNotEmpty {buffer is not empty}`(
+        @StringForgery(regex = ".+") initialContent: String
+    ) {
+        // Given
+        val buffer = StringBuilder(initialContent)
+
+        // When
+        buffer.appendIfNotEmpty(' ')
```
0xnm, line 23: nit: ideally you should append a random char, not test it only with `' '` (same for other tests).

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        // When
+        buffer.appendIfNotEmpty(" ")
+
+        // Then
+
+        assertThat(buffer.toString()).isEqualTo("$initialContent ")
+    }
+
+    @Test
+    fun `M not add any char W addIfNotEmpty {buffer is empty}`() {
```
0xnm, line 46: nit: test name contradicts with test execution - it is always `' '` char added here, not any.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        // Given
+        val buffer = StringBuilder()
+
+        // When
+        buffer.appendIfNotEmpty(' ')
+
+        // Then
+
+        assertThat(buffer.toString()).isEqualTo("")
```
0xnm, line 55:
```suggestion
        // Then
        assertThat(buffer.toString()).isEmpty()
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun `M not add any str W addIfNotEmpty {buffer is empty}`() {
+        // Given
+        val buffer = StringBuilder()
+
+        // When
+        buffer.appendIfNotEmpty(" ")
+
+        // Then
+
+        assertThat(buffer.toString()).isEqualTo("")
```
0xnm, line 68:
```suggestion
        // Then
        assertThat(buffer.toString()).isEmpty()
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
```diff
@@ -0,0 +1,111 @@
...
+import kotlin.math.sqrt
+import kotlin.system.measureNanoTime
+
+private const val ITEMS_TO_JOINT = 10_000
+private const val REPETITION_COUNT = 10_000
+
+@Extensions(
+    ExtendWith(ForgeExtension::class)
+)
+@ForgeConfiguration(value = Configurator::class, seed = 0x50c72968d123L)
```
0xnm, line 28: seed has to be removed, probably it is left after the debugging

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
```diff
@@ -0,0 +1,111 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import kotlin.math.pow
+import kotlin.math.round
+import kotlin.math.sqrt
+import kotlin.system.measureNanoTime
+
+private const val ITEMS_TO_JOINT = 10_000
```
0xnm, line 22:
```suggestion
private const val ITEMS_TO_JOIN = 10_000
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
```diff
@@ -0,0 +1,111 @@
...
+private const val REPETITION_COUNT = 10_000
```
0xnm, line 23: we usually wrap constants in the test in the nested companion object at the bottom of the test class, so that everyone knows what is the layout.

detekt_custom.yml
```diff
@@ -1245,6 +1245,7 @@ datadog:
       - "kotlin.text.StringBuilder()"
       - "kotlin.text.buildString(kotlin.Function1)"
       - "kotlin.text.buildString(kotlin.Int, kotlin.Function1)"
+      - "java.lang.StringBuilder.clear()"
```
0xnm, line 1248: this has to be moved to the right position in this file: it doesn't belong to the `Kotlin String` region and once proper region found, the position should be in the alphabetical order.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
```diff
@@ -0,0 +1,111 @@
...
+                " mean = ${jointToStringExecutionTime.mean}\n" +
+                " std = ${jointToStringExecutionTime.std}\n" +
+                " cv = ${"%.2f".format(jointToStringExecutionTime.cv)}%\n" +
+                " p50 = ${jointToStringExecutionTime.percentile(50)},\n" +
+                " p90 = ${jointToStringExecutionTime.percentile(90)},\n" +
+                " p95 = ${jointToStringExecutionTime.percentile(95)},\n" +
+                " p99 = ${jointToStringExecutionTime.percentile(99)}\n"
+            )
+
+        println(statisticsReport)
```
0xnm, line 82: debugging leftover?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
```diff
@@ -0,0 +1,111 @@
...
+@Extensions(
+    ExtendWith(ForgeExtension::class)
+)
+@ForgeConfiguration(value = Configurator::class, seed = 0x50c72968d123L)
+internal class JointToStringVsStringBuilderPerformanceTest {
+
+    @Test
+    fun `M be faster than joinToString W buildString`(forge: Forge) {
+        val itemsForJoin = forge.aList(ITEMS_TO_JOINT) { forge.aString() }
+        val jointToStringExecutionTime = mutableListOf<Long>()
```
0xnm, line 34:
```suggestion
        val joinToStringExecutionTime = mutableListOf<Long>()
```

dd-sdk-android-core/api/apiSurface
satween: moved, thanks for clarifications

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
satween: 👍

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
satween: 👍

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
satween: 👍

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
satween: 👍

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
satween: done, will do next time right way

detekt_custom.yml
satween: I thought that if it's extension function it should be in kotlin space, anyway, moved

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/JointToStringVsStringBuilderPerformanceTest.kt
satween: The idea was to print this information, but without failed test nobody will read it. Changed to `withFailMessage` instruction

detekt_custom.yml
0xnm: Then it should be `kotlin.text.StringBuilder.clear()` here.

satween: `java.lang.StringBuilder.clear()` is how detekt sees it. Probably there is some issue in `UnsafeThirdPartyFunctionCall` but adding `kotlin.text.StringBuilder.clear()` to whitelist will not help.

dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+
+@Extensions(
+    ExtendWith(ForgeExtension::class)
+)
+internal class StringBuilderExtKtTest {
+
+    @Test
+    fun `M add char W addIfNotEmpty {buffer is not empty}`(
```
0xnm, line 23: nit: the method name is actually a bit different (same for other tests in this file)

```suggestion
    fun `M add char W appendIfNotEmpty {buffer is not empty}`(
```

dd-sdk-android-internal/src/test/java/com/datadog/internal/utils/StringBuilderExtKtTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    ) {
+        // Given
+        val buffer = StringBuilder(initialContent)
+
+        // When
+        buffer.appendIfNotEmpty(' ')
+
+        // Then
+
+        assertThat(buffer.toString()).isEqualTo("$initialContent ")
```
0xnm, line 34:
```suggestion
        // Then
        assertThat(buffer.toString()).isEqualTo("$initialContent ")
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/StringBuilderExt.kt
satween: decided to keep as is

0xnm (approved):

PR #2457, opened 2024-12-17
RUM-7549 Handle trace with coroutines

xgouchet:
### What does this PR do?

The main focus is to avoid sharing the same list of scopes across tracers, and add some internal logs to better diagnose issues with tracing.


features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
```diff
@@ -84,9 +89,27 @@ public long getCurrentTimeNano() {
...
+        false,
+        new HashMap<>()
+      );
       return;
     }
     if (!traceId.equals(span.context().getTraceId())) {
+      internalLogger.log(
+        InternalLogger.Level.ERROR,
+        InternalLogger.Target.USER,
+        () -> "Span " + span.getOperationName() + " not registered because of traceId mismatch; " +
```
0xnm, line 107: is it useful here also to log the value of `span.context().getTraceId()`?

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
```diff
@@ -96,19 +119,55 @@ public void registerSpan(final DDSpan span) {
...
+          false,
+          new HashMap<>()
+      );
       return;
     }
     if (!traceId.equals(span.context().getTraceId())) {
+      internalLogger.log(
+          InternalLogger.Level.ERROR,
+          InternalLogger.Target.USER,
+          () -> "Span " + span.getOperationName() + " not expired because of traceId mismatch; " +
```
0xnm, line 152: same as above

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
```diff
@@ -84,9 +89,27 @@ public long getCurrentTimeNano() {
 
   public void registerSpan(final DDSpan span) {
     if (traceId == null || span.context() == null) {
+      internalLogger.log(
+        InternalLogger.Level.ERROR,
```
0xnm, line 93: In `DDSpan` when we are dropping the span `warn` level is used, but here it is an `error` in the whole file. Should they be aligned?

0xnm (approved):
I see that these changes only for OpenTracing, should we do something similar for OpenTelemetry.

Anyway, lgtm.

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
xgouchet: In DDSpan, we use Warning as the cause of the drop is (most likely) misuse of the SDK, usually by calling `finish()` more than once.
Here the drop reason are ones that _should not happen™️_, hence the use of Error level.

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
xgouchet: Makes sense, yes.

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
xgouchet: 👍

PR #2468, opened 2024-12-20
Fix internal telemetry on invalid loading time usage

xgouchet:
### What does this PR do?

The way our telemetry was set up, we could send warnings when there shouldn't have been : 
- if a previous RumViewScope was stopped but kept to wait for children event to be sent, it would always send a warning saying there's no active view, even if an actual active view would actually handle the event.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
@@ -135,10 +135,13 @@ internal class RumViewManagerScope(
         event: RumRawEvent,
         writer: DataWriter<Any>
     ) {
+        val hasNoView = childrenScopes.isEmpty()
         val iterator = childrenScopes.iterator()
+        var hasActiveView = false
         @Suppress("UnsafeThirdPartyFunctionCall") // next/remove can't fail: we checked hasNext
         while (iterator.hasNext()) {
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -149,6 +152,21 @@ internal class RumViewManagerScope(
...
+            sdkCore.internalLogger.log(
+                InternalLogger.Level.WARN,
+                InternalLogger.Target.USER,
+                { NO_ACTIVE_VIEW_FOR_LOADING_TIME_WARNING_MESSAGE }
+            )
+            sdkCore.internalLogger.logApiUsage {
+                InternalTelemetryEvent.ApiUsage.AddViewLoadingTime(
+                    overwrite = event.overwrite,
+                    noView = hasNoView,
+                    noActiveView = !hasNoView
```
jonathanmos, line 166: This is a bit confusing to me. If we get here then it means we're inside a block where hasActiveView is necessarily false, but we could end up passing false to AddViewLoading time for noActiveView indicating that there is actually an active view?

xgouchet: Here I'm replicating the intended usage of this telemetry, where only one of the two attributes can be true at a time.

jonathanmos (approved):

PR #2470, opened 2024-12-30
Fix proguard rules for compose checkbox

jonathanmos:
### What does this PR do?
Checkbox is not visible in obfuscated apks. From an investigation it looks like we need to use `keep` over `keepClassMembers`, and also that the path for CheckboxKt should be different.

### Motivation

### Additional Notes


0xnm (approved):

ambushwork:
Just to clarify, this PR doesn't fix what you've mentioned in the description "also that the path for CheckboxKt should be different.", does it?

ambushwork:
Question about the description "that the path for CheckboxKt should be different."

it should be different from what? from debug?

ambushwork (approved):

PR #2483, opened 2025-01-13
RUM-8051: Implement Head-based sampling for network instrumentation

0xnm:
### What does this PR do?

This PR implements [Head-based sampling](https://docs.datadoghq.com/tracing/trace_pipeline/ingestion_mechanisms/?tab=java#head-based-sampling) for the network instrumentation to align with default sampling existing in other Datadog tracers.

In short, Head-based sampling (HBS) allows to have a single sampling decision for the whole trace: it is kept or dropped, instead of the individual spans.

This also aligns with the following changes in iOS SDK done in the past:

* https://github.com/DataDog/dd-sdk-ios/pull/1783
* https://github.com/DataDog/dd-sdk-ios/pull/1793
* https://github.com/DataDog/dd-sdk-ios/pull/1794

Now, if there is a parent span (propagated through the headers or via request tag; OpenTracing or OpenTelemetry), then its sampling decision will be taken into account instead of the sampling decision made by the network instrumentation.

If there is no parent tracing context for the incoming request, then sampling decision will be made by interceptor.

The necessary integration tests are added, they uncovered some issues we had with trace propagation.

Also this PR bring the following change: spans with `DROP` sampling decision are not written, they are rejected by the intake anyway, so it makes no sense to send them.

There will also be a subsequent PR, allowing to make an _additional_ sampling decision (to drop or keep) for the network span if there is a parent tracing context. 

### DIFF
--- a/reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/HeadBasedSamplingTest.kt
+++ b/reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/HeadBasedSamplingTest.kt
@@ -1,0 +1,670 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.okhttp
+
+import com.datadog.android.Datadog
+import com.datadog.android.api.SdkCore
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.core.internal.utils.toHexString
+import com.datadog.android.core.stub.StubSDKCore
+import com.datadog.android.okhttp.otel.addParentSpan
+import com.datadog.android.okhttp.tests.assertj.SpansPayloadAssert
+import com.datadog.android.okhttp.tests.elmyr.OkHttpConfigurator
+import com.datadog.android.okhttp.trace.TracingInterceptor
+import com.datadog.android.okhttp.trace.parentSpan
+import com.datadog.android.tests.ktx.getString
+import com.datadog.android.trace.AndroidTracer
+import com.datadog.android.trace.Trace
+import com.datadog.android.trace.TraceConfiguration
+import com.datadog.android.trace.TracingHeaderType
+import com.datadog.android.trace.opentelemetry.OtelTracerProvider
+import com.datadog.legacy.trace.api.sampling.PrioritySampling
+import com.datadog.tools.unit.extensions.TestConfigurationExtension
+import com.datadog.tools.unit.getFieldValue
+import com.datadog.tools.unit.getStaticValue
+import com.datadog.tools.unit.setStaticValue
+import com.google.gson.JsonObject
+import com.google.gson.JsonParser
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.annotation.StringForgery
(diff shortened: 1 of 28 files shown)


reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/HeadBasedSamplingTest.kt
```diff
@@ -0,0 +1,663 @@
...
+        // Then
+        val requestSent = mockServer.takeRequest()
+        assertThat(requestSent.getHeader(DATADOG_SAMPLING_PRIORITY_HEADER)).isEqualTo("1")
+        val leastSignificantTraceId = requestSent.getHeader(DATADOG_TRACE_ID_HEADER)
+        checkNotNull(leastSignificantTraceId)
+        val spanId = requestSent.getHeader(DATADOG_SPAN_ID_HEADER)
+        checkNotNull(spanId)
+        val datadogTags = requestSent.getHeader(DATADOG_TAGS_HEADER)
+        checkNotNull(datadogTags)
+        assertThat(datadogTags).startsWith(MOST_SIGNIFICANT_TRACE_ID_KEY)
```
xgouchet, line 122: > [!WARNING]
> Using `startsWith` makes an assumption on the order of the tags in the header. Maybe we should check for the presence of the `MOST_SIGNIFICANT_TRACE_ID_KEY` key but not that it'll be first?

> [!NOTE]
> Are there other tags that we expect and should check here too?

xgouchet (approved):

jonathanmos (approved):

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/HeadBasedSamplingTest.kt
```diff
@@ -0,0 +1,663 @@
...
+        Datadog.stopInstance(stubSdkCore.name)
+        mockServer.shutdown()
+    }
+
+    @Test
+    fun `M use network sampling rate W call is made { no parent context }`(forge: Forge) {
+        // Given
+        if (forge.aBool()) {
+            // outcome should be the same for cases where there is GlobalTracer and when local tracer is used
+            registerGlobalTracer(0.0)
```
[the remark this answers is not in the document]
0xnm: Yes, it is on purpose. Basically we can have 2 tracers used by the network instrumentation: global one or the local one.

In this test we check the case when there is no parent context, meaning that network instrumentation sampling rate should be used, not the sampling rate of the global tracer (or sampling rate of the local tracer).

If this wouldn't be the case, by setting sampling rate here to 0 we won't have any spans/tracing information produced - the test will fail. If this sampling rate is not 0 **and** global tracer is used (while it shouldn't) the test has a chance of being green.

For the test cases when _positive_ sampling decision is carried over from the parent context, there is the opposite: we are setting network sampling rate to 0 and sampling rate of the Global tracer to 100.

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/HeadBasedSamplingTest.kt
0xnm: Good point, I'll update this verification logic.

> Are there other tags that we expect and should check here too?

No, by looking on the `DatadogHttpCodec` implementation only `_dd.p.tid` key/value is set for the `x-datadog-tags` header.

PR #2487, opened 2025-01-17
RUM-7797 Anonymous RUM Identifier

maciejburda:
### What and why?

Adds capability of tracking anonymous id (enabled by default). This data allows linking sessions coming from the same device.

### How?

Following proposal from this [RFC (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/4447830354/RFC+-+Session+Continuity+in+SDKs) it reuses data store mechanism, and based on configuration it generates and reuses stored identifier.

This identifier is attached to `UserInfo` object ([schema already updated](https://github.com/DataDog/rum-events-format/blob/a48cdd6c5dac83b144519fa9d29615ea87d36905/schemas/rum/_common-schema.json#L127)), which is enriching other SDK events.

### Motivation

This extra property adds ability to link sessions coming from the same device.

### Additional Notes

The identifier is read asynchronously from a file, which means RUM events created immediately after calling `RUM.enable()` will not include it. However, this is not an issue since the `anonymous_id` belongs to the Session entity. As long as the `anonymous_id` is present in at least in one event, it is sufficient for linking.

Here's corresponding PR for iOS:
https://github.com/DataDog/dd-sdk-ios/pull/2172

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/UserInfo.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/UserInfo.kt
@@ -22,12 +22,14 @@ import kotlin.jvm.Throws
 
 /**
  * Holds information about the current User.
+ * @property anonymousId a unique anonymous identifier for the device, or null
  * @property id a unique identifier for the user, or null
  * @property name the name of the user, or null
  * @property email the email address of the user, or null
  * @property additionalProperties a dictionary of custom properties attached to the current user
  */
 data class UserInfo(
+    val anonymousId: String? = null,
     val id: String? = null,
     val name: String? = null,
     val email: String? = null,
@@ -37,6 +39,9 @@ data class UserInfo(
     @Suppress("StringLiteralDuplication")
     internal fun toJson(): JsonElement {
         val json = JsonObject()
+        anonymousId?.let { idNonNull ->
+            json.addProperty("anonymous_id", idNonNull)
+        }
         id?.let { idNonNull ->
             json.addProperty("id", idNonNull)
         }
@@ -79,6 +84,7 @@ data class UserInfo(
         @Suppress("StringLiteralDuplication", "ThrowsCount")
         fun fromJsonObject(jsonObject: JsonObject): UserInfo {
             try {
+                val anonymousId = jsonObject.get("anonymous_id")?.asString
                 val id = jsonObject.get("id")?.asString
                 val name = jsonObject.get("name")?.asString
                 val email = jsonObject.get("email")?.asString
@@ -88,7 +94,7 @@ data class UserInfo(
                         additionalProperties[entry.key] = entry.value
                     }
                 }
-                return UserInfo(id, name, email, additionalProperties)
+                return UserInfo(anonymousId, id, name, email, additionalProperties)
             } catch (e: IllegalStateException) {
                 throw JsonParseException(
                     "Unable to parse json into type UserInfo",
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
@@ -8,6 +8,7 @@ package com.datadog.android.api.feature
 
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
+import java.util.UUID
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.ScheduledExecutorService
 
@@ -104,4 +105,11 @@ interface FeatureSdkCore : SdkCore {
      * @param executorContext Context to be used for logging and naming threads running on this executor.
      */
     fun createScheduledExecutorService(executorContext: String): ScheduledExecutorService
+
+    /**
+     * Allows the given feature to set the anonymous ID for the SDK.
+     *
+     * @param anonymousId Anonymous ID to set.
+     */
+    fun setAnonymousId(anonymousId: UUID?)
 }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
@@ -17,6 +17,7 @@ import java.util.concurrent.ScheduledExecutorService
  *
  * SDK core is always guaranteed to implement this interface.
  */
+@Suppress("TooManyFunctions")
 interface FeatureSdkCore : SdkCore {
 
     /**
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
@@ -110,7 +110,7 @@ interface FeatureSdkCore : SdkCore {
     /**
      * Allows the given feature to set the anonymous ID for the SDK.
      *
-     * @param anonymousId Anonymous ID to set.
+     * @param anonymousId Anonymous ID to set. Can be null if feature is disabled.
      */
     fun setAnonymousId(anonymousId: UUID?)
 }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
@@ -270,7 +270,7 @@ internal class DatadogCore(
     }
 
     override fun setAnonymousId(anonymousId: UUID?) {
-        coreFeature.userInfoProvider.setAnonymousId(anonymousId.toString())
+        coreFeature.userInfoProvider.setAnonymousId(anonymousId?.toString())
     }
 
     override fun isCoreActive(): Boolean = isActive
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
@@ -44,6 +44,7 @@ import com.datadog.android.privacy.TrackingConsent
 import com.google.gson.JsonObject
 import java.io.File
 import java.util.Locale
+import java.util.UUID
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.ScheduledExecutorService
@@ -266,6 +267,10 @@ internal class DatadogCore(
     /** @inheritDoc */
     override fun createScheduledExecutorService(executorContext: String): ScheduledExecutorService {
         return coreFeature.createScheduledExecutorService(executorContext)
+    }
+
+    override fun setAnonymousId(anonymousId: UUID?) {
+        coreFeature.userInfoProvider.setAnonymousId(anonymousId.toString())
     }
 
     override fun isCoreActive(): Boolean = isActive
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
@@ -22,6 +22,7 @@ import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.privacy.TrackingConsent
 import com.google.gson.JsonObject
 import java.io.File
+import java.util.UUID
 import java.util.concurrent.Callable
 import java.util.concurrent.Delayed
 import java.util.concurrent.ExecutionException
@@ -127,6 +128,8 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
     override fun createScheduledExecutorService(executorContext: String): ScheduledExecutorService {
         return NoOpScheduledExecutorService()
     }
+
+    override fun setAnonymousId(anonymousId: UUID?) = Unit
 
     // endregion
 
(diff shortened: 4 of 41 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAnonymousIdentifierManager.kt
```diff
@@ -0,0 +1,74 @@
...
+) : AnonymousIdentifierManaging {
+
+    override fun manageAnonymousId(shouldTrack: Boolean) {
+        val anonymousIdKey = "anonymous"
+        if (shouldTrack) {
+            dataStore.value(
+                anonymousIdKey,
+                null,
+                AnonymousIdentifierReadCallback { anonymousId ->
+                    if (anonymousId == null) {
```
satween, line 33: Are you sure wee need this lambda here? 🤔 It basically just add more instructions than saves

Just to consider:
You have an object, implementing `DataStoreReadCallback `: 

````kotlin

object : DataStoreReadCallback<UUID>{

 override fun onSuccess(dataStoreContent: DataStoreContent<UUID>?) {
        core.setAnonymousId(dataStoreContent?.data)
    }

    override fun onFailure() {
       val newAnonymousId = UUID.randomUUID()
                        dataStore.setValue(
                            anonymousIdKey,
                            newAnonymousId,
                            0,
                            null,
                            AnonymousIdentifierSerializer()
                        )
                        core.setAnonymousId(newAnonymousId)
    }
}

````

And that's it, no AnonymousIdentifierReadCallback class, no if(anonymousId == null) and rest logic...

File: features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/Rum.kt [hunk 2]
```diff
@@ -73,6 +73,14 @@ object Rum {
 
         sdkCore.registerFeature(rumFeature)
 
+        sdkCore.getFeature(rumFeature.name)?.dataStore?.let {
+            val anonymousIdentifierManager = AnonymousIdentifierManager(
```
satween, line 77: > [!NOTE]
> Not sure that we need to create an instance of stateless class just to call one method... Maybe some static function will be better here?

satween:
Running 
```
./local_ci.sh -n -c -t -a  
```

from the root of the project will help you to overcome detekt problems.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAnonymousIdentifierManager.kt
maciejburda: If got everything right, I think we would still need the `== null` logic. Implemented callback makes failure and success with null value work the same way:

```
internal class AnonymousIdentifierReadCallback(
    private val onFinished: (UUID?) -> Unit
) : DataStoreReadCallback<UUID> {
    override fun onSuccess(dataStoreContent: DataStoreContent<UUID>?) {
        onFinished(dataStoreContent?.data)
    }

    override fun onFailure() {
        onFinished(null)
    }
}
```

With inline object I would need to repeat this code on both success with no value and failure:
```
val newAnonymousId = UUID.randomUUID()
  dataStore.setValue(
      anonymousIdKey,
      newAnonymousId,
      0,
      null,
      AnonymousIdentifierSerializer()
  )
  core.setAnonymousId(newAnonymousId)
  ```

Comment (maciejburda) [reply features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAnonymousIdentifierManager.kt:33]: Something like this would work, but feels a little bit bloated:

```kotlin
val datastoreCallback = object : DataStoreReadCallback<UUID>{

override fun onSuccess(dataStoreContent: DataStoreContent<UUID>?) {
    handle(dataStoreContent?.data)
}

override fun onFailure() {
    handle(null)
}

fun handle(anonymousId: UUID?) {
    if (anonymousId == null) {
        val newAnonymousId = UUID.randomUUID()
        dataStore.setValue(
            anonymousIdKey,
            newAnonymousId,
            0,
            null,
            AnonymousIdentifierSerializer()
        )
        core.setAnonymousId(newAnonymousId)
    } else {
        core.setAnonymousId(anonymousId)
    }
}
}
dataStore.value(
anonymousIdKey,
null,
datastoreCallback,
AnonymousIdentifierDeserializer()
)
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/Rum.kt
maciejburda: Definitely took the shortcut here. I'll look into it 🙌

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/UserInfo.kt
```diff
@@ -22,12 +22,14 @@ import kotlin.jvm.Throws
...
 /**
  * Holds information about the current User.
+ * @property anonymousId a unique anonymous identifier for the device, or null
  * @property id a unique identifier for the user, or null
  * @property name the name of the user, or null
  * @property email the email address of the user, or null
  * @property additionalProperties a dictionary of custom properties attached to the current user
  */
 data class UserInfo(
+    val anonymousId: String? = null,
```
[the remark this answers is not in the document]
maciejburda: I think that conceptually browser SDK team decided to avoid that. Browser initially used `deviceId` and they went for more generic naming after a couple of iterations.

features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
```diff
@@ -123,6 +123,30 @@
           "type": "string",
           "description": "Email of the user",
           "readOnly": true
+        },
+        "anonymous_id": {
+          "type": "string",
+          "description": "Identifier of the user across sessions",
```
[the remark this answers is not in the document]
maciejburda: It's a mobile vs browser type of problem. In our silo it's less likely for a mobile device to be used by multiple users (but imagine tablet questionnaire in a shopping mall). In some cases it's pointing to device in some to the user.

In reality it's a static id that's likely to be consistent between sessions. It should allow us to know that session is coming from the same source (device/user), but by design we are not bulletproof to app reinstalls, app local data clearing.

From the Product Analytics business perspective they need an identifier that will allows us to soft-link sessions that don't have any user data explicitly set by the SDK customer.

maciejburda:
Relates to https://github.com/DataDog/dd-sdk-ios/pull/2172

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAnonymousIdentifierManager.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.datadog.android.api.storage.datastore.DataStoreReadCallback
+import com.datadog.android.core.internal.persistence.Deserializer
+import com.datadog.android.core.persistence.Serializer
+import com.datadog.android.core.persistence.datastore.DataStoreContent
+import java.util.UUID
+
+/**
+ * This class is responsible for managing the anonymous ID.
+ */
+object RumAnonymousIdentifierManager {
```
0xnm, line 20: blocking: Let's avoid having singletons and convert it to the simple class.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumAnonymousIdentifierManagerTest.kt
```diff
@@ -0,0 +1,94 @@
...
+class RumAnonymousIdentifierManagerTest {
+
+    private lateinit var dataStore: DataStoreHandler
+    private lateinit var core: FeatureSdkCore
+
+    @Before
+    fun setUp() {
+        dataStore = mock()
+        core = mock()
+    }
```
0xnm, line 33:
```suggestion

@Extensions(
    ExtendWith(MockitoExtension::class)
)
class RumAnonymousIdentifierManagerTest {

    @Mocl
    lateinit var dataStore: DataStoreHandler
    @Mock
    lateinit var core: FeatureSdkCore
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumAnonymousIdentifierManagerTest.kt
```diff
@@ -0,0 +1,94 @@
...
+                any<AnonymousIdentifierReadCallback>(),
+                any()
+            )
+        ).thenAnswer {
+            val callback = it.arguments[2] as AnonymousIdentifierReadCallback
+            callback.onSuccess(null)
+        }
+
+        // When
+        RumAnonymousIdentifierManager.manageAnonymousId(true, dataStore, core)
```
0xnm, line 51: this shouldn't compile, `RumAnonymousIdentifierManager` is not `object` class anymore, we need an instance here. Same for tests below.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumAnonymousIdentifierManagerTest.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        // Then
+        verify(dataStore).setValue(eq("anonymous_id_key"), any(), eq(0), eq(null), any())
```
0xnm, line 54:
```suggestion
        verify(dataStore).setValue(eq("anonymous_id_key"), any(), eq(0), isNull, any())
```

0xnm (approved):
I left some comments, but nothing blocking for me.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -268,6 +269,10 @@ internal class DatadogCore(
         return coreFeature.createScheduledExecutorService(executorContext)
     }
 
+    override fun setAnonymousId(anonymousId: UUID?) {
```
satween, line 272: > [!TIP]
> Do we need nullability support here? if `anonymousId` == null then `anonymousId .toString()` will return "null" so if someone will set null it actually will similar to set "null". Not sure if it's an issue, but it's seems like `null` should drop the value instead

satween (approved):
Check my comment above please, bisides that LGTM

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
maciejburda: Yes - when developer using the SDK decided disabling this feature, we would clear the id.

satween (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
```diff
@@ -104,4 +106,11 @@ interface FeatureSdkCore : SdkCore {
      * @param executorContext Context to be used for logging and naming threads running on this executor.
      */
     fun createScheduledExecutorService(executorContext: String): ScheduledExecutorService
+
+    /**
+     * Allows the given feature to set the anonymous ID for the SDK.
+     *
+     * @param anonymousId Anonymous ID to set.
```
ambushwork, line 113: in `@param` line, it's better to explain when should `anonymousId` be given null

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
```diff
@@ -128,6 +129,8 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
         return NoOpScheduledExecutorService()
     }
 
+    override fun setAnonymousId(anonymousId: UUID?) = Unit
```
ambushwork, line 132: you can just omit the return value if it is Unit

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumAnonymousIdentifierManager.kt
```diff
@@ -0,0 +1,91 @@
...
+        dataStore.removeValue(ANONYMOUS_ID_KEY)
+        core.setAnonymousId(null)
+    }
+
+    companion object {
+        private const val ANONYMOUS_ID_KEY = "anonymous_id_key"
+    }
+}
+
+@Suppress("SwallowedException", "UnsafeThirdPartyFunctionCall")
```
ambushwork, line 66: usually we suppress "UnsafeThirdPartyFunctionCall" at the exact call site with the comment why it can be suppress

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumAnonymousIdentifierManager.kt
```diff
@@ -0,0 +1,91 @@
...
+internal class AnonymousIdentifierDeserializer : Deserializer<String, UUID> {
+    override fun deserialize(model: String): UUID? {
+        return try {
+            UUID.fromString(model)
+        } catch (e: IllegalArgumentException) {
+            null
```
ambushwork, line 72: Is it expected that we don't report it to telemetry or even just log it?

ambushwork (dismissed):

jonathanmos (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumAnonymousIdentifierManager.kt
maciejburda: It's very unlikely to happen as we fully control creation. Only some file corruption/deletion would cause this.

There's no fallback other than recreating the uuid, so it should be fine to ignore.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
maciejburda: Other functions in this file follow `= Unit`. I think we should keep the consistency.

satween (approved):

jonathanmos (approved):

PR #2489, opened 2025-01-20
Improve updateFeatureContext performances

xgouchet:
### What does this PR do?

It slightly improve the performance of context sharing across modules. 

### Motivation

Context sharing is what allows different modules (RUM, SR, APM) to share context (e.g.: RUM Session Id) to link events together. 

### Additional Notes

We focused this PR on several bottlenecks. 

- The `getFeatureContext()` method returned a copy of the source-of-truth Map, which was unnecessary as it was already typed as immutable;
- The `updateFeatureContext()` method would create a copy of the source-of-truth map for every module it notified;
- The `updateFeatureContext()` method would use the `.filter{}` utility on a map, which creates an unnecessary copy


detekt_custom.yml
```diff
@@ -1009,6 +1010,7 @@ datadog:
       - "kotlin.collections.MutableList.toTypedArray()"
       - "kotlin.collections.MutableList.withIndex()"
       - "kotlin.collections.MutableList.firstOrNull()"
+      - "kotlin.collections.MutableMap.asSequence()"
```
0xnm, line 1013: it seems there is no usage of this method in the PR?

0xnm (approved):

jonathanmos (approved):

PR #2490, opened 2025-01-23
Adjust telemetry metrics sampling rates

xgouchet:
### What does this PR do?

Adjust the metrics sampling rate, to ensure we have consistent data across platforms, and don't overwhelm our intake.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
@@ -15,7 +15,6 @@ import com.datadog.android.core.internal.persistence.file.advanced.FeatureFileOr
 import com.datadog.android.core.internal.persistence.file.existsSafe
 import com.datadog.android.core.internal.persistence.file.lengthSafe
 import com.datadog.android.core.internal.time.TimeProvider
-import com.datadog.android.core.metrics.MethodCallSamplingRate
 import com.datadog.android.privacy.TrackingConsent
 import java.io.File
 import java.util.Locale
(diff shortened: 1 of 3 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -43,7 +42,7 @@ internal class BatchMetricsDispatcher(
             internalLogger.logMetric(
                 messageBuilder = { BATCH_DELETED_MESSAGE },
                 additionalProperties = it,
-                samplingRate = MethodCallSamplingRate.LOW.rate
```
0xnm, line 46: this will increase the volume by 15x times, is it expected?

xgouchet: Yep, this is to align with iOS metrics

0xnm: By looking on the volume for the "Batch Closed" telemetry for the latest SDK releases (2.17.0 for Android, 2.22.0 for iOS) the difference is like 8x, but only 3x for the 2.16.0 vs 2.21.0, so probably increasing this by 15x will create big disparity (and maybe this value should be lowered on iOS?), but if the goal is to have the same effective sampling rate between the 2 platforms, I'm fine with this change.

0xnm (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/metrics/BatchMetricsDispatcher.kt
```diff
@@ -43,7 +42,7 @@ internal class BatchMetricsDispatcher(
...
+                samplingRate = 1.5f
```
jonathanmos, line 45: For the other sampling rates we've used values from MethodCallSamplingRate. Do we not want to add an additional value there for this rate (do we not expect 1.5f to be reused)?

jonathanmos (approved):

PR #2495, opened 2025-01-28
RUM View Ended" Telemetry

satween:
### What does this PR do?

Adds mandatory telemetry events for TTNS and ITNV

[Spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/4534173727/RUM+View+Ended+Telemetry+-+Spec) 


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewInitializationMetricsState.kt
```diff
@@ -0,0 +1,55 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+package com.datadog.android.rum.internal.metric
+
+internal data class ViewInitializationMetricsState(
+    val initializationTime: Long?,
+    val config: ViewInitializationMetricsConfig,
+    val noValueReason: NoValueReason?
+)
```
xgouchet, line 12: > [!NOTE]
Nitpick, feel free to disregard. Although both the metrics inv

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewMetricDispatcher.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.datadog.tools.annotation.NoOpImplementation
+
+@NoOpImplementation
+internal interface ViewMetricDispatcher {
+    fun sendViewEnded(invState: ViewInitializationMetricsState, tnsState: ViewInitializationMetricsState)
+    fun onDurationResolved(newDuration: Long)
+    fun onViewLoadingTimeResolved(newLoadingTime: Long)
+
+    enum class ViewType {
+        APPLICATION,
```
xgouchet, line 17: > [!NOTE]
> Nitpick, if this relates to the App Launch, we might want to name it that: `APPLICATION_LAUNCH`.

xgouchet (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewInitializationMetricsState.kt
satween: Sorry, I didn't get the question, what do you mean here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -828,6 +832,11 @@ internal open class RumViewScope(
             memoryVitalMonitor.unregister(memoryVitalListener)
             frameRateVitalMonitor.unregister(frameRateVitalListener)
             networkSettledMetricResolver.viewWasStopped()
+
+            viewEndedMetricDispatcher.sendViewEnded(
```
0xnm, line 836: Is it intentional that we are using a different lingvo: `ended` vs `stopped`? Shouldn't we align?

View may be stopped, but still be in the list of the scopes (but not active ones), waiting for the child scopes to finish.

Once all childs scopes finish, the view is considered as `complete` (see https://github.com/DataDog/dd-sdk-android/blob/6c54695c4f36678a840cb45c3016882a90185b90/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L1314-L1322).

 Is `ended` here an equivalent of `stopped` or `complete`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1388,6 +1397,16 @@ internal open class RumViewScope(
             interactionToNextViewMetricResolver: InteractionToNextViewMetricResolver,
             networkSettledResourceIdentifier: InitialResourceIdentifier
         ): RumViewScope {
+            val networkSettledMetricResolver = NetworkSettledMetricResolver(
+                networkSettledResourceIdentifier,
+                sdkCore.internalLogger
+            )
+
+            val viewEndedMetricDispatcher = ViewEndedMetricDispatcher(
+                viewType = ViewMetricDispatcher.ViewType.CUSTOM,
```
0xnm, line 1406: shouldn't we inherit it as an argument of `fromEvent` call and it is up to the caller to provide a relevant value?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,162 @@
...
+    private val samplingRate: Float = DEFAULT_SAMPLE_RATE
+) : ViewMetricDispatcher {
+
+    private var duration: Long? = null
+    private var loadingTime: Long? = null
+    private var metricSent: Boolean = false
+
+    override fun sendViewEnded(invState: ViewInitializationMetricsState, tnsState: ViewInitializationMetricsState) {
+        if (metricSent) {
+            return internalLogger.log(
```
0xnm, line 24:
```suggestion
           internalLogger.log(
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    companion object {
+        const val DEFAULT_SAMPLE_RATE: Float = 0.75f
+
+        internal const val VIEW_ENDED_MESSAGE = "[Mobile Metric] RUM View Ended"
+
+        internal const val KEY_METRIC_TYPE = "metric_type"
+
+        private const val VALUE_METRIC_TYPE = "rum view ended"
```
0xnm, line 92: should this one be with underscores?

```suggestion
        private const val VALUE_METRIC_TYPE = "rum_view_ended"
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,162 @@
...
+        @JvmStatic
+        @VisibleForTesting
+        internal fun toAttributeValue(config: ViewInitializationMetricsConfig): String = when (config) {
+            ViewInitializationMetricsConfig.CUSTOM -> VALUE_CUSTOM
+            ViewInitializationMetricsConfig.TIME_BASED_DEFAULT -> VALUE_TIME_BASED_DEFAULT
+            ViewInitializationMetricsConfig.TIME_BASED_CUSTOM -> VALUE_TIME_BASED_CUSTOM
+        }
+
+        @JvmStatic
+        @VisibleForTesting
```
0xnm, line 141: why `JvmStatic` attribute is needed here, if they are supposed only be used from Kotlin?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/networksettled/NetworkSettledMetricResolver.kt
```diff
@@ -30,12 +33,16 @@ internal class NetworkSettledMetricResolver(
...
+    private var currentViewDiagnostic = Diagnostic()
+
     fun viewWasCreated(eventTimestampInNanos: Long) {
         viewCreatedTimestamp = eventTimestampInNanos
+        currentViewDiagnostic = Diagnostic()
     }
 
     fun resourceWasStarted(context: InternalResourceContext) {
         if (viewWasStopped) return
+        currentViewDiagnostic.started += 1
```
jonathanmos, line 45: Question: networkSettledMetricResolver can be called only on a single thread, or we could have calls coming from several threads? Because some of the variables in this class are volatile, but diagnostic does not seem to be explicitly threadsafe.

jonathanmos (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewEndedMetricDispatcher.kt
ncreated: No, it is fine and it follows the spec (and iOS implementation).
```
“metric_type”: “rum view ended”,
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,162 @@
...
+
+        internal const val VIEW_ENDED_MESSAGE = "[Mobile Metric] RUM View Ended"
+
+        internal const val KEY_METRIC_TYPE = "metric_type"
+
+        private const val VALUE_METRIC_TYPE = "rum view ended"
+
+        internal const val KEY_RVE = "rve"
+        internal const val KEY_DURATION = "duration"
+        internal const val KEY_LOADING_TIME = "loading_time_value"
```
ncreated, line 96: change-request/ Quoting the spec:
```
  “loading_time”:    { “value”: <int> }
```
it should be rather a `loading_time.value`.

It seems that `put(KEY_LOADING_TIME, buildMap { put(KEY_VALUE, loadingTime) })` will create `"loading_time_value": {"value": ...}` instead, no?

ncreated (changes requested):
I only validated the spec values, finding a problem around `loading_time` key - requesting a change. Other that this, it looks very good, well done 🙂 💪.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewEndedMetricDispatcher.kt
satween: True, thanks, fixed !

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/networksettled/NetworkSettledMetricResolver.kt
satween: Thanks for spotting, fixed 👍

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
satween: Nice catch here - changes followed from this note lead me to found potential issue. 
The `ended` here comes from metric name rather than android sdk notation so I prefer to keep it as it is ( to be aligned with `SessionEndedMetric`), but entire method call now checks for child scopes

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
satween: I found that there is `RumViewType` enum that does actually the same so removed duplication and reusing  `RumViewType` instead. There is no reason to get the value from method parameters because  `RumViewType` is not provided at this moment, but thanks for spotting - it helped to reduce code duplication

ncreated (approved):
Looks good to me 👍

xgouchet (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1388,6 +1386,16 @@ internal open class RumViewScope(
...
         ): RumViewScope {
+            val networkSettledMetricResolver = NetworkSettledMetricResolver(
+                networkSettledResourceIdentifier,
+                sdkCore.internalLogger
+            )
+
+            val viewEndedMetricDispatcher = ViewEndedMetricDispatcher(
+                viewType = RumViewType.FOREGROUND,
+                internalLogger = sdkCore.internalLogger
+            )
```
0xnm, line 1397: Is it a bold assumption that `RumViewScope#fromEvent` will always be called for foreground flow?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/ViewEndedMetricDispatcher.kt
```diff
@@ -0,0 +1,164 @@
...
+    companion object {
+        const val DEFAULT_SAMPLE_RATE: Float = 75f
+
+        internal const val VIEW_ENDED_MESSAGE = "[Mobile Metric] RUM View Ended"
+
+        internal const val KEY_METRIC_TYPE = "metric_type"
+
+        private const val VALUE_METRIC_TYPE = "rum view ended"
+
+        internal const val KEY_RVE = "rve"
```
0xnm, line 98: maybe it is worth adding a comment saying what "rve" means? Or use full form in the constant name for the clarity, like we do below for `KEY_TIME_TO_NETWORK_SETTLED = "tns"`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolver.kt
```diff
@@ -74,22 +72,36 @@ internal class InteractionToNextViewMetricResolver(
...
+
+        // Third - the case where there is no interaction from the previous view.
+        if (lastInteractions[previousViewId] == null) {
+            return NoValueReason.InteractionToNextView.NO_ACTION
         }
-        return null
+
+        // Finally - checking that the previous view interaction is eligible
+        resolveLastInteraction(previousViewId, currentViewCreatedTimestamp)
+            ?: return NoValueReason.InteractionToNextView.NO_ELIGIBLE_ACTION
```
0xnm, line 101: There is a corner case when view ID may change if session ID has changed https://github.com/DataDog/dd-sdk-android/blob/7387cc9494ff1321c969ad2664f6e36ff3dbeca1/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L226-L229, but it is hard to address it anyway.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/interactiontonextview/InteractionToNextViewMetricResolverTest.kt
```diff
@@ -110,6 +113,104 @@ internal class InteractionToNextViewMetricResolverTest {
...
+        assertThat(result.noValueReason).isEqualTo(NoValueReason.InteractionToNextView.NO_ACTION)
+    }
+
+    @Test
+    fun `M return NO_ELIGIBLE_ACTION W getState { all actions was filtered out by LastInteractionIdentifier }`(
+        forge: Forge
+    ) {
+        // Given
+        val fakePreviousActions = forge.generateFakeActions(1, 10, fakeFirstViewId)
+        val fakeViewCreatedTimestamp = System.nanoTime()
```
0xnm, line 152: isn't it better to generate a random value here? not a big deal though

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/ViewEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,280 @@
...
+        // Then
+        verify(mockInternalLogger).log(
+            eq(InternalLogger.Level.WARN),
+            target = eq(Target.TELEMETRY),
+            messageBuilder = any(),
+            throwable = eq(null),
+            onlyOnce = eq(false),
+            additionalProperties = eq(null)
+        )
+        verifyNoMoreInteractions(mockInternalLogger)
```
0xnm, line 111: minor: It is better to have a single `When` and single `Then` if possible. Here you can do both `dispatcherUnderTest.sendViewEnded` in `When` and have both assertions below, because they have different arguments to assert.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/ViewEndedMetricDispatcherTest.kt
```diff
@@ -0,0 +1,280 @@
...
+        // When
+        dispatcherUnderTest.sendViewEnded(fakeInvState, fakeTnsState)
+
+        // Then
+        verify(mockInternalLogger).logMetric(
+            messageBuilder = argThat { invoke() == VIEW_ENDED_MESSAGE },
+            additionalProperties = eq(expectedAttributes()),
+            samplingRate = eq(fakeSampleRate),
+            creationSampleRate = eq(null)
+        )
```
0xnm, line 129: to verify `InternalLogger` calls you can use `verifyLog` extension function, this will remove the need to call `invoke` for message assertion and also it won't be any need to use `eq`.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/interactiontonextview/TimeBasedInteractionIdentifierTest.kt
```diff
@@ -117,5 +117,30 @@ internal class TimeBasedInteractionIdentifierTest : ObjectTest<TimeBasedInteract
         assertThat(result).isFalse()
     }
 
+    @Test
+    fun `M return true W defaultThresholdUsed{default delay used}`() {
```
0xnm, line 121:
```suggestion
    fun `M return true W defaultThresholdUsed {default delay used}`() {
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/interactiontonextview/TimeBasedInteractionIdentifierTest.kt
```diff
@@ -117,5 +117,30 @@ internal class TimeBasedInteractionIdentifierTest : ObjectTest<TimeBasedInteract
...
+
+        // When
+        val result = testedValidator.defaultThresholdUsed()
+
+        // Then
+        assertThat(result).isTrue()
+    }
+
+    @Test
+    fun `M return false W defaultThresholdUsed{custom delay used}`(forge: Forge) {
```
0xnm, line 133:
```suggestion
    fun `M return false W defaultThresholdUsed {custom delay used}`(forge: Forge) {
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/metric/interactiontonextview/TimeBasedInteractionIdentifierTest.kt
```diff
@@ -117,5 +117,30 @@ internal class TimeBasedInteractionIdentifierTest : ObjectTest<TimeBasedInteract
...
+        // Given
+        val testedValidator = TimeBasedInteractionIdentifier(
+            TimeBasedInteractionIdentifier.DEFAULT_TIME_THRESHOLD_MS + forge.aLong(min = 0, max = 100)
```
0xnm, line 136: minor: `forge.aLong(min = 0, max = 100)` can be replaced by the `@LongForgery(min = 0, max = 100)` test method argument.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/forge/TelemetryViewInitializationMetricsStateForgeryFactory.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import com.datadog.android.rum.internal.metric.NoValueReason
+import com.datadog.android.rum.internal.metric.ViewInitializationMetricsConfig
+import com.datadog.android.rum.internal.metric.ViewInitializationMetricsState
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+internal class TelemetryViewInitializationMetricsStateForgeryFactory : ForgeryFactory<ViewInitializationMetricsState> {
+    override fun getForgery(forge: Forge): ViewInitializationMetricsState {
+        val initializationTime = forge.aNullable { forge.aLong(min = 0L) }
```
0xnm, line 17: minor: `Forge` methods taking lambda will have `Forge` as `this`, so such things can be written as `forge.aNullable { aLong(min = 0L) }`

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -201,7 +205,16 @@ internal class RumViewScopeTest {
     private lateinit var mockNetworkSettledMetricResolver: NetworkSettledMetricResolver
 
     @Mock
-    private lateinit var mockInteractionToNextViewMetricResolver: InteractionToNextViewMetricResolver
+    private lateinit var mockInteractionNextViewMetricResolver: InteractionToNextViewMetricResolver
```
0xnm, line 208: is it expected rename? Should't it still have `To` in the name as in the interface name?

satween: looks like auto-renamed by AS

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
satween: Yes, basically it's default value for the RumViewScope. I will pass it explicitly

PR #2496, opened 2025-01-30
RUM-8314 allow disabling 404 span redaction

xgouchet:
### What does this PR do?

Adds an option to keep the resource name for network requests ending in `404` as original, instead of having them redacted to `"404"`.

### DIFF
--- a/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
+++ b/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
@@ -84,6 +84,7 @@ internal constructor(
     internal val traceOrigin: String?,
     internal val traceSampler: Sampler<Span>,
     internal val traceContextInjection: TraceContextInjection,
+    internal val redacted404ResourceName: Boolean,
     internal val localTracerFactory: (SdkCore, Set<TracingHeaderType>) -> Tracer
 ) : Interceptor {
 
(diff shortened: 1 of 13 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -882,6 +889,18 @@ internal constructor(
             return getThis()
         }
 
+        /**
+         * Set whether network requests returning a 404 status code should have their resource name redacted.
+         * In order to reduce the cardinality of resource names in APM, 404 urls are automatically redacted to
+         * "404"
+         * @param redacted if true, all 404 requests will have a resource name set to "404", else the resource name
+         * will be the URL
+         */
```
0xnm, line 898:
```suggestion
        /**
         * Set whether network requests returning a 404 status code should have their resource name redacted.
         * In order to reduce the cardinality of resource names in APM, 404 URLs are automatically redacted to
         * "404".
         * @param redacted if true, all 404 requests will have a resource name set to "404", else the resource name
         * will be the URL
         */
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -882,6 +889,18 @@ internal constructor(
...
+        fun set404ResourcesRedacted(redacted: Boolean): R {
```
0xnm, line 899: side thought: This API may be confusing for `DatadogInterceptor` users, because this option applies only to APM actually, but they may think that they will see RUM Resource name as "404". But since it is a shared builder, we cannot remove this method from Builder for `DatadogInterceptor`.

0xnm (dismissed):

0xnm (approved):

PR #2502, opened 2025-02-04
RUM-8366 Forward RUM Session ID in trace headers

xgouchet:
### What does this PR do?

As part of our Joint Sampling project, we forward the RUM session ID as tag in the tracing headers (DD and W3C style).

### DIFF
--- a/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/DatadogHttpCodec.java
+++ b/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/DatadogHttpCodec.java
@@ -31,7 +31,9 @@ class DatadogHttpCodec {
 
     public static final String OT_BAGGAGE_PREFIX = "ot-baggage-";
     public static final String LEAST_SIGNIFICANT_TRACE_ID_KEY = "x-datadog-trace-id";
-    public static final String MOST_SIGNIFICANT_TRACE_ID_KEY = "_dd.p.tid";
+    public static final String MOST_SIGNIFICANT_TRACE_ID_TAG_KEY = "_dd.p.tid";
+    public static final String RUM_SESSION_ID_TAG_KEY = "_dd.p.rsid";
+    public static final String RUM_SESSION_ID_KEY = "session_id";
     public static final String DATADOG_TAGS_KEY = "x-datadog-tags";
     public static final String SPAN_ID_KEY = "x-datadog-parent-id";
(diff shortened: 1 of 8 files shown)


features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/DatadogHttpCodec.java
```diff
@@ -31,7 +31,9 @@ class DatadogHttpCodec {
 
     public static final String OT_BAGGAGE_PREFIX = "ot-baggage-";
     public static final String LEAST_SIGNIFICANT_TRACE_ID_KEY = "x-datadog-trace-id";
-    public static final String MOST_SIGNIFICANT_TRACE_ID_KEY = "_dd.p.tid";
+    public static final String MOST_SIGNIFICANT_TRACE_ID_TAG_KEY = "_dd.p.tid";
+    public static final String RUM_SESSION_ID_TAG_KEY = "_dd.p.rsid";
+    public static final String RUM_SESSION_ID_KEY = "session_id";
```
0xnm, line 36: won't this key have a risk to overlap with customer-set key? maybe we should prefix it with Datadog/RUM?

0xnm (approved):

jonathanmos (approved):

0xnm (changes requested):
I actually realised that we need to do similar changes to the codes used by Otel as well? like https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/propagation/DatadogHttpCodec.java

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/DatadogHttpCodec.java
xgouchet: This key is a tag we set ourselves in the Android Tracer on the span, and the customer doesn't have access to that span before the generation of the headers.

0xnm: Won't APM backend process this tag as well in the `SpanEvent` received? Or it is used only for the propagation?

xgouchet: It's only used for propagation

jonathanmos (approved):

0xnm (approved):

PR #2504, opened 2025-02-10
[RUM-8375]  View ended instrumentation type attribute support

satween:
### What does this PR do?

Adds support for optional instrumentation_type attributes into view ended telemetry metric.

### Motivation
[Spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/4534173727/RUM+View+Ended+Telemetry+-+Spec)

[Previous](https://github.com/DataDog/dd-sdk-android/pull/2495) PR was too big to add this logic, I'd decided to split them into two separate ones 

### Additional Notes

I noticed that we were using the additionalAttributes map as a container for some local values that shouldn't be passed 
directly to the backend, but used to determine some values just before the actual events are sent, so as a side effect of this PR the local attributes logic has been refactored: Added a single place for such attributes and their cleanup.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/BundleExt.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/BundleExt.kt
@@ -13,8 +13,8 @@ internal const val ARGUMENT_TAG = "view.arguments"
 /**
  * Converts this bundle into a Map of attributes to be included in a RUM View event.
  */
-fun Bundle?.convertToRumViewAttributes(): MutableMap<String, Any?> {
-    if (this == null) return mutableMapOf()
+fun Bundle?.convertToRumViewAttributes(): Map<String, Any?> {
+    if (this == null) return emptyMap()
 
     val attributes = mutableMapOf<String, Any?>()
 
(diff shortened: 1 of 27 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/BundleExt.kt
```diff
@@ -13,8 +13,8 @@ internal const val ARGUMENT_TAG = "view.arguments"
 /**
  * Converts this bundle into a Map of attributes to be included in a RUM View event.
  */
-fun Bundle?.convertToRumViewAttributes(): Map<String, Any?> {
-    if (this == null) return emptyMap()
+fun Bundle?.convertToRumViewAttributes(): MutableMap<String, Any?> {
```
ambushwork, line 16: I have a bit doubt if we need to change the return value for the enrichment. Usually we return always immutable map in case of unintended modification.  In your `enrichWithConstantAttribute` function, you can call `toMutableMap` to create a new map with the cloned content inside.

ambushwork (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/BundleExt.kt
satween: I just remembered last performance issues with toMap/toMuttableMap thait @xgouchet fixed, and thought that we could save some transformations. But I agree, immutable is better and we should do some performance adjustments only if we face real problem, will fix that

jonathanmos (approved):

ambushwork (approved):

ambushwork (approved):

PR #2505, opened 2025-02-10
[RUM-8098] Support for configuration schema updates for time based strategy of TNS and INV metrics

satween:
### What does this PR do?

Adds support for INV and TNS time-based thresholds if they used for metric collection
Small side effect on LocalAttribute - making `string` field private

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/attributes/LocalAttribute.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/attributes/LocalAttribute.kt
@@ -18,22 +18,23 @@ interface LocalAttribute {
      * Enumeration of all local attributes keys used in the application.
      * Made via **enum** to make sure that all such attributes will be removed before sending the event to the backend.
      *
-     * @property string - Unique string value for a local attribute key.
+     * @param string - Unique string value for a local attribute key.
      */
     @InternalApi
     enum class Key(
-        val string: String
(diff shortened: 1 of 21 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/attributes/LocalAttribute.kt
```diff
@@ -18,22 +18,23 @@ interface LocalAttribute {
      * Enumeration of all local attributes keys used in the application.
      * Made via **enum** to make sure that all such attributes will be removed before sending the event to the backend.
      *
-     * @property string - Unique string value for a local attribute key.
+     * @param string - Unique string value for a local attribute key.
      */
     @InternalApi
     enum class Key(
-        val string: String
+        private val string: String
```
ambushwork, line 25: Can you elaborate the purpose of make it private and expose it by `toString()` function? it seems the main change of this PR, but I don't read the explanation in the description

satween: Yep, this change shadowed main purpose a bit. The idea behind this change is to reduce duplicity of the LocalProperty.Key object. As all java/kotlin objects has `toString` method by default. I just missed this fact in the previous PR and noticed that making `sting` property public makes no sense.

ambushwork (approved):

PR #2509, opened 2025-02-20
RUM-8673 Update UserInfo API to make id mandatory.

maciejburda:
### What does this PR do?

Makes id parameter of setUserInfo() mandatory. Old signature is kept with deprecation warning.

### Motivation

Change requested by PA.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -249,6 +272,11 @@ object Datadog {
      */
     @JvmStatic
     @JvmOverloads
+    @Deprecated(
+        "UserInfo id property is now mandatory.",
+        ReplaceWith("setUserInfo(id!!, name, email, extraInfo, sdkCore)")
```
0xnm, line 277: I think we shouldn't put `ReplaceWith` here, because it uses force-unwrap. Essentially method name is the same, and user needs to adjust just call parameter.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -249,6 +272,11 @@ object Datadog {
...
+    )
+    @JvmName("setUserInfoDeprecated")
```
0xnm, line 279: Having this is fine for me, it is only for Java-first users, which is probably the minority.

Another option to avoid conflicting overload would be to not modify the signature, but log a warning in case if `null`  is received, but many users won't notice it.

0xnm (commented):
Overall the solution looks okay to me, I left some thoughts.

0xnm (approved):

jonathanmos (approved):

Raheelnaz:
what if there is no userInfo? for example in a logged out state?

PR #2512, opened 2025-02-21
RUM-8729 ensure span logs use 128 bits trace id as hex string

xgouchet:
### What does this PR do?

Ensures that the logs created from a `DDSpan.log()` behave the same way as the ones created from our `Logger` with `bundleWithTrace=true`.

### DIFF
--- a/sample/kotlin/src/main/kotlin/com/datadog/android/sample/traces/TracesViewModel.kt
+++ b/sample/kotlin/src/main/kotlin/com/datadog/android/sample/traces/TracesViewModel.kt
@@ -368,14 +368,20 @@ internal class TracesViewModel(
             withinSpan("AsyncOperation", activeSpanInMainThread) {
                 logger.v("Starting Async Operation...")
 
-                for (i in 0..100) {
+                val count = (Random().nextInt() % 50) + 50
+                log("Async op loops $count times")
+                var actualCount = 0
+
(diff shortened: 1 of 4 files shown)


sample/kotlin/src/main/kotlin/com/datadog/android/sample/traces/TracesViewModel.kt
```diff
@@ -368,14 +368,20 @@ internal class TracesViewModel(
...
                 logger.v("Starting Async Operation...")
 
-                for (i in 0..100) {
+                val count = (Random().nextInt() % 50) + 50
+                log("Async op loops $count times")
+                var actualCount = 0
+
+                for (i in 0 until count) {
                     if (isCancelled) {
+                        log("Async operation cancelled")
```
jonathanmos, line 377: minor/ do we want these log statements?

jonathanmos (approved):

sample/kotlin/src/main/kotlin/com/datadog/android/sample/traces/TracesViewModel.kt
xgouchet: Those are logs on the Span object, so they're here to test that the Span.logs are tied to the right Span in the UI

jonathanmos (approved):

PR #2516, opened 2025-02-27
[RUM-8652] Allow definition of custom implementations of specific SR methods

cdn34dd:
### What does this PR do?

- Exposes a new SessionReplayConfigutation option to allow clients to pass their own implementations of specific methods used in the Session Replay SDK. 

- Updates the SDK code to use the overriden function if provided and falling back to the original implementation otherwise. 
( Currently only `getCurrentActivity` is allowed to be redefined, but there's a high likelihood that more will be needed ) 

### Motivation

Certain parts of the session replay SDK don't  work as intended when ran in a react-native environment, leading to broken replays. In order to keep platform specific code out of the android SDK, the goal is to allow each client to add their own implementations of the parts that don't work.

### Additional Notes

We keep the `setInternalCallback` only accessible through a internal proxy in order to not expose internal config options. In practice it would look something like this:
```
        val internalCallback = ReactNativeSessionReplayInternalCallback(reactContext)
        
        val configuration = SessionReplayConfiguration.Builder(replaySampleRate.toFloat())
            /* .... */
            .setImagePrivacy(/* .... */)
            .setTouchPrivacy(/* .... */)
            .setTextAndInputPrivacy(/* .... */)
            .let { builder ->
                _SessionReplayInternalProxy(builder).setInternalCallback(internalCallback)
            }
```

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
@@ -8,6 +8,7 @@ package com.datadog.android.sessionreplay
 
 import androidx.annotation.FloatRange
 import com.datadog.android.api.InternalLogger
+import com.datadog.android.sessionreplay.internal.recorder.SessionReplayRecorder
 import com.datadog.android.sessionreplay.recorder.OptionSelectorDetector
 import com.datadog.android.sessionreplay.utils.DrawableToColorMapper
 import java.util.Locale
@@ -27,7 +28,8 @@ data class SessionReplayConfiguration internal constructor(
     internal val touchPrivacy: TouchPrivacy,
     internal val textAndInputPrivacy: TextAndInputPrivacy,
     internal val dynamicOptimizationEnabled: Boolean,
-    internal val systemRequirementsConfiguration: SystemRequirementsConfiguration
+    internal val systemRequirementsConfiguration: SystemRequirementsConfiguration,
+    internal val internalCallback: SessionReplayInternalCallback
 ) {
 
     /**
@@ -73,6 +75,7 @@ data class SessionReplayConfiguration internal constructor(
         private var extensionSupportSet: MutableSet<ExtensionSupport> = mutableSetOf()
         private var dynamicOptimizationEnabled = true
         private var systemRequirementsConfiguration = SystemRequirementsConfiguration.NONE
+        private var internalCallback: SessionReplayInternalCallback = NoOpSessionReplayInternalCallback()
 
         /**
          * Adds an extension support implementation. This is mostly used when you want to provide
@@ -211,6 +214,19 @@ data class SessionReplayConfiguration internal constructor(
         }
 
         /**
+         * Allows definition of custom callback functions for Session Replay
+         * that may require platform-specific behavior.
+         * Currently, this enables defining:
+         * - [SessionReplayInternalCallback.getCurrentActivity]:
+         *   Used in [SessionReplayRecorder] to register fragment lifecycle callbacks
+         *   for clients initialized after the `Application.onCreate` phase.
+         */
+        internal fun setInternalCallback(internalCallback: SessionReplayInternalCallback): Builder {
+            this.internalCallback = internalCallback
+            return this
+        }
+
+        /**
          * Builds a [SessionReplayConfiguration] based on the current state of this Builder.
          */
         fun build(): SessionReplayConfiguration {
@@ -226,7 +242,8 @@ data class SessionReplayConfiguration internal constructor(
                 sampleRate = sampleRate,
                 startRecordingImmediately = startRecordingImmediately,
                 dynamicOptimizationEnabled = dynamicOptimizationEnabled,
-                systemRequirementsConfiguration = systemRequirementsConfiguration
+                systemRequirementsConfiguration = systemRequirementsConfiguration,
+                internalCallback = internalCallback
             )
         }
 
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/_SessionReplayInternalProxy.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/_SessionReplayInternalProxy.kt
@@ -1,0 +1,38 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay
+
+import com.datadog.android.lint.InternalApi
+
+/**
+ * This class exposes internal methods that are used by other Datadog modules and cross platform
+ * frameworks. It is not meant for public use.
+ *
+ * DO NOT USE this class or its methods if you are not working on the internals of the Datadog SDK
+ * or one of the cross platform frameworks.
+ *
+ * Methods, members, and functionality of this class  are subject to change without notice, as they
+ * are not considered part of the public interface of the Datadog SDK.
+ */
+@InternalApi
+@Suppress(
+    "ClassName",
+    "ClassNaming"
+)
+class _SessionReplayInternalProxy(private val builder: SessionReplayConfiguration.Builder) {
+    /**
+     * Sets an internal callback for session replay.
+     *
+     * @param internalCallback callback instance to override specific parts of the codebase.
+     * @return [SessionReplayConfiguration.Builder] instance.
+     */
+    fun setInternalCallback(
+        internalCallback: SessionReplayInternalCallback
+    ): SessionReplayConfiguration.Builder {
+        return builder.setInternalCallback(internalCallback)
+    }
+}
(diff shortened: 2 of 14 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayCustomCallbacks.kt
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import android.view.Window
+
+data class SessionReplayCustomCallbacks(
+    val getCurrentWindows: (() -> List<Window>)? = null
+)
```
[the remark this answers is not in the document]
cdn34dd: The general idea was to make a simple configuration object we could add to the configuration builder that holds optional callbacks, allowing clients to set only the ones they needed. But here both would work.

0xnm:
I think we shouldn't go this way (by exposing it in the public API along with other methods), it should be an internal API accessible from the proxy.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/_SessionReplayInternalProxy.kt
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * DO NOT USE this class or its methods if you are not working on the internals of the Datadog SDK
+ * or one of the cross platform frameworks.
+ *
+ * Methods, members, and functionality of this class  are subject to change without notice, as they
+ * are not considered part of the public interface of the Datadog SDK.
+ */
+@InternalApi
+@Suppress(
+    "UndocumentedPublicClass",
```
ambushwork, line 24: This class has been documented above, why do we still need this Suppression here?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayCustomCallbacks.kt
```diff
@@ -0,0 +1,13 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.sessionreplay.internal.recorder
+
+import android.view.Window
+
+data class SessionReplayCustomCallbacks(
+    val getCurrentWindows: (() -> List<Window>)? = null
```
ambushwork, line 12: I think instead of having a `data class` with only one callback inside, you can just make it as an `interface`
```
@NoOpImplementation
interface SessionReplayCustomCallback{
      fun getCurrentWindows() : List<Window>
}
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -210,6 +213,15 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Specifies callbacks functions to override standard behaviours in Session Replay
+         * that need to behave differently depending on the platform.
+         */
```
ambushwork, line 219: While reading this documentation, I would expect to understand what's the exact behaviours it will override. can you elaborate it?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -73,6 +75,7 @@ data class SessionReplayConfiguration internal constructor(
         private var extensionSupportSet: MutableSet<ExtensionSupport> = mutableSetOf()
         private var dynamicOptimizationEnabled = true
         private var systemRequirementsConfiguration = SystemRequirementsConfiguration.NONE
+        private var customCallbacks: SessionReplayCustomCallbacks = SessionReplayCustomCallbacks()
```
ambushwork, line 78: is it expected that the class name and field name are all pluriel?

0xnm:
Please ignore my comment above, I had an old version of this PR opened. I will review a new one.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
cdn34dd: I would say in this case yes, but what would you suggest ?

ambushwork: usually we keep class name singular to represent a single instance of an object, unless the class itself present a pluriel collection such as `data class Users(val users: List<User>)`, and for the field name it should be like:

```
private var customCallback : SessionReplayCustomCallback = SessionReplayCustomCallback()

// or 

private var customCallbacks : List<SessionReplayCustomCallback> = SessionReplayCustomCallback()

```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayCustomCallbacks.kt
0xnm, line 12: minor: there is no need to use `get` prefix for `val` property, it can be just `val currentWindows`

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
```diff
@@ -169,8 +172,7 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
             onDrawListenerProducer = DefaultOnDrawListenerProducer(
                 snapshotProducer = SnapshotProducer(
                     DefaultImageWireframeHelper(
-                        logger = internalLogger,
-                        resourceResolver = resourceResolver,
+                        logger = internalLogger, resourceResolver = resourceResolver,
```
0xnm, line 175: to be reverted

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayCustomCallbacks.kt
0xnm, line 12: maybe some explanation/docs here will be useful

0xnm (approved):
I left few suggestions, but nothing blocking. LGTM.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayCustomCallbacks.kt
ambushwork: Instead of having several callbacks feature, is that possible to have only one callback but with more functions in it?

cdn34dd: Not sure what that means, can you explain it a bit more or give an example ? Just for context, all these functions we need to override from RN or other project can all be unrelated to each other and can affect different parts of the sdk.

ambushwork: If I understand correctly, what you intent to do should be like:

```
data class SessionReplayCustomCallbacks(
    val getCurrentWindows: (() -> List<Window>)? = null
    val doSomethingElseA: ((SomeInputA) -> SomeOutputA)? = null
    val doSomethingElseB: ((SomeInputB) -> List<SomeOutputB>)? = null
)
```

Instead of doing that, you can do :
```
@NoOpImplementation
interface SessionReplayCustomCallback{

     fun getCurrentWindows() : List<Window>

     fun doSomethingElseA(SomeInputA) : SomeOutputA

     fun doSomethingElseA(SomeInputB) : SomeOutputB
}
```

In that case you will have only one callback but can be called in different placesto serve for your need.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/_SessionReplayInternalProxy.kt
```diff
@@ -0,0 +1,40 @@
...
+    "UndocumentedPublicFunction",
+    "UndocumentedPublicProperty",
+    "ClassName",
+    "ClassNaming",
+    "VariableNaming"
+)
+class _SessionReplayInternalProxy {
```
ambushwork, line 31: _SessionReplayInternalProxy.setCustomCallbacks(builder, customCallbacks)

_SessionReplayInternalProxy(builder).setCustomCallbacks(customCallbacks)

ambushwork (approved):

xgouchet (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayLifecycleCallback.kt
```diff
@@ -21,6 +21,24 @@ internal class SessionReplayLifecycleCallback(
 
     private val currentActiveWindows = WeakHashMap<Window, Any?>()
 
+    fun setCurrentWindow(activity: Activity) {
```
ambushwork, line 24: Here the function name is `setCurrentWindow` but you are taking activity as the argument, which makes a bit confusing,
you can just change to `setCurrentWindow(window: Window)` and let the call site worry about how to retrieve the window.

ambushwork (approved):

PR #2525, opened 2025-03-06
RUM-8041: Upload quality telemetry

jonathanmos:
### What does this PR do?
Adds details about upload quality to Session Ended telemetry - information about cycle count, blockers and network failures.

Cycle count are sent as part of session ended telemetry. The structure is:

```
rse: {
  ...
  upload_cycle: {
    rum: <int>,
    tracing: <int>,
    logs: <int>,
    ...
  }
}
```

Blockers and failures are sent as type mobile metrics batch blocked. The structure is:

```
{
  blockers: <string[]> // optional - if there were blockers (can be lpm, low_battery or offline)
  failure: <string> // optional - if there was a failure this would be the error code (e.g. 500)
  batch_count: <int> // indicates the number of batches blocked by the failure/blocker. 
  For example, if there were 20 batches in an upload trigger 
  and the failure happened on the fourth upload this number would be 16.
}
```

### Motivation
Increase visibility into upload quality. 

### Additional Notes


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
```diff
@@ -86,15 +95,41 @@ internal class DataUploadRunnable(
 
     private fun isNetworkAvailable(): Boolean {
         val networkInfo = networkInfoProvider.getLatestNetworkInfo()
-        return networkInfo.connectivity != NetworkInfo.Connectivity.NETWORK_NOT_CONNECTED
+        val isOffline = networkInfo.connectivity == NetworkInfo.Connectivity.NETWORK_NOT_CONNECTED
+
+        if (isOffline) {
+            sendUploadQualityEvent(
```
satween, line 101: Minor: Not sure if this was expected but from `isNetworkAvailable` I expect that method just return some boolean value, and there is no side effects like calling`sendUploadQualityEvent`. I suggest call `sendUploadQualityEvent` in some separate pipeline or rename method to something like `resolveNetworkAvailability`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
```diff
@@ -116,6 +151,12 @@ internal class DataUploadRunnable(
         batchMeta: ByteArray?
     ): UploadStatus {
         val status = dataUploader.upload(context, batch, batchMeta, batchId)
+        if (status.code != HTTP_SUCCESS_CODE) {
```
satween, line 154: minor: maybe it's better to have condition like `status.code >= 200 && status.code < 300`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -72,6 +76,41 @@ internal class SessionEndedMetric(
         missedEventCountByType[missedEventType] = (missedEventCountByType[missedEventType] ?: 0) + 1
     }
 
+    fun onUploadQualityTracked(event: UploadQualityEvent) {
+        val track = uploadQuality.getOrPut(event.track) {
+            mutableMapOf<String, Any?>().apply {
```
satween, line 81: tip: Maybe it's better to have `mutableMapOf<UploadQualityMetric, Any?>()` ? then you'll need to call `.key` only when serializing it

satween (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
jonathanmos: I don't think that in practice we're receiving non-202 success codes. In `DataOkHttpUploader` we're specifically checking for 202 as success, and log unexpected status codes. Also in `DefaultUploadSchedulerStrategy` we're specifically using 202 as the basis for decreasing the backoff.  We can expand the definition of success, but it should perhaps be a wider discussion.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
```diff
@@ -125,9 +177,20 @@ internal class DataUploadRunnable(
...
+                specificType = specificType
+            )
+        )
+    }
+
     // endregion
 
     companion object {
         internal const val LOW_BATTERY_THRESHOLD = 10
+        private const val HTTP_SUCCESS_CODE = 202
```
0xnm, line 198: minor: 202 is `Accepted`. This is as well how it is named in `DataOkHttpUploader.HTTP_ACCEPTED`

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -692,6 +708,324 @@ internal class DataUploadRunnableTest {
...
+        // When
+        testedRunnable.run()
+
+        // Then
+        argumentCaptor<UploadQualityEvent> {
+            verify(mockRumFeature, times(2)).sendEvent(
+                event = capture()
+            )
+            val firstEvent = firstValue as? UploadQualityEvent
+            val secondEvent = secondValue as? UploadQualityEvent
```
0xnm, line 753: Why do we use safe cast here? I think we can use `as` directly, because we expect these to be that type.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -692,6 +708,324 @@ internal class DataUploadRunnableTest {
...
+            category = UploadQualityCategories.COUNT,
+            specificType = null
+        )
+        val expectedBlocker = UploadQualityEvent(
+            track = fakeFeatureName,
+            category = UploadQualityCategories.BLOCKER,
+            specificType = UploadQualityBlockers.OFFLINE.key
+        )
+        val fakeExpected = listOf(expectedCount, expectedBlocker)
+        assertThat(allValues.containsAll(fakeExpected)).isTrue()
```
0xnm, line 872:
```suggestion
        assertThat(allValues).containsAll(fakeExpected)
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -692,6 +708,324 @@ internal class DataUploadRunnableTest {
...
+            category = UploadQualityCategories.COUNT,
+            specificType = null
+        )
+        val expectedBlocker = UploadQualityEvent(
+            track = fakeFeatureName,
+            category = UploadQualityCategories.BLOCKER,
+            specificType = UploadQualityBlockers.LOW_BATTERY.key
+        )
+        val fakeExpected = listOf(expectedCount, expectedBlocker)
+        assertThat(allValues.containsAll(fakeExpected)).isTrue()
```
0xnm, line 934:
```suggestion
        assertThat(allValues).containsAll(fakeExpected)
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -692,6 +708,324 @@ internal class DataUploadRunnableTest {
...
+            category = UploadQualityCategories.BLOCKER,
+            specificType = UploadQualityBlockers.LOW_BATTERY.key
+        )
+        val expectedBlocker2 = UploadQualityEvent(
+            track = fakeFeatureName,
+            category = UploadQualityCategories.BLOCKER,
+            specificType = UploadQualityBlockers.LOW_POWER_MODE.key
+        )
+        val fakeExpected = listOf(expectedCount, expectedBlocker, expectedBlocker2)
+        assertThat(allValues.containsAll(fakeExpected)).isTrue()
```
0xnm, line 1004:
```suggestion
        assertThat(allValues).containsAll(fakeExpected)
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnableTest.kt
```diff
@@ -692,6 +708,324 @@ internal class DataUploadRunnableTest {
...
+    }
+
+// endregion
```
0xnm, line 1007:
```suggestion
    // endregion
```

dd-sdk-android-internal/api/apiSurface
```diff
@@ -28,6 +28,17 @@ sealed class com.datadog.android.internal.telemetry.InternalTelemetryEvent
     class AddViewLoadingTime : ApiUsage
       constructor(Boolean, Boolean, Boolean, MutableMap<String, Any?> = mutableMapOf())
   object InterceptorInstantiated : InternalTelemetryEvent
+enum com.datadog.android.internal.telemetry.UploadQualityCategories
+  - COUNT
+  - BLOCKER
+  - FAILURE
+enum com.datadog.android.internal.telemetry.UploadQualityBlockers
```
0xnm, line 35: should these enums be names in a singular form? Like `UploadQualityCategory` and `UploadQualityBlocker` respectively.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -128,9 +128,10 @@ internal class DatadogCore(
     /** @inheritDoc */
     override fun registerFeature(feature: Feature) {
         val sdkFeature = SdkFeature(
-            coreFeature,
-            feature,
-            internalLogger
+            featureSdkCore = this,
```
0xnm, line 131: Alternatively we can just add new method to `InternalLogger`. But not a big deal.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/UploadQualityEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+ * The different categories of upload quality events.
+ * - COUNT: The number of uploads.
+ * - BLOCKER: The count of different blockers that prevented an upload.
+ * - FAILURE: The count of different failure codes.
+ */
+enum class UploadQualityCategories {
+    COUNT,
+    BLOCKER,
+    FAILURE
+}
```
0xnm, line 19: should we back values by the string literal as for the `UploadQualityBlockers `

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/UploadQualityEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.telemetry
+
+/**
+ * The different categories of upload quality events.
+ * - COUNT: The number of uploads.
+ * - BLOCKER: The count of different blockers that prevented an upload.
+ * - FAILURE: The count of different failure codes.
```
0xnm, line 13: minor: these should probably go to the individual values doc instead of enum doc. Same for enum below.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/SessionEndedMetric.kt
```diff
@@ -37,6 +39,8 @@ internal class SessionEndedMetric(
     private var lastTrackedView: TrackedView? = null
     private var wasStopped: Boolean = false
 
+    private var uploadQuality: MutableMap<String, MutableMap<UploadQualityMetric, Any?>> = mutableMapOf()
```
0xnm, line 42: seems like it can be `val`?

```suggestion
    private val uploadQuality = mutableMapOf<String, MutableMap<UploadQualityMetric, Any?>>()
```

0xnm, line 42: we never clean it up? to remove obsolete keys. Seems like this one will grow indefinitely?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -332,6 +335,158 @@ class RumSessionEndedMetricTest {
...
+
+        // Then
+        val attributes = sessionEndedMetric.toMetricAttributes(fakeNtpOffsetAtEnd)
+        val rse = attributes[SessionEndedMetric.RSE_KEY] as Map<*, *>
+        val uploadQuality = rse[UploadQualityMetric.UPLOAD_METRIC_KEY.key] as Map<*, *>
+        val eventTrack = uploadQuality[fakeTrack] as Map<*, *>
+        val cycleCount = eventTrack[UploadQualityMetric.CYCLE_COUNT_KEY.key] as Int
+        assertThat(cycleCount).isEqualTo(1)
+        val failures = eventTrack[UploadQualityMetric.FAILURE_COUNT_KEY.key] as Map<*, *>
+        assertThat(failures).isEqualTo(emptyMap<String, Any>())
```
0xnm, line 368:
```suggestion
        assertThat(failures).isEmpty()
```

same can be applied to the similar lines below

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/metric/RumSessionEndedMetricTest.kt
```diff
@@ -332,6 +335,158 @@ class RumSessionEndedMetricTest {
...
+        val sessionEndedMetric = stubSessionEndedMetric()
+        val fakeTrack = forge.anElementFrom(
+            listOf(
+                "rum",
+                "logs",
+                "tracing",
+                "session-replay",
+                "session-replay-resources"
+            )
+        )
```
0xnm, line 385: minor: this method supports `vararg`, so `listOf` call can be removed.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
```diff
@@ -32,18 +34,25 @@ internal class DataUploadRunnable(
     private val systemInfoProvider: SystemInfoProvider,
     internal val uploadSchedulerStrategy: UploadSchedulerStrategy,
     internal val maxBatchesPerJob: Int,
-    private val internalLogger: InternalLogger
+    private val internalLogger: InternalLogger,
+    private val uploadQualityListener: UploadQualityListener
 ) : UploadRunnable {
 
+    private var delayMs: Long = 0
+    private var batchConsumerAvailableAttempts: Int = 0
```
0xnm, line 42: I think we shouldn't extract these 2 as properties, this will make `DataUploadRunnable` stateful and things may become dangerous.

Let's keep them as method-local if possible?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -399,6 +425,24 @@ internal class RumFeature(
...
+
+        if (event is UploadQualityEvent.UploadQualityBlockerEvent) {
+            properties[UploadBlockerMetric.BATCH_COUNT_KEY.key] = event.batchCount
+            properties[UploadBlockerMetric.BLOCKERS_KEY.key] = event.blockers
+        } else if (event is UploadQualityEvent.UploadQualityFailureEvent) {
+            properties[UploadBlockerMetric.BATCH_COUNT_KEY.key] = event.batchCount
+            properties[UploadBlockerMetric.FAILURE_KEY.key] = event.failure
+        }
+
+        return properties.toMap()
```
0xnm, line 443: why an additional call to `toMap` is needed? It seems to me that it is redundant.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/UploadBlockerMetric.kt
```diff
@@ -0,0 +1,17 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.metric
+
+internal enum class UploadBlockerMetric(val key: String) {
+    METRICS_KEY("[Mobile Metric] Batch Blocked"),
```
0xnm, line 10: I have a feeling that in this enum it is actually a mix of values and keys, am I wrong?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/RumFeatureTest.kt
```diff
@@ -1235,6 +1273,109 @@ internal class RumFeatureTest {
...
+    ) {
+        // Given
+        testedFeature.onInitialize(appContext.mockInstance)
+
+        // When
+        testedFeature.onReceive(fakeUploadQualityBlockerEvent)
+
+        // Then
+        val expectedAdditionalProperties = mapOf(
+            KEY_METRIC_TYPE to UploadBlockerMetric.BATCH_BLOCKED_KEY.key,
```
0xnm, line 1327: so here `UploadBlockerMetric.BATCH_BLOCKED_KEY` is not a key, but value

PR #2565, opened 2025-03-24
RUM-9130: Fix flaky TraceWriterTest M log error and proceed W write() { serialization failed }

aleksandr-gringauz:
### What does this PR do?

Fixed a flaky test. Had to parametrize `SpanForgeryFactory` in order to control forging the `samplingPriority` field. I am not sure it is a canonical way of doing things so feel free to criticize.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/TraceWriter.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/TraceWriter.kt
@@ -92,5 +92,7 @@ internal class TraceWriter(
     companion object {
         internal const val ERROR_SERIALIZING = "Error serializing %s model"
         internal val DROP_SAMPLING_PRIORITIES = setOf(PrioritySampling.SAMPLER_DROP, PrioritySampling.USER_DROP)
+        private val KEEP_SAMPLING_PRIORITIES = setOf(PrioritySampling.SAMPLER_KEEP, PrioritySampling.USER_KEEP)
+        internal val KEEP_AND_UNSET_SAMPLING_PRIORITIES = KEEP_SAMPLING_PRIORITIES + PrioritySampling.UNSET
     }
 }
--- a/features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/TraceWriterTest.kt
+++ b/features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/TraceWriterTest.kt
@@ -20,6 +20,7 @@ import com.datadog.android.trace.internal.storage.ContextAwareSerializer
 import com.datadog.android.trace.model.SpanEvent
 import com.datadog.android.trace.utils.verifyLog
 import com.datadog.android.utils.forge.Configurator
+import com.datadog.android.utils.forge.SpanForgeryFactory
 import com.datadog.opentracing.DDSpan
 import com.datadog.tools.unit.forge.aThrowable
 import fr.xgouchet.elmyr.Forge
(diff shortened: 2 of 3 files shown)


features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/TraceWriterTest.kt
```diff
@@ -333,4 +328,12 @@ internal class TraceWriterTest {
...
 
     // endregion
+
+    private fun generateSpanListWithPriorities(forge: Forge, priorities: Set<Int>): MutableList<DDSpan> {
+        return forge.aList {
+            SpanForgeryFactory { frg, span ->
+                span.samplingPriority = frg.anElementFrom(priorities)
+            }.getForgery(forge)
+        }.toMutableList()
+    }
```
0xnm, line 338: it can be done simpler, without the need to do changes in `SpanForgeryFactory` (plus it can be an extension method):

```suggestion
    private fun Forge.spansWithPriorities(forge: Forge, priorities: Set<Int>): MutableList<DDSpan> {
        return aList {
            getForgery<DDSpan>().apply { span.samplingPriority = anElementFrom(priorities) }
        }.toMutableList()
    }
```

but I'm okay with keeping the proposed change as well.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/TraceWriter.kt
```diff
@@ -92,5 +92,7 @@ internal class TraceWriter(
     companion object {
         internal const val ERROR_SERIALIZING = "Error serializing %s model"
         internal val DROP_SAMPLING_PRIORITIES = setOf(PrioritySampling.SAMPLER_DROP, PrioritySampling.USER_DROP)
+        private val KEEP_SAMPLING_PRIORITIES = setOf(PrioritySampling.SAMPLER_KEEP, PrioritySampling.USER_KEEP)
+        internal val KEEP_AND_UNSET_SAMPLING_PRIORITIES = KEEP_SAMPLING_PRIORITIES + PrioritySampling.UNSET
```
0xnm, line 96: probably these are needed only for tests? maybe put them in the `TraceWriterTest`?

0xnm (dismissed):

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/TraceWriterTest.kt
aleksandr-gringauz: > it can be done simpler

I did it the way you describe at first. But this actually is a different solution that didn't quite work. 

1. Suppose `SpanForgeryFactory` sets `samplingPriorityHere` to `PrioritySampling.USER_DROP` [here](https://github.com/DataDog/dd-sdk-android/blob/f807ef6c3e71a3580e6a80f2bee0f28904d02d9e/features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/utils/forge/SpanForgeryFactory.kt#L39).
2. Then we change it the way you describe to `PrioritySampling.UNSET`. `PrioritySampling.UNSET` satisfies the condition `!in DROP_SAMPLING_PRIORITIES`, it was used in the test before, so we probably should not forget about it.
3. [Here](https://github.com/DataDog/dd-sdk-android/blob/f807ef6c3e71a3580e6a80f2bee0f28904d02d9e/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/DDSpanContext.java#L222) `PrioritySampling.UNSET` is ignored.
4. We end up with a failed test because we have `PrioritySampling.USER_DROP` as priority.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/TraceWriter.kt
aleksandr-gringauz: I thought it is better this way because if you put `KEEP_AND_UNSET_SAMPLING_PRIORITIES` in test source set, when it is needed in the main source set later we will end up with duplicate code with very high chance. These things should probably be grouped together even if some are needed only in tests for now.

jonathanmos (dismissed):

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/TraceWriterTest.kt
0xnm: okay, makes sense then for `PrioritySampling.UNSET` case

aleksandr-gringauz:
just resolved conflicts

PR #2571, opened 2025-03-28
Fix flaky tests in OtelTraceWriter

jonathanmos:
### What does this PR do?
Some tests in the OtelTraceWriter are flaky because of the method used to generate lists of DDSpans. In some cases, all the elements of the list can be filtered out leaving empty lists, and this causes the tests to fail. 

Solution: ensure the list of spans generated by the tests is not empty.

### Motivation
Flakiness noticed in the tests.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriterTest.kt
+++ b/features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriterTest.kt
@@ -106,9 +106,10 @@ internal class OtelTraceWriterTest {
     @Test
     fun `M write spans W write()`(forge: Forge) {
         // GIVEN
-        val ddSpans = forge.aList { getForgery<DDSpan>() }
-            .filter { it.samplingPriority() !in OtelTraceWriter.DROP_SAMPLING_PRIORITIES }
-            .toMutableList()
+        val ddSpans = createNonEmptyDdSpans(
+            forge = forge,
+            includeDropSamplingPriority = false
+        )
         val spanEvents = ddSpans.map { forge.getForgery<SpanEvent>() }
         val serializedSpans = ddSpans.map { forge.aString() }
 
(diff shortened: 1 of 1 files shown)


features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriterTest.kt
```diff
@@ -301,5 +307,21 @@ internal class OtelTraceWriterTest {
...
+            { it.samplingPriority() in OtelTraceWriter.DROP_SAMPLING_PRIORITIES }
+        } else {
+            { it.samplingPriority() !in OtelTraceWriter.DROP_SAMPLING_PRIORITIES }
+        }
+
+        var spans = emptyList<DDSpan>()
+        while (spans.isEmpty()) {
+            spans = forge.aList { getForgery<DDSpan>() }
+                .filter(predicate)
+        }
```
0xnm, line 321: we can also use approach which @aleksandr-gringauz did in https://github.com/DataDog/dd-sdk-android/pull/2565 (there it is a bit more flexible), but here it is also okay

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriterTest.kt
```diff
@@ -301,5 +307,21 @@ internal class OtelTraceWriterTest {
         verifyNoMoreInteractions(mockSdkCore, mockTracingFeatureScope)
     }
 
+    private fun createNonEmptyDdSpans(forge: Forge, isDropSampling: Boolean): MutableList<DDSpan> {
```
0xnm, line 310: maybe for the clarity we can use `includeDropSamplingPriority` here instead of `isDropSampling` or something similar?

0xnm (approved):

0xnm (approved):

PR #2563, opened 2025-03-21
RUM-593: [Android] - Add traceSampleRate to Telemetry Configuration Event

aleksandr-gringauz:
### What does this PR do?

It adds a sampling rate from OkHttp interceptor to the telemetry configuration event. In the `init` block of `TracingInterceptor` we put the sampling rate into the feature context of the tracing feature. And then extract it in `DatadogCore.sendCoreConfigurationTelemetryEvent` when we send the event to the Rum feature.

Problems that remain unsolved:
1. If `TracingInterceptor` is created before `DatadogCore` then we will not send the sampling rate, because there is no way to submit the feature context without having an instance of `DatadogCore`. It happened in the sample app. I fixed it by making necessary things lazy.
2. There can be multiple instances of `TracingInterceptor` in the app and we don't know which one's sampling rate will be sent.

But for our purposes it looks like that these problems aren't blocking.

demo [internal]: [link](https://dd-rum-telemetry.datadoghq.com/logs?query=%40type%3Aconfiguration%20%40configuration.trace_sample_rate%3A%3E0%20source%3Aandroid&agg_m=%40session.id&agg_m_source=base&agg_q=%40configuration.trace_sample_rate&agg_q_source=base&agg_t=cardinality&cols=host%2Cservice&fromUser=true&messageDisplay=inline&panel=%7B%22timeRange%22%3A%7B%22from%22%3A1742561778000%2C%22to%22%3A1742565378000%2C%22live%22%3Atrue%7D%2C%22queryString%22%3A%22%40type%3Aconfiguration%20%40configuration.trace_sample_rate%3A%3E0%20source%3Aandroid%20AND%20%40configuration.trace_sample_rate%3A100%22%7D&refresh_mode=sliding&storage=hot&stream_sort=desc&top_n=10&top_o=top&viz=sunburst&x_missing=true&from_ts=1742561778000&to_ts=1742565378000&live=true)

### Motivation

What inspired you to submit this pull request?

### Additional Notes

Anything else we should know when reviewing?

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](CONTRIBUTING.md) doc)

PR #2572, opened 2025-03-28
RUM-562: Report configured distributed tracing headers as part of Configuration telemetry

aleksandr-gringauz:
### What does this PR do?

We provide distributed tracing header types to the `selectedTracingPropagators` field of the `TelemetryConfigurationEvent` in a similar way we did with the `traceSampleRate` here https://github.com/DataDog/dd-sdk-android/pull/2563. The disadvantages described in that pr also apply here.

demo [internal]:
[link](https://dd-rum-telemetry.datadoghq.com/logs?query=%40type%3Aconfiguration%20service%3Add-sdk-android&agg_m=%40session.id&agg_m_source=base&agg_q=%40configuration.selected_tracing_propagators&agg_q_source=base&agg_t=cardinality&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&top_n=10&top_o=top&viz=sunburst&x_missing=true&from_ts=1742808269390&to_ts=1743413069390&live=true) logs from the sample app. Select at least 1 week window so that you can see events from Friday March 28th.

### Motivation

### Additional Notes


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreInitializationTest.kt
```diff
@@ -371,8 +373,13 @@ internal class DatadogCoreInitializationTest {
         testedCore.features += Feature.TRACING_FEATURE_NAME to mockTracingFeature
 
         val traceSampleRate = forge.aFloat(min = 0f, max = 100f)
+        val hostsWithHeaderTypes: HostsWithHeaderTypes? = forge.aNullable<HostsWithHeaderTypes>()
```
0xnm, line 376:
```suggestion
        val hostsWithHeaderTypes = forge.aNullable<HostsWithHeaderTypes>()
```

dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/HostsWithHeaderTypesForgeryFactory.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import com.datadog.android.internal.telemetry.HostsWithHeaderTypes
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+class HostsWithHeaderTypesForgeryFactory : ForgeryFactory<HostsWithHeaderTypes> {
+    override fun getForgery(forge: Forge): HostsWithHeaderTypes {
+        return HostsWithHeaderTypes(
+            hosts = forge.aMap {
+                aString() to aList {
```
0xnm, line 17: minor: maybe generate something like a host look-like? by using say `aStringMatching("https://[a-z0-9]+\\.com")`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import com.datadog.android.internal.telemetry.HostsWithHeaderTypes
+import com.datadog.android.telemetry.model.TelemetryConfigurationEvent
+
+internal object HostsWithHeaderTypesMapper {
+    fun mapToSelectedTracingPropagators(
+        headerTypes: HostsWithHeaderTypes
+    ): List<TelemetryConfigurationEvent.SelectedTracingPropagator> {
+        return headerTypes.hosts
+            .values
```
0xnm, line 17: so why do we collect (or in other words, have it in the model) hosts information if in the end we only send header types to the configuration telemetry?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.telemetry.internal
+
+import com.datadog.android.internal.telemetry.HostsWithHeaderTypes
+import com.datadog.android.telemetry.model.TelemetryConfigurationEvent
+
+internal object HostsWithHeaderTypesMapper {
+    fun mapToSelectedTracingPropagators(
```
0xnm, line 13: suggestion: maybe have it as an extension method is simpler?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapperTest.kt
```diff
@@ -0,0 +1,59 @@
...
+                host3 to emptySet()
+            )
+        )
+
+        val result = HostsWithHeaderTypesMapper.mapToSelectedTracingPropagators(hostsWithHeaderTypes)
+
+        assertThat(result)
+            .containsExactlyInAnyOrder(mappedHeaderType1, mappedHeaderType2)
+            .doesNotHaveDuplicates()
+    }
```
0xnm, line 58:
```suggestion
        // Given
        val host1 = forge.aString()
        val host2 = forge.aString()
        val host3 = forge.aString()

        val headerType1 = forge.aValueFrom(HostsWithHeaderTypes.HeaderType::class.java)
        val headerType2 = forge.aValueFrom(HostsWithHeaderTypes.HeaderType::class.java, exclude = listOf(headerType1))

        val mappedHeaderType1 = headerType1.toSelectedTracingPropagator()
        val mappedHeaderType2 = headerType2.toSelectedTracingPropagator()

        val hostsWithHeaderTypes = HostsWithHeaderTypes(
            hosts = mapOf(
                host1 to setOf(headerType1),
                host2 to setOf(headerType2, headerType1),
                host3 to emptySet()
            )
        )

        // When
        val result = HostsWithHeaderTypesMapper.mapToSelectedTracingPropagators(hostsWithHeaderTypes)

        // Then
        assertThat(result)
            .containsExactlyInAnyOrder(mappedHeaderType1, mappedHeaderType2)
            .doesNotHaveDuplicates()
    }
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandlerTest.kt
```diff
@@ -1549,6 +1549,10 @@ internal class TelemetryEventHandlerTest {
             .hasUseLocalEncryption(internalConfigurationEvent.useLocalEncryption)
             .hasIsMainProcess(fakeDatadogContext.processInfo.isMainProcess)
             .hasTraceSampleRate(internalConfigurationEvent.okhttpInterceptorSampleRate?.toLong())
+            .hasSelectedTracingPropagators(
+                internalConfigurationEvent.hostsWithHeaderTypes
+                    ?.let(HostsWithHeaderTypesMapper::mapToSelectedTracingPropagators)
+            )
```
0xnm, line 1555: if we drop hosts from `HostsWithHeaderTypes`, then we don't need `mapToSelectedTracingPropagators` call and can probably have simply `.hasSelectedTracingPropagators(internalConfigurationEvent.hostsWithHeaderTypes)` call

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapper.kt
aleksandr-gringauz: I felt that if we are making some information (header types) available to the outside it is better to expose "the original" information, not the result of some function which removes most of it.

Suppose we need something else connected to these hosts and headers (maybe a set of hosts?), now we can just use the thing I did here.

However I actually don't have a strong opinion here, one can argue for example that it is better to expose as little as possible to the outside.

I have no objections to refactor the code to expose only a set of headers.

0xnm: yes, let's change this PR to have only tracing header types passed around, it will greatly simply the things. We can add hosts information later if needed.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandlerTest.kt
aleksandr-gringauz: No, I don't think so.

1. `selectedTracingPropagators` field of `TelemetryConfigurationEvent` has the type `List<SelectedTracingPropagator>`.
2. `SelectedTracingPropagator` is located in `dd-sdk-android-rum` and is not available in `dd-sdk-android-core` where `InternalTelemetryEvent.Configuration` is [constructed](https://github.com/DataDog/dd-sdk-android/blob/8464f0f45d12983188b8cf13433767f66165b114/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt#L519). 
3. We still have to pass a set of enums in `internalConfigurationEvent.hostsWithHeaderTypes` and then map it to  `List<SelectedTracingPropagator>` in the `TelemetryEventHandler`.
4. And the same mapping has to be done in tests.
5. We could make `internalConfigurationEvent.hostsWithHeaderTypes` a `Set<String>` and then construct `SelectedTracingPropagator` from `String`. But I think it is a bad approach.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreInitializationTest.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapper.kt
aleksandr-gringauz: done

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapper.kt
aleksandr-gringauz: done

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapper.kt
```diff
@@ -0,0 +1,32 @@
...
+package com.datadog.android.telemetry.internal
+
+import com.datadog.android.internal.telemetry.InternalTracingHeaderType
+import com.datadog.android.internal.telemetry.InternalTracingHeaderTypesSet
+import com.datadog.android.telemetry.model.TelemetryConfigurationEvent
+
+internal object HostsWithHeaderTypesMapper {
+    fun InternalTracingHeaderTypesSet.toSelectedTracingPropagators() = types.map { it.toSelectedTracingPropagator() }
+
+    private fun InternalTracingHeaderType.toSelectedTracingPropagator():
```
0xnm, line 16: suggestion: maybe make them 1st class citizens? without `HostsWithHeaderTypesMapper` object wrapper (which should also be renamed). Given that they have custom receiver, it seems there is no benefit.

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTest.kt
```diff
@@ -1572,7 +1574,7 @@ internal open class TracingInterceptorTest {
     }
 
     @Test
-    fun `M add trace sample rate to tracing feature context W constructor called`(
+    fun `M add trace sample rate and header types to tracing feature context W constructor called`(
```
0xnm, line 1577: maybe split in 2 different tests?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/trace/TracingHeaderTypeMapper.kt
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.okhttp.internal.trace
+
+import com.datadog.android.internal.telemetry.InternalTracingHeaderType
+import com.datadog.android.internal.telemetry.InternalTracingHeaderTypesSet
+import com.datadog.android.trace.TracingHeaderType
+
+internal object TracingHeaderTypeMapper {
```
ambushwork, line 13: In my opinion, We try to avoid to use Kotlin `object` if it doesn't have a strong argument in our SDK, what's the necessity of this one?

ambushwork (approved):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/trace/TracingHeaderTypeMapper.kt
aleksandr-gringauz: No problem, I will remove the objects. To me it is just a question of style and preference.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTracingHeaderType.kt
```diff
@@ -0,0 +1,15 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.telemetry
+
+@Suppress("UndocumentedPublicClass")
+enum class InternalTracingHeaderType {
```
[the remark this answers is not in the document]
aleksandr-gringauz: good point, I'll try to do this

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/trace/TracingHeaderTypeMapper.kt
ambushwork: For our SDK it's just not just a style, the main problem of using object is that it's hard to mock it since we don't use `mockk` for static mocking. And it also stays globally in the memory even if it is never used.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTracingHeaderType.kt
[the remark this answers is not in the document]
aleksandr-gringauz: ok, I thought about it

> maybe drop Internal prefix from the name? Is it useful to have it?

My idea was to distinguish it from an already existing `TracingHeaderType`. [link](https://github.com/DataDog/dd-sdk-android/blob/8464f0f45d12983188b8cf13433767f66165b114/dd-sdk-android-core/src/main/kotlin/com/datadog/android/trace/TracingHeaderType.kt#L14) which is public and can be used by clients of the sdk. Thus the prefix - internal.

> Also dd-sdk-android-okhttp already depends on the dd-sdk-android-rum module, so we can reference SelectedTracingPropagator from there directly, without having a dedicated intermediate enum with basically same values. Is it useful to have it?

Now these header types are passed to `InternalTelemetryEvent.Configuration` [here](https://github.com/DataDog/dd-sdk-android/blob/aleksandr-gringauz/rum-562/report-tracing-headers/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTelemetryEvent.kt#L45). It is in the `dd-sdk-android-internal` and can't access neighter `TracingHeaderType` nor `SelectedTracingPropagator`.

**There are 2 choices:**
1. Continue using `InternalTelemetryEvent.Configuration`, but pass the `Set<SelectedTracingPropagator>` as `Any`. Not the best because we have `Any`.
2. Extract the set of headers right in the `TelemetryEventHandler` similar to [this](https://github.com/DataDog/dd-sdk-android/blob/8464f0f45d12983188b8cf13433767f66165b114/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt#L288). Not the best because we in `TelemetryEventHandler` (rum feature) we rely on the fact that there is tracing feature somewhere (features start knowing about one another).

I like approach with `InternalTelemetryEvent.Configuration` (the way I did it in pr, not with `Any`) because:
1. It uses static typing (at least in the `TelemetryConfigurationHandler`).
2. The code in `TelemetryConfigurationHandler.createConfigurationEvent` is "cleaner" - just a function that accepts `InternalTelemetryEvent.Configuration` as parameter and creates an event (if we don't take [this](https://github.com/DataDog/dd-sdk-android/blob/8464f0f45d12983188b8cf13433767f66165b114/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt#L288) into account).

**I suggest 2 options:**
1. Either I leave it the way it is now (`InternalTracingHeaderType` remains).
2. Stop putting things in `InternalTelemetryEvent.Configuration` and just extract everything from different features' contexts in `TelemetryEventHandler`. I can event change my implementation for `okhttpInterceptorSampleRate` from the previous pr.

I like option 1 more. The fact that we have a duplicate for `TracingHeaderType` isn't a big problem imho. Duplication - yes. But we have kind of bigger separation of public api `TracingHeaderType` and internal api `InternalTracingHeaderType`. There is close to no risk that we forget to change `TracingHeaderType` if `InternalTracingHeaderType` changes because it won't compile. And we can change `InternalTracingHeaderType` more freely because it is not public api. Maybe `dd-sdk-android-internal` will become a place where we store shared models that can be accessed by all features instead of using HashMaps to communicate between modules.

But I have no problem refactoring to option 2.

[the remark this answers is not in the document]
0xnm: > My idea was to distinguish it from an already existing TracingHeaderType. [link](https://github.com/DataDog/dd-sdk-android/blob/8464f0f45d12983188b8cf13433767f66165b114/dd-sdk-android-core/src/main/kotlin/com/datadog/android/trace/TracingHeaderType.kt#L14) which is public and can be used by clients of the sdk. Thus the prefix - internal.

What is being added already have a different package name and moreover it is a part of `dd-sdk-android-internal`, which is added as `implementation` everywhere, so it is not re-exported to the public API.

> I suggest 2 options:

Let's do option 1 then, I'm fine with that.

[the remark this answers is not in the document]
aleksandr-gringauz: > What is being added already have a different package name (with internal in the path) and moreover it is a part of dd-sdk-android-internal, which is added as implementation everywhere, so it is not re-exported to the public API.

I will clarify a bit. I mean that when we (sdk devs) try to search for the class in Android Studio, we don't see 2 classes with exactly the same name. I understand that `dd-sdk-android-internal` is added as implementation everywhere and its classes are not visible to the clients of the sdk.

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTest.kt
aleksandr-gringauz: I don't see why I should. The test does `verifyNoMoreInteractions` in the end. If I split it in two, the separate tests would not be able to do that.

If you have some specific idea in mind, please tell me.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/InternalTracingHeaderType.kt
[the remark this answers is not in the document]
aleksandr-gringauz: > Let's do option 1 then, I'm fine with that.

done

> idea: maybe drop Internal prefix from the name? Is it useful to have it?

done, droped internal prefix from both `TracingHeaderType` and `TracingHeaderTypesSet`

dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/HostsWithHeaderTypesForgeryFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package com.datadog.android.internal.tests.elmyr
+
+import com.datadog.android.internal.telemetry.InternalTracingHeaderType
+import com.datadog.android.internal.telemetry.InternalTracingHeaderTypesSet
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+class HostsWithHeaderTypesForgeryFactory : ForgeryFactory<InternalTracingHeaderTypesSet> {
```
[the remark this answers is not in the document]
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/HostsWithHeaderTypesMapper.kt
aleksandr-gringauz: removed the objects that I created, changed to top-level functions

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -598,5 +602,6 @@ internal class DatadogCore(
         internal val startupTimeNs: Long = System.nanoTime()
 
         internal const val OKHTTP_INTERCEPTOR_SAMPLE_RATE = "okhttp_interceptor_sample_rate"
+        internal const val OKHTTP_INTERCEPTOR_HOSTS_WITH_HEADER_TYPES = "okhttp_interceptor_hosts_with_header_types"
```
0xnm, line 605: minor: there is no hosts anymore

```suggestion
        internal const val OKHTTP_INTERCEPTOR_HEADER_TYPES = "okhttp_interceptor_header_types"
```

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -299,6 +299,7 @@ interface com.datadog.android.rum.tracking.ViewTrackingStrategy : TrackingStrate
 class com.datadog.android.sqlite.DatadogDatabaseErrorHandler : android.database.DatabaseErrorHandler
   constructor(String? = null, android.database.DatabaseErrorHandler = DefaultDatabaseErrorHandler())
   override fun onCorruption(android.database.sqlite.SQLiteDatabase)
+fun com.datadog.android.internal.telemetry.TracingHeaderTypesSet.toSelectedTracingPropagators()
```
0xnm: this function should be `internal`

0xnm (approved):
lgtm, just some minor naming update suggestions

features/dd-sdk-android-rum/api/apiSurface
aleksandr-gringauz: fixed

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
aleksandr-gringauz: fixed

0xnm (approved):

PR #2573, opened 2025-03-31
RUM-9309: Add `ActionTrackingStrategy` Interface to decouple find view logic

ambushwork:
### What does this PR do?

Prior to this PR, `GesturesListener` is coupled with Android View framework to find a target view of user interaction, making it difficult to extend to Jetpack Compose.

To make the Jetpack compose view finding possible within `GesturesListener`, this PR adds an interface called `ActionTrackingStrategy`, and move all the Android View framework logic from `GestureListener` into the `AndroidActionTrackingStrategy` implementation.  This change makes `GesturesListener` agnostic about the view framework, prepares for the Compose actions tracking integration

### Motivation

RUM-9309

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
@@ -1,0 +1,157 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.instrumentation.gestures
+
+import android.view.View
+import android.view.ViewGroup
+import android.widget.AbsListView
+import android.widget.ScrollView
+import androidx.core.view.ScrollingView
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.internal.instrumentation.gestures.GesturesListener.Companion.MSG_NO_TARGET_SCROLL_SWIPE
+import com.datadog.android.rum.internal.instrumentation.gestures.GesturesListener.Companion.MSG_NO_TARGET_TAP
+import com.datadog.android.rum.tracking.ActionTrackingStrategy
+import com.datadog.android.rum.tracking.ViewTarget
+import java.util.LinkedList
+
+/**
+ * Implementation of [ActionTrackingStrategy] for Android View, used to locate the target view
+ * with given coordinates when tapping or scrolling on Android View.
+ *
+ * @param internalLogger used to log information when finding the view.
+ */
+internal class AndroidActionTrackingStrategy(
+    private val internalLogger: InternalLogger
+) : ActionTrackingStrategy {
+
+    private val coordinatesContainer = IntArray(2)
+
+    override fun findTargetForTap(decorView: View, x: Float, y: Float): ViewTarget {
+        val queue = LinkedList<View>()
+        queue.addFirst(decorView)
+        var target: View? = null
+        var notifyMissingTarget = true
+
+        while (queue.isNotEmpty()) {
+            // removeFirst can't fail because we checked isNotEmpty
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            val view = queue.removeFirst()
+            // TODO RUM-9289: Move Jetpack Compose Logic to the Compose implementation.
+            if (queue.isEmpty() && isJetpackComposeView(view)) {
+                notifyMissingTarget = false
+            }
+
+            if (isValidTapTarget(view)) {
+                target = view
+            }
+
+            if (view is ViewGroup) {
+                handleViewGroup(view, x, y, queue, coordinatesContainer)
+            }
+        }
+
+        if (target == null && notifyMissingTarget) {
+            internalLogger.log(
+                InternalLogger.Level.INFO,
+                InternalLogger.Target.USER,
+                { MSG_NO_TARGET_TAP }
+            )
+        }
+        return ViewTarget(target)
+    }
+
+    override fun findTargetForScroll(decorView: View, x: Float, y: Float): ViewTarget? {
+        val queue = LinkedList<View>()
+        queue.add(decorView)
+
+        var notifyMissingTarget = true
+        while (queue.isNotEmpty()) {
+            // removeFirst can't fail because we checked isNotEmpty
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            val view = queue.removeFirst()
+            if (queue.isEmpty() && isJetpackComposeView(view)) {
+                notifyMissingTarget = false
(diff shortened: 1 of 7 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/ActionTrackingStrategy.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import android.view.View
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Strategy interface for tracking user interaction targets within a UI,
+ * such as taps and scrolls.
+ */
+@NoOpImplementation
+interface ActionTrackingStrategy {
```
aleksandr-gringauz, line 17: Shouldn't this class have an InternalApi annotation? Does the user of the sdk need to use it explicitly?

ambushwork: Good question, it's mainly for internal implementation of Android View and Jetpack Compose in our SDK, but we will have a public API in `RumConfiguration` for the manual instrumentation, so I don't see a strong reason that we should warn users that it is only internal use  if they really have a need to use their custom implementation.

aleksandr-gringauz (approved):

0xnm (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
```diff
@@ -0,0 +1,154 @@
...
+                InternalLogger.Level.INFO,
+                InternalLogger.Target.USER,
+                { MSG_NO_TARGET_SCROLL_SWIPE }
+            )
+        }
+
+        return null
+    }
+
+    private fun isJetpackComposeView(view: View): Boolean {
```
mariusc83, line 97: why do we mix things in the strategy ? Shouldn't you have an strategy for JepackCompose also ? In this case why do we want to skip notification in case the last view in the tree is jetpack compose view ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
```diff
@@ -0,0 +1,154 @@
...
+ * Implementation of [ActionTrackingStrategy] for Android View, used to locate the target view
+ * with given coordinates when tapping or scrolling on Android View.
+ *
+ * @param internalLogger used to log information when finding the view.
+ */
+internal class AndroidActionTrackingStrategy(
+    private val internalLogger: InternalLogger
+) : ActionTrackingStrategy {
+    private val coordinatesContainer = IntArray(2)
+    override fun findTargetForTap(decorView: View, x: Float, y: Float): ViewTarget {
```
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
ambushwork: Good point, this is from current code, I tried just move the current logic from `GestureListener` into this class without breaking anything in this PR, I added a todo on the call site of this function. It will be moved out when compose implementation is done

0xnm: This is from the existing code, not something new. This method is used just to filter out Compose wrapper in the classic view, nothing more.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
```diff
@@ -0,0 +1,154 @@
...
+package com.datadog.android.rum.internal.instrumentation.gestures
+
+import android.view.View
+import android.view.ViewGroup
+import android.widget.AbsListView
+import android.widget.ScrollView
+import androidx.core.view.ScrollingView
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.internal.instrumentation.gestures.GesturesListener.Companion.MSG_NO_TARGET_SCROLL_SWIPE
+import com.datadog.android.rum.internal.instrumentation.gestures.GesturesListener.Companion.MSG_NO_TARGET_TAP
```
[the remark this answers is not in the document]
ambushwork: These seem like the message that I can reuse for Compose implementation, so for the moment I will keep it as-is.

jonathanmos (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
mariusc83: I understand this part is just one thing I do not understand: if we provide a `JetpackActionTrackingStrategy` then in your application configuration you will either use one or another. If I have Jetpack views I will use `JetpackActionTrackingStrategy` so there is no need to check for jetpack compose view in the default strategy. Am I missing something ?

ambushwork: Here the impact is only a log message, `AndroidActionTrackingStrategy` will not able to locate the exact Compose View, but it should not be considered that we miss the target, so we don't send a log to the user.
This is the current log, it can be optimised when `ComposeActionTrackingStrategy` comes in

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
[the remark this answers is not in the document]
0xnm: still there, it seems. but anyway, not blocking.

0xnm (approved):

mariusc83 (approved):

0xnm (dismissed):

0xnm (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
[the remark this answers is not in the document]
ambushwork: updated

PR #2575, opened 2025-04-01
RUM-9297: Add Compose actions tracking strategy API in RumConfiguration

ambushwork:
### What does this PR do?

Add API `setComposeTrackingStrategy` in RumConfiguration which allows user to add `ComposeActionsTrackingStategy`.

So that `GestureListener` will be able to use this strategy to locate target in Jetpack Compose.

### Motivation

RUM-9297


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -311,6 +312,22 @@ data class RumConfiguration internal constructor(
...
+         * @param composeActionTrackingStrategy actions tracking strategy for Jetpack compose.
+         */
+        fun setComposeActionTrackingStrategy(
+            composeActionTrackingStrategy: ActionTrackingStrategy
+        ): Builder {
+            rumConfig = rumConfig.copy(
+                composeActionTrackingStrategy = composeActionTrackingStrategy
+            )
+            return this
+        }
```
0xnm, line 329: Maybe we should define this function as extension function in the `dd-sdk-android-compose` module and then use internal proxy to set things in `RumConfiguration.Builder`?

My concern here is the leak of Compose-topic into non-Compose module.

Also it is written that we should use `ComposeActionTrackingStategy`, but method has argument of type `ActionTrackingStrategy`. So we kind of saying by type definition that we allow any implementation, but in reality we expect only one.

By moving it to the extension function in `dd-sdk-android-compose` we can have a single function like `enableComposeActionTracking` without strategy argument (if we expect a concrete instance) or `setComposeActionTrackingStrategy` with default `ComposeActionTrackingStategy` instance where type can be open for the extension if needed.

WDYT?

aleksandr-gringauz: > By moving it to the extension function in dd-sdk-android-compose we can have a single function like enableComposeActionTracking without strategy argument

I like this idea. To me it looks like the user doesn't need to know anything about `ComposeActionTrackingStrategy`, it should be internal and `enableComposeActionTracking` should create and add it to the [configuration](https://github.com/DataDog/dd-sdk-android/blob/05e490a9704f80842738d4035774962c66e51520/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt#L602) of Rum feature.

ambushwork: Good Idea, I updated the function name and document to get rid of "Compose" in RumConfiguration, and mark it as `@InternalApi`

0xnm: It still should be a different way: the function itself should be `internal` and be called via `_RumInternalProxy` from `dd-sdk-android-compose`.

0xnm: This one is only for the action naming/attributes extraction logic. I'm not sure it answers the need.

ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -726,11 +733,17 @@ internal class RumFeature(
...
+            composeActionTrackingStrategy: ActionTrackingStrategy,
             internalLogger: InternalLogger
         ): DatadogGesturesTracker {
             val defaultProviders = arrayOf(JetpackViewAttributesProvider())
             val providers = customProviders + defaultProviders
-            return DatadogGesturesTracker(providers, interactionPredicate, internalLogger)
+            return DatadogGesturesTracker(
+                providers,
+                interactionPredicate,
+                composeActionsTrackingStrategy = composeActionTrackingStrategy,
```
0xnm, line 744:
```suggestion
                composeActionTrackingStrategy,
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -365,6 +366,17 @@ data class RumConfiguration internal constructor(
...
 
+        /**
+         * Set custom [ActionTrackingStrategy] RUM actions tracking.
+         *
+         * @param actionTrackingStrategy custom actions tracking strategy.
+         */
+        internal fun setCustomActionTrackingStrategy(
+            actionTrackingStrategy: ActionTrackingStrategy
+        ): Builder {
+            rumConfig = rumConfig.copy(composeActionTrackingStrategy = actionTrackingStrategy)
```
0xnm, line 377: there is a bit of misalignment here: method is called `setCustomActionTrackingStrategy` but it is setting `composeActionTrackingStrategy`, not something generic. Should it be renamed?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -708,12 +713,14 @@ internal class RumFeature(
...
             touchTargetExtraAttributesProviders: Array<ViewAttributesProvider>,
             interactionPredicate: InteractionPredicate,
+            composeActionTrackingStrategy: ActionTrackingStrategy,
             internalLogger: InternalLogger
         ): UserActionTrackingStrategy {
             val gesturesTracker =
                 provideGestureTracker(
                     touchTargetExtraAttributesProviders,
                     interactionPredicate,
+                    composeActionTrackingStrategy = composeActionTrackingStrategy,
```
0xnm, line 723:
```suggestion
                    composeActionTrackingStrategy,
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/DatadogGesturesTracker.kt
```diff
@@ -106,7 +108,8 @@ internal class DatadogGesturesTracker(
                 targetAttributesProviders,
                 interactionPredicate,
                 WeakReference(context),
-                internalLogger
+                composeActionTrackingStrategy = composeActionsTrackingStrategy,
+                internalLogger = internalLogger
```
0xnm, line 112: do we need to use named arguments if variable name matches?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumConfigurationBuilderTest.kt
```diff
@@ -78,39 +81,43 @@ internal class RumConfigurationBuilderTest {
...
-                vitalsMonitorUpdateFrequency = VitalsUpdateFrequency.AVERAGE,
-                sessionListener = NoOpRumSessionListener(),
-                additionalConfig = emptyMap(),
-                initialResourceIdentifier = TimeBasedInitialResourceIdentifier(),
-                lastInteractionIdentifier = TimeBasedInteractionIdentifier(),
-                trackAnonymousUser = true,
-                slowFramesConfiguration = null
-            )
-        )
+        rumConfiguration.featureConfiguration.apply {
```
0xnm, line 84: minor: maybe `with(rumConfiguration.featureConfiguration)` it will be more descriptive? like to signify that we are not going to mutate it.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/RumConfigurationExt.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.datadog.android.compose
+
+import com.datadog.android.rum.RumConfiguration
+import com.datadog.android.rum._RumInternalProxy
+import com.datadog.android.rum.tracking.ActionTrackingStrategy
+
+/**
+ * Enable [ActionTrackingStrategy] for Jetpack Compose automatic actions tracking, such as
```
0xnm, line 14: probably we should remove mention of `ActionTrackingStrategy` from this public doc, because it is not something that customer may set in the public API and it is not supposed to be used directly.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/RumConfigurationExt.kt
```diff
@@ -0,0 +1,25 @@
...
+ * tap and scroll. Jetpack Compose actions tracking will be disabled by not calling this API,
+ * which is the default behavior.
```
0xnm, line 16:
```suggestion
 * tap and scroll. Jetpack Compose actions tracking will be disabled if this API is not called,
 * which is the default behavior.
```

integrations/dd-sdk-android-compose/src/test/kotlin/ComposeActionTrackingStrategyTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.mockito.Mock
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.quality.Strictness
+
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
```
0xnm, line 22: probably it should go before imports?

0xnm (commented):
overall lgtm! I've added few comments and suggestions.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/DatadogGesturesTracker.kt
ambushwork: Personally I prefer always adding named argument for function call, for mainly these reasons:
* avoid argument orders mistakes if some arguments have same type.
* more readable in PR on GH, especially the variable name doesn't match exactly with argument name.
* be alerted with the silent break of argument name in the function declaration.

Maybe none of these advantages apply here in this PR, but I think it's a good thing to have more in our code base, WDYT?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
ambushwork: I was thinking about it, on one hand we want to hide the `compose` concept in the public API, but on the other hand, in the end `GestureListener` where this `composeActionTrackingStrategy` is used, actually should know about it is exactly a compose strategy. so we need a place to do this kind of naming transformation.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/DatadogGesturesTracker.kt
0xnm: yes, sure, makes sense. just then maybe we can make it consistent and add all names for the remaining arguments in this function call?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
0xnm: Yes, but this is not a public API anymore, because it is `internal` method, so we can simply call it `setComposeActionTrackingStrategy`. And we can rename it at any time if needed.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
ambushwork: updated with full named arguments

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
ambushwork: updated with full named arguments

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/DatadogGesturesTracker.kt
ambushwork: updated with full named arguments

0xnm (approved):
nice!

PR #2579, opened 2025-04-02
RUM-9340: Fix RumAction Tap is added twice for every ACTION_UP

ambushwork:
### What does this PR do?

Prior to this PR, `RumMonitor.addActions(RumActionType.Tap)` will always be called twice for a single tap action, this is because that in `GestureListener` both `onSingleTapUp` and `onUp` are called by `GesturesDetectorWrapper` when a ACTION_UP is received. both functions will call `addActions` for this tap.

The fix is to make `onSingleTapUp` return true, indicating that the tap is already consumed and should not be handled it again.

### Motivation

RUM-9340

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesDetectorWrapper.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesDetectorWrapper.kt
@@ -24,7 +24,9 @@ internal class GesturesDetectorWrapper(
     )
 
     fun onTouchEvent(event: MotionEvent) {
-        defaultGesturesDetector.onTouchEvent(event)
+        if (defaultGesturesDetector.onTouchEvent(event)) {
+            return
+        }
         val action = event.actionMasked
         if (action == MotionEvent.ACTION_UP) {
             gestureListener.onUp(event)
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesDetectorWrapper.kt
```diff
@@ -24,7 +24,9 @@ internal class GesturesDetectorWrapper(
     )
 
     fun onTouchEvent(event: MotionEvent) {
-        defaultGesturesDetector.onTouchEvent(event)
+        if (defaultGesturesDetector.onTouchEvent(event)) {
```
mariusc83, line 27: oh I see so returning true from the `onSingleTapUp` will fix your code here. Nice one ;)

mariusc83 (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
```diff
@@ -53,7 +53,7 @@ internal class GesturesListener(
     override fun onSingleTapUp(e: MotionEvent): Boolean {
         val decorView = windowReference.get()?.decorView
         handleTapUp(decorView, e)
-        return false
+        return true
```
0xnm, line 56: are you sure this has no impact on the customer side? normally we shouldn't consume events, but just read them and let them propagate further. And it is customer code to decide where it should be consumed.

ambushwork: If I understand correctly, we use `WindowCallbackWrapper` as the interceptor of touch event, and indeed in `WindowCallbackWrapper` we always return false to let event propagate. 

`GesturesListner` derives from  `WindowCallbackWrapper` to handle the touch event, and its return value is not taken into account the final return value of `WindowCallbackWrapper`, so I don't think it will affect user behavior.

But since it's a quite impactful change, it will be very helpful if you can double check in the code if what I am saying is correct.

Also cc @mariusc83

jonathanmos (approved):

PR #2586, opened 2025-04-07
RUM-9298: Implement Compose actions tracking strategy

ambushwork:
### What does this PR do?

Implement `ComposeActionsTrackingStrategy` to find the target of action in Jetpack Compose.

The main idea of the whole process:
* `GestureListener` receives the motion event first and start to iterate the view tree from the decor view of the window.
* During the iteration of the view tree we use android and compose strategy separately:
     * Use `AndroidActionTrackingStrategy` to simply test the view is clickable/scrollable and hit by the action.
     * Use `ComposeActionTrackingStrategy` to continue the iteration if the view is a `Owner` of compose layout node.
* In `ComposeActionTrackingStrategy`, there is no more view tree, instead we can still iterate the layout node tree to find out which node is clickable/scrollable and hit by the motion event coordinates.

### Motivation

RUM-9298

### Additional Notes

Just in case of confusion, there are some back and forth from previous PR, the logic of iterating the view tree was moved out from `GesturesListener` to `AndroidActionTrackingStrategy`, now it is moved back to `GestureListener`, because before finding the compose view, we still need to iterate from `DecorView` to `AndroidComposeView`, so this part should always be the common logic between Android view and Jetpack Compose.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
```diff
@@ -121,14 +121,68 @@ internal class GesturesListener(
...
+            )
+        }
+        return target
     }
 
-    private fun findTargetForTap(decorView: View, x: Float, y: Float): ViewTarget? {
-        return androidActionTrackingStrategy.findTargetForTap(decorView, x, y)
-            ?: composeActionTrackingStrategy.findTargetForTap(decorView, x, y)
+    private fun findTargetForScroll(view: View, x: Float, y: Float): ViewTarget? {
+        // return most inside scrollable element
```
0xnm, line 165: is there a missing word here? same for the similar comment below

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActionTrackingStrategy.kt
```diff
@@ -19,22 +19,22 @@ interface ActionTrackingStrategy {
     /**
      * Finds the target element at the given screen coordinates in response to a tap action.
      *
-     * @param decorView The root [View] of the current window hierarchy.
+     * @param view The root [View] of finding the target.
```
0xnm, line 22:
```suggestion
     * @param view The root [View] to start search from, top-down.
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActionTrackingStrategy.kt
```diff
@@ -19,22 +19,22 @@ interface ActionTrackingStrategy {
...
      *         or `null` if no target was found.
      */
-    fun findTargetForTap(decorView: View, x: Float, y: Float): ViewTarget?
+    fun findTargetForTap(view: View, x: Float, y: Float): ViewTarget?
 
     /**
      * Finds the target element at the given screen coordinates in response to a scroll action.
      *
-     * @param decorView The root [View] of the current window hierarchy.
+     * @param view The root [View] of finding the target.
```
0xnm, line 33:
```suggestion
     * @param view The root [View] to start search from, top-down.
```

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/ComposeActionTrackingStrategy.kt
```diff
@@ -0,0 +1,91 @@
...
+        val queue: Queue<LayoutNode> = LinkedList()
+        try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // Any exception will be caught.
+            queue.add(owner.root)
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+            // All exceptions should be caught and return in case of crash.
+            logAddQueueException(e)
+            // The iteration of the layout nodes tree has failed, return null.
+            return null
+        }
```
0xnm, line 57: in which case we can end up here? like `owner.root` call can throw?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/ComposeActionTrackingStrategy.kt
```diff
@@ -0,0 +1,91 @@
...
+        isScrollEvent: Boolean
+    ): ViewTarget? {
+        val queue: Queue<LayoutNode> = LinkedList()
+        try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // Any exception will be caught.
+            queue.add(owner.root)
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+            // All exceptions should be caught and return in case of crash.
+            logAddQueueException(e)
+            // The iteration of the layout nodes tree has failed, return null.
```
0xnm, line 55: is it `owner.root` call who does the iteration under the hood?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/ComposeActionTrackingStrategy.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private fun hitTest(layoutNode: LayoutNode, x: Float, y: Float): Boolean {
+        return layoutNodeUtils.getLayoutNodeBoundsInWindow(layoutNode)?.let { bounds ->
+            x >= bounds.left && x <= bounds.right && y >= bounds.top && y <= bounds.bottom
+        } ?: false
+    }
+
+    private fun logAddQueueException(e: Exception) {
+        (Datadog.getInstance() as? FeatureSdkCore)?.internalLogger?.log(
```
0xnm, line 84: We should use the instance for the feature. Or is it like that because of the initialization sequence?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    private fun <T> runSafe(action: () -> T): T? {
+        try {
+            return action()
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+            // We rely on visibility suppression to access internal field,
+            // any runtime exception must be caught here.
+            (Datadog.getInstance() as? FeatureSdkCore)?.internalLogger?.log(
```
0xnm, line 81: same - is it possible to use feature-specific instance?

integrations/dd-sdk-android-compose/src/test/kotlin/ComposeActionTrackingStrategyTest.kt
```diff
@@ -34,38 +46,302 @@ class ComposeActionTrackingStrategyTest {
...
+
+    /**
+     *           0
+     *         /   \
+     *        1(c)  2
+     *      /     \
+     *     3(c)   4
+     */
+    @Test
+    fun `M return most inside clickable node W findTargetForTap`(forge: Forge) {
```
0xnm, line 98: "most inside" -> maybe "most relevant" or "lowest level"?

0xnm (commented):
nice work! I left some questions, bit overall lgtm.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/ComposeActionTrackingStrategy.kt
ambushwork: we have instance per feature? how to get it from here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
0xnm: I think that "most inside" is not the term to use in such case. Maybe "bottom-most" then? Because there is for sure "top-most" which exists.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/ComposeActionTrackingStrategy.kt
0xnm: yes, you normally have and instance of `FeatureSdkCore` passed during the `RumFeature` initialization, and this is the instance which should be used.

Is there a chance to get it here? But I understand, that it may be not available when we just create a configuration object.

ambushwork: yes we create it when building the configuration, so we don't have the instance of `FeatureSdkCore` here

0xnm: maybe it is worth then also implement `com.datadog.android.rum.tracking.TrackingStrategy` interface? There you can get `SdkCore` instance, you can check how `ActivityLifecycleTrackingStrategy` is using it, for example.

ambushwork: updated, now I make `ActionTrackingStrategy` extend `TrackingStrategy` and the registration of `sdkCore` happens in `GesturesListener`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
```diff
@@ -34,19 +37,24 @@ internal class GesturesListener(
...
+    private var scrollTargetReference: WeakReference<ViewTarget?> = WeakReference(null)
     private var onTouchDownXPos = 0f
     private var onTouchDownYPos = 0f
 
     // region GesturesListener
 
+    init {
+        contextRef.get()?.let {
+            composeActionTrackingStrategy.register(sdkCore, it)
+            androidActionTrackingStrategy.register(sdkCore, it)
```
0xnm, line 54: normally we need to unregister as well, but I guess if SDK core is stopped, then it won't be referenced from user code and even though it will have a reference from action tracking strategies, it will be still deallocated.

integrations/dd-sdk-android-compose/src/test/kotlin/ComposeActionTrackingStrategyTest.kt
```diff
@@ -34,38 +46,302 @@ class ComposeActionTrackingStrategyTest {
...
+
+    /**
+     *           0
+     *         /   \
+     *        1(c)  2
+     *      /     \
+     *     3(c)   4
+     */
+    @Test
+    fun `M return lowest level clickable node W findTargetForTap`(forge: Forge) {
```
0xnm, line 98: minor: let's use `bottom-most` instead of `lowest level`, seems like it is the opposite of the quite common `top-most`.

integrations/dd-sdk-android-compose/src/test/kotlin/ComposeActionTrackingStrategyTest.kt
```diff
@@ -34,38 +46,302 @@ class ComposeActionTrackingStrategyTest {
...
+
+    /**
+     *           0
+     *         /  \
+     *       1(c)  2
+     *      /    \
+     *     3(c)   4
+     */
+    @Test
+    fun `M return most inisde scrollable node W findTargetForScroll`(forge: Forge) {
```
0xnm, line 126:
```suggestion
    fun `M return bottom-most scrollable node W findTargetForScroll`(forge: Forge) {
```

0xnm (approved):

0xnm (approved):

PR #2601, opened 2025-04-15
RUM-9427: Add telemetry for compose instrumentation functions

ambushwork:
### What does this PR do?

Before releasing the compose auto instrumentation, we need to add telemetry about the usage of auto instrumentation.
Since it is almost impossible to add telemetry directly, the idea of this PR is to send telemetry when the instrumented function is call.

Main changes in this PR:
* Before we have two public APIs `datadog` and `NavigationViewTrackingEffect` which can be called by user manually and also instrumented by KCP automatically. Now `instrumentedDatadog` and `instrumentedNavigationViewTrackingEffect` internal functions are added, and they are reserved to KCP for auto instrumentation. The public APIs stay the same.
* The public APIs and internal functions will execute the same core functions under the hood, they only difference is that they will send telemetry with different attributes to indicate the way of instrumentation (manual or auto).
* Two functions send telemetry will different messages, both of them are sent only once.
image

* There are two additional attributes:
image

### Motivation

RUM-9427

### DIFF
--- a/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation.kt
+++ b/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation.kt
@@ -4,12 +4,13 @@
  * Copyright 2016-Present Datadog, Inc.
  */
 
-@file:Suppress("MatchingDeclarationName")
+@file:Suppress("MatchingDeclarationName", "PackageNameVisibility")
 
 package com.datadog.android.compose
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.NonRestartableComposable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.rememberUpdatedState
@@ -19,6 +20,8 @@ import androidx.navigation.NavDestination
 import com.datadog.android.Datadog
 import com.datadog.android.api.SdkCore
 import com.datadog.android.compose.internal.ComposeNavigationObserver
+import com.datadog.android.compose.internal.InstrumentationType
+import com.datadog.android.compose.internal.sendTelemetry
 import com.datadog.android.rum.GlobalRumMonitor
 import com.datadog.android.rum.tracking.AcceptAllNavDestinations
 import com.datadog.android.rum.tracking.ComponentPredicate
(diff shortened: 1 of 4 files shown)


integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation.kt
```diff
@@ -41,6 +43,51 @@ fun NavigationViewTrackingEffect(
     trackArguments: Boolean = true,
     destinationPredicate: ComponentPredicate<NavDestination> = AcceptAllNavDestinations(),
     sdkCore: SdkCore = Datadog.getInstance()
+) {
+    sendTelemetry(autoInstrumented = false, InstrumentationType.ViewTracking)
```
0xnm, line 47: is it ok to not wrap this in `remember`? won't we have this called many times?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation.kt
```diff
@@ -41,6 +43,51 @@ fun NavigationViewTrackingEffect(
...
+    InternalNavigationViewTrackingStrategy(
```
0xnm, line 48: I think this is used only here? Do we need to have a dedicated composable or we may just inline it?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/Telemetry.kt
```diff
@@ -0,0 +1,45 @@
...
+        messageBuilder = { message },
+        onlyOnce = true,
+        additionalProperties = attributes
+    )
+}
+
+internal enum class InstrumentationType {
+    Semantics,
+    ViewTracking
+}
```
0xnm, line 36: This will be obfuscated in the release build, so `${instrumentationType.name}` will give obfuscated value.

We have to define strict string literals for the values.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation.kt
ambushwork: `InternalNavigationViewTrackingStrategy` is used both by public API and internal instrumented function

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/Telemetry.kt
```diff
@@ -0,0 +1,45 @@
...
+    instrumentationType: InstrumentationType
+) {
+    val message = "$DATADOG_SEMANTICS_TELEMETRY_LOG: ${instrumentationType.value}"
+    val attributes = mapOf(
+        KEY_COMPOSE_INSTRUMENTATION to mapOf(
+            KEY_ENABLED to autoInstrumented,
+            KEY_INSTRUMENTATION_TYPE to instrumentationType
+        )
+    )
+    (Datadog.getInstance() as? FeatureSdkCore)?.internalLogger?.log(
```
0xnm, line 24: maybe we should have `sdkCore = Datadog.getInstance()` argument for this function? In case of modifiers we don't have an access to `SdkCore`, but in case of navigation effect and interaction tracking we do.

0xnm (dismissed):

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/Telemetry.kt
```diff
@@ -0,0 +1,45 @@
...
+        level = InternalLogger.Level.INFO,
+        target = InternalLogger.Target.TELEMETRY,
+        messageBuilder = { message },
+        onlyOnce = true,
```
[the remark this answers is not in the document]
ambushwork: you mean give an additional sample rate on top of Rum telemetry sample rate?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/Telemetry.kt
ambushwork: updated.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/Telemetry.kt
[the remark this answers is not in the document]
0xnm: yes, since here it will be invoked on every usage of the component, potentially, so it may still have a big volume.

0xnm (approved):

PR #2602, opened 2025-04-15
RUM-8366 propagate session id in baggage header

xgouchet:
### What does this PR do?

Update the header used to propagate the RUM session ID when tracing


features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/propagation/DatadogHttpCodec.java
```diff
@@ -77,13 +78,11 @@ public void inject(final DDSpanContext context, final TextMapInject carrier) {
...
             tags.append('=');
             tags.append(mostSignificantTraceId);
+            carrier.put(DATADOG_TAGS_KEY, tags.toString());
+
             if(sessionId != null) {
-                tags.append(',');
-                tags.append(RUM_SESSION_ID_TAG_KEY);
-                tags.append('=');
-                tags.append(sessionId);
+                carrier.put(BAGGAGE_KEY, RUM_SESSION_ID_BAGGAGE_KEY + '='+sessionId);
```
0xnm, line 84: question: we already have `OT_BAGGAGE_PREFIX` (`ot-baggage-`) here, which is used with a particular baggage item key.

won't we have any issues with having `ot-baggage-xxx` and `baggage` at the same time at the same time?

0xnm, line 84: do we need to have a similar change for the `com.datadog.trace.core.propagation.DatadogHttpCodec`, etc.?

xgouchet: Indeed, done

jonathanmos (dismissed):

0xnm (dismissed):

0xnm (approved):

PR #2607, opened 2025-04-17
RUMS-4305 Fix the memory leak in the `PendingTrace#cleaner`

mariusc83:
### What does this PR do?

After analysing the borrowed code from the deprecated APM java tracer following some internal investigations `PendingTrace` was actually leaking in 2 places:

1. In one place when using `dropSpan` the cleaner could not actually clean the `PendingTrace` reference and this was persisted forever in the `pendingTraces` set. Basically when the cleaner was trying to `clean` that method was doing nothing as there was no pending Span to write and the reference was not removed. 
2. Because of using a mutable `LinkedList` as an entry key to a `Set<PendingTrace>`. Basically because the key is resolved from the `LinkedList#hashCode()` method which is dependent on the values contained in the list first time a `PendingTrace` is created the `hashCode()` is always `1` and when later we're trying to remove it through `clean` the `hashCodes` no longer match. 
The problem was not that visible when instrumenting the `httpRequest` as the `hashCode` when adding the list was always `1` so only 1 `PendingTrace` was persisted but it became more visible when using a very trivial example as:
```
val tracer = GlobalTracer.get()
            repeat(100){
                val span = tracer.buildSpan("AsyncOperation")
                    .start()
                Thread.sleep(1)
                span.finish()
            }
```
In this moment we're going to leak 200 `PendingTraces` because:

- when adding each first `PendingTrace` the set will contain `[1]->[(pendingTrace1, key1)]`
- when adding second `PendingTrace` the set will check if key `1` exist and then will equal the keys by values but because at the lists are equaling the elements in the `equal` method in this moment the first trace has 1 element and the one to be added has 0 so they do not match so the set will contain `[1]->[(pendingTrace1, key1), (pendingTrace2, key2)]
- when later trying to remove each of these it will first try to match the `hashCode` but because they all have elements in them the `hashCode` will not longer match `1` key so the will all be leaked.
This problem is not reproducible when using the `DatadogInterceptor` because there we are using `drop` instead of `finish` in the Span and this will not add the span into the `PendingTrace` a.k.a `LinkedList`. In that situation the `pendingTraces` set will only contain one empty `PendingTrace`: `[1]->[(PendingTrace, key)]`. Every time a new `PendingTrace` will be created it will match through `hashCode` and `equal` to the one in the `Set` and it will not be added. In this situation only the first added `PendingTrace` will be leaked and this is because of the reason number 1 described above.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
+++ b/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
@@ -6,6 +6,8 @@
 
 package com.datadog.opentracing;
 
+import androidx.annotation.VisibleForTesting;
+
 import com.datadog.android.api.InternalLogger;
 import com.datadog.exec.CommonTaskExecutor;
 import com.datadog.exec.CommonTaskExecutor.Task;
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/PendingTraceTest.kt
```diff
@@ -67,6 +81,175 @@ class PendingTraceTest {
...
+        }
+        countDownLatch.await(20000, TimeUnit.MILLISECONDS)
+
+        // When
+        Thread {
+            PendingTrace.getSpanCleaner().run()
+        }.apply {
+            start()
+            join(20000)
+        }
```
0xnm, line 166: we shouldn't have a timeout here, but isn't 20000ms too much for such repetitions count? probably even 1s will be enough.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/PendingTraceTest.kt
```diff
@@ -67,6 +81,175 @@ class PendingTraceTest {
...
+
+        // When
+        PendingTrace.getSpanCleaner().run()
+
+        // Then
+        assertThat(PendingTrace.getPendingTracesSize()).isEqualTo(0)
+    }
+
+    @Test
+    fun `M not leak the PendingTraces W finishSpan from different threads`(
```
0xnm, line 124:
```suggestion
    fun `M not leak the PendingTraces W finish from different threads`(
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/PendingTraceTest.kt
```diff
@@ -67,6 +81,175 @@ class PendingTraceTest {
         }
     }
 
+    @Test
+    fun `M not leak the PendingTraces W finishSpan`(
```
0xnm, line 85:
```suggestion
    fun `M not leak the PendingTraces W finish`(
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/PendingTraceTest.kt
```diff
@@ -67,6 +81,175 @@ class PendingTraceTest {
...
+            start()
+            join(20000)
+        }
+
+        // Then
+        assertThat(PendingTrace.getPendingTracesSize()).isEqualTo(0)
+    }
+
+    @Test
+    fun `M not leak the PendingTraces W dropSpan`(
```
0xnm, line 173:
```suggestion
    fun `M not leak the PendingTraces W drop`(
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,184 @@
...
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(value = Configurator::class)
+internal class SpanCleanerTest {
+
+    private lateinit var testeCleaner: SpanCleaner
+
+    @BeforeEach
+    fun setUp() {
+        testeCleaner = SpanCleaner()
+    }
```
0xnm, line 43:
```suggestion
    private val testedCleaner: SpanCleaner = SpanCleaner()
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,184 @@
...
+    // region cleanup
+
+    @Test
+    fun `M remove random traces W add and remove`(forge: Forge) {
+        // Given
+        val pendingTraces = forge.aList(size = forge.anInt(min = 10, max = 50)) { mock<PendingTrace>() }
+        pendingTraces.forEach {
+            testeCleaner.addPendingTrace(it)
+        }
+        assertThat(testeCleaner.pendingTraces.values).containsExactlyInAnyOrderElementsOf(pendingTraces)
```
0xnm, line 54: should it be `assume` here as precondition?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,184 @@
...
+
+    // region concurrency
+
+    @Test
+    fun `M not leak any PendingTrace W add, mutate and remove { different threads }`(forge: Forge) {
+        // Given
+        val pendingTraces = forge.aList(size = forge.anInt(min = 10, max = 50)) { mockPendingTrace(forge) }
+        val addCountDownLatch = CountDownLatch(pendingTraces.size)
+        val removeCountDownLatch = CountDownLatch(pendingTraces.size)
+        pendingTraces.take(pendingTraces.size).forEach { pendingTrace ->
```
0xnm, line 129: I'm confused by the `pendingTraces.take(pendingTraces.size)` -  it is essentially the same as without `take` call, no? It is effectively the same collection. Should it be `pendingTraces.take(pendingTraces.size / 2)` here?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,184 @@
...
+                if (forge.aBool()) {
+                    repeat(forge.anInt(min = 1, max = 10)) {
+                        pendingTrace.addSpan(mock())
+                    }
+                }
+                testeCleaner.addPendingTrace(pendingTrace)
+                repeat(forge.anInt(min = 1, max = 10)) {
+                    pendingTrace.addSpan(mock())
+                }
+                addCountDownLatch.countDown()
```
0xnm, line 149: there is a small difference with a block above, it is using `if (forge.aBool())`, is it intended?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,184 @@
...
+            }.start()
+        }
+        addCountDownLatch.await(3, TimeUnit.SECONDS)
```
0xnm, line 152: we will hit this timeout if for the 2 calls to `forEach` above each time we take `pendingTraces.size / 2` elements. If number of elements is odd, `addCountDownLatch` will never reach 0.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
mariusc83: Let me think...

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
mariusc83: yeah I just wanted to test random case where the pending trace has something before was added, maybe should I add 2 tests here ? Isn't going to be too much ?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
mariusc83: Hmm not sure I want that, I want to start fresh for each test.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
mariusc83: I don't like `assume` as it doesn't have all the list assertions :(

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
0xnm: it will start fresh for each test - each test will have its own enclosing instance of the test class

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
mariusc83: yup because that means the cleaner doesn't work correctly.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,187 @@
...
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(value = Configurator::class)
+internal class SpanCleanerTest {
+
+    private lateinit var testeCleaner: SpanCleaner
```
ambushwork, line 39: typo: ~~testeCleaner~~ -> testedCleaner

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
```diff
@@ -25,410 +27,469 @@
...
-  private static class SpanCleanerTask implements Task<SpanCleaner> {
+    /**
+     * This class is used to create a unique key for the pending trace map.
+     * It uses the identity hash code of the object to create hash code.
+     * In case of hash code collision, it compares the object references to ensure uniqueness.
+     * The JVM system guarantees that `System.identityHashCode` will return same value for the lifetime of the
+     * object. This will ensure that add/remove operations are legit for the same object.
+     */
+    @VisibleForTesting
+    static class IdentityKey {
```
ambushwork, line 460: I am not sure that I understand the logic of `IdentityKey` here, for me when comparing two instances of `IdentityKey`,  you are still comparing if the `pendingTraces` that put inside are the same objects in the end. so for me you can just use a `Set<PendingTrace>` to store it as before.

To verify this, I pulled your branch, and replaced `Map<IdentityKey, PendingTrace>` by `Set<PendingTrace>` and then ran the `SpanCleanerTest` that you added, all the tests still pass, so it proves that using `Set<PendingTraces>` will work.

Do I miss something here?

mariusc83: Nope, that doesn't work, I need to actually add maybe an extra unit test for that case. Basically what happens if you use a `Set<PendingTrace>` it will first use the `PendingTrace#hashCode` method to generate a key for the bucket. That already brakes the set because the `hashCode` will return different values if the `PendingTrace` is empty or has some elements in it so when trying to remove it will no longer work. The other problem is the `equals` method which right now it uses `equals` on the `LinkedList/PendingTrace` elements so in the case for example I am adding 2 `PendingTraces` with the same spans in them to the set it will see them as same and it will drop the second one which is not what we want as they are 2 different traces.

ambushwork (approved):
LGTM

xgouchet (dismissed):
Nice work

features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java
```diff
@@ -25,410 +27,469 @@
...
+                // preserve throughput count.
+                // Don't report the trace because the data comes from buggy uses of the api and is suspect.
+                tracer.incrementTraceCount();
+            }
+            count++;
+            expireReference();
+        }
+        if (count > 0) {
+            // TODO attempt to flatten and report if top level spans are finished. (for accurate metrics)
+        } else if (pendingReferenceCount.get() == 0) {
```
0xnm, line 347: I'm not quite sure about this change: let's say we created a new `PendingTrace` without any spans inside and we add spans only _after_ first cycle of the `SPAN_CLEANER` is completed - with this change `removePendingTrace` will be called even before we manage to register first span, no? 🤔

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/PendingTraceTest.kt
```diff
@@ -89,7 +272,6 @@ class PendingTraceTest {
                         "childSpan_$i",
                         internalLogger
                     )
-                    pendingTrace.registerSpan(span)
```
0xnm, line 92: what is the need for this change?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,204 @@
...
+        // Given
+        val pendingTraces = forge.aList(size = forge.anInt(min = 10, max = 50)) { mock<PendingTrace>() }
+        pendingTraces.forEach {
+            testedCleaner.addPendingTrace(it)
+        }
+        Assumptions.assumeTrue(testedCleaner.pendingTraces.values.size == pendingTraces.size)
+        Assumptions.assumeTrue(testedCleaner.pendingTraces.values.containsAll(pendingTraces))
+
+        // When
+        pendingTraces.shuffled().forEach {
```
0xnm, line 53:
```suggestion
        pendingTraces.shuffled(Random(forge.seed)).forEach {
```

but what is the need to call `shuffled` anyway? to have a different order between `addPendingTrace` and `removePendingTrace` calls?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
```diff
@@ -0,0 +1,204 @@
...
+                    }
+                }
+                testedCleaner.addPendingTrace(pendingTrace)
+                repeat(forge.anInt(min = 1, max = 10)) {
+                    pendingTrace.addSpan(mock())
+                }
+                addCountDownLatch.countDown()
+            }.start()
+        }
+        addCountDownLatch.await(3, TimeUnit.SECONDS)
```
0xnm, line 172: `addCountDownLatch` will never reach zero, because we initialize it with `pendingTraces.size` counter, but `countDown` is called only `pendingTraces.size - half` times.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/SpanCleanerTest.kt
mariusc83: yup

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/opentracing/PendingTraceTest.kt
mariusc83: the `registerSpan` it is already called in the constructor of `DdSpan`

PR #2611, opened 2025-04-23
RUM-7924: Fix Session Replay is not resumed after the session has expired before

ambushwork:
### What does this PR do?

This is a bug fix based on a support ticket saying they don't see session replay for some of their sessions, after investigating there some key points which lead to the issue.
* All the sessions missing replay are created after the expiration of previous sessions
* Client set `backgroundTrackingEnabled` to false in `RumConfiguration`
* Client have sampleRate 100% and never call start/stop recording manually.

After looking into code, it turns out that the expiration of the session will make `keepSession` false, and stop recording in `SessionReplayFeature`, when `keepSession` becomes true due to new user interactions, session replay will not be able to resume.

This PR fixes this case and add a unit test to protect against this corner case. To achieve that, there are some import flag changes in `SessionReplayFeature`:
* `shouldRecord` flag now becomes a definitive state if the recording should be started or stopped, meaning that it will be decided by the other flags (`userOption`, `keepSessions`, `isSampled`), it should not be changed directly outside `modifyShouldRecordState` function
* `userOption` flag is added to indicate the current user decision about recording, it uses `startRecordingImmediately` as default value, and can be set to true if user calls `manuallyStartRecording` and can be set to false if user calls `manuallyStopRecording`.

now in `SessionReplayFeature` the state flow becomes:

image

### Motivation

RUM-7924

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
@@ -99,11 +99,15 @@ internal class SessionReplayFeature(
     private lateinit var appContext: Context
 
     // should we record the session - a combination of rum sampling, sr sampling
-    // and sr stop/start state
-    private val shouldRecord = AtomicBoolean(startRecordingImmediately)
+    // and user option.
+    private val shouldRecord = AtomicBoolean(false)
 
-    // used to monitor changes to an active session due to manual stop/start
-    private val recordingStateChanged = AtomicBoolean(false)
+    // Indicates the user's intend on recording, it starts with `startRecordingImmediately`
+    // in configuration, can be changed by calling start/stop recordings API.
+    private val userIntentToRecord = AtomicBoolean(startRecordingImmediately)
+
+    // used to monitor changes to user's intend on recording state
+    private val userIntentToRecordChanged = AtomicBoolean(false)
 
     // are we recording at the moment
     private val isRecording = AtomicBoolean(false)
@@ -199,14 +203,14 @@ internal class SessionReplayFeature(
     // region Manual Recording
 
     internal fun manuallyStopRecording() {
-        if (shouldRecord.compareAndSet(true, false)) {
-            recordingStateChanged.set(true)
+        if (userIntentToRecord.compareAndSet(true, false)) {
+            userIntentToRecordChanged.set(true)
         }
     }
 
     internal fun manuallyStartRecording() {
-        if (shouldRecord.compareAndSet(false, true)) {
-            recordingStateChanged.set(true)
+        if (userIntentToRecord.compareAndSet(false, true)) {
+            userIntentToRecordChanged.set(true)
         }
     }
 
@@ -259,7 +263,7 @@ internal class SessionReplayFeature(
     }
 
     private fun shouldHandleSession(alreadySeenSession: Boolean): Boolean {
-        return !alreadySeenSession || recordingStateChanged.get()
+        return !alreadySeenSession || userIntentToRecordChanged.get()
     }
 
     private fun applySampling(alreadySeenSession: Boolean) {
@@ -269,9 +273,14 @@ internal class SessionReplayFeature(
     }
 
     private fun modifyShouldRecordState(sessionData: SessionData) {
-        val isSampledIn = sessionData.keepSession && isSessionSampledIn.get()
-        if (!isSampledIn) {
-            if (shouldRecord.compareAndSet(true, false)) {
+        val isSessionEligible = sessionData.keepSession && isSessionSampledIn.get()
+        if (isSessionEligible) {
+            shouldRecord.set(userIntentToRecord.get())
+        } else {
+            shouldRecord.set(false)
+            if (!sessionData.keepSession) {
+                logNotKeptMessage()
+            } else {
                 logSampledOutMessage()
             }
         }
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -99,11 +99,15 @@ internal class SessionReplayFeature(
...
 
     // should we record the session - a combination of rum sampling, sr sampling
-    // and sr stop/start state
-    private val shouldRecord = AtomicBoolean(startRecordingImmediately)
+    // and user option.
+    private val shouldRecord = AtomicBoolean(false)
+
+    // Indicates the user options on recording, it starts with `startRecordingImmediately`
+    // in configuration, can be changed by calling start/stop recordings API.
+    private val userOption = AtomicBoolean(startRecordingImmediately)
```
jonathanmos, line 107: Perhaps something more descriptive? `isManualRecordingMode`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -99,11 +99,15 @@ internal class SessionReplayFeature(
...
 
     // used to monitor changes to an active session due to manual stop/start
-    private val recordingStateChanged = AtomicBoolean(false)
+    private val userOptionChanged = AtomicBoolean(false)
```
jonathanmos, line 110: How about `isManualRecordingStarted` or something like that?

jonathanmos (dismissed):

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeatureTest.kt
```diff
@@ -801,6 +801,121 @@ internal class SessionReplayFeatureTest {
...
+            verify(mockRecorder).resumeRecorders()
+            verify(mockRecorder).stopRecorders()
+        }
+        verifyNoMoreInteractions(mockRecorder)
+    }
+
+    @Test
+    fun `M resume recording in new session W user call startRecording {startRecordingImmediately = false}`() {
+// Given
+        val fakeSessionId3 = UUID.randomUUID().toString()
```
aleksandr-gringauz, line 854: minor. `Elmyr` can generate UUIDs ([link](https://github.com/xgouchet/Elmyr/blob/0ab6c3a673fd8771e9fc02f0676a3a886992a1ef/jvm/src/main/kotlin/fr/xgouchet/elmyr/jvm/factories/UUIDForgeryFactory.kt#L11)), it is probably better to generate these fake session using `forge`.

But I see, that in this file `UUID.randomUUID().toString()` is used.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
ambushwork: I thought about this kind of naming, the reason why I want to have a different name is because it contains not only the "manual recording" state, but also "startRecordingImmediately" state here. 
For example if `startRecordingImmediately` is true, the flag will be true at the beginning, but actually user never call `startRecording`, in this case calling `isManualRecordingMode` is not appropriate.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -269,9 +273,14 @@ internal class SessionReplayFeature(
...
 
     private fun modifyShouldRecordState(sessionData: SessionData) {
-        val isSampledIn = sessionData.keepSession && isSessionSampledIn.get()
-        if (!isSampledIn) {
-            if (shouldRecord.compareAndSet(true, false)) {
+        val isSessionEligible = sessionData.keepSession && isSessionSampledIn.get()
+        if (isSessionEligible) {
+            shouldRecord.set(userOption.get())
+        } else {
+            shouldRecord.compareAndSet(true, false)
```
aleksandr-gringauz, line 280: 1. In the previous version of the code the logging was done only when `compareAndSet` succeeded. Now it is done in any case. Is it ok? If `modifyShouldRecordState ` is called from 2 threads concurrently now (with the same `sessionData.keepSession` and `isSessionSampledIn`), the logging could happen twice. If we do `if (shouldRecord.compareAndSet(true, false))`, it will happen once.
2. `compareAndSet` seems unnecessary the way code is written now, `set` is enough.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeatureTest.kt
ambushwork: updated

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
ambushwork: Good point, so here I don't want to log only when `compareAndSet` is successful, for example sessionA is sampled out, we log the message, `shouldRecord` is false, after a while sessionB is created, it is also sampled out, for me we should log it again because it's a new session, but in previous version `compareAndSet` will fail and will skip the log.

and you are right since we don't use the return value, I change it from `compareAndSet` to `set`

jonathanmos (approved):

aleksandr-gringauz (approved):

PR #2631, opened 2025-05-06
RUM-9854: Introduce event processing thread

0xnm:
### What does this PR do?

This PR adds a dedicated context event processing thread which will be used when `FeatureScope.withWriteContext` is called.

Before this change when `FeatureScope.withWriteContext` was called, the content of the block was executed on the thread from I/O pool and block arguments were `DatadogContext` and `EventBatchWriter`.

After this change the block arguments are `DatadogContext` and `EventWriteScope`, where `EventWriteScope` is block-accepting lambda where argument is `EventWriteScope`. This allows now have 2 pipelines inside `withWriteContext` call:

* the content of the block passed to the `withWriteContext` will be executed on the single context processing thread shared between all the features.
* the context of the block passed to the `EventWriteScope` will be executed on the thread from I/O pool.

To illustrate it better:

```
// before
withWriteContext { datadogContext, eventBatchWriter ->
  // all on the I/O thread
  // ...
  // there is no switch to the I/O thread inside the write call, because we are already on the I/O thread
  eventBatchWriter.write(...)
}

// after
withWriteContext { datadogContext, writeScope ->
  // on the context processing thread
  // ...
  writeScope { eventBatchWriter ->
    // on the I/O thread
    eventBatchWriter.write(...)
  }
}
```

This will allow to introduce the wait mechanism for the feature context synchronization without imposing such wait on the client thread.

Also this PR moves `withWriteContext` call up to the root scope in the RUM processing, so that we can have context earlier.

The diff looks big, but its majority is in the tests where the main volume of changes comes from formatting and renaming.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
@@ -59,6 +59,7 @@ import com.datadog.android.core.internal.system.NoOpAppVersionProvider
 import com.datadog.android.core.internal.system.NoOpSystemInfoProvider
 import com.datadog.android.core.internal.system.SystemInfoProvider
 import com.datadog.android.core.internal.thread.BackPressureExecutorService
+import com.datadog.android.core.internal.thread.DatadogThreadFactory
 import com.datadog.android.core.internal.thread.LoggingScheduledThreadPoolExecutor
 import com.datadog.android.core.internal.thread.ScheduledExecutorServiceFactory
 import com.datadog.android.core.internal.time.AppStartTimeProvider
@@ -97,8 +98,10 @@ import java.lang.ref.WeakReference
 import java.util.Locale
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ExecutorService
+import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.ScheduledThreadPoolExecutor
+import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 
@@ -144,6 +147,7 @@ internal class CoreFeature(
 
     internal lateinit var uploadExecutorService: ScheduledThreadPoolExecutor
     internal lateinit var persistenceExecutorService: FlushableExecutorService
+    internal lateinit var contextExecutorService: ThreadPoolExecutor
     internal lateinit var backpressureStrategy: BackPressureStrategy
 
     internal var localDataEncryption: Encryption? = null
@@ -286,6 +290,7 @@ internal class CoreFeature(
     fun drainAndShutdownExecutors() {
         val tasks = arrayListOf<Runnable>()
 
+        contextExecutorService.queue.drainTo(tasks)
         persistenceExecutorService.drainTo(tasks)
 
         uploadExecutorService
@@ -295,9 +300,11 @@ internal class CoreFeature(
         // we need to make sure we drain the runnable list in both executors first
         // then we shut them down by using the await termination method to make sure we block
         // the thread until the active task is finished.
+        contextExecutorService.shutdown()
         persistenceExecutorService.shutdown()
         uploadExecutorService.shutdown()
 
+        contextExecutorService.awaitTermination(DRAIN_WAIT_SECONDS, TimeUnit.SECONDS)
         persistenceExecutorService.awaitTermination(DRAIN_WAIT_SECONDS, TimeUnit.SECONDS)
         uploadExecutorService.awaitTermination(DRAIN_WAIT_SECONDS, TimeUnit.SECONDS)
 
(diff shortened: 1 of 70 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
```diff
@@ -169,16 +174,43 @@ internal class SdkFeature(
...
-        val contextProvider = coreFeature.contextProvider
-        if (contextProvider is NoOpContextProvider) return
-        val context = contextProvider.context
-        storage.writeCurrentBatch(context) { callback(context, it) }
+        coreFeature.contextExecutorService
+            .executeSafe("withWriteContext-${wrappedFeature.name}", internalLogger) {
+                val contextProvider = coreFeature.contextProvider
+                if (contextProvider is NoOpContextProvider) return@executeSafe
+                val context = contextProvider.context
+                val eventBatchWriteScope = storage.getEventWriteScope(context)
```
aleksandr-gringauz, line 184: [minor] Looks like a duplicate code similar to the one in `getWriteContextSync`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -613,6 +620,19 @@ internal class CoreFeature(
...
         )
+        // TODO RUM-9851 Switch to the executor which is aware of backpressure, but only logs it
+        @Suppress("UnsafeThirdPartyFunctionCall") // all parameters are safe
+        contextExecutorService = ThreadPoolExecutor(
+            // core pool size
+            1,
+            // max pool size,
+            1,
+            // keep-alive time
+            0L,
```
mariusc83, line 631: why not keeping it alive for a while ?

features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/LogsFeature.kt
```diff
@@ -172,29 +172,36 @@ internal class LogsFeature(
...
-                    bundleWithTraces = true,
-                    bundleWithRum = true,
-                    networkInfo = null,
-                    userInfo = null,
-                    threadName = jvmCrash.threadName,
-                    threads = jvmCrash.threads,
-                    tags = emptySet()
-                )
+        // TODO RUM-9852 Implement better passthrough mechanism for the JVM crash scenario
+        val writeContext = sdkCore.getFeature(name)
```
aleksandr-gringauz, line 176: Is my understanding correct that we need this `getWriteContextSync` only because we have to log the crash synchronously, because otherwise the process might die before we finish (because we do it in `Thread.UncaughtExceptionHandler`)?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt
0xnm: for now, it is, yes. There is a TODO item attached to the `getWriteContextSync`, so maybe this method will be gone by the end of the development.

mariusc83 (approved):
LGTM

features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/LogsFeature.kt
0xnm: yes, this is right. That is why we have a TODO item as well, we want to have a quick passthrough for crashes avoiding switching threads. The only thread switch we can allow is for the I/O write.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
0xnm: This is to align with defaults of the `Executors.newSingleThreadExecutor` call ([link](https://cs.android.com/android/platform/superproject/main/+/main:libcore/ojluni/src/main/java/java/util/concurrent/Executors.java;l=197-202))

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/ThreadSafety.kt
```diff
@@ -189,5 +192,28 @@ class ThreadSafety(
         }
     }
 
+    private fun resolveCallFullNames(resolvedCall: ResolvedCall<out CallableDescriptor>): Set<String> {
```
aleksandr-gringauz, line 195: Could you please explain what this change is for? I'm just not familiar with writing custom detekt rules and don't understand what changed.

aleksandr-gringauz: ah, I see the new tests, it is clearer now

0xnm: yes, it is basically for `typealias` thing, because `ResolvedCall` will have a direct type signature, with `typealias` erased, so we need to generate the signature with `typealias` included in addition.

aleksandr-gringauz (approved):

xgouchet (approved):

PR #2635, opened 2025-05-07
RUM-9881: Push context changes from public API to the context thread

0xnm:
### What does this PR do?

To align the sequence of context updates/reads caused by the SDK API calls, we need to update context on the context thread only, so that something like this is correct:

```
Datadog.setUserInfo(A)

// RUM View should get User Info = A and not User Info = B (which may happen in case of concurrency).
RumMonitor.startView(...)

Datadog.setUserInfo(B)
```

### DIFF
--- a/reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/SdkCoreTest.kt
+++ b/reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/SdkCoreTest.kt
@@ -58,8 +58,8 @@ class SdkCoreTest : MockServerTest() {
 
     private lateinit var fakeTrackingConsent: TrackingConsent
 
-    private var featureSdkCore: FeatureSdkCore? = null
-    private var testedSdkCore: SdkCore? = null
+    private lateinit var featureSdkCore: FeatureSdkCore
+    private lateinit var testedSdkCore: SdkCore
 
     @Before
     fun setUp() {
@@ -71,13 +71,15 @@ class SdkCoreTest : MockServerTest() {
         fakeTrackingConsent = forge.aValueFrom(TrackingConsent::class.java)
         fakeUserAdditionalProperties = forge.exhaustiveAttributes(excludedKeys = setOf("id", "name", "email"))
         val configuration: Configuration = forge.getForgery()
-        testedSdkCore = Datadog.initialize(
-            ApplicationProvider.getApplicationContext(),
-            configuration,
-            fakeTrackingConsent
+        testedSdkCore = checkNotNull(
+            Datadog.initialize(
+                ApplicationProvider.getApplicationContext(),
+                configuration,
+                fakeTrackingConsent
+            )
         )
-        featureSdkCore = testedSdkCore as? FeatureSdkCore
-        featureSdkCore?.registerFeature(stubFeature)
+        featureSdkCore = testedSdkCore as FeatureSdkCore
+        featureSdkCore.registerFeature(stubFeature)
     }
 
     @After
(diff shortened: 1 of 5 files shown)


ambushwork:
is it possible to add unit test to protect against the concurrency issue that you mentioned in the description?

ambushwork (approved):

jonathanmos (approved):

xgouchet (approved):

mariusc83 (approved):

0xnm:
@ambushwork Good point, totally makes sense. I've added the necessary integration test.

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/SdkCoreTest.kt
```diff
@@ -375,18 +377,38 @@ class SdkCoreTest : MockServerTest() {
...
+        featureSdkCore.getFeature(stubFeature.name)?.withWriteContext { datadogContext, _ ->
             trackingConsent = datadogContext.trackingConsent
             countDownLatch.countDown()
         }
         countDownLatch.await(SHORT_WAIT_MS, TimeUnit.MILLISECONDS)
         assertThat(trackingConsent).isEqualTo(expectedTrackingConsent)
     }
 
+    @Test
+    fun must_not_have_concurrency_when_setTrackingConsent() {
```
ambushwork, line 394: This test have a very high chance of false positive since it is not concurrent enough, here is my suggestion:

```
    @Test
    fun must_not_have_concurrency_when_setTrackingConsent() {
        // Given
        val repeatTimes = forge.anInt(10, 100)
        val countDownLatch = CountDownLatch(repeatTimes)
        val expectedList = mutableListOf<TrackingConsent>()
        for (i in 0 until repeatTimes) {
            expectedList.add(forge.aValueFrom(TrackingConsent::class.java))
        }
        val resultList = mutableListOf<TrackingConsent>()
        // When
        for (i in 0 until repeatTimes) {
            testedSdkCore.setTrackingConsent(expectedList[i])
            featureSdkCore.getFeature(stubFeature.name)?.withWriteContext { datadogContext, _ ->
                resultList.add(datadogContext.trackingConsent)
                countDownLatch.countDown()
            }
            testedSdkCore.setTrackingConsent(forge.aValueFrom(TrackingConsent::class.java))
        }
        
        // Then
        countDownLatch.await(SHORT_WAIT_MS, TimeUnit.MILLISECONDS)
        assertThat(resultList).isEqualTo(expectedList)
    }
```

0xnm: Given the frequency of the execution of the tests CI job it is enough to have only one run I think to have a true positive _at some point_, but I pushed a suggested change for the better local reproducibility.

ambushwork: Here the problem that I see is that not matter how many times you execute the test that you had before, it will always pass even without your fix. (you can approve it by adding `@RepeatedTest(100)` in local ). So for me the frequency of the execution on CI can't actually reproduce the concurrency the issue.

The main point is that we need to have intensive execution in the single test run.

0xnm: > it will always pass

It depends on the env, I had this test failing (without `repeat`) on my machine without the change in production files.

Unfortunately, we cannot use `RepeatedTest` here, because it is instrumented test, so JUnit 4, that is why relying on the `repeat` in the test is better for the local reproducibility.

ambushwork (approved):

PR #2645, opened 2025-05-13
RUM-9940: Make `getDatadogContext` read on the context thread

0xnm:
### What does this PR do?

This PR makes `DatadogCore.getDatadogContext` call read `DatadogContext` on the context thread in the blocking manner to align with context changes.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
@@ -35,6 +35,7 @@ import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
 import com.datadog.android.core.internal.time.DefaultAppStartTimeProvider
 import com.datadog.android.core.internal.utils.executeSafe
+import com.datadog.android.core.internal.utils.getSafe
 import com.datadog.android.core.internal.utils.scheduleSafe
 import com.datadog.android.core.internal.utils.submitSafe
 import com.datadog.android.core.thread.FlushableExecutorService
@@ -47,9 +48,7 @@ import java.io.File
 import java.util.Locale
 import java.util.UUID
 import java.util.concurrent.Callable
-import java.util.concurrent.CancellationException
 import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.ExecutionException
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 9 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -371,7 +358,9 @@ internal class DatadogCore(
     }
 
     override fun getDatadogContext(): DatadogContext? {
-        return contextProvider?.context
+        return coreFeature.contextExecutorService
+            .submitSafe("getDatadogContext", internalLogger, Callable { contextProvider?.context })
```
mariusc83, line 362: is this going to block if the queue is not ready ?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProvider.kt
```diff
@@ -13,6 +13,7 @@ internal class DatadogUserInfoProvider(
     internal val dataWriter: DataWriter<UserInfo>
 ) : MutableUserInfoProvider {
 
+    @Volatile
```
mariusc83, line 16: shouldn't this be atomic instead ? Volatile only guarantees the visibility between 2 threads.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
0xnm: Yes, it will. This call is blocking on purpose.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProvider.kt
0xnm: There is no difference between `AtomicReference.get` / `AtomicReference.set` usage and `volatile` modifier usage ([link](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/util/concurrent/atomic/AtomicReference.java#L60-L93)). There is a difference in case of CAS usage, but it is not needed here.

aleksandr-gringauz: Should things like `CoreFeature.clientToken` [link](https://github.com/DataDog/dd-sdk-android/blob/8be1d29860375c75497f72af6f8d7a2c2dbeb8b6/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogContextProvider.kt#L22) become `volatile` as well?

0xnm: This is the thing to address in the future PRs, but so far it is good without it, because it is used only in the upload cycle which is scheduled only few seconds after SDK is initialized (-> when `CoreFeature` gets `clientToken` set).

And it is reset to empty string only when SDK is stopped anyway.

Overall I will align the properties used in the `DatadogContext` creation later.

mariusc83 (approved):

aleksandr-gringauz (approved):

PR #2647, opened 2025-05-14
RUM-9958: Align datadog initialization with ios for benchmark app

aleksandr-gringauz:
### What does this PR do?

As we discussed in one of our meetings, we should align Datadog SDK initialization for baseline runs of our benchmark scenarios between iOS and Android. The consensus was to do it the same way as on iOS - for baseline runs no Datadog components should be initialized. For example for session replay [link](https://github.com/DataDog/dd-sdk-ios/blob/develop/BenchmarkTests/Runner/Scenarios/SessionReplay/SessionReplaySwiftUI/SessionReplaySwiftUIScenario.swift#L22).

What changed:
1. Datadog SDK Core isn't initialized in baseline run. 
2. Rum isn't initialized in baseline run for session replay.

### Motivation

### Additional Notes


sample/benchmark/src/main/java/com/datadog/benchmark/sample/DatadogFeaturesInitializer.kt
```diff
@@ -85,13 +91,13 @@ internal class DatadogFeaturesInitializer @Inject constructor(
...
         val logsConfig = LogsConfiguration.Builder().build()
-        Logs.enable(logsConfig, sdkCore)
+        Logs.enable(logsConfig, sdkCore.get())
     }
 
     private fun needToEnableRum(config: BenchmarkConfig): Boolean {
         return when (isInstrumentedRun(config)) {
             true -> isSessionReplayScenario(config) || isRumScenario(config)
-            false -> isSessionReplayScenario(config)
+            false -> false
```
0xnm, line 100: just curious: is there any benefit of `when(boolean)` vs `if-else`? is it the same bytecode after all?

sample/benchmark/src/main/java/com/datadog/benchmark/sample/di/activity/DatadogActivityModule.kt
```diff
@@ -59,5 +94,29 @@ internal interface DatadogActivityModule {
...
+        env = BuildConfig.BUILD_TYPE
+    )
+        .setBatchSize(BatchSize.SMALL)
+        .setUploadFrequency(UploadFrequency.FREQUENT)
+
+    configBuilder.setBackpressureStrategy(
+        BackPressureStrategy(
+            CAPACITY_BACK_PRESSURE_STRATEGY,
+            { Log.w("BackPressure", "THRESHOLD REACHED!") },
+            { Log.e("BackPressure", "ITEM DROPPED $it!") },
```
0xnm, line 110: let's avoid screaming in the logs :)

```suggestion
            { Log.w("BackPressure", "Threshold reached") },
            { Log.e("BackPressure", "Item dropped: $it") },
```

0xnm (approved):

sample/benchmark/src/main/java/com/datadog/benchmark/sample/DatadogFeaturesInitializer.kt
aleksandr-gringauz: https://gist.github.com/aleksandr-gringauz/8850479ea36934f937be162d79ef7fc0

this is what I got when decompiling kotlin code for desktop jvm

looks like when adds some overhead related to throwing `NoWhenBranchMatchedException`. Why is it doing this - I currently don't know, it is an exhaustive when, compiler should be able to figure it out.

Changed to `if` in my case, probably `if` is more idiomatic anyway.

PR #2650, opened 2025-05-15
RUM-9977: Update RUM feature context only after event processing completion

0xnm:
### What does this PR do?

This PR brings the following change: now we will update RUM feature context only once, once RUM events processing is complete.

This change requires breaking a bit encapsulation by exposing concrete sub-types of `RumScope`, but advantage is that we remove multiple feature context update call sites and leave only one place where it is done. Anyway, the hierarchy of the scopes is always the same, we never changed it.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -700,6 +698,7 @@ internal class DatadogRumMonitor(
                     executorService.executeSafe("Rum event handling", sdkCore.internalLogger) {
                         synchronized(rootScope) {
                             rootScope.handleEvent(event, datadogContext, writeScope, writer)
+                            updateFeatureContext()
```
ambushwork, line 701: Question: If I understand correctly, prior to this PR, `RumViewSceope` only update RUM feature context for `onStartAction`, with this change, we try to update the context for every event, is that correct?

ambushwork: If we take `RumRawEvent.ActionSent` as an example, this event will be delegated to `RumViewScope.onActionSent`, when handling this event, I see it will only update the context for `has_replay` attribute.

However with this PR, it will  have an additional `updateFeatureContext` here, if it is true, we are updating feature context more frequently than before?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
```diff
@@ -67,9 +66,17 @@ internal class RumApplicationScope(
         )
     )
 
-    val activeSession: RumScope?
+    val activeSession: RumSessionScope?
         get() {
-            return childScopes.find { it.isActive() }
+            val activeSessions = childScopes.filter { it.isActive() }
+            if (activeSessions.size > 1) {
+                sdkCore.internalLogger.log(
```
mariusc83, line 73: is this related with this PR ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
mariusc83: @ambushwork I understand what you are saying but I don't think it is too much harm in that, I guess removing this complexity is worthy, what do you think ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt
0xnm: yes, it is. It is just a safeguard to inform that we have to make a selection from the multiple active sessions.

mariusc83 (approved):

PR #2655, opened 2025-05-16
RUM-9908 Create view attributes update methods

xgouchet:
### What does this PR do?

Add methods for customer to update the current View's attributes

### DIFF
--- a/features/dd-sdk-android-rum/api/apiSurface
+++ b/features/dd-sdk-android-rum/api/apiSurface
@@ -120,6 +120,8 @@ interface com.datadog.android.rum.RumMonitor
   fun clearAttributes()
   fun stopSession()
   fun addViewLoadingTime(Boolean)
+  fun addViewAttributes(Map<String, Any?>)
+  fun removeViewAttributes(Collection<String>)
   var debug: Boolean
   fun _getInternal(): _RumInternalProxy?
 enum com.datadog.android.rum.RumPerformanceMetric
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -216,6 +216,8 @@ internal open class RumViewScope(
 
             is RumRawEvent.UpdatePerformanceMetric -> onUpdatePerformanceMetric(event)
             is RumRawEvent.AddViewLoadingTime -> onAddViewLoadingTime(event, writer)
+            is RumRawEvent.AddViewAttributes -> onAddViewAttributes(event)
+            is RumRawEvent.RemoveViewAttributes -> onRemoveViewAttributes(event)
 
             else -> delegateEventToChildren(event, writer)
         }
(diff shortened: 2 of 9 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -182,10 +184,9 @@ internal class RumResourceScope(
         eventTime: Time,
         writer: DataWriter<Any>
     ) {
-        attributes.putAll(GlobalRumMonitor.get(sdkCore).getAttributes())
```
mariusc83, line 185: I think this PR contains the changes from the other one ? I think you need a rebase here ?

satween:
I think this PR misses attribute Precedence:

> Attributes are overwritten in events as they become more specific, so the precedence order is “Local, View, Global”.

For example see this [line](https://github.com/DataDog/dd-sdk-android/blob/dc61c078d0f55a9edddc77d61736b93a2164a077/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L900) if I am not mistaken it's seems like global attributes overrides event-level attributes.

There are probably some other similar places when global attributes overrides event-level ones

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -319,6 +321,18 @@ internal open class RumViewScope(
         sendViewUpdate(event, writer)
     }
 
+    @WorkerThread
+    private fun onAddViewAttributes(event: RumRawEvent.AddViewAttributes) {
+        viewAttributes.putAll(event.attributes)
```
ambushwork, line 326: is it thread-safe if `putAll` and `remove` are called in different threads?

xgouchet: In theory this can't be called on different threads

ambushwork (dismissed):

mariusc83 (dismissed):

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -120,6 +120,8 @@ interface com.datadog.android.rum.RumMonitor
   fun clearAttributes()
   fun stopSession()
   fun addViewLoadingTime(Boolean)
+  fun addViewAttributes(Map<String, Any?>)
```
simaoseica-dd, line 123: question/ Any reason for not having an interface for single attributes? 
Additionally, is it planned an interface to add multiple global attributes?

xgouchet: The RFC didn't mention adding a single attribute although it's feasible anyway with the current API. 
I can create another PR which adds the option to add multiple global attributes 👍

mariusc83 (approved):

ambushwork (approved):

PR #2662, opened 2025-05-20
RUM-10040: Perform lazy capture of `DatadogContext` at the span creation

0xnm:
### What does this PR do?

Right now when span is started and `bundleWithRum=true` in OpenTracing or OpenTelemetry tracers configuration, we are doing a call to get RUM context in the synchronous way, thus blocking a caller thread.

With the migration to the context processing thread, we cannot allow long blocking on the caller thread, so in this PR we do a lazy capture of the `DatadogContext` at the span creation time and add a captor as a span tag.

Later on it is safe to access value captured at the span write stage, because `withWriteContext` will also be executed on the context processing thread, meaning by that time the call we did before (`withContext`) to capture initial context will be completed.

This PR aims to glue the blocking/sync API of tracing and async API of the event/context processing.

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/concurrent/CompletableFuture.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/concurrent/CompletableFuture.kt
@@ -1,0 +1,25 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.concurrent
+
+/**
+ * java.util.concurrent.CompletableFuture is available only starting from API 24, so here
+ * is a simple class mimicking very basics of it.
+ */
+@Suppress("UndocumentedPublicFunction", "UndocumentedPublicProperty")
+class CompletableFuture<T : Any> {
+    @Volatile
+    lateinit var value: T
+        private set
+
+    fun isComplete(): Boolean = this::value.isInitialized
+
+    fun complete(value: T) {
+        if (isComplete()) return
+        this.value = value
+    }
+}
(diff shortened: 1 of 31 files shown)


features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
```diff
@@ -70,6 +75,34 @@ internal class OtelTraceWriter(
...
+                    LogAttributes.RUM_APPLICATION_ID,
+                    rumContext["application_id"] as? String
+                )
+                setTag(LogAttributes.RUM_SESSION_ID, rumContext["session_id"] as? String)
+                setTag(LogAttributes.RUM_VIEW_ID, rumContext["view_id"] as? String)
+                setTag(LogAttributes.RUM_ACTION_ID, rumContext["action_id"] as? String)
+            } else {
+                internalLogger.log(
+                    InternalLogger.Level.ERROR,
+                    InternalLogger.Target.USER,
```
ambushwork, line 98: is it an actionable log for user? maybe we should log to telemetry to monitor the frequency of this error?

ambushwork (approved):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/data/OtelTraceWriter.kt
0xnm: Normally, this should never happen ™️, but here if it happens and user is missing linking between RUM and Traces, I would like to have a support ticket opened by the user instead, with necessary details.

jonathanmos (approved):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/concurrent/CompletableFuture.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.datadog.android.internal.concurrent
+
+/**
+ * java.util.concurrent.CompletableFuture is available only starting from API 24, so here
+ * is a simple class mimicking very basics of it.
+ */
+@Suppress("UndocumentedPublicFunction", "UndocumentedPublicProperty")
+class CompletableFuture<T : Any> {
```
aleksandr-gringauz, line 14: very minor

there is no need for `T` here to be non-nullable, `CompletableFuture` should have no problem returning a nullable result.

0xnm: it has to be non-nullable, because `lateinit` doesn't allow having a nullable type.

aleksandr-gringauz (approved):

PR #2666, opened 2025-05-21
RUM-10064: Read RUM context in Session Replay in non-blocking manner

0xnm:
### What does this PR do?

The main goal of this PR is to move away from the `getFeatureContext()` calls in Session Replay to read RUM context, because it will become a blocking call and we cannot allow ourselves waiting in case if this call happens on the main thread.

This is achieved by listening to the RUM context updates instead and this is the same way iOS Session Replay it taking RUM context ([link](https://github.com/DataDog/dd-sdk-ios/blob/4e28d23b616dce388fde0e2fafd641fd16fb175b/DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift#L69)).

It can be a bit of the miss when associating Session Replay wireframes to the RUM views due to the following:

> Session Replay receives updates of the RUM context and this is the only way of getting it there. This means that if we started a new view and captured a new snapshot belonging to this view, but by the time we need to associate this snapshot with a RUM context new RUM startView didn’t finish processing yet (since it is a dedicated context queue, not main), we will associate this wireframe with a wrong RUM context, from the previous RUM view.

However, it is not a problem, because a miss will be very small and is not visible in the Session Replay player when switching between RUM views on the timeline.

Moreover, even with the current approach (by using `getFeatureContext()`) such mismatch is possible due to the concurrent nature of the event processing.

In general, this cannot be solved, because Session Replay and RUM tracking strategies are using different callbacks to report the data, so it cannot be aligned perfectly.

Also this PR does a minor change to the context update subscription mechanism: if there is a context available for the feature, it will be immediately emitted to the listener upon subscription.


ambushwork (approved):
LGTM!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayFeature.kt
```diff
@@ -170,6 +173,7 @@ internal class SessionReplayFeature(
 
     override fun onStop() {
         stopRecording()
+        sdkCore.removeContextUpdateReceiver(Feature.RUM_FEATURE_NAME, rumContextProvider)
```
mariusc83, line 176: isn't there a mistake here ? I think you are not using the same key right to register and remove ?

PR #2668, opened 2025-05-22
RUM-10088: Propagate `has_replay` flag to `RumContext` object

0xnm:
### What does this PR do?

This PR correctly sets `has_replay` flag from the Session Replay context in the `RumContext` object which is returned by the `RumScope.getRumContext`. It wasn't set before.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -126,6 +126,7 @@ internal open class RumViewScope(
     internal var version: Long = 1
     internal val customTimings: MutableMap<String, Long> = mutableMapOf()
     internal val featureFlags: MutableMap<String, Any?> = mutableMapOf()
+    internal var hasReplay = false
 
     internal var stopped: Boolean = false
 
@@ -258,7 +259,7 @@ internal open class RumViewScope(
                 viewType = type,
                 viewTimestamp = eventTimestamp,
                 viewTimestampOffset = serverTimeOffsetInMs,
-                hasReplay = false
+                hasReplay = hasReplay
             )
     }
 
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -258,7 +259,7 @@ internal open class RumViewScope(
                 viewType = type,
                 viewTimestamp = eventTimestamp,
                 viewTimestampOffset = serverTimeOffsetInMs,
-                hasReplay = false
+                hasReplay = hasReplay
```
ambushwork, line 262: Just wondering why it doesn't cause any bug if it is not correct set as said in the description, maybe we never use `hasReplay` from the context we retrieve it from this function? Or maybe children scope have `has_replay` false when using this function but other events have `has_replay` true so the whole session will have `has_replay` true anyway?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -927,20 +928,18 @@ internal open class RumViewScope(
...
+        )
+
         sdkCore.newRumEventWriteOperation(datadogContext, writeScope, writer, eventType) {
-            val currentViewId = rumContext.viewId.orEmpty()
             val user = datadogContext.userInfo
-            val hasReplay = featuresContextResolver.resolveViewHasReplay(
-                datadogContext,
-                currentViewId
-            )
-            sdkCore.updateFeatureContext(Feature.RUM_FEATURE_NAME) { currentRumContext ->
```
ambushwork, line 937: is this call removed because we have other place to update the rum context?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
0xnm: yes, because before we only put manual maps in the `updateFeatureContext` call. This PR removes the call which was doing that. And `RumContext.hasReplay` exists, but wasn't explicitly queried. Although, it is queried now https://github.com/DataDog/dd-sdk-android/blob/773728520138e5fe8a71877ba778709c61b483a0/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt#L736-L741.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: OK I see, we put `has_replay` in View scope context, and wait for here to transfer to rum context.

ambushwork (approved):

PR #2670, opened 2025-05-23
RUM-10120: Better project-type dependency resolution for Detekt custom rules execution

0xnm:
### What does this PR do?

I noticed that during custom Detekt rules job execution there were some warnings regarding the types that cannot be resolved.

This PR removes (mostly) the need to specify project dependencies for Detekt run, although there is a pitfall: `api` configuration cannot be resolved, so we cannot know transitive project dependencies. Maybe this can be solved with detached configuration, but for now let's just specify most common modules `:dd-sdk-android-core` and `:dd-sdk-android-internal` by default for everything.

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/config/DetektCustomConfig.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/config/DetektCustomConfig.kt
@@ -8,6 +8,7 @@ package com.datadog.gradle.config
 
 import com.android.build.gradle.LibraryExtension
 import org.gradle.api.Project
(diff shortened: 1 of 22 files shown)


satween (approved):
Looks good to me, although it would be great to have some linked JIRA ticket with more details ( like warnings examples) in case of any troubles in a future. 

Also we probably need to add some information somewhere that when new module is being added developer should revice DetectCostomConfig.kt file and check if he need to adjust logic that adds dependencies. Such internal changes is not frequent but on the other hand it's tricky to find it

0xnm:
@satween I linked a Jira ticket. Not sure where info about new module setup should be added.

PR #2672, opened 2025-05-26
RUM-9514 remove deprecated userInfo methods

xgouchet:
### What does this PR do?

Remove the option to set a null userId

### DIFF
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
@@ -53,36 +53,51 @@ internal class DatadogUserInfoProviderTest {
 
     @Test
     fun `M return saved userInfo W setUserInfo() and getUserInfo()`(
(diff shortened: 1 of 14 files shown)


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/user/DatadogUserInfoProviderTest.kt
```diff
@@ -53,36 +53,51 @@ internal class DatadogUserInfoProviderTest {
...
         // Then
-        verify(mockWriter).write(userInfo)
+        verify(mockWriter).write(validUserInfo)
     }
 
     @Test
     fun `M keep existing properties W addUserProperties() is called`(
         @Forgery userInfo: UserInfo,
+        @StringForgery userId: String,
         @StringForgery forge: Forge
```
0xnm, line 91: minor: strange that there is a `StringForgery` annotation here, it shouldn't have any effect and can be removed.

0xnm (dismissed):

0xnm (approved):

0xnm (approved):

PR #2675, opened 2025-05-27
RUM-9501 Don't send fatal errors to Logs, only send them to RUM

xgouchet:
### What does this PR do?

let the crash handlers (`DatadogExceptionHandler` and `DatadogNdkCrashHandler`) only send errors to RUM


satween (dismissed):
lgtm, but let's have another opinion here, @0xnm's maybe?

ambushwork (dismissed):

ambushwork:
> PR LGTM, but I am curious what's the motivation behind this one? (I don't see it neither in the description nor in the ticket)

I see it in V3 scope doc, OK for me

dd-sdk-android-core/src/main/kotlin/com/datadog/android/ndk/internal/DatadogNdkCrashHandler.kt
```diff
@@ -44,9 +43,6 @@ internal class DatadogNdkCrashHandler(
     internal var lastNetworkInfo: NetworkInfo? = null
```
0xnm, line 43: Probably we can remove `lastNetworkInfo` and `lastUserInfo` since they were used only for Logs? We can also remove the part where we write them to the disk.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/ndk/internal/DatadogNdkCrashHandler.kt
```diff
@@ -285,38 +199,6 @@ internal class DatadogNdkCrashHandler(
...
-        sdkCore: FeatureSdkCore,
-        errorLogMessage: String,
-        logAttributes: Map<String, String>,
-        ndkCrashLog: NdkCrashLog,
-        lastNetworkInfo: NetworkInfo?,
-        lastUserInfo: UserInfo?
-    ) {
-        val logsFeature = sdkCore.getFeature(Feature.LOGS_FEATURE_NAME)
-        if (logsFeature != null) {
-            logsFeature.sendEvent(
```
0xnm, line 300: we also need to remove related code in the `LogsFeature`

0xnm (changes requested):

0xnm (dismissed):

0xnm (approved):

PR #2694, opened 2025-06-03
RUM-10240: Add public APIs for set account information

ambushwork:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
@@ -314,6 +314,75 @@ object Datadog {
         sdkCore.clearAllData()
     }
 
+    /**
+     * Sets current account information.
+     *
+     * Those will be added to logs, traces and RUM events automatically.
+     *
+     * @param id Account ID.
+     * @param name representing the account, if exists.
+     * @param extraInfo Account's custom attributes, if exists.
+     * @param sdkCore SDK instance to set account information. If not provided, default SDK
+     * instance will be used.
+     */
+    @JvmStatic
+    @JvmOverloads
+    fun setAccountInfo(
+        id: String,
+        name: String? = null,
+        extraInfo: Map<String, Any?> = emptyMap(),
+        sdkCore: SdkCore = getInstance()
+    ) {
+        sdkCore.setAccountInfo(
+            id = id,
+            name = name,
+            extraInfo = extraInfo
+        )
+    }
+
+    /** Add custom attributes to the current account information
+     *
+     * This extra info will be added to already existing extra info that is added
+     * to Logs, Traces and RUM events automatically.
+     *
+     * @param extraInfo Account's additional custom attributes.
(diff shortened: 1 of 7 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
0xnm, line 1: I still find the difference between `userInfo` and `accountInfo` (same for `userExtraInfo` vs `accountExtraInfo`) very confusing.

I think we need to add a clear explanation to all `user`/ `account` APIs describing the difference (and use-cases) between the two.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
```diff
@@ -314,6 +314,75 @@ object Datadog {
...
+            id = id,
+            name = name,
+            extraInfo = extraInfo
+        )
+    }
+
+    /** Add custom attributes to the current account information
+     *
+     * This extra info will be added to already existing extra info that is added
+     * to logs traces and RUM events automatically.
```
aleksandr-gringauz: minor

`logs, traces` (comma is missing)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
ambushwork: `user` and `account` are just two different fields that will be provided in all the events, so that client can query with them without using custom attributes. About the what a `user` or  an `account` it is, it depends on how client wants to use it. Most common case in my mind is that `user` represents an application end user, this user can have multiple accounts for this app, `setAccountInfo` or  `clearAccountInfo` should be called when login/logout.

0xnm: My point here is that we need to update doc of each related public API method, describing the difference and usage. Otherwise it only creates a confusion when we have two options for almost the same.

ambushwork: Yes I understand, my point is that the definition of `account` and `user` heavily depends on how the client see them, in documentation we should probably not assume how they should use it, like saying "userInfo corresponds to end user, and accountInfo corresponds to user's account" etc.
But sure it's important to make it clear, I will discuss with IOS team how we should address it properly, (currently they have the similar doc on their side), and a [ticket](https://datadoghq.atlassian.net/browse/RUM-10247?atlOrigin=eyJpIjoiODExOTI5Y2RjNjRhNDM3NzgxOWIyYWQzNjY5NTQ3ZGEiLCJwIjoiaiJ9) is created for this.

aleksandr-gringauz (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
0xnm: Also we may want to update the doc of `setAnonymousId` with that development.

PR #2695, opened 2025-06-03
RUM-3100 generate checksum xml

xgouchet:
### What does this PR do?

Gradle allows developers to add a `verification-metadata.xml` file to their project to pin dependencies with a specific hash or PGP signature. This prevents attack if the artifact repository is compromised. 

> [!NOTE]
> When publishing an artifact to Maven, the hash is automatically computed, and downloaded by Gradle with the artifact itself. If the `verification-metadata.xml` is present, it will compare the downloaded hash with the one in the xml file and abort the build if they don't match.

In order to improve our customer's trust in our artifact, this PR will generate a `verification-metadata.xml` file for each published artifact (as part of the `publish:<module>` job), and a new job will create a merged `verification-metadata.xml` that we can publish on GitHub and customers can refer to to ensure at any given time that the hash present in Maven is the expected one.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -518,6 +518,11 @@ publish:release-core:
   script:
     - !reference [.snippets, set-publishing-credentials]
     - ./gradlew :dd-sdk-android-core:publishToSonatype --stacktrace --no-daemon
+  artifacts:
+    when: on_success
+    expire_in: 7 days
+    paths:
+      - dd-sdk-android-core/verification-metadata.xml
 
 publish:release-internal:
   tags: [ "arch:amd64" ]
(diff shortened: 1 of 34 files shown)


merge_verification_metadata.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2019-Present Datadog, Inc
+
+import sys
```
mariusc83, line 8: Great job !!
I am wondering where is this merger going to be used ? I am not seeing it yet in the code, is that correct ?

.gitlab-ci.yml
```diff
@@ -577,9 +604,14 @@ publish:release-glide:
   timeout: 30m
   script:
     - !reference [.snippets, set-publishing-credentials]
-    - ./gradlew :integrations:dd-sdk-android-glide:publishToSonatype --stacktrace --no-daemon
+    - ./gradlew :features:dd-sdk-android-rum:publishToSonatype --stacktrace --no-daemon
+  artifacts:
```
mariusc83, line 608: what's this config doing ?

.gitlab-ci.yml
```diff
@@ -875,3 +1004,45 @@ notify:dogfood-gradle-plugin:
...
+    - publish:release-trace-coroutines
+    - publish:release-rum-coroutines
+    - publish:release-rx
+    - publish:release-sqldelight
+    - publish:release-timber
+    - publish:release-android-tv
+    - publish:release-okhttp
+    - publish:release-okhttp-otel
+    - publish:release-benchmark
+  script:
```
mariusc83, line 1040: ok I see the script is being used here, I wonder what's the purpose of this merged file ?

mariusc83 (commented):
I am thinking, should we add also some docs on how this should be verified on their end ? how should they use this `xml` ?

.gitlab-ci.yml
```diff
@@ -794,6 +917,12 @@ publish:release-benchmark:
   script:
     - !reference [.snippets, set-publishing-credentials]
     - ./gradlew :tools:benchmark:publishToSonatype --stacktrace --no-daemon
+  artifacts:
+    when: on_success
+    expire_in: 7 days
+    paths:
+      - dd-sdk-android-internal/verification-metadata.xml
```
0xnm: probably it should be `tools/dd-sdk-android-benchmark-internal/verification-metadata.xml` here?

anyway, we can skip metadata publication here, because this one is only consumed by our internal apps and we don't want to enable verification there, we don't need it.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/verification/GenerateVerificationXmlTask.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        val aarSha256 = aarFile.sha256()
+        val pomSha256 = pomFile.sha256()
+        val moduleSha256 = moduleFile.sha256()
+
+        val content = xml(TAG_ROOT) {
+            xmlns = NS_DEPS_VERIF
+            TAG_CONFIGURATION {
+                TAG_VERIF_METADATA { -true.toString() }
+                TAG_VERIF_SIGNATURE { -false.toString() } // TODO-RUM-3104 add signature verification
```
0xnm:
```suggestion
                TAG_VERIF_SIGNATURE { -false.toString() } // TODO RUM-3104 add signature verification
```

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/verification/GenerateVerificationXmlTask.kt
```diff
@@ -0,0 +1,121 @@
...
+
+import com.datadog.gradle.config.AndroidConfig
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.TaskAction
+import org.redundent.kotlin.xml.PrintOptions
+import org.redundent.kotlin.xml.xml
+import java.io.File
+import java.security.MessageDigest
+
+open class GenerateVerificationXmlTask : DefaultTask() {
```
0xnm, line 17: so it seems there is no Gradle built-in way to generate it? It seems that one can use https://github.com/sigstore/sigstore-java, but indeed it may be simpler to generate it manually.

0xnm, line 17: should we have a test/validation that generated XML indeed matches the schema?

0xnm (dismissed):
lgtm! nice job.

I left some comments, but nothing blocking.

mariusc83 (dismissed):

.gitlab-ci.yml
```diff
@@ -875,3 +1004,45 @@ notify:dogfood-gradle-plugin:
     - pip3 install GitPython requests
     - aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.gh_token --with-decryption --query "Parameter.Value" --out text >> ./gh_token
     - python3 dogfood.py -v $CI_COMMIT_TAG -t gradle-plugin
+
+notify:merge-verification-metadata:
+  tags: [ "arch:amd64" ]
+  only:
+    - tags
+    - develop
```
ambushwork: Question: why do we need to do this on develop branch?

ambushwork (dismissed):

.gitlab-ci.yml
xgouchet: good call indeed, gradle's verification mecanism doesn't work with snapshot versions 👍

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/verification/GenerateVerificationXmlTask.kt
xgouchet: Gradle can only generate it for the depdencies your project is using but not for a library you're publishing (I couldn't find any doc for library maintainers at all), so for now I think it's better to do it manually. 
I did check that the `sha256` computed here matches the one when we publish to Maven (checked with `publishToMavenLocal`)

xgouchet: That's a good point, I'll try and see if I can write a test for that 👍

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/verification/GenerateVerificationXmlTask.kt
xgouchet: Good call

.gitlab-ci.yml
xgouchet: This will extract the generated XML for each module (as each one is published to Maven in its own job) so that it can be merged later in the new job.

.gitlab-ci.yml
xgouchet: Good catch

.gitlab-ci.yml
xgouchet: The purpose is to have a single XML file to provide our customers with all `sha256` there. 
I thought about publishing all the individual xml files, but that would be a hassle, compared to just copy-pasting the content of a single file in their own `verification-metadata.xml`. Even if they have hashes for unnecessary libs, it'll still work.

merge_verification_metadata.py
xgouchet: It's used in a new Gitlab job

xgouchet:
@mariusc83 
> should we add also some docs on how this should be verified on their end ? how should they use this xml ?

I'll add a doc in confluence on the process, the idea is to have the doc published in our release notes on github.

ambushwork (approved):

PR #2699, opened 2025-06-04
RUM-10099: Create a new RumViewScope when the session is renewed

ambushwork:
### What does this PR do?

Major Changes in this PR:
* `RumViewScope`
    * `sessionId`, `viewId` and `eventTime` become `val` in `RumViewScope`, they will not be able to renew in the same `RumViewScope` instance.
    * `oldViewIds` is removed, `RumViewScope` doesn't remember this anymore since `viewId` now is immutable.
    * Event callback doesn't check `oldViewIds` anymore before handling the event since `oldViewIds` is removed, meaning that we don't handle the pending event from the old view scope now.
* `RumViewManagerScope`
   * when the scope is asked to renew, it will clone all the children scope with the new event time and replace them.
* `RumSessionScope`
   * when session needs to renew, call proactively `RumViewManagerScope` to renew the children scope.
   * Replace the renew session timestamp with the event time, otherwise the session start time will be later than the first event in the session, which is not logical.

### Motivation

RUM-10099

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
@@ -14,6 +14,7 @@ import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.rum.RumSessionListener
 import com.datadog.android.rum.internal.domain.RumContext
+import com.datadog.android.rum.internal.domain.Time
 import com.datadog.android.rum.internal.metric.SessionMetricDispatcher
 import com.datadog.android.rum.internal.metric.slowframes.SlowFramesListener
 import com.datadog.android.rum.internal.utils.percent
@@ -60,7 +61,7 @@ internal class RumSessionScope(
     private val noOpWriter = NoOpDataWriter<Any>()
 
     @Suppress("LongParameterList")
-    internal var childScope: RumScope? = RumViewManagerScope(
+    internal var childScope: RumViewManagerScope? = RumViewManagerScope(
         this,
         sdkCore,
         sessionEndedMetricDispatcher,
@@ -121,7 +122,7 @@ internal class RumSessionScope(
         writer: DataWriter<Any>
     ): RumScope? {
         if (event is RumRawEvent.ResetSession) {
-            renewSession(System.nanoTime(), StartReason.EXPLICIT_STOP)
+            renewSession(event.eventTime, StartReason.EXPLICIT_STOP)
         } else if (event is RumRawEvent.StopSession) {
             stopSession()
         }
@@ -131,7 +132,7 @@ internal class RumSessionScope(
         val actualWriter = if (sessionState == State.TRACKED) writer else noOpWriter
 
         if (event !is RumRawEvent.SdkInit) {
-            childScope = childScope?.handleEvent(event, actualWriter)
+            childScope = childScope?.handleEvent(event, actualWriter) as? RumViewManagerScope
         }
 
         return if (isSessionComplete()) {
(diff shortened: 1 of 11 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -81,26 +81,15 @@ internal open class RumViewScope(
...
+    internal val viewId: String = UUID.randomUUID().toString()
 
-    private val oldViewIds = mutableSetOf<String>()
     private val startedNanos: Long = eventTime.nanoTime
     internal var stoppedNanos: Long = eventTime.nanoTime
     internal var viewLoadingTime: Long? = null
 
     private val serverTimeOffsetInMs = sdkCore.time.serverTimeOffsetMs
-    internal val eventTimestamp = eventTime.timestamp + serverTimeOffsetInMs
+    internal var eventTimestamp = eventTime.timestamp + serverTimeOffsetInMs
```
aleksandr-gringauz: is this change `val -> var` needed? I don't see where `eventTimestamp` is changed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -121,7 +122,7 @@ internal class RumSessionScope(
         writer: DataWriter<Any>
     ): RumScope? {
         if (event is RumRawEvent.ResetSession) {
-            renewSession(System.nanoTime(), StartReason.EXPLICIT_STOP)
+            renewSession(event.eventTime, StartReason.EXPLICIT_STOP)
```
xgouchet, line 125: Good catch !

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -178,6 +167,11 @@ internal open class RumViewScope(
         networkSettledMetricResolver.viewWasCreated(eventTime.nanoTime)
         interactionToNextViewMetricResolver.onViewCreated(viewId, eventTime.nanoTime)
         slowFramesListener?.onViewCreated(viewId, startedNanos)
+        if (rumContext.syntheticsTestId != null) {
+            Log.i(RumScope.SYNTHETICS_LOGCAT_TAG, "_dd.application.id=${rumContext.applicationId}")
+            Log.i(RumScope.SYNTHETICS_LOGCAT_TAG, "_dd.session.id=${rumContext.sessionId}")
+            Log.i(RumScope.SYNTHETICS_LOGCAT_TAG, "_dd.view.id=$viewId")
+        }
```
xgouchet: > [!WARNING]
> Is this for debugging purpose ? If we want to keep those in, let's use the InternalLogger instead.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -56,11 +56,18 @@ internal class RumViewManagerScope(
...
+    internal val childrenScopes = mutableListOf<RumViewScope>()
     internal var stopped = false
     private var lastStoppedViewTime: Time? = null
 
     // region RumScope
+    fun renewViewScopes(eventTime: Time) {
+        val newChildScope = childrenScopes.map { rumViewScope ->
+            rumViewScope.renew(eventTime)
+        }
+        childrenScopes.clear()
```
xgouchet, line 68: > [!NOTE]
> This means that the existing Views won't ever receive updates that might be relevant (e.g.: Action Sent, Stop Resource, …) leading to an incomplete view. It might be relevant to mark those session as stopped but still keep them to ensure they have the relevant action/resource/error/... count.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: nice catch, updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: I just move this part as-is, before this was in the place where the view id is updated, now since view id can't change, it is moved into the `init()`, not sure why it's `Log.i` before.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
ambushwork: This is an expected change in this PR since it has been discussed in the RFC.

But keeping the old scopes to handle the events while creating new ones is an interesting idea, the only concern is that if we copy the old view scopes, but we never know if there is an action of these views coming in the future, we will always keep it in the collections, causing collection size always growing.

mariusc83 (commented):
LGTM, waiting for cleaning up those `Log.i` and then I can accept.

ambushwork:
> LGTM, waiting for cleaning up those `Log.i` and then I can accept.

@mariusc83 updated

mariusc83 (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1395,6 +1400,14 @@ internal open class RumViewScope(
         }
     }
 
+    private fun logSynthetics(key: String, value: String) {
+        sdkCore.internalLogger.log(
+            level = InternalLogger.Level.INFO,
+            target = InternalLogger.Target.USER,
```
aleksandr-gringauz, line 1434: Are you sure the user needs these logs? Maybe `InternalLogger.Target.MAINTAINER` is better?

xgouchet (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: I think the goal here is to print the log on the synthetics which is the "user" of the app.

aleksandr-gringauz: ok

aleksandr-gringauz (dismissed):

aleksandr-gringauz (approved):

aleksandr-gringauz (approved):

PR #2702, opened 2025-06-05
RUM-10241: Add AccountInfo into CoreFeature and DatadogContext

ambushwork:
### What does this PR do?

Add AccountInfo into CoreFeature and DatadogContext

### Motivation

RUM-10241

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/DatadogContext.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/DatadogContext.kt
@@ -29,6 +29,7 @@ import com.datadog.android.privacy.TrackingConsent
  * @property networkInfo information about the current network availability and quality
  * @property deviceInfo information about device
  * @property userInfo information about the current user
+ * @property accountInfo information about the current account
  * @property trackingConsent information about the current tracking consent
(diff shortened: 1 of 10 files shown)


aleksandr-gringauz (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/DatadogContext.kt
```diff
@@ -48,6 +49,7 @@ data class DatadogContext(
     val networkInfo: NetworkInfo,
     val deviceInfo: DeviceInfo,
     val userInfo: UserInfo,
+    val accountInfo: AccountInfo?,
```
aleksandr-gringauz, line 52: very minor, no change needed

`AccountInfo` is nullable, but `UserInfo` is not. Though they are very similar.

xgouchet (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/DatadogContext.kt
ambushwork: `AccountInfo` should be optional and nullable, `UserInfo` use empty data class as a null user, I think it's because none of the `UserInfo` property is mandatory including `id`, so it has an dedicated function to check if it is a valid `UserInfo`

PR #2704, opened 2025-06-05
RUM-386: Process feature context on the context thread

0xnm:
### What does this PR do?

This PR changes the way feature context is read and updated. Now, by default, all calls to update feature context or to read it will delegate to the context processing thread, this will guarantee that sequence of the updates is corresponding to the sequence of the SDK API calls causing it.

There is also a way to update/read feature context from the caller thread, because sometimes calls are already done from the context thread (and sometimes we are ok doing such calls not from the context thread).

Since feature context is becoming a shared resource, it is guarded by the locks (write lock and read lock, read more about [ReentrantReadWriteLock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html)).

To avoid deadlock, write lock has a timeout during which it will try to acquire lock. If it is not possible to do during the timeout, update operation will be aborted.

This change now guarantees that, for example, there is the following API sequence:

```
--> RUM Context A
--> call to RUM updating RUM Context to Context B
--> logs/traces call
```

Then there won't be a situation when Logs/Traces read Context A instead of Context B just because they are processed faster than RUM.

This invariant is covered by the integration test.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -564,19 +603,76 @@ internal class DatadogCore(
...
+    private fun Lock.safeTryWithLock(time: Long, unit: TimeUnit, block: () -> Unit) {
+        val locked = try {
+            // NullPointerException cannot happen, time unit is not null
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            tryLock(time, unit)
+        } catch (e: InterruptedException) {
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                { "Couldn't acquire ${javaClass.simpleName} due to the exception thrown, aborting operation." },
```
ambushwork, line 615: is proguard rules needed here?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -564,19 +603,76 @@ internal class DatadogCore(
...
+                e
+            )
+            return
+        }
+        if (!locked) {
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                {
+                    "Couldn't acquire ${javaClass.simpleName} due to" +
```
ambushwork, line 625: same as above

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -260,10 +282,12 @@ internal class DatadogCore(
             )
         } else {
             feature.setContextUpdateListener(listener)
-            features.keys.forEach {
-                val currentContext = contextProvider?.getFeatureContext(it)
-                if (!currentContext.isNullOrEmpty()) {
-                    listener.onContextUpdate(it, currentContext)
+            features.forEach {
+                if (it.key != featureName) {
+                    val currentContext = it.value.featureContext
```
aleksandr-gringauz: shouldn't this read operation of `featureContext` be guarded by `featureContextLock`?

0xnm: indeed, good catch

ambushwork (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
0xnm: No, it is a platform class, so it is not obfuscated unlike app/library class.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogContextProvider.kt
```diff
@@ -64,19 +67,13 @@ internal class DatadogContextProvider(val coreFeature: CoreFeature) : ContextPro
                 // Values at the top 2 levels are frozen: feature-name key,
                 // and feature-specific-name key.
                 featuresContext = mutableMapOf<String, Map<String, Any?>>().apply {
-                    val source = coreFeature.featuresContext
-                    source.forEach { (key, value) ->
-                        this[key] = value.toMap()
+                    featureContextProvider.getFeaturesContexts().forEach {
+                        val value = it.second()
+                        if (value.isNotEmpty()) {
+                            this[it.first] = value.toMap()
```
aleksandr-gringauz, line 73: you already do the copying in `getFeatureContexts`, additional `toMap` seems unnecessary

I see the comment above, but I still don't understand...

0xnm: yes, this is just a leftover from the current code, but anyway this part will change in the next PR

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -398,19 +428,28 @@ internal class DatadogCore(
...
+            appContext,
             instanceId,
             mutableConfig,
             TrackingConsent.PENDING
         )
 
+        contextProvider = DatadogContextProvider(coreFeature) {
+            features.keys.map {
+                it to {
+                    // already on the context thread
```
aleksandr-gringauz, line 440: Just a suggestion. No change needed.

For these things it might be worth having some sort of `assertIsContextThread()` functions that crash if it is not the right thread. And make them crash only in our sample app.

We could put these in many places.

I see that we currently have [isDeveloperModeEnabled](https://github.com/DataDog/dd-sdk-android/blob/60a194593d21ce050e0ce9f46451b390a370b304/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt#L125). But this could be used by the customer [setUseDeveloperModeWhenDebuggable](https://github.com/DataDog/dd-sdk-android/blob/60a194593d21ce050e0ce9f46451b390a370b304/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt#L104).

 You might also want to crash in debug mode [here](https://github.com/DataDog/dd-sdk-android/pull/2704/files#diff-7502deb1ea8251815bb840fbd19753ec18219813f6fa83bf0c5bd7bfccfec1e6R621) instead of logging. But if we have some kind of monitors on these logs in our sample, it is fine.

aleksandr-gringauz: > instead of logging

in addition to logging

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -564,19 +603,76 @@ internal class DatadogCore(
...
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                { "Couldn't acquire ${javaClass.simpleName} due to the exception thrown, aborting operation." },
+                e
+            )
+            return
+        }
+        if (!locked) {
+            internalLogger.log(
```
aleksandr-gringauz, line 621: Just to confirm. Deadlock shouldn't be possible, you are just protecting from a possible error in code. Or am I missing something?

0xnm: No, `locked = false` will be in case of `tryLock` call didn't succeed, because of timeout. And timeout may happen not because of deadlock, but simply because lock was acquired by some other thread for a long time, so we need to report such cases, because it may be the indicator of the poor performance.

aleksandr-gringauz: > To avoid deadlock, write lock has a timeout during which it will try to acquire lock. If it is not possible to do during the timeout, update operation will be aborted.

ok, thanks for clarifying

Reading this, I thought you're just being extra cautious to not introduce a deadlock in case you made some kind of mistake.

aleksandr-gringauz (approved):

ambushwork (approved):

PR #2708, opened 2025-06-06
[RUM-9899]: TracingInterceptor migration

satween:
### What does this PR do?

This PR contains major change for tracer migration. 
I tried to reduce amount of changes with some tricks with  kotlin's `typealias` but it still huge. Maybe some of you will find it easier to review by commits. 

## Main changes: 
* Migration from `AndroidTracer` to `CoreTracer` in `TracingInterceptor` and it descendants. 
* Refactor and removing duplication logic in unit tests.

## Please pay attention to:

Despite the fact that the APIs of CoreTracer and AndroidTracer are similar, some methods were either not supported or supported at a different level of abstraction. In such cases, I implemented the closest and simplest solution and added comments so you could spot them.

## Known issues
* In  some tests StringForgery with regex inside generated 33 length string instead of 32 which was leading to expeption in DDTraceId parsing. Temporary replaced regex "[a-f0-9]{31}". Going to dig deeper in separate ticket. 
* There is no replacement for `GlobalTracer`, but `TracingInterceptor` relies on some globally stored `CoreTracer` instance. Going to discuss this with the team and provide solution with separate PR.


features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/interceptor/MutableSpan.java
```diff
@@ -78,4 +78,6 @@ default Object getTag(String key) {
    * @return The root span for the current trace fragment.
    */
   MutableSpan getLocalRootSpan();
+
+  void drop();
```
satween, line 82: > [!IMPORTANT]
> We need to drop spans for requests that failed. There was no drop method in `CoreTracer`.

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/PendingTrace.java
```diff
@@ -223,6 +223,14 @@ void registerSpan(final DDSpan span) {
         }
     }
 
+    public void unregisterSpan(final DDSpan span){
```
satween: > [!IMPORTANT]
> Another point that I want to pay attention to: Dropping span wasn't supported in `CoreTracer` and I am not sure that removing span from `ROOT_SPAN` is the right thing here. Please let me know if I missing something

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -78,7 +77,8 @@ open class DatadogInterceptor internal constructor(
     traceSampler: Sampler<Span>,
     traceContextInjection: TraceContextInjection,
     redacted404ResourceName: Boolean,
-    localTracerFactory: (SdkCore, Set<TracingHeaderType>) -> Tracer
+    localTracerFactory: (SdkCore, Set<TracingHeaderType>) -> Tracer,
+    globalTracerProvider: () -> Tracer?
```
satween: > [!IMPORTANT]
> There is no direct replacement for `GlobalTracer` for `CoreTracer`.
The only alternative we have is  [OtelTracerProvider](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt)  but there is no way to get `CoreTracer` instance from it. 
Another problem is that `OtelTracerProvider` is the part of `-otel` modules which will force customers to enable desugaring. So seems like we need some replacement for `GlobalTracer`. 
For the migration purposes I replaced `GlobalTracer` with `globalTracerProvider` lambda but at this moment no value provided for it. Let me know if you have other ideas for `GlobalTracer` replacements beside creating some other static container

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/otel/TraceContextExt.kt
```diff
@@ -7,34 +7,17 @@
...
-        spanIdAsBigInteger,
+        DDTraceId.fromHexOrDefault(traceId, DDTraceId.ZERO),
+        DDSpanId.fromHexOrDefault(spanId, DDSpanId.ZERO),
         samplingPriority,
         null,
-        emptyMap(),
-        emptyMap()
+        null,
+        null
     )
```
satween: > [!IMPORTANT]
> Pay attention to this code as well. I kept same logic with creating `ExtractedContext` from `TraceContext`

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -871,12 +847,18 @@ internal constructor(
         internal const val OKHTTP_INTERCEPTOR_HEADER_TYPES = "okhttp_interceptor_header_types"
 
         private const val AGENT_PSR_ATTRIBUTE = "_dd.agent_psr"
+
         private val DEFAULT_LOCAL_TRACER_FACTORY: (SdkCore, Set<TracingHeaderType>) -> Tracer =
-            { sdkCore, tracingHeaderTypes ->
```
satween, line 876: > [!IMPORTANT]
> There is no way to provide  `tracingHeaderTypes` directly into builder so I replaced it with properties, as they created in [OpenTracerProvider.Builder](https://github.com/DataDog/dd-sdk-android/blob/60a194593d21ce050e0ce9f46451b390a370b304/features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt#L266-L299)

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/DatadogInterceptorWithoutTracesTest.kt
```diff
@@ -133,26 +137,28 @@ internal class DatadogInterceptorWithoutTracesTest {
...
     var fakeRedacted404Resources: Boolean = true
     // endregion
 
     @BeforeEach
     fun `set up`(forge: Forge) {
-        whenever(mockLocalTracer.buildSpan(TracingInterceptor.SPAN_NAME)) doReturn mockSpanBuilder
-        whenever(mockSpanBuilder.withOrigin(DatadogInterceptor.ORIGIN_RUM)) doReturn mockSpanBuilder
-        whenever(mockSpanBuilder.asChildOf(null as SpanContext?)) doReturn mockSpanBuilder
-        whenever(mockSpanBuilder.start()) doReturn mockSpan
+        fakeTraceId = forge.aDDTraceId(fakeTraceIdString)
```
satween: > [!IMPORTANT]
> There are bunch of classes that tests `TracerInterceptor` with different configurations. The problem is that `set up` methods of them mostly copy-pasted. To reduce amount of code duplication I've extracted everything into [TracingInterceptorTestsExt](https://github.com/DataDog/dd-sdk-android/blob/tvaleev/feature/RUM-9899_2/integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTestsExt.kt)

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -799,6 +801,11 @@ internal constructor(
             return getThis()
         }
 
+        internal fun setGlobalTracerProvider(globalTracerProvider: () -> Tracer?) : R {
```
mariusc83: This provider and the localTracerFactory are still using the `io.opentracing.Tracer` ? I am not sure what is this `Tracer` interface here ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -873,11 +880,12 @@ internal constructor(
         private const val AGENT_PSR_ATTRIBUTE = "_dd.agent_psr"
         private val DEFAULT_LOCAL_TRACER_FACTORY: (SdkCore, Set<TracingHeaderType>) -> Tracer =
             { sdkCore, tracingHeaderTypes ->
-                AndroidTracer.Builder(sdkCore)
-                    // set sample rate to 100 to avoid sampling in the tracer, we are going to sample in the interceptor
-                    .setSampleRate(ALL_IN_SAMPLE_RATE)
-                    .setTracingHeaderTypes(tracingHeaderTypes)
-                    .build()
+                CoreTracer.CoreTracerBuilder((sdkCore as FeatureSdkCore).internalLogger).build()
+//                AndroidTracer.Builder(sdkCore)
```
mariusc83: why is this commented out ? shouldn't we provide the header types here ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -323,17 +323,7 @@ internal constructor(
...
-        val openTracingSpan = request.tag(Span::class.java)
-        if (openTracingSpan != null && openTracingSpan.context() is DDSpanContext) {
-            // very hacky, this is to trigger sampling decision and avoid
-            // making DDTracer#setSamplingPriorityIfNecessary public
-            tracer.inject(openTracingSpan.context(), Format.Builtin.TEXT_MAP_INJECT, TextMapInject { _, _ -> })
-            return (openTracingSpan.context() as? DDSpanContext)
-                ?.let { if (it.samplingPriority == PrioritySampling.UNSET) null else it.samplingPriority > 0 }
-        }
-
+        // TODO: should we add some dd-trace-java specific here?
```
mariusc83: I think it is ok to remove everything Opentracing related here so no need for this todo.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
mariusc83: I think I saw the typealias after. I guess here we will need to re-work a bit following the discussion we have internally. Maybe we need to introduce our local `Tracer` interface from which `OtelTracer` should inherit.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
mariusc83: so as discussed in the Slack thread we should invent our own interface `DatadogTracer` from which the `OtelTracer` will inherit and have only the needed methods in the `DatadogTracer` which will be implemented in the `OtelTracer` (`start`, `finish`, `drop`).

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/CoreTracer.java
```diff
@@ -972,6 +975,13 @@ public AgentTracer.SpanBuilder withLink(AgentSpanLink link) {
             return this;
         }
 
+        @NonNull
+        @Override
+        public AgentTracer.SpanBuilder withOrigin(@Nullable String origin) {
```
mariusc83, line 913: should we add the equivalent unit tests for this ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -387,20 +380,20 @@ internal constructor(
     }
 
     private fun extractParentContext(tracer: Tracer, request: Request): SpanContext? {
-        val tagContext = request.tag(Span::class.java)?.context()
-            ?: request.tag(TraceContext::class.java)?.toOpenTracingContext()
+        val tagContext = request.tag(Span::class.java)?.context() ?: request.tag(TraceContext::class.java)
```
mariusc83: why do we still check for `request.tag(Span::class.java)` ? this was from OpenTracing if I am not mistaken and should not be here anymore.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -439,19 +433,17 @@ internal constructor(
...
-                        DATADOG_SPAN_ID_HEADER,
-                        DATADOG_ORIGIN_HEADER -> requestBuilder.addHeader(key, value)
-                    }
+                requestBuilder
+            ) { carrier, key, value ->
+                when (key) {
+                    DATADOG_LEAST_SIGNIFICANT_64_BITS_TRACE_ID_HEADER,
+                    DATADOG_TAGS_HEADER,
+                    DATADOG_SPAN_ID_HEADER,
+                    DATADOG_ORIGIN_HEADER -> carrier.addHeader(key, value)
```
mariusc83, line 460: is the `carrier` the `requestBuilder` in this case ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -631,7 +619,9 @@ internal constructor(
         }
     }
 
-    private fun Span.drop() = (this as? MutableSpan)?.drop()
+    private fun Span.drop() {
+        // TODO (this as? MutableSpan)?.drop()
```
mariusc83: I think you should already open a JIRA ticket here and mark it with `TODO: RUM???` to not forget about it.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/otel/TraceContextExt.kt
```diff
@@ -7,34 +7,17 @@
...
-private const val BASE_16_RADIX = 16
-
-internal fun TraceContext.toOpenTracingContext(): SpanContext {
-    val traceIdAsBigInteger = parseToBigInteger(traceId)
-    val spanIdAsBigInteger = parseToBigInteger(spanId)
+internal fun TraceContext.toAgentSpanContext(): ExtractedContext {
     return ExtractedContext(
-        traceIdAsBigInteger,
-        spanIdAsBigInteger,
+        DDTraceId.fromHexOrDefault(traceId, DDTraceId.ZERO),
```
mariusc83: I think it looks fine

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/interceptor/MutableSpan.java
mariusc83: I am afraid we need to implement one to keep same logic as before.

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/DDSpan.java
```diff
@@ -322,6 +322,11 @@ public DDSpan getLocalRootSpan() {
     return context.getTrace().getRootSpan();
   }
 
+  @Override
+  public void drop() {
+    context.getTrace().unregisterSpan(this);
```
mariusc83, line 327: nice, do we have tests for this ?

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/PendingTrace.java
mariusc83: Hmm...not sure it's ok here, I am thinking about the case where you have a parent span already and you only want to create the Okhttp span for the ids. In this case dropping the OkHttp span will completely drop all the trace ?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/OtelMigration.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.okhttp.trace
+
+import com.datadog.trace.bootstrap.instrumentation.api.AgentSpan
+import com.datadog.trace.bootstrap.instrumentation.api.AgentTracer
+
+// This code aimed to reduce amount of lines in PR for review. Gonna replace with actual class just before merging v3
+typealias Tracer = AgentTracer.TracerAPI
```
mariusc83: not sure how much value is this bringing, I think it creates a bit more confusion actually at least when I do the reviews.

mariusc83 (commented):
LGTM in general. We're missing the part where we are able to inject our own tracer into the `DatadogInterceptor#builder`, see the Slack thread but I guess this can still be added into this PR.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: `Tracer` here is a typealias for AgentTracer.TracerAPI. I made this typealias only for PR ( to reduce amount of changes made and make reviewing process bit easier). I will remove typealiases with different PR, that wouldn't have any logic. You can be sure that there is no `io.opentracing.Tracer` usage anywhere where `Tracer` is used

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: yes. Using `requestBuiler` will keep context of the outer (caller) function so I use only lambda's parameters

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: It's seems like you've reviewed some outdated version of the PR. `drop()` method implementation has been added to the PR ( sorry for the inconvenience)

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/OtelMigration.kt
satween: Ok, gonna revert this

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: this is just a typealias magic. Removed all of that confusing lines

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/CoreTracer.java
satween: added 👍

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/DDSpan.java
satween: yep

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: provided 👍

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/CastExt.kt
```diff
@@ -0,0 +1,28 @@
...
+ * now could be written as:
+ * ```kotlin
+    val writer = featuredSdkCore.getFeature(Feature.TRACING_FEATURE_NAME)
+        ?.unwrap<Feature>()
+        ?.tryCastTo<com.datadog.android.trace.InternalCoreWriterProvider>()
+        ?.getCoreTracerWriter()
+ *  ```
+ */
+inline fun <reified R> Any.tryCastTo(): R? {
+    return this as R
```
aleksandr-gringauz: if it is `tryCastTo` shouldn't it be `as?` instead of `as`? Because `as` will throw an exception if it can't cast

satween: removed it completley

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/otel/TraceContextExt.kt
satween: replaced with `DatadogTracingToolkit.propagationHelper.createExtractedContext`

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: Resolved with -api layer

aleksandr-gringauz:
lgtm

but better wait also for @mariusc83 approval

aleksandr-gringauz (dismissed):

features/dd-sdk-android-trace/build.gradle.kts
```diff
@@ -45,13 +45,17 @@ android {
...
+    }
 }
 
 dependencies {
     api(project(":dd-sdk-android-core"))
+    api(project(":features:dd-sdk-android-trace-api"))
     implementation(project(":dd-sdk-android-internal"))
-    // TODO RUM-9902 This is temporary for compilation. Gonna change to implementation after removing opentracing code
-    api(project(":features:dd-sdk-android-trace-internal"))
+    implementation(project(":features:dd-sdk-android-trace-internal"))
```
mariusc83, line 58: any reason why this is exposed through `implementation` dependency now ?

features/dd-sdk-android-trace-api/build.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+    id("org.jetbrains.kotlinx.kover")
+
+    // Internal Generation
+    id("com.datadoghq.dependency-license")
+    id("apiSurface")
+    id("transitiveDependencies")
+    id("binary-compatibility-validator")
+}
+
+android {
```
mariusc83, line 42: I am still in doubts about this module, I see that is currently just used by the `dd-sdk-android-trace` module. In that sense I wonder if it's not too much ? I wonder what the team has to say about this: @0xnm , @aleksandr-gringauz ?

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/bootstrap/instrumentation/api/TagContext.java
```diff
@@ -170,7 +170,12 @@ public final Map<String, String> getTags() {
   }
 
   @Override
-  public final int getSamplingPriority() {
+  public final int getTraceSamplingPriority() {
```
mariusc83, line 173: why do you have 2 methods with different signature but returning same value ? This can be miss - leading.

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/config/TracerConfig.java
```diff
@@ -44,6 +50,7 @@ public final class TracerConfig {
   public static final String TRACE_SAMPLING_OPERATION_RULES = "trace.sampling.operation.rules";
   // JSON rules
   public static final String TRACE_SAMPLING_RULES = "trace.sampling.rules";
+  public static final String SDK_V2_COMPATIBILITY_FLAG = "v2.compatibility.enabled";
```
mariusc83, line 53: what is this flag needed for ? I see that for now it is not set from anywhere in the sdk.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
```diff
@@ -221,6 +195,16 @@ class SampleApplication : Application() {
             }
         }.build()
         Trace.enable(tracesConfig)
+
+        GlobalDatadogTracer.registerIfAbsent(
```
mariusc83, line 199: Are we keeping this only for the OkHttp instrumentation ?

mariusc83: if this is the case this will look very weird. IMO the tracing feature in the new state should do 2 things:
 - initialise with a configuration
 - provide otel implementation by using the dd-sdk-android-otel module on top
 - provide the `DatadogInterceptor` from the `dd-sdk-android-okhttp` mdule and in case this requires the `DatadogTracer` to create/open spans it should register this under the hood.

features/dd-sdk-android-trace-api/build.gradle.kts
satween: We discussed this in person, but I will put few key points, why this module was added: 

* Hides implementation details from public API available for users 
* Makes possible to easily migrate to the new version of dd-trace-java or any other tracing implementation
* Reducing ability to silently bounding any integration module to a specific implementation details. (i.e  it's not possible to cast `Trace` to `CoreTracer` anywhere except `-trace` module)

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/config/TracerConfig.java
satween: Context: `AndroidTracer` user SAMPLER_KEEP and SAMPLER_DROP constant for sampling priority. `CoreTracer` uses USER_KEEP and USER_DROP for sampling priority. 
1. If we change all SAMPLER_KEEP/SAMPLER_DROP to USER_KEEP/USER_DROP everywhere - it could affect manual traces.
2. If we change all USER_KEEP/USER_DROP to SAMPLER_KEEP/SAMPLER_DROP  it could affect manual traces made with otel (because otel implementation already in public and implemented with `CoreTracer`). 

To keep `Sample` object JSON looks same as before migration, but keep Otel implementation working same way as  it was before v3 we need to add this flag, which set to `true` if tracer created from otel provider.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
satween: >Are we keeping this only for the OkHttp instrumentation ? 

Nope, we have bunch of places where global instance is being used:

[SqliteDatabaseExt](https://github.com/DataDog/dd-sdk-android/blob/c6b7a4358626509e0fb027d25863ce74ad3afc9a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/sqlite/SqliteDatabaseExt.kt#L29), [SpanExt](https://github.com/DataDog/dd-sdk-android/blob/c33379fe77a9371b013c36fb03160759aa940581/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/SpanExt.kt#L28), [CoroutineExt](https://github.com/DataDog/dd-sdk-android/blob/c6b7a4358626509e0fb027d25863ce74ad3afc9a/integrations/dd-sdk-android-trace-coroutines/src/main/kotlin/com/datadog/android/trace/coroutines/CoroutineExt.kt#L41) and so on. 


>if this is the case this will look very weird. IMO the tracing feature in the new state should do 2 things:
> * initialise with a configuration
> * provide otel implementation by using the dd-sdk-android-otel module on top
> * provide the DatadogInterceptor from the dd-sdk-android-okhttp mdule and in case this requires the DatadogTracer to create/open spans it should register this under the hood.

That would be our best scenario, but we have few limitation here: 
1. If we force customers using `Otel` version - they would have to add desugaring support. To make it possible avoid otel dependencies we need `DatadogTracer` itself.
2. If we remove `GlobalDatadogTracer` we would have to migrate our integration modules (like okhttp, sqldelight, etc) to use `Otel` instead and they gonna require desugaring as well. Otherwise all of those integrations will have to create it's own instance locally and could miss some specific parameters that customer wants (and will only provide on configuration side).

TL;DR We need to wait until minSdk = 26 for the SDK to get rid from `GlobalDatadogTracer

features/dd-sdk-android-trace/build.gradle.kts
satween: This module contains adapters mapping api interfaces into dd-trace-java implementation. 
`:features:dd-sdk-android-trace-internal` added as `implementation` dependency in order to make it visible to current module, but hide from others.

Not  sure that I've got the question right, please correct me if you meant something else

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
mariusc83: I am not saying to have these module depending on Otel. I am suggesting that you keep the `GlobalDatadogTracer` but you register yourself this instance under the hood and you do not expose this API to the users. Correct me if I am wrong but the intent for this instance is to be used internally by all our instrumentations that require spans/traces right now as as OkHttp. ?

features/dd-sdk-android-trace/build.gradle.kts
mariusc83: ok I see it now, you need these APIs later in module like `dd-sdk-android-trace-otel` to delegate to ?

satween: Let me put bit more context here: 
* `:features:dd-sdk-android-trace-internal` - is basically dd-trace-java implementation
* `:features:dd-sdk-android-trace-api` - module that's being used everywhere ( in other modules like `-okhttp`, `-otel` and so on) to interact with tracing feature. But it only contain interfaces, without implementation. 
*`:features:dd-sdk-android-trace` - bounds `-api` to `-internal`.

 So `-trace` module is aware about `-api` and `-internal` but don't expose `-internal` for descendants that's why it is included as `implementation`

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
satween: The problem with that approach is that we need some place where `DatadogTracer` is being configured by the customer before we could put it into `GlobalDatadogTracer`. We cannot create it by ourselves because customer should provide some specific parameters (like tracingHeadersTypes for example). 

Technically we could put registering logic inside `build()` method of a `DatadogTracerBuilder` implementation, but it could lead to issues, because it's not expected from developer side that creating builder makes them globally accessible, and I personally don't like such side effects of a functions.  But I am open to discuss any other possible solution if you know some

mariusc83: Let's discuss this also tomorrow when @0xnm will be in the office, I want to hear also his opinion on this matter before going forward.

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/bootstrap/instrumentation/api/TagContext.java
satween: Yep, this method gonna be removed for TagContext. Thanks for spotting.

ci/pipelines/default-pipeline.yml
```diff
@@ -526,6 +526,23 @@ publish:release-internal:
...
+  tags: [ "arch:amd64" ]
+  only:
+    - tags
+    - develop
+  image: $CI_IMAGE_DOCKER
+  stage: publish
+  timeout: 30m
+  script:
+    - !reference [.snippets, set-publishing-credentials]
+    - ./gradlew :features:dd-sdk-android-trace-api:publishToSonatype closeSonatypeStagingRepository --stacktrace --no-daemon
```
0xnm, line 539: `closeSonatypeStagingRepository` should also be added to the task `publish:release-trace-internal` below. I cannot comment there, so commenting here.

detekt_custom.yml
```diff
@@ -759,8 +755,11 @@ datadog:
       - "java.util.concurrent.atomic.AtomicLong.constructor(kotlin.Long)"
       - "java.util.concurrent.atomic.AtomicLong.get()"
       - "java.util.concurrent.atomic.AtomicLong.set(kotlin.Long)"
+      - "java.util.concurrent.atomic.AtomicReference.compareAndSet(com.datadog.trace.core.CoreTracer?, com.datadog.trace.core.CoreTracer?)"
       - "java.util.concurrent.atomic.AtomicReference.compareAndSet(com.datadog.android.api.SdkCore?, com.datadog.android.api.SdkCore?)"
-      - "java.util.concurrent.atomic.AtomicReference.compareAndSet(io.opentracing.Tracer?, io.opentracing.Tracer?)"
+      - "java.util.concurrent.atomic.AtomicReference.compareAndSet(com.datadog.trace.bootstrap.instrumentation.api.AgentTracer.TracerAPI?, com.datadog.trace.bootstrap.instrumentation.api.AgentTracer.TracerAPI?)"
+      - "java.util.concurrent.atomic.AtomicReference.compareAndSet(com.datadog.android.trace.api.tracer.DatadogTracer?, com.datadog.android.trace.api.tracer.DatadogTracer?)"
+      - "java.util.concurrent.atomic.AtomicReference.compareAndSet(UNKNOWN, UNKNOWN)"
```
0xnm, line 762: we probably need to adjust the check to avoid having this `UNKNOWN`, let's create a ticket.

features/dd-sdk-android-logs/api/apiSurface
```diff
@@ -35,17 +35,17 @@ data class com.datadog.android.log.LogsConfiguration
     fun setEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.log.model.LogEvent>): Builder
     fun build(): LogsConfiguration
 data class com.datadog.android.log.model.LogEvent
-  constructor(Device, Os, Status, kotlin.String, kotlin.String, kotlin.String, Logger, Usr? = null, Account? = null, Network? = null, Error? = null, kotlin.String? = null, kotlin.String, kotlin.collections.MutableMap<kotlin.String, kotlin.Any?> = mutableMapOf())
+  constructor(LogEventDevice, Os, Status, kotlin.String, kotlin.String, kotlin.String, Logger, Dd, Usr? = null, Account? = null, Network? = null, Error? = null, kotlin.String? = null, kotlin.String, kotlin.collections.MutableMap<kotlin.String, kotlin.Any?> = mutableMapOf())
```
0xnm, line 38: it seems these changes should already be in v3, is `develop` only merged in the tracing branch?

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/propagation/DatadogPropagation.kt
```diff
@@ -0,0 +1,50 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+package com.datadog.android.trace.api.propagation
+
+import com.datadog.android.trace.api.span.DatadogSpanContext
+
+/**
+ * Provides an interface for injecting and extracting span context to/from specified carriers.
+ * Used for propagating context to http headers
```
0xnm:
```suggestion
 * Used for propagating context to http headers.
```

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/span/DatadogSpan.kt
```diff
@@ -0,0 +1,204 @@
...
+    fun addThrowable(throwable: Throwable)
+
+    /**
+     * Associates a throwable with the current span, marking it as an error
+     * and capturing the provided throwable for additional context with a specified error priority.
+     *
+     * @param throwable The throwable to associate with the current span.
+     * @param errorPriority The priority level of the error, represented as a byte.
+     */
+    fun addThrowable(throwable: Throwable, errorPriority: Byte)
```
0xnm: probably this shouldn't be exposed to the end-user, at least in a such shape, because it is not clear what are the possible values for `errorPriority` (and what is maps to).

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/span/DatadogSpanWriter.kt
```diff
@@ -3,13 +3,10 @@
...
-public interface DDScopeEvent {
-
-  void start();
-
-  void finish();
-}
+/**
+ * A writer is responsible for sending collected spans.
+ * This wrapper is required to not expose CoreTracer wrapper to dependent modules
+ * */
```
0xnm:
```suggestion
/**
 * A writer is responsible for sending collected spans.
 * This wrapper is required to not expose [CoreTracer] wrapper to dependent modules
 */
```

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/tracer/DatadogTracer.kt
```diff
@@ -0,0 +1,80 @@
...
+     * requiring an active span will use this one until it is explicitly deactivated.
+     *
+     * @param span The span to be activated. Represents the logical unit of work being traced.
+     * @return An instance of [DatadogScope] representing the activated scope.
+     */
+    fun activateSpan(span: DatadogSpan): DatadogScope?
+
+    /**
+     * Activates the provided span within the current context of the tracer.
+     * If `asyncPropagating` is set to true, the span is propagated asynchronously.
```
0xnm: what does it mean exactly? should we expose this method?

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/tracer/DatadogTracer.kt
```diff
@@ -0,0 +1,80 @@
...
+     * @return the active [DatadogSpan].
+     */
+    fun activeSpan(): DatadogSpan?
+
+    /**
+     * Provides an implementation of the [DatadogPropagation] interface to be used for span propagation.
+     *
+     * @return An instance of [DatadogPropagation]
+     */
+    fun propagate(): DatadogPropagation = NoOpDatadogPropagation()
```
0xnm, line 33: what is the need to return default value in the interface declaration?

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/DDTraceId.java
```diff
@@ -54,6 +54,23 @@ public static DDTraceId fromHex(String s) throws NumberFormatException {
...
 
+  /**
+   * Creates a new {@link DDTraceId} from the given {@link #toHexString() hexadecimal
+   * representation}.
+   *
+   * @param s The hexadecimal {@link String} representation of a {@link DD128bTraceId} to parse.
+   * @return A new {@link DDTraceId} instance or DDTraceId.ZERO if the given {@link #toHexString()
+   * hexadecimal String} does not represent a valid number.
+   */
+  public static DDTraceId fromHexOrDefault(String s, DDTraceId defaultValue) throws NumberFormatException {
```
0xnm: in fact it doesn't throw `NumberFormatException`, it is handled inside

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/PendingTrace.java
```diff
@@ -323,7 +330,7 @@ private PublishState decrementRefAndMaybeWrite(boolean isRootSpan) {
             // Finished root with pending work ... delay write
             pendingTraceBuffer.enqueue(this);
             return PublishState.ROOT_BUFFERED;
-        } else if (partialFlushMinSpans > 0 && size() >= partialFlushMinSpans) {
```
0xnm, line 326: is this change needed?

features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/common/sampling/SamplerTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    ) {
+        // Given
+        whenever(mockConfig.traceSampleRate).thenReturn(fakeSampleRate)
+
+        // When
+        val actual = Sampler.Builder.forConfig(mockConfig, null)
+
+        // Then
+        assertThat(actual).isInstanceOf(RateByServiceTraceSampler::class.java)
+        assertThat((actual as RateByServiceTraceSampler).sampleRate).isEqualTo(fakeSampleRate)
```
0xnm: to avoid additional cast after type check you can do like this:

```suggestion
        check(actual is RateByServiceTraceSampler)
        assertThat(actual.sampleRate).isEqualTo(fakeSampleRate)
```

in case of the wrong type the error message will also be clear enough

features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/core/DDSpanTest.kt
```diff
@@ -507,7 +509,28 @@ internal class DDSpanTest : DDCoreSpecification() {
             .isEqualTo(SamplingMechanism.SPAN_SAMPLING_RATE)
         assertThat(span.getTag(DDSpanContext.SPAN_SAMPLING_RULE_RATE_TAG)).isEqualTo(rate)
         assertThat(span.getTag(DDSpanContext.SPAN_SAMPLING_MAX_PER_SECOND_TAG)).isEqualTo(expectedLimit)
-        assertThat(span.samplingPriority()).isEqualTo(PrioritySampling.UNSET.toInt())
+        assertThat(span.traceSamplingPriority).isNull()
+    }
+
+    @Test
+    fun `W drop T unregisterSpan is called`() {
```
0xnm:
```suggestion
    fun `M call unregisterSpan W drop is called`() {
```

features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/core/PendingTraceTest.kt
```diff
@@ -187,6 +187,37 @@ internal class PendingTraceTest : PendingTraceTestBase() {
         assertThat(trace.spans).containsExactlyInAnyOrder(unfinishedSpan, unfinishedSpan2)
     }
 
+    @Test
+    fun `do not drop root span when drop is called`() {
```
0xnm:
```suggestion
    fun `do not drop root span when unregisterSpan is called`() {
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/GlobalDatadogTracer.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    private val instance = AtomicReference<DatadogTracer?>()
+
+    /**
+     * Registers the provided tracer as the global tracer if no tracer is currently registered.
+     *
+     * @param tracer The tracer to register as the global tracer.
+     * @return `true` if the tracer was successfully registered, or `false` if a tracer was already registered.
+     */
+    @Synchronized
```
0xnm: it is not really needed here, because backing `AtomicReference` will do the proper atomic swap using `compareAndSet`.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/GlobalDatadogTracer.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun getOrNull(): DatadogTracer? = instance.get()
+
+    /**
+     * Clears the current instance of the global Datadog tracer.
+     *
+     * This method sets the internal tracer instance to null, effectively
+     * removing any active tracer currently held in the global state.
+     * The general purpose is to use it for test implementation.
+     */
+    @Synchronized
```
0xnm: even if multiple methods enter here at the same time, it shouldn't cause the problem since they do the same

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/DatadogPropagationAdapter.kt
```diff
@@ -0,0 +1,46 @@
...
+    override fun <C> extract(
+        carrier: C,
+        getter: (carrier: C, classifier: (String, String) -> Boolean) -> Unit
+    ): DatadogSpanContext? {
+        return delegate.extract(carrier) { car, cls -> getter(car, cls::accept) }
+            ?.let { DatadogSpanContextAdapter(it) }
+    }
+
+    private fun constructErrorMessage(klass: KClass<*>) = "DatadogPropagationAdapter supports only" +
+        " DatadogSpanContextAdapter instancies for injection but ${klass.simpleName} is given"
```
0xnm:
```suggestion
        " DatadogSpanContextAdapter instances for injection but ${klass.simpleName} is given"
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/DatadogTracingTest.kt
```diff
@@ -0,0 +1,160 @@
...
+        val tracer = DatadogTracing.newTracerBuilder(mockSdkCore).build() as DatadogTracerAdapter
+
+        // Then
+        val coreTracer = tracer.delegate as CoreTracer
+        val writer: Writer = coreTracer.getFieldValue("writer")
+        assertThat(writer).isInstanceOf(NoOpWriter::class.java)
+    }
+
+    @Test
+    fun `M log a maintainer error W build { TracingFeature not implementing InternalCoreTracerWriterProvider }`() {
```
0xnm: `InternalCoreWriterProvider` probably?

features/dd-sdk-android-trace/src/main/json/trace/span-schema.json
```diff
@@ -16,17 +16,9 @@
     "error",
     "type",
     "meta",
-    "metrics",
-    "device",
-    "os"
+    "metrics"
   ],
   "properties": {
-    "device": {
```
0xnm, line 24: why do we move them under the `meta` now? is it aligned with iOS?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/DatadogPropagationAdapterTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        val carrier = Any()
+
+        // When
+        testedPropagation.inject(UnsupportedDatadogSpanContextImplementation(), carrier) { _, _, _ -> }
+
+        // Then
+        mockInternalLogger.verifyLog(
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.USER,
+            "DatadogPropagationAdapter supports only DatadogSpanContextAdapter instancies for injection " +
```
0xnm:
```suggestion
            "DatadogPropagationAdapter supports only DatadogSpanContextAdapter instances for injection " +
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/DatadogSpanContextAdapterTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        assertThat(actual).isEqualTo(expectedMap)
+    }
+
+    @Test
+    fun `M return delegate#traceId W traceId is called`(forge: Forge) {
+        // Given
+        val ddTraceId = forge.getForgery<DDTraceId>()
+
+        // When
+        whenever(mockAgentSpanContext.traceId).thenReturn(ddTraceId)
```
0xnm: this should be in `Given`

instrumented/integration/proguard-rules.pro
```diff
@@ -11,9 +11,10 @@
     *;
 }
 
-# Required because we need access to GlobalTracer isRegistered property to reset it through reflection
--keepnames class io.opentracing.util.GlobalTracer {
-    private boolean isRegistered;
+# Required because we need access to GlobalDatadogTracer getOrNull method to reset it through reflection
```
0xnm, line 14: this is already declared in the `trace` module, so I guess re-declaration is not needed?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -210,13 +204,18 @@ internal constructor(
         sdkCore: InternalSdkCore,
         chain: Interceptor.Chain,
         request: Request,
-        tracer: Tracer
+        tracer: DatadogTracer
     ): Response {
-        val span = buildSpan(tracer, request)
-        val isSampled = span.sample(tracer, request)
+        val span: DatadogSpan = buildSpan(tracer, request)
```
0xnm:
```suggestion
        val span = buildSpan(tracer, request)
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -488,20 +479,20 @@ internal constructor(
...
+    private fun handleW3CNotSampledHeaders(span: DatadogSpan, requestBuilder: Request.Builder) {
         if (traceContextInjection == TraceContextInjection.ALL) {
-            val traceId = span.context().traceIdAsHexString()
-            val spanId = span.context().toSpanId()
+            val traceId = span.context().traceId.toHexString()
+            val spanId = span.context().spanId.toString()
             requestBuilder.addHeader(
                 W3C_TRACEPARENT_KEY,
-                @Suppress("UnsafeThirdPartyFunctionCall") // Format string is static
+                @Suppress("UnsafeThirdPartyFunctionCall", "InvalidStringFormat") // Format string is static
```
0xnm, line 489: why `InvalidStringFormat` is raised?

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/utils/forge/TraceContextFactory.kt
```diff
@@ -13,7 +13,7 @@ import fr.xgouchet.elmyr.ForgeryFactory
 internal class TraceContextFactory : ForgeryFactory<TraceContext> {
 
     override fun getForgery(forge: Forge): TraceContext {
-        val fakeTraceId = forge.aStringMatching("[0-9a-f]{32}")
+        val fakeTraceId = forge.aStringMatching("[0-9a-f]{31}")
```
0xnm, line 16: why we are changing this?

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/HeadBasedSamplingTest.kt
```diff
@@ -135,7 +138,7 @@ class HeadBasedSamplingTest {
                     leastSignificantTraceId.toLong().toHexString().padStart(16, '0')
                 )
                 hasMostSignificant64BitsTraceId(mostSignificantTraceId)
-                hasSpanId(spanId.toLong().toHexString())
+                hasSpanId(spanId.cleanDecimalFormat())
```
0xnm: nit: it doesn't really clear what `clean` means in the method name, maybe it is worth renaming it?

detekt_custom.yml
satween: RUM-11338

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/span/DatadogSpan.kt
satween: Good catch, moved to internal side

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/utils/forge/TraceContextFactory.kt
satween: There is an issue in forgery lib logic, which gonna create 33 symbol string from this regexp, which will lead to exception thrown from [this](https://github.com/DataDog/dd-sdk-android/blob/c6b7a4358626509e0fb027d25863ce74ad3afc9a/features/dd-sdk-android-trace/src/main/java/com/datadog/trace/api/DD128bTraceId.java#L85) line

instrumented/integration/proguard-rules.pro
satween: Not tested yet, gonna remove in separate ticket RUM-11362 (after we make a decision regarding publicitly available DatadogTracer)

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/utils/forge/TraceContextFactory.kt
0xnm: Maybe then we can create 2 strings of 16 chars and concatenate them to have a right 32 chars string. Not sure what is the exact issue.

satween: Well then we would have to declare them as "[0-9a-f]{15}".  "[0-9a-f]{16}" works now because the limit is 32, so it parsed correctly. I think it's more complicated that the record above, WDYT?

features/dd-sdk-android-trace/src/main/json/trace/span-schema.json
satween: We found on dogfooding that we lost all manual traces. 
iOS also lost them. That was happened because we put `device` and `os` in the root of the object, but APM intake was expecting them only in meta object. iOS also reverted this

features/dd-sdk-android-logs/api/apiSurface
satween: This change appears because we have to revert `os` and `device` back to meta in this [commit](https://github.com/DataDog/dd-sdk-android/pull/2708/commits/d6bebddadc05313bddf938914a4683aa6e402e60).  The change was made in this branch, but not in separate ticket to speed up next dogfooding interation

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: This is an issue with our detekt rule, gonna be fixed in separate ticket

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/tracer/DatadogTracer.kt
satween: As described [here](https://github.com/DataDog/dd-sdk-android/blob/tvaleev/feature/RUM-9899_2/features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/propagation/NoOpDatadogPropagation.kt) our NoOpImplementation doesn't support generics so such trick is the only way to make  NoOpDatadogTracer generated.

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/tracer/DatadogTracer.kt
satween: this method is called only from OTel and looks like good candidate to be internal. Left only for compatibility reasons

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/PendingTrace.java
satween: here we aligning new tracer implementation with [previous](https://github.com/DataDog/dd-sdk-android/blob/98865fe531895d61ff8fd9a2188d0256822bcd5e/features/dd-sdk-android-trace/src/main/java/com/datadog/opentracing/PendingTrace.java#L303) one

instrumented/integration/proguard-rules.pro
0xnm: What I mean here is that it is already declared in the Proguard rules file of the Tracing module and so will be added from there.

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/span/DatadogSpan.kt
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Represents the duration of a span in nanoseconds.
+     */
+    val durationNano: Long
+
+    /**
+     * Represents the start time of the span in nanoseconds.
+     */
+    val startTimeNanos: Long
```
0xnm, line 67: minor: it is better to have consistent naming, so it should be `durationNanos`

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -1,24 +1,13 @@
...
-  override fun toString(): String
-  companion object 
-    fun logThrowable(io.opentracing.Span, Throwable)
-    fun logErrorMessage(io.opentracing.Span, String)
+object com.datadog.android.trace.DatadogTracing
+  fun newTracerBuilder(com.datadog.android.api.SdkCore = Datadog.getInstance()): com.datadog.android.trace.api.tracer.DatadogTracerBuilder
+object com.datadog.android.trace.GlobalDatadogTracer
+  fun registerIfAbsent(com.datadog.android.trace.api.tracer.DatadogTracer): Boolean
+  fun get(): com.datadog.android.trace.api.tracer.DatadogTracer
+  fun getOrNull(): com.datadog.android.trace.api.tracer.DatadogTracer?
```
0xnm, line 6: do we need both? maybe we should leave only one of them?

0xnm (dismissed):

0xnm (approved):

features/dd-sdk-android-trace/api/apiSurface
satween: Gonna address this in next PR if possible

features/dd-sdk-android-trace-api/src/main/kotlin/com/datadog/android/trace/api/span/DatadogSpan.kt
satween: Gonna address this in next PR if possible

PR #2711, opened 2025-06-06
RUM-9917 improve speed of running core-it suite

xgouchet:
### What does this PR do?

Speed up the `core-it` test suite and reduce the time spent in CI

### DIFF
--- a/reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
+++ b/reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
@@ -120,7 +120,7 @@ class FeatureScopeTest : MockServerTest() {
                 .withTrackingConsent(trackingConsent)
                 .receivedData(fakeBatchData, fakeBatchMetadata)
             true
-        }.doWait(LONG_WAIT_MS)
+        }.doWait(MEDIUM_WAIT_MS)
     }
 
     @Test
@@ -155,7 +155,7 @@ class FeatureScopeTest : MockServerTest() {
                 .withTrackingConsent(trackingConsent)
                 .receivedData(fakeBatchData, fakeBatchMetadata)
             true
-        }.doWait(LONG_WAIT_MS)
+        }.doWait(MEDIUM_WAIT_MS)
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/FeatureScopeTest.kt
```diff
@@ -120,7 +120,7 @@ class FeatureScopeTest : MockServerTest() {
                 .withTrackingConsent(trackingConsent)
                 .receivedData(fakeBatchData, fakeBatchMetadata)
             true
-        }.doWait(LONG_WAIT_MS)
+        }.doWait(MEDIUM_WAIT_MS)
```
satween, line 123: Any option to avoid real time awaiting? I remember that we have several `TimeProvider`s ....

xgouchet: The `MEDIUM_WAIT_MS` here is only a timeout, so it will wait until the condition inside the lambda passes, not a millisecond more.

ambushwork (dismissed):

reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/InternalSdkCoreTest.kt
```diff
@@ -406,8 +408,8 @@ class InternalSdkCoreTest : MockServerTest() {
             .apply {
                 _InternalProxy.allowClearTextHttp(this)
             }
-            .setBatchSize(forge.getForgery())
-            .setUploadFrequency(forge.getForgery())
+            .setBatchSize(BatchSize.SMALL)
```
0xnm, line 412: probably it should be different here? Like big batches + `setBatchProcessingLevel(BatchProcessingLevel.HIGH)`, since we want to upload as much as possible as fast as possible.

xgouchet: Smaller batches means that they are ready to be uploaded faster. 
Good idea on `setBatchProcessingLevel`, I'll update the PR

0xnm: > Smaller batches means that they are ready to be uploaded faster

oh yeah, this is true as well. We can have more upload iterations for them, but `BatchProcessingLevel.HIGH` should solve that.

PR #2716, opened 2025-06-11
RUM-10314: Read feature context only when explicitly requested by the caller

0xnm:
### What does this PR do?

We now have a possibility to update feature context from multiple threads, which imposes the need to use write/read locks.

This means that if `withWithContext`, `withContext`, etc. methods are made which provide `DatadogContext` in the callback, if there is any feature context update is ongoing, we need to wait for its completion (since write lock is acquired, making it impossible to read until unlocked) to put the right value.

However, caller may not need the particular feature context, so this wait may be for nothing.

This PR adds an additional argument `withFeatureContext` to the methods mentioned above which is used to **explicitly** indicate what feature contexts caller is interested in. This makes API a bit more verbose, but at the same time we avoid the possible redundant wait time.

That change won't be needed if all feature updates/reads would happen on the context processing thread, but it is not the case.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
@@ -24,6 +24,9 @@ interface FeatureScope {
 
     /**
      * Utility to write an event, asynchronously.
+     * @param withFeatureContexts Feature contexts ([DatadogContext.featuresContext] property) to include
+     * in the [DatadogContext] provided. The value should be the feature names as declared by [Feature.name].
+     * Default is empty, meaning that no feature contexts will be included.
      * @param callback an operation called with an up-to-date [DatadogContext]
      * and an [EventWriteScope]. Callback will be executed on a single context processing worker thread. Execution of
      * [EventWriteScope] will be done on a worker thread from I/O pool.
@@ -31,16 +34,21 @@ interface FeatureScope {
      */
     @AnyThread
     fun withWriteContext(
+        withFeatureContexts: Set<String> = emptySet(),
         callback: (datadogContext: DatadogContext, write: EventWriteScope) -> Unit
     )
 
     /**
      * Utility to read current [DatadogContext], asynchronously.
+     * @param withFeatureContexts Feature contexts ([DatadogContext.featuresContext] property) to include
+     * in the [DatadogContext] provided. The value should be the feature names as declared by [Feature.name].
+     * Default is empty, meaning that no feature contexts will be included.
      * @param callback an operation called with an up-to-date [DatadogContext].
      * [DatadogContext] will have a state created at the moment this method is called.
      */
     @AnyThread
     fun withContext(
+        withFeatureContexts: Set<String> = emptySet(),
         callback: (datadogContext: DatadogContext) -> Unit
     )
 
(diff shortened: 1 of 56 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
```diff
@@ -44,7 +44,7 @@ internal class DataUploadRunnable(
         var uploadAttempts = 0
         var lastBatchUploadStatus: UploadStatus? = null
         if (isNetworkAvailable() && isSystemReady()) {
-            val context = contextProvider.context
+            val context = contextProvider.context(withFeatureContexts = emptySet())
```
mariusc83: shouldn't be emptySet() by default ?

mariusc83 (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/data/upload/DataUploadRunnable.kt
0xnm: No, this is `ContextProvider` interface here which is internal and doesn't have default arguments (since in most of the cases it should just pass argument from the caller, like `withWriteContext`, etc.).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/ContextProvider.kt
```diff
@@ -12,5 +12,5 @@ internal interface ContextProvider {
     // TODO RUM-3784 lifecycle checks may be needed for the cases when context is requested
     //  when datadog is not initialized yet/anymore (case of UploadWorker, other calls site
     //  should be in sync with lifecycle)
-    val context: DatadogContext
+    fun context(withFeatureContexts: Set<String>): DatadogContext
```
ambushwork: since we change `val` to `fun`, maybe we need to rename it to `getContext` or even `get()` since `ContextProvider` already indicates the thing it provides?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/InternalSdkCore.kt
```diff
@@ -119,5 +119,5 @@ interface InternalSdkCore : FeatureSdkCore {
      * @return the current DatadogContext, or null
      */
     @InternalApi
-    fun getDatadogContext(): DatadogContext?
+    fun getDatadogContext(withFeatureContexts: Set<String> = emptySet()): DatadogContext?
```
ambushwork, line 125: It's better to update the documentation of the argument here even though it's an internal one.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -24,35 +24,41 @@ interface FeatureScope {
 
     /**
      * Utility to write an event, asynchronously.
+     * @param withFeatureContexts Feature contexts to include in the [DatadogContext] provided. Default is empty.
```
ambushwork: With this explanation I am not sure what does the string represents in the `Set`, maybe it's better to explain that the feature name should be provided in the `Set`?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureScope.kt
```diff
@@ -24,35 +24,41 @@ interface FeatureScope {
...
      * Same as [withWriteContext] but will be executed in the blocking manner.
      *
+     * @param withFeatureContexts Feature contexts to include in the [DatadogContext] provided. Default is empty.
+     *
      * **NOTE**: This API is for the internal use only and is not guaranteed to be stable.
      */
     @AnyThread
     @InternalApi
-    fun getWriteContextSync(): Pair<DatadogContext, EventWriteScope>?
+    fun getWriteContextSync(withFeatureContexts: Set<String> = emptySet()): Pair<DatadogContext, EventWriteScope>?
```
ambushwork, line 67: Should we be more precise in the documentation that what's the consequence by giving a set of feature names as the description of the PR does? and what the default values will lead, like not waiting for other feature context update, etc..?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/ContextProvider.kt
0xnm: Updated to `getContext` rather than `get` in order to have a bit simpler text-based search.

ambushwork (approved):

PR #2718, opened 2025-06-11
RUM-6171 use FileObserver to limit calls to filesystem

xgouchet:
### What does this PR do?

Use the file observer to keep an up to date list of batch files instead of calling `listFiles` each time. 

### Motivation

Performance improvement

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
@@ -6,6 +6,11 @@
 
 package com.datadog.android.core.internal.persistence.file.batch
 
+import android.os.FileObserver
+import android.os.FileObserver.CREATE
+import android.os.FileObserver.DELETE
+import android.os.FileObserver.MOVED_TO
+import android.os.FileObserver.MOVED_FROM
 import androidx.annotation.WorkerThread
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.core.internal.metrics.BatchClosedMetadata
@@ -20,7 +25,6 @@ import com.datadog.android.core.internal.persistence.file.lengthSafe
 import com.datadog.android.core.internal.persistence.file.listFilesSafe
 import com.datadog.android.core.internal.persistence.file.mkdirsSafe
 import java.io.File
-import java.io.FileFilter
 import java.util.Locale
 import java.util.concurrent.atomic.AtomicInteger
 import kotlin.math.roundToLong
@@ -36,8 +40,6 @@ internal class BatchFileOrchestrator(
     private val pendingFiles: AtomicInteger = AtomicInteger(0)
 ) : FileOrchestrator {
 
-    private val fileFilter = BatchFileFilter()
-
     // Offset the recent threshold for read and write to avoid conflicts
     // Arbitrary offset as ±5% of the threshold
     @Suppress("UnsafeThirdPartyFunctionCall") // rounded Double isn't NaN
@@ -51,6 +53,37 @@ internal class BatchFileOrchestrator(
     private var previousFileItemCount: Long = 0
     private var lastFileAccessTimestamp: Long = 0L
     private var lastCleanupTimestamp: Long = 0L
+
+    private val knownFiles: MutableSet<File> =
+        rootDir.listFilesSafe(internalLogger)?.toMutableSet() ?: mutableSetOf()
+
+    @Suppress("DEPRECATION") // Recommended constructor only available in API 29 Q
+    internal val fileObserver = object : FileObserver(rootDir.path, FILE_OBSERVER_MASK) {
+        override fun onEvent(event: Int, name: String?) {
+            if (!name.isNullOrEmpty() && name.isBatchFileName) {
+                val file = File(rootDir, name)
+                when (event) {
+                    MOVED_TO, CREATE -> {
+                        synchronized(knownFiles) {
+                            knownFiles.add(file)
+                        }
+
+                    }
+
+                    MOVED_FROM, DELETE -> {
+                        synchronized(knownFiles) {
+                            knownFiles.remove(file)
+                        }
+                    }
+                }
+            }
+        }
+
+    }
+
+    init {
+        fileObserver.startWatching()
+    }
 
     // region FileOrchestrator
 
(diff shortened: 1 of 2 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -52,6 +53,37 @@ internal class BatchFileOrchestrator(
...
+                        }
+                    }
+                }
+            }
+        }
+
+    }
+
+    init {
+        fileObserver.startWatching()
```
mariusc83, line 85: should we stop watching at some point ? not sure when though or if needed

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -52,6 +53,37 @@ internal class BatchFileOrchestrator(
...
     private var lastCleanupTimestamp: Long = 0L
 
+    private val knownFiles: MutableSet<File> =
+        rootDir.listFilesSafe(internalLogger)?.toMutableSet() ?: mutableSetOf()
+
+    @Suppress("DEPRECATION") // Recommended constructor only available in API 29 Q
+    internal val fileObserver = object : FileObserver(rootDir.path, CREATE or DELETE) {
+        override fun onEvent(event: Int, name: String?) {
+            when (event) {
+                CREATE -> {
```
ambushwork: Question: is it possible that `MOVED_FROM` or `MOVED_TO` also cause the change of the file list in the path?

ambushwork (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
xgouchet: In theory it shouldn't but I'll add it just to be sure

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
xgouchet: In theory yes but not sure when either

ambushwork (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -52,6 +50,37 @@ internal class BatchFileOrchestrator(
     private var lastFileAccessTimestamp: Long = 0L
     private var lastCleanupTimestamp: Long = 0L
 
+    private val knownFiles: MutableSet<File> =
+        rootDir.listFilesSafe(internalLogger)?.toMutableSet() ?: mutableSetOf()
+
+    @Suppress("DEPRECATION") // Recommended constructor only available in API 29 Q
+    internal val fileObserver = object : FileObserver(rootDir.path, CREATE or DELETE) {
```
0xnm: Mask is saying that only `CREATE` or `DELETE` events will be watched, but `MOVED_TO`, `MOVED_FROM` are mentioned in the code as well.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -52,6 +50,37 @@ internal class BatchFileOrchestrator(
...
+                        synchronized(knownFiles) {
+                            knownFiles.add(File(rootDir, name))
+                        }
+                    }
+                }
+
+                MOVED_FROM, DELETE -> {
+                    if (!name.isNullOrEmpty() && name.isBatchFileName) {
+                        synchronized(knownFiles) {
+                            knownFiles.removeAll { it.name == name }
```
0xnm: given that `knownFiles` is a `Set<File>` and `File.equals` relies on the path comparison anyway ([link](https://cs.android.com/android/platform/superproject/+/android15-qpr2-release:libcore/ojluni/src/main/java/java/io/File.java;l=2109)), won't it be simpler to use `.remove` call instead of iterating?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
0xnm: We could use `Object.finalize`, but here it is Kotlin `Any`, so no counterpart method.

We probably need to still find a way to stop watching, because if `Datadog.stop` is called, we will leak this subscription.

0xnm: Apparently we should be good, because there is `stopWatching` call in the `FileObserver.finalize` ([link](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/os/FileObserver.java;l=258-260?q=FileObserver)).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -52,6 +54,37 @@ internal class BatchFileOrchestrator(
     private var lastFileAccessTimestamp: Long = 0L
     private var lastCleanupTimestamp: Long = 0L
 
+    private val knownFiles: MutableSet<File> =
+        rootDir.listFilesSafe(internalLogger)?.toMutableSet() ?: mutableSetOf()
```
0xnm, line 58: I realized the it may bring the following concern: this will be called on the user thread during SDK initialization and since it is I/O call, it may slow down SDK start (especially if there is a lot of files on the disk).

Instead, we maybe can to initial read in the `getWritableFile` / `getReadableFile` methods (well, on any initial access of this property), this will shift it to the worker thread improving SDK initialization.

Same may apply to the `root` folder creation.

satween:
We have similar PR: https://github.com/DataDog/dd-sdk-android/pull/2720

PR #2720, opened 2025-06-12
RUM-6171 use FileObserver to limit calls to filesystem

0xnm:
### What does this PR do?

The same as https://github.com/DataDog/dd-sdk-android/pull/2718, just branch moved directly to the repo.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -148,55 +173,58 @@ internal class BatchFileOrchestrator(
...
+                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
+                { ERROR_CANT_CREATE_ROOT.format(Locale.US, rootDir.path) }
+            )
+        }
+        created
+    }
+
+    @Synchronized
+    private fun startFileObserverIfNotStarted() {
+        if (isFileObserverStarted.compareAndSet(false, true)) {
```
0xnm: In theory CAS operation `isFileObserverStarted.compareAndSet(false, true)` already guards the logical branch from parallel execution, so `@Synchronized` is not needed. Do I miss something?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -148,55 +173,58 @@ internal class BatchFileOrchestrator(
         return pendingFiles.decrementAndGet()
     }
 
-    @Suppress("LiftReturnOrAssignment", "ReturnCount")
+    @Suppress("ReturnCount")
     private fun isRootDirValid(): Boolean {
```
0xnm, line 175: This will be called from `getRootDir` <- ... <- `SdkFeature.prepareStorage`, meaning it will be called on the caller thread. In most of the cases it will be main thread during app startup, so we need to make sure that these changes don't have significant impact perf-wise at the SDK initialization.

ambushwork (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
kikoveiga: Removed it.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
kikoveiga: I ran some benchmarks on my machine, not forgetting that real Android devices might be slower (3x-5x). In the worst case of having 10 000 files, there is a one-time init cost of ~35 ms (given ~7 ms was the result on my machine × 5 for adjustment to Android device). After init, each operation becomes basically free (~0.2 ms) since it reads from memory. The old approach would pay ~35 ms on every call to `getWritableFile()`, `getReadableFile()`, `getAllFiles()` or `getFlushableFiles()`.

0xnm: Can you please attach the table with different scenarios (as I also mentioned above, it would be nice to see the cost of write call in case of different volume of files on the disk)? 35ms for the init on the main thread at the app startup may be costly.

kikoveiga: Android Estimates (×5 multiplier for slower storage)

  | Files  | OLD: Per Op | NEW: Init (one-time) | NEW: Per Op |
  |--------|-------------|----------------------|-------------|
  | 0      | 0.07 ms     | 0.1 ms               | 0.07 ms     |
  | 100    | 0.44 ms     | 0.4 ms               | 0.01 ms     |
  | 1,000  | 3.5 ms      | 3.6 ms               | 0.09 ms     |
  | 10,000 | 36.5 ms     | 36 ms                | 0.9 ms      |

  Summary:
  - OLD approach: No init cost, but pays full disk I/O on every getWritableFile(), getReadableFile(), getAllFiles(), getFlushableFiles() call
  - NEW approach: One-time init cost at SDK startup, then each operation reads from memory (~40x faster)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -52,7 +54,32 @@ internal class BatchFileOrchestrator(
     private var previousFile: File? = null
     private var previousFileItemCount: Long = 0
     private var lastFileAccessTimestamp: Long = 0L
-    private var lastCleanupTimestamp: Long = 0L
+    private val lastCleanupTimestamp = AtomicLong(0L)
+    private var isFileObserverStarted = AtomicBoolean(false)
```
0xnm:
```suggestion
    private val isFileObserverStarted = AtomicBoolean(false)
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -150,55 +175,57 @@ internal class BatchFileOrchestrator(
...
+        } else {
+            createRootDirectory()
+        }
+
+        if (isValid) startFileObserverIfNotStarted()
+        return isValid
+    }
+
+    private fun createRootDirectory(): Boolean = synchronized(rootDir) {
+        val created = rootDir.existsSafe(internalLogger) || rootDir.mkdirsSafe(internalLogger)
```
0xnm, line 207: Previous code was making sure that there is no race in the directory creation by imposing `synchronized` block and making another check inside that folder is indeed not created yet at the beginning of that block.

Should we also use such logic inside this function?

kikoveiga: I think this is still done because this function uses `synchronized` and the short-circuit `||`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
kikoveiga: Done, thank you!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
aleksandr-gringauz: > In most of the cases it will be main thread during app startup

IIUC on startup `getRootDir` is called in [ConscentAwareFileMigrator](https://github.com/DataDog/dd-sdk-android/blob/19c66239fe54d6c1a962711f3b1b8cd7471301f8/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/ConsentAwareFileMigrator.kt#L50).

And this migration is scheduled on the [persistanceExecutor](https://github.com/DataDog/dd-sdk-android/blob/19c66239fe54d6c1a962711f3b1b8cd7471301f8/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/advanced/ConsentAwareFileOrchestrator.kt#L97)

So it shouldn't be called on main thread.

PR #2722, opened 2025-06-12
Xgouchet/automotive

xgouchet:
### What does this PR do?

A adds a sample app for automotive support


sample/auto/.gitignore
0xnm, line 1: maybe we should create a folder `sample/automotive` instead of `sample/auto`? Android Auto is a different thing, so it may be a confusion here by looking on the name.

sample/auto/build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+        multiDexEnabled = true
+
+        buildFeatures {
+            buildConfig = true
+        }
+
+        configureFlavorForAutoApp(project.rootDir)
+    }
+
+    namespace = "com.datadog.sample.auto"
```
0xnm: following the suggestion above

```suggestion
    namespace = "com.datadog.sample.automotive"
```

sample/auto/build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+
+dependencies {
+    // Datadog Libraries
+    implementation(project(":features:dd-sdk-android-logs"))
+    implementation(project(":features:dd-sdk-android-rum"))
+    implementation(project(":integrations:dd-sdk-android-okhttp"))
+    implementation(project(":integrations:dd-sdk-android-okhttp-otel"))
+
+    // Desugaring SDK
+    coreLibraryDesugaring(libs.androidDesugaringSdk)
```
0xnm: do we need this if min API is 29? It should be good for Otel support already.

sample/auto/proguard-rules.pro
0xnm, line 1: probably not needed since empty

sample/auto/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,48 @@
...
+                <category android:name="androidx.car.app.category.NAVIGATION"/>
+            </intent-filter>
+        </service>
+
+        <meta-data
+            android:name="androidx.car.app.minCarApiLevel"
+            android:value="1"/>
+    </application>
+
+</manifest>
```
0xnm: minor: missing newline at the EOF. This applies also to other XML files here.

```suggestion
</manifest>

```

0xnm (commented):
useful addition 👍

PR #2723, opened 2025-06-12
RUM-9527: Defer drawable copy to work thread in Session Replay

ambushwork:
### What does this PR do?

This PR defers the drawable copy from main thread to work thread in Session Replay to fix the crash when Android View wants to draw the original drawable on UI thread.

#### Why the crash happens?
To draw the drawable to generate bitmap for Session Replay image wireframe, we will copy the drawable in case of any impact of the original one, for most of the implementations of `Drawable` it has no problem, except `MaterialShapeDrawable`.

`MaterialShapeDrawable` holds an singleton of [`ShapeAppearancePathProvider`](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/shape/MaterialShapeDrawable.java#L174)on main thread, if we copy the drawable on main thread, the same path held by this singleton will be used for both drawing, which is not allowed by the system. So deferring the copy to the work thread will make it generate a new instance for the thread, and the path will not be shared with main thread any more.

### Motivation

RUM-9527

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
@@ -137,21 +137,24 @@ internal class ResourceResolver(
             return
         }
 
-        val copiedDrawable = drawableCopier.copy(originalDrawable, resources)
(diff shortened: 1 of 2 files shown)


0xnm:
Won't moving the copy to the worker thread have the following impact, though: by the time we make a copy the content of Drawable will be different from what user saw at the moment we have to record the resource?

aleksandr-gringauz (dismissed):

0xnm (approved):

PR #2731, opened 2025-06-16
RUM-10224: Introduce check release pipeline

aleksandr-gringauz:
### What does this PR do?

### Motivation

### Additional Notes


ci/pipelines/common-snippets.yml
```diff
@@ -0,0 +1,5 @@
+.common-snippets:
```
aleksandr-gringauz, line 1: Extracted this to a separate file so that it can be reused both in the `default-pipeline.yml` and `check-release-pipeline.yml`.

ci/pipelines/default-pipeline.yml
```diff
@@ -0,0 +1,1030 @@
+include:
```
aleksandr-gringauz, line 1: there are no changes here except moving `set-github-installation-token` into `common-snippets` and fixing the paths to moved scripts.

ambushwork: Suggestion: next time you can separate modification and moving codes into two commits, it will be easier for reviewing the diff part.

ci/scripts/check_latest_release_is_published.sh
```diff
@@ -0,0 +1,48 @@
...
+  "dd-sdk-android-rum"
+  "dd-sdk-android-logs"
+  "dd-sdk-android-trace"
+  "dd-sdk-android-trace-otel"
+  "dd-sdk-android-webview"
+  "dd-sdk-android-session-replay"
+  "dd-sdk-android-session-replay-compose"
+  "dd-sdk-android-session-replay-material"
+  "dd-sdk-android-ndk"
+  "dd-sdk-android-coil"
```
ambushwork: missing `dd-sdk-android-fresco` and `benchmark` here?

ci/pipelines/default-pipeline.yml
```diff
@@ -0,0 +1,1030 @@
...
+notify:merge-verification-metadata:
+  tags: [ "arch:amd64" ]
+  only:
+    - tags
+  image: $CI_IMAGE_DOCKER
+  stage: notify
+  when: on_success
+  dependencies:
+    - publish:release-core
+    - publish:release-internal
```
ambushwork, line 1000: I think we stop publishing `internal` module?

aleksandr-gringauz: I am not sure about that. Modules that are published depend on `dd-sdk-android-internal`. How are they supposed to use it if it isn't published?

ci/scripts/check_latest_release_is_published.sh
aleksandr-gringauz: thanks, I'll add `fresco`!

for benchmark I think we'll have to add it after [these](https://github.com/DataDog/dd-sdk-android/commit/ccd79322895a6ba135e2b73b32005fb4aeb5c31c) changes are released

ci/pipelines/default-pipeline.yml
ambushwork: The other modules depends on it by using `implementation(project(":dd-sdk-android-internal"))`, so they still work even if we don't publish `internal` right?

I just don't remember if we made the decision of stopping publish `internal`

aleksandr-gringauz: I don't think so.

https://central.sonatype.com/artifact/com.datadoghq/dd-sdk-android-rum/dependencies - here you can see that `dd-sdk-android-rum` depends on `dd-sdk-android-internal`.

The code of `dd-sdk-android-internal` should somehow reach the user, it is not compiled into other modules, it is published as a separate artifact and other libraries depend on it.

ambushwork: You are right, artifacts don't include the transitive dependencies unless we play some trick on it, but we are fine with that!

ambushwork (dismissed):
Well done

ci/pipelines/default-pipeline.yml
```diff
@@ -0,0 +1,1030 @@
...
+stages:
+  - ci-image
+  - security
+  - analysis
+  - test # TODO RUM-1622 cleanup eventually
+  - test-pyramid
+  - publish
+  - notify
+
+.snippets:
```
0xnm, line 16: minor: having `.snippets` here and also a dedicated `common-snippets` file can create some confusion, because it is better to have a single place for them.

Maybe get rid of `common-snippets` for now and create it and move everything there in some other commit?

upd: ah, I see, you need `common-snippets` to reference it from `check-release-pipeline`.

ci/pipelines/check-release-pipeline.yml
```diff
@@ -0,0 +1,24 @@
+include:
+  - "ci/pipelines/common-snippets.yml"
+  - 'https://gitlab-templates.ddbuild.io/slack-notifier/v1/template.yml'
+
+stages:
+  - check-release
+  - notify
+
+check-release:is-published:
```
0xnm, line 9: what is the schedule to run this?

ci/scripts/check_latest_release_is_published.sh
```diff
@@ -0,0 +1,49 @@
...
+# Copyright 2016-Present Datadog, Inc.
+#
+
+set -o pipefail
+
+tag_name=$(curl -s -H "Authorization: Bearer $GITHUB_TOKEN" \
+     -H "Accept: application/vnd.github+json" \
+     https://api.github.com/repos/DataDog/dd-sdk-android/releases/latest | jq -r .tag_name)
+
+repos_list=(
```
0xnm: suggestion: maybe we can think of building it dynamically. Essentially, all folders starting with `dd-sdk-android-` represent published modules, at least it is true as of now.

0xnm (approved):

ci/pipelines/default-pipeline.yml
aleksandr-gringauz: > ah, I see, you need common-snippets to reference it from check-release-pipeline

yes, I created `common-snippets` so that I could use `set-github-installation-token` from both `default-pipeline` and `check-release-pipeline`

ci/pipelines/check-release-pipeline.yml
aleksandr-gringauz: I plan to do the same as the [synthetic test](https://mobile-integration.datadoghq.com/synthetics/edit/hrg-bj3-s2k?from_details=true) - every 12 hours.

ci/scripts/check_latest_release_is_published.sh
aleksandr-gringauz: done

created `listAllPublishedArtifactIds` gradle task and use its output in `check_latest_release_is_published.sh`.

0xnm: thanks! having a single source of truth for the list of published artifacts is a nice thing 👍

build.gradle.kts
```diff
@@ -224,3 +224,21 @@ tasks.register("printSdkDebugRuntimeClasspath") {
         File("sdk_classpath").writeText(result.joinToString(File.pathSeparator) { it.absolutePath })
     }
 }
+
+tasks.register("listAllPublishedArtifactIds") {
+    doLast {
+        val artifactIds = rootProject.subprojects.flatMap { subproject ->
+            val publishing = subproject.extensions.findByName("publishing") as? PublishingExtension
```
0xnm: minor: `subproject.extensions.findByType<PublishingExtension>()` can be probably simpler

aleksandr-gringauz: done

PR #2732, opened 2025-06-16
RUM-10346: Avoid copying hardware bitmap in Session Replay

ambushwork:
### What does this PR do?

In Session Replay especially for Jetpack Compose, we are copying the original bitmap before compressing and transforming it into byte data on work thread in case that the original one is recycled. 

But copying [hardware bitmap](https://bumptech.github.io/glide/doc/hardwarebitmaps.html) can be very slow especially we are doing it on main thread. With some manual test we noticed that the average copy time of software bitmap is around 30\~60us, and for hardware bitmap it can be up to 10\~20ms. And it throws strict mode warning if client enables it in their app.

So in this PR we decide to skip hardware bitmap copying, passing the original instance until the byte data processing, if it is recycled during the time, then we skip it by reporting it to telemetry.

### Motivation

RUM-10346:

### DIFF
--- a/features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/SemanticsUtils.kt
+++ b/features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/SemanticsUtils.kt
@@ -7,6 +7,7 @@
 package com.datadog.android.sessionreplay.compose.internal.utils
 
 import android.graphics.Bitmap
+import android.os.Build
 import android.view.View
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.ui.Modifier
@@ -28,6 +29,7 @@ import androidx.compose.ui.unit.Density
 import com.datadog.android.Datadog
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.core.sampling.RateBasedSampler
 import com.datadog.android.sessionreplay.ImagePrivacy
 import com.datadog.android.sessionreplay.TextAndInputPrivacy
 import com.datadog.android.sessionreplay.TouchPrivacy
@@ -42,7 +44,10 @@ import com.datadog.android.sessionreplay.compose.internal.mappers.semantics.Text
 import com.datadog.android.sessionreplay.utils.GlobalBounds
 
 @Suppress("TooManyFunctions")
-internal class SemanticsUtils(private val reflectionUtils: ReflectionUtils = ReflectionUtils()) {
+internal class SemanticsUtils(
+    private val reflectionUtils: ReflectionUtils = ReflectionUtils(),
+    private val sampler: RateBasedSampler<Unit> = RateBasedSampler(BITMAP_TELEMETRY_SAMPLE_RATE)
+) {
 
     internal fun findRootSemanticsNode(view: View): SemanticsNode? {
         reflectionUtils.apply {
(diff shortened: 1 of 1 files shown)


features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/SemanticsUtils.kt
```diff
@@ -273,6 +274,10 @@ internal class SemanticsUtils(private val reflectionUtils: ReflectionUtils = Ref
             }
         }
 
+        // Avoid copying hardware bitmap because it is slow and may violate [StrictMode#noteSlowCall]
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && bitmap?.config == Bitmap.Config.HARDWARE) {
+            return BitmapInfo(bitmap, isContextualImage)
```
0xnm, line 288: do we know if on the latest devices/API versions hardware bitmaps is majority or minority?

aleksandr-gringauz:
Didn't we stop copying bitmap in main thread in [this](https://github.com/DataDog/dd-sdk-android/pull/2723/files) pr?

ambushwork:
> Didn't we stop copying bitmap in main thread in [this](https://github.com/DataDog/dd-sdk-android/pull/2723/files) pr?

That PR is the fix for copying the drawable, it only applies to Android classic view since we can't access to drawable directly in Jetpack Compose, This PR fixes the bitmap copy issue when we retrieve the bitmap by using reflection in Jetpack Compose.

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/SemanticsUtils.kt
ambushwork: It depends on how Android decodes the image, if we use `Icon`,`Image` they are likely to use hardware bitmap for better performance while `Coil` [claim they disable the hardware bitmaps by default](https://github.com/coil-kt/coil/issues/1560#issuecomment-1456889783).

0xnm: I guess it is minority then. Maybe it is worth to add some metric checking how many bitmaps are Hardware ones? Not sure if it is actionable though, since probably there is no other way to get pixels fast into the app memory.

0xnm (approved):

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/SemanticsUtils.kt
ambushwork: Good idea, besides collecting the bitmap config, I also added some additional information like contextual image, bitmap size etc, it may help debugging the issue in the future.

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/SemanticsUtils.kt
```diff
@@ -282,6 +293,25 @@ internal class SemanticsUtils(private val reflectionUtils: ReflectionUtils = Ref
         }
     }
 
+    private fun sendBitmapInfoTelemetry(bitmap: Bitmap, isContextual: Boolean) {
+        val sampler: RateBasedSampler<Unit> = RateBasedSampler(BITMAP_TELEMETRY_SAMPLE_RATE)
```
0xnm: we can avoid creating a new instance at each method call and just have this either a class property, or even put in the `companion` object.

ambushwork: Moved to the constructor

PR #2736, opened 2025-06-17
RUM-10283: Make node wireframes immutable before flattening it

ambushwork:
### What does this PR do?

When iterating the wireframes with `node.wireframes.map`, it can throw ConcurrentModificationException because some wireframes are added as a mutable list, to prevent this, we need to make a copy before iterating it.

### Motivation

RUM-10283

### Additional Notes


0xnm (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/NodeFlattener.kt
```diff
@@ -19,7 +19,7 @@ internal class NodeFlattener(private val wireframeUtils: WireframeUtils = Wirefr
         stack.push(root)
         while (stack.isNotEmpty()) {
             val node = stack.pop()
-            node.wireframes
+            node.wireframes.toList()
```
aleksandr-gringauz, line 22: Does this actually fix the problem?

`toList` is still doing iteraction over a collection. If `wireframes` is a `MutableList` and during this iteration some  other thread changes it, then `ConcurrentModificationException` can be thrown.

ambushwork: Yes it does, `toList` creates a new list instance of original one, so when you modifying the original list, the new list instance will not be affected anymore during the iteration.

This can be proved by the unit test I added, when removing `toList` the test fails, when adding it the test passes

aleksandr-gringauz: > creates a new list instance of original one

it does it doing an iteraction over an original list. It all comes down to [this](https://github.com/openjdk/jdk/blob/57266064a7bb995c3c614c19b936687af35e9aa4/src/java.base/share/classes/java/util/ArrayList.java#L181) (copying ArrayList using a constructor)

if during this copying another thread modifies the original list, the copyied list will be in invalid state or copying will crash

0xnm: Most probably `toList` doesn't create iterator, but is using `System.arrayCopy` under the hood, which is `native`, so behavior may be different in case of concurrent modification.

Indeed, using the test provided outcome is different with and without `toList` call.

aleksandr-gringauz: yes, [ConcurrentModificationException](https://docs.oracle.com/javase/8/docs/api/java/util/ConcurrentModificationException.html) is thrown when the code of the collection itself is able to detect that something is wrong (concurrent modification occured).

The fact that it isn't thrown doesn't mean that the code is thread-safe.

`System.arrayCopy` as it is native most likely can't throw this exception. But it is doing the copy of the array without any locking. So if array is concurrently modified, the result of the copy could be bad.

aleksandr-gringauz: Looking at the [code](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SnapshotProducer.kt#L95) I am actually not sure that the problem is with this `wireframes` field. It might be some other list.

**upd** No, I looked at the wrong place. It is most likely `wireframes` field.

0xnm: Valid point about detection of concurrent modification when things are done in JVM vs native layer 👍

ambushwork: The overall crash rate  of ConcurrentModificationException prior to this PR is already rare, around [2 times in the past 2 months](https://dd-rum-telemetry.datadoghq.com/error-tracking/issue/4e977544-2521-11f0-a618-da7ad0900002?query=source%3Aandroid%20service%3Add-sdk-android%20%40has_build_id%3Atrue%20%40type%3Asdk_crash%20-%40error.kind%3Ajava.lang.OutOfMemoryError%20-%40error.first_datadog_frame.symbol%3A%28com.datadog.android.rum.internal.instrumentation.gestures.WindowCallbackWrapper.dispatchTouchEvent%20OR%20com.datadog.android.rum.internal.instrumentation.gestures.WindowCallbackWrapper.onMenuItemSelected%20OR%20com.datadog.android.rum.internal.instrumentation.gestures.WindowCallbackWrapper.dispatchKeyEvent%20OR%20com.datadog.android.rum.internal.instrumentation.gestures.WindowCallbackWrapper.onCreatePanelMenu%20OR%20com.datadog.android.compose.internal.TapActionTracker.invoke%29%20-version%3A1.%2A&index=&refresh_mode=sliding&from_ts=1744976312209&to_ts=1750246712209&live=true), based on that the chance of getting this exception during `toList` iteration is negligible I think.

The most proper way is to use lock or `CopyOnWriteArrayList` if the list is too large to iterate, but I think it's overkilled for our case and brings performance penatly.

aleksandr-gringauz: Did you actually find the place where this list is concurrently modified at the same time as `NodeFlattener.flattenNode` is called? I looked a bit at the code and it seems fine, I don't see the problem.

ambushwork: Yes, the problem comes from [here](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/TreeViewTraversal.kt#L100).
Each WireframeMappers have different ways of returning the list of wireframes, some use `listOf(wf1,wf2,wf3)`, some use `mutableList` to add all the wireframes then directly return. 

We are not able to reproduce exactly which mapper causes the problem, so not sure about the reason why the original list is modified after returning it.

aleksandr-gringauz: If you create a `MutableList`, populate it with values and then never modify it again, then at some point pass it to another thread using `executorService.execute`, it should be [fine](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#:~:text=Actions%20in%20a%20thread%20prior%20to%20the%20submission%20of%20a%20Runnable%20or%20Callable%20task%20to%20an%20ExecutorService%20happen%2Dbefore%20any%20actions%20taken%20by%20that%20task).

So the only possibility as far as I understand is that the MutableList is modified after it is passed to another thread.

Does it work in the way that we create these wireframes on UI thread and then pass them to the worker thread and they are only read from there? If so, it's strange that we have such problem.

ambushwork: theoretically session replay wireframes generation works like you described, it should never be modified afterwards. But like I said we currently not able to locate the place where it is modified, only thing we can do is to prevent exception from modification.

mariusc83: I think the problem lies here: https://github.com/DataDog/dd-sdk-android/blob/edfe513aa7f67304af4838fdf0143a0c5df37112/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/DefaultInteropViewCallback.kt#L24 and is because you are using this callback on the UIThread. So in this moment you can have a WorkerThread iterating your `wireframes` from one side and this callback iterating the same collection on the `UIThread`. None of this logic is doing any alteration though on the collection so not 100% sure how the `ConcurrentModificationException` might be triggered but I am pretty sure is somewhere around this logic.

mariusc83: Also @ambushwork even if you're doing a copy there you need to synchronise the code block where you are actually performing the copy at it might happen to have a concurrent modification in the middle of the copy operation.

ambushwork: As I explained to Sasha, adding `synchronise` block can bring performance punishment, and since the crash rate of this `ConcurrentModificationException` is already very low before this fix, the chance of triggering `ConcurrentModificationException` when copying a small size of list can be negligible, I think it's a good trade-off

PR #2745, opened 2025-06-19
RUM-9851: Monitor backpressure of context executor

0xnm:
### What does this PR do?

We need to monitor if the queue behind context executor is becoming big, so this PR modifies `BackPressuredBlockingQueue` a bit to:

* add constructor to create unbounded queue
* notify only if a given threshold is reached, separating this value from the `capacity` value which is basically `Int.MAX_VALUE` for unbounded queue.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
@@ -12,13 +12,55 @@ import com.datadog.android.core.configuration.BackPressureStrategy
 import com.datadog.android.internal.thread.NamedExecutionUnit
 import java.util.concurrent.TimeUnit
 
-internal class BackPressuredBlockingQueue<E : Any>(
-    private val logger: InternalLogger,
-    private val executorContext: String,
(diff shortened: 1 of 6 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BackPressuredBlockingQueue.kt
```diff
@@ -33,52 +75,59 @@ internal class BackPressuredBlockingQueue<E : Any>(
...
-            if (remainingCapacity() == 0) {
-                onThresholdReached()
+            if (size == notifyThreshold) {
+                notifyThresholdReached()
             }
             return true
         }
     }
 
+    override fun put(e: E) {
```
aleksandr-gringauz, line 85: Did you just decide to improve or `ExecutorService` started using `put` when the queue's size became `Int.MAX_VALUE`?

0xnm: It is just to have a notification fired when `notifyThreshold` is reached, which may be different from the capacity.

We are not calling this method in the production code, but it is called in the tests and to have a consistent behavior I did this update.

aleksandr-gringauz (approved):

PR #2763, opened 2025-06-25
RUM-10143: Stop posting recorded data item from main thread

ambushwork:
### What does this PR do?

It turns out we are posting the task from main thread for nothing, and it causes item dropped if the client's UI thread is busy, see in the ticket [RUM-10143].

### Motivation

RUM-10143


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/async/RecordedDataQueueRefs.kt
```diff
@@ -28,9 +24,6 @@ internal data class RecordedDataQueueRefs(
     }
 
     internal fun tryToConsumeItem() {
-        mainThreadHandler.post {
```
0xnm, line 31: can you please elaborate a bit more here: are we already on the main thread or we are on the worker thread and posting to the main thread is useless?

if it is a former, then maybe we need to add a `MAINTAINER` log entry saying when we are not on the main thread?

ambushwork: `tryToConsumeItem` is called from different places, it can be on main thread or also on work thread, we found it's useless to switch to main thread, because `tryToConsumeItems` function just check if the queue is empty then execute it with `executorService` anyway.

0xnm (approved):

jonathanmos (approved):

PR #2766, opened 2025-06-27
RUM-10575: Bump Kotlin version used to 2.0.21

0xnm:
### What does this PR do?

This PR bumps Kotlin version used to 2.0.21 and also updates Dokka version used.

Explicit reference to the Compose Compiler extension is also removed, because it is now managed by Compose Plugin.

### DIFF
--- a/sample/kotlin/build.gradle.kts
+++ b/sample/kotlin/build.gradle.kts
@@ -18,9 +18,10 @@ import com.datadog.gradle.plugin.InstrumentationMode
 plugins {
     id("com.android.application")
     kotlin("android")
+    alias(libs.plugins.composeCompilerPlugin)
     kotlin("kapt")
     id("com.github.ben-manes.versions")
-    id("org.jetbrains.dokka")
+    id("org.jetbrains.dokka-javadoc")
     id("com.squareup.sqldelight")
     id("com.google.devtools.ksp")
     alias(libs.plugins.datadogGradlePlugin)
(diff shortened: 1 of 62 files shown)


ambushwork (approved):

sample/kotlin/build.gradle.kts
```diff
@@ -65,14 +66,6 @@ android {
         java17()
     }
 
-    buildFeatures {
-        compose = true
-    }
```
aleksandr-gringauz: Are you sure that it is ok to remove this?

https://developer.android.com/develop/ui/compose/setup#setup-compose

does compose work in the sample without this buildFeature enabled?

0xnm: The trick here is that AGP doesn't look at this flag value if Compose Compiler Plugin is applied https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/scope/BuildFeatureValuesImpl.kt;l=58-60, but let's put it back to avoid confusion, because migration guide doesn't mention removal of this flag (even though it is not used).

I've updated PR.

aleksandr-gringauz: thanks for the explanation

I am ok with the previous solution as well, just wanted to double check that not too much was removed

not removing

```
buildFeatures {
    compose = true
}
```

is ok

lgtm

aleksandr-gringauz (approved):

PR #2772, opened 2025-07-01
RUM-8875 Add `updateExternalRefreshRate` to internal RUM API

awforsythe:
refs: RUM-8875

### What does this PR do?

This PR adds a new `updateExternalRefreshRate` function to the internal `AdvancedRumMonitor` interface, then updates `RumViewScope` so that it will populate `refresh_rate_average` and related properties based on the data received via this API function, if any such calls have been made.

If no calls have been made to `updateExternalRefreshRate` in the context of the current view, `RumViewScope` will continue to use the data collected from `frameRateVitalListener` as before.

Regardless of whether any calls are made to `updateExternalRefreshRate`, the vitals monitoring system within `com.datadog.android.rum.internal.vitals` will continue operating exactly as it does now. We make no attempt to disable or otherwise interfere with the `frameRateVitalListener`.

### Motivation

Per [RUM-8875](https://datadoghq.atlassian.net/browse/RUM-8875): `FPSVitalListener` uses `JankStats` to sample framerate metrics, but this approach does not work with Unity on Android, as Unity manages its rendering pipeline internally in a way that doesn't expose frame timing information to `JankStats`. As a result, Unity apps simply do not receive useful data for `refresh_rate_average` et al. in RUM View events.

Providing a new `updateExternalRefreshRate` will allow the Unity SDK to periodically self-report frame timing data collected from the Unity runtime, ensuring that the Android SDK can populate refresh-rate-related View properties with useful data.

### Additional Notes

I'm new to this codebase and I'm not terribly familiar with Kotlin, so please feel free to treat these changes with extra scrutiny.

#### Summary of code changes

- **`AdvancedRumMonitor.kt`**: Add `fun updateExternalRefreshRate(frameTimeSeconds: Double)` to `AdvancedRumMonitor` interface
- **`apiSurface`**: Regenerated
- **`dd-sdk-android-rum.api`**: Regenerated
- **`_RumInternalProxy.kt`**: Proxy calls to `AdvancedRum` impl
- **`RumRawEvent.kt`**: Define `UpdateExternalRefreshRate` event
- **`RumViewManagerScope.kt`**: Register that event in `silentOrphanEventTypes`, so that calls to `updateExternalRefreshRate` made when no view is active will be dropped without error
- **`RumViewScope.kt`**:
    - Add `externalRefreshRateInfo` as a nullable `VitalInfo` property
    - Handle `UpdateExternalRefreshRate` events by calling `onUpdateExternalRefreshRate`
    - In `onUpdateExternalRefreshRate`, set or update `externalRefreshRateInfo` property based on supplied value, guarding against divide-by-zero
    - When constructing view events, prefer `externalRefreshRateInfo` for refresh rate metrics if set; use `lastFrameRateInfo` otherwise
- **`DatadogRumMonitor.kt`**:
    - Handle `updateExternalRefreshRate` calls by constructing an `UpdateExternalRefreshRate` event and passing it to `handleEvent`

#### Summary of tests added

- **`RumInternalProxyTest.kt`**: Add a test to exercise proxying of internal API calls for new function
- **`RumRawEventExt.kt`**: Ensure that new event type is included in `Forge.anyRumEvent`
- **`RumViewScopeTest.kt`**: Add test coverage for changes to `RumViewScope`, asserting that:
    - if an external refresh rate sample is supplied, that value is written to subsequent view events
    - when multiple samples are supplied, they are averaged as expected
    - if a refresh rate of zero seconds is supplied, it is ignored and no divide by zero occurs
    - if a negative refresh rate is supplied, it is ignored
    - if refresh rate samples have been received from both `updateExternalRefreshRate` _and_ the internal `JankStats` listener, the external data will be preferred
    - if only internal `JankStats` data is present, it will be used as normal
    - if the view is stopped, external refresh rate events will be ignored
    - both average and min are computed precisely and accurately from external samples

[RUM-8875]: https://datadoghq.atlassian.net/browse/RUM-8875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
@@ -631,6 +631,10 @@ internal class DatadogRumMonitor(
         handleEvent(RumRawEvent.UpdatePerformanceMetric(metric, value))
     }
 
+    override fun updateExternalRefreshRate(frameTimeNanos: Long) {
+        handleEvent(RumRawEvent.UpdateExternalRefreshRate(frameTimeNanos))
+    }
+
     override fun setInternalViewAttribute(key: String, value: Any?) {
         handleEvent(RumRawEvent.SetInternalViewAttribute(key, value))
     }
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
@@ -631,8 +631,8 @@ internal class DatadogRumMonitor(
         handleEvent(RumRawEvent.UpdatePerformanceMetric(metric, value))
     }
 
-    override fun updateExternalRefreshRate(frameTimeNanos: Long) {
-        handleEvent(RumRawEvent.UpdateExternalRefreshRate(frameTimeNanos))
+    override fun updateExternalRefreshRate(frameTimeSeconds: Double) {
+        handleEvent(RumRawEvent.UpdateExternalRefreshRate(frameTimeSeconds))
     }
 
     override fun setInternalViewAttribute(key: String, value: Any?) {
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
@@ -7863,6 +7863,248 @@ internal class RumViewScopeTest {
 
     // endregion
 
+    // region External Refresh Rate
+
+    @Test
+    fun `M send update W handleEvent(UpdateExternalRefreshRate+KeepAlive) { single value }`(
+        forge: Forge
+    ) {
+        // GIVEN
+        val frameTimeNanos = forge.aLong(min = 1_000_000L, max = 50_000_000L) // 1ms to 50ms
+        val expectedRefreshRate = 1_000_000_000.0 / frameTimeNanos.toDouble()
+
+        // WHEN
+        testedScope.handleEvent(
+            RumRawEvent.UpdateExternalRefreshRate(frameTimeNanos),
+            mockWriter
+        )
+        val result = testedScope.handleEvent(
+            RumRawEvent.KeepAlive(),
+            mockWriter
+        )
+
+        // THEN
+        argumentCaptor<ViewEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture(), eq(EventType.DEFAULT))
+            assertThat(lastValue)
+                .hasRefreshRateMetric(expectedRefreshRate, expectedRefreshRate)
+        }
+        verifyNoMoreInteractions(mockWriter)
+        assertThat(result).isSameAs(testedScope)
+    }
+
+    @Test
+    fun `M send update W handleEvent(UpdateExternalRefreshRate+KeepAlive) { multiple values }`(
+        forge: Forge
+    ) {
+        // GIVEN
+        val frameTimesNanos = forge.aList(size = 5) {
+            aLong(min = 8_000_000L, max = 20_000_000L) // ~50-125 FPS range
+        }
+
+        var sum = 0.0
+        var min = Double.MAX_VALUE
+        var max = -Double.MAX_VALUE
+        val refreshRates = mutableListOf<Double>()
+
+        // WHEN
+        frameTimesNanos.forEach { frameTime ->
+            val refreshRate = 1_000_000_000.0 / frameTime.toDouble()
+            refreshRates.add(refreshRate)
+            sum += refreshRate
+            min = kotlin.math.min(min, refreshRate)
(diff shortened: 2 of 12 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -631,6 +631,10 @@ internal class DatadogRumMonitor(
         handleEvent(RumRawEvent.UpdatePerformanceMetric(metric, value))
     }
 
+    override fun updateExternalRefreshRate(frameTimeSeconds: Double) {
```
0xnm, line 634: how often this method will be called? is it on the scale of every few milliseconds?

awforsythe: No, I think it would be unwise of us to call it every frame. My initial plan was that the Unity SDK would respect the `VitalsUpdateFrequency` setting that (presumably) configures the internal vital monitors, so depending on that setting, it would be called roughly every 100ms, every 500ms, every 1000ms, or never.

(It may also be called immediately before `stopView`, in the event that we're stopping a view in which we've never called `updateExternalRefreshRate`.)

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7863,6 +7863,248 @@ internal class RumViewScopeTest {
...
+        val result = testedScope.handleEvent(
+            RumRawEvent.KeepAlive(),
+            mockWriter
+        )
+
+        // THEN
+        argumentCaptor<ViewEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture(), eq(EventType.DEFAULT))
+            assertThat(lastValue)
+                .hasRefreshRateMetric(expectedRefreshRate, expectedRefreshRate)
```
0xnm: minor: maybe it is better use named arguments here to avoid questions why both arguments are the same

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7863,6 +7863,248 @@ internal class RumViewScopeTest {
...
+        var max = -Double.MAX_VALUE
+        val refreshRates = mutableListOf<Double>()
+
+        // WHEN
+        frameTimesSeconds.forEach { frameTime ->
+            val refreshRate = 1.0 / frameTime
+            refreshRates.add(refreshRate)
+            sum += refreshRate
+            min = kotlin.math.min(min, refreshRate)
+            max = kotlin.math.max(max, refreshRate)
```
0xnm:
```suggestion
            min = min(min, refreshRate)
            max = max(max, refreshRate)
```

and the add the imports for `kotlin.math.min` and `kotlin.math.max`.

Alternatively `coerceAtMost` / `coerceAtLeast` methods can be used.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7863,6 +7863,248 @@ internal class RumViewScopeTest {
...
+
+        // AND - Add internal refresh rate data (should be ignored)
+        vitalListener.onVitalUpdate(VitalInfo(1, internalRefreshRate, internalRefreshRate, internalRefreshRate))
+
+        val result = testedScope.handleEvent(
+            RumRawEvent.KeepAlive(),
+            mockWriter
+        )
+
+        // THEN - Should use external data, not internal
```
0xnm: minor: we can probably remove comments from `When/And/Then` as it is pretty much clear from the code itself.

Same for below.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -7863,6 +7863,248 @@ internal class RumViewScopeTest {
...
+
+    @Test
+    fun `M prioritize external data W handleEvent(UpdateExternalRefreshRate+VitalUpdate+KeepAlive)`(
+        forge: Forge
+    ) {
+        // GIVEN
+        val externalFrameTime = forge.aDouble(min = 0.016, max = 0.017) // ~60 FPS
+        val expectedExternalRefreshRate = 1.0 / externalFrameTime
+
+        val internalRefreshRate = forge.aDouble(min = 30.0, max = 45.0) // Different range
```
0xnm: probably we can put some very wide range here since it is about priority? so range doesn't probably matter. Otherwise `min = 0.016, max = 0.017` looks a bit questionable.

Same can be applied for the tests below.

0xnm (dismissed):
nice!

awforsythe:
Thanks for the review, @0xnm! I tweaked the new unit tests as you recommended.

jonathanmos (approved):

ambushwork (approved):

PR #2776, opened 2025-07-03
Introduce _RumInternalProxy.setRumSessionTypeOverride

aleksandr-gringauz:
### What does this PR do?

We want to be able to force `sessionType=user` for sessions that happen in synthetic tests.

Introducing internal api `_RumInternalProxy.setRumSessionType` that sets `sessionType` that will be propagated to all RUM events regardless of whether we have a synthetic test or not.

The pr is big, but most of the changes are related to using named arguments in various places. For example [this](https://github.com/DataDog/dd-sdk-android/blob/3b18911419de9351ec82ddbfd82a72a530ed7e19/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt#L31) constructor has a lot of arguments of primitive types (`bool`, `long`), one can easily make mistake in order of the items. Also some of them have default values and if I create a new argument in this constructor and place it in the end, I have to provide the values for previous arguments manually.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
@@ -37,6 +37,7 @@ import com.datadog.android.internal.telemetry.InternalTelemetryEvent
 import com.datadog.android.rum.GlobalRumMonitor
 import com.datadog.android.rum.RumErrorSource
 import com.datadog.android.rum.RumSessionListener
+import com.datadog.android.rum.RumSessionType
 import com.datadog.android.rum.configuration.SlowFramesConfiguration
 import com.datadog.android.rum.configuration.VitalsUpdateFrequency
 import com.datadog.android.rum.internal.anr.ANRDetectorRunnable
@@ -628,7 +629,8 @@ internal class RumFeature(
         val slowFramesConfiguration: SlowFramesConfiguration?,
         val composeActionTrackingStrategy: ActionTrackingStrategy,
         val additionalConfig: Map<String, Any>,
-        val trackAnonymousUser: Boolean
+        val trackAnonymousUser: Boolean,
+        val rumSessionTypeOverride: RumSessionType?
     )
 
     internal companion object {
@@ -677,7 +679,8 @@ internal class RumFeature(
             composeActionTrackingStrategy = NoOpActionTrackingStrategy(),
             additionalConfig = emptyMap(),
             trackAnonymousUser = true,
-            slowFramesConfiguration = null
+            slowFramesConfiguration = null,
+            rumSessionTypeOverride = null
         )
 
         internal const val EVENT_MESSAGE_PROPERTY = "message"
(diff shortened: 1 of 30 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScope.kt
```diff
@@ -230,11 +233,13 @@ internal class RumActionScope(
                 resultId = rumContext.syntheticsResultId
             )
         }
-        val sessionType = if (syntheticsAttribute == null) {
-            ActionEvent.ActionEventSessionType.USER
-        } else {
-            ActionEvent.ActionEventSessionType.SYNTHETICS
+
+        val sessionType = when {
```
aleksandr-gringauz, line 237: If we have `rumSessionType` use it. If we are inside a synthetic test, use `sessionType=SYNTHETICS`, otherwise `sessionType=USER`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -628,7 +629,8 @@ internal class RumFeature(
         val slowFramesConfiguration: SlowFramesConfiguration?,
         val composeActionTrackingStrategy: ActionTrackingStrategy,
         val additionalConfig: Map<String, Any>,
-        val trackAnonymousUser: Boolean
+        val trackAnonymousUser: Boolean,
+        val rumSessionType: RumSessionType?
```
0xnm: I'm wondering if we should call this something like `rumSessionTypeOverride`, because we will have set it up from the intent attributes and this value is just what overrides the normal flow. Or maybe add some comment.

Although in the symmetric PR for iOS SDK suffix `Override` is used only in configuration https://github.com/DataDog/dd-sdk-ios/pull/2383/files.

0xnm (dismissed):
looks good 👍 

since we have many lines changed here, it would be nice to have one more review.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
aleksandr-gringauz: > I'm wondering if we should call this something like rumSessionTypeOverride

done

> Or maybe add some comment.

I added a comment [here](https://github.com/DataDog/dd-sdk-android/pull/2776/files#diff-606fbc699313cd534ac3d7e8faf6464a4d3814843e1ead34f467fbd9e5a5ff80) 

> Although in the symmetric PR for iOS SDK suffix Override is used only in configuration

Looks like on iOS the situation is slightly different, they are able to determine final `rumSessionType` in one place and propagate the value to all rum events [link](https://github.com/DataDog/dd-sdk-ios/blob/10c8c58640b08d64bcdf2191aa74be5e43dfb134/DatadogRUM/Sources/RUMMonitor/Scopes/RUMScopeDependencies.swift#L123). So it is not "override", it is the final one.

On Android we create first RUM scopes ([RumApplicationScope](https://github.com/DataDog/dd-sdk-android/blob/edd95ecf7b743906ebbcca3932aaab363d748bd8/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumApplicationScope.kt#L28) for example) before [this](https://github.com/DataDog/dd-sdk-android/blob/edd95ecf7b743906ebbcca3932aaab363d748bd8/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityLifecycleTrackingStrategy.kt#L97) is called and account for the synthetic test attributes only when the events are created (for example like [this](https://github.com/DataDog/dd-sdk-android/blob/aleksandr-gringauz/force-user-session-type/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L509)). So for us is is kind of "override" indeed.

ambushwork (approved):
LGTM!

PR #2784, opened 2025-07-08
RUM-10703: Fix Session Replay NPE when getting TextView padding

ambushwork:
### What does this PR do?

Here is the crash stack trace about this issue:
```
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'int android.text.Layout.getHeight()' on a null object reference
  at android.widget.TextView.getBottomVerticalOffset (TextView.java:9021)
  at android.widget.TextView.getTotalPaddingBottom (TextView.java:3451)
  at com.datadog.android.sessionreplay.recorder.mapper.TextViewMapper.resolvePadding (TextViewMapper.kt:193)
  at com.datadog.android.sessionreplay.recorder.mapper.TextViewMapper.resolveTextPosition (TextViewMapper.kt:185)
  at com.datadog.android.sessionreplay.recorder.mapper.TextViewMapper.createTextWireframe (TextViewMapper.kt:142)
  at com.datadog.android.sessionreplay.recorder.mapper.TextViewMapper.map (TextViewMapper.kt:62)
  at com.datadog.android.sessionreplay.recorder.mapper.TextViewMapper.map (TextViewMapper.kt:30)
```

From the documentation of `TextView.getLayout()` we know that it's probably caused by the text update:
```
/**
 * Gets the {@link android.text.Layout} that is currently being used to display the text.
 * This value can be null if the text or width has recently changed.
 * @return The Layout that is currently being used to display the text.
 */
```

Based on that, we just need to make sure that the `layout` is present before reaching the `getTotalPaddingBottom` function.

### Motivation

RUM-10703

### Additional Notes

The whole function is called from MainThread, so it's unlikely that between checking `textView.layout` and using `textView.totalPaddingBottom.` the `layout` is set to null again due to the concurrency


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/TextViewMapper.kt
```diff
@@ -188,12 +188,21 @@ open class TextViewMapper<in T : TextView>(
...
+                top = textView.totalPaddingTop.densityNormalized(pixelsDensity).toLong(),
+                bottom = textView.totalPaddingBottom.densityNormalized(pixelsDensity).toLong(),
+                left = textView.totalPaddingStart.densityNormalized(pixelsDensity).toLong(),
+                right = textView.totalPaddingEnd.densityNormalized(pixelsDensity).toLong()
+            )
+        } else {
+            MobileSegment.Padding(
+                top = textView.paddingTop.densityNormalized(pixelsDensity).toLong(),
+                bottom = textView.paddingBottom.densityNormalized(pixelsDensity).toLong(),
+                left = textView.paddingLeft.densityNormalized(pixelsDensity).toLong(),
```
jonathanmos: should be paddingStart

aleksandr-gringauz (approved):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/recorder/mapper/TextViewMapper.kt
```diff
@@ -188,12 +188,21 @@ open class TextViewMapper<in T : TextView>(
     }
 
     private fun resolvePadding(textView: TextView, pixelsDensity: Float): MobileSegment.Padding {
-        return MobileSegment.Padding(
-            top = textView.totalPaddingTop.densityNormalized(pixelsDensity).toLong(),
-            bottom = textView.totalPaddingBottom.densityNormalized(pixelsDensity).toLong(),
-            left = textView.totalPaddingStart.densityNormalized(pixelsDensity).toLong(),
-            right = textView.totalPaddingEnd.densityNormalized(pixelsDensity).toLong()
-        )
+        return if (textView.layout != null) {
```
0xnm, line 191: are we sure it is a fix that will be enough in general when fetching properties of `TextView`, given we cannot reproduce it?

To me the issue is that `layout` is `null` is related to the fact that we either try to create wireframe too early or too late in the `TextView` lifecycle and there is no guarantee that we fix this call site, but then we can have a crash somewhere later if `layout` is still `null`.

Is there a way to check that a view is in a good and reliable state in general before we start pulling the properties for the wireframe generation?

ambushwork: The client is not able to reproduce the issue systemically neither on their side, so they can't tell which `TextView` exact is causing the issue. Even though they gave us the code snippet, we only know that they are using viewbinding, and programmatically assign the text content in their activity, otherwise nothing irregular.

According to the source code, null `layout` can also happen between `TextView` cleaning `layout` and making new `layout`, and in this case we are not able to prevent it by checking if the view is well loaded or not, the view can update at anytime. so in my point of view, checking `layout` is the best thing that we can do to make sure it.

0xnm (approved):

PR #2793, opened 2025-07-21
RUM-10316 - properties referencing support

satween:
### What does this PR do?

In order to align `os` and `device` properties across different events we need to have single source of truth for `device` and `os` objects. Current `JsonSchemaReader` supports referencing only if referenced object was declared in `definitions` object. 

This PR adds support for referencing object declared in `properties`. Note that only whole object referencing is supported because referencing by id requires name resolution, which makes referenced type a bit different every time it referenced.

see [/external_nested_description_properties.json](https://github.com/DataDog/dd-sdk-android/compare/feature/v3...tvaleev/feature/RUM-10316-properties-ref-support?expand=1#diff-9531c9c12b39b22e54422de1b17ca0bc57ac00185bbe01af4f8ebf65c34d66f4) for example of new syntax supported

### Motivation

We need this change in order to use rum/_common_schema.json as a source for `device` and `os` objects. 

### Additional Notes

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/JsonSchemaReader.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/JsonSchemaReader.kt
@@ -62,7 +62,9 @@ class JsonSchemaReader(
             return loadDefinitionFromFileRef(path, localRef, fromFile)
         }
 
-        val name = REF_DEFINITION_REGEX.matchEntire(ref)?.groupValues?.get(1)
+        val fieldGroups = REF_NAME_REGEX.matchEntire(ref)?.groupValues
+        val type = fieldGroups?.get(1)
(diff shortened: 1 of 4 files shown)


satween:
> Can you please give more details which referenced properties are needed to be supported? Indeed, `device` is [declared](https://github.com/search?q=repo%3ADataDog%2Frum-events-format+device+path%3A%2F%5Eschemas%5C%2F%2F+path%3A%2F%5Eschemas%5C%2Frum%5C%2F%2F&type=code) in the common schema, but events generated for views, actions, etc. already have `device` and `os` properties in the model.
> 
> Do you mean they need to be referenced from the Logs and Traces schema instead of direct declaration there?

Yes, thats correct, see https://github.com/DataDog/dd-sdk-android/pull/2794 for more details.

jonathanmos (approved):

mariusc83 (approved):

PR #2794, opened 2025-07-21
RUM-10316: using common-schema for common object generation

satween:
### What does this PR do?

This PR continues changes started in https://github.com/DataDog/dd-sdk-android/pull/2793

There are several changes were mage under this PR:

1. `cloneCommonSchema` task was added for `-logs` and `-trace` modules - making possible to clone most recent schema file from schema repository
2. `generateLogModelsFromJson` and `generateTraceModelFromJson` tasks were updated to ignore objects from `_common_schema.json` from  being generated directly
3. `span-schema.json` and `log-schema.json` files were updated to support property referencing added in https://github.com/DataDog/dd-sdk-android/pull/2793
4. Forgery factories, tests and mappers were updated in order to support new event structure

### Motivation
We aiming to align event schema for v3 on both platforms.
Similar changes were made for ios https://github.com/DataDog/dd-sdk-ios/pull/2322

### Additional Notes


features/dd-sdk-android-logs/clone_common_schema.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+import com.datadog.gradle.plugin.gitclone.GitCloneDependenciesTask
+
+tasks.register<GitCloneDependenciesTask>("cloneCommonSchema") {
+    extension.apply {
+        clone(
+            "https://github.com/DataDog/rum-events-format.git",
+            "schemas/rum",
```
aleksandr-gringauz, line 12: Just out of curiosity. Will it work if I write here `schemas/rum/_common-schema.json` without specifying all other files in `excludedPrefixes`?

satween: it would but it will download all those files that is unnecessary for logs schema. 
Specifying single file is not supported by clone gradle plugin at this moment

satween:
> Is my understanding correct that another solution would be just to change schemas for `-logs` and `-traces` manually?
> 
> Yes, there will not be a single source of truth but we would avoid all complicated changes here.

Well, not exactly. We need to have same schema as Ios has. So in case if we just manually adjust *.json files we still gonna have lot of changes in generated code and as well as in mapper logic. So the only difference will be actually in schema file.

aleksandr-gringauz (dismissed):

aleksandr-gringauz (approved):

jonathanmos (approved):

PR #2797, opened 2025-07-23
Collect Locale attributes

jonathanmos:
### What does this PR do?
Collects the following Locale attributes:

currentLocale - the current locale defined for the user or falls back to the device default
locales - an ordered list of locales the user has defined
timeZone - the timeZone of the device

These new attributes are sent within the view, error, action and resource scopes within the application and device categories. A known edge case at the moment is that if the user updates the values of locale or timezone while the sdk is running this change will not be reflected in what we send - the values are set at the time that the context is created and cached. This means that only a restart of the SDK would send updated values. However, users don't often change their timezone or locale settings so this may be an acceptable limitation. The alternative could be to set broadcast receivers for these changes, but that may be overkill.

### Motivation
Part of the initiative to collect additional attributes.

### Additional Notes


dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/LocaleInfo.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Provides information about locale.
+ *
+ * @property locales Ordered list of the user’s preferred system languages as IETF language tags.
+ * @property currentLocale The user's current locale as a language tag (language + region), computed from their preferences and the app's supported languages, e.g. 'es-FR'.
+ * @property timeZone The device’s current time zone identifier, e.g. 'Europe/Berlin'.
+ */
+data class LocaleInfo(
+    val locales: List<String>,
```
aleksandr-gringauz, line 17: do elements of `locales` have the same meaning as `currentLocale`? Can `es-FR` be in `locales`?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/system/DefaultAndroidInfoProvider.kt
```diff
@@ -50,6 +51,46 @@ internal class DefaultAndroidInfoProvider(
         }
     }
 
+    override val locales: List<String> by lazy(LazyThreadSafetyMode.PUBLICATION) {
```
ambushwork, line 54: Question: what will happen if "locales/currentLocale/timeZone" change during the app lifecycle and these attributes are queried again after the changing?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/LocaleInfo.kt
jonathanmos: yes, locales is an array of the currently set values in the same format. For example: 

["en-US", "he-IL"]

aleksandr-gringauz (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/system/DefaultAndroidInfoProvider.kt
jonathanmos: I've tested this manually and changing the values while the sdk is running doesn't appear to cause issues. We don't actually query the values again, as they are immutable (cached) in the context when it is created so they won't update to new values until the sdk restarts.

ambushwork (approved):

PR #2806, opened 2025-07-29
RUM-11121: Make accessibility send only mutations

jonathanmos:
### What does this PR do?
**Previous state:**
Accessibility attributes were collected and sent only when the flag `collectAccessibilitySettings` was set, this flag being an internal flag in `RumInternalProxy`. Attributes were sent as custom tags in view update events.

**New state:**
Accessibility attributes will be sent for all users who set the feature flag `collectAccessibility` (which is now public in `RumConfiguration` and has been renamed to match the ios api). The attributes will be sent as part of the view schema and not as custom tags. Additionally these attributes are `send-only-once`, meaning that accessibility attributes are only sent the first time they have a non-null value, and subsequently only if they change.

Api for accessibility:

.collectAccessibility(enabled = true/false)

### Motivation
Part of the effort to expand metadata visibility.

### Additional Notes
Contains an update to the RUM schema to support sending the RTL Enabled attribute - this attribute indicates whether right-to-left layout was set on the user's device. 

Contains a change to how view filtering works. Previously we sent only the view update with the max doc version, however this causes issues with the send-only-once mechanism. Therefore we now send if either the previous condition was true OR the update contains accessibility changes.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
@@ -60,6 +60,17 @@ data class RumConfiguration internal constructor(
         }
 
         /**
+         * Whether to collect accessibility attributes inside the RUM view update event.
+         * This is disabled by default.
+         *
+         * @param enabled whether collecting accessibility attributes is enabled or not.
+         */
+        fun collectAccessibility(enabled: Boolean): Builder {
+            rumConfig = rumConfig.copy(collectAccessibility = enabled)
+            return this
+        }
+
+        /**
          * Sets the sample rate for Internal Telemetry (info related to the work of the
          * SDK internals). Default value is 20.
          *
@@ -387,15 +398,5 @@ data class RumConfiguration internal constructor(
             rumConfig = rumConfig.copy(rumSessionTypeOverride = rumSessionTypeOverride)
             return this
         }
-
-        /**
-         * Sets a flag to collect accessibility settings inside the RUM view end event.
-         * By default these settings are not collected.
-         */
-        internal fun collectAccessibilitySettings(): Builder {
-            rumConfig = rumConfig.copy(collectAccessibilitySettings = true)
-            return this
-        }
-        // endregion
     }
 }
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/DefaultAccessibilitySnapshotManager.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/DefaultAccessibilitySnapshotManager.kt
@@ -1,0 +1,34 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.domain.accessibility
+
+internal class DefaultAccessibilitySnapshotManager(
+    private val accessibilityReader: AccessibilityReader
+) : AccessibilitySnapshotManager {
+    private val lastSnapshot = mutableMapOf<String, Any>()
+
+    @Synchronized
+    override fun latestSnapshot(): Accessibility {
+        val newAccessibilityState = accessibilityReader.getState()
+        val newSnapshot = mutableMapOf<String, Any>()
+
+        // remove the ones we saw already
+        for (key in newAccessibilityState.keys) {
+            val newValue = newAccessibilityState[key]
+                ?: continue
+
+            val oldValue = lastSnapshot[key]
+
+            if (newValue != oldValue) {
+                newSnapshot[key] = newValue
+                lastSnapshot[key] = newValue
+            }
+        }
+
+        return Accessibility.fromMap(newSnapshot)
+    }
+}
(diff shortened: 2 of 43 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/DefaultAccessibilitySnapshotManager.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package com.datadog.android.rum.internal.domain.accessibility
+
+internal class DefaultAccessibilitySnapshotManager(
+    private val accessibilityReader: AccessibilityReader
+) : AccessibilitySnapshotManager {
+    private val lastSnapshot = mutableMapOf<String, Any>()
+
+    @Synchronized
+    override fun latestSnapshot(): Accessibility {
```
ambushwork, line 15: Function logics looks clear to me, just wondering if the name of the function match with the logic. 
For me it will be more appropriate to name as:
* getChangedAccessibilityState
* diffAccessibilityState

Also I am not sure why we need a dedicated class for this, can it just be inside `AccessibilityReader`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumEventMeta.kt
```diff
@@ -62,8 +64,14 @@ internal sealed class RumEventMeta {
                     VIEW_TYPE_VALUE -> {
                         val viewId = model.get(VIEW_ID_KEY).asString
                         val docVersion = model.get(DOCUMENT_VERSION_KEY).asLong
-
-                        View(viewId, docVersion)
+                        val hasAccessibilityElement = model.get(HAS_ACCESSIBILITY_KEY)
+                        val hasAccessibility = when {
+                            hasAccessibilityElement == null -> false // Missing field (backward compatibility)
+                            hasAccessibilityElement.isJsonNull -> null // Explicit null in JSON
```
ambushwork: Question: which use case corresponds to explicit null in JSON?

ambushwork (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -990,6 +984,17 @@ internal open class RumViewScope(
...
 
+        val accessibilityState = accessibilitySnapshotManager.latestSnapshot()
+        val accessibility = ViewEvent.Accessibility(
+            textSize = accessibilityState.textSize,
+            invertColorsEnabled = accessibilityState.isColorInversionEnabled,
+            singleAppModeEnabled = accessibilityState.isScreenPinningEnabled,
+            screenReaderEnabled = accessibilityState.isScreenReaderEnabled,
+            closedCaptioningEnabled = accessibilityState.isClosedCaptioningEnabled,
+            reducedAnimationsEnabled = accessibilityState.isReducedAnimationsEnabled,
+            rtlEnabled = accessibilityState.isRtlEnabled
```
aleksandr-gringauz, line 995: Just to understand better. Suppose we send `rtlEnabled = true`. The next time view event is sent `rtlEnabled = null`. Will backend understand that `rtlEnabled` is actually still `true` and not override it to `null`?

jonathanmos: For null values we should be safe because in the ViewEvent generated code only non-null values are allowed in the json

aleksandr-gringauz (dismissed):

aleksandr-gringauz (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -59,6 +59,17 @@ data class RumConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Whether to collect accessibility attributes inside the RUM view update event.
```
0xnm, line 63: Customers unlikely know that RUM View update event is. Maybe we can simply remove `inside the RUM view update event` part?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/event/RumViewEventFilter.kt
```diff
@@ -37,10 +37,12 @@ internal class RumViewEventFilter(
...
                 @Suppress("UnsafeThirdPartyFunctionCall") // we checked the key before
                 val viewMeta = viewMetaByEvent.getValue(it)
-                // we need to leave only view event with a max doc version for a give viewId in the
-                // batch, because backend will do the same during the reduce process
+
+                // we need to leave only view events with accessibility OR view event with a max doc version
+                // for a give viewId in the batch, because backend will do the same during the reduce process
                 @Suppress("UnsafeThirdPartyFunctionCall") // if there is a meta, there is a max doc version
-                viewMeta.documentVersion == maxDocVersionByViewId.getValue(viewMeta.viewId)
+                viewMeta.hasAccessibility == true ||
```
0xnm, line 44: so if we have accessibility enabled, we simply don't have any view reducing on the client side? do I get it right?

Or not every view update has `hasAccessibility` even if accessibility collection is enabled?

features/dd-sdk-android-rum/src/testFixtures/kotlin/com/datadog/android/rum/utils/forge/AccessibilityForgeryFactory.kt
```diff
@@ -0,0 +1,25 @@
...
+internal class AccessibilityForgeryFactory : ForgeryFactory<Accessibility> {
+    override fun getForgery(forge: Forge): Accessibility {
+        return Accessibility(
+            textSize = forge.aNullable { forge.aString() },
+            rtlEnabled = forge.aNullable { forge.aBool() },
+            screenReaderEnabled = forge.aNullable { forge.aBool() },
+            increaseContrastEnabled = forge.aNullable { forge.aBool() },
+            reducedAnimationsEnabled = forge.aNullable { forge.aBool() },
+            invertColorsEnabled = forge.aNullable { forge.aBool() },
+            singleAppModeEnabled = forge.aNullable { forge.aBool() }
```
0xnm, line 22: minor, `aNullable` already has `Forge` as a receiver in the lambda, so it can be `forge.aNullable { aString }`, etc.

PR #2814, opened 2025-08-08
[RUM-11249][FO]: public API

satween:
### What does this PR do?

This PR adds public API without any implementation for the FO feature. 

### Motivation

This changes should unblock RN integration implementation

### DIFF
--- a/features/dd-sdk-android-rum/api/apiSurface
+++ b/features/dd-sdk-android-rum/api/apiSurface
@@ -124,6 +124,9 @@ interface com.datadog.android.rum.RumMonitor
   fun clearAttributes()
   fun stopSession()
   fun addViewLoadingTime(Boolean)
+  fun startFeatureOperation(String, String? = null, Map<String, Any?> = emptyMap())
+  fun succeedFeatureOperation(String, String? = null, Map<String, Any?> = emptyMap())
+  fun failFeatureOperation(String, String? = null, com.datadog.android.rum.featureoperations.FailureReason, Map<String, Any?> = emptyMap())
   var debug: Boolean
   fun _getInternal(): _RumInternalProxy?
 enum com.datadog.android.rum.RumPerformanceMetric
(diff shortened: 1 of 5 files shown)


aleksandr-gringauz (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -334,6 +335,74 @@ interface RumMonitor {
     @ExperimentalRumApi
     fun addViewLoadingTime(overwrite: Boolean)
 
+    /**
+     * Reports that the [name] feature operation has started.
```
ambushwork: Opinionated: The documentation doesn't match exactly with the function name. The function name feels like is about to start a feature operation, but actually it should be called after user has started a feature operation, so my question is that, on other platforms, is it the same name?

satween: Yeah, this is where I couldn't decide which word is better for the documentation. Changed a bit to align with the method name. Thanks for spotting that

ambushwork (dismissed):

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -124,6 +124,11 @@ interface com.datadog.android.rum.RumMonitor
   fun clearAttributes()
   fun stopSession()
   fun addViewLoadingTime(Boolean)
+  fun startFeatureOperation(String, String? = null, Map<String, Any?> = emptyMap())
+  fun succeedFeatureOperation(String, String? = null, Map<String, Any?> = emptyMap())
+  fun failFeatureOperation(String, String? = null, com.datadog.android.rum.featureoperations.FailureReason, Map<String, Any?> = emptyMap())
+  fun retryFeatureOperation(String, String? = null, Map<String, Any?> = emptyMap())
```
mariedm: We should only have 3 public APIs for now: `startFeatureOperation`, `succeedFeatureOperation`, and `failFeatureOperation`. `updateFeatureOperation` and `retryFeatureOperation` won't be supported for now.

mariedm (changes requested):
LGMT, but we must only expose 3 APIs: start, end successfully, and with failure.

features/dd-sdk-android-rum/api/apiSurface
satween: hmm. I thought "now" means at the very beginning of the development, but it would be added shortly. Thanks for spotting this out.  I will remove this then

mariedm: Hopefully next quarter we can add those

mariedm (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt
```diff
@@ -334,6 +335,49 @@ interface RumMonitor {
...
+     * provided during [startFeatureOperation] invocation.
+     * @param failureReason the reason for the operation failure.
+     * @param attributes additional custom attributes to attach to the feature operation. Can be
+     * used to add some result data produced as the result of the operation.
+     */
+    @ExperimentalRumApi
+    fun failFeatureOperation(
+        name: String,
+        operationKey: String? = null,
+        failureReason: FailureReason,
```
0xnm, line 377: FYI: on iOS this attribute is called just `reason`, not `failureReason`, so it will be nice to align

https://github.com/DataDog/dd-sdk-ios/pull/2427/files#diff-8a79dd186c7b2bd17b0d53e57ebddbfd05909f395af9a6d01e9ac6c46b9a1f22R341

PR #2815, opened 2025-08-10
RUM-11120: Add battery and display attributes

jonathanmos:
### What does this PR do?
Adds some battery and display attributes to view updates. The following fields were added:

`battery_level`: The current battery charge level, expressed as a float from 0.0 (empty) to 1.0 (full). This is to align with ios, so we lose some granularity as android can measure to within 1% of the battery level while ios shows only to the nearest 10%. Monitored by polling once every 60 seconds - it didn't seem a good idea to register a listener as the listener was firing sometimes multiple times per second.

`low_power_mode`: A boolean indicating whether the device is currently in Low Power Mode. Monitored with a broadcast receiver.

`brightness_level`: The current screen brightness, normalized as a float between 0.0 (darkest) and 1.0 (brightest). Monitored with a content observer.

### Motivation
Part of adding metadata fields.

### Additional Notes
Refactor of the architecture of the info providers since we now have three providers (accessibility/battery/display) - all classes now implement the InfoProvider interface and use the same generic NoOp class. The generic interface returns a kv map and we use the concrete data classes to convert between the objects and this representation.


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
```diff
@@ -0,0 +1,133 @@
...
+        registerReceivers()
+        buildInitialState()
+    }
+
+    @Synchronized
+    override fun getState(): Map<String, Any> {
+        // while we could register a receiver for battery level,
+        // it fires far too often (multiple times per second)
+        // so it seems better to only poll battery charge state once in a period of time
+        val now = System.currentTimeMillis()
```
aleksandr-gringauz: As far as I know, it is better to use `SystemClock.uptimeMillis() or SystemClock.elapsedRealtime` to measure time intervals (not `System.currentTimeMillis`). The reason is explained [here](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/os/SystemClock.java;l=45?q=public%20final%20class%20SystemClock).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/display/DefaultDisplayInfoProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+    }
+
+    init {
+        registerListeners()
+        buildInitialState()
+    }
+
+    @Synchronized
```
aleksandr-gringauz: very minor

Is this `@Synchronized` actually necessary? `currentState` is a volatile reference to an immutable `data class`. And there is nothing happening inside `getState` that requires a critical section.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -306,17 +319,26 @@ internal class RumFeature(
...
+
+    private fun cleanupInfoProviders() {
         if (configuration.collectAccessibility) {
             accessibilityReader.cleanup()
-            accessibilityReader = NoOpAccessibilityReader()
+            accessibilityReader = NoOpInfoProvider()
             accessibilitySnapshotManager = NoOpAccessibilitySnapshotManager()
         }
 
-        GlobalRumMonitor.unregister(sdkCore)
```
satween, line 315: This line removed on purpose? Don't see it moved somewhere else

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/BatteryInfo.kt
```diff
@@ -0,0 +1,35 @@
...
+package com.datadog.android.rum.internal.domain.battery
+
+/**
+ * Provides information about the battery state.
+ *
+ * @property batteryLevel the current battery charge level, expressed as a float from 0.0f (empty) to 1.0f (full).
+ * @property lowPowerMode a boolean indicating whether the device is currently in Low Power Mode.
+ */
+internal data class BatteryInfo(
+    val batteryLevel: Float? = null,
```
satween: nit maybe it make sense to add [@FloatRange](https://developer.android.com/reference/kotlin/androidx/annotation/FloatRange)(min, max) annotation here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
```diff
@@ -0,0 +1,133 @@
...
+        )
+    }
+
+    private fun registerReceivers() {
+        val powerSaveFilter = IntentFilter(ACTION_POWER_SAVE_MODE_CHANGED)
+        applicationContext.registerReceiver(powerSaveModeReceiver, powerSaveFilter)
+    }
+
+    private fun getLowPowerMode(): Boolean? {
+        return powerManager?.isPowerSaveMode
```
satween, line 102: Shouldn't it be just false if `powerManager` is null?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
```diff
@@ -0,0 +1,133 @@
...
+    @Volatile
+    private var currentState = BatteryInfo()
+
+    private var lastTimeBatteryLevelChecked = AtomicLong(System.currentTimeMillis())
+
+    private val powerSaveModeReceiver = object : BroadcastReceiver() {
+        override fun onReceive(context: Context, intent: Intent) {
+            val isPowerSaveMode = powerManager?.isPowerSaveMode
+            isPowerSaveMode?.let {
+                currentState = currentState.copy(
```
satween: Should this write be also synchronized? Also for init{} ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
jonathanmos: it moved to line 324

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
jonathanmos: This means that we can't get the state of power management so I'm not sure we should set a value as it could be misleading

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
aleksandr-gringauz: > Should this write be also synchronized?

+1

> Also for init{} ?

If we put `buildInitialState` before `registerReceivers` it shouldn't be needed.

jonathanmos: I'm not sure because the variable is volatile and we're only replacing the reference so there's no partial state. Let's discuss

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/InfoProvider.kt
```diff
@@ -4,12 +4,12 @@
...
 
-package com.datadog.android.rum.internal.domain.accessibility
+package com.datadog.android.rum.internal.domain
 
 import com.datadog.tools.annotation.NoOpImplementation
 
 @NoOpImplementation
-internal interface AccessibilityReader {
+internal interface InfoProvider {
     fun getState(): Map<String, Any>
```
jonathanmos: Need to change getState to use generics

aleksandr-gringauz (dismissed):

aleksandr-gringauz (approved):

satween (approved):

PR #2823, opened 2025-08-19
Some additional accessibility test coverage

jonathanmos:
### What does this PR do?
Refactor `DefaultAccessibilitySnapshotManager` to return a nullable object - this simplifies the later logic in terms of knowing if anything changed in the accessibility state. 

Adds some additional test coverage for accessibility around the logic for view reduction

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/AccessibilitySnapshotManager.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/AccessibilitySnapshotManager.kt
@@ -10,5 +10,5 @@ import com.datadog.tools.annotation.NoOpImplementation
 
 @NoOpImplementation
 internal interface AccessibilitySnapshotManager {
-    fun latestSnapshot(): AccessibilityInfo
+    fun getIfChanged(): AccessibilityInfo?
 }
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/RumDataWriterTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/RumDataWriterTest.kt
@@ -17,7 +17,6 @@ import com.datadog.android.rum.model.ErrorEvent
 import com.datadog.android.rum.model.LongTaskEvent
 import com.datadog.android.rum.model.ResourceEvent
 import com.datadog.android.rum.model.ViewEvent
-import com.datadog.android.rum.model.ViewEvent.Accessibility
 import com.datadog.android.rum.utils.config.GlobalRumMonitorTestConfiguration
 import com.datadog.android.rum.utils.forge.Configurator
 import com.datadog.tools.unit.annotations.TestConfigurationsProvider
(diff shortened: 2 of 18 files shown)


features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/RumDataWriterTest.kt
```diff
@@ -252,7 +254,63 @@ internal class RumDataWriterTest {
 
     // endregion
 
-    private fun isAccessibilityPopulated(accessibility: ViewEvent.Accessibility?): Boolean {
+    // region accessibility
+
+    @Test
+    fun `M hasAccessibility false W write() { empty accessibility object }`(
```
0xnm: I think there is a disconnect between `RumViewScope` (or `AccessibilitySnapshotManager`) and `RumDataWriter` leading to the existence of `RumDataWriter.isAccessibilityPopulated` method and following issues:

* one may update `DefaultAccessibilitySnapshotManager.latestSnapshot` method if new properties are added, but forget to update `RumDataWriter.isAccessibilityPopulated`
* the number of properties participating in `DefaultAccessibilitySnapshotManager.latestSnapshot` is not the same as in `RumDataWriter.isAccessibilityPopulated`, is it expected?
* as a consequence of above, we have some properties which are not checked for the change in `DefaultAccessibilitySnapshotManager.latestSnapshot`, meaning we will always have `hasAccessibility=true`, isn't it?

Maybe we should simply return `null` from `DefaultAccessibilitySnapshotManager.latestSnapshot` if nothing changed? And rename this method somehow.

This would allow us to have better encapsulation and smaller change surface.

ambushwork (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/AccessibilityInfo.kt
```diff
@@ -27,4 +27,32 @@ internal data class AccessibilityInfo(
...
+    override fun hashCode(): Int {
+        var result = isScreenReaderEnabled?.hashCode() ?: 0
+        result = 31 * result + (isColorInversionEnabled?.hashCode() ?: 0)
+        result = 31 * result + (isClosedCaptioningEnabled?.hashCode() ?: 0)
+        result = 31 * result + (isReducedAnimationsEnabled?.hashCode() ?: 0)
+        result = 31 * result + (isScreenPinningEnabled?.hashCode() ?: 0)
+        result = 31 * result + (isRtlEnabled?.hashCode() ?: 0)
+        result = 31 * result + (textSize?.hashCode() ?: 0)
+        return result
+    }
```
0xnm: why this change is needed? isn't it the same as `data` class generates under the hood?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/AccessibilitySnapshotManager.kt
```diff
@@ -10,5 +10,5 @@ import com.datadog.tools.annotation.NoOpImplementation
 
 @NoOpImplementation
 internal interface AccessibilitySnapshotManager {
-    fun latestSnapshot(): AccessibilityInfo
+    fun latestSnapshot(): AccessibilityInfo?
```
0xnm: by the usage at the call site maybe it is better to call this as `getIfChanged`, but not a blocker for me, just a suggestion.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/accessibility/AccessibilityInfo.kt
jonathanmos: I thought it may be necessary, but you're right that it's redundant here - removed

0xnm (approved):

ambushwork (approved):

Issue #2809, opened 2025-07-31
Improve OkHTTP Client Initialization


Laimiux:
### Feature description

This is a two-fold request:
- Use `Call.Factory` to move `OkHttpClient` initialization from main thread to the background thread.
- OkHttpClients Should Be Shared ([documentation](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/index.html)). Specifically, thread and connection pools should be shared for more efficient resource management. It would be ideal DataDog accepted `() -> OkHttpClient.Builder` as part of configuration.

Laimiux:
> Hello [@Laimiux](https://github.com/Laimiux)!
> 
> Let me clarify our decision.
> 
> * Shared OkHttpClient instances
> 
> We are not going to share `OkHttpClient` with a customer code, because we have a [specific configuration](https://github.com/DataDog/dd-sdk-android/blob/e83db329d86fbf98e0a7c3d4baf37ad64bcc4e12/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L577-L609) and we don't want to have it modified. Note, that some requirements, like the set of cipher suites are imposed by the security (e.g. to not allow cipher suite downgrade). Improper setup of `OkHttpClient` may create customer issues.
> 
> Moreover, by doing this we keep the delivery rate of the observability data to the Datadog intake high, because when instance is shared we may experience connection pool starvation, depending on the customer application nature.
> 
> Keep in mind that you have [SDK configuration](https://github.com/DataDog/dd-sdk-android/blob/e83db329d86fbf98e0a7c3d4baf37ad64bcc4e12/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/configuration/Configuration.kt#L172-L191) to control how batches are uploaded and since they are uploaded in sequential manner for the feature (RUM, Logs, Traces, etc.), each feature has no more than 1 (or 2 in case of Session Replay) connection opened at the same time.
> 
> * `OkHttpClient` initialization
> 
> If I get it right, the chart you've posted is for the Datadog SDK initialization in general in your application, with all feature included?
> 
[...]

PR #2824, opened 2025-08-19
Move Okhttp client initialization to background thread.

Laimiux:
### What does this PR do?

Moving OkHttp Client initialization to the background thread.

### Motivation
It's a standard practice to initialize the Okhttp client on the background thread. It's a pretty trivial thing to do since there already exists `Call.Factory` interface. 

### Additional Notes
https://github.com/DataDog/dd-sdk-android/issues/2809

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
@@ -89,10 +89,12 @@ import com.datadog.android.security.Encryption
 import com.google.gson.JsonObject
 import com.lyft.kronos.AndroidClockFactory
 import com.lyft.kronos.KronosClock
+import okhttp3.Call
 import okhttp3.CipherSuite
 import okhttp3.ConnectionSpec
 import okhttp3.OkHttpClient
 import okhttp3.Protocol
+import okhttp3.Request
 import okhttp3.TlsVersion
 import java.io.File
 import java.io.FileNotFoundException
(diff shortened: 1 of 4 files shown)


Laimiux:
> Hello @Laimiux! Thanks for your contribution.
> 
> Before moving forward, can you please tell what is the impact of the OkHttp initialization on the overall Datadog SDK initialization duration in your app?
> 
> As I mentioned [here](https://github.com/DataDog/dd-sdk-android/issues/2809#issuecomment-3199809043), for our sample app running on emulator it is around 1%.
> 
> > For more context https://www.zacsweers.dev/dagger-party-tricks-deferred-okhttp-init/
> > OkHttpClient initialization can take upwards of 100ms on some Android devices in the wild due to its use of TrustManagerFactory. Initializing the Cache is also not free, as it creates an internal Executor and may incur some disk IO if you want to prepare a cache directory for it first.
> 
> Yes, but this article talks about generic setup, so it is a generic advice. Datadog SDK have a specific setup which doesn't use cache. SDK indeed is using TLS, so `TrustManagerFactory` will be invoked and the particular provider backing it up depends on the OEM (that is why its initialization time may vary), but on the chart you've posted [here](https://github.com/DataDog/dd-sdk-android/issues/2809#issuecomment-3198252109) percentile is pretty much uniformed (so there are probably no spikes in the initialization duration for different OEMs due to runtime implementation issues).

I see it take about ~9% of total initialization. While the other areas are bigger, this one is easy to address. 

Screenshot 2025-03-20 at 12 00 25 PM

ambushwork (approved):

0xnm:
Closing in favour of #2829.

PR #2825, opened 2025-08-20
RUM-11385: Create Nav3 demo screen in sample app

ambushwork:
### What does this PR do?

Create Nav3 demo screen in sample app.

The screen use `NavDisplay` and navigates among 3 simple compose screen: "Home", "Discovery" and "Settings"

### Motivation

RUM-11385

### Additional Notes

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -95,6 +95,9 @@ timber = "5.0.1"
 coroutines = "1.4.2"
 exoplayer = "2.19.1"
 newPipeExtractor = "v0.24.6"
+navigation3 = "1.0.0-alpha01"
+kotlinxSerializationCore = "1.8.1"
+lifecycleViewmodelNav3 = "1.0.0-alpha01"
 
 # Local Server
 ktor = "2.3.13"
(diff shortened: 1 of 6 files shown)


sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/Navigation3Activity.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContent {
+            Nav3Screen()
+        }
+    }
+
+    @Composable
+    private fun Nav3Screen() {
```
0xnm: there is also a file called `Navigation3Screen`. Maybe we need to call it somehow differently to avoid confusion? Or the other.

0xnm (dismissed):

sample/kotlin/src/main/kotlin/com/datadog/android/sample/compose/Navigation3Activity.kt
ambushwork: renamed this one to `NavDisplaySample`

0xnm (dismissed):

gradle/libs.versions.toml
```diff
@@ -263,8 +265,10 @@ exoplayerOkHttp = { module = "com.google.android.exoplayer:extension-okhttp", ve
 newPipeExtractor = { module = "com.github.TeamNewPipe.NewPipeExtractor:extractor", version.ref = "newPipeExtractor" }
 
 # Core Navigation 3 libraries
-androidx-navigation3-runtime = { module = "androidx.navigation3:navigation3-runtime", version.ref = "nav3Core" }
-androidx-navigation3-ui = { module = "androidx.navigation3:navigation3-ui", version.ref = "nav3Core" }
+androidx-navigation3-runtime = { module = "androidx.navigation3:navigation3-runtime", version.ref = "navigation3" }
+androidx-navigation3-ui = { module = "androidx.navigation3:navigation3-ui", version.ref = "navigation3" }
```
0xnm, line 269: Just wondering: since Navigation 3 has a dedicated artifact ID and also package names, maybe we don't need a dedicated module and can fit everything as a part of `-compose` module?

The downside for the customer will be pulling what they don't need as transitive dependencies, but they will need to pull `compose` module anyway if they want to use manual action tracking I guess.

ambushwork: Actually it makes sense, the tracking side-effect for Navigation3 doesn't even require any dependencies from Nav3, let me create a new feature branch

ambushwork: Retargeted this branch directly to develop since it is a complete task as a sample screen, and removed `:integrations:dd-sdk-android-navigation3` dependency since we don't want to create it anymore.

0xnm (approved):

PR #2828, opened 2025-08-20
[RUM-11251][FO]: Fix RumVitalEvent serialization logic

satween:
### What does this PR do?

It turned out that vital event wasn't sent to backend because there was no serialization logic added.
_(back-end wasn't ready at the previous iterations, so it was hard to notice any issues)_

I've added serialization same way it's done for other events like `LongTaskEvent`, `ViewEvent` etc.
As (according to schema) `RumVitalEvent` also contains `usr`, `context` and `account` sanitization for those attributes was also added. 

I've also added support for `setVitalEventMapper ` and added custom mapper for `sample` and `benchmarking` apps.

Please let me know if some of those changes were unnecessary. 

### DIFF
--- a/features/dd-sdk-android-rum/api/apiSurface
+++ b/features/dd-sdk-android-rum/api/apiSurface
@@ -82,6 +82,7 @@ data class com.datadog.android.rum.RumConfiguration
     fun setActionEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.ActionEvent>): Builder
     fun setErrorEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.ErrorEvent>): Builder
     fun setLongTaskEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.LongTaskEvent>): Builder
+    fun setVitalEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.RumVitalEvent>): Builder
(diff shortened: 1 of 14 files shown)


features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -82,6 +82,7 @@ data class com.datadog.android.rum.RumConfiguration
     fun setActionEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.ActionEvent>): Builder
     fun setErrorEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.ErrorEvent>): Builder
     fun setLongTaskEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.LongTaskEvent>): Builder
+    fun setVitalEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.RumVitalEvent>): Builder
```
0xnm, line 85: It should be called `VitalEvent`, because for other events `Rum` prefix is removed. Can you please do this change as a part of this PR?

ambushwork (approved):

mariusc83 (approved):

features/dd-sdk-android-rum/api/apiSurface
satween: I will make it in the separate one because the amount of changes will be big enough

PR #2830, opened 2025-08-21
RUM-11386: Implement Navigation3 view tracking side effect

ambushwork:
### What does this PR do?

 Implement Navigation3 view tracking side effect

### Motivation

RUM-11386

### Demo

Sample app with `Navigation3TrackingEffect` send correct view events when navigating.

image

### DIFF
--- a/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
+++ b/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
@@ -1,0 +1,124 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+@file:Suppress("MatchingDeclarationName", "PackageNameVisibility")
+
+package com.datadog.android.compose
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.NonRestartableComposable
+import com.datadog.android.Datadog
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.SdkCore
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.compose.internal.InstrumentationType
+import com.datadog.android.compose.internal.SupportLibrary
+import com.datadog.android.compose.internal.sendTelemetry
+import com.datadog.android.internal.attributes.ViewScopeInstrumentationType
+import com.datadog.android.internal.attributes.enrichWithConstantAttribute
+import com.datadog.android.rum.ExperimentalRumApi
+import com.datadog.android.rum.GlobalRumMonitor
+import com.datadog.android.rum.tracking.ComponentPredicate
+
+/**
+ * A side effect which should be used to track view navigation with the Navigation3
+ * for Jetpack Compose setup.
+ *
+ * @param T the type of the key of navigation back stack.
+ * @param backStack backStack of the navigation to watch.
+ * @param keyPredicate to accept the backstack key that will be taken into account as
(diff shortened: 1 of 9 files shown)


integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
```diff
@@ -0,0 +1,122 @@
...
+    InternalNavigation3TrackingStrategy(
+        backStack = backStack,
+        destinationPredicate = keyPredicate,
+        attributesResolver = attributesResolver,
+        sdkCore = sdkCore
+    )
+}
+
+@Composable
+@NonRestartableComposable
```
aleksandr-gringauz, line 87: I know it existed before in `InternalNavigationViewTrackingStrategy`, but curious. Why do we need this `NonRestartableComposable` thing?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
```diff
@@ -0,0 +1,122 @@
...
+@ExperimentalRumApi
+fun <T> Navigation3TrackingEffect(
+    backStack: List<T>,
+    keyPredicate: ComponentPredicate<T> = AcceptAllNavKeyPredicate(),
+    attributesResolver: AttributesResolver<T>? = null,
+    sdkCore: SdkCore = Datadog.getInstance()
+) {
+    sendTelemetry(
```
aleksandr-gringauz: Shouldn't you wrap this in `LaunchedEffect(Unit)` ? So that telemetry isn't sent on every recomposition.

ambushwork: nice catch, fixed

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
ambushwork: If I understand correctly it tells compiler that this function is stable, never needs to be recomposed independently, it's a recomposition optimisation

aleksandr-gringauz (approved):

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
```diff
@@ -0,0 +1,124 @@
...
+    val topKey = backStack.lastOrNull()
+    LaunchedEffect(topKey) {
+        topKey?.takeIf { destinationPredicate.accept(it) }?.let { current ->
+            try {
+                val attributes =
+                    attributesResolver?.resolveAttributes(current)?.toMutableMap()
+                        ?.enrichWithConstantAttribute(ViewScopeInstrumentationType.COMPOSE)
+                        ?: emptyMap()
+
+                GlobalRumMonitor.get(sdkCore).startView(
```
aleksandr-gringauz, line 104: I found [this](https://github.com/DataDog/dd-sdk-android/blob/0606a315338a2026db105ce26e16c6348153dfdc/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/ComposeNavigationObserver.kt#L36) `stopView` call when `onPause` in lifecycle happens. Shouldn't we do a similar thing here?

In all our "tracking strategies" (for example [here](https://github.com/DataDog/dd-sdk-android/blob/0606a315338a2026db105ce26e16c6348153dfdc/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityViewTrackingStrategy.kt#L74)) we stop the view when the app goes to background.

ambushwork: yes, you are totally right, actually that will be my next commit, sorry for not being clear here

PR #2832, opened 2025-08-22
RUM-11386: Make Navigation3 tracking listen to lifecycle

ambushwork:
### What does this PR do?

This PR mainly attach the `Navigation3TrackingEffect` to the lifecycle of the lifecycle owner.
Also there is some improvements:
* Move the core logic to a dedicated internal non-composable function so it can be easily unit tested.
* Adding `BackstackKeyResolver` interface which can accept the custom key resolving, by default we resolve it with hashcode.
* 

### Motivation

### Additional Notes

### DIFF
--- a/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackStackKeyResolver.kt
+++ b/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackStackKeyResolver.kt
@@ -1,0 +1,29 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.compose
+
+/**
+ * Resolves a stable key for an item in a back stack.
+ * A stable key means that the same item will always return the same key during its lifetime in the
+ * back stack, and that two different items should not return the same key.
+ *
+ * This is used in [com.datadog.android.rum.RumMonitor.startView] and
+ * [com.datadog.android.rum.RumMonitor.stopView] as the key to identify items in the back
+ * stack and track them as Views in RUM.
+ *
+ * @param T the type of item in the back stack.
+ */
+interface BackStackKeyResolver<T : Any> {
+
+    /**
+     * Returns a stable key for the given item.
+     *
+     * @param item the item to get the stable key for.
+     * @return a stable key for the item.
+     */
+    fun getStableKey(item: T): String
+}
--- a/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/HashcodeBackStackKeyResolver.kt
+++ b/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/HashcodeBackStackKeyResolver.kt
@@ -1,0 +1,21 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.compose
+
+/**
+ * A [BackStackKeyResolver] that uses the [Any.hashCode] of the item as its stable key.
+ *
+ * @param T the type of item in the back stack.
+ */
+class HashcodeBackStackKeyResolver<T : Any> : BackStackKeyResolver<T> {
+
+    override fun getStableKey(
+        item: T
+    ): String {
+        return item.hashCode().toString()
+    }
+}
--- a/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
+++ b/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
@@ -10,6 +10,12 @@ package com.datadog.android.compose
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.NonRestartableComposable
+import androidx.compose.runtime.State
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.produceState
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleEventObserver
 import com.datadog.android.Datadog
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
@@ -19,8 +25,8 @@ import com.datadog.android.compose.internal.SupportLibrary
 import com.datadog.android.compose.internal.sendTelemetry
 import com.datadog.android.internal.attributes.ViewScopeInstrumentationType
 import com.datadog.android.internal.attributes.enrichWithConstantAttribute
-import com.datadog.android.rum.ExperimentalRumApi
 import com.datadog.android.rum.GlobalRumMonitor
+import com.datadog.android.rum.RumMonitor
 import com.datadog.android.rum.tracking.ComponentPredicate
 
 /**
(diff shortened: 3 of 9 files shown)


integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
```diff
@@ -31,15 +37,17 @@ import com.datadog.android.rum.tracking.ComponentPredicate
...
  * @param sdkCore the SDK instance to use. If not provided, default instance will be used.
  */
 @Composable
 @NonRestartableComposable
-@ExperimentalRumApi
+@ExperimentalTrackingApi
 fun <T> Navigation3TrackingEffect(
     backStack: List<T>,
     keyPredicate: ComponentPredicate<T> = AcceptAllNavKeyPredicate(),
+    backstackKeyResolver: BackstackKeyResolver<T> = HashcodeBackstackKeyResolver(),
```
0xnm: minor: we need to align spelling of `backStack` vs `backstack`

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackstackKeyResolver.kt
```diff
@@ -0,0 +1,26 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.compose
+
+/**
+ * Resolves a stable key for an item in a backstack.
+ * A stable key means that the same item will always return the same key during its lifetime in the
+ * backstack.
```
0xnm: should we also add the statement that it shouldn't be also 2 different items with the same key?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackstackKeyResolver.kt
```diff
@@ -0,0 +1,26 @@
...
+ * This is used to identify items in the backstack and track them as Views in RUM.
```
0xnm: maybe we can make a reference/link to the `RumMonitor.startView` and `RumMonitor.stopView` APIs here, and to the key argument there specifically.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackstackKeyResolver.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+interface BackstackKeyResolver<T> {
+
+    /**
+     * Returns a stable key for the given item.
+     *
+     * @param item the item to get the stable key for.
+     * @return a stable key for the item.
+     */
+    fun getStableKey(item: T): String
```
0xnm: just a side note: do we constrain ourselves here to `String`? Key in the RUM API can be `Any`, so in theory return type can also be generic.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/HashcodeBackstackKeyResolver.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.compose
+
+/**
+ * A [BackstackKeyResolver] that uses the [hashCode][Any.hashCode] of the item as its stable key.
```
0xnm:
```suggestion
 * A [BackstackKeyResolver] that uses the [Any.hashCode] of the item as its stable key.
```

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
ambushwork: To follow the Google API grammar, `backStack` is picked for argument name and `back stack` for documentation.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackstackKeyResolver.kt
ambushwork: added.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackstackKeyResolver.kt
ambushwork: The problem I see here is that, if we allow `Any` as the return type, than we need to trust that client resolves the identity of this return type correctly, and I believe some (or most) of clients will just do:
```
override fun fun getStableKey(item: MyType): Any{
      return item
}
```

Forcing the return type to string will make it more intuitive if they are equal or not.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
```diff
@@ -88,23 +99,48 @@ internal fun <T> InstrumentedNavigation3TrackingEffect(
...
+    isResumed: Boolean,
+    keyPredicate: ComponentPredicate<T> = AcceptAllNavKeyPredicate(),
+    backStackKeyResolver: BackStackKeyResolver<T>,
+    attributesResolver: AttributesResolver<T>? = null,
+    rumMonitor: RumMonitor,
+    internalLogger: InternalLogger
+) {
+    val viewKey = backStackKeyResolver.getStableKey(topKey)
+    if (isResumed) {
+        if (keyPredicate.accept(topKey)) {
```
0xnm: should we log info if it wasn't accepted?

integrations/dd-sdk-android-compose/src/test/kotlin/com/datadog/android/compose/HashcodeBackStackKeyResolverTest.kt
```diff
@@ -0,0 +1,73 @@
...
+
+@ExtendWith(ForgeExtension::class)
+internal class HashcodeBackStackKeyResolverTest {
+
+    private lateinit var testedResolver: HashcodeBackStackKeyResolver<Any>
+
+    @BeforeEach
+    fun `set up`() {
+        testedResolver = HashcodeBackStackKeyResolver()
+    }
```
0xnm:
```suggestion
    private val testedResolver = HashcodeBackStackKeyResolver()
```

0xnm (dismissed):
lgtm, but it would be nice to have another pair of eyes

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
```diff
@@ -88,23 +99,48 @@ internal fun <T> InstrumentedNavigation3TrackingEffect(
...
+            keyPredicate = destinationPredicate,
+            backStackKeyResolver = backStackKeyResolver,
+            attributesResolver = attributesResolver,
+            rumMonitor = GlobalRumMonitor.get(sdkCore),
+            internalLogger = internalLogger
+        )
+    }
+}
+
+internal fun <T> trackBackStack(
```
aleksandr-gringauz: nit

you can add a lower bound on T `<T : Any>` in order to signal that it is non-null (it is, because you do `val topKey = backStack.lastOrNull() ?: return` before).

this way there will be no need to have a case for null in `resolveDefaultViewName`

aleksandr-gringauz: upd

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation3/navigation3-ui/src/commonMain/kotlin/androidx/navigation3/ui/NavDisplay.kt;l=146?q=public%20fun%20%3CT%20:%20Any%3E%20NavDisplay

you should make `<T : Any>` everywhere you use generics, because on the Compose side it can't be null

0xnm: nice point

ambushwork: Nice catch, fixed

0xnm (dismissed):

aleksandr-gringauz (dismissed):

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/HashcodeBackStackKeyResolver.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package com.datadog.android.compose
+
+/**
+ * A [BackStackKeyResolver] that uses the [Any.hashCode] of the item as its stable key.
+ *
+ * @param T the type of item in the back stack.
+ */
+class HashcodeBackStackKeyResolver<T> : BackStackKeyResolver<T> {
```
aleksandr-gringauz: you should place `<T:Any>` here as well I think

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/BackStackKeyResolver.kt
```diff
@@ -0,0 +1,29 @@
...
+ * A stable key means that the same item will always return the same key during its lifetime in the
+ * back stack, and that two different items should not return the same key.
+ *
+ * This is used in [com.datadog.android.rum.RumMonitor.startView] and
+ * [com.datadog.android.rum.RumMonitor.stopView] as the key to identify items in the back
+ * stack and track them as Views in RUM.
+ *
+ * @param T the type of item in the back stack.
+ */
+interface BackStackKeyResolver<T> {
```
aleksandr-gringauz: and here

aleksandr-gringauz (dismissed):

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/Navigation3.kt
ambushwork: added a debug log to user target.

aleksandr-gringauz (approved):

PR #2845, opened 2025-09-01
RUM-11196: Support apollo graphql

jonathanmos:
### What does this PR do?
Adds support for Apollo GraphQL requests, which is essentially native support for graphQL (until now only available from cross-platform).

Caveats:
Support only for V4 of Apollo-Kotlin. If there will be enough demand for V3 we may extend support there.

How does it work:
- Added a new `ApolloInterceptor` plugin under a new integration module: `dd-sdk-android-apollo`
- This plugin adds some headers to graphQL requests containing request fields: `operationName`, `operationType`, `variables` and `payload`.
- Added logic inside `DatadogInterceptor` to catch these headers and send them as part of `RumResourceScope`. After extracting the data from these headers they are removed so they won't be sent onwards to the host.
- All the fields are sent by default with the exception of payload. To send it the user must set a new configuration option inside the DatadogApolloInterceptor constructor -> `DatadogApolloInterceptor(sendGraphQLPayloads = true)`
- To setup, the user adds our ApolloInterceptor plugin to their ApolloClient configuration (as described in the Readme)

### Motivation
Add GraphQL support to native.

### Additional Notes

- Notice that we can't test this at the moment as part of the sample because it's graphQL. This means that the payload flag is also not added to the sampleApplication because it would have no effect. I'm testing it manually but we'll have to create an automatic testing solution. 
- The new module adds about 4MB in terms of dependency size and two additional MIT dependency licenses (apollo-kotlin and a uuid dependency that apollo-kotlin uses internally).


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -335,6 +335,16 @@ data class RumConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Enable/Disable the collection of graphQL payloads.
+         *
+         * By default the SDK will not collect these payloads.
+         */
+        fun captureGraphQLPayloads(enabled: Boolean): Builder {
```
0xnm: I don't think it should be here, because it is a technology-specific API. WDYT about moving it to `_RumInternalProxy` and calling only from `Apollo` module?

Also it can be just the attribute passed to the `startResource/stopResource`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -194,6 +195,10 @@ internal class RumFeature(
         backgroundEventTracking = configuration.backgroundEventTracking
         trackFrustrations = configuration.trackFrustrations
 
+        sdkCore.updateFeatureContext(RUM_FEATURE_NAME) {
+            it[SEND_GRAPHQL_PAYLOADS_KEY] = configuration.captureGraphQLPayloads
```
0xnm: minor: it is better to align wording - send vs capture.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -214,11 +216,48 @@ internal class RumResourceScope(
         val finalTiming = timing ?: extractResourceTiming(
             attributes.remove(RumAttributes.RESOURCE_TIMINGS) as? Map<String, Any?>
         )
+
+        val rumFeatureContext = sdkCore.getFeatureContext(RUM_FEATURE_NAME)
```
0xnm: if context is needed, we should get it from `newRumEventWriteOperation(writer) { datadogContext` below

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -214,11 +216,48 @@ internal class RumResourceScope(
...
+        val shouldSendGraphQLPayloads = rumFeatureContext[SEND_GRAPHQL_PAYLOADS_KEY] as? Boolean ?: false
```
0xnm: but anyway, I don't think it should be a property of the context. It is a property of `RumConfiguration` which should be read when `DatadogRumMonitor` is created and passed down to the `RumResourceScope` constructor, because it is a property which is a) constant; b) is there for the whole lifetime of the RUM feature; c) written and read only in the boundaries of `rum` module

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAttributes.kt
```diff
@@ -109,22 +109,42 @@ object RumAttributes {
...
+
+    /**
+     * GraphQL operation type coming from Apollo Kotlin plugin (String).
+     */
+    const val APOLLO_GRAPHQL_OPERATION_TYPE: String = "_dd-custom-header-graph-ql-operation-type"
+
+    /**
+     * GraphQL payload coming from Apollo Kotlin plugin (String).
+     */
+    const val APOLLO_GRAPHQL_PAYLOAD: String = "_dd-custom-header-graph-ql-payload"
```
0xnm: we cannot reuse existing attributes?

0xnm (changes requested):
I briefly reviewed the RUM module part of it and left some suggestions.

It is good to see finally GraphQL support in the SDK!

integrations/dd-sdk-android-apollo/transitiveDependencies
```diff
@@ -0,0 +1,17 @@
...
+com.apollographql.apollo:apollo-api-jvm:4.3.3                   :  330 Kb
+com.apollographql.apollo:apollo-mpp-utils-jvm:4.3.3             : 1477 b 
+com.apollographql.apollo:apollo-runtime-android:4.3.3           :  379 Kb
+com.benasher44:uuid-jvm:0.8.2                                   :    8 Kb
+com.squareup.okhttp3:okhttp:4.12.0                              :  771 Kb
+com.squareup.okio:okio-jvm:3.9.0                                :  363 Kb
+org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.8.21                  :  963 b 
+org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.21                  :  968 b 
+org.jetbrains.kotlin:kotlin-stdlib:2.0.0                        : 1689 Kb
+org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.9.0         : 1429 Kb
```
ambushwork, line 13: is `coroutine` introduced by `apollo` or by our implementation?

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
```diff
@@ -0,0 +1,91 @@
...
+import com.apollographql.apollo.interceptor.ApolloInterceptorChain
+import kotlinx.coroutines.flow.Flow
+import okio.IOException
+
+/**
+ * A Datadog Apollo interceptor for GraphQL operations.
+ */
+class DatadogApolloInterceptor(
+    // wrap the extension function (for unit tests)
+    private val variablesExtractor: (
```
ambushwork: is it possible to expose it to unit test by using a function? here is two potential concerns that I see:
* User will wonder what it is when constructing the object and still be able to pass something inside.
* It might break binary compatibility of the constructor of `DatadogApolloInterceptor` when this `variablesExtractor` changes.

integrations/dd-sdk-android-apollo/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    testImplementation(libs.bundles.jUnit5)
+    testImplementation(libs.bundles.testTools)
+    testImplementation(libs.okHttpMock)
+}
+
+kotlinConfig(jvmBytecodeTarget = JvmTarget.JVM_11)
+junitConfig()
+androidLibraryConfig()
+publishingConfig(
+    projectDescription = "An interceptor for handling graphql requests"
```
ambushwork: should we mention apollo here?

ambushwork (dismissed):

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,57 @@
...
+    implementation "com.datadoghq:dd-sdk-android-okhttp:<latest-version>"
+    implementation "com.datadoghq:dd-sdk-android-rum:<latest-version>"
+    implementation "com.datadoghq:dd-sdk-android-apollo:<latest-version>"
+}
+```
+
+### Initial Setup
+
+1. Setup RUM monitoring, see the dedicated [Datadog Android RUM Collection documentation][2] to learn how.
+2. Setup OkHttp instrumentation with Datadog RUM SDK, see the [dedicated documentation][3] to learn how.
```
urseberry:
```suggestion
1. Set up RUM monitoring with [Datadog Android RUM][2].
2. Set up OkHttp instrumentation with the [Datadog RUM SDK][3].
```

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Add the Datadog interceptor to your apollo client setup:
```
urseberry:
```suggestion
Add the Datadog interceptor to your Apollo Client setup:
```

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,57 @@
+# Datadog Integration for Apollo-Kotlin
+
+## Getting Started
```
urseberry:
```suggestion
## Getting started
```

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```groovy
+dependencies {
+    implementation "com.datadoghq:dd-sdk-android-okhttp:<latest-version>"
+    implementation "com.datadoghq:dd-sdk-android-rum:<latest-version>"
+    implementation "com.datadoghq:dd-sdk-android-apollo:<latest-version>"
+}
+```
+
+### Initial Setup
```
urseberry:
```suggestion
### Initial setup
```

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,57 @@
...
+    .serverUrl([graphQLEndpoint])
+    .addInterceptor(DatadogApolloInterceptor())
+    .okHttpClient([okhttpClientConfiguration])
+    .build()
+```
+
+This automatically adds Datadog headers to your GraphQL requests, allowing them to be tracked
+by Datadog. Note that while `query` and `mutation` type operations are tracked, `subscription` operations are not.
+
+#### Sending GraphQL Payloads
```
urseberry:
```suggestion
#### Sending GraphQL payloads
```

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+GraphQL payload capture is disabled by default, so if you want to use it you must set the `captureGraphQLPayloads` flag in the RumConfiguration as follows:
```
urseberry:
```suggestion
GraphQL payload capture is disabled by default. To use it, set the `captureGraphQLPayloads` flag in the RumConfiguration as follows: 
```

integrations/dd-sdk-android-apollo/transitiveDependencies
jonathanmos: Apollo Kotlin is [introducing](https://github.com/apollographql/apollo-kotlin/blob/b6b45145b09e227049949aed4b7deaf2b1b4d5a2/gradle/libraries.toml) it

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAttributes.kt
jonathanmos: after another look at this it does seem we can reuse these attributes and I've changed the code accordingly

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -1230,6 +1231,30 @@ data class com.datadog.android.rum.model.LongTaskEvent
     fun toJson(): com.google.gson.JsonElement
     companion object 
       fun fromJson(kotlin.String): InvokerType
+data class com.datadog.android.rum.model.ProfilingInternalContextSchema
```
0xnm: this and new classes below shouldn't be there, you probably need to regenerate models, considering https://github.com/DataDog/dd-sdk-android/pull/2834

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAttributes.kt
```diff
@@ -107,22 +107,22 @@ object RumAttributes {
     const val RESOURCE_TIMINGS: String = "_dd.resource_timings"
 
     /**
-     * GraphQL operation type coming from cross-platform SDK (String).
+     * GraphQL operation type K (String).
```
0xnm:
```suggestion
     * GraphQL operation type (String).
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -775,6 +780,9 @@ internal class RumFeature(
         internal const val FAILED_TO_ENABLE_JANK_STATS_TRACKING_MANUALLY =
             "Manually enabling JankStats tracking threw an exception."
 
+        internal const val SEND_GRAPHQL_PAYLOADS_KEY =
```
0xnm: it doesn't seem to be used

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,25 +547,40 @@ internal class RumResourceScope(
...
 
+    private fun truncateGraphQLPayload(payload: String): String {
+        val payloadBytes = payload.toByteArray(Charsets.UTF_8)
+
+        return if (payloadBytes.size <= MAX_GRAPHQL_PAYLOAD_SIZE) {
+            payload
+        } else {
+            // Truncate to MAX_GRAPHQL_PAYLOAD_SIZE and ensure we don't break UTF-8 encoding
+            val truncatedBytes = payloadBytes.copyOf(MAX_GRAPHQL_PAYLOAD_SIZE)
+            String(truncatedBytes, Charsets.UTF_8)
```
0xnm: We do break UTF-8 encoding here, because each code point in UTF-8 is 1 to 4 bytes (https://en.wikipedia.org/wiki/UTF-8#Description).

So by truncating like that we _may_ get invalid codepoint.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,25 +547,40 @@ internal class RumResourceScope(
...
+        }
+    }
+
     // endregion
 
     companion object {
 
+        @VisibleForTesting
+        internal const val MAX_GRAPHQL_PAYLOAD_SIZE = 30 * 1024
```
0xnm: probably it is in bytes?

```suggestion
        internal const val MAX_GRAPHQL_PAYLOAD_SIZE_BYTES = 30 * 1024
```

integrations/dd-sdk-android-apollo/api/apiSurface
```diff
@@ -0,0 +1,9 @@
+class com.datadog.android.apollo.DatadogApolloInterceptor : com.apollographql.apollo.interceptor.ApolloInterceptor
+  constructor((com.apollographql.apollo.api.Operation<*>, com.apollographql.apollo.api.CustomScalarAdapters) -> String? = { operationToVariablesJson, adapters ->
```
0xnm: probably this file needs to be re-generated, here it is the implementation body, not type/function definition

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
```diff
@@ -0,0 +1,87 @@
...
+            requestBuilder.addHttpHeader(DD_GRAPHQL_TYPE_HEADER, operationType)
+        }
+
+        operationVariables?.let {
+            requestBuilder.addHttpHeader(DD_GRAPHQL_VARIABLES_HEADER, operationVariables)
+        }
+
+        operationPayload.let {
+            requestBuilder.addHttpHeader(DD_GRAPHQL_PAYLOAD_HEADER, operationPayload)
+        }
```
0xnm, line 72: blocker: I don't think we need to go into this direction, because these headers will be sent to the final host for nothing, this is only info we need to collect, but not to send out with the request itself.

Instead, you need to use `Request.tag` and put the necessary structure there and later on read in `DatadogInterceptor`. Then we must mention in the docs that `DatadogApolloInterceptor` should be placed before `DatadogInterceptor` (to keep the right order).

This will also remove the need for the `sendGraphQlPayloads` RUM configuration, because this flag can be a part of `DatadogApolloInterceptor` configuration and make it to the `Request.tag` as well, then extracted and passed as an attribute to the `stopResource` call.

0xnm (changes requested):

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
jonathanmos: Request.tag unfortunately isn't available in ApolloInterceptor because it is intercepting at the Apollo layer before reaching okhttp, and this is the reason that also in the cross platform solution they used custom headers. Regarding the concern of sending the headers to the final host, I could remove them inside DatadogInterceptor as soon as they have been read.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
jonathanmos: I'll introduce a more robust method

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,57 @@
+# Datadog Integration for Apollo-Kotlin
+
+## Getting started
+
+Note that we support only Apollo version 4.
```
urseberry:
```suggestion
**Note:** The integration only supports Apollo version 4.
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
jonathanmos: I've moved this to be a constructor parameter of the `DatadogApolloInterceptor`

urseberry (dismissed):

detekt_custom_safe_calls.yml
```diff
@@ -265,6 +265,16 @@ datadog:
...
       - "com.google.android.material.tabs.TabLayout.TabView.getChildAt(kotlin.Int)"
       # endregion
+      # region Apollo Kotlin
+      - "com.apollographql.apollo.api.json.buildJsonString(kotlin.String?, kotlin.Function1)"
+      - "com.apollographql.apollo.api.Operation.composeJsonRequest(com.apollographql.apollo.api.json.JsonWriter, com.apollographql.apollo.api.CustomScalarAdapters)"
+      - "com.apollographql.apollo.api.Operation.document()"
+      - "com.apollographql.apollo.api.Operation.name()"
+      - "com.apollographql.apollo.api.ApolloRequest.newBuilder()"
+      - "com.apollographql.apollo.api.ApolloRequest.Builder.addHttpHeader(kotlin.String, kotlin.String)"
+      - "com.apollographql.apollo.api.ApolloRequest.Builder.build()"
```
0xnm: very nit:`com.apollographql.apollo.api.Operation` must be after `com.apollographql.apollo.api.ApolloRequest`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -242,11 +242,20 @@ internal class RumResourceScope(
         val finalTiming = timing ?: extractResourceTiming(
             resourceAttributes.remove(RumAttributes.RESOURCE_TIMINGS) as? Map<String, Any?>
         )
+
+        val graphqlOperationName = resourceAttributes.remove(RumAttributes.GRAPHQL_OPERATION_NAME) as? String
+        val graphqlOperationType = resourceAttributes.remove(RumAttributes.GRAPHQL_OPERATION_TYPE) as? String
+        val graphqlVariables = resourceAttributes.remove(RumAttributes.GRAPHQL_VARIABLES) as? String
+
+        // The decision whether to send payloads is determined by a datadogInterceptor parameter
```
0xnm: isn't it a `DatadogApolloInterceptor` parameter?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,25 +542,61 @@ internal class RumResourceScope(
...
+            payload.substring(0, payload.length - charactersToRemove)
+        }
+    }
+
     // endregion
 
     companion object {
 
+        @VisibleForTesting
+        internal const val MAX_GRAPHQL_PAYLOAD_SIZE_BYTES = 30 * 1024
```
0xnm: `@VisibleForTesting` annotation doesn't really add any value in Kotlin, it was rather designed for the usage in Java code.

```suggestion
        internal const val MAX_GRAPHQL_PAYLOAD_SIZE_BYTES = 30 * 1024
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScopeTest.kt
```diff
@@ -2816,10 +2819,154 @@ internal class RumResourceScopeTest {
...
+        // When
+        testedScope.handleEvent(mockEvent, fakeDatadogContext, mockEventWriteScope, mockWriter)
+
+        // Then
+        argumentCaptor<ResourceEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture(), eq(EventType.DEFAULT))
+            val actualPayload = firstValue.resource.graphql?.payload
+
+            assertThat(actualPayload).isNotNull()
+            assertThat(actualPayload!!.length).isLessThan(originalPayload.length)
```
0xnm:
```suggestion
            checkNotNull(actualPayload)
            assertThat(actualPayload.length).isLessThan(originalPayload.length)
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScopeTest.kt
```diff
@@ -2816,10 +2819,154 @@ internal class RumResourceScopeTest {
...
+            assertThat(actualPayload.toByteArray(Charsets.UTF_8).size)
+                .isLessThanOrEqualTo(RumResourceScope.MAX_GRAPHQL_PAYLOAD_SIZE_BYTES)
+
+            assertThat(actualPayload).startsWith(originalPayload.take(10))
```
0xnm: I think it is better to write something like `assertThat(originalPayload).startsWith(actualPayload)` here

integrations/dd-sdk-android-apollo/.gitignore
```diff
@@ -0,0 +1,18 @@
...
+# Java class files
+*.class
+
+# Generated files
+bin/
+gen/
+out/
+
+# Gradle files
+build/
```
0xnm:
```suggestion
build/

```

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,55 @@
+# Datadog Integration for Apollo-Kotlin
```
0xnm: it is `Apollo Kotlin` in the official docs https://www.apollographql.com/docs/kotlin

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Getting started
+
+**Note:** The integration only supports Apollo version 4.
+
+To include the integration for [Apollo-Kotlin][1] in your project, add the
```
0xnm:
```suggestion
To include the integration for [Apollo Kotlin][1] in your project, add the
```

integrations/dd-sdk-android-apollo/README.md
```diff
@@ -0,0 +1,55 @@
...
+    .okHttpClient([okhttpClientConfiguration])
+    .build()
+```
+
+This automatically adds Datadog headers to your GraphQL requests, allowing them to be tracked
+by Datadog. Note that while `query` and `mutation` type operations are tracked, `subscription` operations are not.
+
+#### Sending GraphQL payloads
+
+GraphQL payload sending is disabled by default. To enable it, set the `sendGraphQLPayloads` flag in the DatadogApolloInterceptor constructor as follows:
```
0xnm:
```suggestion
GraphQL payload sending is disabled by default. To enable it, set the `sendGraphQLPayloads` flag in the `DatadogApolloInterceptor` constructor as follows:
```

integrations/dd-sdk-android-apollo/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    testImplementation(libs.bundles.jUnit5)
+    testImplementation(libs.bundles.testTools)
+    testImplementation(libs.okHttpMock)
+}
+
+kotlinConfig(jvmBytecodeTarget = JvmTarget.JVM_11)
+junitConfig()
+androidLibraryConfig()
+publishingConfig(
+    projectDescription = "An Apollo interceptor for handling graphql requests"
```
0xnm:
```suggestion
    projectDescription = "An Apollo interceptor for handling GraphQL requests to use with the Datadog monitoring library for Android applications."
```

integrations/dd-sdk-android-apollo/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+</manifest>
```
0xnm:
```suggestion
</manifest>

```

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
```diff
@@ -0,0 +1,91 @@
...
+import com.apollographql.apollo.api.variablesJson
+import com.apollographql.apollo.interceptor.ApolloInterceptor
+import com.apollographql.apollo.interceptor.ApolloInterceptorChain
+import kotlinx.coroutines.flow.Flow
+import okio.IOException
+
+/**
+ * A Datadog Apollo interceptor for GraphQL operations.
+ */
+open class DatadogApolloInterceptor(
```
0xnm: it seems there is no need for this calls to be `open`

```suggestion
class DatadogApolloInterceptor(
```

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
```diff
@@ -0,0 +1,91 @@
...
+    private val sendGraphQLPayloads: Boolean = false
```
0xnm, line 27: documentation is missing for this public property

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
```diff
@@ -0,0 +1,91 @@
...
+    private fun <D : Operation.Data> extractPayload(operation: Operation<D>, adapters: CustomScalarAdapters) =
+        buildJsonString {
+            operation.composeJsonRequest(this, adapters)
+        }
+
+    internal companion object {
+        const val DD_GRAPHQL_NAME_HEADER = "_dd-custom-header-graph-ql-operation-name"
+        const val DD_GRAPHQL_VARIABLES_HEADER = "_dd-custom-header-graph-ql-variables"
+        const val DD_GRAPHQL_TYPE_HEADER = "_dd-custom-header-graph-ql-operation_type"
+        const val DD_GRAPHQL_PAYLOAD_HEADER = "_dd-custom-header-graph-ql-payload"
```
0xnm: maybe you can extract this constants into the `dd-sdk-android-internal` module and then reference them from both Apollo and OkHttp modules?

This will be better, because if there is some typo/mismatch in these literals between Apollo and OkHttp declarations, we won't find this, because there is no integration test (which we could add btw in the `integrations/single-fit/okttp`).

integrations/dd-sdk-android-apollo/src/test/java/com/datadog/android/apollo/DatadogApolloInterceptorTest.kt
```diff
@@ -0,0 +1,446 @@
...
+        val (_, originalRequest, requestBuilder) = setupBasicMocks("GetUser", "query GetUser { user { id: name } }")
+        val chain = mock<ApolloInterceptorChain>()
+
+        // When
+        testedInterceptor.intercept(originalRequest, chain)
+
+        // Then
+        val headerNameCaptor = argumentCaptor<String>()
+        val headerValueCaptor = argumentCaptor<String>()
+        verify(requestBuilder, atLeastOnce()).addHttpHeader(headerNameCaptor.capture(), headerValueCaptor.capture())
```
0xnm: why `atLeastOnce` is needed here?

integrations/dd-sdk-android-apollo/src/test/java/com/datadog/android/apollo/DatadogApolloInterceptorTest.kt
```diff
@@ -0,0 +1,446 @@
...
+
+    private fun setupApolloInterceptor(sendGraphQLPayloads: Boolean): DatadogApolloInterceptor =
+        object : DatadogApolloInterceptor(
+            sendGraphQLPayloads
+        ) {
+            override fun extractVariables(operation: Operation<*>, adapters: CustomScalarAdapters): String? {
+                return when (operation.name()) {
+                    "GetUser" -> """{"userId": "123", "filters": ["active"]}"""
+                    "SetUser" -> """{"userId": "123", "filters": ["active"]}"""
+                    "CreateUser" -> """{"input": {"name": "John", "email": "john@example.com"}}"""
```
0xnm: minor: probably some random values can be used here

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -361,6 +439,9 @@ open class DatadogInterceptor internal constructor(
                 "Make sure you initialized the Datadog SDK with a valid Application Id, " +
                 "and that RUM features are enabled."
 
+        internal const val ERROR_FAILED_BUILD_REQUEST =
+            "Failed to build interceptor chain after removing DD headers. Falling back to original chain"
```
0xnm:
```suggestion
            "Failed to build interceptor chain after removing DD headers. Falling back to original chain."
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -192,6 +215,49 @@ open class DatadogInterceptor internal constructor(
...
+                    target = InternalLogger.Target.MAINTAINER,
+                    messageBuilder = { ERROR_FAILED_BUILD_REQUEST },
+                    throwable = e
+                )
+                originalChain // fallback to the original request
+            }
+        } else {
+            originalChain
+        }
+    }
```
0xnm, line 253: I don't think we need to add such complexity by building the chain. What you can do here is just remove headers before `.proceed()` call and move them under the `Request.tag` and take them out from there when response arrives.

By doing this they won't go out and you still have access to the necessary information.

0xnm (changes requested):
Looks alright! I've added a few comments and suggestions.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,24 +542,58 @@ internal class RumResourceScope(
...
+                // Once we've removed enough bytes to fit within the limit, stop
+                if (bytesToRemove >= excessBytes) {
+                    break
+                }
+            }
+
+            @Suppress("UnsafeThirdPartyFunctionCall") // indexOutOfBounds cant be thrown here
+            payload.substring(0, payload.length - charactersToRemove)
+        }
+    }
```
0xnm: there is a simpler and more efficient way to do this (in the end, manipulating something over 30Kb in the memory may have performance impact):

*  we can avoid redundant `val payloadBytes = payload.toByteArray(Charsets.UTF_8)` allocation by checking `payload.length` first: since strings are stored as UTF-16 in runtime and UTF-16 is two 16-bit code units max for code point, if `payload.length == codepoint number` it means that if `payload.length` is less than 7500, it is **definitely** below 30Kb limit (7500*4). If it is more, than we need to have a more precise look by getting exact bytes.
* we can truncate easier: you can take advantage from the fact that in UTF-8 continuation byte is always starting from `10` https://en.wikipedia.org/wiki/UTF-8#Description, so we just need to slice `(MAX_BYTES - something)` where something is 0 if `payloadBytes[MAX_BYTES]` starts from 0 or some number which can be found by going from the tail till the closest byte starting from `11`.

Not sure about the performance impact, but it seems it will at least have less allocations.

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.apollographql.apollo.interceptor.ApolloInterceptor
+import com.apollographql.apollo.interceptor.ApolloInterceptorChain
+import com.datadog.android.apollo.internal.DefaultVariablesExtractor
+import com.datadog.android.apollo.internal.VariablesExtractor
+import com.datadog.android.internal.network.GraphQLHeaders
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * A Datadog Apollo interceptor for GraphQL operations.
+ * @param sendGraphQLPayloads should graphQL payloads be reported. This is disabled by default.
```
0xnm:
```suggestion
 * @param sendGraphQLPayloads Should GraphQL payloads be reported or not. This is disabled by default.
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -192,6 +216,49 @@ open class DatadogInterceptor internal constructor(
...
+
+    private fun removeGraphQLHeaders(requestBuilder: Request.Builder) {
+        requestBuilder.removeHeader(GraphQLHeaders.DD_GRAPHQL_NAME_HEADER.headerValue)
+        requestBuilder.removeHeader(GraphQLHeaders.DD_GRAPHQL_TYPE_HEADER.headerValue)
+        requestBuilder.removeHeader(GraphQLHeaders.DD_GRAPHQL_VARIABLES_HEADER.headerValue)
+        requestBuilder.removeHeader(GraphQLHeaders.DD_GRAPHQL_PAYLOAD_HEADER.headerValue)
```
0xnm:
```suggestion
        GraphQLHeaders.values().forEach { requestBuilder.removeHeader(it.headerValue) }
```

It will be more resilient if we add new values to the `GraphQLHeaders` enum

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -277,6 +344,13 @@ open class DatadogInterceptor internal constructor(
         }
     }
 
+    private fun hasGraphQLHeaders(headers: Headers): Boolean {
+        return headers[GraphQLHeaders.DD_GRAPHQL_NAME_HEADER.headerValue] != null ||
```
0xnm: we can do similar thing here `GraphQLHeaders.values().any { headers[it] != null }`

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -192,6 +216,49 @@ open class DatadogInterceptor internal constructor(
...
+        return if (hasGraphQLHeaders(originalChain.request().headers)) {
+            try {
+                object : Interceptor.Chain by originalChain {
+                    override fun proceed(request: Request): Response {
+                        val cleanedRequest = request.newBuilder().apply {
+                            removeGraphQLHeaders(this)
+                        }.build()
+                        return originalChain.proceed(cleanedRequest)
+                    }
+                }
```
0xnm: I'm not sure about the behavior here when we have both Apollo and OkHttp instrumentation. Can you please add a test in `reliability/single-fit/okhttp` with the setup with ApolloInterceptor and DatadogInterceptor and checking that we have expected expected calls of `RumMonitor` and that we have expected headers going out to the server.

Maybe it is worth to check also calls with `TracingInterceptor` to make sure that trace context propagation headers are not affected.

0xnm (changes requested):
LGTM overall, I just added few suggestions.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
jonathanmos: Good catch! I've refactored to use this method

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,24 +542,68 @@ internal class RumResourceScope(
...
         }
 
         return null
     }
 
+    private fun truncateGraphQLPayload(payload: String): String {
+        // Fast path: if string length is less than 7500 characters, it's definitely under 30KB
+        // UTF-16 uses max 2 code units per code point, and UTF-8 uses max 4 bytes per code point
+        // So if payload.length < 7500, then UTF-8 size < 7500 * 4 = 30KB
+        if (payload.length < MAX_GRAPHQL_PAYLOAD_SIZE_BYTES / UTF8_BYTES_PER_CODEPOINT) {
```
0xnm: the comment is a bit wrong: VM stores strings in UTF-16, where code point is 2 16 bit code units max, so when getting `payload.length`, we are not talking about UTF-8, but UTF-16. This `UTF8_BYTES_PER_CODEPOINT` should be called `UTF16_MAX_BYTES_PER_CODEPOINT`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,24 +542,68 @@ internal class RumResourceScope(
...
+            // If we're in the middle of a multi-byte character, we need to back up
+            // UTF-8 continuation bytes start with 10xxxxxx (0x80-0xBF)
+            // UTF-8 start bytes are either 0xxxxxxx or 11xxxxxx
+            // here we apply a mask with 11 to keep only the top two bits, and we compare
+            // these two bits to 10 to see if it's a continuation
+            while (truncateIndex > 0 &&
+                (payloadBytes[truncateIndex].toInt() and UTF8_LEADING_BYTE_MASK) == UTF8_CONTINUATION_BYTE
+            ) {
+                truncateIndex--
+            }
```
0xnm: if I get it right, we are doing the following: imagine a 2 code points (one of 1 byte, another is 4 bytes) like `0xxxxxxx 11xxxxxx 10xxxxxx 10xxxxxx 10xxxxx` [0, 1, 2, 3, 4]. Originally truncate index is at position 3, and according to the loop we move it to the 1 and what we will have as resulting string is `0xxxxxxx`, because we take length.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,24 +542,68 @@ internal class RumResourceScope(
...
+
+            // cannot throw IndexOutOfBounds as offset/length > 0 and offset <= length - bytes
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            String(
```
0xnm: seems like this constructor:

> This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement string.

So we shouldn't have an exception if we cut things wrong.

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,587 @@
...
+        whenever(mockSdkCore.getFeature(Feature.RUM_FEATURE_NAME)) doReturn mockFeatureScope
+        featureScopes[Feature.RUM_FEATURE_NAME] = mockFeatureScope
+        val fakeTraceConfiguration = TraceConfiguration.Builder().build()
+        Trace.enable(fakeTraceConfiguration, stubSdkCore)
+
+        // Setup GlobalRumMonitor mock
+        GlobalRumMonitor::class.declaredFunctions.first { it.name == "registerIfAbsent" }.apply {
+            isAccessible = true
+            call(GlobalRumMonitor::class.objectInstance, mockRumMonitor, stubSdkCore)
+        }
```
0xnm: Is all this reflection needed? It seems redundant to me. Normally you have `StubSDKCore` instance and it should be used with `RUM.enable` and `Trace.enable` and no need to mock anything.

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,587 @@
...
+
+        mockServer = MockWebServer()
+        mockServer.start()
+    }
+
+    @AfterEach
+    fun `tear down`() {
+        Datadog.stopInstance(stubSdkCore.name)
```
0xnm, line 152: this and `GlobalRumMonitor` call below shouldn't be needed, we can ignore these calls. Am I missing something?

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,587 @@
...
+            .build()
+
+        // Simulate what Apollo interceptor would do - add GraphQL headers
+        val requestWithGraphQLHeaders = Request.Builder()
+            .url(mockServer.url("/graphql"))
+            .addHeader(GraphQLHeaders.DD_GRAPHQL_NAME_HEADER.headerValue, fakeOperationName)
+            .addHeader(GraphQLHeaders.DD_GRAPHQL_TYPE_HEADER.headerValue, fakeOperationType)
+            .addHeader(GraphQLHeaders.DD_GRAPHQL_VARIABLES_HEADER.headerValue, fakeVariables)
+            .addHeader(GraphQLHeaders.DD_GRAPHQL_PAYLOAD_HEADER.headerValue, fakePayload)
+            .build()
```
0xnm: I was rather thinking we could use `DatadogApolloInterceptor` here instead of simulating things. So the tests in this file should just setup `DatadogApolloInterceptor` + `DatadogInterceptor` and use Apollo client to shoot requests and then assert what was received at the `MockWebServer`.

0xnm (commented):
Overall lgtm, I've added some questions

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -533,24 +542,68 @@ internal class RumResourceScope(
...
+                operationName = operationName,
+                variables = variables,
+                payload = payload
             )
         }
 
         return null
     }
 
+    private fun truncateGraphQLPayload(payload: String): String {
```
aleksandr-gringauz: I had a conversation with a certain LLM and it proposed the following solution:

```
/**
 * Returns a prefix of this string whose UTF-8 encoding is <= maxBytes bytes.
 */
fun String.truncateToUtf8Bytes(maxBytes: Int): String {
    val encoder = StandardCharsets.UTF_8.newEncoder()
    val dst = ByteBuffer.allocate(maxBytes)
    val src = CharBuffer.wrap(this)

    // Encode as much as fits. The encoder will not consume a character
    // if doing so would overflow the byte buffer.
    encoder.encode(src, dst, true)

    // src.position() == number of characters successfully encoded
    return substring(0, src.position())
}
```

This solution doesn't require any knowledge of inner workings of UTF-8.

About complexity:
Suppose **N** - number of characters in the input string, **R** - the limit on number of bytes in result.

space: O(**R**) - additional buffer `dst` and a new resulting string are both O(**R**).
time: O(**R**) - filling `dst` and then creating the resulting string are both O(**R**).

You current solution (if I am not mistaken):
space: O(**N + R**) - `payloadBytes` is O(**N**) and the resulting string is O(**R**).
time: O(**N + R**) - creating `payloadBytes` is O(**N**) and filling the resulting string is O(**R**).

This isn't a blocking comment, but I think this is a good info to share. I didn't test this proposed solution a lot, maybe there are some edge cases.

0xnm: Seems like this one is also something like `O(N+R)`? Because first encoder should iterate through the bytes and then `substring` _might_ be linear (hard to say, because it is backed by native code).

aleksandr-gringauz: > Because first encoder should iterate through the bytes 

the [docs](https://cs.android.com/android/platform/superproject/main/+/main:libcore/ojluni/src/main/java/java/nio/charset/CharsetEncoder.java;l=510?q=CharsetEncoder) say "At most {@link Buffer#remaining in.remaining()} characters will be read and at most {@link Buffer#remaining out.remaining()} bytes will be written." So not all of the bytes will be read. Only until `dst` is full.

(I didn't dig deeper than this)

> then substring might be linear

it _should_ copy only the bytes from 0 to `src.position()`. So it is **O(R)**. Linear to the number of bytes in the result. What makes you think that it will want to iterate through the whole buffer? Do you suspect `arraycopy` does something weird?

integrations/dd-sdk-android-apollo/src/main/java/com/datadog/android/apollo/DatadogApolloInterceptor.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        operationVariables?.let {
+            requestBuilder.addHttpHeader(GraphQLHeaders.DD_GRAPHQL_VARIABLES_HEADER.headerValue, operationVariables)
+        }
+
+        if (sendGraphQLPayloads) {
+            val operationPayload = extractPayload(operation, adapters)
+            operationPayload.let {
+                requestBuilder.addHttpHeader(GraphQLHeaders.DD_GRAPHQL_PAYLOAD_HEADER.headerValue, operationPayload)
```
aleksandr-gringauz, line 70: Just to better understand.

Why aren't we truncating payload here like in `RumResourceScope`?

aleksandr-gringauz: understood myself

these headers aren't sent to the backend, they are just used to pass data to `DatadogInterceptor` and are then removed by it

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -192,6 +216,46 @@ open class DatadogInterceptor internal constructor(
         )
     }
 
+    private fun chainWithoutDDHeaders(
+        internalLogger: InternalLogger,
+        originalChain: Interceptor.Chain
+    ): Interceptor.Chain {
+        return if (hasGraphQLHeaders(originalChain.request().headers)) {
+            try {
```
aleksandr-gringauz: Are you sure this try/catch is in the right place? 

The "Try" block contains only a creation of `object : Interceptor.Chain` and does nothing else. It seems like it can't throw any exceptions.

jonathanmos: `build` can throw an illegalStateException and `proceed` can throw IOException

aleksandr-gringauz: > build can throw an illegalStateException and proceed can throw IOException

The way it is now - you can only catch exceptions that happen during execution of construction of `object : Interceptor.Chain by originalChain`. 

As far as I understand it should be like this:

```
object : Interceptor.Chain by originalChain {
    override fun proceed(request: Request): Response {
        return try {
            val cleanedRequest = request.newBuilder().apply {
                removeGraphQLHeaders(this)
            }.build()
            originalChain.proceed(cleanedRequest)
        } catch (IllegalStateException | ....) {
        }
    }
}
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -525,32 +538,79 @@ internal class RumResourceScope(
...
+                variables = variables,
+                payload = payload
             )
         }
 
         return null
     }
 
+    @Suppress("ReturnCount", "SwallowedException")
+    fun String.truncateToUtf8Bytes(maxBytes: Int): String {
```
aleksandr-gringauz: shouldn't this function be private?

aleksandr-gringauz (dismissed):

reliability/single-fit/okhttp/build.gradle.kts
```diff
@@ -21,6 +21,7 @@ plugins {
 
     // Tests
     id("de.mobilej.unmock")
+    id("com.apollographql.apollo") version "4.3.3"
```
0xnm: can you please add this one to the version catalog? so that it can be used through the `alias` without the need to specify the version.

reliability/single-fit/okhttp/build.gradle.kts
```diff
@@ -47,14 +48,29 @@ dependencies {
...
     testImplementation(libs.gson)
+    testImplementation(libs.apolloRuntime)
     unmock(libs.robolectric)
 }
 
+apollo {
+    service("testService") {
+        srcDir("src/test/graphql")
+        packageName.set("com.datadog.android.testgraphql")
+        schemaFiles.from("src/test/graphql/schema.graphqls")
```
0xnm: shouldn't GraphQL source files be in `resources` folder?

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,310 @@
...
+        return TracingInterceptor.Builder(tracedHosts = listOf(mockServer.hostName)).build()
+    }
+
+    @AfterEach
+    fun `tear down`() {
+        GlobalDatadogTracer.clear()
+        mockServer.shutdown()
+    }
+
+    // region graphQL headers
```
0xnm, line 156: apart of verification that GraphQL headers are removed, it is worth also checking that RUM events were created indeed with expected values. You should be able to get the necessary events by calling `stubSdkCore.eventsWritten(Feature.RUM_FEATURE_NAME)`

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,310 @@
...
+    private lateinit var stubSdkCore: StubSDKCore
+    private lateinit var mockServer: MockWebServer
+    private lateinit var apolloClient: ApolloClient
+
+    @BeforeEach
+    fun `set up`(forge: Forge) {
+        mockServer = MockWebServer()
+
+        val fakeApplicationId = forge.anAlphabeticalString()
+        stubSdkCore = StubSDKCore(forge)
```
0xnm: you also need to do a bit of setup like here https://github.com/DataDog/dd-sdk-android/blob/c5026317a833532e40c2928b95a93209505d2b58/reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/HeadBasedSamplingTest.kt#L66-L68 to register Stub instance name and then do `Datadog.stopInstance(stubSdkCore.name)` in tear down and also use `setSdkInstanceName` at the interceptors setup. Otherwise they will try to use default instance name, which is not registered, so effectively interceptors won't do anything and be just pass-through.

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,310 @@
...
+        val requestSent = mockServer.takeRequest()
+        assertThat(requestSent.getHeader(GraphQLHeaders.DD_GRAPHQL_NAME_HEADER.headerValue)).isNull()
+        assertThat(requestSent.getHeader(GraphQLHeaders.DD_GRAPHQL_TYPE_HEADER.headerValue)).isNull()
+        assertThat(requestSent.getHeader(GraphQLHeaders.DD_GRAPHQL_VARIABLES_HEADER.headerValue)).isNull()
+        assertThat(requestSent.getHeader(GraphQLHeaders.DD_GRAPHQL_PAYLOAD_HEADER.headerValue)).isNull()
+    }
+
+    // endregion
+
+    // region TracingInterceptor
```
0xnm: are these tests useful? I would say we don't care about what happens when `TracingInterceptor` is used instead of `DatadogInterceptor`, because it is incorrect setup.

0xnm (changes requested):

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,385 @@
...
+    private fun createDatadogInterceptor(): DatadogInterceptor {
+        return DatadogInterceptor
+            .Builder(tracedHosts = listOf(mockServer.hostName))
+            .setSdkInstanceName(stubSdkCore.name)
+            .build()
+    }
+
+    private fun createTracingInterceptor(): TracingInterceptor {
+        return TracingInterceptor
+            .Builder(tracedHosts = listOf(mockServer.hostName)).build()
```
0xnm: `setSdkInstanceName` should be used here as well

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,385 @@
...
+        apolloClient.query(
+            FakeQuery(
+                userId = fakeUserId,
+                filters = Optional.present(listOf(fakeFilter1, fakeFilter2))
+            )
+        ).execute()
+
+        // Then
+        val requestSent = mockServer.takeRequest()
+        assertThat(requestSent.getHeader(GraphQLHeaders.DD_GRAPHQL_NAME_HEADER.headerValue)).isNull()
```
0xnm: minor: you can simply do `GraphQLHeaders.values().forEach { assertThat(it.headerValue).isNull() }`, this way you won't need to update this code if new values are added

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/ApolloIntegrationTest.kt
```diff
@@ -0,0 +1,385 @@
...
+
+        // When
+        apolloClient.query(
+            FakeQuery(
+                userId = fakeUserId,
+                filters = Optional.present(listOf(fakeFilter1, fakeFilter2))
+            )
+        ).execute()
+
+        Thread.sleep(100)
```
0xnm: is this needed? normally there is no threading in `StubSdkCore`

0xnm (dismissed):

0xnm (approved):

aleksandr-gringauz (approved):

ambushwork (approved):

PR #2846, opened 2025-09-01
[RUM-11016][v3]: Fixing disabled tests

satween:
### What does this PR do?

There are few tests that was disabled during tracing migration. 
This PR enables them back and refactors SdkCoreTest.

### Motivation

We need to have all tests working before next release.

### DIFF
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
@@ -91,7 +91,6 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Assumptions.assumeFalse
 import org.junit.jupiter.api.Assumptions.assumeTrue
 import org.junit.jupiter.api.BeforeEach
-import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.extension.ExtendWith
@@ -3576,7 +3575,6 @@ internal class RumViewScopeTest {
     // region Error
 
     @Test
-    @Disabled // RUM-11016 disabling because it's flaky
     fun `M send event W handleEvent(AddError) on active view`(
         @StringForgery message: String,
         @Forgery source: RumErrorSource,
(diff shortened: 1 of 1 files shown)


reliability/core-it/src/androidTest/kotlin/com/datadog/android/core/integration/tests/SdkCoreTest.kt
```diff
@@ -105,45 +102,36 @@ class SdkCoreTest : MockServerTest() {
...
-        // When
-        testedSdkCore.setUserInfo(fakeUserId, fakeUserName, fakeUserEmail, fakeUserAdditionalProperties)
-
-        // Then
-        val countDownLatch = CountDownLatch(1)
-        var readUserInfo: UserInfo? = null
-        featureSdkCore.getFeature(stubFeature.name)?.withWriteContext { datadogContext, _ ->
-            readUserInfo = datadogContext.userInfo
-            countDownLatch.countDown()
+        testedSdkCore.withImmediateExecutorService {
```
0xnm: I'm not sure we need to do this. The whole point of integration test is to test things as a whole, here we are just stubbing a very essential piece.

Is there a need for the changes in this test file? `getFeature.withWriteContext` must return a  _right_ context with the changes already applied by calling `setUserInfo` above reflected in the `DatadogContext` returned, so it is also a part of the context pipeline verification.

satween: The issue with this test is that setUser/clearUser could be called in random order leading to the case when there is no user info even if it expected. 

I have another option - just to add `Thread.sleep(1)` and it will also reduce the amount of false alarms. 
WDYT?

0xnm: > setUser/clearUser could be called in random order

no, they should guarantee sequential execution, because they delegate to the [context executor which is a single thread always](https://github.com/DataDog/dd-sdk-android/blob/b03a1eb3fa7ce16920521e7fa8d4167a78508862/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L626), this is a property of the execution pipeline.

`withWriteContext` is called on the context executor and that is why there is a `CountdownLatch` as a sync primitive to block test execution until the latch is lifted.

There is probably the issue somewhere under the hood? Is it flaky often?

0xnm (approved):

PR #2859, opened 2025-09-05
Use single source of truth for `os` and `device` properties definitions for Logs and Traces

0xnm:
### What does this PR do?

By relying on the separate files definitions, although they are taken from the same source, they may have different revisions.

This PR instead points to the single location of the data model definition in RUM, guaranteeing  the same model across RUM, Logs and Traces.


features/dd-sdk-android-logs/src/main/json/log/log-schema.json
```diff
@@ -6,10 +6,10 @@
   "description": "Structure holding information about a Log",
   "properties": {
     "device": {
-      "$ref": "_common-schema.json#/properties/device"
+      "$ref": "../../../../../dd-sdk-android-rum/src/main/json/rum/_common-schema.json#/properties/device"
     },
     "os": {
-      "$ref": "_common-schema.json#/properties/os"
+      "$ref": "../../../../../dd-sdk-android-rum/src/main/json/rum/_common-schema.json#/properties/os"
```
satween, line 12: Such approach does not keep the file which was used to generate code. In case if any trouble we can't find source for the generated code. Plus, if you make any changes in the `dd-sdk-android-rum` project (even if you change the branch) it will lead to files re-generation. What is the advantage we gonna have with such configuration compared to the previous one?

features/dd-sdk-android-logs/build.gradle.kts
```diff
@@ -79,7 +79,6 @@ unMock {
 }
 
 apply(from = "generate_log_models.gradle.kts")
-apply(from = "clone_common_schema.gradle.kts")
```
satween, line 82: Don't you want to add some project-level to copy the file from the schema repo?

0xnm: no, we are not going to pull it from repo _here_, we are going to read already pulled in the RUM module.

jonathanmos (approved):

satween (approved):

PR #2861, opened 2025-09-08
RUM-11555: Implement AppStartupTypeDetector

aleksandr-gringauz:
### What does this PR do?

This is the first PR in the series of PRs about app launch improvements.

It implements the algorithm of detecting application start type according to the algorithm described in the RFC [here](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5439422959/RFC+-+App+Launch+Observability+in+Android#Distinguishing-between-SCENARIO_WARM_FIRST_ACTIVITY-and-SCENARIO_WARM_AFTER_ON_DESTROY).

Implemention of TTID calculation and other changes will be in the next PRs.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/api/dd-sdk-android-rum.api
+++ b/features/dd-sdk-android-rum/api/dd-sdk-android-rum.api
@@ -1,3 +1,12 @@
+public final class com/datadog/android/rum/BuildConfig {
+	public static final field BUILD_TYPE Ljava/lang/String;
+	public static final field DEBUG Z
+	public static final field LIBRARY_PACKAGE_NAME Ljava/lang/String;
+	public static final field MIN_SDK I
+	public static final field TARGET_SDK I
+	public fun <init> ()V
+}
+
 public final class com/datadog/android/rum/CloseableExtKt {
 	public static final fun useMonitored (Ljava/io/Closeable;Lcom/datadog/android/api/SdkCore;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
 	public static synthetic fun useMonitored$default (Ljava/io/Closeable;Lcom/datadog/android/api/SdkCore;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Ljava/lang/Object;
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetector.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetector.kt
@@ -24,7 +24,7 @@ internal interface RumAppStartupDetector {
             sdkCore: InternalSdkCore,
             listener: Listener
         ): RumAppStartupDetector {
-            val impl = RumAppStartupDetectorImpl(
+            return RumAppStartupDetectorImpl(
                 application = application,
                 buildSdkVersionProvider = BuildSdkVersionProvider.DEFAULT,
                 appStartupTimeProvider = { sdkCore.appStartTimeNs },
@@ -32,7 +32,6 @@ internal interface RumAppStartupDetector {
                 timeProviderNanos = { System.nanoTime() },
                 listener
             )
-            return impl
         }
     }
 }
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetector.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetector.kt
@@ -1,0 +1,38 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.startup
+
+import android.app.Application
+import com.datadog.android.core.InternalSdkCore
+import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import com.datadog.android.rum.DdRumContentProvider
+
+internal interface RumAppStartupDetector {
+    interface Listener {
+        fun onAppStartupDetected(scenario: RumStartupScenario)
+    }
+
+    fun destroy()
+
+    companion object {
+        fun create(
+            application: Application,
+            sdkCore: InternalSdkCore,
+            listener: Listener
+        ): RumAppStartupDetector {
+            val impl = RumAppStartupDetectorImpl(
+                application = application,
+                buildSdkVersionProvider = BuildSdkVersionProvider.DEFAULT,
+                appStartupTimeProvider = { sdkCore.appStartTimeNs },
+                processImportanceProvider = { DdRumContentProvider.processImportance },
+                timeProviderNanos = { System.nanoTime() },
+                listener
+            )
+            return impl
+        }
+    }
+}
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
@@ -28,7 +28,7 @@ internal class RumAppStartupDetectorImpl(
 
     private var numberOfActivities: Int = 0
     private var isChangingConfigurations: Boolean = false
-    private var isFirstActivityForProcess = true
+    private var isFirstActivityForProcess: Boolean = true
 
     init {
         application.registerActivityLifecycleCallbacks(this)
@@ -107,7 +107,6 @@ internal class RumAppStartupDetectorImpl(
         }
 
         isFirstActivityForProcess = false
-        isChangingConfigurations = false
     }
 
     override fun destroy() {
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
@@ -12,6 +12,7 @@ import android.app.Application
 import android.os.Build
 import android.os.Bundle
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import java.lang.ref.WeakReference
 import kotlin.time.Duration.Companion.nanoseconds
 import kotlin.time.Duration.Companion.seconds
 
@@ -78,26 +79,27 @@ internal class RumAppStartupDetectorImpl(
 
             val gap = (now - processStartTime).nanoseconds
             val hasSavedInstanceStateBundle = savedInstanceState != null
+            val weakActivity = WeakReference(activity)
 
             val scenario = if (isFirstActivityForProcess) {
                 if (!processStartedInForeground || gap > START_GAP_THRESHOLD) {
                     RumStartupScenario.WarmFirstActivity(
                         initialTimeNanos = now,
                         hasSavedInstanceStateBundle = hasSavedInstanceStateBundle,
-                        activity = activity
+                        activity = weakActivity
                     )
                 } else {
                     RumStartupScenario.Cold(
                         initialTimeNanos = processStartTime,
                         hasSavedInstanceStateBundle = hasSavedInstanceStateBundle,
-                        activity = activity
+                        activity = weakActivity
                     )
                 }
             } else {
                 RumStartupScenario.WarmAfterActivityDestroyed(
                     initialTimeNanos = now,
                     hasSavedInstanceStateBundle = hasSavedInstanceStateBundle,
-                    activity = activity
+                    activity = weakActivity
                 )
             }
 
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
@@ -15,8 +15,6 @@ import com.datadog.android.core.internal.system.BuildSdkVersionProvider
 import kotlin.time.Duration.Companion.nanoseconds
 import kotlin.time.Duration.Companion.seconds
 
-private val START_GAP_THRESHOLD = 5.seconds
-
 internal class RumAppStartupDetectorImpl(
     private val application: Application,
     private val buildSdkVersionProvider: BuildSdkVersionProvider,
@@ -111,5 +109,9 @@ internal class RumAppStartupDetectorImpl(
 
     override fun destroy() {
         application.unregisterActivityLifecycleCallbacks(this)
+    }
+
+    companion object {
+        private val START_GAP_THRESHOLD = 5.seconds
     }
 }
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
@@ -1,0 +1,116 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.startup
+
+import android.app.Activity
+import android.app.ActivityManager
+import android.app.Application
+import android.os.Build
+import android.os.Bundle
+import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Duration.Companion.seconds
+
+private val START_GAP_THRESHOLD = 5.seconds
+
+internal class RumAppStartupDetectorImpl(
+    private val application: Application,
+    private val buildSdkVersionProvider: BuildSdkVersionProvider,
+    private val appStartupTimeProvider: () -> Long,
+    private val processImportanceProvider: () -> Int,
+    private val timeProviderNanos: () -> Long,
+    private val listener: RumAppStartupDetector.Listener
+) : RumAppStartupDetector, Application.ActivityLifecycleCallbacks {
+
+    private var numberOfActivities: Int = 0
+    private var isChangingConfigurations: Boolean = false
+    private var isFirstActivityForProcess = true
+
+    init {
+        application.registerActivityLifecycleCallbacks(this)
+    }
+
+    override fun onActivityPreCreated(activity: Activity, savedInstanceState: Bundle?) {
+        if (buildSdkVersionProvider.version >= Build.VERSION_CODES.Q) {
+            onBeforeActivityCreated(activity, savedInstanceState)
+        }
+    }
+
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        if (buildSdkVersionProvider.version < Build.VERSION_CODES.Q) {
+            onBeforeActivityCreated(activity, savedInstanceState)
+        }
+    }
+
+    override fun onActivityDestroyed(activity: Activity) {
+        numberOfActivities--
+        if (numberOfActivities == 0) {
+            isChangingConfigurations = activity.isChangingConfigurations
+        }
+    }
+
+    override fun onActivityPaused(activity: Activity) {
+    }
+
+    override fun onActivityResumed(activity: Activity) {
+    }
+
+    override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle) {
+    }
+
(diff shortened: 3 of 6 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetector.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    fun destroy()
+
+    companion object {
+        fun create(
+            application: Application,
+            sdkCore: InternalSdkCore,
+            listener: Listener
+        ): RumAppStartupDetector {
+            val impl = RumAppStartupDetectorImpl(
```
0xnm: very nit: maybe just `return RumAppStartupDetectorImpl`, unless it is simple to step during debugging.

Or we can move even further: get rid of dedicated class and use `object: RumAppStartupDetector  { ... }`, like here https://github.com/DataDog/dd-sdk-android/blob/6e78b75d121be56a0b63ff660f10e27b82bf015e/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/vitals/JankStatsProvider.kt#L25

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -0,0 +1,116 @@
...
+import android.app.Activity
+import android.app.ActivityManager
+import android.app.Application
+import android.os.Build
+import android.os.Bundle
+import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Duration.Companion.seconds
+
+private val START_GAP_THRESHOLD = 5.seconds
```
0xnm: is there advantage of pulling it outside of `RumAppStartupDetectorImpl`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -0,0 +1,116 @@
...
+    private val buildSdkVersionProvider: BuildSdkVersionProvider,
+    private val appStartupTimeProvider: () -> Long,
+    private val processImportanceProvider: () -> Int,
+    private val timeProviderNanos: () -> Long,
+    private val listener: RumAppStartupDetector.Listener
+) : RumAppStartupDetector, Application.ActivityLifecycleCallbacks {
+
+    private var numberOfActivities: Int = 0
+    private var isChangingConfigurations: Boolean = false
+    private var isFirstActivityForProcess = true
```
0xnm:
```suggestion
    private var isChangingConfigurations: Boolean = false
    private var isFirstActivityForProcess: Boolean = true
```

just for the consistency: either we show type for both, or we don't

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -0,0 +1,116 @@
...
+                    hasSavedInstanceStateBundle = hasSavedInstanceStateBundle,
+                    activity = activity
+                )
+            }
+
+            listener.onAppStartupDetected(scenario)
+        }
+
+        isFirstActivityForProcess = false
+        isChangingConfigurations = false
```
0xnm: are we sure we can set it to `false` here?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
```diff
@@ -0,0 +1,436 @@
...
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class RumAppStartupDetectorImplTest {
+    @Mock
+    private lateinit var application: Application
+
+    @Mock
+    private lateinit var listener: RumAppStartupDetector.Listener
+
+    @IntForgery(min = Build.VERSION_CODES.M, max = Build.VERSION_CODES.BAKLAVA)
```
0xnm: should we limit upper bound?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
```diff
@@ -0,0 +1,436 @@
...
+    private lateinit var buildSdkVersionProvider: BuildSdkVersionProvider
+
+    @Mock
+    private lateinit var activity: Activity
+
+    private var currentTime: Duration = 0.nanoseconds
+
+    @BeforeEach
+    fun `set up`() {
+        currentTime = 0.nanoseconds
```
0xnm: it seems this call is redundant, because we do the same during the property initialization

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
```diff
@@ -0,0 +1,436 @@
...
+            application = application,
+            buildSdkVersionProvider = buildSdkVersionProvider,
+            appStartupTimeProvider = { 0 },
+            processImportanceProvider = { processImportance },
+            timeProviderNanos = { currentTime.inWholeNanoseconds },
+            listener
+        )
+
+        verify(application).registerActivityLifecycleCallbacks(any())
+        verifyNoMoreInteractions(application)
```
0xnm: let's move out this to a dedicated test, maybe? `M call registerActivityLifecycleCallbacks  W init`

otherwise a verification during supposedly `Given` stage is a bit odd

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
```diff
@@ -0,0 +1,436 @@
...
+                initialTimeNanos = currentTime.inWholeNanoseconds,
+                hasSavedInstanceStateBundle = hasSavedInstanceStateBundle3,
+                activity = activity3
+            )
+        )
+
+        verifyNoMoreInteractions(listener)
+    }
+
+    private fun launchColdAndVerify(
```
0xnm: I would propose to not use a method which has "Given + When + Then" inside, and then after its call there is another "When + Then", because it is hard to read the overall test and its flow.

Maybe you can just copy-paste its contents directly where it is used? Yes, it will be some repetition, but it will be easier to read.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    private fun onBeforeActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        numberOfActivities++
+        val now = timeProviderNanos()
+
+        if (numberOfActivities == 1 && !isChangingConfigurations) {
+            val processStartTime = appStartupTimeProvider()
+
+            val processStartedInForeground =
+                processImportanceProvider() == ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
```
ambushwork, line 78: Question, what's the advantage of exposing process Importance with an interface? will it differ from device how the process importance is provided?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetector.kt
aleksandr-gringauz: > maybe just return RumAppStartupDetectorImpl, unless it is simple to step during debugging.

done

> Or we can move even further: get rid of dedicated class and use object: RumAppStartupDetector  { ... }, like here

`RumAppStartupDetectorImpl` still requires some arguments that have no defaults - `Application` and `Listener`, I can't have a `DEFAULT` one.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: Yes.

1. If you put a constant at the top level of the file it will compile into a Java's `static final` inside the `RumAppStartupDetectorImpl` .
2. If you put it inside a `companion object`, the code for this `companion object` will be created.

So technically speaking, if we don't already have a companion object, option 1 is better.

However, IMHO this difference is so negligible that it isn't worth thinking about. And focus on code style. I see that in most places `companion object` is used -> will do the same here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: It works in the following way.

1. We detect a startup only if the `Activity` that is being created is the first one (no other `Activity` exists currently). This is achieved using `numberOfActivities`.
2. Suppose the app has a single `Activity` that is currently visible and the screen rotates. `onDestroy` is called, then `onCreate`. We need to distinguish this from a real startup. This is what `isChangingConfigurations` is for.

To answer your question. This line is probably not needed, I'll remove it. Because `if (numberOfActivities == 1 && !isChangingConfigurations) {` will not pass if it isn't the first activity anyway. And when the last activity is destroyed we will set `isChangingConfigurations = activity.isChangingConfigurations` in `onDestroy`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
jonathanmos: question: why did we choose specifically 5 seconds and not some other interval?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
```diff
@@ -0,0 +1,436 @@
...
+            activity = activity2,
+            hasSavedInstanceStateBundle = hasSavedInstanceStateBundle2
+        )
+
+        // Then
+        verifyNoMoreInteractions(listener)
+    }
+
+    @Test
+    fun `M detect WarmAfterActivityDestroyed W RumAppStartupDetector {1st Cold, activity destroyed another created}`(
```
jonathanmos, line 315: nit: There is some inconsistency. In some test names it is RumAppStartupDetector and in others it is RumAppStartupDetectorImpl

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
```diff
@@ -0,0 +1,33 @@
...
+    data class Cold(
+        override val initialTimeNanos: Long,
+        override val hasSavedInstanceStateBundle: Boolean,
+        override val activity: Activity
+    ) : RumStartupScenario
+
+    data class WarmFirstActivity(
+        override val initialTimeNanos: Long,
+        override val hasSavedInstanceStateBundle: Boolean,
+        override val activity: Activity
```
ambushwork: Why do we need to hold the activity instance inside the data class? will there be some risk of memory leak?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
aleksandr-gringauz: Good question.

I think we actually should, because if we don't `max` will be `Int.MAX_VALUE` and we will in 99% cases test API versions that don't exist.

But when new API appears, we will most likely forget to fix this place.

The only idea I have - make `min = MIN_SDK` and `max = TARGET_SDK`. 

Done.

jonathanmos (approved):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
aleksandr-gringauz: done

`M registerActivityLifecycleCallbacks W RumAppStartupDetectorImpl constructor`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
aleksandr-gringauz: > Why do we need to hold the activity instance inside the data class?

Because later it will be needed to subscribe to the `decorView.onDraw` inside it.

> will there be some risk of memory leak?

I plan to do something like 
```
AppStartupDetector.create(....) {
    scenario.activity.onNextDraw { reportTTID() }
}
```

But it will be in future PRs. I don't plan to hold the instance of `Activity` for too long.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: > what's the advantage of exposing process Importance with an interface? 

I want to mock it in tests. I could mock `DdRumContentProvider` but using an interface seemed cleaner to me here. No strong opinion though.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: No good reason actually. Competitors' thresholds vary from single digit seconds to a minute.

~~My suspicion is that this gap heuristic will be needed very rarely~~. I also plan to log this gap into telemetry to see its distribution to better understand if the threshold is adequate.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
aleksandr-gringauz: fixed, now `RumAppStartupDetector ` everywhere

I had `RumAppStartupDetectorImpl` at first but ran into a maxLine lint rule and started cutting where I could.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
0xnm: > If you put a constant at the top level of the file it will compile into a Java's static final inside the `RumAppStartupDetectorImpl`

I see a different behavior for the following example:

```
private val VALUE = 5

class Something {
    fun doSomething() {
        print(VALUE)
    }
}
```

decompiles to:

```
@Metadata(...)
public final class Something {
   public final void doSomething() {
      int var1 = SomethingKt.access$getVALUE$p();
      System.out.print(var1);
   }
}
// SomethingKt.java
package something;

import kotlin.Metadata;

@Metadata(...)
public final class SomethingKt {
   private static final int VALUE = 5;

   // $FF: synthetic method
   public static final int access$getVALUE$p() {
      return VALUE;
   }
}
```

so a dedicated class is generated.

Dedicated class is generated by the `companion object` as well, but if we add `companion object` in the future to the `RumAppStartupDetectorImpl` and we keep external property, we will have 3 classes, instead of 2 if we just move property to `companion object` right now.

No a blocker for me, but it seems it is a case where we can slightly optimize things by moving this to `companion object`, it  is `private` anyway.

aleksandr-gringauz: > I see a different behavior for the following example

yes, actually you are right, checked myself

changed to `companion object`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
ambushwork: I understand you will deal with the reference of the class carefully, however you will not be able to prevent other developers or even yourself use the reference incorrectly by accident in the future. 
So will it be possible to wrap the activity with `WeakReference` if needed?

aleksandr-gringauz: changed to `WeakReference`

ambushwork: Since you've changed to WeakReference, you need to override the `hashCode` and `isEqualsTo` method.
Because we treat it same if the instance of activity in the `WeakRefence` is equal, but not to check the instance of the wrapper.

And adding unit test of equality on data class is recommended

aleksandr-gringauz: > you need to override the hashCode and isEqualsTo method

I'm not sure if it is possible to do correctly.

Let's say we have a class `class WeakWrapper(val ref: WeakReference<Activity>)` for which we implemented `hashCode` and `equals` somehow. And `equals` actually uses `ref.get`.

Let's consider the situation:
```
val weakActivity = WeakReference(activity)
val obj1 = WeakWrapper(weakActivity)
val obj2 = WeakWrapper(weakActivity)

obj1.equals(obj2) // returns true

....

// activity gets garbage collected, weakActivity.get() becomes null

obj1.equals(obj2) // should return true here, because for 2 objects equals should always return the same value
```

But if we return `true` in the last line, it means that we have to return `true` for any `obj3.equals(obj4)` if the `obj3.ref.get() == null && obj4.ref.get() == null`, even if some time ago those `ref`s inside `obj3` and `obj4` pointed to different activities => we will have to return `true` for every `obj3.equals(obj4)` always => problem.

@ambushwork WDYT?

ambushwork: you don't have to add additional wrapper on top of WeakReference, you ca just override `hashCode` and `equals` on Scenario class, just like this one:
https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ViewTarget.kt#L29

aleksandr-gringauz: discussed and agreed to change `data class` -> `class` without implementing `hashCode` and `equals`.

features/dd-sdk-android-rum/api/dd-sdk-android-rum.api
```diff
@@ -1,3 +1,12 @@
+public final class com/datadog/android/rum/BuildConfig {
```
0xnm, line 1: if it is needed only for tests, is it possible to make it available only in test sources?

aleksandr-gringauz: https://github.com/DataDog/dd-sdk-android/pull/2883 - I have this idea. I didn't find a way to enable `BuildConfig` only for tests. Propose to use `System.getProperty(...)`.

PR #2863, opened 2025-09-08
Split detekt custom rules to handle code point limit

jonathanmos:
### What does this PR do?
We are currently receiving errors of `The incoming YAML document exceeds the limit: 102400 code points` when running custom rules, as soon as the detekt_custom.yml exceeds a certain number of code points. This error is actually not originating in detekt, but rather in snakeYAML that is a detekt dependency. While we might be able to override the limit by using a custom YAML parser, this is complicated by our also running detekt through a shell script. Therefore I'm using a simpler alternative here that will work for both methods that we use for running the rules - splitting the yml file. 

In this PR I split the detekt_custom.yml into three separate parts:
 - a general section
 - safe calls
 - unsafe calls

and these parts are then concatenated together when detekt is run.

### Motivation
Work around the code point limit.

### Additional Notes


detekt_custom_unsafe_calls.yml
0xnm, line 1: are you sure that by splitting the rule into 2 different files it gets executed as a single rule, just once and not twice, like rules with different configs?

jonathanmos: I was able to find this quote about how detekt handles multiple configs: "Currently Detekt supports specifying multiple config files, which will be merged together at runtime with priority in reverse of the order they are specified. If a later file defines a given rule, that definition will supersede any previous definitions of that rule."

0xnm (approved):

ambushwork (approved):

PR #2881, opened 2025-09-15
[RUM-11495]: Supporting baggage header updates

satween:
### What does this PR do?

Adds a logic to properly support W3C's [baggage header.](https://www.w3.org/TR/baggage/). Existing logic just re-write the `baggage` header value every time new item was added.  This PR changes this logic and allows to update header value. SDK will replace item only if item keys are equal.

Note that Baggage implementation is taken from `dd-trace-java` and modified a bit. 

### DIFF
--- a/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/Baggage.java
+++ b/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/Baggage.java
@@ -1,0 +1,112 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.trace.core.propagation;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.datadog.trace.util.PercentEscaper;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.util.HashMap;
+import java.util.Map;
+
+// Source: https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/baggage/BaggagePropagator.java
+public final class Baggage {
+    private static final char PAIR_SEPARATOR = ',';
+    private static final char KEY_VALUE_SEPARATOR = '=';
+    private static final PercentEscaper UTF_ESCAPER = PercentEscaper.create();
+
+    private final HashMap<String, String> values;
+
+    public Baggage() {
+        this.values = new HashMap<>();
+    }
+
+    public boolean isEmpty() {
+        return values.isEmpty();
+    }
+
+    public Baggage put(@NonNull String key, @NonNull String value) {
+        values.put(key, value);
+        return this;
+    }
+
+    @NonNull
+    public Baggage mergeWith(@Nullable Baggage other) {
+        if (other != null) {
+            values.putAll(other.asMap());
+        }
+        return this;
+    }
+
+    @NonNull
+    public String toString() {
+        int processedItems = 0;
+        StringBuilder baggageText = new StringBuilder();
+        for (final Map.Entry<String, String> entry : asMap().entrySet()) {
+            if (processedItems > 0) baggageText.append(',');
+
+            PercentEscaper.Escaped escapedKey = UTF_ESCAPER.escapeKey(entry.getKey());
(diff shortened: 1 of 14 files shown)


features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/Baggage.java
```diff
@@ -0,0 +1,112 @@
...
+public final class Baggage {
+    private static final char PAIR_SEPARATOR = ',';
+    private static final char KEY_VALUE_SEPARATOR = '=';
+    private static final PercentEscaper UTF_ESCAPER = PercentEscaper.create();
+
+    private final HashMap<String, String> values;
+
+    public Baggage() {
+        this.values = new HashMap<>();
+    }
```
0xnm:
```suggestion
    private final HashMap<String, String> values = new HashMap<>();
```

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/Baggage.java
```diff
@@ -0,0 +1,112 @@
...
+            kvSeparatorInd = input.indexOf(KEY_VALUE_SEPARATOR, pairSeparatorInd + 1);
+            pairSeparatorInd = input.indexOf(PAIR_SEPARATOR, pairSeparatorInd + 1);
+            pairSeparatorInd = pairSeparatorInd == -1 ? input.length() : pairSeparatorInd;
+            start = end + 1;
+        }
+
+        return baggage;
+    }
+
+    private static String decode(final String value) {
```
0xnm: minor suggestion, otherwise it is not clear what is decoded, because both input and output are `String`

```suggestion
    private static String urlDecode(final String value) {
```

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/HttpCodec.java
```diff
@@ -363,4 +365,16 @@ static String firstHeaderValue(final String value) {
     int firstComma = value.indexOf(',');
     return firstComma == -1 ? value : value.substring(0, firstComma).trim();
   }
+
+  @NonNull
+  static Baggage composeBaggage(@NonNull DDSpanContext context) {
+    Baggage baggage = new Baggage();
+    final String sessionId = (String) context.getTags().get(RUM_SESSION_ID_KEY);
```
0xnm, line 372: shouldn't we take everything from the `context`? because if it is extracted context and there were other tags there, we won't propagate them

features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/core/propagation/BaggageTest.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.params.provider.MethodSource
+
+internal class BaggageTest {
+
+    @ParameterizedTest
+    @MethodSource("invalidBaggageHeaderSamples")
+    fun `M return empty W from { invalid header value }`(
+        headerValue: String?
+    ) {
+        assertThat(Baggage.from(headerValue).asMap()).isEqualTo(emptyMap<String, String>())
```
0xnm: probably there is `isEmpty` for `Map`?

```suggestion
        assertThat(Baggage.from(headerValue).asMap()).isEmpty()
```

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/HttpCodec.java
satween: Not clear what "everything" we should take from the context. This code just refactor of the existing logic in `DatadogHttpCodec` and `W3CHttpCodec`.
 
Are you suggesting to get `sessionId` directly from context, but not from the `tags` ?

ambushwork (approved):

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/HttpCodec.java
0xnm: Nevermind. I was thinking that this baggage should go into `ExtractedContext` if we are working with the request already having propagation context, but it seems Codecs logic of working with `baggageItems` is about `ot-baggage` (so I was thinking that proper merge should be done in the `inject` as well, preserving items from `DDSpanContext`), but it seems now the work of supporting this is shifted to `mergeBaggage` in `TracingInterceptor`.

0xnm (approved):

PR #2886, opened 2025-09-18
feat: context management

typotter:
# Feature Flags Evaluat Context Implementation

## What does this PR do?

Implements synchronous evaluation context management for feature flag evaluation, inclusion in precompute-assignments request, for exposure logging via in-memory storage and enrichment from the SDK core. 

This establishes the foundation for feature flag targeting by allowing developers to set evaluation context that gets automatically enhanced with device, app, and user data from the Datadog SDK core.

Context, Targeting Data, Attributes; these are all words that mean the set of properties the flag evaluation system can use to determine the effective value of a flag for a particular user or subject (or context). The `targetingKey` is the pillar of the evaluation context, impacting both targeting rules and the selection of traffic for Experiments (such as an A/B test). An inconsistent `targetingKey` for a user leads to inconsistent bucketing.

## Motivation

The feature flags implementation needs context management to:
- Enable proper flag targeting based on user/device attributes
- Pull context data from existing Datadog SDK infrastructure
- Provide a developer-friendly API following Kotlin best practices
- Include context information in network requests for precomputed flag evaluation
- Prepare the foundation for future atomic configuration management

### 🎯 Start Here - Core API Changes
1. **`FlagsProvider.kt`** - See the new `setContext(targetingKey, attributes)` API
2. **`FlagsConfiguration.kt`** - New `useDatadogContextForEvaluations` option
3. **`EvaluationContext.kt`** - New Kotlin DSL for context creation

### 🏗️ Architecture Overview
1. **`FlagsContextManager.kt`** - Central context management with SDK core integration
2. **`DefaultFlagsRepository.kt`** - Context storage and network coordination
3. **`DefaultFlagsNetworkManager.kt`** - Context inclusion in network requests

### 🧪 Understanding Through Tests
1. **`EvaluationContextTest.kt`** - See the new DSL syntax examples
2. **`FlagsContextIntegrationTest.kt`** - End-to-end context flow
3. **`FlagsContextManagerTest.kt`** - SDK core integration examples

### 💡 Understanding the Flow
1. **Start at:** `DatadogFlagsProvider.setContext()` (line 37)
2. **Follow to:** `DefaultFlagsRepository.updateContext()` (line 53)
3. **See enhancement:** `FlagsContextManager.enhanceWithCoreContext()` (line 100+)
4. **Network inclusion:** `DefaultFlagsNetworkManager.buildRequestBody()` (line 178+)

### Basic Context Setting
```kotlin
val provider = FlagsClient.get()

provider.setContext("user123", mapOf(
        "region" to "us-east",
        "tier" to "gold",
        "beta" to true
    ))

val enabled = provider.resolveBooleanValue("new-feature", false)
```

### SDK Core Integration
```kotlin
// Automatically enhanced with:
// - device.type, device.brand, device.model, device.architecture
// - app.version, app.build, env, service, version
// - user.id, user.name, user.email, user.* (additional properties)
```


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
```diff
@@ -35,41 +34,17 @@ object Flags {
      */
     @JvmOverloads
     @JvmStatic
-    @Suppress("UnusedPrivateMember") // todo: remove this when we start using the config
     fun enable(
-        configuration: FlagsConfiguration,
+        configurationBuilder: FlagsConfiguration.Builder,
```
jonathanmos: The standard elsewhere in the SDK is to take a prebuilt configuration. Why do we prefer to take a builder here?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/TestUtils.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.datadog.android.api.context.DeviceInfo
+import com.datadog.android.api.context.UserInfo
+import com.datadog.android.flags.featureflags.EvaluationContext
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+
+/**
+ * Test utilities for creating mock objects and test data for feature flags tests.
+ */
+object TestUtils {
```
jonathanmos: We can do this through forgery factories

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsContextPerformanceTest.kt
```diff
@@ -0,0 +1,191 @@
...
+
+            flagsProvider.setContext(context)
+
+            // Evaluate some flags
+            flagsProvider.resolveBooleanValue("test-flag-$iteration", false)
+            flagsProvider.resolveStringValue("test-string-$iteration", "default")
+        }
+
+        // Force garbage collection
+        System.gc()
```
jonathanmos: This won't actually force garbage collection so the question is whether it would make the test flaky

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/FlagsContextManagerTest.kt
```diff
@@ -0,0 +1,296 @@
...
+                env = "test",
+                clientToken = "test-token"
+            )
+
+        contextManager = FlagsContextManager(featureSdkCore, internalLogger, configuration)
+    }
+
+    @Test
+    fun `initializes with SDK core context when available`() {
+        // Given
```
jonathanmos: For all the tests we should consider forging random values

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -17,16 +21,33 @@ data class FlagsConfiguration(
...
         }
 
-        fun build(): FlagsConfiguration {
+        fun setUseDatadogContextForEvaluations(enabled: Boolean): Builder {
+            flagsConfig = flagsConfig.copy(useDatadogContextForEvaluations = enabled)
+            return this
+        }
+
+        internal fun build(
+            site: String,
```
jonathanmos: Users will get these parameters from here or from datadogContext?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/EvaluationContext.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+        
+        fun build() = EvaluationContext(
+            targetingKey = targetingKey,
+            targetingAttributes = attributes.toMap()
+        )
+    }
+
+    init {
+        require(targetingKey.isNotBlank()) { "Targeting key cannot be blank" }
```
jonathanmos: Will this work if targetingKey is not defined (null)? Also, I'm not sure we should use 'require' because it will throw an exception

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProviderExtensions.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+fun FlagsProvider.addContextAttribute(key: String, value: Any) {
+    val currentContext = getCurrentContext()
+
+    if (currentContext != null) {
+        val updatedAttributes = currentContext.targetingAttributes.toMutableMap()
+        updatedAttributes[key] = value
+        setContext(currentContext.targetingKey, updatedAttributes)
+    } else {
+        throw IllegalStateException(
```
jonathanmos: Do we want to throw an exception here?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagsContextManager.kt
```diff
@@ -0,0 +1,164 @@
...
+        return EvaluationContext.builder(userContext.targetingKey).addAll(coreAttributes).addAll(userContext.targetingAttributes).build()
+    }
+
+    /**
+     * Builds attribute map from SDK core context data.
+     */
+    private fun buildCoreAttributes(coreContext: DatadogContext): Map<String, Any> {
+        return buildMap {
+            // Device attributes
+            put("device.type", coreContext.deviceInfo.deviceType.toString())
```
jonathanmos: nit: We can extract these keys to constants

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultFlagsNetworkManager.kt
```diff
@@ -143,48 +143,70 @@ internal class DefaultFlagsNetworkManager(
...
+                        attributesJson.put(key, jsonArray)
+                    }
+                    is Map<*, *> -> {
+                        // Handle nested objects
+                        val jsonObject = JSONObject(value as Map<String, Any>)
+                        attributesJson.put(key, jsonObject)
+                    }
+                    else -> attributesJson.put(key, value.toString())
+                }
+            }
```
ncreated: question/ It looks like we allow multiple data types in `targeting_attributes`, but the [endpoint spec](https://datadoghq.atlassian.net/wiki/x/bQC7RwE) defines them as `Map<String, String>`:
```
targeting_attributes:
   type: object
   description: Key/value attributes for subject targeting.
   additionalProperties:
      type: string
   example:
      attr1: value1
      companyId: "1"
``` 

The `js-client` [stringifies `targeting_attributes`](https://github.com/DataDog/openfeature-js-client/blob/15e44841ecbc3e666e32c1f12c226f4a5aeb3b10/packages/browser/src/transport/fetchConfiguration.ts#L38-L41), and @sameerank did the same in the Swift client [here](https://github.com/DataDog/dd-sdk-ios/pull/2477/files#diff-a1f685b61ba49c938a8ab03d3216d67d602abaf2b511ca8694cf2020fc52baf0).  

To better support this in a typed language like Swift, we decided to require a `[String: String]` map in the `EvaluationContext` API. This could be relaxed later (by stringifying various types internally), but the plan is to handle that in future iterations.  

Just double-checking here—this implementation doesn’t seem consistent with the spec. Is that intentional?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/EvaluationContext.kt
typotter: Good catch. I've changed signatures to require the targetingKey for the builder and catch the exception in the one place we build the context internally.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/FlagsContextManagerTest.kt
typotter: great tip. thanks!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultFlagsNetworkManager.kt
typotter: not intentional - got carried away a bit; have fixed!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: ddcontext

jonathanmos (approved):

features/dd-sdk-android-flags/api/apiSurface
```diff
@@ -9,12 +9,24 @@ object com.datadog.android.flags.featureflags.FlagsClient
...
+  class Builder
+    constructor(String, com.datadog.android.api.InternalLogger)
+    fun addAttribute(String, String): Builder
+    fun addAttribute(String, Number): Builder
+    fun addAttribute(String, Boolean): Builder
+    fun addAttribute(String, Any): Builder
+    fun addAll(Map<String, Any>): Builder
+    fun build(): EvaluationContext
+  companion object 
+    fun builder(String, com.datadog.android.api.InternalLogger)
```
0xnm, line 29: We shouldn't expose `InternalLogger` in the public API, it is not supposed to be used by the customer.

Is this class `EvaluationContext` should be internal and made public by mistake?

features/dd-sdk-android-flags/api/apiSurface
```diff
@@ -9,12 +9,24 @@ object com.datadog.android.flags.featureflags.FlagsClient
   fun isRegistered(com.datadog.android.api.SdkCore = Datadog.getInstance()): Boolean
   fun get(com.datadog.android.api.SdkCore = Datadog.getInstance()): FlagsProvider
 interface com.datadog.android.flags.featureflags.FlagsProvider
```
0xnm, line 11: What is the reason to have `flags.featureflags` in the package name? Shouldn't it be only `flags` or `featureflags`? Using both seems redundant.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
```diff
@@ -15,9 +14,10 @@ import org.json.JSONObject
 interface FlagsProvider {
     /**
      * Set the context for the provider.
-     * @param newContext The new context to use for the provider.
+     * @param targetingKey The targeting key to use for flag evaluation.
```
0xnm, line 17: We probably need better documentation for that, it is not clear what is `targetingKey` by looking just on the javadoc.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
```diff
@@ -15,9 +14,10 @@ import org.json.JSONObject
...
+     * @param attributes The attributes to use for targeting.
      */
-    fun setContext(newContext: ProviderContext)
+    fun setContext(targetingKey: String, attributes: Map<String, Any>)
```
0xnm, line 20: Usually the type of `attributes` (at least in RUM and Logs) is `Map<String, Any?>` (allowing nullable value). Is there a reason for using non-nullable value type?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
```diff
@@ -25,32 +28,58 @@ internal class DatadogFlagsProvider(
...
         if (flagsRepository is NoOpFlagsRepository) {
             flagsRepository = DefaultFlagsRepository(
-                featureSdkCore = featureSdkCore,
-                executorService = executorService,
-                flagsContext = flagsContext
+                featureSdkCore = featureSdkCore
             )
         }
+
+        // Create orchestrator with network manager and dependencies
```
0xnm, line 40: nit: no need to have comments like that, they don't add any value

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
```diff
@@ -25,32 +28,58 @@ internal class DatadogFlagsProvider(
...
+    override fun setContext(targetingKey: String, attributes: Map<String, Any>) {
+        try {
+            // Create evaluation context with attribute filtering
+            val evaluationContext = EvaluationContext.builder(targetingKey, featureSdkCore.internalLogger)
+                .addAll(attributes)
+                .build()
+
+            // Pass to orchestrator to handle network request and atomic storage
+            flagsOrchestrator.updateEvaluationsForContext(evaluationContext)
+        } catch (e: IllegalArgumentException) {
```
0xnm, line 67: what exactly throws this exception? if it is our code, we shouldn't throw any.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
```diff
@@ -0,0 +1,98 @@
...
+                    // Network request failed, store context with empty flags
+                    flagsRepository.setFlagsAndContext(context, emptyMap())
+
+                    internalLogger.log(
+                        InternalLogger.Level.WARN,
+                        InternalLogger.Target.MAINTAINER,
+                        { "Network request failed for context ${context.targetingKey}, stored with empty flags" }
+                    )
+                }
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
```
0xnm, line 72: This `try-catch` is very wide and should be avoided. Is there a way to narrow it down?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
```diff
@@ -0,0 +1,98 @@
...
+            internalLogger = internalLogger
+        ) {
+            try {
+                internalLogger.log(
+                    InternalLogger.Level.DEBUG,
+                    InternalLogger.Target.MAINTAINER,
+                    { "Processing evaluation context: ${context.targetingKey}" }
+                )
+
+                // Make network request to fetch precomputed flags
```
0xnm, line 47: it is better to remove comments like that, they don't add value

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
```diff
@@ -6,10 +6,4 @@
...
 package com.datadog.android.flags.featureflags.internal.model
 
-internal data class FlagsContext(
-    val applicationId: String?,
-    val clientToken: String,
-    val site: String,
-    val env: String,
-    val targetingKey: String
-)
+internal data class FlagsContext(val applicationId: String?, val clientToken: String, val site: String, val env: String)
```
0xnm, line 9: in which case we can have `null` application ID?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
```diff
@@ -10,55 +10,51 @@ import com.datadog.android.api.InternalLogger
...
-            }
-            DOMAIN_D0G -> {
-                "$customerDomain.ff-cdn.datad0g.com"
-            }
-            in siteConfig -> {
-                val siteConfiguration = siteConfig[site]
-                if (siteConfiguration != null) {
-                    val dc = siteConfiguration.optString("dc", "")
-                    val tld = siteConfiguration.optString("tld", "com")
+    internal fun buildEndpointHost(site: String, customerDomain: String = "preview"): String = when (site) {
```
0xnm, line 15: what is the case for the default value of `customerDomain` argument?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProviderTest.kt
```diff
@@ -413,4 +414,186 @@ internal class DatadogFlagsProviderTest {
...
+            "is_beta" to forge.aBool(),
+            "score" to forge.aDouble()
+        )
+
+        // When
+        testedProvider.setContext(fakeTargetingKey, fakeAttributes)
+
+        // Then
+        // Since orchestrator runs async, we can't directly verify the EvaluationContext
+        // But we can verify that the method completes without throwing
```
0xnm, line 437: then `assertNotThrows` should be used to make the intention to assert it clear

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProviderTest.kt
```diff
@@ -413,4 +414,186 @@ internal class DatadogFlagsProviderTest {
...
+        // Then
+        // addAll() now calls addAttribute() which filters, so warnings should be logged
+        verify(mockInternalLogger, times(3)).log(
+            eq(InternalLogger.Level.WARN),
+            eq(InternalLogger.Target.USER),
+            any(),
+            eq(null),
+            eq(false),
+            eq(null)
+        )
```
0xnm, line 469: We are checking only logging and we don't check that we have proper attributes as a result of filtering.

If underlying code will filter out valid attributes and keep invalid, this test will still pass.

You need to assert that only `validAttributes` are kept.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProviderTest.kt
```diff
@@ -413,4 +414,186 @@ internal class DatadogFlagsProviderTest {
...
+        // Given
+        val fakeTargetingKey = forge.anAlphabeticalString()
+        val emptyAttributes = emptyMap<String, Any>()
+
+        // When
+        testedProvider.setContext(fakeTargetingKey, emptyAttributes)
+
+        // Then
+        // Should complete without errors and not log any warnings
+        verify(mockInternalLogger, times(0)).log(
```
0xnm, line 483: same: we check only logging here, but not checking that attributes didn't make it.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/model/EvaluationContextTest.kt
```diff
@@ -0,0 +1,568 @@
...
+            "user_id" to forge.anInt()
+        )
+
+        // When
+        val context = EvaluationContext(fakeTargetingKey, fakeAttributes)
+
+        // Then
+        assertThat(context.targetingKey).isEqualTo(fakeTargetingKey)
+        assertThat(context.attributes).isEqualTo(fakeAttributes)
+    }
```
0xnm, line 57: is this test useful?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/model/EvaluationContextTest.kt
```diff
@@ -0,0 +1,568 @@
...
+@MockitoSettings(strictness = Strictness.LENIENT)
+internal class EvaluationContextTest {
+
+    @Mock
+    lateinit var mockInternalLogger: InternalLogger
+
+    @BeforeEach
+    fun `set up`(forge: Forge) {
+        ForgeConfigurator.configure(forge)
+    }
```
0xnm, line 38: this can be replace by `@ForgeConfiguration(ForgeConfigurator::class)` annotation on the class

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/model/EvaluationContextTest.kt
```diff
@@ -0,0 +1,568 @@
...
+        // Then
+        assertThat(context.attributes[fakeKey]).isEqualTo(fakeValue)
+        verify(mockInternalLogger, times(0)).log(
+            any<InternalLogger.Level>(),
+            any<InternalLogger.Target>(),
+            any<() -> String>(),
+            any(),
+            any(),
+            any()
+        )
```
0xnm, line 97: `verifyNoInteractions(mockInternalLogger)`, same for similar blocks below

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
typotter: We don't have targeting for null attributes at this time, but it's possible we will in the future. I've refactored the EvaluationContext to an internal and external model where the former will carry only the attributes supported by the flagging system and the external model is more freeform and doesn't use the internal logger

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/model/EvaluationContextTest.kt
typotter: 🧑‍🍳 thanks. much cleaner!

features/dd-sdk-android-flags/api/apiSurface
typotter: The `EvaluationContext` is going to be an external model, so I've split it out now, with an internal model where we can do validation/filtering.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
typotter: Have filled out the docs on the `EvaluationContext` for `targetingKey` and have updated the signature here.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/model/EvaluationContextTest.kt
typotter: ✅

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/model/EvaluationContextTest.kt
typotter: not particularly, I suppose

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
typotter: yes.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
typotter: ApplicationID is a RUM-specific value, so we may not have it until the RUM feature is enabled.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
typotter: the customer domain is for future use. The parameter with default makes it more explicit that `preview` is not intended in perpetuity.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProviderTest.kt
typotter: this test is moved to DatadogEvaluationContextTest

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProviderTest.kt
typotter: moved

features/dd-sdk-android-flags/api/apiSurface
typotter: I do agree we need to settle on a single name here. I think `flags` is sufficient. Because we're two working on a lot of the same files right now, we'll make this rename in this feature branch in an independent pr

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
typotter: agree

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
typotter: I've shifted around the validation here to avoid the exception, thank you

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProviderTest.kt
typotter: hoisted the evaluations manager so we can mock it; makes this a lot less awkward

typotter (commented):
Thank you for the thoughtful review, @0xnm 
I've a WIP drat PR #2896 to address these. It's too late in my day to quite get #2896 ready for review, but it is most of the way.

PR #2895, opened 2025-09-25
Send exposures as batches

jonathanmos:
### What does this PR do?
This is part of the feature flags project.
Writes feature flags exposures to batches and sends them to the endpoint using the core mechanism.

### Motivation

### Additional Notes


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
```diff
@@ -45,6 +48,11 @@ internal class DatadogFlagsProvider(
 
         val precomputeMapper = PrecomputeMapper(featureSdkCore.internalLogger)
 
+        val exposureEventsProcessor = featureSdkCore.getFeature(FLAGS_FEATURE_NAME)
```
aleksandr-gringauz: You can put `processor` directly into `DatadogFlagsProvider` constructor. This way you will avoid a strange situation when `exposureEventsProcessor` is nullable here.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.datadog.android.flags.featureflags.model.EvaluationContext
+import com.datadog.android.flags.internal.model.ExposureEvent
+import com.datadog.android.flags.internal.model.Identifier
+import com.datadog.android.flags.internal.model.Subject
+import com.datadog.android.flags.internal.storage.RecordWriter
+
+internal class ExposureEventsProcessor(
+    private val writer: RecordWriter
+) : EventsProcessor {
+    private val exposuresSentCache = mutableSetOf<String>()
```
aleksandr-gringauz: Will `ExposureEventsProcessor `be used only from one thread? If not, we probably should make `exposuresSentCache` thread-safe.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
```diff
@@ -45,6 +48,11 @@ internal class DatadogFlagsProvider(
...
+            ?.unwrap<FlagsFeature>()?.processor
+
+        exposureEventsProcessor?.clearExposureCache()
```
aleksandr-gringauz: Why is this `clearExposureCache` call necessary? The way I see it it is called just after (or almost after) construction of `EventsProcessor`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
aleksandr-gringauz: **upd**

Looked a bit at the code. If it is non-trivial, leave it as it is now.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
```diff
@@ -0,0 +1,70 @@
...
+) : EventsProcessor {
+
+    private data class CacheKey(
+        val targetingKey: String,
+        val flagName: String,
+        val allocationKey: String,
+        val variationKey: String
+    )
+
+    private val exposuresSentCache = ConcurrentHashMap<CacheKey, Unit>()
```
aleksandr-gringauz: very nit: you can use `Collections.newSetFromMap`.

aleksandr-gringauz (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
```diff
@@ -76,27 +75,51 @@ internal class DatadogFlagsProvider(
 
     override fun resolveBooleanValue(flagKey: String, defaultValue: Boolean): Boolean {
         val precomputedFlag = flagsRepository.getPrecomputedFlag(flagKey)
-        return precomputedFlag?.variationValue?.toBooleanStrictOrNull() ?: defaultValue
+        val result = precomputedFlag?.variationValue?.toBooleanStrictOrNull()
```
typotter: should we extract to an internal generic `getValue` method so that our `writeExposureEvent` needs only one touchpoint?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/FlagsRequestFactory.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.context.DatadogContext
+import com.datadog.android.api.net.Request
+import com.datadog.android.api.net.RequestExecutionContext
+import com.datadog.android.api.net.RequestFactory
+import com.datadog.android.api.storage.RawBatchEvent
+import com.datadog.android.core.internal.utils.join
+import java.util.UUID
+
+internal class FlagsRequestFactory(
```
typotter: We will want to have a `RequestFactory` for the assignments request as well. Could this be the `ExposuresRequestFactory` and I'll add an `AssignmentsRequestFactory`?

typotter (dismissed):
Looks great!
 Just a couple small name/refactor comments.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -10,19 +10,21 @@ package com.datadog.android.flags
...
+) {
     /**
      * A Builder class for a [FlagsConfiguration].
      */
     class Builder {
-        private var flagsConfig = DEFAULT_FEATURE_FLAGS_CONFIG
+        private var enableExposureLogging: Boolean = false
 
         /**
          * Sets whether exposures should be logged to RUM. This is disabled by default.
```
0xnm: just curious: what is the impact if it is enabled? billing?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -10,19 +10,21 @@ package com.datadog.android.flags
  * Describes configuration to be used for the Flags feature.
  * @param enableExposureLogging Log exposure events to RUM.
  */
-data class FlagsConfiguration(val enableExposureLogging: Boolean) {
+data class FlagsConfiguration(
```
0xnm: blocking: constructor should be `internal` - we don't want to give a possibility to create this class explicitly, given we have a builder.

`enableExposureLogging` property here should be `internal` as well, otherwise it is possible to overwrite it in the public API by calling `copy`.

aleksandr-gringauz (approved):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -8,34 +8,73 @@ package com.datadog.android.flags.internal
...
 
 /**
  * An implementation of [Feature] for getting and reporting
  * feature flags to the RUM dashboard.
  */
-internal class FlagsFeature(private val sdkCore: FeatureSdkCore, internal var applicationId: String? = null) :
-    Feature,
+internal class FlagsFeature(
+    private val sdkCore: FeatureSdkCore,
+    internal var applicationId: String? = null,
```
0xnm: Can Flags work without RUM completely?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -8,34 +8,73 @@ package com.datadog.android.flags.internal
...
-        if (featureName == Feature.RUM_FEATURE_NAME) {
-            applicationId = context["application_id"].toString()
+    override val name: String = FLAGS_FEATURE_NAME
+
+    override fun onContextUpdate(
+        featureName: String,
+        context: Map<String, Any?>
+    ) {
+        if (applicationId == null) {
+            applicationId = context[RUM_APPLICATION_ID]?.toString()
```
0xnm, line 57: can read/writes to `applicationId` happen from different threads?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -8,34 +8,73 @@ package com.datadog.android.flags.internal
...
+            internalLogger = sdkCore.internalLogger
+        )
+
+    override val storageConfiguration =
+        FeatureStorageConfiguration(
+            maxItemSize = MAX_ITEM_SIZE,
+            maxItemsPerBatch = MAX_ITEMS_PER_BATCH,
+            maxBatchSize = MAX_BATCH_SIZE,
+            oldBatchThreshold = OLD_BATCH_THRESHOLD
+        )
```
0xnm: nit: it is better to pull properties up, before the methods

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/model/ExposureEvent.kt
```diff
@@ -0,0 +1,55 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.flags.internal.model
+
+import com.google.gson.JsonElement
+import com.google.gson.JsonObject
+
+internal data class ExposureEvent(
```
0xnm, line 14: should we maybe simply create json schema locally and then generate event from there?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/storage/FlagsExposureEvent.kt
```diff
@@ -0,0 +1,32 @@
...
+    val key: String
+)
+
+internal data class ExposureFlag(
+    val key: String
+)
+
+internal data class ExposureVariant(
+    val key: String
+)
```
0xnm: do we need data classes with a single member? or are they going to be extended in the future?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/storage/RecordWriter.kt
```diff
@@ -0,0 +1,20 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.flags.internal.storage
+
+import com.datadog.android.flags.internal.model.ExposureEvent
+
+/**
+ * Will persists the serialized ExposureEvent in the allocated caching location.
```
0xnm:
```suggestion
 * Will persists the serialized [ExposureEvent] in the allocated caching location.
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EventsProcessor.kt
```diff
@@ -0,0 +1,17 @@
...
+package com.datadog.android.flags.internal
+
+import com.datadog.android.flags.featureflags.internal.model.PrecomputedFlag
+import com.datadog.android.flags.featureflags.model.EvaluationContext
+import com.datadog.tools.annotation.NoOpImplementation
+
+@NoOpImplementation
+internal interface EventsProcessor {
+    fun processEvent(flagName: String, context: EvaluationContext, data: PrecomputedFlag)
+    fun clearExposureCache()
```
0xnm: at which point this method is supposed to be called?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
0xnm: effectively we allow this to be unbounded. what is the strategy to clean it up?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -8,34 +8,73 @@ package com.datadog.android.flags.internal
...
+
+    private fun createDataWriter(): RecordWriter {
+        return ExposureEventRecordWriter(sdkCore)
     }
 
     internal companion object {
-        const val FLAGS_FEATURE_NAME = "flags"
+        const val MAX_BATCH_SIZE = 4L * 1024 * 1024 // 4 MB
+        const val OLD_BATCH_THRESHOLD = 18L * 60L * 60L * 1000L // 18 hours
+        const val MAX_ITEMS_PER_BATCH = 50
```
ambushwork, line 77: I see only this one is different from `DEFAULT`, in this case you can use DEFAULT.copy( maxItemsPerBatch = 50) above, and please explain why it's different with a comment.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
jonathanmos: Yes, if RUM exists then we can make use of the applicationId and can report also to RUM, but we don't block if RUM isn't available

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/storage/FlagsExposureEvent.kt
jonathanmos: This was an unused file that I missed in the rebase - I'm deleting it

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
jonathanmos: I've converted this to a bounded lrucache (max 100 elements)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/model/ExposureEvent.kt
jonathanmos: Added a todo to do this in another pr

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
```diff
@@ -0,0 +1,75 @@
...
+internal class ExposureEventsProcessor(
+    private val writer: RecordWriter
+) : EventsProcessor {
+
+    private data class CacheKey(
+        val targetingKey: String,
+        val flagName: String,
+        val allocationKey: String,
+        val variationKey: String
+    )
```
0xnm, line 26: with 4 unrelated properties making a cache key do we expect a high rate of cache miss?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
```diff
@@ -0,0 +1,75 @@
...
+            allocationKey = data.allocationKey,
+            variationKey = data.variationKey
+        )
+
+        @Suppress("UnsafeThirdPartyFunctionCall") // LruCache.get() is safe with non-null key
+        val alreadySent = exposuresSentCache[cacheKey]
+
+        if (alreadySent == null) {
+            @Suppress("UnsafeThirdPartyFunctionCall") // LruCache.put() is safe with non-null key
+            exposuresSentCache.put(cacheKey, true)
```
0xnm: we need to put those into the 3rd party calls file, it seems that declarations are safe actually

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,175 @@
...
+        val configuration1 = FlagsConfiguration(
+            enableExposureLogging = true
+        )
+        val configuration2 = FlagsConfiguration(
+            enableExposureLogging = true
+        )
+
+        // Then
+        assertThat(configuration1).isEqualTo(configuration2)
+        assertThat(configuration1.hashCode()).isEqualTo(configuration2.hashCode())
```
0xnm: do we need this test (and above, and below)? seems like we are checking how `data` classes are working, and this is the property of Kotlin, not the logic we add.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/model/PrecomputedFlagTest.kt
```diff
@@ -0,0 +1,479 @@
...
+        )
+
+        // Then
+        assertThat(precomputedFlag.extraLogging).isEqualTo(complexExtraLogging)
+        assertThat(precomputedFlag.extraLogging.length()).isGreaterThan(0)
+    }
+
+    // endregion
+
+    // region Equality and Hash Code
```
0xnm, line 136: do we need all these tests? seems like they don't add value

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputeMapperTest.kt
```diff
@@ -0,0 +1,597 @@
...
+            )
+        )
+
+        // When
+        val result = testedMapper.map(json)
+
+        // Then
+        assertThat(result).hasSize(1)
+        val flag = result[fakeFlagName]
+        assertThat(flag).isNotNull
```
0xnm:
```suggestion
        checkNotNull(flag)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputeMapperTest.kt
```diff
@@ -0,0 +1,597 @@
...
+
+        // Then
+        assertThat(result).isEmpty()
+
+        val messageCaptor = argumentCaptor<() -> String>()
+        verify(mockInternalLogger).log(
+            eq(InternalLogger.Level.WARN),
+            eq(InternalLogger.Target.MAINTAINER),
+            messageCaptor.capture(),
+            org.mockito.kotlin.any(),
```
0xnm: should go to import

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: You've got it. It's RUM ingestion and eventually, experiments ingestion events that would hit billing. The other impact is that feedback for Feature Flags won't be available in parts of the Datadog Feature Flags & Experimentation app, so the customer loses observability.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
typotter: The number of misses should be low once all of the Flags in an app have been evaluated for the first time.
`targetingKey` is likely to remain stable as it typically references the user. The length of the cache is essentially the number of Flags that are evaluated times the churn in assignment.
The `allocationKey` & `variationKey` only change if the `evaluationContext` changes is reset _and_ the resulting flag assignment changes. The flux in assignment varies depending on the kind of feature flag (i.e. a rollout vs A/B test) but would generally be pretty stable during a user's session in the app. As a result, churn on Allocation and Variation would be very low.

typotter (approved):
LGTM

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
0xnm: We probably need to put more focus on explaining what it is, because it impacts billing, so it is kind of important parameter to change.

features/dd-sdk-android-flags/api/apiSurface
```diff
@@ -1,9 +1,8 @@
 object com.datadog.android.flags.Flags
   fun enable(FlagsConfiguration, com.datadog.android.api.SdkCore = Datadog.getInstance())
 data class com.datadog.android.flags.FlagsConfiguration
-  constructor(Boolean)
   class Builder
-    fun setEnableExposureLogging(Boolean): Builder
+    fun setTrackExposures(Boolean): Builder
```
0xnm: do we need `set`? seems other `trackXXX` public methods around codebase drop it.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -8,21 +8,22 @@ package com.datadog.android.flags
...
     /**
      * A Builder class for a [FlagsConfiguration].
      */
     class Builder {
-        private var flagsConfig = DEFAULT_FEATURE_FLAGS_CONFIG
+        private var trackExposures: Boolean = true
 
         /**
-         * Sets whether exposures should be logged to RUM. This is disabled by default.
+         * Sets whether exposures should be logged to the dedicated exposures intake endpoint. This is enabled by default.
```
0xnm, line 22: so it is not about RUM anymore? Should customer care here more about intake endpoint or cross-linking with other feature in wording?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -73,8 +79,7 @@ internal class DatadogFlagsClient(
      * @return The boolean value of the flag, or the default value if unavailable or unparseable.
      */
     override fun resolveBooleanValue(flagKey: String, defaultValue: Boolean): Boolean {
-        val precomputedFlag = flagsRepository.getPrecomputedFlag(flagKey)
-        return precomputedFlag?.variationValue?.toBooleanStrictOrNull() ?: defaultValue
+        return getValue(flagKey, defaultValue) { it.toBooleanStrictOrNull() }
```
0xnm: is `toBooleanStrictOrNull` usage here is on purpose compared to just `toBooleanOrNull`?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+@ExtendWith(ForgeExtension::class)
+internal class FlagsConfigurationTest {
+
+    @BoolForgery
+    var fakeTrackExposures: Boolean = false
+
+    // region Data Class Tests
+
+    @Test
+    fun `M create FlagsConfiguration W constructor { with all parameters }`() {
```
0xnm: do we need to test this given that constructor is internal? moreover, we just testing the getter here, I don't think it adds any value.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+        val builder = FlagsConfiguration.Builder()
+
+        // When
+        val result = builder.setTrackExposures(false)
+        val configuration = builder.build()
+
+        // Then
+        assertThat(result).isSameAs(builder) // Fluent interface
+        assertThat(configuration.trackExposures).isFalse()
+    }
```
0xnm: you can have it just in a single test by providing random value to `trackExposures`

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+    fun `M set exposure logging disabled W Builder { setTrackExposures false }`() {
+        // Given
+        val builder = FlagsConfiguration.Builder()
+
+        // When
+        val result = builder.setTrackExposures(false)
+        val configuration = builder.build()
+
+        // Then
+        assertThat(result).isSameAs(builder) // Fluent interface
```
0xnm: is it an important property to test?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+    @Test
+    fun `M build configuration with all options W Builder { reverse chaining order }`() {
+        // When
+        val configuration = FlagsConfiguration.Builder()
+            .setTrackExposures(true)
+            .build()
+
+        // Then
+        assertThat(configuration.trackExposures).isTrue()
+    }
```
0xnm: I'm not sure what this test means

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+    @Test
+    fun `M build configuration with all options W Builder { fluent chaining }`() {
+        // When
+        val configuration = FlagsConfiguration.Builder()
+            .setTrackExposures(true)
+            .build()
+
+        // Then
+        assertThat(configuration.trackExposures).isTrue()
+    }
```
0xnm: probably we don't need this test as well

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClientTest.kt
```diff
@@ -69,7 +65,8 @@ internal class DatadogFlagsClientTest {
         testedClient = DatadogFlagsClient(
             featureSdkCore = mockFeatureSdkCore,
             evaluationsManager = mockEvaluationsManager,
-            flagsRepository = mockFlagsRepository
+            flagsRepository = mockFlagsRepository,
+            flagsConfiguration = FlagsConfiguration.Builder().setTrackExposures(true).build()
```
0xnm: if configuration doesn't really matter, you can create a `ForgeryFactory` for it and just generate a random one (overriding the necessary property by doing the `copy` if really needed).

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputeMapperTest.kt
```diff
@@ -0,0 +1,598 @@
...
+                                            // Missing other required fields
+                                        }
+                                    )
+                                }
+                            )
+                        }
+                    )
+                }
+            )
+        }.toString()
```
0xnm: minor: maybe you can just create extension functions to build JSON with less indents in the code

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessorTest.kt
```diff
@@ -0,0 +1,667 @@
...
+            targetingKey = fakeTargetingKey,
+            attributes = mapOf("user_id" to forge.anAlphabeticalString())
+        )
+
+        // When
+        testedProcessor.processEvent(fakeFlagName, fakeContext, fakeFlag)
+        testedProcessor.processEvent(fakeFlagName, fakeContext, fakeFlag) // Duplicate
+
+        // Then
+        verify(mockRecordWriter, times(1)).write(any())
```
0xnm: `times(1)` is already a default mode

```suggestion
        verify(mockRecordWriter).write(any())
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessorTest.kt
```diff
@@ -0,0 +1,667 @@
...
+            attributes = mapOf("user_id" to forge.anAlphabeticalString())
+        )
+        val fakeFlag2 = fakeFlag.copy(allocationKey = forge.anAlphabeticalString())
+
+        // When
+        testedProcessor.processEvent(fakeFlagName, fakeContext, fakeFlag)
+        testedProcessor.processEvent(fakeFlagName, fakeContext, fakeFlag2)
+
+        // Then
+        verify(mockRecordWriter, times(2)).write(any())
```
0xnm: we are not interested in asserting the properties of object being written in all these tests, is it intended?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessorTest.kt
```diff
@@ -0,0 +1,667 @@
...
+        val context1 = EvaluationContext(targetingKey = "user1")
+        val context2 = EvaluationContext(targetingKey = "user2")
+
+        val flag1 = PrecomputedFlag(
+            variationType = forge.anAlphabeticalString(),
+            variationValue = forge.anAlphabeticalString(),
+            doLog = forge.aBool(),
+            allocationKey = "alloc1",
+            variationKey = "var1",
+            extraLogging = org.json.JSONObject(),
```
0xnm:
```suggestion
            extraLogging = JSONObject(),
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/ExposuresRequestFactoryTest.kt
```diff
@@ -0,0 +1,143 @@
...
+        // Then
+        assertThat(request1.id).isNotEqualTo(request2.id)
+        assertThat(request1.headers[RequestFactory.HEADER_REQUEST_ID])
+            .isNotEqualTo(request2.headers[RequestFactory.HEADER_REQUEST_ID])
+        assertThat(request1.id).isEqualTo(request1.headers[RequestFactory.HEADER_REQUEST_ID])
+        assertThat(request2.id).isEqualTo(request2.headers[RequestFactory.HEADER_REQUEST_ID])
+    }
+
+    @Test
+    fun `M create request with valid UUID format W create()`(
```
0xnm: is this important property to check? ID may be not UUID, there is no such requirement probably

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/ExposureEventRecordWriterTest.kt
```diff
@@ -0,0 +1,325 @@
...
+        whenever(mockSdkCore.getFeature(Feature.FLAGS_FEATURE_NAME)).thenReturn(mockFeature)
+        whenever(
+            mockFeature.withWriteContext(
+                any(),
+                any<
+                    (
+                        DatadogContext,
+                        ((EventBatchWriter) -> Unit) -> Unit
+                    ) -> Unit
+                    >()
```
0xnm: can it be just `any()` here or type inference fails?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/ExposureEventRecordWriterTest.kt
```diff
@@ -0,0 +1,325 @@
...
+        // Then
+        val eventCaptor = argumentCaptor<RawBatchEvent>()
+        verify(mockEventBatchWriter).write(
+            event = eventCaptor.capture(),
+            batchMetadata = isNull(),
+            eventType = eq(EventType.DEFAULT)
+        )
+
+        val capturedEvent = eventCaptor.firstValue
+        assert(capturedEvent.data.contentEquals(expectedBytes))
```
0xnm: `assert` is no-op unless there is JVM flag enabled, probably you meant to use `assertThat`?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/ExposureEventRecordWriterTest.kt
```diff
@@ -0,0 +1,325 @@
...
+                    (
+                        DatadogContext,
+                        ((EventBatchWriter) -> Unit) -> Unit
+                    ) -> Unit
+                    >()
+            )
+        )
+            .thenAnswer { invocation ->
+                @Suppress("UNCHECKED_CAST")
+                val callback = invocation.arguments[1] as (DatadogContext, ((EventBatchWriter) -> Unit) -> Unit) -> Unit
```
0xnm: you can `EventWriteScope` instead of `((EventBatchWriter) -> Unit) -> Unit`

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/ExposureEventRecordWriterTest.kt
```diff
@@ -0,0 +1,325 @@
...
+            subject = Subject(fakeSubjectId, fakeAttributes)
+        )
+
+        whenever(mockSdkCore.getFeature(Feature.FLAGS_FEATURE_NAME)).thenReturn(null)
+
+        // When
+        testedWriter.write(fakeEvent)
+
+        // Then
+        verify(mockEventBatchWriter, never()).write(any(), any(), any())
```
0xnm: should it be simply `verifyNoInteractions(mockEventBatchWriter)` here?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
jonathanmos: I don't have access to toBooleanOrNull, but I've added in lowercase to handle True/TRUE etc. cases

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -73,8 +79,7 @@ internal class DatadogFlagsClient(
      * @return The boolean value of the flag, or the default value if unavailable or unparseable.
      */
     override fun resolveBooleanValue(flagKey: String, defaultValue: Boolean): Boolean {
-        val precomputedFlag = flagsRepository.getPrecomputedFlag(flagKey)
-        return precomputedFlag?.variationValue?.toBooleanStrictOrNull() ?: defaultValue
+        return getValue(flagKey, defaultValue) { it.lowercase().toBooleanStrictOrNull() }
```
0xnm: minor: it is always better to specify `Locale` as well, `Locale.US` in our case.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import fr.xgouchet.elmyr.annotation.BoolForgery
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+
+@ExtendWith(ForgeExtension::class)
+internal class FlagsConfigurationTest {
+
+    // region Data Class Tests
```
0xnm: redundant marker

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
```diff
@@ -55,26 +59,25 @@ internal class FlagsTest {
     @StringForgery
     lateinit var fakeEnv: String
 
+    lateinit var fakeConfiguration: FlagsConfiguration
+
     @BeforeEach
-    fun `set up`() {
+    fun `set up`(forge: Forge) {
```
0xnm: minor: declaring `@Forgery lateinit var fakeConfiguration: FlagsConfiguration` should also work

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/ExposureEventRecordWriterTest.kt
```diff
@@ -0,0 +1,300 @@
...
+        whenever(mockSdkCore.getFeature(Feature.FLAGS_FEATURE_NAME)).thenReturn(mockFeature)
+        whenever(
+            mockFeature.withWriteContext(
+                any(),
+                any()
+            )
+        )
+            .thenAnswer { invocation ->
+                @Suppress("UNCHECKED_CAST")
+                val callback = invocation.arguments[1] as (DatadogContext, EventWriteScope) -> Unit
```
0xnm: to avoid `UNCHEKED_CAST` suppression, you can use `invation.getArgument<(DatadogContext, EventWriteScope) -> Unit>(1)` instead

PR #2896, opened 2025-09-25
fix and improve tests, context management

typotter:
What does this PR do?

  This PR refactors the EvaluationContext class and tests in the Flags feature to address comments in #2886 

  Motivation

  This PR addresses review comments from PR #2886:

  1. Public API Issues: Removed InternalLogger exposure and Builder pattern complexity
  2. SDK Consistency: Changed to Map<String, Any?> to match RUM/Logs patterns
  3. API Spec Compliance: Internal normalization ensures Map<String, String> for network calls
  4. Code Quality: Improved documentation, error handling, and test coverage
  5. Architecture: Clean separation between public (flexible) and internal (normalized) APIs

  Additional Notes

  Key Architectural Changes:
  - Public EvaluationContext: Accepts any attribute types for developer convenience
  - Internal DatadogEvaluationContext: Handles validation, normalization, and API compliance
  - Dependency Injection: DatadogFlagsProvider now accepts EvaluationsManager for better testability

  Testing Improvements:
  - Added comprehensive attribute normalization tests
  - Enabled proper unit testing with mock EvaluationsManager
  - Applied SDK testing conventions (Forge, Mockito patterns)


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
```diff
@@ -0,0 +1,124 @@
...
+            return normalized
+        }
+
+        /**
+         * Normalizes a single attribute value to a String.
+         * Returns null for unsupported types or null values, logging appropriate warnings.
+         */
+        private fun normalizeValue(key: String, value: Any?, logger: InternalLogger): String? = when (value) {
+            null -> {
+                logger.log(
```
aleksandr-gringauz: Question.

I see that we support `null` as values in `attributes: Map<String, Any?>`. Then here we log a warning if we encounter `null` and skip it in `normalizeAttributes` function. Why don't we just have `attributes: Map<String, Any>` to express using the type system that we don't actually allow nulls?

Is it a fix for [this](https://github.com/DataDog/dd-sdk-android/pull/2886#discussion_r2375979672) comment?

If in the future we start supporting nulls, we can return to `attributes: Map<String, Any?>`, it will be backward-compatible.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
```diff
@@ -6,6 +6,7 @@
 
 package com.datadog.android.flags.featureflags
 
+import com.datadog.android.flags.featureflags.model.EvaluationContext
 import org.json.JSONObject
 
 /**
```
aleksandr-gringauz, line 12: > An interface for defining Flags providers.

I'm not sure I understand this comment. This interface is called `FlagsProvider`. How does it define other Flags providers?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
```diff
@@ -46,9 +46,9 @@ internal class DefaultFlagsRepository(
         return null
     }
 
-    override fun getEvaluationContext(): EvaluationContext? = atomicState.get()?.context
+    override fun getEvaluationContext(): DatadogEvaluationContext? = atomicState.get()?.context
 
-    override fun getPrecomputedFlagWithContext(key: String): Pair<PrecomputedFlag, EvaluationContext>? {
+    override fun getPrecomputedFlagWithContext(key: String): Pair<PrecomputedFlag, DatadogEvaluationContext>? {
         val state = atomicState.get()
```
aleksandr-gringauz: very very nit

```
val state = atomicState.get() ?: return null
val flag = state.flags[key] ?: return null
return flag to state.context
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
```diff
@@ -89,10 +92,32 @@ object Flags {
             env = env
         )
 
-        return DatadogFlagsProvider(
+        // Create repository
```
0xnm: suggestion: it is better to avoid comments which don't add any value

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
```diff
@@ -14,10 +15,9 @@ import org.json.JSONObject
 interface FlagsProvider {
     /**
      * Set the context for the provider.
-     * @param targetingKey The targeting key to use for flag evaluation.
-     * @param attributes The attributes to use for targeting.
+     * @param context The evaluation context containing targeting key and attributes.
```
0xnm:
```suggestion
     * @param context The [EvaluationContext] containing targeting key and attributes.
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
```diff
@@ -10,68 +10,27 @@ import com.datadog.android.api.InternalLogger
...
-    private val flagsContext: FlagsContext,
-    private var flagsRepository: FlagsRepository = NoOpFlagsRepository()
+    private val evaluationsManager: EvaluationsManager,
+    private val flagsRepository: FlagsRepository
 ) : FlagsProvider {
 
-    private val flagsOrchestrator: EvaluationsManager
+    override fun setContext(context: EvaluationContext) {
+        // Convert public context to internal normalized context
+        val datadogContext = DatadogEvaluationContext.from(context, featureSdkCore.internalLogger)
```
0xnm: minor: we also have `DatadogContext` carrying the basic SDK information, so maybe to make it clear that it is not `DatadogContext`, but `EvaluationContext` converted from the public representation it is better to call this one `datadogEvaluationContext`?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
```diff
@@ -27,71 +27,47 @@ internal class EvaluationsManager(
     private val precomputeMapper: PrecomputeMapper
 ) {
 
+    // region EvaluationsManager
```
0xnm: seems like there is no other region?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
```diff
@@ -27,71 +27,47 @@ internal class EvaluationsManager(
...
-                }
-            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+            val response = flagsNetworkManager.downloadPrecomputedFlags(context)
+            val flagsMap = if (response != null) {
+                precomputeMapper.map(response)
+            } else {
                 internalLogger.log(
-                    InternalLogger.Level.ERROR,
+                    InternalLogger.Level.WARN,
                     InternalLogger.Target.MAINTAINER,
```
0xnm: could this log message be useful for user as well?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
```diff
@@ -0,0 +1,124 @@
...
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.flags.featureflags.model.EvaluationContext
+
+/**
+ * Internal representation of evaluation context that contains only normalized
+ * attributes that Datadog's targeting system can parse and use in evaluation rules.
+ *
+ * This class handles the conversion from the flexible public EvaluationContext
+ * to the strictly typed format required by the API (Map<String, String>).
```
0xnm:
```suggestion
 * This class handles the conversion from the flexible public [EvaluationContext]
 * to the strictly typed format required by the API (`Map<String, String>`).
```

0xnm: If network API can handle only non-nullable values, shouldn't we also use the same in the public API?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
```diff
@@ -0,0 +1,124 @@
...
+ * - Filters out unsupported types and null values
+ * - Provides validation methods for context integrity
+ *
+ * @param targetingKey The identifier used for bucketing and targeting. Must remain consistent
+ *                     for the same entity to ensure consistent flag evaluation across requests.
+ *                     Must not be blank or whitespace-only.
+ * @param attributes Map of normalized attributes where all values are String representations
+ *                   of the original attribute values from the public EvaluationContext.
+ */
+internal class DatadogEvaluationContext internal constructor(
```
0xnm:
```suggestion
internal class DatadogEvaluationContext constructor(
```

Class is already `internal`, so no need for the `internal` modifier on the constructor

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
```diff
@@ -0,0 +1,124 @@
...
+ * - Validates targeting key is not blank/whitespace-only
+ * - Normalizes all attribute values to String type
+ * - Filters out unsupported types and null values
+ * - Provides validation methods for context integrity
+ *
+ * @param targetingKey The identifier used for bucketing and targeting. Must remain consistent
+ *                     for the same entity to ensure consistent flag evaluation across requests.
+ *                     Must not be blank or whitespace-only.
+ * @param attributes Map of normalized attributes where all values are String representations
+ *                   of the original attribute values from the public EvaluationContext.
```
0xnm:
```suggestion
 * @param attributes Map of normalized attributes where all values are [String] representations
 *                   of the original attribute values from the public [EvaluationContext].
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
```diff
@@ -0,0 +1,124 @@
...
+    /** The identifier used for bucketing and targeting in flag evaluation. */
+    val targetingKey: String,
+    val attributes: Map<String, String>
+) {
+
+    /**
+     * Validates the Datadog evaluation context.
+     * @return true if the context is valid, false otherwise.
+     */
+    fun isValid(): Boolean = targetingKey.isNotBlank()
```
0xnm: just out of curiosity: so context with non-empty targeting key and empty attributes is still a valid context?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
```diff
@@ -0,0 +1,124 @@
...
+    companion object {
+        private const val UNSUPPORTED_VALUE_WARNING =
+            "Unsupported attribute value type for key '%s': %s. " +
+                "Only String, Number, and Boolean values are supported. Attribute skipped."
+        private const val NULL_VALUE_WARNING =
+            "Null attribute value for key '%s'. Attribute skipped."
+
+        /**
+         * Creates a DatadogEvaluationContext from a public EvaluationContext,
+         * normalizing all attributes to String key-value pairs that the API can handle.
```
0xnm:
```suggestion
         * Creates a [DatadogEvaluationContext] from a public [EvaluationContext],
         * normalizing all attributes to [String] key-value pairs that the API can handle.
```

It is better to put `[]` around class names to have a link-like navigation when reading docs. See KDoc syntax reference https://kotlinlang.org/docs/kotlin-doc.html.

Same applies to other KDoc entries in the module codebase.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
```diff
@@ -6,4 +6,13 @@
 
 package com.datadog.android.flags.featureflags.internal.model
 
+/**
+ * Internal context containing Datadog configuration needed for feature flag requests.
+ *
+ * @param applicationId The Datadog application ID. May be null when the SDK is not fully initialized
+ *                      or when running in certain test environments where app ID is not required.
+ * @param clientToken The client token for authenticating requests to Datadog
+ * @param site The Datadog site (e.g., US1, EU1) for routing requests
```
0xnm, line 17: is it possible to use `DatadogSite` enum here which we have https://github.com/DataDog/dd-sdk-android/blob/develop/dd-sdk-android-core/src/main/kotlin/com/datadog/android/DatadogSite.kt?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProviderTest.kt
```diff
@@ -418,132 +425,45 @@ internal class DatadogFlagsProviderTest {
...
-        // But we can verify that the method completes without throwing
-        // Integration tests would verify the full flow
+        val contextCaptor = argumentCaptor<DatadogEvaluationContext>()
+        verify(mockEvaluationsManager).updateEvaluationsForContext(contextCaptor.capture())
+
+        val capturedContext = contextCaptor.firstValue
+        assertThat(capturedContext.targetingKey).isEqualTo(fakeTargetingKey)
+        // Verify all attributes were normalized to strings
+        assertThat(capturedContext.attributes).hasSize(4)
+        assertThat(capturedContext.attributes).containsKeys("plan", "region", "user_id", "is_beta")
```
0xnm: minor suggestion: for even more strict validation we can use `containsOnlyKeys`

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerTest.kt
0xnm, line 1: this file is called `.kt.disabled`, so it is not processed as a test source file, but it is still a part of commit. Should it drop `.disabled` extension or not be a part of commit?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContextTest.kt
```diff
@@ -0,0 +1,316 @@
...
+                "numberAttr" to numberValue,
+                "booleanAttr" to booleanValue
+            )
+        )
+
+        // When
+        val datadogContext = DatadogEvaluationContext.from(publicContext, mockInternalLogger)
+
+        // Then
+        assertThat(datadogContext).isNotNull
```
0xnm: minor trick: `checkNotNull(datadogContext)` can be used here (it also throws assertion error) and then you can benefit from smart cast to non-nullable type below.

Same can be done for the tests below.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContextTest.kt
```diff
@@ -0,0 +1,316 @@
...
+        // Then
+        // Only non-null attributes should be present
+        assertThat(datadogContext?.attributes).isEqualTo(
+            mapOf(
+                "validAttr" to validValue
+            )
+        )
+
+        // Warning should be logged for null value
+        verify(mockInternalLogger, times(1)).log(
```
0xnm: `times(1)` is a default validation mode.

```suggestion
        verify(mockInternalLogger).log(
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContextTest.kt
```diff
@@ -0,0 +1,316 @@
...
+        assertThat(result).isNull()
+    }
+
+    @Test
+    fun `M return true W isValid() { valid context }`() {
+        // Given
+        val context = DatadogEvaluationContext("user123", mapOf("plan" to "premium"))
+
+        // When
+        val isValid = context?.isValid()
```
0xnm: `context` is not `null` here

```suggestion
        val isValid = context.isValid()
```

0xnm (changes requested):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
typotter: fair critique. tbh. we threw up a lot of rough code to bootstrap our spike and not everything merged to this feature branch is quite right. The FlagsProvider concept is merged over to the FlagsClient in #2899.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
typotter: thanks!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
typotter: Would it make sense to dual log a warning to the user and an error to the maintainer here? The maintainer will want to know when the SDK internals fail, but to the user it should be mostly transparent and a warning would be appropriate.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContextTest.kt
typotter: thank you! That's really helpful

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultFlagsNetworkManager.kt
```diff
@@ -30,7 +30,7 @@ internal class DefaultFlagsNetworkManager(
 ) : FlagsNetworkManager {
     internal lateinit var callFactory: OkHttpCallFactory
 
-    internal class OkHttpCallFactory(factory: () -> OkHttpClient) : Call.Factory {
+    internal open class OkHttpCallFactory(factory: () -> OkHttpClient) : Call.Factory {
```
jonathanmos: Do we want this to be an open class?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
```diff
@@ -12,6 +12,13 @@ import org.json.JSONObject
 
 internal class EndpointsHelper(private val internalLogger: InternalLogger) {
 
+    /**
+     * Builds the endpoint host for feature flag requests based on the Datadog site.
+     *
```
jonathanmos: nit: since this is an internal class I'm not sure we need this documentation

typotter: ✅

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
typotter: The nullable external API allows for more flexibility for the developer. Similar to RUM accepting nullable values. We filter internally for what is allowed by the network API at the time ( we can extend capabilities going forward)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
typotter: Correct
The targeting key itself can be used in the targeting rules of a Feature Flag. Similarly, an A/B experiment flag may have no targeting rules; it relies on sharding to distribute traffic across the variants, which in turn, relies on the targetingKey (at the present time)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
typotter: Thanks. Appreciate the Kotlin tips and reminders.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
typotter: Yes, this wider-scoping was in response to that previous comment.

Is it better to have an external surface that is congruent with other Features (Logs, RUM) providing flexibility to the developer, or is it better here to limit the external API to just the types that are supported, providing a more concise API for the developer and eliminating the need for filtering warnings?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
typotter: Yes.. good idea.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
typotter: 🆒 I am beginning to really like Kotlin

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultFlagsNetworkManager.kt
typotter: No. I had this open for testing, but not needed any longer.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerTest.kt
typotter: Including this in the commit. Still had some work to do for these tests.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
typotter: Detekt complains about the number of return statements though. Is this a reasonable use of `@Suppress("ReturnCount")` ?

typotter (commented):
Thank you for the reviews! PTAL

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
0xnm: Usually it makes sense to keep both `MAINTAINER` and `USER`, because:

* `MAINTAINER` - log happens only in the debug SDK builds (what is published on Maven is release builds)
* `USER` - log happens in both debug+release SDK builds.

So if it is useful for the `USER` and is actionable somehow from their part (or can help in the investigation with Datadog support), it should be shown to the user.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
aleksandr-gringauz: > Detekt complains about the number of return statements though. Is this a reasonable use of @Suppress("ReturnCount") ?

I didn't expect that... I'm afraid our current detekt rules are a bit too much. We might need to revamp them at some point, a bit annoying in this particular case. You can either have your original code or do `@Suppress("ReturnCount")`, as you want. No need to spend much time.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
aleksandr-gringauz: Ok, I get it, thanks for the clarification.

It's probably better to ask @0xnm opinion.

I would say that `Map<String, Any>` is better so that the user can't accidentally put nulls there that will not really be uploaded. Especially if `null`  support will not be added soon. But I don't have a strong opinion here, up to you. No blockers from me.

0xnm: I think the latter (expressing the limitations in the type system) will be better here, because it will reduce the number of questions/support tickets we receive.

Alignment with other features only makes sense if backend API is the same in terms of handling these "key-value" pairs, but here it is different.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManager.kt
typotter: Thank you for the details on the target mechanisms.

I've combined the log calls into a single ERROR.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/DatadogEvaluationContext.kt
typotter: OK. Thanks everyone for the insight.

Parity with iOS implementation of Flags is going to take priority over ergonomics and consistency. We are going to limit the attributes to Map<String, String> and put the conversion on the devs. This limited API can always be expanded if/when we expose new types or eliminate the need to stringifiy.

typotter (commented):
Thanks again. ptal

aleksandr-gringauz (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
```diff
@@ -6,56 +6,69 @@
...
-     * @param flagKey The name of the key to query.
-     * @param defaultValue The value to return if the key cannot be retrieved.
-     * @return The value of the flag, or the default value if it cannot be retrieved.
+     * Resolves a numeric flag value.
+     *
+     * @param flagKey The unique identifier of the flag to resolve.
+     * @param defaultValue The value to return if the flag cannot be retrieved or parsed.
+     * @return The numeric value of the flag, or the default value if unavailable.
      */
     fun resolveNumberValue(flagKey: String, defaultValue: Number): Number
```
0xnm, line 55: `Number` is rarely used as return type for the functions, because it implies caller should downcast it to a proper type, but probably it is coming from the open standard?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultFlagsNetworkManager.kt
```diff
@@ -72,7 +72,7 @@ internal class DefaultFlagsNetworkManager(
             internalLogger.log(
                 InternalLogger.Level.ERROR,
                 InternalLogger.Target.MAINTAINER,
-                { "Unable to find host for site ${flagsContext.site}; we will retry later." },
+                { "Unable to find host for site ${flagsContext.site.name}; we will retry later." },
```
0xnm: very minor: it is better to use `.siteName` here instead of `.name`, because it can be that build is obfuscated, so `.name` will print enum item name, which will be obfuscated unlike `.siteName` string literal.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
```diff
@@ -6,64 +6,79 @@
...
     }
 
     private val siteConfig: Map<String, JSONObject> = mapOf(
-        "US1" to JSONObject(),
-        "US3" to createSiteConfigObject(dc = "us3"),
-        "US5" to createSiteConfigObject(dc = "us5"),
-        "AP1" to createSiteConfigObject(dc = "ap1"),
-        "AP2" to createSiteConfigObject(dc = "ap2"),
-        "EU" to createSiteConfigObject(tld = "eu")
+        "us1" to JSONObject(),
```
0xnm: should it be `createSiteConfigObject(dc = "us1")` here?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
```diff
@@ -6,64 +6,79 @@
...
+        "us3" to createSiteConfigObject(dc = "us3"),
+        "us5" to createSiteConfigObject(dc = "us5"),
+        "ap1" to createSiteConfigObject(dc = "ap1"),
+        "ap2" to createSiteConfigObject(dc = "ap2"),
+        "eu1" to createSiteConfigObject(tld = "eu"),
+        "us1_fed" to createSiteConfigObject(tld = "com"),
+        "staging" to createSiteConfigObject(tld = "com")
```
0xnm: should it be the same value for config object for these?

0xnm (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsProvider.kt
typotter: yes. The standard requires Number but puts Int/Float as optionals to implement. We will add these specifics as we get closer to adding an OF Provider adapter layer.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultFlagsNetworkManager.kt
typotter: very helpful catch.

typotter: hmm. siteName is internal, so I've changed to use the `intakeEndpoint` for now.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
typotter: Similar to the intake endpoint, the URL of `US1` does not actually have `us1` in it (same with omitting DC from eu). Added a comment.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
typotter: actually, I don't need these entries since they're special cased below. In fact, I don't need this map at all - just an unnecessary extra layer.

0xnm (approved):

PR #2898, opened 2025-09-25
[Feature flags] Add a persistence layer

jonathanmos:
### What does this PR do?
Part of the feature flag development.
• Uses the datastore to persist precomputed flags between sessions.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
@@ -16,9 +16,11 @@ import com.datadog.android.flags.featureflags.internal.DatadogFlagsClient
 import com.datadog.android.flags.featureflags.internal.evaluation.EvaluationsManager
 import com.datadog.android.flags.featureflags.internal.model.FlagsContext
 import com.datadog.android.flags.featureflags.internal.repository.DefaultFlagsRepository
+import com.datadog.android.flags.featureflags.internal.repository.NoOpFlagsRepository
 import com.datadog.android.flags.featureflags.internal.repository.net.DefaultFlagsNetworkManager
 import com.datadog.android.flags.featureflags.internal.repository.net.PrecomputeMapper
 import com.datadog.android.flags.internal.FlagsFeature
+import com.datadog.android.flags.internal.FlagsFeature.Companion.FLAGS_FEATURE_NAME
 
 /**
  * Entry point for the Flags feature.
@@ -100,10 +102,17 @@ object Flags {
             env = env
         )
 
-        val flagsRepository = DefaultFlagsRepository(
-            featureSdkCore = sdkCore,
-            instanceName = "default"
-        )
+        val datastore = (sdkCore as? FeatureSdkCore)?.getFeature(FLAGS_FEATURE_NAME)
+            ?.dataStore
+        val flagsRepository = if (datastore != null) {
+            DefaultFlagsRepository(
+                featureSdkCore = sdkCore,
+                dataStore = datastore,
+                instanceName = "default"
+            )
+        } else {
+            NoOpFlagsRepository()
+        }
 
         val flagsNetworkManager = DefaultFlagsNetworkManager(
             internalLogger = sdkCore.internalLogger,
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
@@ -9,6 +9,7 @@ package com.datadog.android.flags
 import com.datadog.android.Datadog
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
+import com.datadog.android.api.feature.Feature.Companion.FLAGS_FEATURE_NAME
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.flags.featureflags.FlagsClient
@@ -20,7 +21,6 @@ import com.datadog.android.flags.featureflags.internal.repository.NoOpFlagsRepos
 import com.datadog.android.flags.featureflags.internal.repository.net.DefaultFlagsNetworkManager
 import com.datadog.android.flags.featureflags.internal.repository.net.PrecomputeMapper
 import com.datadog.android.flags.internal.FlagsFeature
-import com.datadog.android.flags.internal.FlagsFeature.Companion.FLAGS_FEATURE_NAME
 
 /**
  * Entry point for the Flags feature.
@@ -102,7 +102,7 @@ object Flags {
             env = env
         )
 
-        val datastore = (sdkCore as? FeatureSdkCore)?.getFeature(FLAGS_FEATURE_NAME)
+        val datastore = (sdkCore as FeatureSdkCore).getFeature(FLAGS_FEATURE_NAME)
             ?.dataStore
         val flagsRepository = if (datastore != null) {
             DefaultFlagsRepository(
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
@@ -101,7 +101,8 @@ object Flags {
         )
 
         val flagsRepository = DefaultFlagsRepository(
-            featureSdkCore = sdkCore
+            featureSdkCore = sdkCore,
+            instanceName = "default"
         )
 
         val flagsNetworkManager = DefaultFlagsNetworkManager(
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
@@ -57,6 +57,7 @@ object Flags {
         }
     }
 
+    @Suppress("LongMethod")
     private fun createClient(
         configuration: FlagsConfiguration,
         sdkCore: FeatureSdkCore,
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/PrecomputedFlagConstants.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/PrecomputedFlagConstants.kt
@@ -6,34 +6,38 @@
 
 package com.datadog.android.flags.featureflags.internal.model
 
-/**
- * Constants for PrecomputedFlag field values.
- */
-internal object PrecomputedFlagConstants {
+internal enum class JsonKeys(val value: String) {
+    // EvaluationContext keys
+    EVALUATION_CONTEXT("evaluationContext"),
+    TARGETING_KEY("targetingKey"),
+    ATTRIBUTES("attributes"),
 
-    /**
-     * Valid variation type values.
-     */
-    object VariationType {
-        const val BOOLEAN = "boolean"
-        const val STRING = "string"
-        const val INTEGER = "integer"
-        const val DOUBLE = "double"
-        const val JSON = "json"
+    // FlagsStateEntry keys
+    FLAGS("flags"),
+    LAST_UPDATE_TIMESTAMP("lastUpdateTimestamp"),
 
-        val ALL_VALUES = listOf(BOOLEAN, STRING, INTEGER, DOUBLE, JSON)
-    }
+    // PrecomputedFlag keys
+    VARIATION_TYPE("variationType"),
+    VARIATION_VALUE("variationValue"),
+    DO_LOG("doLog"),
+    ALLOCATION_KEY("allocationKey"),
+    VARIATION_KEY("variationKey"),
+    EXTRA_LOGGING("extraLogging"),
+    REASON("reason")
+}
 
-    /**
-     * Valid reason values.
-     */
-    object Reason {
-        const val DEFAULT = "DEFAULT"
-        const val TARGETING_MATCH = "TARGETING_MATCH"
-        const val RULE_MATCH = "RULE_MATCH"
-        const val PREREQUISITE_FAILED = "PREREQUISITE_FAILED"
-        const val ERROR = "ERROR"
+internal enum class VariationType(val value: String) {
+    BOOLEAN("boolean"),
+    STRING("string"),
+    INTEGER("integer"),
+    DOUBLE("double"),
+    JSON("json")
+}
 
-        val ALL_VALUES = listOf(DEFAULT, TARGETING_MATCH, RULE_MATCH, PREREQUISITE_FAILED, ERROR)
-    }
+internal enum class FlagReason(val value: String) {
+    DEFAULT("DEFAULT"),
+    TARGETING_MATCH("TARGETING_MATCH"),
+    RULE_MATCH("RULE_MATCH"),
+    PREREQUISITE_FAILED("PREREQUISITE_FAILED"),
+    ERROR("ERROR")
 }
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsPersistenceManager.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsPersistenceManager.kt
@@ -16,7 +16,7 @@ import com.datadog.android.flags.featureflags.internal.model.PrecomputedFlag
 import com.datadog.android.flags.featureflags.model.EvaluationContext
 
 internal class FlagsPersistenceManager(
-    private val dataStore: DataStoreHandler?,
+    private val dataStore: DataStoreHandler,
     private val instanceName: String,
     private val internalLogger: InternalLogger,
     onStateLoaded: (FlagsStateEntry?) -> Unit
@@ -41,23 +41,16 @@ internal class FlagsPersistenceManager(
             lastUpdateTimestamp = System.currentTimeMillis()
         )
 
-        dataStore?.setValue(
+        dataStore.setValue(
             key = getFlagsStateKey(),
             data = entry,
             serializer = serializer,
             callback = callback
-        ) ?: run {
-            internalLogger.log(
-                InternalLogger.Level.ERROR,
-                InternalLogger.Target.MAINTAINER,
-                { "Unable to access flags feature datastore for saving state" }
-            )
-            callback?.onFailure()
-        }
+        )
     }
 
     private fun loadFlagsState(onStateLoaded: (FlagsStateEntry?) -> Unit) {
-        dataStore?.value(
+        dataStore.value(
             key = getFlagsStateKey(),
             deserializer = deserializer,
             callback = object : DataStoreReadCallback<FlagsStateEntry> {
@@ -75,14 +68,7 @@ internal class FlagsPersistenceManager(
                     onStateLoaded(null)
                 }
             }
-        ) ?: run {
-            internalLogger.log(
-                InternalLogger.Level.WARN,
-                InternalLogger.Target.MAINTAINER,
-                { "Unable to access flags feature datastore for loading state" }
-            )
-            onStateLoaded(null)
-        }
+        )
     }
 
     companion object {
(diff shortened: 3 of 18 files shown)


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
```diff
@@ -25,6 +25,7 @@ internal class DatadogFlagsProvider(
     private val executorService: ExecutorService,
     private val featureSdkCore: FeatureSdkCore,
     private val flagsContext: FlagsContext,
+    private val instanceName: String,
```
0xnm: How it is supposed to be used? It is never read, just initialized, and the only value we pass is `default`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/PrecomputedFlagConstants.kt
```diff
@@ -6,34 +6,38 @@
...
-    object Reason {
-        const val DEFAULT = "DEFAULT"
-        const val TARGETING_MATCH = "TARGETING_MATCH"
-        const val RULE_MATCH = "RULE_MATCH"
-        const val PREREQUISITE_FAILED = "PREREQUISITE_FAILED"
-        const val ERROR = "ERROR"
+internal enum class VariationType(val value: String) {
+    BOOLEAN("boolean"),
+    STRING("string"),
+    INTEGER("integer"),
```
0xnm, line 32: `Long` values are not supported?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
```diff
@@ -9,35 +9,48 @@ package com.datadog.android.flags.featureflags.internal.repository
...
     override fun setFlagsAndContext(context: EvaluationContext, flags: Map<String, PrecomputedFlag>) {
         val newState = FlagsState(context, flags)
         atomicState.set(newState)
 
-        internalLogger.log(
-            InternalLogger.Level.DEBUG,
-            InternalLogger.Target.MAINTAINER,
-            { "Set flags and context: ${flags.size} flags for context: ${context.targetingKey}" }
-        )
+        if (isInitialized.get()) {
```
aleksandr-gringauz: Just a heads-up. Not sure if it is critical here.

Suppose `setFlagsAndContext` is called before `persistenceManager` finishes reading the first value from disk. In this case `isInitialized` is `false` here and the new values will not be written to disk. 

They will be written to `atomicState`. 2 situations are possible next:
1. They are later overriden by the value that is read by `persistenceManager` from disk. Here we completely lose the new value.
2. They are not overriden, because `atomicState.set(newState)` happened between `atomicState.set(loadedState)` and `isInitialized.set(true)`. Here we don't persist the new value.

I would maybe do something like this:
```
internal class DefaultFlagsRepository(
    private val featureSdkCore: FeatureSdkCore,
    private val instanceName: String,
    private val internalLogger: InternalLogger = featureSdkCore.internalLogger
) : FlagsRepository {
    private data class FlagsState(val context: EvaluationContext, val flags: Map<String, PrecomputedFlag>)
    private val atomicState = AtomicReference<FlagsState?>(null)

    private val persistenceManager = FlagsPersistenceManager(
        dataStore = featureSdkCore.getFeature(FLAGS_FEATURE_NAME)?.dataStore,
        instanceName = instanceName,
        internalLogger = internalLogger
    ) { persistedState ->
        persistedState?.let {
            val loadedState = FlagsState(it.evaluationContext, it.flags)
            atomicState.set(loadedState)
        }
    }

    override fun setFlagsAndContext(context: EvaluationContext, flags: Map<String, PrecomputedFlag>) {
        val newState = FlagsState(context, flags)

        persistenceManager.saveFlagsState(
            context = context,
            flags = flags,
            callback = object : DataStoreWriteCallback {
                override fun onSuccess() {
                    atomicState.set(newState)
                }

                override fun onFailure() {
                }
            }
        )
    }
...
}

```

The executor inside the `DataStoreHandler` is single threaded (afaik) and thus the reading of the first value by `persistenceManager` is guaranteed to happen before the write in the first call of the `setFlagsAndContext`.

But in this case we lose the property that if we call `getPrecomputedFlag` after `setFlagsAndContext` we will get the values we recently set. Not sure this is acceptable here.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializer.kt
```diff
@@ -0,0 +1,116 @@
...
+                e
+            )
+            null
+        }
+    }
+
+    @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+    private fun deserializeEvaluationContext(contextJson: JSONObject): EvaluationContext {
+        val targetingKey = contextJson.getString(JsonKeys.TARGETING_KEY.value)
+        val attributesJson = contextJson.optJSONObject(JsonKeys.ATTRIBUTES.value)
```
0xnm: sometimes it is `getJSONObject`, sometimes it is `optJSONObject`. I understand that former throws exception, but apart of that, what is the consideration to use between the two in general in this class?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateSerializer.kt
```diff
@@ -0,0 +1,76 @@
...
+
+            json.toString()
+        } catch (e: JSONException) {
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                InternalLogger.Target.MAINTAINER,
+                { "Failed to serialize FlagsStateEntry to JSON" },
+                e
+            )
+            "{}"
```
0xnm: is it useful to return empty JSON object and not an empty string?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
```diff
@@ -9,35 +9,48 @@ package com.datadog.android.flags.featureflags.internal.repository
...
 internal class DefaultFlagsRepository(
     private val featureSdkCore: FeatureSdkCore,
+    private val instanceName: String,
     private val internalLogger: InternalLogger = featureSdkCore.internalLogger
 ) : FlagsRepository {
-    // Atomic state - ensures context and flags are always consistent
     private data class FlagsState(val context: EvaluationContext, val flags: Map<String, PrecomputedFlag>)
     private val atomicState = AtomicReference<FlagsState?>(null)
 
+    private val isInitialized = AtomicBoolean(false)
```
aleksandr-gringauz: is `isInitialized` necessary?

is it possible to just check `atomicState` for null?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsPersistenceManager.kt
```diff
@@ -0,0 +1,91 @@
...
+import com.datadog.android.api.storage.datastore.DataStoreHandler
+import com.datadog.android.api.storage.datastore.DataStoreReadCallback
+import com.datadog.android.api.storage.datastore.DataStoreWriteCallback
+import com.datadog.android.core.persistence.datastore.DataStoreContent
+import com.datadog.android.flags.featureflags.internal.model.FlagsStateEntry
+import com.datadog.android.flags.featureflags.internal.model.PrecomputedFlag
+import com.datadog.android.flags.featureflags.model.EvaluationContext
+
+internal class FlagsPersistenceManager(
+    private val dataStore: DataStoreHandler?,
```
aleksandr-gringauz: Why is `dataStore` optional? Is there any situation when `FlagsPersistenceManager` exists without `dataStore`?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
aleksandr-gringauz: **upd**

I think this solution is better:

```
private val persistenceManager = FlagsPersistenceManager(
        dataStore = featureSdkCore.getFeature(FLAGS_FEATURE_NAME)?.dataStore,
        instanceName = instanceName,
        internalLogger = internalLogger
    ) { persistedState ->
        persistedState?.let {
            val loadedState = FlagsState(it.evaluationContext, it.flags)
            atomicState.compareAndSet(null, loadedState)
        }
    }

    override fun setFlagsAndContext(context: EvaluationContext, flags: Map<String, PrecomputedFlag>) {
        val newState = FlagsState(context, flags)
        
        atomicState.set(newState)

        persistenceManager.saveFlagsState(
            context = context,
            flags = flags
        )
    }
```

> But in this case we lose the property that if we call getPrecomputedFlag after setFlagsAndContext we will get the values we recently set. Not sure this is acceptable here.

and this problem goes away

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializer.kt
```diff
@@ -0,0 +1,116 @@
...
+                variationValue = flagJson.getString(JsonKeys.VARIATION_VALUE.value),
+                doLog = flagJson.getBoolean(JsonKeys.DO_LOG.value),
+                allocationKey = flagJson.getString(JsonKeys.ALLOCATION_KEY.value),
+                variationKey = flagJson.getString(JsonKeys.VARIATION_KEY.value),
+                extraLogging = flagJson.getJSONObject(JsonKeys.EXTRA_LOGGING.value),
+                reason = flagJson.getString(JsonKeys.REASON.value)
+            )
+        } catch (e: JSONException) {
+            internalLogger.log(
+                InternalLogger.Level.WARN,
```
ambushwork: is there any reason that it throws warning here but above it throws error?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
jonathanmos: Good catch - I agree that this is the better approach here

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsProvider.kt
jonathanmos: atm default is just a stub value. When we change the api a bit later to receive an instanceName I'll refactor to pass it

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializer.kt
jonathanmos: I'll remove all optJsonObjects so that it's consistent

aleksandr-gringauz (dismissed):

typotter (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
```diff
@@ -92,9 +94,17 @@ object Flags {
             env = env
         )
 
-        val flagsRepository = DefaultFlagsRepository(
-            featureSdkCore = sdkCore
-        )
+        val datastore = (sdkCore as? FeatureSdkCore)?.getFeature(FLAGS_FEATURE_NAME)
```
0xnm: minor: you can use strict cast here

```suggestion
        val datastore = (sdkCore as FeatureSdkCore).getFeature(FLAGS_FEATURE_NAME)
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsPersistenceManager.kt
```diff
@@ -0,0 +1,77 @@
...
+    onStateLoaded: (FlagsStateEntry?) -> Unit
+) {
+    private val serializer = FlagsStateSerializer(internalLogger)
+    private val deserializer = FlagsStateDeserializer(internalLogger)
+
+    init {
+        loadFlagsState(onStateLoaded)
+    }
+
+    private fun getFlagsStateKey(): String = "$FLAGS_STATE_KEY_PREFIX-$instanceName"
```
0xnm:
```suggestion
    private val flagsStateKey: String = "$FLAGS_STATE_KEY_PREFIX-$instanceName"
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializer.kt
```diff
@@ -0,0 +1,125 @@
...
+            val contextJson = json.getJSONObject(JsonKeys.EVALUATION_CONTEXT.value)
+            val evaluationContext = deserializeEvaluationContext(contextJson)
+
+            @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+            val flagsJson = json.getJSONObject(JsonKeys.FLAGS.value)
+            val flags = deserializeFlags(flagsJson)
+
+            @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+            val timestamp = try {
+                json.getLong(JsonKeys.LAST_UPDATE_TIMESTAMP.value)
```
0xnm: maybe `optLong` will be better here? or should we handle missing key here at all?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
```diff
@@ -8,36 +8,43 @@ package com.datadog.android.flags.featureflags.internal.repository
...
     override fun setFlagsAndContext(context: EvaluationContext, flags: Map<String, PrecomputedFlag>) {
         val newState = FlagsState(context, flags)
         atomicState.set(newState)
-
-        internalLogger.log(
-            InternalLogger.Level.DEBUG,
-            InternalLogger.Target.MAINTAINER,
-            { "Set flags and context: ${flags.size} flags for context: ${context.targetingKey}" }
-        )
+        persistenceManager.saveFlagsState(context, flags)
```
0xnm: if save operation failed by some reason, is there any serious impact of having that value still in `atomicState` only?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
```diff
@@ -28,8 +28,16 @@ internal class EndpointsHelper(private val internalLogger: InternalLogger) {
...
+                val dc = try {
+                    siteConfiguration.getString("dc")
+                } catch (_: JSONException) {
+                    ""
+                }
+                val tld = try {
+                    siteConfiguration.getString("tld")
+                } catch (_: JSONException) {
+                    "com"
+                }
```
0xnm: what is the reason for such change? why exception is more preferred?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializerTest.kt
```diff
@@ -0,0 +1,227 @@
...
+                }
+            )
+            put("lastUpdateTimestamp", timestamp)
+        }
+
+        // When
+        val result = testedDeserializer.deserialize(json.toString())
+
+        // Then
+        assertThat(result).isNotNull()
```
0xnm:
```suggestion
        checkNotNull(result)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateSerializerTest.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    @Mock
+    lateinit var mockInternalLogger: InternalLogger
+
+    private lateinit var testedSerializer: FlagsStateSerializer
+
+    @BeforeEach
+    fun `set up`() {
+        testedSerializer = FlagsStateSerializer(mockInternalLogger)
+    }
```
0xnm, line 34:
```suggestion
    private val testedSerialize = FlagsStateSerializer(mockInternalLogger)
```

very minor

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateSerializerTest.kt
```diff
@@ -0,0 +1,132 @@
...
+        val json = JSONObject(serialized)
+        assertThat(json.has("evaluationContext")).isTrue()
+        assertThat(json.has("flags")).isTrue()
+        assertThat(json.has("lastUpdateTimestamp")).isTrue()
+
+        val contextJson = json.getJSONObject("evaluationContext")
+        assertThat(contextJson.getString("targetingKey")).isEqualTo(targetingKey)
+
+        val flagsJson = json.getJSONObject("flags")
+        assertThat(flagsJson.has("flag1")).isTrue()
```
0xnm: maybe for all these JSON assertions it is better to create a dedicated `Assert` object

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/DefaultFlagsRepository.kt
jonathanmos: there won't be persistence between sessions, but we need to store the state in order to continue serving flags. I'll add in the datastore callback here so that we can log an error

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializer.kt
```diff
@@ -0,0 +1,121 @@
...
+                e
+            )
+            null
+        }
+    }
+
+    @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+    private fun deserializeEvaluationContext(contextJson: JSONObject): EvaluationContext {
+        val targetingKey = contextJson.getString(JsonKeys.TARGETING_KEY.value)
+        val attributesJson = try {
```
0xnm: it can be slightly simplified:

```
val attributes = try {
  deserializeAttributes(contextJson.getJSONObject(JsonKeys.ATTRIBUTES.value))
} catch {
  emptyMap()
}
```

and then you don't need to have nullable type in `deserializeAttributes` and do null-check there.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializerTest.kt
0xnm: comment is resolved, but change is not made.

0xnm (dismissed):

aleksandr-gringauz (dismissed):

aleksandr-gringauz (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/persistence/FlagsStateDeserializer.kt
```diff
@@ -0,0 +1,121 @@
...
+            deserializeAttributes(contextJson.getJSONObject(JsonKeys.ATTRIBUTES.value))
+        } catch (_: JSONException) {
+            emptyMap()
+        }
+
+        return EvaluationContext(targetingKey, attributes)
+    }
+
+    @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+    private fun deserializeAttributes(attributesJson: JSONObject?): Map<String, String> {
```
0xnm: argument will never be null though, you will have `JSONException` is there is no such key or it is not of type `JSONObject` in `contextJson.getJSONObject(JsonKeys.ATTRIBUTES.value)` call.

0xnm (dismissed):

PR #2899, opened 2025-09-25
chore: Refactor FlagsProvider and FlagsClient into one

typotter:
## Flags Feature - FlagsClient
_Feature Flagging mobile SDK spike_
stacked on #2896 
[FFL-1116](https://datadoghq.atlassian.net/browse/FFL-1116)

### What does this PR do?

Refactors the WIP `dd-sdk-android-flags` package to consolidate `FlagsProvider` functionality into `FlagsClient` using an interface + companion object pattern. This provides cleaner static access (`FlagsClient.instance()`) while maintaining instance-based flag operations, aligning with mobile-api-spec.md requirements.

### Motivation

The existing dual-class pattern (`FlagsClient` object + `FlagsProvider` interface) was confusing and didn't align with mobile-api-spec.md. The spec requires static access for client instances but instance methods for flag operations. This refactoring simplifies the API to a single `FlagsClient` class that provides both patterns cleanly and adheres to the agreed upon spec for mobile flags SDKs

#### Out of scope
- Creating/getting clients by name is tracked in [FFL-1118](https://datadoghq.atlassian.net/browse/FFL-1118)

 
#### **API Target**

```kotlin
val client = FlagsClient.instance()
client.resolveBooleanValue("flag", false)
```

#### **Key Changes Made**

1. **Consolidated Classes**: Eliminated `FlagsProvider` separate class
2. **Enhanced Interface**: Added companion object with static registry to `FlagsClient`
3. **Simplified API**: Single `FlagsClient.instance()` entry point
4. **Updated Tests**: Rewrote tests to use interface + companion pattern
5. **Documentation**: Added comprehensive KDoc for all methods and companion object

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

[FFL-1118]: https://datadoghq.atlassian.net/browse/FFL-1118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[...]

PR #2900, opened 2025-09-26
Flags.enable and FlagsClient APIs

typotter:
## Flags Feature: Enable method
_Feature Flagging mobile SDK MVP spike_
stacked on #2899
[FFL-1117](https://datadoghq.atlassian.net/browse/FFL-1117)
[FFL-1118](https://datadoghq.atlassian.net/browse/FFL-1118)

### What does this PR do?

This PR implements the complete Flags feature initialization and client management APIs, bringing both `Flags.enable()` and `FlagsClient` up to the mobile API specification. The PR includes configuration management, multi-domain client support, and graceful degradation patterns.

**Key Changes:**

**1. Flags.enable() Configuration**
- Enhanced `FlagsConfiguration` with `customExposureEndpoint`
- Made configuration parameter optional with default configuration for convenience
- Updated `FlagsFeature` to accept and store `FlagsConfiguration` for future client creation
- Configuration is properly passed through to `FlagsContext` and network layer

**2. FlagsClient Factory API**
- Implemented `get()` method with multi-tenant support: `get(name: String = "default", sdkCore: SdkCore)`
  - Auto-creates default client when name == "default"
  - Returns NOPClient for custom names not found (safe fallback)
- Implemented Builder pattern for explicit client creation
  - Returns existing client with warning if duplicate requested
  - Never crashes - always returns valid FlagsClient (real or NOP)
- Renamed `setContext()` to `setEvaluationContext()` for OpenFeature alignment

**3. Multi-Client Management**
- Support for multiple SDK cores and multiple named clients per core
- Lifecycle management with `unregister()` and `clear()` methods

### Motivation

This change fully implements the mobile API specification for Feature Flagging:

```kotlin
// Feature initialization
Flags.enable()                                    // default config, default core
Flags.enable(flagsConfig)                         // custom config, default core
Flags.enable(flagsConfig, core)                   // advanced, custom core

// Client access
val client = FlagsClient.get()                    // get/create default client
val customClient = FlagsClient.get("custom")      // get custom named client (or NOP)
val newClient = FlagsClient.builder("name").build()        // explicitly create named client

// Flag evaluation
client.setEvaluationContext(context)              // set targeting context
val enabled = client.resolveBooleanValue("feature-key", false)
```

[FFL-1117]: https://datadoghq.atlassian.net/browse/FFL-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FFL-1118]: https://datadoghq.atlassian.net/browse/FFL-1118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -10,32 +10,44 @@ import android.content.Context
...
         sdkCore.setContextUpdateReceiver(this)
     }
 
-    override fun onStop() {
-        sdkCore.removeContextUpdateReceiver(this)
-    }
+    // endregion
 
     internal companion object {
+        // TODO move this to com.datadog.android.api.feature.Feature
```
aleksandr-gringauz: Just a reminder. Maybe simply do this now and forget about it.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -10,32 +10,44 @@ import android.content.Context
...
 /**
  * An implementation of [Feature] for getting and reporting
  * feature flags to the RUM dashboard.
  */
-internal class FlagsFeature(private val sdkCore: FeatureSdkCore, internal var applicationId: String? = null) :
+internal class FlagsFeature(private val sdkCore: FeatureSdkCore, internal var flagsConfiguration: FlagsConfiguration) :
     Feature,
     FeatureContextUpdateReceiver {
 
+    internal var applicationId: String? = null
```
aleksandr-gringauz: should it be private?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -10,32 +10,44 @@ import android.content.Context
...
 import com.datadog.android.api.feature.FeatureContextUpdateReceiver
 import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.flags.FlagsConfiguration
 
 /**
  * An implementation of [Feature] for getting and reporting
  * feature flags to the RUM dashboard.
  */
-internal class FlagsFeature(private val sdkCore: FeatureSdkCore, internal var applicationId: String? = null) :
+internal class FlagsFeature(private val sdkCore: FeatureSdkCore, internal var flagsConfiguration: FlagsConfiguration) :
```
aleksandr-gringauz: `internal var flagsConfiguration: FlagsConfiguration` -> `private val flagsConfiguration: FlagsConfiguration`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
+
+            if (flagsFeature == null) {
+                return NoOpFlagsClient(
+                    name = name,
+                    reason = "Flags feature not enabled",
+                    logger = (sdkCore as? FeatureSdkCore)?.internalLogger
+                )
+            }
+
+            val key = Companion.ClientKey(sdkCore, name)
```
aleksandr-gringauz: nit:

You can omit `Companion.` prefix here and in similar places.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
+                if (existingClient != null) {
+                    logWarning(
+                        sdkCore,
+                        "Attempted to create a FlagsClient named '$name', but one already exists. " +
+                            "Existing client will be used, and new configuration will be ignored."
+                    )
+                    return existingClient
+                }
+
+                // Merge configuration (selective override using data class copy)
```
aleksandr-gringauz: > selective override using data class copy

I don't see usage of the `copy` method in code.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
+                val featureConfig = flagsFeature.flagsConfiguration
+                val mergedConfig = FlagsConfiguration(
+                    customExposureEndpoint = explicitCustomEndpoint ?: featureConfig.customExposureEndpoint,
```
aleksandr-gringauz: We have `Flags` object that has `FlagsConfiguration` and we have `FlagsClient` that has a builder that has exactly the same configuration parameters.

`flagsConfiguration` is inside `FlagsFeature` class and can possibly be used there.

So we might end up with a client that has different configuration than `FlagsFeature`. Is this OK?

You can look for inspiration at `RumFeature`/`RumConfiguration` or `SessionReplayFeature`/`SessionReplayConfiguration`.

If you want to have multiple clients with different configurations, I would just drop the configuration that is inside `FlagsFeature`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/Flags.kt
```diff
@@ -7,130 +7,35 @@
...
+     * @param configuration configuration to use with feature flags and experiments. If not provided, the default
      * configuration will be used.
      * @param sdkCore SDK instance to register feature in. If not provided, a default SDK instance
      * will be used.
      */
     @JvmOverloads
     @JvmStatic
     fun enable(
-        @Suppress("UNUSED_PARAMETER") configuration: FlagsConfiguration,
+        configuration: FlagsConfiguration = FlagsConfiguration.defaultConfiguration(),
```
0xnm: normally we always request configuration explicitly when enabling features in SDK, but I am not personally against having this.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -8,21 +8,53 @@ package com.datadog.android.flags
...
- * @param enableExposureLogging Log exposure events to RUM.
+ * @param customExposureEndpoint Custom endpoint URL for uploading exposure events. If null, the default endpoint will be used.
+ * @param customFlagEndpoint Custom endpoint URL for proxying precomputed flag assignment requests. If null, the default endpoint will be used.
+ * @param enableExposureLogging Whether to enable exposure event logging. Defaults to true.
  */
-data class FlagsConfiguration(val enableExposureLogging: Boolean) {
+data class FlagsConfiguration internal constructor(
+    val customExposureEndpoint: String? = null,
+    val customFlagEndpoint: String? = null,
+    val enableExposureLogging: Boolean = true
```
0xnm: these properties should also be `internal`, otherwise they will be visible at `copy` method.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -31,13 +63,26 @@ data class FlagsConfiguration(val enableExposureLogging: Boolean) {
...
-    internal companion object {
+    /**
+     * Companion object that provides factory methods for creating the default [FlagsConfiguration] instance.
+     */
+    companion object {
+        /**
+         * Creates a [FlagsConfiguration] with default settings.
+         * @return a new [FlagsConfiguration] instance with default configuration values.
+         */
+        fun defaultConfiguration(): FlagsConfiguration = DEFAULT_FEATURE_FLAGS_CONFIG
```
0xnm: probably can be simply called `default`, but I am not sure if it a reserved word in Kotlin.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
+     * ```
+     */
+    class Builder {
+        private val name: String
+        private val sdkCore: SdkCore
+
+        // Optional configuration overrides (null = use feature default)
+        private var explicitCustomEndpoint: String? = null
+        private var explicitFlaggingProxy: String? = null
+        private var explicitEnableExposureLogging: Boolean? = null
```
0xnm: I think this is where confusion may be coming: we have also similar params in `FlagsConfiguration`. I think we should have a single place where we set them up.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
+        /**
+         * Creates a builder for a named [FlagsClient].
+         *
+         * The name is used to identify and retrieve this [FlagsClient] later via [get].
+         * Multiple [FlagsClient] instances with different names can coexist within the same [SdkCore].
+         *
+         * @param name the client name. Must be non-empty.
+         * @param sdkCore the SDK instance to associate with this client. Defaults to main instance.
+         */
+        constructor(name: String, sdkCore: SdkCore = Datadog.getInstance()) {
```
0xnm: probably it can be just like the following and then you don't need an explicit constructor above:

```suggestion
        constructor(name: String = DEFAULT_CLIENT_NAME, sdkCore: SdkCore = Datadog.getInstance()) {
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
+         * If a [FlagsClient] with the same name already exists for this [SdkCore]:
+         * - Logs a warning (WARN level, USER target)
+         * - Returns the existing [FlagsClient]
+         *
+         * @return the created [FlagsClient], existing client, or [NoOpFlagsClient].
+         */
+        @Suppress("ReturnCount")
+        fun build(): FlagsClient {
+            // Validate that the Flags feature is enabled
+            val flagsFeature = (sdkCore as? FeatureSdkCore)
```
0xnm: minor: `SdkCore` is always `FeatureSdkCore`, so you can do strict cast:

```suggestion
            val flagsFeature = (sdkCore as FeatureSdkCore)
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
0xnm: I completely agree with this point.

I don't think we need to support the case where clients are using the different parameters from the feature declaration. What can be use-case for that?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
      *
      * This companion manages the registration and retrieval of [FlagsClient] instances
      * per SDK core, ensuring thread-safe access and proper lifecycle management.
      */
     companion object {
-        private val registeredClients: MutableMap<SdkCore, FlagsClient> = mutableMapOf()
+        /**
+         * Composite key for storing multiple named clients per SDK core.
+         */
+        private data class ClientKey(val sdkCore: SdkCore, val name: String)
```
0xnm: minor: it is better to use `sdkCore.name` as a key here, to avoid storing the reference to `SdkCore`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -77,109 +108,351 @@ interface FlagsClient {
...
+                        listOf(InternalLogger.Target.USER, InternalLogger.Target.MAINTAINER),
+                        {
+                            "No FlagsClient with name '$name' exists for SDK instance '${sdkCore.name}'. " +
+                                if (name == DEFAULT_CLIENT_NAME) {
+                                    "Create a client first using: FlagsClient.Builder().build(). "
+                                } else {
+                                    "Create a client first using: FlagsClient.Builder(\"$name\").build(). "
+                                } +
+                                "Returning NoOpFlagsClient which always returns default values."
+                        }
```
0xnm: minor: just to make it easier to read

```suggestion
                if (client == null) {
                    val logger = (sdkCore as FeatureSdkCore).internalLogger
                    val clientBuildStatement = if (name == DEFAULT_CLIENT_NAME) {
                        "Create a client first using: FlagsClient.Builder().build(). "
                    } else {
                        "Create a client first using: FlagsClient.Builder(\"$name\").build(). "
                    }

                    logger.log(
                        InternalLogger.Level.ERROR,
                        listOf(InternalLogger.Target.USER, InternalLogger.Target.MAINTAINER),
                        {
                            "No FlagsClient with name '$name' exists for SDK instance '${sdkCore.name}'. " +
                                " $clientBuildStatement. Returning NoOpFlagsClient which always returns default values."
                        }
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/NoOpFlagsClient.kt
```diff
@@ -15,54 +16,97 @@ import org.json.JSONObject
...
+        return defaultValue
+    }
+
+    /**
+     * Logs a critical error with both USER and MAINTAINER targets.
+     * This ensures visibility in both debug builds (MAINTAINER) and production (USER, if verbosity allows).
+     *
+     * @param operation The operation being called (e.g., "resolveBooleanValue for flag 'my-flag'")
+     */
+    private fun logCriticalError(operation: String) {
```
0xnm, line 101: is it useful to log in the `NoOpFlagsClient` given that we will already print a warning to logcat when this client is created?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/NoOpFlagsClient.kt
```diff
@@ -15,54 +16,97 @@ import org.json.JSONObject
...
+        logger?.log(
+            InternalLogger.Level.ERROR,
+            listOf(InternalLogger.Target.USER, InternalLogger.Target.MAINTAINER),
```
0xnm: normally it should be either `USER` or `MAINTAINER`, mixing both is not useful.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
```diff
@@ -7,19 +7,50 @@
...
     val site: DatadogSite,
-    val env: String
-)
+    val env: String,
+    val customExposureEndpoint: String? = null,
+    val customFlagEndpoint: String? = null
+) {
+    companion object {
+        /**
+         * Creates a FlagsContext from core SDK context and feature configuration.
```
0xnm:
```suggestion
         * Creates a [FlagsContext] from core SDK context and feature configuration.
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
```diff
@@ -7,19 +7,50 @@
...
+         *
+         * @param datadogContext The core SDK context containing authentication and routing info
+         * @param applicationId The application ID (may be null if RUM context not yet available)
+         * @param flagsConfiguration The feature-level configuration from user
+         * @return A complete FlagsContext combining core and feature configuration
```
0xnm:
```suggestion
         * @return A complete [FlagsContext] combining core and feature configuration
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
0xnm: if read and write can happen from different threads, it should be `@Volatile`

0xnm (commented):
I reviewed production files and left some comments.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: I agree as well. I was of the impression that we were cascading options to allow client-specific overrides

We can simplify this for now and look at how to best provide for client-specific overrides are explicitly needed. The FlagsClient may have its own options to set in the future, so we will keep an Builder here to future-proof.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: agreed. The cascading overrides is un-necessary at this point so I have removed the duplication (triplication, really) in the client.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: agreed. There was a call in the spec for
`FlagsClient.Builder(customSdkCore).build()` for a custom core, but `FlagsClient.Builder(sdkCore = customSdkCore)` works as well here.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: thanks!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: I've moved the clients map to the FlagsFeature, so this composite key is un-needed _and_ the clients will unregister `onStop`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: done, thanks!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: added annotation

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/NoOpFlagsClient.kt
typotter: It is intentional for the NoOp Client to make a lot of noise when it is used (in addition to when it is created).
The spec for all the [error messaging also just changed ](https://docs.google.com/document/d/1_VG6soopfKamINSVCdXpVLNLWIB5U-Smb8oBMSWN8WE/edit?tab=t.0)
I will implement those changes in a followup to keep it focused.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/NoOpFlagsClient.kt
typotter: 👍

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: It was looking like #2895 (which has this change) would land before mine. It has not yet, so I have made the change in this PR

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: updated this block heavily

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: copy that

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: appears to work!

typotter (commented):
thank you! ptal

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -8,21 +8,53 @@ package com.datadog.android.flags
 
 /**
  * Describes configuration to be used for the Flags feature.
- * @param enableExposureLogging Log exposure events to RUM.
+ * @param customExposureEndpoint Custom endpoint URL for uploading exposure events. If null, the default endpoint will be used.
+ * @param customFlagEndpoint Custom endpoint URL for proxying precomputed flag assignment requests. If null, the default endpoint will be used.
+ * @param enableExposureLogging Whether to enable exposure event logging. Defaults to true.
  */
-data class FlagsConfiguration(val enableExposureLogging: Boolean) {
+data class FlagsConfiguration(
```
0xnm: constructor should be made `internal`: we want to allow creating the instance of this class only through `Builder`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -78,105 +109,227 @@ interface FlagsClient {
...
          */
-        internal fun unregister(sdkCore: SdkCore = Datadog.getInstance()) {
-            synchronized(registeredClients) {
-                registeredClients.remove(sdkCore)
+        @JvmOverloads
+        @JvmStatic
+        fun get(
+            name: String = DEFAULT_CLIENT_NAME,
+            sdkCore: FeatureSdkCore = Datadog.getInstance() as FeatureSdkCore
+        ): FlagsClient {
```
0xnm: `sdkCore` should be of `SdkCore` type and cast must be made in the function body, because we don't want to expose internal interface in the public API for the customer.

```suggestion
        fun get(
            name: String = DEFAULT_CLIENT_NAME,
            sdkCore: SdkCore = Datadog.getInstance()
        ): FlagsClient {
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -78,105 +109,227 @@ interface FlagsClient {
...
+            val logger = sdkCore.internalLogger
+
+            val flagsFeature = sdkCore.getFeature(FLAGS_FEATURE_NAME)?.unwrap<FlagsFeature>()
+
+            if (flagsFeature == null) {
+                logger.log(
+                    InternalLogger.Level.ERROR,
+                    listOf(InternalLogger.Target.USER, InternalLogger.Target.MAINTAINER),
```
0xnm: it should be either `USER` or `MAINTAINER`, mixing the two is not useful (we have anyway verbose SDK logging set up in the sample app, so log entry will be printed there twice since it is using dev build).

This happens in some other places around the module as well.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -78,105 +109,227 @@ interface FlagsClient {
...
+                    {
+                        "No FlagsClient with name '$name' exists for SDK instance '${sdkCore.name}'. " +
+                            if (name == DEFAULT_CLIENT_NAME) {
+                                "Create a client first using: FlagsClient.Builder().build(). "
+                            } else {
+                                "Create a client first using: FlagsClient.Builder(\"$name\").build(). "
+                            } +
```
0xnm: it is better to pull it out into the dedicated variable for a better readability.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/NoOpFlagsClient.kt
0xnm: Yes, but the target here is `MAINTAINER`, not `USER`. Is this intended?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -7,35 +7,91 @@
...
  */
-internal class FlagsFeature(private val sdkCore: FeatureSdkCore, internal var applicationId: String? = null) :
+internal class FlagsFeature(private val sdkCore: FeatureSdkCore, private val flagsConfiguration: FlagsConfiguration) :
     Feature,
     FeatureContextUpdateReceiver {
+    @Volatile
+    internal var applicationId: String? = null
+        private set
+
+    internal fun getFlagsConfiguration(): FlagsConfiguration = flagsConfiguration
```
0xnm: looks like this is not needed: any caller can read `flagsConfiguration` property directly (which can be made `internal` if needed), there is no additional logic in this method.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationBuilderTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        // When
+        val flagsConfiguration = testedBuilder.build()
+
+        // Then
+        assertThat(flagsConfiguration.customExposureEndpoint).isNull()
+        assertThat(flagsConfiguration.customFlagEndpoint).isNull()
+    }
+
+    @Test
+    fun `M build configuration with custom endpoint W useCustomEndpoint() and build()`(
```
0xnm:
```suggestion
    fun `M build configuration with custom endpoint W useCustomExposureEndpoint() and build()`(
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationBuilderTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        // When
+        val flagsConfiguration = testedBuilder.useCustomExposureEndpoint(customEndpointUrl).build()
+
+        // Then
+        assertThat(flagsConfiguration.customExposureEndpoint).isEqualTo(customEndpointUrl)
+        assertThat(flagsConfiguration.customFlagEndpoint).isNull()
+    }
+
+    @Test
+    fun `M build configuration with flagging proxy W useFlaggingProxy() and build()`(
```
0xnm:
```suggestion
    fun `M build configuration with flagging proxy W useCustomFlagEndpoint() and build()`(
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationBuilderTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        // When
+        val flagsConfiguration = testedBuilder.useCustomFlagEndpoint(flaggingProxyUrl).build()
+
+        // Then
+        assertThat(flagsConfiguration.customExposureEndpoint).isNull()
+        assertThat(flagsConfiguration.customFlagEndpoint).isEqualTo(flaggingProxyUrl)
+    }
+
+    @Test
+    fun `M build configuration with all options W setAll() and build()`(
```
0xnm: there is no `setAll` call

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationBuilderTest.kt
```diff
@@ -0,0 +1,97 @@
...
+            .useCustomFlagEndpoint(flaggingProxyUrl)
+            .build()
+
+        // Then
+        assertThat(flagsConfiguration.customExposureEndpoint).isEqualTo(customEndpointUrl)
+        assertThat(flagsConfiguration.customFlagEndpoint).isEqualTo(flaggingProxyUrl)
+    }
+
+    @Test
+    fun `M build configuration with null custom endpoint W useCustomEndpoint(null) and build()`() {
```
0xnm:
```suggestion
    fun `M build configuration with null custom endpoint W useCustomExposureEndpoint(null) and build()`() {
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationBuilderTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        // When
+        val flagsConfiguration = testedBuilder.useCustomExposureEndpoint(null).build()
+
+        // Then
+        assertThat(flagsConfiguration.customExposureEndpoint).isNull()
+        assertThat(flagsConfiguration.customFlagEndpoint).isNull()
+    }
+
+    @Test
+    fun `M build configuration with null flagging proxy W useFlaggingProxy(null) and build()`() {
```
0xnm:
```suggestion
    fun `M build configuration with null flagging proxy W useCustomFlagEndpoint(null) and build()`() {
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,202 @@
...
+import fr.xgouchet.elmyr.annotation.StringForgery
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+
+@ExtendWith(ForgeExtension::class)
+internal class FlagsConfigurationTest {
+
+    // region defaultConfiguration()
```
0xnm:
```suggestion
    // region default()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,202 @@
...
+
+        val config2 = FlagsConfiguration.Builder()
+            .useCustomExposureEndpoint(fakeExposureEndpoint2)
+            .build()
+
+        // Then
+        assertThat(config1).isNotEqualTo(config2)
+    }
+
+    // endregion
```
0xnm: I don't think we need all this, is it useful to test something we don't control?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -0,0 +1,202 @@
...
+        val config2 = FlagsConfiguration.Builder()
+            .useCustomExposureEndpoint(fakeExposureEndpoint)
+            .setExposureLoggingEnabled(false)
+            .build()
+
+        // Then
+        assertThat(config1).isNotEqualTo(config2)
+    }
+
+    // endregion
```
0xnm: is this test useful? seems like once again just checking `data` class behaviour properties, we don't control this.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
```diff
@@ -86,106 +47,128 @@ internal class FlagsTest {
...
+        argumentCaptor<FlagsFeature> {
+            verify(mockSdkCore).registerFeature(capture())
+            assertThat(lastValue.getFlagsConfiguration().customExposureEndpoint).isEqualTo(fakeCustomEndpoint)
         }
     }
 
     @Test
-    fun `M log error W enable() { missing clientToken and site }`() {
+    fun `M register FlagsFeature with flagging proxy W enable() { flagging proxy }`(
+        @StringForgery(regex = "https://[a-z]+\\.com(/[a-z]+)+") fakeFlaggingProxy: String
```
0xnm: just for the naming consistency:

```suggestion
        @StringForgery(regex = "https://[a-z]+\\.com(/[a-z]+)+") fakeCustomFlagEndpoint: String
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
```diff
@@ -51,11 +57,61 @@ internal class FlagsClientTest {
         whenever(mockSecondSdkCore.internalLogger).thenReturn(mockInternalLogger)
         whenever(mockSdkCore.name).thenReturn("test-sdk")
         whenever(mockSecondSdkCore.name).thenReturn("test-sdk-2")
+
+        // Enable the Flags Feature for first SDK core
+        Flags.enable(sdkCore = mockSdkCore)
+
+        // Capture the registered feature and mock getFeature() to return a FeatureScope wrapping it
+        argumentCaptor<com.datadog.android.api.feature.Feature> {
```
0xnm: must be in the `import`

```suggestion
        argumentCaptor<Feature> {
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
```diff
@@ -51,11 +57,61 @@ internal class FlagsClientTest {
...
+            verify(mockSdkCore).registerFeature(capture())
+            val capturedFeature = lastValue as com.datadog.android.flags.internal.FlagsFeature
+            val mockFeatureScope = mock<com.datadog.android.api.feature.FeatureScope>()
```
0xnm:
```suggestion
            val capturedFeature = lastValue as FlagsFeature
            val mockFeatureScope = mock<FeatureScope>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
```diff
@@ -51,11 +57,61 @@ internal class FlagsClientTest {
...
+            whenever(
+                mockFeatureScope.unwrap<com.datadog.android.flags.internal.FlagsFeature>()
+            ).thenReturn(capturedFeature)
```
0xnm, line 74:
```suggestion
            whenever(
                mockFeatureScope.unwrap<FlagsFeature>()
            ).thenReturn(capturedFeature)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
```diff
@@ -230,11 +259,138 @@ internal class FlagsClientTest {
...
+    @Test
+    fun `M return existing client W Builder#build() {client already exists with default name}`() {
+        // Given
+        registerClient(mockFlagsClient, mockSdkCore, "default")
+
+        // When - second build with same name
+        val client = FlagsClient.Builder(sdkCore = mockSdkCore).build()
+
+        // Then - should return existing client, NOT NOPClient
+        assertThat(client).isEqualTo(mockFlagsClient)
```
0xnm, line 286: The calls and assertions are already self-descriptive

```suggestion
        // Given
        registerClient(mockFlagsClient, mockSdkCore, "default")

        // When
        val client = FlagsClient.Builder(sdkCore = mockSdkCore).build()

        // Then
        assertThat(client).isEqualTo(mockFlagsClient)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContextTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    @StringForgery
+    lateinit var fakeSiteName: String
+
+    @StringForgery
+    lateinit var fakeEnv: String
+
+    @Test
+    fun `M create FlagsContext with all parameters W create() { complete configuration }`(
+        @StringForgery(regex = "https://[a-z]+\\.com(/[a-z]+)+") fakeCustomEndpoint: String,
+        @StringForgery(regex = "https://[a-z]+\\.com(/[a-z]+)+") fakeProxyUrl: String
```
0xnm:
```suggestion
        @StringForgery(regex = "https://[a-z]+\\.com(/[a-z]+)+") fakeExposureEndpoint: String,
        @StringForgery(regex = "https://[a-z]+\\.com(/[a-z]+)+") fakeFlagEndpoint: String
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/net/EndpointsHelperTest.kt
```diff
@@ -159,4 +164,78 @@ internal class EndpointsHelperTest {
     }
 
     // endregion
+
+    // region getFlaggingEndpoint
+
+    @Test
+    fun `M return proxy URL W getFlaggingEndpoint() {flagging proxy configured}`(
```
0xnm:
```suggestion
    fun `M return proxy URL W getFlaggingEndpoint() {flagging endpoint configured}`(
```

for the naming consistency it is better to use `endpoint` instead of `proxy`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -7,35 +7,91 @@
...
+     */
+    internal val registeredClients: MutableMap<String, FlagsClient> = mutableMapOf()
+
+    /**
+     * Gets a registered [FlagsClient] by name.
+     *
+     * @param name The client name to lookup
+     * @return The registered [FlagsClient], or null if not found
+     */
+    internal fun getClient(name: String): FlagsClient? = registeredClients[name]
```
aleksandr-gringauz: synchronized is missing here

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -7,35 +7,91 @@
...
+    internal var applicationId: String? = null
+        private set
+
+    internal fun getFlagsConfiguration(): FlagsConfiguration = flagsConfiguration
+
+    /**
+     * Registry of [FlagsClient] instances by name.
+     * This map stores all clients created for this feature instance.
+     */
+    internal val registeredClients: MutableMap<String, FlagsClient> = mutableMapOf()
```
aleksandr-gringauz: should it be private? do we access it from anywhere outside of this class?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: I get a compiler warning that since this a data class, the `internal` constructor is exposed via the public `copy` method and this would become an error in a future version of Kotlin. I don't think we need this to be a data class so we can get the desired visibility

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: you're right. I moved all the access to `registeredClients` here as well.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: done

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: The FlagClient factory reads the `applicationId` to include in network requests. This can be refactored to be directly applied in the request factory in #2917

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: updated log statements to single targets, mostly `USER`.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationBuilderTest.kt
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: `internal val` so it can still be accessed by the `FlagsClient`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
0xnm: This is fine for now, it is solved with a proper compiler annotation. This issue is for all Configuration classes for our features, so we will address it later.

For now let's just make constructor `internal` (or `private`).

features/dd-sdk-android-flags/async-set-context.md
0xnm, line 1: This file shouldn't be committed, probably?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
```diff
@@ -86,106 +47,128 @@ internal class FlagsTest {
...
 
         // When
         Flags.enable(fakeConfiguration, mockSdkCore)
 
         // Then
-        verify(mockSdkCore).createSingleThreadExecutorService(FLAGS_EXECUTOR_NAME)
+        argumentCaptor<FlagsFeature> {
+            verify(mockSdkCore).registerFeature(capture())
+            // Verify the feature was created (implicit by successful registration)
+            assertThat(lastValue).isNotNull
```
0xnm: this assertion is kind of useless, because if `registerFeature` was called, it means it was called with non-null value according to its signature which doesn't allow nullable type.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClientTest.kt
```diff
@@ -499,7 +501,7 @@ internal class DatadogFlagsClientTest {
         val fakeContext = EvaluationContext(fakeTargetingKey, fakeAttributes)
 
         // When
-        testedClient.setContext(fakeContext)
+        testedClient.setEvaluationContext(fakeContext)
```
0xnm, line 507: comment is regarding `verify` below: instead of `any` we could actually capture the argument is verify its properties.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClientTest.kt
```diff
@@ -487,7 +487,9 @@ internal class DatadogFlagsClientTest {
     }
 
     @Test
-    fun `M process context and store flags W setContext() { complete flow with mock repository }`(forge: Forge) {
+    fun `M process context and store flags W setEvaluationContext() { complete flow with mock repository }`(
```
0xnm:
```suggestion
    fun `M process context and store flags W setEvaluationContext() { complete flow }`(
```

0xnm (approved):

0xnm (changes requested):
Resetting review status, because there is file which shouldn't be probably committed.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
typotter: updated to check the feature's name.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClientTest.kt
typotter: 👍

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: done

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/NoOpFlagsClient.kt
typotter: fixed, thank you

typotter (commented):
alas, I have not been able to quite get through the merge conflicts from Jonathan's merges. tomorrow!

aleksandr-gringauz (approved):

0xnm (approved):

PR #2902, opened 2025-09-29
RUM-11831: Integrate Perfetto ProfilingManager API

ambushwork:
### What does this PR do?

This PR creates `PerfettoProfiler` to delegate Perfetto `ProfilingManager` API:
* Expose `start` and `stop` to internal module to start and stop profiling
* Send telemetry when start and stop
 
### Motivation

RUM-11831

### Additional Notes
* Add `TimeProvider` Interface and default implementation in `internal` module since we need it in `PerfettoProfiler`
* To use Profiling API, we need to upgrade `androidx.core` version to 1.17.0 which deprecates `GestureDetectorCompat` in RUM `GesturesDetectorWrapper`, but we  Suppress it in this PR.


dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/TimeProvider.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.utils
+
+/**
+ * Provides the current device timestamp.
+ * Meant for internal usage.
+ */
+interface TimeProvider {
```
0xnm: we already have `com.datadog.android.core.internal.time.TimeProvider` and also `com.datadog.android.sessionreplay.internal.utils.TimeProvider`.

Maybe it is worth to simply pull the one from `core` to `internal`? Otherwise we have 3 `TimeProvider` interfaces.

detekt_custom_safe_calls.yml
```diff
@@ -441,6 +451,7 @@ datadog:
       - "java.util.concurrent.locks.ReadWriteLock.readLock()"
       - "java.util.concurrent.locks.ReadWriteLock.writeLock()"
       - "java.util.concurrent.locks.ReentrantReadWriteLock.constructor()"
+      -
       # endregion
```
0xnm, line 464:
```suggestion
      - "java.util.concurrent.locks.ReentrantReadWriteLock.constructor()"
      # endregion
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
```diff
@@ -0,0 +1,123 @@
...
+ * @param internalLogger the logger instance to use for logging.
+ * @param timeProvider The time provider to use to get the current time.
+ */
+@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+internal class PerfettoProfiler(
+    private val internalLogger: InternalLogger,
+    private val timeProvider: TimeProvider
+) : Profiler {
+
+    private val profilingExecutor = Executors.newSingleThreadExecutor()
```
0xnm: is it worth to pull it to the constructor to be able to inject the "same-thread" executor for tests?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
```diff
@@ -0,0 +1,123 @@
...
+    private val resultCallback: Consumer<ProfilingResult>
+
+    private var profilingStarted: AtomicBoolean = AtomicBoolean(false)
+
+    private var profilingStartTime = 0L
+
+    init {
+        requestBuilder = StackSamplingRequestBuilder()
+            .setCancellationSignal(stopSignal)
+            .setTag(PROFILING_TAG_APPLICATION_LAUNCH)
```
0xnm, line 51: this is going to be configurable?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
```diff
@@ -0,0 +1,123 @@
...
+                    TELEMETRY_KEY_DURATION to duration
+                )
+            )
+        )
+    }
+
+    companion object {
+        private val PROFILING_MAX_DURATION_MS = TimeUnit.MINUTES.toMillis(1).toInt()
+        private const val PROFILING_TAG_APPLICATION_LAUNCH = "ApplicationLaunch"
+        private const val BUFFER_SIZE_KB = 5120 // 5MB
```
0xnm, line 117: It is worth to add the explanation where this number comes from/why it is chosen like that. Same maybe for the value below.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
```diff
@@ -0,0 +1,123 @@
...
+            if (result.errorCode == ProfilingResult.ERROR_NONE) {
+                // TODO RUM-11853: Implement the upload logic of profiling result.
+            }
+            sendProfilingEndTelemetry(result = result, duration = duration)
+        }
+    }
+
+    override fun start(appContext: Context) {
+        // profiling can start only once, so we never set `profilingStarted` back to false.
+        if (profilingStarted.compareAndSet(false, true)) {
```
0xnm, line 67: since there are no other places where we read it, I guess `profilingStarted` is just to protect from the situation when `start` is called multiple times?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/PerfettoProfilerTest.kt
```diff
@@ -0,0 +1,189 @@
...
+            .requestProfiling(
+                eq(ProfilingManager.PROFILING_TYPE_STACK_SAMPLING),
+                any<Bundle>(),
+                any<String>(),
+                any<CancellationSignal>(),
+                any(),
+                callbackCaptor.capture()
+            )
+
+        val mockResult = org.mockito.kotlin.mock<ProfilingResult> {
```
0xnm: to be moved to `import`

0xnm (commented):
🚀

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
aleksandr-gringauz: Question.

Shouldn't we use these [factory methods](https://github.com/DataDog/dd-sdk-android/blob/fa39327ad28cfe5a1792db2a7f966526e75a1aec/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt#L108) to create executors everywhere?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/TimeProvider.kt
ambushwork: Rebased the branch and replaced it with the new merged `TimeProvider` in `internal`

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
ambushwork: Updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
ambushwork: Not for the moment since our first goal is only about app launch. But we may have other tags if we extend our profiling scope.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
ambushwork: Comments added:
```
// Currently we give an estimated maximum size of profiling result to 5MB, it can be
// increased or configurable if needed.
private const val BUFFER_SIZE_KB = 5120 // 5MB

// Currently we give 100HZ frequency to balance the sampling accuracy and performance
// overhead, it can be increased or configurable if needed.
private const val PROFILING_SAMPLING_RATE = 100 // 100Hz
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
ambushwork: It can be read in other places in the future, for example at some point Rum feature would like to know what's the current status of the profiling.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
ambushwork: Actually good point, but this method needs to call with `sdkCore`, so it will be done in the future commit where `PerfettoProfiler` is constructed.

aleksandr-gringauz (approved):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/PerfettoProfiler.kt
0xnm: My point is to make it clear then that 5 MB is not something like intake limitation and is some artbitrary value chosen by us and which can go up if needed.

0xnm (approved):

PR #2904, opened 2025-09-29
[RUM-11854]: Proper support propagation of RUM context for automatic http tracing via baggage header

satween:
### What does this PR do?

1. Fixes issue with missing `RUM context` in tracing feature that uses `CoreTracer` for automatic tracing. 
2. Ensures that `RUM context` is being added in one single place and works for both custom tracing and automatic tracing
3. Ensures that sampling decision based on sampleId/traceId values works properly


features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/DatadogTracerAdapter.kt
```diff
@@ -67,13 +48,25 @@ internal class DatadogTracerAdapter(
...
+    private fun wrapSpan(span: AgentTracer.SpanBuilder) = DatadogSpanBuilderAdapter(span, spanLogger)
+        .withRumContextIfNeeded()
+
+    private fun DatadogSpanBuilder.withRumContextIfNeeded() = apply {
+        if (!bundleWithRumEnabled) return this
+
+        sdkCore.getFeature(Feature.RUM_FEATURE_NAME)
+            ?.withContext(withFeatureContexts = setOf(Feature.RUM_FEATURE_NAME)) { datadogContext ->
+                val rumContext = datadogContext.featuresContext[Feature.RUM_FEATURE_NAME].orEmpty()
+                withTag(LogAttributes.RUM_VIEW_ID, rumContext["view_id"])
```
aleksandr-gringauz: Question.

IIUC, the callback in `withContext` is called asynchronously. Is it ok, if tags are added a bit later to the builder (maybe even after `builder.start` is called)?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/DatadogTracerAdapter.kt
```diff
@@ -67,13 +48,25 @@ internal class DatadogTracerAdapter(
...
+                withTag(LogAttributes.RUM_ACTION_ID, rumContext["action_id"])
+                withTag(LogAttributes.RUM_SESSION_ID, rumContext["session_id"])
+                withTag(LogAttributes.RUM_APPLICATION_ID, rumContext["application_id"])
+            }
```
0xnm: I think we still may have a concurrency issue: this callback will be called on the context thread and by the time we have to do a trace propagation _it may not yet be called_ (it all depends on how busy the context thread is).

Should we still maybe use `CompletableFuture` for the resolution and block the code responsible for the trace propagation (when it is time to flush trace context into headers/carrier) until it completes?

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/RumContextPropagationTest.kt
```diff
@@ -0,0 +1,278 @@
...
+                put("application_id", anAlphabeticalString())
+                sessionId?.let {
+                    put(
+                        RUM_CONTEXT_SESSION_ID,
+                        "aaaaaaaa-bbbb-Mccc-Nddd-${(sessionId).toULong().toString(16)}"
+                    )
+                }
+            }
+
+        private fun Forge.prepareSubSdkCore(rumContext: Map<String, Any?>): StubSDKCore {
```
0xnm:
```suggestion
        private fun Forge.prepareStubSdkCore(rumContext: Map<String, Any?>): StubSDKCore {
```

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/RumContextPropagationTest.kt
```diff
@@ -0,0 +1,278 @@
...
+            )
+
+            val sdkCoreStub = StubSDKCore(this, datadogContext = datadogContext)
+
+            Datadog::class.java
+                .getStaticValue<Datadog, Any>("registry")
+                .getFieldValue<MutableMap<String, SdkCore>, Any>("instances")
+                .also { instances -> instances += sdkCoreStub.name to sdkCoreStub }
+
+            Trace.enable(TraceConfiguration.Builder().build(), sdkCoreStub)
```
0xnm: this is not related to the core setup, should it be moved out?

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/RumContextPropagationTest.kt
```diff
@@ -0,0 +1,278 @@
...
+
+        // When
+        testedClient.makeNetworkCall()
+
+        // Then
+        assertSentRequest {
+            assertThat(getHeader(HEADER_BAGGAGE)).isEqualTo(
+                "session.id=${fakeRumContext[RUM_CONTEXT_SESSION_ID]}"
+            )
+        }
```
0xnm, line 93: Most probably all these tests will fail if you try to run them with a real SDK core which have a dedicated thread for context processing, because of the concurrency issue described above.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/DatadogTracerAdapter.kt
satween: This change is made with assumption that back-end gonna be able to merge added context to existing trace by `traceId`/`sessionId`.  But yes, `CompletableFuture` might be bit safer here

0xnm: If client portion of the trace consists only of the single span (in most cases it is when only OkHttp instrumentation is used) which missed it, then it won't be the case.

satween: reverted to lazy mode

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/DatadogTracerAdapter.kt
satween: reverted to lazy mode

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/RumContextPropagationTest.kt
satween: I'll keep this tests as it at least test the case when rum context is resolved. 

The rest of the cases is covered in a different places (I reverted the lazy context resulution logic back with slight refactor).
 
Note that I didn't add any synchronization for the RUM context future before data is being sent by span event or http header, but added MAINTAINER logs level to see if we have such issue.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextHelper.kt
```diff
@@ -0,0 +1,89 @@
...
+     * A constant used to store RUM context future.
+     */
+    const val DATADOG_INITIAL_CONTEXT: String = "_dd.datadog_initial_context"
+    internal const val INITIAL_DATADOG_CONTEXT_NOT_AVAILABLE_ERROR = "Initial span creation Datadog context" +
+        " is not available at the write time."
+
+    /**
+     * Helper function that adds RUM context future to the tags for lazy context.
+     */
+    fun DatadogSpanBuilder.injectRumContextFeature(rumFeature: FeatureScope) = apply {
```
aleksandr-gringauz: Is this used outside of `dd-sdk-android-trace` module? If not, should we make it `internal`?

satween: yes, it is being used in `trace-http` and `trace-http-otel` modules

aleksandr-gringauz (dismissed):

aleksandr-gringauz (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/sampling/DeterministicSampler.kt
```diff
@@ -82,12 +82,22 @@ open class DeterministicSampler<T : Any>(
...
+        /**
+         * Represents the maximum allowable sample rate in the sampling process.
+         */
+        const val SAMPLE_ALL_RATE: Float = 100f
 
-        // Good number for Knuth hashing (large, prime, fit in 64 bit long)
-        private const val SAMPLER_HASHER: ULong = 1111111111111111111u
+        /**
+         * A constant value used as a multiplier for generating deterministic hash values
+         * within the `DeterministicSampler` implementation. This value is a good number for
```
0xnm:
```suggestion
         * within the [DeterministicSampler] implementation. This value is a good number for
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextHelper.kt
```diff
@@ -0,0 +1,89 @@
...
+     */
+    fun DatadogSpan.extractRumContextFeature(internalLogger: InternalLogger) = apply {
+        extractRumContextFeature(this, internalLogger)
+    }
+
+    internal fun DDSpan.extractRumContextFeature(internalLogger: InternalLogger) = apply {
+        extractRumContextFeature(this, internalLogger)
+    }
+
+    private fun extractRumContextFeature(instance: Any, internalLogger: InternalLogger) {
```
0xnm: I see 2 cases here:

* when context is extracted at the Span write time -> context should be there by that time, given that write thread is always switched on after the context thread, so if context is not there (there is a `Future`, which is not completed), this is an error. We have nothing to do other than log error and move on.
* when context is extracted for the trace context propagation. -> context _may_ not be there yet. In this case, since most probably call is made at the I/O (network thread), we can wait I think, meaning we can block the execution until `Future` is completed (limiting await time probably).

I think we need to address second case by passing an argument to this function saying if we should await context or not if it is not yet there. WDYT?

satween: I was thinking to check if we have such issue firs, by analyzing logs. 
In case if we face this issues - we could make some changes. Probably I need to add source parameter here to distinguish `TraceInterceptor`  from `CoreSpanWriter`.

0xnm: I think we need to address this before shipping this change, it is a concurrency issue, so it is hard to predict if it will be frequent or not, it all depends on the context thread load.

satween: Okay, so blocking `DatadogContext` resolution has been added for `TracingInterceptor`. For `CoreSpanWriter` logic stays the same.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    private val sdkCore: FeatureSdkCore?
+        get() = sdkCoreProvider()
+
+    private val internalLogger: InternalLogger?
+        get() = sdkCore?.internalLogger
+
+    private fun injectRumContext(builder: DatadogSpanBuilder) {
+        sdkCore?.getFeature(Feature.RUM_FEATURE_NAME)
+            ?.getContextFuture(withFeatureContexts = setOf(Feature.RUM_FEATURE_NAME))
```
0xnm, line 38: question: is there advantage for creating `getContextFuture` method instead of using `withContext`? Seems like we added a method to the core API just for a single use-case, where `withContext` can be used.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
```diff
@@ -0,0 +1,141 @@
...
+            ?.let { lazyContext -> builder.withTag(DATADOG_INITIAL_CONTEXT, lazyContext) }
+    }
+
+    private fun extractRumContextInternal(instance: Any, block: Boolean = false) {
+        val future = instance.getTag<Future<DatadogContext?>>(DATADOG_INITIAL_CONTEXT)
+
+        if (future != null) {
+            val datadogContext = when {
+                block -> future.getOrNull(CONTEXT_RESOLUTION_TIMEOUT, TimeUnit.SECONDS)
```
0xnm, line 47: minor: we may be interested in logging to the telemetry cases when it wasn't resolved on timeout. It would be some signal to investigate (although I'm not sure if it would be easily actionable). WDYT?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
```diff
@@ -0,0 +1,141 @@
...
+                    .orEmpty()
+
+                instance.setTag(LogAttributes.RUM_APPLICATION_ID, rumContext["application_id"])
+                instance.setTag(LogAttributes.RUM_SESSION_ID, rumContext["session_id"])
+                instance.setTag(LogAttributes.RUM_VIEW_ID, rumContext["view_id"])
+                instance.setTag(LogAttributes.RUM_ACTION_ID, rumContext["action_id"])
+            } else {
+                internalLogger?.log(
+                    InternalLogger.Level.ERROR,
+                    listOf(InternalLogger.Target.USER, InternalLogger.Target.MAINTAINER),
```
0xnm: `USER` and `MAINTAINER` shouldn't be used together. Either of the two should be enough. Same for the similar usages below.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
```diff
@@ -0,0 +1,141 @@
...
+            { ERROR_FUTURE_GET_FAILED }
+        )
+        null
+    }
+
+    companion object {
+        /**
+         * A key used to store RUM context future.
+         */
+        const val DATADOG_INITIAL_CONTEXT: String = "_dd.datadog_initial_context"
```
0xnm: should it be public?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -549,7 +553,7 @@ internal constructor(
             setSampledOutHeaders(tracedRequestBuilder, tracingHeaderTypes, span, tracer)
         } else {
             tracer.propagate().inject(
-                span.context(),
+                span.extractRumContext(rumContextPropagator, block = true).context(),
```
0xnm: is this call needed? we will try to get the result of the `Future` for a second time, setting same value again, because we already resolved them in `interceptAndTrace` call up in the stack.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
satween: This is what I've highlighted in DM: It's not being added to the public API. It's an internal method that accessible only through extension method. 

The advantage of having a Future here is that it allows us to: 
- Avoid reinventing the wheel with our own `CompletableFuture`, which could lead to potential issues. The Java `Future` class feels more stable. All required logic is also supported by Future.
- Properly block  the thread when the result is obtained.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
satween: well, I was keeping it public for tests only, changed to internal

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
satween: Good catch. Removed

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
satween: fixed

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt
0xnm: Yes, but then you need to do type-check for `SdkFeature`, because it is not a part of interface (which is for good).

The goal is to have a limited ways to work with context, to simplify the DevEx, like a single simple thing which do the stuff, otherwise we may create a problem of choice.

But I'm ok with having it like that, given it doesn't make it to the interface.

0xnm (dismissed):

0xnm (approved):

PR #2908, opened 2025-10-01
Move DdRumContentProvider to core and use its creation time as app launch time

aleksandr-gringauz:
### What does this PR do?

Currently we use [DatadogCore.startupTimeNs](https://github.com/DataDog/dd-sdk-android/blob/6f16289fa5f3e0fd7ed5ce41c774cf03fff6e732/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProvider.kt#L28C21-L28C46) as app launch timestamp for API < 24 (now it is only 23).

Fields of the companion object get initialized during the first usage of the class in code (in our case `DatadogCore`) which happens somewhere inside `Application.onCreate`. Using `DdRumContentProvider.createTimeNs` (proposed in this pr) is better, because content providers are initialized earlier than `Application.onCreate` and thus we include more user code inside our application launch metrics.

Stacktrace to illustrate the point:
```
lambda 'run' in 'Companion':746, DatadogCore (com.datadog.android.core.internal)
<clinit>:745, DatadogCore (com.datadog.android.core.internal)
lambda 'synchronized' in 'initialize':91, Datadog (com.datadog.android)
initialize:87, Datadog (com.datadog.android)
initialize:125, Datadog (com.datadog.android)
initializeDatadog:145, SampleApplication (com.datadog.android.sample)
onCreate:108, SampleApplication (com.datadog.android.sample)
```

This change requires `DdRumContentProvider` to be used from `DefaultAppStartTimeProvider`, which is in `dd-sdk-android-core`. So I moved `DdRumContentProvider` to `dd-sdk-android-internal` and renamed it.

### Motivation

### Additional Notes

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/rum/DdRumContentProvider.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/rum/DdRumContentProvider.kt
@@ -64,19 +64,19 @@ class DdRumContentProvider : ContentProvider() {
         return 0
     }
 
-    internal companion object {
+    companion object {
         internal const val DEFAULT_IMPORTANCE: Int =
             ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
-        internal var processImportance = 0
 
-        @Suppress("unused") // Used for instrumented tests
-        @JvmStatic
-        private fun overrideProcessImportance(importance: Int) {
-            Log.w(
-                "DdRumContentProvider",
-                "override processImportance: $processImportance -> $importance"
-            )
-            processImportance = importance
-        }
+        /**
+         * Process importance at the moment of creating [DdRumContentProvider]
+         * Should be set from the outside only in tests.
+         */
+        var processImportance: Int = 0
+
+        /**
+         * fallback for APIs below Android N, see [DefaultAppStartTimeProvider].
+         */
+        val createTimeNs: Long = System.nanoTime()
     }
 }
--- a/instrumented/integration/build.gradle.kts
+++ b/instrumented/integration/build.gradle.kts
@@ -95,6 +95,7 @@ dependencies {
     implementation(project(":features:dd-sdk-android-logs"))
     implementation(project(":features:dd-sdk-android-trace"))
     implementation(project(":features:dd-sdk-android-rum"))
+    implementation(project(":dd-sdk-android-internal"))
     implementation(project(":integrations:dd-sdk-android-okhttp"))
 
     implementation(libs.gson)
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
@@ -743,8 +743,5 @@ internal class DatadogCore(
             "SDK core already has \"%s\" listener registered."
 
         internal val CONFIGURATION_TELEMETRY_DELAY_MS = TimeUnit.SECONDS.toMillis(5)
-
-        // fallback for APIs below Android N, see [DefaultAppStartTimeProvider].
-        internal val startupTimeNs: Long = System.nanoTime()
     }
 }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProvider.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProvider.kt
@@ -10,8 +10,8 @@ import android.annotation.SuppressLint
 import android.os.Build
 import android.os.Process
 import android.os.SystemClock
-import com.datadog.android.core.internal.DatadogCore
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import com.datadog.android.rum.DdRumContentProvider
 import java.util.concurrent.TimeUnit
 
 internal class DefaultAppStartTimeProvider(
@@ -25,7 +25,7 @@ internal class DefaultAppStartTimeProvider(
                 val diffMs = SystemClock.elapsedRealtime() - Process.getStartElapsedRealtime()
                 System.nanoTime() - TimeUnit.MILLISECONDS.toNanos(diffMs)
             }
-            else -> DatadogCore.startupTimeNs
+            else -> DdRumContentProvider.createTimeNs
         }
     }
 }
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProviderTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProviderTest.kt
@@ -9,8 +9,8 @@ package com.datadog.android.core.internal.time
 import android.os.Build
 import android.os.Process
 import android.os.SystemClock
-import com.datadog.android.core.internal.DatadogCore
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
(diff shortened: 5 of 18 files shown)


dd-sdk-android-internal/src/main/java/com/datadog/android/rum/DdRumContentProvider.kt
```diff
@@ -64,16 +59,18 @@ class DdRumContentProvider : ContentProvider() {
...
 
-    internal companion object {
-        internal const val DEFAULT_IMPORTANCE: Int =
+    companion object {
+        const val DEFAULT_IMPORTANCE: Int =
             ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
-        internal var processImportance = 0
+        var processImportance: Int = 0
+
+        val createTimeNs: Long = System.nanoTime()
```
0xnm, line 80: if I understand it right, the application start time will be a bit different now **only** for API 23? This may skew customer data a bit, but I guess the impact is very small both in terms of the affected devices and also in the value of data skew?

aleksandr-gringauz: Yes, it is only API 23.

We already know the the metric `rum.measure.app.startup_time` is used by very few customers.

We don't know about the duration of the application_start action and the ApplicationLaunch view, but I suspect not big either. And together with the fact that it is only API 23, I think it is ok to just fix. So that our new TTID and TTFD are better from the start.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/DdCoreContentProvider.kt
```diff
@@ -18,7 +12,8 @@ import android.util.Log
 /**
  * A Content provider used to monitor the Application startup time efficiently.
  */
-class DdRumContentProvider : ContentProvider() {
+@Suppress("UndocumentedPublicProperty")
```
ambushwork: Why not just add documentation on the property even if it is just "internal" public

dd-sdk-android-internal/src/main/java/com/datadog/android/rum/DdRumContentProvider.kt
aleksandr-gringauz: > value of data skew

it shouldn't be big in most cases, but it of course depends on the app. If the app does something heavy in Application.onCreate, it will start being included in application launch time after the fix

dd-sdk-android-internal/src/test/java/com/datadog/android/internal/DdCoreContentProviderTest.kt
```diff
@@ -98,7 +98,7 @@ class DdRumContentProviderTest {
         testedProvider.onCreate()
 
         // Then
-        assertThat(DdRumContentProvider.processImportance).isEqualTo(processImportance)
+        Assertions.assertThat(DdCoreContentProvider.processImportance).isEqualTo(processImportance)
```
ambushwork: static import can be used here

features/dd-sdk-android-rum/src/main/AndroidManifest.xml
```diff
@@ -9,8 +9,8 @@
     <application>
 
         <provider
-            android:name=".DdRumContentProvider"
-            android:authorities="${applicationId}.provider.datadog.rum"
```
ambushwork, line 13: > [!IMPORTANT]
> This seems a breaking change for some clients, `DdRumContentProvider` is public but `DdCoreContentProvider` becomes "internal" public, we should consider if there is any edge case that clients still need it in their code, for example override some tags of this provider in their manifest, or some other usage like what we do in [integration test](https://github.com/DataDog/dd-sdk-android/blob/develop/instrumented/integration/src/main/kotlin/com/datadog/android/sdk/integration/rum/ActivityTrackingPlaygroundActivity.kt)

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/DdCoreContentProvider.kt
```diff
@@ -18,7 +12,8 @@ import android.util.Log
...
+class DdCoreContentProvider : ContentProvider() {
```
ambushwork: Why it's renamed to xxCorexx? can it just be `DdContentProvider`?

instrumented/integration/build.gradle.kts
```diff
@@ -95,6 +95,7 @@ dependencies {
     implementation(project(":features:dd-sdk-android-logs"))
     implementation(project(":features:dd-sdk-android-trace"))
     implementation(project(":features:dd-sdk-android-rum"))
+    implementation(project(":dd-sdk-android-internal"))
```
ambushwork, line 98: Personal opinion: Here is the thing I mentioned above, theoretically `integration` module should simulate the client scenario with activities and fragments to test the integrated functionality, meaning that we should not directly include `internal` in production dependency, if it's not the case, then client will probably face the same issue.

features/dd-sdk-android-rum/src/main/AndroidManifest.xml
aleksandr-gringauz: I could do the same as I did in this PR, but move it to "dd-sdk-android-core" package instead without renaming and maintaining package name.

WDYT, @ambushwork ?

aleksandr-gringauz: Or even better. Move to `dd-sdk-android-internal` like I already did but maintaining package and name.

Yes, it is true that `DdRumContentProvider` class was accessible from kotlin code. But there were no public methods or fields besides those that come from the `ContentProvider` class. I doubt that you can do anything reasonable with it.

aleksandr-gringauz: > Move to dd-sdk-android-internal like I already did but maintaining package and name.

did this

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/DdCoreContentProvider.kt
aleksandr-gringauz: renamed back to `DdRumContentProvider`

dd-sdk-android-internal/src/main/java/com/datadog/android/rum/DdRumContentProvider.kt
```diff
@@ -64,19 +64,19 @@ class DdRumContentProvider : ContentProvider() {
...
 
-    internal companion object {
+    companion object {
         internal const val DEFAULT_IMPORTANCE: Int =
             ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
-        internal var processImportance = 0
 
-        @Suppress("unused") // Used for instrumented tests
-        @JvmStatic
-        private fun overrideProcessImportance(importance: Int) {
```
aleksandr-gringauz, line 74: Removed this used using reflection in instrumented tests. Instead use `processImportance` field in the tests directly.

Now `DdRumContentProvider` is in the `dd-sdk-android-internal` module and isn't accessible to the sdk clients. So I think it is pretty safe to just use `processImportance` without the need to use reflection.

instrumented/integration/build.gradle.kts
aleksandr-gringauz: see this comment https://github.com/DataDog/dd-sdk-android/pull/2908/files#r2394830743

basically I stopped using reflection (which is kind of relying on hidden internals) and started using `dd-sdk-android-internal` (which is also hidden internals).

ambushwork (approved):

0xnm (approved):

PR #2910, opened 2025-10-02
RUM-11785: Introduce telemetry for TTID

aleksandr-gringauz:
### What does this PR do?

This pr connects together `RumAppStartupDetector`, `RumTTIDReporter` and reports TTID value using telemetry along with some additional information. You can read the spec of the telemetry [here](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5561778603/Android+-+AppLaunch+TTID+Telemetry+-+Spec).

Reporting the values to RUM intake using Vitals will be done in future PRs, because currently there are some frontend problems that prevent them from being displayed properly.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
@@ -72,6 +72,10 @@ import com.datadog.android.rum.internal.metric.slowframes.SlowFramesListener
 import com.datadog.android.rum.internal.monitor.AdvancedRumMonitor
 import com.datadog.android.rum.internal.monitor.DatadogRumMonitor
 import com.datadog.android.rum.internal.net.RumRequestFactory
+import com.datadog.android.rum.internal.startup.RumAppStartupDetector
+import com.datadog.android.rum.internal.startup.RumStartupScenario
+import com.datadog.android.rum.internal.startup.RumTTIDInfo
+import com.datadog.android.rum.internal.startup.RumTTIDReporter
 import com.datadog.android.rum.internal.thread.NoOpScheduledExecutorService
 import com.datadog.android.rum.internal.tracking.JetpackViewAttributesProvider
 import com.datadog.android.rum.internal.tracking.NoOpInteractionPredicate
@@ -167,6 +171,7 @@ internal class RumFeature(
     internal val rumContextUpdateReceivers = mutableSetOf<FeatureContextUpdateReceiver>()
 
     private val lateCrashEventHandler by lazy { lateCrashReporterFactory(sdkCore as InternalSdkCore) }
+    private var rumAppStartupDetector: RumAppStartupDetector? = null
 
     // region Feature
 
@@ -250,6 +255,8 @@ internal class RumFeature(
 
         sessionListener = configuration.sessionListener
 
+        initRumAppStartupDetector()
+
         sdkCore.setEventReceiver(name, this)
 
         initialized.set(true)
@@ -332,6 +339,9 @@ internal class RumFeature(
         sessionListener = NoOpRumSessionListener()
 
         cleanupInfoProviders()
+
+        rumAppStartupDetector?.destroy()
+        rumAppStartupDetector = null
 
         GlobalRumMonitor.unregister(sdkCore)
         initialized.set(false)
@@ -661,6 +671,30 @@ internal class RumFeature(
 
     private fun addSessionReplaySkippedFrame() {
         (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor)?.addSessionReplaySkippedFrame()
+    }
+
+    private fun initRumAppStartupDetector() {
+        rumAppStartupDetector = RumAppStartupDetector.create(
+            application = appContext.applicationContext as Application,
+            sdkCore = sdkCore as InternalSdkCore,
+            listener = object : RumAppStartupDetector.Listener {
+                private val rumTTIDReporter = RumTTIDReporter.create(
+                    sdkCore = sdkCore,
+                    listener = object : RumTTIDReporter.Listener {
+                        override fun onTTIDCalculated(
+                            info: RumTTIDInfo
+                        ) {
+                            (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor)
+                                ?.sendTTIDEvent(info)
+                        }
+                    }
+                )
+
+                override fun onAppStartupDetected(scenario: RumStartupScenario) {
+                    rumTTIDReporter.onAppStartupDetected(scenario)
+                }
+            }
+        )
     }
 
     // endregion
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumRawEvent.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumRawEvent.kt
@@ -17,6 +17,7 @@ import com.datadog.android.rum.featureoperations.FailureReason
 import com.datadog.android.rum.internal.RumErrorSourceType
 import com.datadog.android.rum.internal.domain.Time
 import com.datadog.android.rum.internal.domain.event.ResourceTiming
+import com.datadog.android.rum.internal.startup.RumTTIDInfo
 import com.datadog.android.rum.model.ActionEvent
 
 internal sealed class RumRawEvent {
@@ -270,5 +271,10 @@ internal sealed class RumRawEvent {
         val attributes: Map<String, Any?>,
         val failureReason: FailureReason? = null,
         override val eventTime: Time = Time()
+    ) : RumRawEvent()
+
+    internal class AppStartTTIDEvent(
+        override val eventTime: Time = Time(),
+        val info: RumTTIDInfo
     ) : RumRawEvent()
 }
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
@@ -24,6 +24,7 @@ import com.datadog.android.rum.internal.domain.battery.BatteryInfo
 import com.datadog.android.rum.internal.domain.display.DisplayInfo
 import com.datadog.android.rum.internal.metric.SessionMetricDispatcher
 import com.datadog.android.rum.internal.metric.slowframes.SlowFramesListener
+import com.datadog.android.rum.internal.startup.RumAppStartupTelemetryReporter
 import com.datadog.android.rum.internal.utils.percent
 import com.datadog.android.rum.internal.vitals.VitalMonitor
 import com.datadog.android.rum.metric.interactiontonextview.LastInteractionIdentifier
@@ -56,7 +57,8 @@ internal class RumSessionScope(
     private val displayInfoProvider: InfoProvider<DisplayInfo>,
     private val sessionInactivityNanos: Long = DEFAULT_SESSION_INACTIVITY_NS,
     private val sessionMaxDurationNanos: Long = DEFAULT_SESSION_MAX_DURATION_NS,
-    rumSessionTypeOverride: RumSessionType?
+    rumSessionTypeOverride: RumSessionType?,
+    private val rumAppStartupTelemetryReporter: RumAppStartupTelemetryReporter
 ) : RumScope {
 
     internal var sessionId = RumContext.NULL_UUID
@@ -70,6 +72,8 @@ internal class RumSessionScope(
     private val random = SecureRandom()
 
     private val noOpWriter = NoOpDataWriter<Any>()
+
+    private var appStartIndex: Int = 0
 
     @Suppress("LongParameterList")
     internal var childScope: RumViewManagerScope? = RumViewManagerScope(
@@ -149,9 +153,21 @@ internal class RumSessionScope(
 
         val actualWriter = if (sessionState == State.TRACKED) writer else noOpWriter
 
-        if (event !is RumRawEvent.SdkInit) {
-            childScope =
-                childScope?.handleEvent(event, datadogContext, writeScope, actualWriter) as? RumViewManagerScope
+        when (event) {
+            is RumRawEvent.AppStartTTIDEvent -> {
+                if (sessionState == State.TRACKED) {
+                    rumAppStartupTelemetryReporter.reportTTID(
+                        info = event.info,
+                        indexInSession = appStartIndex
+                    )
+                    appStartIndex++
+                }
+            }
+            is RumRawEvent.SdkInit -> {}
+            else -> {
+                childScope =
+                    childScope?.handleEvent(event, datadogContext, writeScope, actualWriter) as? RumViewManagerScope
+            }
         }
 
         return if (isSessionComplete()) {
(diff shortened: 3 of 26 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/DdRumContentProvider.kt
```diff
@@ -68,6 +68,7 @@ class DdRumContentProvider : ContentProvider() {
         internal const val DEFAULT_IMPORTANCE: Int =
             ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
         internal var processImportance = 0
+        internal val creationTimeNs: Long = System.nanoTime()
```
aleksandr-gringauz: This is needed only until [this](https://github.com/DataDog/dd-sdk-android/pull/2908) pr is merged.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -149,9 +153,21 @@ internal class RumSessionScope(
 
         val actualWriter = if (sessionState == State.TRACKED) writer else noOpWriter
 
-        if (event !is RumRawEvent.SdkInit) {
-            childScope =
-                childScope?.handleEvent(event, datadogContext, writeScope, actualWriter) as? RumViewManagerScope
+        when (event) {
+            is RumRawEvent.AppStartTTIDEvent -> {
+                if (sessionState == State.TRACKED) {
+                    rumAppStartupTelemetryReporter.reportTTID(
```
aleksandr-gringauz, line 159: I expect that reporting the actual value to the RUM intake using Vitals will take place exactly next to this `reportTTID` call in `RumSessionScope`. 

The new vital doesn't require a view, so no need to put this in `RumViewScope`.

It will be in the future PRs.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumRawEvent.kt
```diff
@@ -271,4 +272,9 @@ internal sealed class RumRawEvent {
         val failureReason: FailureReason? = null,
         override val eventTime: Time = Time()
     ) : RumRawEvent()
+
+    internal class AppStartTTIDEvent(
+        override val eventTime: Time,
```
satween: in other places `eventTime` is an optional parameter, should it be optional here too? 

```kotlin
override val eventTime: Time = Time()
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
satween: I am not 100% sure about this, but as far as I know we bound every event to a view. I mean it's impossible to report an event without being bound to a view. Probably you should use background view or so? cc @0xnm

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
satween, line 237: why we should renew session with app start event? Wouldn't it be some side effects from it? (like  if user launch app from HOT/WARM state?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupTelemetryReporterImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+                    put(KEY_PROCESS_START_IMPORTANCE, processStartImportance)
+
+                    put(KEY_HAS_SAVED_INSTANCE_STATE, info.scenario.hasSavedInstanceStateBundle)
+
+                    info.scenario.gap()?.let {
+                        put(KEY_GAP, it.inWholeMilliseconds)
+                    }
+                }
+            ),
+            samplingRate = 15.0f
```
satween: could be a constant value to re-use in tests

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupTelemetryReporterImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val sdkCore: InternalSdkCore,
+    private val contentProviderCreationTimeNanos: Long,
+    private val processStartImportance: Int
+) : RumAppStartupTelemetryReporter {
+
+    override fun reportTTID(
+        info: RumTTIDInfo,
+        indexInSession: Int
+    ) {
+        internalLogger.logMetric(
```
satween, line 22: why TTID is being reported as a log event, but not as a VitalEvent? 
I think it's better to have unified format for TTID for both platforms

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
```diff
@@ -32,3 +35,19 @@ internal sealed interface RumStartupScenario {
         override val activity: WeakReference<Activity>
     ) : RumStartupScenario
 }
+
+internal fun RumStartupScenario.name(): String {
+    return when (this) {
+        is RumStartupScenario.Cold -> "cold"
```
satween: nit: maybe add `name` as a property for `RumStartupScenario` instance or convert this function to a `val` property ?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
```diff
@@ -32,3 +35,19 @@ internal sealed interface RumStartupScenario {
...
+        is RumStartupScenario.WarmAfterActivityDestroyed -> "warm_after_activity_destroyed"
+        is RumStartupScenario.WarmFirstActivity -> "warm_first_activity"
+    }
+}
+
+internal fun RumStartupScenario.gap(): Duration? {
```
satween: nit: same here with the property

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumTTIDReporterImpl.kt
```diff
@@ -95,10 +95,13 @@ internal class RumTTIDReporterImpl(
     }
 
     private fun onFirstDraw(scenario: RumStartupScenario) {
-        val duration = (timeProviderNanos() - scenario.initialTimeNanos).nanoseconds
+        val now = timeProviderNanos()
+        val duration = (now - scenario.initialTimeNanos).nanoseconds
```
satween: nit: It is looks like `timeProviderNanos()` does not provide nanoseconds but some `Date` object?
In that case maybe it make sense to rename `timeProviderNanos()` to just `timeProvider` and `initialTimeNanos` to `initialTime`. Otherwise maybe worth to store nanoseconds in `initialTimeNanos`. This would make everything more consistent

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -663,6 +673,30 @@ internal class RumFeature(
         (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor)?.addSessionReplaySkippedFrame()
     }
 
+    private fun initRumAppStartupDetector() {
+        rumAppStartupDetector = RumAppStartupDetector.create(
+            application = appContext as Application,
```
satween: nit: appContext.getApplicationContext() could be bit safer I guess

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -1406,6 +1424,93 @@ internal class RumSessionScopeTest {
...
+        // When
+        val result = testedScope.handleEvent(
+            event = event,
+            datadogContext = fakeDatadogContext,
+            writeScope = mockEventWriteScope,
+            writer = mockWriter
+        )
+
+        // Then
+        val context1 = result!!.getRumContext()
```
satween:
``` kotlin
val context = checkNotNull(result?.getRumContext())
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
aleksandr-gringauz: We plan to actually not require view for TTID. Simao created a special type of vital for TTID and TTFD and made it so that it doesn't require view. However for now there is some frontend work that needs to be done so that this could be displayed properly.

https://github.com/DataDog/rum-events-format/pull/302
https://github.com/DataDog/rum-events-format/pull/306

That's why I said in the description that the code for sending the vitals will be in the future PRs. This PR only contains telemetry.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
aleksandr-gringauz: > why we should renew session with app start event?

There is a second `if` below `if (isNewSession || isExpired || isTimedOut) {`. We renew only in this case, not every time TTID event is received.

I added this `isAppStartEvent` as a thing similar to `isInteraction`, so that if TTID happens before the first view is started we still start the session and send it. In practice this will most likely be an impossible scenario, because TTID will happen somewhere inside the first real RUM View.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupTelemetryReporterImpl.kt
aleksandr-gringauz: This is just Telemetry, not actual data for the users. See this comment https://github.com/DataDog/dd-sdk-android/pull/2910/files#r2398954697

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumTTIDReporterImpl.kt
aleksandr-gringauz: > It is looks like timeProviderNanos() does not provide nanoseconds but some Date object?

now it returns nanoseconds

Do you propose using kotlin `Duration` type everywhere instead of passing `Long`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumRawEvent.kt
aleksandr-gringauz: done

I'm not a big fan of default arguments. I feel I can get trapped by them, forget that a certain argument exists, it silently sets to default, but this is not the behavior that I wanted. So I myself almost never tend to use them.

But this is pretty opinionated.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
aleksandr-gringauz: changed to computed property

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
aleksandr-gringauz: changed to computed property

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupTelemetryReporterImpl.kt
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.startup
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.InternalSdkCore
+
+internal class RumAppStartupTelemetryReporterImpl(
+    private val internalLogger: InternalLogger,
```
ambushwork, line 12: minor: since you have `sdkCore`, you can get internalLogger with it, don't have to pass it again.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
```diff
@@ -17,13 +18,15 @@ internal sealed interface RumStartupScenario {
     class Cold(
         override val initialTimeNanos: Long,
         override val hasSavedInstanceStateBundle: Boolean,
-        override val activity: WeakReference<Activity>
+        override val activity: WeakReference<Activity>,
+        val gap: Duration
```
ambushwork: Should we give a more explicit name to explain which kind of gap it is

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupTelemetryReporterImpl.kt
aleksandr-gringauz: Stopped passing `sdkCore`, started passing `appStartupTimeNs`. `sdkCore` was needed for only this.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumStartupScenario.kt
aleksandr-gringauz: renamed to `appStartActivityOnCreateGap`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumTTIDReporterImpl.kt
satween: I am basically concerning regarding misleading of the name and the meaning. If we providing nanos we should use `Long`, if we provide `Duration` we probably should not call it `...Nanos`

satween: Otherwise this code is confusing: 
```
(now - scenario.initialTimeNanos).nanoseconds
```

we subscrating nanoseconds and then taking nanoseconds from the result. But it already should be nanoseconds, right?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
satween: I am not argue with that, but as TTID stands fro time to initial **display** maybe view still required to tell to the user which exact view is responsible for the initial display?

aleksandr-gringauz: > I am not argue with that, but as TTID stands fro time to initial display maybe view still required to tell to the user which exact view is responsible for the initial display?

Currently we plan not to send the viewId with TTID vital. However the frontend implementation for this isn't ready and there is indeed a chance that we might encounter some unexpected problems and have to actually send the viewId. If so, ok - we will change our implementation.

We plan to show the TTID value only in session level UI, not on View level.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumTTIDReporterImpl.kt
aleksandr-gringauz: Indeed the code is a bit confusing. Though it was correct I think.

I stopped using `Duration` class, changed the naming everywhere to `...somethingNs`. Should be cleaner now.

ambushwork (approved):

PR #2911, opened 2025-10-02
[RUM-11854]: Account id  and User id propagated in baggage header

satween:
### What does this PR do?

Adds support for `accountId` and `userId` propagation through `baggage` header.
Add and refactor tests that is related to rum context propagation


features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/core/propagation/DatadogHttpCodecTest.kt
```diff
@@ -71,7 +79,7 @@ internal class DatadogHttpCodecTest {
                 capture()
             )
 
-            assertThat(firstValue).isEqualTo("session.id=$fakeSessionId")
+            assertThat(firstValue).isEqualTo("account.id=$fakeAccountId,user.id=$fakeUserId,session.id=$fakeSessionId")
```
ambushwork: I think it's also nice to test  the case when `accountId` or `userId` is null or empty, we don't have the string like "account.id=,user.id=,"

satween: yep, make sense, added

PR #2915, opened 2025-10-06
RUM-11853: Implement Profiling writer and request factory

ambushwork:
### What does this PR do?

This PR implements the logic how Perfetto Result should be written into batch file and how the profile request should be built.

### Motivation

RUM-11853

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/TimeExt.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/TimeExt.kt
@@ -1,0 +1,28 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.utils
+
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+
+/**
+ * Formats the given epoch milliseconds to an ISO-8601 formatted string in UTC timezone.
+ *
+ * Example: 2020-02-14T12:34:56.000Z
+ *
+ * @param epochMillis The time in milliseconds to format.
+ * @return The formatted date string.
+ */
+@Suppress("UnsafeThirdPartyFunctionCall")
+fun formatIsoUtc(epochMillis: Long): String {
+    // NPE cannot happen here, ISO_8601 pattern is valid
+    val sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US)
+    sdf.timeZone = TimeZone.getTimeZone("UTC")
+    return sdf.format(Date(epochMillis))
+}
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
@@ -1,0 +1,94 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.profiling.internal
+
+import com.datadog.android.api.context.DatadogContext
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.api.storage.EventType
+import com.datadog.android.api.storage.RawBatchEvent
+import com.datadog.android.core.internal.persistence.file.readBytesSafe
+import com.datadog.android.internal.utils.formatIsoUtc
+import com.datadog.android.profiling.model.ProfilingEvent
+import java.io.File
+
+internal class ProfilingDataWriter(
+    private val sdkCore: FeatureSdkCore
+) : ProfilingWriter {
+    override fun write(
+        profilingResult: PerfettoResult
+    ) {
+        sdkCore.getFeature(Feature.PROFILING_FEATURE_NAME)
+            ?.withWriteContext { context, writeScope ->
+                writeScope { writer ->
+                    val rawBatchEvent = buildRawBatchEvent(
+                        context = context,
+                        profilingResult = profilingResult
+                    )
+                    if (rawBatchEvent != null) {
+                        synchronized(this) {
+                            writer.write(
+                                event = rawBatchEvent,
+                                batchMetadata = null,
+                                eventType = EventType.DEFAULT
+                            )
+                        }
+                    }
+                }
+            }
+    }
+
+    private fun buildRawBatchEvent(
+        context: DatadogContext,
+        profilingResult: PerfettoResult
+    ): RawBatchEvent? {
+        val byteData = readProfilingData(profilingResult.resultFilePath)
+        if (byteData == null || byteData.isEmpty()) {
+            return null
+        }
+        val profilingEvent = createProfilingEvent(context, profilingResult)
+        val serializedEvent =
+            profilingEvent.toJson().toString().toByteArray(Charsets.UTF_8)
+        return RawBatchEvent(data = serializedEvent, metadata = byteData)
+    }
+
+    private fun createProfilingEvent(
+        context: DatadogContext,
+        profilingResult: PerfettoResult
+    ): ProfilingEvent {
+        return ProfilingEvent(
+            start = formatIsoUtc(profilingResult.start),
+            end = formatIsoUtc(profilingResult.end),
+            attachments = listOf(PERFETTO_ATTACHMENT_NAME),
+            family = ANDROID_FAMILY_NAME,
+            version = VERSION_NUMBER,
+            tagsProfiler = buildTags(context)
+        )
+    }
+
+    private fun buildTags(context: DatadogContext): String = buildString {
+        append("$TAG_KEY_SERVICE:${context.service}")
+        append(",")
+        append("$TAG_KEY_VERSION:${context.version}")
+    }
+
+    private fun readProfilingData(profilingPath: String): ByteArray? {
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        // profilingPath is not null in kotlin, it can't throw NPE
+        return File(profilingPath).readBytesSafe(internalLogger = sdkCore.internalLogger)
+    }
+
+    companion object {
+        private const val TAG_KEY_SERVICE = "service"
+        private const val TAG_KEY_VERSION = "version"
+        private const val PERFETTO_ATTACHMENT_NAME = "perfetto.proto"
+        private const val ANDROID_FAMILY_NAME = "android"
+
+        // Only `4` is supported by profiling Backend
+        private const val VERSION_NUMBER = "4"
+    }
+}
--- a/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingDataWriterTest.kt
+++ b/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingDataWriterTest.kt
@@ -1,0 +1,161 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.profiling.internal
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.context.DatadogContext
+import com.datadog.android.api.feature.EventWriteScope
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureScope
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.api.storage.EventBatchWriter
+import com.datadog.android.api.storage.EventType
+import com.datadog.android.api.storage.RawBatchEvent
+import com.datadog.android.internal.utils.formatIsoUtc
+import com.datadog.android.profiling.forge.Configurator
+import com.datadog.android.profiling.model.ProfilingEvent
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.annotation.Forgery
+import fr.xgouchet.elmyr.junit5.ForgeConfiguration
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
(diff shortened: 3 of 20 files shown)


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+                                event = rawBatchEvent,
+                                batchMetadata = null,
+                                eventType = EventType.DEFAULT
+                            )
+                        }
+                        isWritten = true
+                    }
+                }
+            }
+        return isWritten
```
aleksandr-gringauz: This will always return `false`, because `withWriteContext` is calling the callback asynchronously. Why do we need to return a boolean here?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+        val serializedEvent =
+            profilingEvent.toJson().toString().toByteArray(Charsets.UTF_8)
+        return RawBatchEvent(data = serializedEvent, metadata = byteData)
+    }
+
+    private fun createProfilingEvent(
+        context: DatadogContext,
+        profilingResult: PerfettoResult
+    ): ProfilingEvent {
+        sdkCore.getFeatureContext(Feature.PROFILING_FEATURE_NAME)
```
aleksandr-gringauz: This line seems pointless. The result of `getFeatureContext` call isn't used.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        return isWritten
+    }
+
+    private fun buildRawBatchEvent(
+        context: DatadogContext,
+        profilingResult: PerfettoResult
+    ): RawBatchEvent? {
+        val byteData = readProfilingData(profilingResult.resultFilePath)
```
aleksandr-gringauz, line 49: nit:

I would just return `null` from `readProfilingData` and check the result later without creating `ByteArray(0)` and checking `byteData.isEmpty()`.

`readProfilingData` in this way has a cleaner api, `null` value becomes an explicit sign of error.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/TimeExt.kt
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Example: 2020-02-14T12:34:56Z
+ *
+ * @param epochMillis The time in milliseconds to format.
+ * @return The formatted date string.
+ */
+@Suppress("UnsafeThirdPartyFunctionCall")
+fun formatIsoUtc(epochMillis: Long): String {
+    // NPE cannot happen here, ISO_8601 pattern is valid
+    val sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US)
```
0xnm: are we intentionally dropping milliseconds here? it is different from what we use here https://github.com/DataDog/dd-sdk-android/blob/8f38ae839101239d1024206008ca9cb08a947d63/features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/utils/LogUtils.kt#L13

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+import java.io.File
+
+internal class ProfilingDataWriter(
+    private val sdkCore: FeatureSdkCore
+) : ProfilingWriter {
+    override fun write(
+        profilingResult: PerfettoResult
+    ): Boolean {
+        var isWritten = false
+        sdkCore.getFeature(Feature.PROFILING_FEATURE_NAME)
```
0xnm: `withWriteContext` is async method, so most probably most of the time `isWritten` will be `false` at `return` statement.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
0xnm: seems like redundant line?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    private fun buildTags(context: DatadogContext): String = buildString {
+        append("$TAG_KEY_SERVICE:${context.service}")
+        append(",")
+        append("$TAG_KEY_VERSION:${context.version}")
+    }
+
+    private fun readProfilingData(profilingPath: String): ByteArray {
+        return try {
+            File(profilingPath).readBytes()
```
0xnm: minor: you can use `File.safeCall` here if you want.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+            ByteArray(0)
+        }
+    }
+
+    companion object {
+        private const val TAG_KEY_SERVICE = "service"
+        private const val TAG_KEY_VERSION = "version"
+        private const val PERFETTO_ATTACHMENT_NAME = "perfetto.proto"
+        private const val ANDROID_FAMILY_NAME = "android"
+        private const val VERSION_NUMBER = "4"
```
0xnm, line 92: minor: maybe we need to add an explanation of this, why it is `4` for a new development.

0xnm (changes requested):
reviewed production files so far

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+        )
+    }
+
+    private fun buildTags(context: DatadogContext): String = buildString {
+        append("$TAG_KEY_SERVICE:${context.service}")
+        append(",")
+        append("$TAG_KEY_VERSION:${context.version}")
+    }
+
+    private fun readProfilingData(profilingPath: String): ByteArray {
```
aleksandr-gringauz: Question.

Probably not a topic for this PR, but what will happen to the original profile file? Will we delete it later at some point? Or are these files managed by Android SDK?

ambushwork: Google Profiling team claims that these profiling results will be recycled by the OS, will not last long unless we put a debug flag on the phone, so we don't have to care about them.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: explained already in `profiling-schema.json` but I duplicate here for better visibility.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: updated.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: removed

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: `isWritten` removed.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/TimeExt.kt
ambushwork: It was missing in my POC branch, updated now

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: updated.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: removed,

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: Good point,  tried to follow the `DataWriter` pattern at the beginning, but actually it won't be needed, I will remove it.

aleksandr-gringauz (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/TimeExt.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+
+/**
+ * Formats the given epoch milliseconds to an ISO-8601 formatted string in UTC timezone.
+ *
+ * Example: 2020-02-14T12:34:56Z
```
0xnm: very nit: example should include milliseconds as per actual format.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingDataWriterTest.kt
```diff
@@ -0,0 +1,166 @@
...
+        // Don't create the tmp file so it can't be found
+
+        // When
+        testedDataWriterTest.write(
+            profilingResult = fakeResult
+        )
+
+        // Then
+        verifyNoMoreInteractions(mockInternalLogger)
+        verifyNoMoreInteractions(mockEventBatchWriter)
```
0xnm: minor: `verifyNoMoreInteractions` is vararg method, so can be simply `verifyNoMoreInteractions(mockInternalLogger , mockEventBatchWriter)`.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingDataWriterTest.kt
```diff
@@ -0,0 +1,166 @@
...
+        testedDataWriterTest.write(
+            profilingResult = fakeResult.copy(resultFilePath = file.absolutePath)
+        )
+
+        // Then
+        verifyNoMoreInteractions(mockInternalLogger)
+        verifyNoMoreInteractions(mockEventBatchWriter)
+    }
+
+    private fun buildRawBatchEvent(
```
0xnm: alternatively you can simply use `argumentCaptor` and the assert properties of the object captured, because the following approach tend to simply copy-paste code from production files.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingRequestFactoryTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        val batchMetadata = forge.aNullable { batchMetadata.toByteArray() }
+
+        // When
+        val request =
+            testedFactory.create(fakeDatadogContext, executionContext, batchData, batchMetadata)
+
+        // Then
+        requireNotNull(request)
+        assertThat(request.url).isEqualTo("${fakeDatadogContext.site.intakeEndpoint}/api/v2/profile")
+        assertThat(request.contentType).contains("multipart/form-data;")
```
0xnm, line 73: should it be `isEqualTo` here?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingRequestFactoryTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    @Test
+    @Suppress("NAME_SHADOWING")
+    fun `M create a proper request W create()`(
+        @Forgery batchData: List<RawBatchEvent>,
+        @Forgery executionContext: RequestExecutionContext,
+        @StringForgery batchMetadata: String,
+        forge: Forge
+    ) {
+        // Given
+        val batchMetadata = forge.aNullable { batchMetadata.toByteArray() }
```
0xnm: you can avoid `NAME_SHADOWING` suppression by generating string manually, I think it is better:

```suggestion
        val batchMetadata = forge.aNullable { forge.aString().toByteArray() }
```

0xnm (dismissed):
I left a few comments, but nothing blocking.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingRequestFactoryTest.kt
ambushwork: `contentType` contains boundary like this:
```
multipart/form-data; boundary=12726397-1fec-4896-899e-b94cbf460bd7
```
but I think we only need to test `multipart/form-data` exist.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingRequestFactoryTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingDataWriterTest.kt
ambushwork: Good point, updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingDataWriterTest.kt
ambushwork: updated

0xnm (approved):

aleksandr-gringauz (approved):

PR #2917, opened 2025-10-07
feat: custom flagging endpoint and assignment download refactoring.

typotter:
### What does this PR do?

This PR refactors the precomputed assignments request mechanism to follow the SDK-wide **Factory + Executor pattern**, decoupling request construction from execution. The main changes include:

- Created `PrecomputedAssignmentsRequestFactory` to handle request building
- Renamed `DefaultFlagsNetworkManager` → `PrecomputedAssignmentsDownloader` to focus on request execution
- Properly wired custom flag endpoint configuration through the entire request flow
- Added comprehensive unit and integration tests

### Motivation

**Problem**: The existing `DefaultFlagsNetworkManager` tightly coupled request construction (building headers, body, URL) with request execution (network calls, error handling), making it difficult to test request construction independently and inconsistent with patterns used throughout the SDK.

**Goals**:
1. **Decouple request building from execution** - Separate concerns for better testability and maintainability
2. **Follow SDK patterns** - Match the Factory + Executor pattern used by `DataOkHttpUploader`, `ExposuresRequestFactory`, and other features
3. **Improve testability** - Enable testing request construction without network calls and vice versa
4. **Ensure custom endpoint support** - Verify that custom flag endpoint configuration works correctly end-to-end

### Architecture Changes

**Before:**
```
DefaultFlagsNetworkManager
  ├─> downloadPrecomputedFlags()
  ├─> buildHeaders()
  ├─> buildRequestBody()
  └─> executeDownloadRequest()
```

**After (Factory + Executor Pattern):**
```
FlagsFeature
  └─> precomputedRequestFactory (shared, stateless)

FlagsClient
  └─> PrecomputedAssignmentsDownloader(requestFactory, flagsContext)
        ├─> factory.create() → builds Request
        └─> executeDownloadRequest() → executes Request
```

#### 1. PrecomputedAssignmentsRequestFactory
- **Interface**: Defines contract for creating OkHttp requests
- **DefaultPrecomputedAssignmentsRequestFactory**: Implementation that builds requests with:
  - URL determination via `EndpointsHelper` (supports custom endpoints)
  - Authentication headers (client token, application ID)
  - JSON-API formatted request body
  - Proper error handling with logging
- **Scope**: One instance per feature (stateless, shared across clients)

#### 2. PrecomputedAssignmentsDownloader
- **Renamed from**: `DefaultFlagsNetworkManager`
- **Responsibilities**: Request execution only
  - Gets request from factory
  - Executes with OkHttp
  - Handles network errors (`UnknownHostException`, `IOException`, etc.)
  - Processes responses
- **Pattern**: Mirrors `DataOkHttpUploader` error handling

#### 3. EndpointsHelper
- **Refactored to**: Static methods (object)
- **Methods**: `getFlaggingEndpoint()`, `buildEndpointHost()`
- **Created internally**: By request factory when building requests

### Additional Notes

**Pattern Consistency:**
This refactoring aligns the Flags feature with the SDK's standard patterns:
- **Uploads**: `RequestFactory` → `DataOkHttpUploader`
- **Exposures**: `ExposuresRequestFactory` → `DataOkHttpUploader`
- **Precomputed Assignments** (this PR): `PrecomputedAssignmentsRequestFactory` → `PrecomputedAssignmentsDownloader`


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/PrecomputedAssignmentsRequestFactory.kt
```diff
@@ -0,0 +1,135 @@
+/*
```
typotter, line 1: Split out from DefaultFlagsNetworkManager

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+            InternalLogger.Level.ERROR,
+            InternalLogger.Target.MAINTAINER,
+            { "Failed to download flags: ${response.code}" }
+        )
+
+        null
+    }
+
+    private fun setupOkHttpClient() {
+        callFactory = OkHttpCallFactory {
```
aleksandr-gringauz: Maybe this is not a topic for this PR, but I would like to discuss it anyway.

Up until now we've had only one `OkHttpClient` in the whole SDK [here](https://github.com/DataDog/dd-sdk-android/blob/fa39327ad28cfe5a1792db2a7f966526e75a1aec/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L114).

If you create `OkHttpClient` the way you do here it will create its own thread pool and connection pool which is most likely is a waste of resources for no reason. See [this](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/index.html) doc for example.

I suggest we do something like [this](https://github.com/DataDog/dd-sdk-android/commit/2a0a7426c91bb20ebfb3b00a7040ee97d789a0cd) to share the underlying thread pool and connection pool between different `OkHttpClients` in our SDK. 

This also has the benefit that we can add some common settings for all clients (like cache). Not sure that we need it in this particular case though.

WDYT?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+ * @param flagsContext Context containing SDK configuration and authentication
+ * @param requestFactory Factory for creating precomputed assignments requests
+ */
+internal class PrecomputedAssignmentsDownloader(
+    private val internalLogger: InternalLogger,
+    private val flagsContext: FlagsContext,
+    private val requestFactory: PrecomputedAssignmentsRequestFactory
+) : FlagsNetworkManager {
+
+    internal lateinit var callFactory: OkHttpCallFactory
```
aleksandr-gringauz: `internal lateinit var callFactory: OkHttpCallFactory` -> `private val callFactory: OkHttpCallFactory`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
typotter: I think that's a great idea, and thank you for bringing it up here!

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
```diff
@@ -140,6 +140,15 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
         return NoOpScheduledExecutorService()
     }
 
+    override fun createOkHttpCallFactory(block: okhttp3.OkHttpClient.Builder.() -> Unit): okhttp3.Call.Factory {
+        return okhttp3.Call.Factory { request ->
```
aleksandr-gringauz: This isn't actually a "no-op" implementation. It will create a real `OkHttpClient`. You have to create a `NoOpCallFactory`.

Here it is:

```
object NoOpCallFactory: Call.Factory {
    
    object NoOpCall: Call {
        override fun cancel() {
            
        }

        override fun clone(): Call {
            return this
        }

        override fun enqueue(responseCallback: Callback) {
            
        }

        override fun execute(): Response {
            return Response.Builder().build()
        }

        override fun isCanceled(): Boolean {
            return false
        }

        override fun isExecuted(): Boolean {
            return false
        }

        override fun request(): Request {
            return Request.Builder().build()
        }

        override fun timeout(): Timeout {
            return Timeout.NONE
        }

    }
    override fun newCall(request: Request): Call {
        return NoOpCall
    }
}
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -270,49 +273,51 @@ interface FlagsClient {
...
                     flagsConfiguration
                 )
 
                 val flagsRepository = DefaultFlagsRepository(
                     featureSdkCore = sdkCore
                 )
 
-                val flagsNetworkManager = DefaultFlagsNetworkManager(
+                val callFactory = sdkCore.createOkHttpCallFactory {
+                    callTimeout(NETWORK_TIMEOUT_MS, TimeUnit.MILLISECONDS)
```
jonathanmos: we may no longer have to specify these timeouts/protocols because they are specified in the core with the same values

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
aleksandr-gringauz: Why are we using JSON here? You can create something like `data class SiteConfig(val dc: String, val tld: String)`.

And siteConfig can become `Map<DatadogSite, SiteConfig>`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
```diff
@@ -0,0 +1,88 @@
...
+    private val callFactory: Call.Factory,
+    private val internalLogger: InternalLogger,
+    private val flagsContext: FlagsContext,
+    private val requestFactory: PrecomputedAssignmentsRequestFactory
+) : PrecomputedAssignmentsReader {
+
+    @Suppress("ReturnCount", "TooGenericExceptionCaught")
+    override fun readPrecomputedFlags(context: EvaluationContext): String? {
+        val request = requestFactory.create(context, flagsContext) ?: run {
+            internalLogger.log(
```
aleksandr-gringauz: You already have error logging inside `PrecomputedAssignmentsRequestFactory`, this one seems redundant.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
aleksandr-gringauz: Great!

One more question. Do you think that `OkHttpClient` for the flags feature should have the same configuration as the one we already had in the SDK ([link](https://github.com/DataDog/dd-sdk-android/blob/a9a1d1e7ee5afe225c0bf357b69d796c831e52be/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L568))? Or it should be different?

It of course depends on the backend. As far as I see, the host for flags will be different than for the rest of the SDK ([this](https://github.com/DataDog/dd-sdk-android/blob/a9a1d1e7ee5afe225c0bf357b69d796c831e52be/dd-sdk-android-core/src/main/kotlin/com/datadog/android/DatadogSite.kt#L69)).

No strong opinion from me here.

0xnm: It should have the same configuration given Flags are supposed to work in GovCloud as well, which puts restrictions on the set of TLS suites allowed for the FIPS compliance.

typotter: just to note, Flags is not currently supported in GovCloud, (the intent is for it to be, but we've not scoped a GovCloud deploy yet). 

I think it's useful to use the same existing configuration and just allow certain overrides in the FlagsClient on parameters like timeout and number of retries. 
We're not speccing any customization in the API at current, so I think copying the existing configuration makes the most sense, especially given it is already hardened

typotter: (I'll update my code here)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
typotter: agree with you here. changed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
typotter: good catch, yes. I've simplified here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
typotter: thanks!
Would it make sense for the noop call factory to return the `request` it was created with?

aleksandr-gringauz: > Would it make sense for the noop call factory to return the request it was created with?

yes, this is better, indeed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/EndpointsHelper.kt
typotter: actually, #2922 is an even cleaner approach

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
```diff
@@ -0,0 +1,379 @@
...
+        // Shutdown server to simulate network failure
+        mockWebServer.shutdown()
+
+        // When
+        evaluationsManager.updateEvaluationsForContext(evaluationContext)
+
+        // Wait for async execution to complete
+        Thread.sleep(500)
+
+        // Then - When network fails, repository is not updated - error is logged by downloader
```
aleksandr-gringauz: There is no verification/assertions done in this test. Is it intended?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
```diff
@@ -0,0 +1,379 @@
...
+    @Mock
+    lateinit var mockFlagsRepository: FlagsRepository
+
+    private lateinit var mockWebServer: MockWebServer
+    private lateinit var evaluationsManager: EvaluationsManager
+    private lateinit var requestFactory: DefaultPrecomputedAssignmentsRequestFactory
+    private lateinit var downloader: PrecomputedAssignmentsDownloader
+    private lateinit var mapper: PrecomputeMapper
+    private lateinit var callFactory: okhttp3.Call.Factory
+    private val executorService = Executors.newSingleThreadExecutor()
```
aleksandr-gringauz: You can create an executor that executes tasks when they are scheduled like [this](https://github.com/DataDog/dd-sdk-android/blob/4a2b3e4e60c5aa716968f36413ffa31f31e2385f/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/FakeSameThreadExecutorService.kt#L13).

Then you'll not have to do `Thread.sleep(500)`.

The way test is written now - it can potentially be flacky.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
```diff
@@ -0,0 +1,379 @@
...
+        evaluationsManager.updateEvaluationsForContext(evaluationContext)
+
+        // Wait for async execution to complete
+        Thread.sleep(500)
+
+        // Then - Request was made
+        val recordedRequest = mockWebServer.takeRequest()
+        assertThat(recordedRequest.method).isEqualTo("POST")
+
+        // When response fails, repository is not updated - error is logged by downloader
```
aleksandr-gringauz: ok, but how does it affect the test? Does it fail?

features/dd-sdk-android-flags/async-set-context.md
```diff
@@ -0,0 +1,171 @@
+  Based on the test comments indicating async behavior expectations, here's a comprehensive plan to make setContext() return an async result:
```
jonathanmos: This file looks like it shouldn't be committed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/PrecomputedAssignmentsRequestFactoryTest.kt
```diff
@@ -0,0 +1,404 @@
...
+            site = DatadogSite.US1,
+            env = fakeEnv,
+            customFlagEndpoint = null
+        )
+
+        // When
+        val request = testedFactory.create(context, flagsContext)
+
+        // Then
+        assertThat(request).isNotNull
```
jonathanmos: minor: can use checkNotNull to avoid the !! - same for the other occurrences of this

features/dd-sdk-android-flags/async-set-context.md
0xnm: indeed, definitely shouldn't be a part of PR

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/feature/FeatureSdkCore.kt
```diff
@@ -115,6 +115,28 @@ interface FeatureSdkCore : SdkCore {
...
+     * val callFactory = sdkCore.createOkHttpCallFactory {
+     *     callTimeout(45, TimeUnit.SECONDS)
+     *     writeTimeout(30, TimeUnit.SECONDS)
+     * }
+     * ```
+     *
+     * @param block Configuration block to customize the [okhttp3.OkHttpClient.Builder]
+     * @return A [okhttp3.Call.Factory] instance configured with shared resources
+     */
+    fun createOkHttpCallFactory(block: okhttp3.OkHttpClient.Builder.() -> Unit): okhttp3.Call.Factory
```
0xnm:
```suggestion
    fun createOkHttpCallFactory(block: OkHttpClient.Builder.() -> Unit): Call.Factory
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
```diff
@@ -291,4 +300,48 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
...
+
+        override fun enqueue(responseCallback: okhttp3.Callback) = Unit
+
+        @Suppress("UnsafeThirdPartyFunctionCall") // All required fields are set, won't throw
+        override fun execute(): Response {
+            // Response.Builder.build() requires: request, protocol, code, and message
+            // to be set, otherwise it throws IllegalStateException
+            return Response.Builder()
+                .request(originalRequest)
+                .protocol(okhttp3.Protocol.HTTP_1_1)
```
0xnm:
```suggestion
                .protocol(Protocol.HTTP_1_1)
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
```diff
@@ -291,4 +300,48 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
...
+                .code(HttpURLConnection.HTTP_OK)
+                .message("OK")
```
0xnm, line 327: probably this one is not needed?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
```diff
@@ -291,4 +300,48 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
...
+        override fun isExecuted(): Boolean {
+            return false
+        }
+
+        override fun request(): Request {
+            return originalRequest
+        }
+
+        override fun timeout(): okio.Timeout {
+            return okio.Timeout.NONE
```
0xnm:
```suggestion
            return Timeout.NONE
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -179,10 +179,7 @@ internal class DatadogCoreTest {
     }
 
     @Test
-    fun `M register feature W registerFeature()`(
-        @Mock mockFeature: Feature,
-        @StringForgery fakeFeatureName: String
-    ) {
+    fun `M register feature W registerFeature()`(@Mock mockFeature: Feature, @StringForgery fakeFeatureName: String) {
```
0xnm: the previous formatting was actually better to read, maybe it is worth to revert formatting changes in this file?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -1461,10 +1422,50 @@ internal class DatadogCoreTest {
...
+
+        // When
+        val callFactory = testedCore.createOkHttpCallFactory {
+            callTimeout(45, TimeUnit.SECONDS)
+            writeTimeout(30, TimeUnit.SECONDS)
+        }
+
+        // Then
+        assertThat(callFactory).isNotNull()
+        val request = okhttp3.Request.Builder().url("https://example.com").build()
```
0xnm:
```suggestion
        val request = Request.Builder().url("https://example.com").build()
```

probably it is worth to do a search for `okhttp3.` occurrences outside of import statements in this PR.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -1461,10 +1422,50 @@ internal class DatadogCoreTest {
...
+        assertThat(callFactory).isNotNull()
+
+        // Verify the factory can create calls
+        val request = okhttp3.Request.Builder().url("https://example.com").build()
+        val call = callFactory.newCall(request)
+        assertThat(call).isNotNull()
+        assertThat(call.request()).isEqualTo(request)
+
+        // Verify custom timeout was applied
+        assertThat(call.timeout().timeoutNanos()).isGreaterThan(0)
```
0xnm: we can compare directly with `customTimeout`

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -1461,10 +1422,50 @@ internal class DatadogCoreTest {
...
+        val request = okhttp3.Request.Builder().url("https://example.com").build()
+        val call = callFactory.newCall(request)
+        assertThat(call).isNotNull()
+    }
+
+    @Test
+    fun `M create call factory with custom config W createOkHttpCallFactory() { with builder block }`() {
+        // Given
+        testedCore.initialize(fakeConfiguration)
+        val customTimeout = 60_000L
```
0xnm: we can randomize it in the range

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -1461,10 +1422,50 @@ internal class DatadogCoreTest {
...
-    ) : Runnable {
+    @Test
+    fun `M delegate to core feature W createOkHttpCallFactory()`() {
+        // Given
+        testedCore.initialize(fakeConfiguration)
+
+        // When
+        val callFactory = testedCore.createOkHttpCallFactory {
+            callTimeout(45, TimeUnit.SECONDS)
+            writeTimeout(30, TimeUnit.SECONDS)
```
0xnm: these 2 can also be randomized, to avoid clash with possible default values.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
```diff
@@ -1461,10 +1422,50 @@ internal class DatadogCoreTest {
...
+        }
+
+        // Then
+        assertThat(callFactory).isNotNull()
+
+        // Verify the factory can create calls
+        val request = okhttp3.Request.Builder().url("https://example.com").build()
+        val call = callFactory.newCall(request)
+        assertThat(call).isNotNull()
+        assertThat(call.request()).isEqualTo(request)
```
0xnm: seems like we are verifying OkHttp internals here.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -705,9 +705,7 @@ internal class CoreFeatureTest {
     }
 
     @Test
-    fun `M initialize only once W initialize() twice`(
-        @Forgery otherConfig: Configuration
-    ) {
+    fun `M initialize only once W initialize() twice`(@Forgery otherConfig: Configuration) {
```
0xnm: minor: same, probably formatting change can be reverted

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -1510,19 +1483,103 @@ internal class CoreFeatureTest {
...
+    @Test
+    fun `M apply custom configuration W createOkHttpCallFactory() { with custom timeouts }`() {
+        // Given
+        testedFeature.initialize(
+            appContext.mockInstance,
+            fakeSdkInstanceId,
+            fakeConfig,
+            fakeConsent
+        )
+        val customTimeout = 30_000L
```
0xnm: this can be randomized and also we can make its assertion below instead of using `.isGreaterThan(0)`

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -1510,19 +1483,103 @@ internal class CoreFeatureTest {
...
+        assertThat(callFactory1).isNotSameAs(callFactory2)
+
+        // Verify both factories can create calls independently
+        val request = okhttp3.Request.Builder().url("https://example.com").build()
+        val call1 = callFactory1.newCall(request)
+        val call2 = callFactory2.newCall(request)
+        assertThat(call1).isNotNull()
+        assertThat(call2).isNotNull()
+        assertThat(call1.request()).isEqualTo(request)
+        assertThat(call2.request()).isEqualTo(request)
```
0xnm: I'm not sure what exactly we are checking here

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -259,6 +261,7 @@ interface FlagsClient {
         // region Internal
 
         internal const val FLAGS_CLIENT_EXECUTOR_NAME = "flags-client-executor"
+        private val NETWORK_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(45)
```
0xnm: it is already 45 seconds by default https://github.com/DataDog/dd-sdk-android/blob/20de5e01eb16e0f15147b6718909aaa5ec413c64/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L725, we probably don't need to set it again

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -269,49 +272,51 @@ interface FlagsClient {
...
 
                 val flagsRepository = DefaultFlagsRepository(
                     featureSdkCore = sdkCore
                 )
 
-                val flagsNetworkManager = DefaultFlagsNetworkManager(
+                val callFactory = sdkCore.createOkHttpCallFactory {
+                    callTimeout(NETWORK_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+                    writeTimeout(NETWORK_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+                    protocols(listOf(Protocol.HTTP_2, Protocol.HTTP_1_1))
```
0xnm: we don't need this, it is defined in the `core` module

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultPrecomputedAssignmentsRequestFactory.kt
```diff
@@ -0,0 +1,129 @@
...
+ * This factory builds HTTP POST requests for fetching precomputed flag assignments
+ * from the Datadog Feature Flags service. It handles endpoint resolution, authentication
+ * headers, and JSON request body construction.
+ *
+ * @param internalLogger Logger for error and debug messages
+ */
+internal class DefaultPrecomputedAssignmentsRequestFactory(private val internalLogger: InternalLogger) :
+    PrecomputedAssignmentsRequestFactory {
+
+    @Suppress("ReturnCount", "TooGenericExceptionCaught")
```
0xnm: probably `TooGenericExceptionCaught` should be addressed instead of adding it to the suppress list.

why do we need to catch such wide type?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
```diff
@@ -0,0 +1,78 @@
...
+    override fun readPrecomputedFlags(context: EvaluationContext): String? {
+        val request = requestFactory.create(context, flagsContext) ?: return null
+
+        return executeDownloadRequest(request)
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    private fun executeDownloadRequest(request: Request): String? = try {
+        val response = callFactory.newCall(request).execute()
+        @Suppress("UnsafeThirdPartyFunctionCall") // wrapped in try/catch
```
0xnm: it seems `@Suppress` is redundant here, because `handleResponse` is not a 3rd party call

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
```diff
@@ -0,0 +1,379 @@
...
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.kotlin.argumentCaptor
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.verify
+import org.mockito.quality.Strictness
+import java.util.concurrent.Executors
+
+@ExtendWith(MockitoExtension::class, ForgeExtension::class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+internal class EvaluationsManagerIntegrationTest {
```
0xnm: Normally we don't keep integration tests in that folder/module, but I guess if it is executed fast, we can have it here.

Is integration test preferred here due to the complexity of unit testing?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultPrecomputedAssignmentsRequestFactoryTest.kt
```diff
@@ -0,0 +1,404 @@
...
+            site = DatadogSite.US1,
+            env = fakeEnv,
+            customFlagEndpoint = null
+        )
+
+        // When
+        val request = testedFactory.create(context, flagsContext)
+
+        // Then
+        assertThat(request).isNotNull
```
0xnm:
```suggestion
        checkNotNull(request)
```

it will save from `.?` usages below

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
typotter: `Response.Builder.build` will throw if any of request, protocol, or message is not set

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
typotter: agreed.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
typotter: done

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
typotter: done

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
typotter: done

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreTest.kt
typotter: testing just the create Call now.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
typotter: done

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
typotter: done

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
typotter: more than necessary... removed

features/dd-sdk-android-flags/async-set-context.md
typotter: removed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: cleaned up

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: cleaned up

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultPrecomputedAssignmentsRequestFactory.kt
typotter: no longer necessary

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/PrecomputedAssignmentsDownloader.kt
typotter: fixed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/PrecomputedAssignmentsRequestFactoryTest.kt
typotter: done

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
typotter: removing. The components here are already unit tested. We'll write e2e/integration tests for the whole client instead of just the Evaluations Manager.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
typotter: thanks!

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
typotter: removed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/evaluation/EvaluationsManagerIntegrationTest.kt
typotter: not really - was implicitly asserting the call doesn't throw. removed this suite in favour of higher-order testing

aleksandr-gringauz (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
```diff
@@ -340,6 +340,11 @@ internal class DatadogCore(
         return coreFeature.createScheduledExecutorService(executorContext)
     }
 
+    /** @inheritDoc */
+    override fun createOkHttpCallFactory(block: okhttp3.OkHttpClient.Builder.() -> Unit): okhttp3.Call.Factory {
```
0xnm: minor: `okhttp3` should go to import

```suggestion
    override fun createOkHttpCallFactory(block: OkHttpClient.Builder.() -> Unit): Call.Factory {
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -1512,6 +1512,90 @@ internal class CoreFeatureTest {
...
+        // When
+        val callFactory = testedFeature.createOkHttpCallFactory {
+            // Empty configuration block
+        }
+
+        // Then
+        assertThat(callFactory).isNotNull()
+
+        // Verify the factory can create calls
+        val request = okhttp3.Request.Builder().url("https://example.com").build()
```
0xnm:
```suggestion
        val request = Request.Builder().url("https://example.com").build()
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -1512,6 +1512,90 @@ internal class CoreFeatureTest {
...
+        val call = callFactory.newCall(request)
+        assertThat(call).isNotNull()
+        assertThat(call.request()).isEqualTo(request)
+
+        // Verify custom timeout was applied
+        assertThat(call.timeout().timeoutNanos()).isGreaterThan(0)
```
0xnm:
```suggestion
        assertThat(call.timeout().timeoutNanos()).isEqualTo(TimeUnit.SECONDS.toNanos(customTimeout))
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -1512,6 +1512,90 @@ internal class CoreFeatureTest {
...
+        val callFactory = testedFeature.createOkHttpCallFactory {
+            callTimeout(customTimeout, TimeUnit.MILLISECONDS)
+            writeTimeout(customTimeout, TimeUnit.MILLISECONDS)
+        }
+
+        // Then
+        assertThat(callFactory).isNotNull()
+
+        // Verify the factory can create calls
+        val request = okhttp3.Request.Builder().url("https://example.com").build()
```
0xnm:
```suggestion
        val request = Request.Builder().url("https://example.com").build()
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -320,9 +320,12 @@ interface FlagsClient {
                     NoOpFlagsRepository()
                 }
 
-                val flagsNetworkManager = DefaultFlagsNetworkManager(
+                val callFactory = sdkCore.createOkHttpCallFactory {}
```
0xnm: minor: maybe it is worth to use default lambda argument in the `createOkHttpCallFactory` signature? Otherwise such call with empty lambda looks a bit strange.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
0xnm, line 1: minor: seems like previous formatting in this file was better, looks like unnecessary change.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
```diff
@@ -27,44 +27,40 @@ internal class FlagsConfigurationTest {
...
+        @StringForgery fakeCustomFlagEndpoint: String
+    ) {
         // Given
         val builder = FlagsConfiguration.Builder()
 
         // When
-        builder.trackExposures(fakeTrackExposuresState)
+        builder.trackExposures(
+            fakeTrackExposuresState
+        ).useCustomExposureEndpoint(fakeCustomExposureEndpoint).useCustomFlagEndpoint(fakeCustomFlagEndpoint)
```
0xnm: minor: seems like `useCustomFlagEndpoint` call can go on another line. Formatting is off in this call chain.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -56,6 +60,7 @@ internal class FlagsFeatureTest {
     fun `set up`() {
         whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
         whenever(mockSdkCore.createSingleThreadExecutorService(any())) doReturn mockExecutorService
+//        whenever(mockSdkCore.getDatadogContext()) doReturn mockDatadogContext
```
0xnm: seems like this line should be removed and there is no actual usage of `mockDatadogContext` above?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/PrecomputedAssignmentsDownloaderTest.kt
```diff
@@ -0,0 +1,383 @@
...
+        whenever(mockCall.execute()).doReturn(fakeResponse)
+
+        // When
+        testedDownloader.readPrecomputedFlags(fakeEvaluationContext)
+
+        // Then
+        argumentCaptor<EvaluationContext> {
+            verify(mockRequestFactory).create(capture(), eq(fakeFlagsContext))
+            assertThat(lastValue).isEqualTo(fakeEvaluationContext)
+        }
```
0xnm: seems like argument captor is not really needed here and everything can be written simply as:

```suggestion
        verify(mockRequestFactory).create(fakeEvaluationContext, fakeFlagsContext)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/PrecomputedAssignmentsDownloaderTest.kt
```diff
@@ -0,0 +1,383 @@
...
+            verify(mockInternalLogger).log(
+                eq(InternalLogger.Level.ERROR),
+                eq(InternalLogger.Target.MAINTAINER),
+                any(),
+                capture(),
+                eq(false),
+                eq(null)
+            )
+            assertThat(lastValue).isInstanceOf(UnknownHostException::class.java)
+        }
```
0xnm:
```suggestion
        verify(mockInternalLogger).log(
                eq(InternalLogger.Level.ERROR),
                eq(InternalLogger.Target.MAINTAINER),
                isA<UnknownHostException>(),
                capture(),
                eq(false),
                eq(null)
          )
```

similar change can be done for the tests below

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/PrecomputedAssignmentsRequestFactoryTest.kt
0xnm: it actually still the same, the suggested change is not applied here and in the tests below

0xnm (dismissed):
lgtm, but I left some non-blocking suggestions

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsConfigurationTest.kt
typotter: fixed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
typotter: 🧹

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/PrecomputedAssignmentsRequestFactoryTest.kt
typotter: now done

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt
typotter: done

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: added to the interface. empty `()` certainly looks more normal here, thx

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/ExposureEventsProcessor.kt
typotter: fixed; I'll connect w/Jonathan on verifying my ktlint config

typotter (commented):
Thanks again for the review. Sorry for missing that handful of fixes.
The last commit to address comments requires another stamp. PTAL

0xnm (dismissed):

0xnm (approved):

PR #2918, opened 2025-10-07
RUM-11968: Implement Profiling Feature

ambushwork:
### What does this PR do?

* Implement `ProfilingFeature`
* Add `enable` API in `Profiling` object
* Add `ProfilingConfiguration`

### Motivation

RUM-11968

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
@@ -6,9 +6,38 @@
 
 package com.datadog.android.profiling
 
+import android.os.Build
+import androidx.annotation.RequiresApi
+import com.datadog.android.Datadog
+import com.datadog.android.api.SdkCore
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.profiling.internal.ProfilingFeature
+
 /**
  * An entry point to Datadog Profiling feature.
  */
 object Profiling {
-    // TODO RUM-11831: Integrate Perfetto Profiling Manager API
+
+    /**
+     * Enables the Perfetto based profiler to start recording callstack samples during application
+     * launch.
+     *
+     * @param configuration Configuration to use for the feature.
+     * @param sdkCore SDK instance to register feature in. If not provided, default SDK instance
+     * will be used.
+     */
+    @JvmStatic
+    @JvmOverloads
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    fun enable(
+        configuration: ProfilingConfiguration = ProfilingConfiguration.DEFAULT,
+        sdkCore: SdkCore = Datadog.getInstance()
+    ) {
+        val featureSdkCore = sdkCore as FeatureSdkCore
+        val profilingFeature = ProfilingFeature(
+            sdkCore = featureSdkCore,
+            configuration = configuration
+        )
+        featureSdkCore.registerFeature(profilingFeature)
+    }
 }
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
@@ -1,0 +1,49 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.profiling
+
+/**
+ * Describes configuration to be used for the Profiling feature.
+ */
+data class ProfilingConfiguration internal constructor(
+    internal val customEndpointUrl: String?
+) {
+
+    /**
+     * A Builder class for a [ProfilingConfiguration].
+     */
+    class Builder {
+
+        private var customEndpointUrl: String? = null
+
+        /**
+         * Let the Profiling feature target a custom server.
+         * The provided url should be the full endpoint url.
+         */
+        fun useCustomEndpoint(endpoint: String): Builder {
+            customEndpointUrl = endpoint
+            return this
+        }
+
+        /**
+         * Builds a [ProfilingConfiguration] based on the current state of this Builder.
+         */
+        fun build(): ProfilingConfiguration {
+            return ProfilingConfiguration(
+                customEndpointUrl = customEndpointUrl
+            )
+        }
+    }
+
+    companion object {
+
+        /**
+         * A default configuration for the Profiling feature.
+         */
+        val DEFAULT: ProfilingConfiguration = Builder().build()
+    }
+}
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
@@ -1,0 +1,77 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.profiling.internal
+
+import android.content.Context
+import android.os.Build
+import androidx.annotation.RequiresApi
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureEventReceiver
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.api.feature.StorageBackedFeature
+import com.datadog.android.api.net.RequestFactory
+import com.datadog.android.api.storage.FeatureStorageConfiguration
+import com.datadog.android.internal.time.DefaultTimeProvider
+import com.datadog.android.profiling.ProfilingConfiguration
+
+@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+internal class ProfilingFeature(
+    private val sdkCore: FeatureSdkCore,
+    configuration: ProfilingConfiguration
+) : StorageBackedFeature, FeatureEventReceiver {
+
+    private var perfettoProfiler: Profiler = NoOpProfiler()
+
+    private var dataWriter: ProfilingWriter = NoOpProfilingWriter()
+
+    override val requestFactory: RequestFactory = ProfilingRequestFactory(
+        configuration.customEndpointUrl,
+        sdkCore.internalLogger
+    )
+
+    override val storageConfiguration: FeatureStorageConfiguration
+        get() = FeatureStorageConfiguration.Companion.DEFAULT
+
+    override val name: String
+        get() = Feature.Companion.PROFILING_FEATURE_NAME
+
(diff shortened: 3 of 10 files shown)


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -6,9 +6,37 @@
...
+
+    /**
+     * Enables the Perfetto based profiler to start recording callstack samples during application
+     * launch.
+     *
+     * @param configuration Configuration to use for the feature.
+     * @param sdkCore SDK instance to register feature in. If not provided, default SDK instance
+     * will be used.
+     */
+    @JvmStatic
```
0xnm, line 29:
```suggestion
     */
    @JvmOverloads
    @JvmStatic
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -6,9 +6,37 @@
...
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
```
0xnm, line 31: do you think we should simply set `minSdk` to that for this module?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -6,9 +6,37 @@
...
+    fun enable(
+        configuration: ProfilingConfiguration = ProfilingConfiguration.DEFAULT,
```
0xnm, line 33: normally we don't have configuration as default argument in our features, but maybe indeed we can keep it like that

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
```diff
@@ -0,0 +1,53 @@
...
+                customEndpointUrl = customEndpointUrl
+            )
+        }
+    }
+
+    companion object {
+
+        /** Default builder to start building a new [ProfilingConfiguration]. */
+        @JvmStatic
+        fun builder(): Builder = Builder()
```
0xnm: is it needed given `ProfilingConfiguration.Builder()` can/must be called? Otherwise we have 2 APIs to do the same thing: `ProfilingConfiguration.Builder()` and `ProfilingConfiguration.builder()`

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    override fun onStop() {
+        perfettoProfiler.stop()
+        sdkCore.removeEventReceiver(name)
+    }
+
+    override fun onReceive(event: Any) {
+        // Do nothing for the moment.
+    }
```
0xnm, line 68: maybe should be added only when needed? in the next PRs

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
ambushwork: I don't think so, it will force clients to upgrade their application's `minSdk` for using profiling feature. But actually we expect that anyone can import profiling feature, and they are able to enable it only for end user's device above SDK 35

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
ambushwork: Currently we only have `customEndPoint` in the configuration, which is rarely used by clients, so I think it's better to provide default arg here so they can skip it

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
ambushwork: It's mostly reserved for java code convenience, otherwise on Java side you will need to call the constructor like this:
```
ProfilingConfiguration.Builder builder = new ProfilingConfiguration.Builder();
configuration = builder.useCustomEndPoint(url).build();
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
aleksandr-gringauz: Good question.

If you try (change minSdk of a module to 30 and compile the sample), you have something like this:

```
Manifest merger failed : uses-sdk:minSdkVersion 23 cannot be smaller than version 30 declared in library [:integrations:dd-sdk-android-sqldelight] /Users/aleksandr.gringauz/projects/dd-sdk-android/integrations/dd-sdk-android-sqldelight/build/intermediates/merged_manifest/debug/processDebugManifest/AndroidManifest.xml as the library might be using APIs not available in 23
	Suggestion: use a compatible library with a minSdk of at most 23,
		or increase this project's minSdk version to at least 30,
		or use tools:overrideLibrary="com.datadog.android.sqldelight" to force usage (may lead to runtime failures)
```

I'm almost sure it will be the same for the library when it isn't compiled, but downloaded. Because it is manifest merging.

The customer will have to use flavors (like [this](https://developer.android.com/build/build-variants#flavor-dimensions)) to use the profiling module.

But the question is - how to deal with these flavors when uploading to google play. I currently don't know, research needed.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
ambushwork: todo added here

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
0xnm: This will be almost the same with `builder()`, and you need to add `Companion` there:

```
ProfilingConfiguration.Builder builder = ProfilingConfiguration.Companion.builder();
configuration = builder.useCustomEndPoint(url).build();
```

I would suggest to keep only `ProfilingConfiguration.Builder()` to have a single way to create builder and align with similar API in other features.

ambushwork: Well, with `@JvmStatic` you can skip `Companion` and with this you can chain two line together:
```
ProfilingConfiguration.builder().useCustomEndpoint("url").build();
```
otherwise with constructor it looks not great in a chain:
```
(new ProfilingConfiguration.Builder()).useCustomEndpoint("url").build();
```

But it's just a small syntax sugar, we can remove it if you insist.

0xnm: yes, I think we should keep only `.Builder()` constructor, we are not optimizing for Java.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
0xnm: This is important point indeed, thanks for bringing it.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -0,0 +1,77 @@
...
+        get() = FeatureStorageConfiguration.Companion.DEFAULT
+
+    override val name: String
+        get() = Feature.Companion.PROFILING_FEATURE_NAME
+
+    override fun onInitialize(appContext: Context) {
+        perfettoProfiler = PerfettoProfiler(
+            internalLogger = sdkCore.internalLogger,
+            timeProvider = DefaultTimeProvider(),
+            profilingExecutor = sdkCore.createSingleThreadExecutorService(
```
aleksandr-gringauz, line 46: Not blocking for this PR and in general, just topic for discussion.

IMHO creating an executor here is too much. It is used only to call the callback with the result ([link](https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/Profiling/framework/java/android/os/ProfilingManager.java;l=241?q=public%20void%20requestProfiling)).

And for this we are creating a whole thread.

A thread right away eats 1MB of RAM ([link](https://developer.android.com/guide/practices/verifying-apps-art?utm_source=chatgpt.com#Stack_Size)). Maybe for modern devices it isn't much, but still. Also 1MB is probably virtual memory size, but here I am not entirely sure.

For this particular case I have 2 options:
1. `appContext.mainExecutor` - executor for main thread.
2. (requires more work and I am not 100% sure). Use "context thread" that is inside `coreFeature.contextExecutorService`. Probably not the best idea because it has backpressure logic.

The way I see it in the SDK we should have a "fixed" set of threads:
1. "Computational". No more than number of cores on device. More is useless.
2. I/O threads to do blocking I/O. Usually implemented using [ThreadPoolExecutor](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html). We can even put this executor into `OkHttpClient` using [this](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Dispatcher.html), so that OkHttp doesn't create its own thread pool inside ([link](https://github.com/square/okhttp/blob/d4a5be134ef9083a88a80a2e135ec6a730b49673/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Dispatcher.kt#L101)).

But, yes, this is a topic bigger than this PR, I don't propose fixing everything now.

ambushwork: Although most of the heavy operation of profiling is asynchronously executed in another thread, it's still recommended by Google that we use an non-UI thread for the result in case of ANR.
It's good to point out that we may have an memory overhead, but seeing that only API 35 can enable profiling, I can assume they are relatively new devices. 
So let's keep an eye on this, we can tweak it anytime if we see an issue.

aleksandr-gringauz (approved):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
aleksandr-gringauz: > recommended by Google that we use an non-UI thread for the result in case of ANR

fair point

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
0xnm: Let's add TODO item for now.

ambushwork: What's the thing to do here? do we want to bump the `minSdk` for this?

0xnm (approved):

PR #2922, opened 2025-10-08
feat: DatadogSite extension for Flags endpoint

typotter:
### What does this PR do?

This PR changes the `EndpointsHelper` class into a simpler extension on the `DatadogSite` class, more closely mirroring the manner in which `intakeEndpoints` are created, following SDK idioms and improving code organization.

**Key Changes:**
- Replaces `EndpointsHelper` class with `DatadogSite.getFlagsEndpoint()` extension function
- Adds validation for empty/blank customer domains with proper error logging
- Updates `DefaultFlagsNetworkManager` to use the extension directly

### Motivation

The original `EndpointsHelper` class was a utility that didn't follow Kotlin best practices. Since endpoint construction is intrinsically tied to the `DatadogSite` enum, using an extension function:

1. **Aligns with existing patterns**: Matches the `DatadogSite.intakeEndpoint` property pattern used throughout the SDK
2. **Improves discoverability**: Extension functions are easier to find and use via IDE autocomplete
3. **Reduces boilerplate**: Eliminates the need for a separate helper class with static methods
4. **Better encapsulation**: Keeps site-specific endpoint logic close to the `DatadogSite` type

### Additional Notes

**Architecture Decisions:**
- The extension returns the complete URL (including protocol and path) to maintain consistency with `DatadogSite.intakeEndpoint`
- Empty/blank customer domain validation logs an error and returns `null`, following SDK conventions for graceful degradation at runtime
- US1_FED site returns `null` with appropriate error logging as it's not yet supported for flags

**Test Improvements:**
- Parameterized tests use `@StringForgery` for customer domains
- Added test for blank customer domain (whitespace) validation


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Extension function that returns the flags CDN host for this Datadog site.
+ * @param customerDomain The customer-specific subdomain prefix for the flags CDN
+ * @param internalLogger Logger for errors
+ * @return Flags host string in format `<customerDomain>.ff-cdn.<site>.<tld>`, or null if site not supported
+ */
+private fun DatadogSite.flagsHost(customerDomain: String, internalLogger: InternalLogger): String? = when (this) {
+    DatadogSite.US1_FED -> {
+        internalLogger.log(
```
typotter: q: Would it make more sense here to throw an exception to be caught by the caller and have logging happen outside this extension?

aleksandr-gringauz: My opinion, can be different for others.

I would just return `null` in case of `DatadogSite.US1_FED`. Without any logging in this `flagsHost` function and without throwing exceptions.

If we don't support `US1_FED` it can be checked somewhere at the entry point of the feature (like `Flags.enable`) and logged once. No need to log it each time `flagsHost` is called.

Also I think it is a good practice to try to make functions pure (make them only compute input from output without any side effects like logging or throwing exceptions). It isn't possible and reasonable everywhere, but in this particular case when we just construct an url based on some inputs IMHO all of this should be pure functions. If there are some error edge cases, return `null` for errors and let the caller deal with this.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
```diff
@@ -0,0 +1,71 @@
...
+    DatadogSite.EU1 -> buildFlagsHostString(customerDomain, tld = "eu")
+    DatadogSite.US1 -> buildFlagsHostString(customerDomain)
+
+    DatadogSite.US3 -> buildFlagsHostString(customerDomain, "us3", "com")
+    DatadogSite.US5 -> buildFlagsHostString(customerDomain, "us5", "com")
+    DatadogSite.AP1 -> buildFlagsHostString(customerDomain, "ap1", "com")
+    DatadogSite.AP2 -> buildFlagsHostString(customerDomain, "ap2", "com")
+}
+
+private fun buildFlagsHostString(customerDomain: String, dc: String? = null, tld: String = "com"): String {
```
aleksandr-gringauz, line 40: very nit:

I would just make `dc: String` a non-optional parameter and for `DatadogSite.US1` call `buildFlagsHostString(customerDomain, "")`. This `isNotEmpty` checking seems like overcomplicating things.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Gets the complete flags endpoint URL.
+ * @param customerDomain The customer-specific subdomain prefix for the flags CDN (default: "preview").
+ * This is used to construct the full host in the format: `<customerDomain>.ff-cdn.<site>.<tld>`
+ * @param internalLogger Logger for errors
+ * @return Complete flags endpoint URL or null if site not supported or customerDomain is empty
+ */
+@Suppress("ReturnCount")
+internal fun DatadogSite.getFlagsEndpoint(customerDomain: String = "preview", internalLogger: InternalLogger): String? {
```
aleksandr-gringauz: IMHO

```
internal fun DatadogSite.getFlagsEndpoint(customerDomain: String = "preview"): String? {
    val host = flagsHost(customerDomain) ?: return null
    return "https://$host$FLAGS_PATH"
}
```

I would just do it like this. `customerDomain.isBlank()` can be done somewhere when this `customerDomain` is set first.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
```diff
@@ -16,10 +16,12 @@ import com.datadog.android.DatadogSite
...
  * @param env The environment name (e.g., prod, staging) for context
+ * @param customerDomain The customer domain prefix for the flags endpoint (defaults to "preview")
  */
 internal data class FlagsContext(
     val applicationId: String?,
     val clientToken: String,
     val site: DatadogSite,
-    val env: String
+    val env: String,
+    val customerDomain: String = "preview"
```
0xnm, line 26: why do we need to default it to `preview`? Are we going to use non-stable API at the module release?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
0xnm: is it worth to repeat default value here if it is also in the `FlagsContext`?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
0xnm: We avoid throwing exceptions in the SDK, because if they are forgot to be caught, they will crash host app.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensionsTest.kt
```diff
@@ -0,0 +1,174 @@
...
+        val result =
+            DatadogSite.US1_FED.getFlagsEndpoint(customerDomain, mockInternalLogger)
+
+        // Then
+        assertThat(result).isNull()
+        verify(mockInternalLogger).log(
+            eq(InternalLogger.Level.ERROR),
+            eq(InternalLogger.Target.USER),
+            org.mockito.kotlin.argThat { invoke() == "US1_FED is not yet supported for flags endpoints" },
+            org.mockito.kotlin.anyOrNull(),
```
jonathanmos: minor: should remove all the inline packages in the file and just import anyOrNull/any/etc.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensionsTest.kt
```diff
@@ -0,0 +1,174 @@
...
+            Arguments.of(DatadogSite.US1, "ff-cdn.datadoghq.com"),
+            Arguments.of(DatadogSite.US3, "ff-cdn.us3.datadoghq.com"),
+            Arguments.of(DatadogSite.US5, "ff-cdn.us5.datadoghq.com"),
+            Arguments.of(DatadogSite.AP1, "ff-cdn.ap1.datadoghq.com"),
+            Arguments.of(DatadogSite.AP2, "ff-cdn.ap2.datadoghq.com"),
+            Arguments.of(DatadogSite.EU1, "ff-cdn.datadoghq.eu"),
+            Arguments.of(DatadogSite.STAGING, "ff-cdn.datad0g.com")
+        )
+
+        @Suppress("unused")
```
jonathanmos, line 101: why do we want this if it's unused? Is it preparation for a later change? (same question for the next block also)

typotter: It gets picked up as "unused", it is is actually used. It's just referenced by string name in the parameterized annotation `@MethodSource("supportedSitesWithDefaultDomain")`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
typotter: Thanks for the feedback and guidance here, friends!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
typotter: because we have the dot separator after DC, we do need to allow a "not-present" value. Using `listOfNotNull` and then string joining reads a lot cleaner.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/FlagsContext.kt
typotter: The customer domain can currently be anything. We are using `preview` across the SDKs for now until we have a mechanism to produce a per-customer domain. The per-customer domain becomes an offboarding mechanism; traffic from these SDKs would be blocked by DNS instead of hitting any of our compute or caching endpoints. It's an extreme switch to flip, but we found it would have been useful to have it implemented earlier at Eppo.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
typotter: no. `FlagsContext` is the correct place to keep this value.

typotter: 👍

typotter (commented):
Thanks! Made updates. please take another look.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogSiteExtensions.kt
aleksandr-gringauz: got it

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/repository/net/DefaultFlagsNetworkManager.kt
```diff
@@ -214,7 +199,7 @@ internal class DefaultFlagsNetworkManager(
     // TODO replace the environment
     private fun buildEnvPayload(): JSONObject =
         JSONObject()
-            .put("dd_env", "prod")
```
0xnm, line 217: probably the comment above `// TODO replace the environment` is not valid anymore?

0xnm (approved):

aleksandr-gringauz (approved):

PR #2923, opened 2025-10-09
[flags] Generic resolve method

typotter:
# PR #2923: [flags] Generic resolve method

## What does this PR do?

This PR implements the generic `resolve<T>()` method providing OpenFeature-compatible flag resolution with detailed error information, metadata, and variant tracking.

**Key additions:**
_public API_
- Generic `resolve<T>(flagKey, defaultValue)` method that returns `ResolutionDetails<T>`
- `ResolutionDetails<T>` data class containing value, variant, reason, error code, error message, and flag metadata
- `ErrorCode` enum with OpenFeature-spec error codes (FLAG_NOT_FOUND, TYPE_MISMATCH, PARSE_ERROR, etc.)

_internal_
- `FlagValueConverter` object for centralized, side-effect-free type conversion
- Enhanced error handling and type mismatch detection
- Improved internal architecture with `InternalResolution` sealed class

### OpenFeature Specification Alignment

The OpenFeature specification defines a `resolve()` method as the core evaluation API, providing rich resolution details beyond just the flag value. This enables:

- **Error handling**: Applications can detect and handle flag resolution failures (missing flags, type mismatches, parse errors)
- **Variant tracking**: Track which variant of a flag was returned for experimentation
- **Metadata support**: Access additional flag metadata for analytics and debugging
- **Debugging**: Detailed resolution reasons help troubleshoot flag behavior

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClientTest.kt
+++ b/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClientTest.kt
@@ -163,7 +163,8 @@ internal class DatadogFlagsClientTest {
         val fakeDefaultValue = !fakeFlagValue
         val fakeFlag = forge.getForgery<PrecomputedFlag>().copy(
             variationType = VariationType.BOOLEAN.value,
-            variationValue = fakeFlagValue.toString()
+            variationValue = fakeFlagValue.toString(),
+            doLog = true
         )
         val fakeEvaluationContext = EvaluationContext(
             targetingKey = forge.anAlphabeticalString(),
@@ -235,7 +236,8 @@ internal class DatadogFlagsClientTest {
         val fakeDefaultValue = !fakeFlagValue
         val fakeFlag = forge.getForgery<PrecomputedFlag>().copy(
             variationType = VariationType.BOOLEAN.value,
-            variationValue = fakeFlagValue.toString()
+            variationValue = fakeFlagValue.toString(),
+            doLog = true
         )
         val fakeEvaluationContext = EvaluationContext(
             targetingKey = forge.anAlphabeticalString(),
@@ -697,7 +699,8 @@ internal class DatadogFlagsClientTest {
         val fakeFlagValue = forge.anAlphabeticalString()
         val fakeFlag = forge.getForgery<PrecomputedFlag>().copy(
             variationType = VariationType.STRING.value,
-            variationValue = fakeFlagValue
+            variationValue = fakeFlagValue,
+            doLog = true
         )
         val fakeContext = EvaluationContext(
             targetingKey = forge.anAlphabeticalString(),
@@ -858,7 +861,8 @@ internal class DatadogFlagsClientTest {
         val fakeDefaultValue = !fakeFlagValue
         val fakeFlag = forge.getForgery<PrecomputedFlag>().copy(
             variationType = VariationType.BOOLEAN.value,
-            variationValue = fakeFlagValue.toString()
+            variationValue = fakeFlagValue.toString(),
+            doLog = true
         )
         val fakeEvaluationContext = EvaluationContext(
             targetingKey = forge.anAlphabeticalString(),
@@ -893,7 +897,8 @@ internal class DatadogFlagsClientTest {
         val fakeDefaultValue = !fakeFlagValue
         val fakeFlag = forge.getForgery<PrecomputedFlag>().copy(
             variationType = VariationType.BOOLEAN.value,
-            variationValue = fakeFlagValue.toString()
+            variationValue = fakeFlagValue.toString(),
+            doLog = true
         )
         val fakeEvaluationContext = EvaluationContext(
             targetingKey = forge.anAlphabeticalString(),
--- a/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/utils/forge/PrecomputedFlagForgeryFactory.kt
+++ b/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/utils/forge/PrecomputedFlagForgeryFactory.kt
@@ -16,16 +16,7 @@ import org.json.JSONObject
 internal class PrecomputedFlagForgeryFactory : ForgeryFactory<PrecomputedFlag> {
     override fun getForgery(forge: Forge): PrecomputedFlag {
         val variationTypeEnum = forge.anElementFrom(*VariationType.values())
-
-        // Use the same pattern as DatadogFlagsProviderTest
-        val variationType = when (variationTypeEnum) {
-            VariationType.BOOLEAN -> VariationType.BOOLEAN.value
-            VariationType.STRING -> VariationType.STRING.value
-            VariationType.INTEGER -> VariationType.INTEGER.value
-            VariationType.NUMBER -> VariationType.NUMBER.value
-            VariationType.FLOAT -> VariationType.FLOAT.value
-            VariationType.OBJECT -> VariationType.OBJECT.value
-        }
+        val variationType = variationTypeEnum.value
 
         val variationValue = when (variationTypeEnum) {
             VariationType.BOOLEAN -> forge.aBool().toString()
@@ -45,7 +36,7 @@ internal class PrecomputedFlagForgeryFactory : ForgeryFactory<PrecomputedFlag> {
         return PrecomputedFlag(
             variationType = variationType,
             variationValue = variationValue,
-            doLog = true, // Default to true - tests can override if needed
+            doLog = forge.aBool(),
             allocationKey = forge.anAlphabeticalString(),
             variationKey = forge.anAlphabeticalString(),
             extraLogging = forge.aNullable {


features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClientTest.kt
```diff
@@ -105,7 +105,11 @@ internal class DatadogFlagsClientTest {
             variationType = VariationType.BOOLEAN.value,
             variationValue = fakeFlagValue.toString()
         )
-        whenever(mockFlagsRepository.getPrecomputedFlag(fakeFlagKey)) doReturn fakeFlag
+        val fakeContext = EvaluationContext(
+            targetingKey = forge.anAlphabeticalString(),
+            attributes = emptyMap()
+        )
+        whenever(mockFlagsRepository.getPrecomputedFlagWithContext(fakeFlagKey)) doReturn (fakeFlag to fakeContext)
```
typotter, line 113: return both the flag assignment and context as a pair.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ResolutionDetails.kt
```diff
@@ -0,0 +1,30 @@
...
+ * @param T The type of the resolved flag value (Boolean, String, Int, Double, JSONObject).
+ * @property value The resolved flag value. This is always present, either from flag evaluation or the default value.
+ * @property variant Optional string identifier for the resolved variant (e.g., "control", "treatment").
+ * @property reason Optional string explaining why this value was resolved (e.g., "TARGETING_MATCH", "DEFAULT", "ERROR").
+ * @property errorCode Optional error code if the resolution failed. Null indicates successful resolution.
+ * @property errorMessage Optional human-readable error message providing additional context about failures.
+ * @property flagMetadata Optional map of arbitrary metadata associated with the flag (string keys, primitive values).
+ */
+data class ResolutionDetails<T>(
+    val value: T,
```
aleksandr-gringauz, line 24: Can value be null? If not, it is better to do `data class ResolutionDetails<T: Any>`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -73,144 +75,356 @@ internal class DatadogFlagsClient(
...
+    private fun <T> resolveTracked(resolution: InternalResolution<T>): T = when (resolution) {
+        is InternalResolution.Success -> {
+            if (flagsConfiguration.trackExposures && resolution.flag.doLog) {
+                writeExposureEvent(resolution.flagKey, resolution.flag, resolution.context)
+            }
+
+            if (flagsConfiguration.rumIntegrationEnabled && resolution.flag.doLog) {
+                logEvaluation(
+                    key = resolution.flagKey,
+                    value = resolution.value as Any
```
aleksandr-gringauz: If value is never null, you should make `T: Any` everywhere like I described here https://github.com/DataDog/dd-sdk-android/pull/2923/files#r2444011674

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ErrorCode.kt
```diff
@@ -0,0 +1,52 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.flags.model
+
+/**
+ * Error codes for flag resolution failures, aligned with the OpenFeature specification.
+ */
+enum class ErrorCode {
```
aleksandr-gringauz, line 12: I only see that `FLAG_NOT_FOUND`, `PARSE_ERROR` and `TYPE_MISMATCH` are used in this PR. Will others be used in other PRs or are they not needed?

features/dd-sdk-android-flags/api/apiSurface
```diff
@@ -15,10 +15,22 @@ interface com.datadog.android.flags.featureflags.FlagsClient
   fun resolveDoubleValue(String, Double): Double
   fun resolveIntValue(String, Int): Int
   fun resolveStructureValue(String, org.json.JSONObject): org.json.JSONObject
+  fun <T> resolve(String, T): com.datadog.android.flags.model.ResolutionDetails<T>
```
0xnm: Specification says that even for concrete methods above we need to return `ResolutionDetails` https://openfeature.dev/specification/sections/providers, but we are returning just the objects. Is it fine to not use this structure there?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ResolutionDetails.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Details about the resolution of a feature flag, aligned with the OpenFeature specification.
+ *
+ * This structure contains the resolved flag value along with metadata about how the value
+ * was determined. It provides rich diagnostic information for debugging and analytics.
+ *
+ * @param T The type of the resolved flag value (Boolean, String, Int, Double, JSONObject).
+ * @property value The resolved flag value. This is always present, either from flag evaluation or the default value.
+ * @property variant Optional string identifier for the resolved variant (e.g., "control", "treatment").
+ * @property reason Optional string explaining why this value was resolved (e.g., "TARGETING_MATCH", "DEFAULT", "ERROR").
```
aleksandr-gringauz: > e.g., "TARGETING_MATCH", "DEFAULT", "ERROR"

Is it possible to make this a enum instead of a String?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -114,6 +115,40 @@ interface FlagsClient {
...
+     *     println("Experiment metadata: $metadata")
+     * }
+     * ```
+     *
+     * @param T The type of the flag value (Boolean, String, Int, Double, or JSONObject).
+     * @param flagKey The unique identifier of the flag to resolve.
+     * @param defaultValue The value to return if the flag cannot be retrieved or parsed.
+     * @return [ResolutionDetails] containing the value, variant, reason, error info, and metadata.
+     */
+    fun <T> resolve(flagKey: String, defaultValue: T): ResolutionDetails<T>
```
0xnm: Should we support nullable types (values) here? If it is not the case, type should be `T : Any`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,103 @@
...
+     * @param defaultValue The default value (used to determine the type)
+     * @return The type name as a string (e.g., "Boolean", "String", "Int")
+     */
+    fun <T> getTypeName(defaultValue: T): String = when (defaultValue) {
+        is Boolean -> "Boolean"
+        is String -> "String"
+        is Int -> "Int"
+        is Double -> "Double"
+        is JSONObject -> "JSONObject"
+        else -> defaultValue!!::class.simpleName ?: "Unknown"
```
jonathanmos: Theoretically someone could pass null for T. In the current codepaths this would possibly be blocked at `isTypeCompatible`, but I think it's probably better to avoid the !!. We could do something like:

```defaultValue?.let { it.class.simpleName ?: "Unknown" } ?: "Unknown"```

or even directly return "Unknown" depending how important it is to know the exact class name.

File: features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt [hunk 9]
```diff
@@ -73,144 +75,356 @@ internal class DatadogFlagsClient(
...
+     * Resolves a string flag value.
      *
-     * This method returns the stored string value as-is without any type conversion.
-     * This method is thread-safe and performs no network operations.
+     * If the flag cannot be found, is not a string, or an error occurs, the default value is returned.
      *
      * @param flagKey The name of the flag to query. Cannot be null.
-     * @param defaultValue The value to return if the flag cannot be retrieved.
-     * @return The string value of the flag, or the default value if unavailable.
+     * @param defaultValue The value to return if the flag cannot be found resolved for any reason. Cannot be null.
```
0xnm:
```suggestion
     * @param defaultValue The value to return if the flag cannot be found or resolved for any reason. Cannot be null.
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -73,144 +75,356 @@ internal class DatadogFlagsClient(
...
+     * This helper method simplifies the convenience methods by encapsulating the common pattern
+     * of resolving a flag internally and then processing it through the tracking layer.
+     *
+     * @param T The type of the flag value
+     * @param flagKey The name of the flag to query
+     * @param defaultValue The default value to return if resolution fails
+     * @return The resolved value or the default value
+     */
+    private fun <T> resolveValue(flagKey: String, defaultValue: T): T =
+        resolveTracked(resolveInternal(flagKey, defaultValue))
```
0xnm: I think this is confusing: here we have 3 `resolveXXX` methods, and the fact that there is `resolveInternal` with a result passed to `resolveTracked` doesn't add clarity. Should `resolveInternal` be renamed, maybe?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -73,144 +75,356 @@ internal class DatadogFlagsClient(
...
+     * - Int: Parse to integer
+     * - Double: Parse to double
+     * - JSONObject: Parse JSON with error logging
+     *
+     * @param T The type of the flag value (Boolean, String, Int, Double, or JSONObject).
+     * @param flagKey The name of the flag to query.
+     * @param defaultValue The value to return if the flag cannot be retrieved or parsed.
+     * @return [ResolutionDetails] with either the parsed value and metadata, or an error.
+     */
+    override fun <T> resolve(flagKey: String, defaultValue: T): ResolutionDetails<T> {
```
0xnm: minor: it is better to move this up, so that we have interface methods grouped first and only after them we have `private` methods.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -73,144 +75,356 @@ internal class DatadogFlagsClient(
...
+     *
+     * This helper method converts the extraLogging JSONObject from PrecomputedFlag
+     * into a Map<String, Any> suitable for the ResolutionDetails metadata field.
+     * Only string, number, and boolean values are included; other types are omitted.
+     * The returned map is immutable to prevent external modifications.
+     *
+     * @param extraLogging The JSONObject containing additional flag metadata.
+     * @return An immutable map of metadata, or null if the JSONObject is empty.
+     */
+    private fun extractMetadata(extraLogging: JSONObject): Map<String, Any>? {
```
0xnm: do we need to explicitly make return type nullable? Cannot we return empty collection if `JSONObject` is empty?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,103 @@
...
+        val converter = getConverterForType(defaultValue)
+        return try {
+            converter(variationValue)
+        } catch (
+            @Suppress("TooGenericExceptionCaught", "SwallowedException")
+            e: Exception
+        ) {
+            // Catch all conversion exceptions (including JSONException) and return null
+            // The caller is responsible for logging errors as appropriate
+            null
```
0xnm: maybe it is worth to log it with `MAINTAINER` target?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param T The expected type
+     * @param variationType The type specified in the flag metadata (e.g., "boolean", "string", "integer")
+     * @param defaultValue The default value (used to determine the expected type)
+     * @return true if types are compatible, false otherwise
+     */
+    fun <T> isTypeCompatible(variationType: String, defaultValue: T): Boolean = when (defaultValue) {
+        is Boolean -> variationType == VariationType.BOOLEAN.value
+        is String -> variationType == VariationType.STRING.value
+        is Int -> variationType == VariationType.INTEGER.value
```
0xnm: should we also check for `VariationType.NUMBER` here as below?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Converts a flag's string value to the expected type.
+     *
+     * @param T The expected type of the flag value
+     * @param variationValue The raw string value from the flag
+     * @param variationType The type specified in the flag metadata
+     * @param defaultValue The default value (used to infer the target type)
+     * @return The converted value, or null if types are incompatible or conversion fails
+     */
+    @Suppress("UNCHECKED_CAST")
```
aleksandr-gringauz, line 42: > @Suppress("UNCHECKED_CAST")

this is not needed here

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,103 @@
...
+     * Handles special cases like JSONObject parsing with error logging.
+     *
+     * @param T The expected type
+     * @param defaultValue The default value (used to determine the target type)
+     * @return A function that converts a string to type T, or null if conversion fails
+     */
+    @Suppress("UNCHECKED_CAST")
+    private fun <T> getConverterForType(defaultValue: T): (String) -> T? = when (defaultValue) {
+        is Boolean -> { s: String -> s.lowercase(Locale.US).toBooleanStrictOrNull() as? T }
+        is String -> { s: String -> s as? T }
```
0xnm: I think this one can be just `{ s: String -> s  }`: input and output types are the same, so cast is not needed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/PrecomputedFlagConstants.kt
```diff
@@ -30,8 +30,9 @@ internal enum class VariationType(val value: String) {
     BOOLEAN("boolean"),
     STRING("string"),
     INTEGER("integer"),
-    DOUBLE("double"),
-    JSON("json")
+    NUMBER("number"),
+    FLOAT("float"),
```
0xnm, line 34: we still have the usage of `double` term in API, should we keep it here? Or is it the change per spec?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
aleksandr-gringauz: `else -> defaultValue?.let { it::class.simpleName} ?: "Unknown"`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,103 @@
...
+     * Converts a flag's string value to the expected type.
+     *
+     * @param T The expected type of the flag value
+     * @param variationValue The raw string value from the flag
+     * @param variationType The type specified in the flag metadata
+     * @param defaultValue The default value (used to infer the target type)
+     * @return The converted value, or null if types are incompatible or conversion fails
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun <T> convert(variationValue: String, variationType: String, defaultValue: T): T? {
```
aleksandr-gringauz: nit

You don't actually need a defaultValue here. You can pass the class itself using `KClass`.


```
internal object FlagValueConverter {

    fun <T : Any> convert(variationValue: String, variationType: String, resultType: KClass<T>): T? {
        if (!isTypeCompatible(variationType, resultType)) {
            return null
        }
        return try {
            doConvert(variationValue, resultType)
        } catch (
            @Suppress("TooGenericExceptionCaught", "SwallowedException")
            e: Exception
        ) {
            null
        }
    }

    fun isTypeCompatible(variationType: String, resultType: KClass<*>): Boolean = when (resultType) {
        Boolean::class -> variationType == VariationType.BOOLEAN.value
        String::class -> variationType == VariationType.STRING.value
        Int::class -> variationType == VariationType.INTEGER.value
        Double::class -> variationType == VariationType.NUMBER.value || variationType == VariationType.FLOAT.value
        JSONObject::class -> variationType == VariationType.OBJECT.value
        else -> false
    }
    
    @Suppress("UNCHECKED_CAST")
    private fun <T : Any> doConvert(s: String, resultType: KClass<T>): T? = when (resultType) {
        Boolean::class -> s.lowercase(Locale.US).toBooleanStrictOrNull() as? T
        String::class -> s as? T
        Int::class -> s.toIntOrNull() as? T
        Double::class -> s.toDoubleOrNull() as? T
        JSONObject::class -> JSONObject(s) as? T
        else -> null
    }

    fun getTypeName(type: KClass<*>): String = when (type) {
        Boolean::class -> "Boolean"
        String::class -> "String"
        Int::class -> "Int"
        Double::class -> "Double"
        JSONObject::class -> "JSONObject"
        else -> type.simpleName ?: "Unknown"
    }
}
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: removed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: this is great, thank you

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: I was on the fence about this one b/c I wanted to keep the singleton. It would be useful to get a handle on any exception that happens here. I just discovered Kotlin's `Result`, so using that instead

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: yes, you're right. The `number` type that the backend is sending will evolve into sending specifically a `float` or an `integer` variant type so it should be usable as either now.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: 👍

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: Thanks for the explanation. With the refactor to `T: any` and `KClass`, I think this is now addressed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/model/PrecomputedFlagConstants.kt
typotter: These are the types that are sent over the wire. We actually need only one numeric resolution function - the `number` type is what is required for open feature. We'll treat `number`, `float`, and `integer` as satisfying the "resolveDouble" method. `integer` type and `numeric` type will satisfy `resolveInteger`

features/dd-sdk-android-flags/api/apiSurface
typotter: We are not implementing the OpenFeature specification, rather we're exposing enough to allow a shim package to implement the OpenFeature Provider spec. This is the `resolve` method.
The concrete methods above mirror the OpenFeature `Client` interface, which the dev uses in their app. This is to make this SDK still be functional and useful, even without OpenFeature layered on top.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: non null all the way.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
typotter: reworded

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
typotter: yes, I agree. 
I renamed to `readAndParseAssignment`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
typotter: done

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
typotter: good idea. there is no reason to return null

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ResolutionDetails.kt
typotter: thanks for the tip.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
typotter: thanks again

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ErrorCode.kt
typotter: pruned!

typotter (commented):
Thanks for the review. ready for another look

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -195,20 +201,270 @@ internal class DatadogFlagsClient(
...
+                    { "Flag '${resolution.flagKey}': ${resolution.errorMessage}" }
+                )
+            }
+            resolution.defaultValue
+        }
+    }
+
+    // endregion
+
+// region Helper Methods
```
0xnm:
```suggestion
  // region Helper Methods
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -195,20 +201,270 @@ internal class DatadogFlagsClient(
...
+
+    /**
```
0xnm: nit: it seems doc comments like this and below don't add much value, so maybe we can remove them to have more focus on the code itself?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -195,20 +201,270 @@ internal class DatadogFlagsClient(
...
+        extraLogging.keys().forEach { key ->
+            val value = extraLogging.opt(key)
+            // Only include primitive types as per OpenFeature spec
+            when (value) {
+                is String, is Number, is Boolean -> metadata[key] = value
+                // Skip other types (JSONObject, JSONArray, null)
+            }
+        }
+
+        return if (metadata.isNotEmpty()) metadata.toMap() else emptyMap()
```
0xnm: We can just return `metadata` directly, because return type definition is immutable anyway, no need to make a copy.

```suggestion
        return metadata
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -195,20 +201,270 @@ internal class DatadogFlagsClient(
...
     }
+
+// endregion
```
0xnm:
```suggestion
  // endregion
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param T The expected type of the flag value
+     * @param variationValue The raw string value from the flag
+     * @param variationType The type specified in the flag metadata
+     * @param targetType The target type class to convert to
+     * @return Result containing the converted value on success,
+     *         or a Failure with TypeMismatchException if types are incompatible,
+     *         or a Failure with the parse exception if conversion failed
+     */
+    fun <T : Any> convert(variationValue: String, variationType: String, targetType: KClass<*>): Result<T?> {
```
0xnm: should it be `Result<T>` instead of `Result<T?>`? If you do this change, you don't need `parsedValue!!` in the `DatadogFlagsClient.kt`, probably.

0xnm: I also think maybe it is possible to change `targetType: KClass<*>` to `targetType: KClass<T>`? and in `getConverterForType` as well.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -195,20 +201,270 @@ internal class DatadogFlagsClient(
...
+                flagKey = flagKey,
+                defaultValue = defaultValue,
+                errorCode = ErrorCode.FLAG_NOT_FOUND,
+                errorMessage = "Flag not found"
+            )
+        }
+
+        val (flag, context) = flagAndContext
+
+        val conversionResult = FlagValueConverter.convert<T>(
```
0xnm:
```suggestion
        val conversionResult = FlagValueConverter.convert(
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/utils/forge/PrecomputedFlagForgeryFactory.kt
```diff
@@ -22,34 +22,30 @@ internal class PrecomputedFlagForgeryFactory : ForgeryFactory<PrecomputedFlag> {
...
 
-        val reasonEnum = forge.anElementFrom(*FlagReason.values())
-        val reason = when (reasonEnum) {
-            FlagReason.DEFAULT -> FlagReason.DEFAULT.value
-            FlagReason.TARGETING_MATCH -> FlagReason.TARGETING_MATCH.value
-            FlagReason.RULE_MATCH -> FlagReason.RULE_MATCH.value
-            FlagReason.PREREQUISITE_FAILED -> FlagReason.PREREQUISITE_FAILED.value
-            FlagReason.ERROR -> FlagReason.ERROR.value
-        }
+        val reasonEnum = forge.anElementFrom(*ResolutionReason.values())
```
0xnm:
```suggestion
        val reasonEnum = forge.aValueFrom(ResolutionReason::class.java)
```

0xnm (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
aleksandr-gringauz: +1. It should be changed to `KClass<T>`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Gets the appropriate converter function for the given type.
+     * Handles special cases like JSONObject parsing with error logging.
+     *
+     * @param T The expected type
+     * @param targetType The target type class to get converter for
+     * @return A function that converts a string to type T, or null if conversion fails
+     */
+    @Suppress("UNCHECKED_CAST")
+    private fun <T : Any> getConverterForType(targetType: KClass<*>): (String) -> T? = when (targetType) {
```
aleksandr-gringauz: IMHO returning a lambda here instead of simply doing the conversion right aways is overcomplicating things for no reason.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
aleksandr-gringauz: +1

aleksandr-gringauz (approved):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
```diff
@@ -73,119 +76,122 @@ internal class DatadogFlagsClient(
...
+     * @param defaultValue The value to return if the flag cannot be retrieved or parsed.
+     * @return [ResolutionDetails] with either the parsed value and metadata, or an error.
+     */
+    override fun <T : Any> resolve(flagKey: String, defaultValue: T): ResolutionDetails<T> {
+        val resolution = readAndParseAssignment(flagKey, defaultValue)
 
-        if (convertedValue == null) {
-            return defaultValue
+        return when (resolution) {
+            is InternalResolution.Success -> {
```
jonathanmos, line 147: blocking: we're missing evaluation logging to RUM here - there is only exposure batches. Why not `resolveValue` here as in the other api methods?

jonathanmos (changes requested):
Looks good, but see comment about rum logging in `resolve`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: done

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/DatadogFlagsClient.kt
typotter: missed it on the merge. Added now in an extracted method.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: agreed. move the conversion inline

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
typotter: changed, thanks!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,91 @@
...
+     * @param T The expected type of the flag value
+     * @param variationValue The raw string value from the flag
+     * @param variationType The type specified in the flag metadata
+     * @param targetType The target type class to convert to
+     * @return Result containing the converted value on success,
+     *         or a Failure with TypeMismatchException if types are incompatible,
+     *         or a Failure with the parse exception if conversion failed
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun <T : Any> convert(variationValue: String, variationType: String, targetType: KClass<out T>): Result<T> {
```
0xnm:
```suggestion
    fun <T : Any> convert(variationValue: String, variationType: String, targetType: KClass<T>): Result<T> {
```

`out` is not needed here

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/internal/FlagValueConverter.kt
```diff
@@ -0,0 +1,91 @@
...
+        if (!isTypeCompatible(variationType, targetType)) {
+            val expectedType = getTypeName(targetType)
+            return Result.failure(
+                TypeMismatchException("Flag has type '$variationType' but $expectedType was requested")
+            )
+        }
+
+        return runCatching {
+            val result: T? = when (targetType) {
```
0xnm, line 53: tiny nit: it is better to move suppression closer to the site where it happens instead of having it for the whole method

```suggestion
        return runCatching {
            @Suppress("UNCHECKED_CAST")
            val result: T? = when (targetType) {
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/utils/forge/PrecomputedFlagForgeryFactory.kt
```diff
@@ -22,34 +22,30 @@ internal class PrecomputedFlagForgeryFactory : ForgeryFactory<PrecomputedFlag> {
             VariationType.BOOLEAN -> VariationType.BOOLEAN.value
```
jonathanmos: minor: we can replace this whole `when` block with:
```val variationType = variationTypeEnum.value```

File: features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/utils/forge/PrecomputedFlagForgeryFactory.kt [hunk 31]
```diff
@@ -22,34 +22,30 @@ internal class PrecomputedFlagForgeryFactory : ForgeryFactory<PrecomputedFlag> {
...
-            FlagReason.ERROR -> FlagReason.ERROR.value
-        }
+        val reasonEnum = forge.aValueFrom(ResolutionReason::class.java)
+        val reason = reasonEnum.name
 
         return PrecomputedFlag(
             variationType = variationType,
             variationValue = variationValue,
-            doLog = forge.aBool(),
+            doLog = true, // Default to true - tests can override if needed
```
jonathanmos: minor: I think we should consider forging a boolean value and overriding in tests as necessary, otherwise this forgery factory is misleading because it isn't actually forging all cases and this could be missed later on.

aleksandr-gringauz (approved):

PR #2937, opened 2025-10-13
RUM-11784: Update RUM Schema to include app launch vitals

aleksandr-gringauz:
### What does this PR do?

This PR pulls the most recent RUM events schema and updates the code to use it.

The goal is to use the new application launch vital introduced [here](https://github.com/DataDog/rum-events-format/pull/302). But it will be done in future PRs. I plan to merge this PR to develop.

The current PR:
1. Fixes the build errors and adapts code related to feature operations to the schema changes.
2. Fixes the json parser generator so that the following schema works correctly:
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "RequiredForOtherAllOf",
  "type": "object",
  "allOf": [
    {
      "type": "object",
      "properties": {
        "key_1": { "type": "string" }
      }
    },
    {
      "type": "object",
      "required": ["key_1", "key_2"],
      "properties": {
        "key_2": { "type": "string" }
      }
    }
  ]
}
```
`key_1` should be non-null in the resulting `data class`. See other comments in the PR for further explanation. This change is needed to properly support [this](https://github.com/DataDog/rum-events-format/pull/306) change. Otherwise `view` becomes optional for example in `ErrorEvent` which is not what we intended to do.

### Motivation

### Additional Notes

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/TypeDefinition.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/TypeDefinition.kt
@@ -101,6 +101,7 @@ sealed class TypeDefinition {
     data class Class(
         val name: String,
         val properties: List<TypeProperty>,
+        val required: Set<String>,
         override val description: String = "",
         val additionalProperties: TypeProperty? = null,
         val parentType: OneOfClass? = null
@@ -135,18 +136,22 @@ sealed class TypeDefinition {
                     additionalProperties.mergedWith(other.additionalProperties)
                 }
 
+            val mergedRequired = this.required + other.required
+
             return Class(
-                name,
-                mergedFields,
-                "$description\n${other.description}".trim(),
-                mergedAdditionalProperties
+                name = name,
+                properties = mergedFields,
+                required = mergedRequired,
+                description = "$description\n${other.description}".trim(),
+                additionalProperties = mergedAdditionalProperties
             )
         }
 
         override fun matches(other: TypeDefinition): Boolean {
             return (other is Class) &&
                 (other.properties == properties) &&
-                (other.additionalProperties == additionalProperties)
+                (other.additionalProperties == additionalProperties) &&
+                (other.required == required)
         }
 
         fun isConstantClass(): Boolean {
--- a/buildSrc/src/test/resources/input/external_description_complex_path.json
+++ b/buildSrc/src/test/resources/input/external_description_complex_path.json
@@ -9,9 +9,5 @@
     "contact": {
       "$ref": "subfolder/external_description_proxy.json#/definitions/contact"
     }
-  },
-  "required": [
-    "item",
-    "destination"
-  ]
+  }
 }
--- a/features/dd-sdk-android-rum/api/apiSurface
+++ b/features/dd-sdk-android-rum/api/apiSurface
@@ -616,6 +616,34 @@ data class com.datadog.android.rum.model.ActionEvent
     fun toJson(): com.google.gson.JsonElement
     companion object 
       fun fromJson(kotlin.String): Type
+data class com.datadog.android.rum.model.AppLaunchProperties
+  constructor(kotlin.String, kotlin.String? = null, kotlin.String? = null, AppLaunchMetric, kotlin.Number, StartupType? = null, kotlin.Boolean? = null, kotlin.Boolean? = null)
+  val type: kotlin.String
+  fun toJson(): com.google.gson.JsonElement
+  companion object 
+    fun fromJson(kotlin.String): AppLaunchProperties
+    fun fromJsonObject(com.google.gson.JsonObject): AppLaunchProperties
+  enum AppLaunchMetric
+    constructor(kotlin.String)
+    - TTID
+    - TTFD
+    fun toJson(): com.google.gson.JsonElement
+    companion object 
+      fun fromJson(kotlin.String): AppLaunchMetric
+  enum StartupType
+    constructor(kotlin.String)
+    - COLD_START
+    - WARM_START
+    fun toJson(): com.google.gson.JsonElement
+    companion object 
+      fun fromJson(kotlin.String): StartupType
+data class com.datadog.android.rum.model.DurationProperties
+  constructor(kotlin.String, kotlin.String? = null, kotlin.String? = null, kotlin.Number)
+  val type: kotlin.String
+  fun toJson(): com.google.gson.JsonElement
+  companion object 
+    fun fromJson(kotlin.String): DurationProperties
+    fun fromJsonObject(com.google.gson.JsonObject): DurationProperties
 data class com.datadog.android.rum.model.ErrorEvent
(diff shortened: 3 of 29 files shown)


buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/TypeDefinition.kt
```diff
@@ -101,6 +101,7 @@ sealed class TypeDefinition {
     data class Class(
         val name: String,
         val properties: List<TypeProperty>,
+        val required: Set<String>,
```
aleksandr-gringauz, line 104: The idea is the following. 

We store the information about the required fields here instead of storing `val optional: Boolean` in `TypeDefinition`. Because if there is no `TypeDefinition` corresponding to the key in required block, we lose the information that this field is required. And if this key actually occurs in another item of `allOf` we don't treat it as required after merging (this is what I want to fix).

buildSrc/src/test/resources/input/external_description_complex_path.json
```diff
@@ -9,9 +9,5 @@
     "contact": {
       "$ref": "subfolder/external_description_proxy.json#/definitions/contact"
     }
-  },
-  "required": [
```
aleksandr-gringauz, line 13: These `item` and `destination` don't actually exists in this schema. Before my fix the test passed, because the generator only looked at this required block while iterating through the properties (to determine the `val optional: Boolean` field).

features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
```diff
@@ -440,6 +440,19 @@
       "type": "object",
       "description": "User provided context",
       "additionalProperties": true
+    },
+    "stream": {
```
aleksandr-gringauz: This isn't related to vitals, some other new stuff.

0xnm: let's maybe just remove this if not needed for now?

features/dd-sdk-android-rum/generate_rum_models.gradle.kts
```diff
@@ -21,9 +21,12 @@ tasks.register(
         "_perf-metric-schema.json",
         "_profiling-internal-context-schema.json",
         "_rect-schema.json",
+        "_stream-schema.json",
```
0xnm: let's remove this file, we can add it when needed

features/dd-sdk-android-rum/generate_rum_models.gradle.kts
```diff
@@ -21,9 +21,12 @@ tasks.register(
...
         "_view-container-schema.json",
         "_view-accessibility-schema.json",
-        "_view-performance-schema.json"
+        "_view-performance-schema.json",
+        "_vital-common-schema.json",
+        "transition-schema.json"
```
0xnm: This one can be removed as well, it is not needed for now. We can focus only on Vital changes in this PR.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -401,13 +401,12 @@ internal open class RumViewScope(
...
             service = datadogContext.service,
             ddtags = buildDDTagsString(datadogContext),
-            vital = VitalEvent.VitalEventVital(
+            vital = VitalEvent.Vital.FeatureOperationProperties(
                 id = UUID.randomUUID().toString(),
                 name = name,
                 operationKey = operationKey,
                 stepType = stepType,
-                failureReason = failureReason,
-                type = VitalEvent.VitalEventVitalType.OPERATION_STEP
```
0xnm, line 410: we don't need to specify `type` explicitly?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -8727,6 +8723,17 @@ internal class RumViewScopeTest {
                 .hasVersion(fakeDatadogContext.version)
                 .hasServiceName(fakeDatadogContext.service)
                 .hasDDTags(buildDDTagsString(fakeDatadogContext))
+
+            val featureOperationsProps = lastValue.vital
+
+            assertThat(featureOperationsProps).isInstanceOf(VitalEvent.Vital.FeatureOperationProperties::class.java)
+            featureOperationsProps as VitalEvent.Vital.FeatureOperationProperties
```
0xnm:
```suggestion
            check(featureOperationsProps is VitalEvent.Vital.FeatureOperationProperties)
```
and then smart cast will do the job

same can be done for the similar lines below

features/dd-sdk-android-rum/src/main/json/rum/view-schema.json
```diff
@@ -11,6 +11,9 @@
     {
       "$ref": "_view-container-schema.json"
     },
+    {
```
0xnm: can also be removed for now

features/dd-sdk-android-rum/generate_rum_models.gradle.kts
aleksandr-gringauz: Done. Removed everything (new schema files, new fields in existing files) related to `stream` and `transition`.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
aleksandr-gringauz: done

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
aleksandr-gringauz: No. `type` is defined like [this](https://github.com/DataDog/rum-events-format/blob/6429a4b4b9999c3f0cec22d5c1f36f0e47af412a/schemas/rum/vital-feature-operation-schema.json#L17) now as a constant.

In kotlin it will look like `public val type: String = "operation_step"` inside `public data class FeatureOperationProperties`.

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -2106,12 +2178,18 @@ data class com.datadog.android.rum.model.VitalEvent
     companion object 
       fun fromJson(kotlin.String): Configuration
       fun fromJsonObject(com.google.gson.JsonObject): Configuration
-  data class DdVital
+  data class Vital1
```
0xnm, line 2181: not the best naming, but I guess we can live with that

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -2058,7 +2110,7 @@ data class com.datadog.android.rum.model.VitalEvent
       fun fromJson(kotlin.String): Device
       fun fromJsonObject(com.google.gson.JsonObject): Device
   data class Dd
-    constructor(DdSession? = null, Configuration? = null, kotlin.String? = null, kotlin.String? = null, DdVital? = null)
+    constructor(DdSession? = null, Configuration? = null, kotlin.String? = null, kotlin.String? = null, Vital1? = null, Profiling? = null)
```
0xnm, line 2113: just a sanity check: property name is still `vital`?

aleksandr-gringauz: yes https://github.com/DataDog/rum-events-format/blob/6429a4b4b9999c3f0cec22d5c1f36f0e47af412a/schemas/rum/vital-schema.json#L41

PR #2940, opened 2025-10-14
Handle OTEL reserved attributes when using AttributeKeys (Fixes #2935)

### LINKED ISSUE #2935: OTEL reserved attributes are not applied if using setAttribute(AttributeKey<T> key...)
### Describe the bug

When using OpenTelemetry implementation, reserved custom attributes (like `"operation.name"`) are not applied when passing an `AttributeKey` instead of a `String` key.

For example, the following code works:
```kotlin
spanBuilder.setAttribute("operation.name", "op_name")
```

The following code instead doesn't:
```kotlin
spanBuilder.setAttribute(stringKey("operation.name"), "op_name")
```

As a result of this, it's impossible to use `spanBuilder.setAllAttributes(...)` since internally it iterates over a list of `AttributeKey<T>`, ignoring any reserved key passed in the Map.

### Reproduction steps

```kotlin
val spanBuilder: SpanBuilder = otelTracer.spanBuilder("name")
// This handles correctly the reserved attribute "operation.name". On Datadog the operation name will be "op_name"
spanBuilder.setAttribute("operation.name", "op_name")

// This ignores the reserved attributes. On Datadog the operation name with be the default "internal"
spanBuilder.setAttribute(stringKey("operation.name"), "op_name")

// This also ignores the reserved attributes.
val attributes = Attributes.builder().put("operation.name", "op_name").build()
spanBuilder.setAllAttributes(attributes)
```

### Logcat logs

_No response_

### Expected behavior

Reserved attributes should be handled in all cases.

### Affected SDK versions

3.1.0

none

Yes

### Kotlin / Java version

_No response_

### Gradle / AGP version

_No response_

_No response_

### Device Information

_No response_

_No response_

paolorotolo:
### What does this PR do?
This PR extracts the code to handle OTEL reserved attributes when setting a new key with `setAttribute(@NonNull String key, @Nullable String value)` and applies it also to the `setAttribute(AttributeKey<T> key, @Nullable T value)` method.

Test coverage has also been added to `OtelSpanBuilder`.

### Motivation
Setting OTEL reserved attributes with:
```kotlin
spanBuilder.setAttribute(stringKey("operation.name"), "op_name")
```

or 

```kotlin
val attributes = Attributes.builder().put("operation.name", "op_name").build()
spanBuilder.setAllAttributes(attributes)
```

had no result using the latest `dd-sdk-android` release.

See #2935 for more info.

### DIFF
--- a/features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/opentelemetry/trace/OtelSpanBuilderTest.kt
+++ b/features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/opentelemetry/trace/OtelSpanBuilderTest.kt
@@ -22,7 +22,6 @@ import org.junit.jupiter.api.extension.Extensions
 import org.mockito.Mock
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
-import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.verifyNoInteractions
 import org.mockito.kotlin.whenever
@@ -86,7 +85,7 @@ internal class OtelSpanBuilderTest {
         verifyNoInteractions(mockDelegateBuilder)
         testedBuilder.startSpan()
 
-        verify(mockAgentSpan, times(1)).operationName = "op_name"
+        verify(mockAgentSpan).operationName = "op_name"
         verify(mockAgentSpan).setMetric(DatadogTracingConstants.Tags.KEY_ANALYTICS_SAMPLE_RATE, 1)
     }
 
@@ -102,7 +101,7 @@ internal class OtelSpanBuilderTest {
         verifyNoInteractions(mockDelegateBuilder)
         testedBuilder.startSpan()
 
-        verify(mockAgentSpan, times(1)).operationName = "op_name"
+        verify(mockAgentSpan).operationName = "op_name"
         verify(mockAgentSpan).setMetric(DatadogTracingConstants.Tags.KEY_ANALYTICS_SAMPLE_RATE, 0)
     }
 


features/dd-sdk-android-trace-otel/src/main/java/com/datadog/opentelemetry/trace/OtelSpanBuilder.java
```diff
@@ -156,6 +150,11 @@ public <T> SpanBuilder setAttribute(AttributeKey<T> key, @Nullable T value) {
                     }
                 }
                 break;
+            case STRING:
+                if (!handleReservedAttribute(key.getKey(), (String) value)) {
```
satween: I think the correct way will be: 
```
case STRING:
  return setAttribute(key.getKey(), (String) value);
```

isn't it?

paolorotolo: That's true, they end up doing the same thing once the value is a string.

Should I keep the extracted `handleReservedAttribute` method in `setAttribute(@NonNull String key, @Nullable String value)` for clarity or revert that?

satween: Yep, I think it will be good way to dont repeat yourself and keep whole important logic inside one place. This also will be another guarantee (along with tests) that every method call eventually will use same logic under the hood.

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/opentelemetry/trace/OtelSpanBuilderTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        testedBuilder.setAttribute(
+            AttributeKey.stringKey(OtelConventions.OPERATION_NAME_SPECIFIC_ATTRIBUTE), "op_name")
+        testedBuilder.setAttribute(
+            AttributeKey.stringKey(OtelConventions.ANALYTICS_EVENT_SPECIFIC_ATTRIBUTES), "false")
+
+        // Then
+        verifyNoInteractions(mockDelegateBuilder)
+        testedBuilder.startSpan()
+
+        verify(mockAgentSpan, times(1)).operationName = "op_name"
```
satween: I think you can avoid adding `times(1)` as it is a default value for this parameter in `verify` method.

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/opentelemetry/trace/OtelSpanBuilderTest.kt
```diff
@@ -0,0 +1,110 @@
...
+    fun `M add attribute W setAttribute with reserved string key`() {
+        // When
+        testedBuilder.setAttribute(OtelConventions.OPERATION_NAME_SPECIFIC_ATTRIBUTE, "op_name")
+        testedBuilder.setAttribute(OtelConventions.ANALYTICS_EVENT_SPECIFIC_ATTRIBUTES, "true")
+
+        // Then
+        verifyNoInteractions(mockDelegateBuilder)
+        testedBuilder.startSpan()
+
+        verify(mockAgentSpan, times(1)).operationName = "op_name"
```
satween: I think you can avoid adding times(1) as it is a default value for this parameter in verify method.

features/dd-sdk-android-trace-otel/src/test/kotlin/com/datadog/opentelemetry/trace/OtelSpanBuilderTest.kt
paolorotolo: Done, thanks!

features/dd-sdk-android-trace-otel/src/main/java/com/datadog/opentelemetry/trace/OtelSpanBuilder.java
paolorotolo: makes sense, thanks. I'm calling the same method now.

PR #2944, opened 2025-10-15
RUM-11784: Send TTID vital

aleksandr-gringauz:
### What does this PR do?

This PR implements sending of TTID value using Rum `VitalEvent` with `ApplicationLaunchProperties`. Merging into a feature branch for now.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
@@ -16,7 +16,6 @@ import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.rum.RumSessionListener
 import com.datadog.android.rum.RumSessionType
-import com.datadog.android.rum.internal.FeaturesContextResolver
 import com.datadog.android.rum.internal.domain.InfoProvider
 import com.datadog.android.rum.internal.domain.RumContext
 import com.datadog.android.rum.internal.domain.Time
@@ -62,8 +61,7 @@ internal class RumSessionScope(
     private val sessionMaxDurationNanos: Long = DEFAULT_SESSION_MAX_DURATION_NS,
     rumSessionTypeOverride: RumSessionType?,
     private val rumAppStartupTelemetryReporter: RumAppStartupTelemetryReporter,
-    private val rumVitalEventHelper: RumVitalEventHelper,
-    private val featuresContextResolver: FeaturesContextResolver
+    private val rumVitalEventHelper: RumVitalEventHelper
 ) : RumScope {
 
     internal var sessionId = RumContext.NULL_UUID
@@ -315,17 +313,13 @@ internal class RumSessionScope(
     ): VitalEvent {
         val rumContext = getRumContext()
 
-        val hasReplay = childScope?.activeView?.viewId?.let {
-            featuresContextResolver.resolveViewHasReplay(datadogContext, it)
-        }
-
         return rumVitalEventHelper.newVitalEvent(
             timestampMs = event.info.scenario.initialTime.timestamp + sdkCore.time.serverTimeOffsetMs,
             datadogContext = datadogContext,
             eventAttributes = emptyMap(),
             customAttributes = getCustomAttributes(),
             view = null,
-            hasReplay = hasReplay,
+            hasReplay = null,
             rumContext = rumContext,
             vital = VitalEvent.Vital.AppLaunchProperties(
                 id = UUID.randomUUID().toString(),
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
@@ -20,9 +20,8 @@ import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.core.internal.net.FirstPartyHostHeaderTypeResolver
 import com.datadog.android.rum.RumSessionListener
 import com.datadog.android.rum.RumSessionType
-import com.datadog.android.rum.assertj.VitalAppLaunchPropertiesAssert
+import com.datadog.android.rum.assertj.VitalAppLaunchPropertiesAssert.Companion.assertThat
 import com.datadog.android.rum.assertj.VitalEventAssert
-import com.datadog.android.rum.internal.FeaturesContextResolver
 import com.datadog.android.rum.internal.domain.InfoProvider
 import com.datadog.android.rum.internal.domain.RumContext
 import com.datadog.android.rum.internal.domain.accessibility.AccessibilitySnapshotManager
@@ -184,9 +183,6 @@ internal class RumSessionScopeTest {
 
     @Forgery
     lateinit var fakeDisplayInfo: DisplayInfo
-
-    @BoolForgery
-    var fakeHasReplay: Boolean = false
 
     private var fakeVitalSource: VitalEvent.VitalEventSource? = null
 
@@ -1577,17 +1573,6 @@ internal class RumSessionScopeTest {
             syntheticsResultId = null
         )
 
-        val mockView = mock<RumViewScope>()
-        val fakeViewId = forge.aString()
-        whenever(mockView.viewId) doReturn fakeViewId
-        whenever(mockChildScope.activeView) doReturn mockView
-
-        fakeDatadogContext = fakeDatadogContext.copy(
-            featuresContext = fakeDatadogContext.featuresContext.toMutableMap().apply {
-                put(Feature.SESSION_REPLAY_FEATURE_NAME, mapOf(fakeViewId to mapOf("has_replay" to fakeHasReplay)))
-            }
-        )
-
         val info = RumTTIDInfo(
             scenario = scenario,
             durationNs = forge.aLong(min = 0, max = 10000)
@@ -1598,7 +1583,6 @@ internal class RumSessionScopeTest {
         )
 
         // When
-
         val result = testedScope.handleEvent(
             event = event,
             datadogContext = fakeDatadogContext,
@@ -1620,7 +1604,7 @@ internal class RumSessionScopeTest {
                 hasNoSyntheticsTest()
                 hasSessionId(context.sessionId)
                 hasSessionType(fakeRumSessionType?.toVital() ?: VitalEvent.VitalEventSessionType.USER)
-                hasSessionReplay(fakeHasReplay)
+                hasNoSessionReplay()
                 hasNullView()
                 hasSource(fakeVitalSource)
                 hasAccountInfo(fakeDatadogContext.accountInfo)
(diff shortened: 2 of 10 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -281,6 +288,58 @@ internal class RumSessionScope(
...
+            createTTIDVitalEvent(
+                event = event,
+                datadogContext = datadogContext
+            )
+        }.submit()
+
+        rumAppStartupTelemetryReporter.reportTTID(
+            info = event.info,
+            indexInSession = appStartIndex
+        )
```
0xnm, line 305: should this go into `onSuccess` of the write operation above or it is fine to report this if event wasn't written?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -281,6 +288,58 @@ internal class RumSessionScope(
...
+        appStartIndex++
+    }
+
+    private fun createTTIDVitalEvent(
+        event: RumRawEvent.AppStartTTIDEvent,
+        datadogContext: DatadogContext
+    ): VitalEvent {
+        val rumContext = getRumContext()
+
+        val hasReplay = childScope?.activeView?.viewId?.let {
```
0xnm: `viewId` can be taken simply from `RumContext`:

```suggestion
        val hasReplay = rumContext.viewId?.let {
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -281,6 +288,58 @@ internal class RumSessionScope(
...
+            featuresContextResolver.resolveViewHasReplay(datadogContext, it)
+        }
+
+        return rumVitalEventHelper.newVitalEvent(
+            timestampMs = event.info.scenario.initialTime.timestamp + sdkCore.time.serverTimeOffsetMs,
+            datadogContext = datadogContext,
+            eventAttributes = emptyMap(),
+            customAttributes = getCustomAttributes(),
+            view = null,
+            hasReplay = hasReplay,
```
0xnm: I'm not sure if we need to set this flag at all here. Yes, there is such property, but it is only app start, so no real view yet, hence nothing to replay. We have `view = null` above anyway. WDYT?

If we don't set this property, we even simplify the code, because there is not need to pass `FeaturesContextResolver` around.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -163,11 +174,27 @@ internal class RumSessionScopeTest {
...
+    var fakeHasReplay: Boolean = false
+
+    private var fakeVitalSource: VitalEvent.VitalEventSource? = null
+
     @BeforeEach
     fun `set up`(forge: Forge) {
         fakeInitialViewEvent = forge.startViewEvent()
 
-        whenever(mockParentScope.getRumContext()) doReturn fakeParentContext
+        whenever(mockParentScope.getRumContext()).doAnswer { fakeParentContext }
```
0xnm, line 193: does this change anything?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -1506,6 +1565,99 @@ internal class RumSessionScopeTest {
...
+            durationNs = forge.aLong(min = 0, max = 10000)
+        )
+
+        val event = RumRawEvent.AppStartTTIDEvent(
+            info = info
+        )
+
+        // When
+
+        val result = testedScope.handleEvent(
```
0xnm, line 1586:
```suggestion
        // When
        val result = testedScope.handleEvent(
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -1506,6 +1565,99 @@ internal class RumSessionScopeTest {
...
+                hasConnectivityInfo(fakeDatadogContext.networkInfo)
+                hasVersion(fakeDatadogContext.version)
+                hasServiceName(fakeDatadogContext.service)
+                hasDDTags(buildDDTagsString(fakeDatadogContext))
+            }
+
+            val vital = lastValue.vital
+            check(vital is VitalEvent.Vital.AppLaunchProperties)
+
+            VitalAppLaunchPropertiesAssert.assertThat(vital).apply {
```
0xnm: minor: should work as well if you move `VitalAppLaunchPropertiesAssert.assertThat` to import and then it will be just `assertThat` here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
aleksandr-gringauz: Good question, I'm not sure either.

Let's consider the situation when the customer uses [ActivityViewTrackingStrategy](https://github.com/DataDog/dd-sdk-android/blob/641c42dccd480ac8d714e9c8627a3dd2d4c6e16a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityViewTrackingStrategy.kt#L33). 

1. The view starts in [onResume](https://github.com/DataDog/dd-sdk-android/blob/641c42dccd480ac8d714e9c8627a3dd2d4c6e16a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityViewTrackingStrategy.kt#L65).
2. The first draw (and TTID Vital event as a result) will happen slightly (less than a second) after this, so technically the view already exists.
3. The "has_replay" flag for a view that we check [here](https://github.com/DataDog/dd-sdk-android/blob/641c42dccd480ac8d714e9c8627a3dd2d4c6e16a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/FeaturesContextResolver.kt#L19) is set [here](https://github.com/DataDog/dd-sdk-android/blob/641c42dccd480ac8d714e9c8627a3dd2d4c6e16a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/SessionReplayRecordCallback.kt#L26). This happens when a session replay event is recoded [here](https://github.com/DataDog/dd-sdk-android/blob/641c42dccd480ac8d714e9c8627a3dd2d4c6e16a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/storage/SessionReplayRecordWriter.kt#L33).
4. In practice I've never seen `hasReplay = true` in my implementation. As far as I understood the session replay event isn't written fast enough. However if you delay the TTID for 5 seconds `hasReplay` will be true. We have a race here.
5. To wrap up. In practice this `hasReplay` in my case is kind of useless and will in most cases be `false`. I can just pass `hasReplay = null`, I don't think it will be a problem.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
aleksandr-gringauz: I don't think so. This is `RumSessionScope` and [getRumContext](https://github.com/DataDog/dd-sdk-android/blob/641c42dccd480ac8d714e9c8627a3dd2d4c6e16a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt#L182) doesn't fill `viewId` field. It is filled only [here](https://github.com/DataDog/dd-sdk-android/blob/641c42dccd480ac8d714e9c8627a3dd2d4c6e16a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L416) in `RumViewScope`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
aleksandr-gringauz: I think it is fine to report this telemetry even if TTID vital isn't sent. Nothing will break, we'll just have a bit more data in telemetry.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
aleksandr-gringauz: I change this `fakeParentContext` in one of my tests [here](https://github.com/DataDog/dd-sdk-android/blob/928d6c2b5d60f70e2c35031840e72fbfb844c6a5/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt#L1575). When I used `doReturn` `mockParentScope.getRumContext()` returned the old `fakeParentContext`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
0xnm: I think having `hasReplay = null` here is fine. This property is just a sticky flag which is reduced over view updates and pulled to the session level. So basically once it is set, it means that the given session has Session Replay. It means that if _any_ other view (presumably the real one) update delivers it, this will be fine.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
0xnm: Good point, indeed it is called on `RumSessionScope`. But you don't need it anyway if we are not going to set `has_replay` here.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
aleksandr-gringauz: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
aleksandr-gringauz: changed to `hasReplay = null` and simplified the code.

0xnm (approved):

PR #2950, opened 2025-10-17
RUM-11990: Defer requestProfiling to next app launch

ambushwork:
### What does this PR do?

This PR enables the profiling request as early as possible during the app launch:
* Add `DdProfilingContentProvider` to start the profiling request.
* Add `ProfilingStorage` to read/save/clean the profiling flag to defer the profiling request to the next app launch.
 
### Motivation

RUM-11990

### Additional Notes

Profiling Sampling rate is not covered by this PR.


features/dd-sdk-android-profiling/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,20 @@
...
+<manifest xmlns:tools="http://schemas.android.com/tools"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+
+        <provider
+            android:name="com.datadog.android.profiling.DdProfilingContentProvider"
+            android:authorities="${applicationId}.provider.datadog.profiling"
+            android:exported="false"
+            tools:targetApi="35" />
```
aleksandr-gringauz: This `tools:targetApi="35"` just suppresses the lint error ([link](https://developer.android.com/studio/write/tool-attributes#toolstargetapi)).

IMHO it is safer to do:

```
override fun onCreate(): Boolean {
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
        context?.takeIf { isProfilingEnabled(it) }?.let {
            Profiling.start(it)
        }    
    }
    return true
}
```

And not mark `DdProfilingContentProvider` with `@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)`.

Because the way it is now, `DdProfilingContentProvider.onCreate` is still called on API < 35 and it is possible to add code there that crashes on those API levels without the compiler telling you you are wrong.

I understand that now `isProfilingEnabled` will not return `true` on API < 35, because you can't call `Profiling.enable()` on API < 35, but I still think you should be extra safe here and rely on compiler more if it is possible.

ambushwork: Good point, updated as you suggest

aleksandr-gringauz (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/SharedPreferencesStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+    override fun remove(key: String) {
+        prefs.edit().remove(key).apply()
+    }
+
+    override fun clear() {
+        prefs.edit().clear().apply()
+    }
+
+    companion object {
+        internal const val DATADOG_PREFERENCES_FILE_NAME = "dd_prefs"
```
aleksandr-gringauz, line 54: Just a heads up.

There is [this](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5586452507/RFC+Android+SDK+Profiling+Start+Stop+Strategy?focusedCommentId=5614569743) comment in the RFC about the case when we have multiple instances of the SDK in the app.

Currently they will use the same shared preferences which is probably not good. We might need to parametrize this with `sdkInstanceName` string.

Not a blocking comment for this PR. Maybe you'll come up with better ideas.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/PreferencesStorage.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.datadog.tools.annotation.NoOpImplementation
+
+@NoOpImplementation
+internal interface PreferencesStorage {
+
+    fun putInt(key: String, value: Int)
+    fun getInt(key: String, defaultValue: Int = 0): Int
+
+    fun getString(key: String): String?
+    fun putString(key: String, value: String)
```
0xnm, line 35: probably these are unused for now and can be removed? we can add them later

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/SharedPreferencesStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+package com.datadog.android.internal.data
+
+import android.content.Context
+import android.content.SharedPreferences
+
+/**
+ * An implementation of [PreferencesStorage] that uses Android's [SharedPreferences] as storage
+ * backend.
+ */
+class SharedPreferencesStorage(context: Context) : PreferencesStorage {
```
0xnm: `public` class is exposing `internal` interface, there should be warning about this. We need to make interface public as well then.

0xnm: minor: let's make it clear what we expect application context here

```suggestion
class SharedPreferencesStorage(appContext: Context) : PreferencesStorage {
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/SharedPreferencesStorage.kt
0xnm: +1 on this: what we are missing is multiple SDK cores support.

To start recording we need that _any_ of the cores requested it (at least with current configuration where sampling rate is hardcoded), so this is probably fine for the flag to be shared, but we need to:

* make sure that after recording is done it is sent to the proper core which requested it.
* make sure that we don't run several profiles in parallel.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -6,19 +6,27 @@
...
+import com.datadog.android.profiling.internal.perfetto.PerfettoProfiler
+import com.datadog.android.profiling.internal.removeProfilingFlag
+import java.util.concurrent.Executors
 
 /**
  * An entry point to Datadog Profiling feature.
  */
 object Profiling {
 
+    private var profiler: Profiler = NoOpProfiler()
```
0xnm, line 28: this becomes a little bit fragile and probably won't scale for the future development, where we need to start/stop profiling manually and it will be bound to the particular SDK core.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -38,8 +46,19 @@ object Profiling {
         val profilingFeature = ProfilingFeature(
             sdkCore = featureSdkCore,
             configuration = configuration,
-            profilerProvider = PerfettoProfilerProvider()
+            profiler = profiler
```
0xnm, line 49: by doing this it means we can have a possibility to receive profile result for the SDK core instance which didn't request it?

Also it seems by sharing `Profiler` we dispatch result only once, that doesn't work for the multiple SDK cores.

0xnm (changes requested):
I think what we are missing is multiple SDK cores support, we need to refine this flow.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/PreferencesStorage.kt
ambushwork: We don't use it in this PR, but very high change to use it in the future, let's keep it for the integrity of the implementation?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/SharedPreferencesStorage.kt
ambushwork: I agree, I think the multiple SDK core issue is not mainly on SharedPreference flag, but on the `ProfilingFeature` implementation about writing and uploading events as you said. For the dogfooding it's not an issue, I create a [ticket](https://datadoghq.atlassian.net/browse/RUM-12330?atlOrigin=eyJpIjoiOTg5ZjMwMzM4M2E0NDMxZjg5NTYzYmE2YWFkYzAzYWMiLCJwIjoiaiJ9) to handle it in as a separate task

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/SharedPreferencesStorage.kt
ambushwork: updated

ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
ambushwork: will handle in multi-core task mentioned above.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
ambushwork: will handle in multi-core task mentioned above.

aleksandr-gringauz (dismissed):

dd-sdk-android-internal/api/apiSurface
```diff
@@ -22,6 +22,26 @@ class com.datadog.android.internal.collections.EvictingQueue<T> : java.util.Queu
...
+  constructor(android.content.Context)
+  override fun putInt(String, Int)
+  override fun getInt(String, Int): Int
+  override fun getString(String): String?
+  override fun putString(String, String)
+  override fun putBoolean(String, Boolean)
+  override fun getBoolean(String, Boolean): Boolean
+  override fun remove(String)
+  override fun clear()
+  companion object
```
0xnm: we need to make it `internal` or `private`, because it doesn't expose any public members anyway

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/SharedPreferencesStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+package com.datadog.android.internal.data
+
+import android.content.Context
+import android.content.SharedPreferences
+
+/**
+ * An implementation of [PreferencesStorage] that uses Android's [SharedPreferences] as storage
+ * backend.
+ */
+class SharedPreferencesStorage(appContext: Context) : PreferencesStorage {
```
0xnm, line 16: one downside is that we will trigger Strict Mode violation, because we need to read from disk on I/O thread, but I guess we don't have any other choice.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/PreferencesStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Get an integer value from the storage.
+     */
+    fun getInt(key: String, defaultValue: Int = 0): Int
+
+    /**
+     * Put a string value in the storage.
+     */
+    fun getString(key: String): String?
```
0xnm: we are not supplying any default value here as we do with `getInt` and `getBoolean` methods?

features/dd-sdk-android-profiling/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,18 @@
...
+  ~ Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+  ~ This product includes software developed at Datadog (https://www.datadoghq.com/).
+  ~ Copyright 2016-Present Datadog, Inc.
+  -->
+
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+
+        <provider
```
0xnm, line 11: I'm wondering if we also should leverage [initOrder](https://developer.android.com/guide/topics/manifest/provider-element#init) attribute to make sure it is called as early as possible.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStorage.kt
```diff
@@ -0,0 +1,24 @@
...
+
+package com.datadog.android.profiling.internal
+
+import android.content.Context
+import com.datadog.android.internal.data.SharedPreferencesStorage
+
+private const val KEY_PROFILING_ENABLED = "dd_profiling_enabled"
+
+internal fun setProfilingFlag(appContext: Context) {
+    SharedPreferencesStorage(appContext).putBoolean(KEY_PROFILING_ENABLED, true)
```
0xnm: minor: each instance creation will involve doing `appContext.getSharedPreferences` call, which is probably doing I/O, so ideally we could create an instance only once if it is a concern

0xnm (dismissed):

features/dd-sdk-android-profiling/src/main/AndroidManifest.xml
ambushwork: Good point, just wondering what value should we give, max value of Int or just a random large number like 1000

0xnm: I'm not sure, but probably this field is rarely used and apps have very few providers, so something like 1000 or 9999 will make it first in almost all cases.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/PreferencesStorage.kt
ambushwork: fixed

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStorage.kt
ambushwork: Update `ProfilingStage` to singleton with unit test

aleksandr-gringauz (approved):

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingStorageTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    fun `set up`() {
+        whenever(mockContext.getSharedPreferences(any(), any())) doReturn mockPrefs
+        whenever(mockPrefs.edit()) doReturn mockEditor
+        whenever(mockEditor.remove(any())) doReturn mockEditor
+        whenever(mockEditor.putBoolean(any(), any())) doReturn mockEditor
+
+        // Reset the singleton
+        val storageField = ProfilingStorage::class.java.getDeclaredField("sharedPreferencesStorage")
+        storageField.isAccessible = true
+        storageField.set(ProfilingStorage, null)
```
0xnm, line 53: minor: probably it is better to do it in the `AfterEach` (or maybe even in both `BeforeEach` and `AfterEach`), because we will leave the `object` with state set and this object may be used in another test class.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingStorageTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    fun `M return true W isProfilingEnabled() {flag is set}`() {
+        // Given
+        whenever(mockPrefs.getBoolean("dd_profiling_enabled", false)) doReturn true
+
+        // When
+        val isEnabled = ProfilingStorage.isProfilingEnabled(mockContext)
+
+        // Then
+        Assertions.assertThat(isEnabled).isTrue()
```
0xnm, line 75: nit: can be a static import

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingStorageTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        repeat(10) {
+            Thread {
+                ProfilingStorage.setProfilingFlag(mockContext)
+                latch.countDown()
+            }.start()
+        }
+        latch.await(5, TimeUnit.SECONDS)
+
+        // Then
+        verify(mockContext, times(1)).getSharedPreferences(
```
0xnm, line 115:
```suggestion
        verify(mockContext).getSharedPreferences(
```

0xnm (approved):

PR #2963, opened 2025-10-22
feat: Graceful mode and error logging policies

typotter:
### What does this PR do?

Implements graceful mode error handling for the Flags SDK, providing three-tier policy-based error handling (Graceful/Error/Strict) that adapts behavior based on build type and configuration.

**Key changes:**
- Added `gracefulModeEnabled` configuration parameter (defaults to `true`)
- Implemented debug build detection using `ApplicationInfo.FLAG_DEBUGGABLE`
- Added policy-aware error logging in `FlagsFeature.logErrorWithPolicy()`
- Refactored `NoOpFlagsClient` to use policy-based logging functions
- Updated client creation and retrieval to crash in strict mode for misuse scenarios

**Policy behavior:**
- **Graceful Policy** (Release builds): Conditional logging through SDK logger
- **Error Policy** (Debug + graceful enabled): Always log to `System.err`, non-crashing
- **Strict Policy** (Debug + graceful disabled): Crash immediately on misuse with
`IllegalStateException`

### Motivation

The Flags SDK needs configurable error handling to balance developer experience across
different environments:

- **Development/QA**: Fail fast with crashes to catch configuration errors early
- **Staging/Dogfooding**: Loud warnings without crashes for visibility without disruption
- **Production**: Silent fallbacks for maximum stability

This aligns with OpenFeature error handling patterns and provides flexibility for teams
with different development workflows (small teams vs. large modular apps, monoliths vs.
DI-based architectures).

**Related:** FFL-1187

### Additional Notes

**Implementation highlights:**

1. **Debug Detection:** Uses `ApplicationInfo.FLAG_DEBUGGABLE` (same pattern as
`DatadogCore.isAppDebuggable()`) with performance caching via `@Volatile` field primed
during `onInitialize()`

2. **Thread Safety:**
   - `@Volatile` fields for `appContext`, `debugBuildCache`, and `initialized` flag
   - Double-check locking pattern in `getOrRegisterNewClient()`
   - Proper lifecycle management with cleanup in `onStop()`

3. **Crash Points (Strict Mode Only):**
   - Duplicate client creation: `FlagsFeature.getOrRegisterNewClient()`
   - Client not found: `FlagsClient.get()`
   - Uses Kotlin's `error()` function to throw `IllegalStateException`

4. **Safe Fallback:** When `FlagsFeature` not enabled, always uses graceful behavior (can't determine policy without config)

**Documentation:**
- Comprehensive KDoc with environment-specific guidance
- Added `@throws IllegalStateException` documentation to `build()` and `get()` methods

**Intentional deviations from spec:**
- Kept `NoOpFlagsClient` name instead of `FallbackClient`

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
@@ -251,7 +251,8 @@ interface FlagsClient {
                 createInternal(
                     configuration = flagsFeature.flagsConfiguration,
                     featureSdkCore = sdkCore,
-                    flagsFeature = flagsFeature
+                    flagsFeature = flagsFeature,
+                    name = name
                 )
             }
         }
@@ -344,7 +345,7 @@ interface FlagsClient {
             configuration: FlagsConfiguration,
             featureSdkCore: FeatureSdkCore,
             flagsFeature: FlagsFeature,
-            name: String = DEFAULT_CLIENT_NAME
+            name: String
         ): FlagsClient {
             val executorService = featureSdkCore.createSingleThreadExecutorService(
                 executorContext = FLAGS_CLIENT_EXECUTOR_NAME
@@ -380,7 +381,7 @@ interface FlagsClient {
                     }
                 )
             } else {
-                // Build a [DatadogFlagsClient] instance from its dependencies.
+                // Build the various dependencies for the [DatadogFlagsClient]
 
                 val flagsContext = FlagsContext(
                     applicationId = applicationId,
--- a/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/NoOpFlagsClientTest.kt
+++ b/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/NoOpFlagsClientTest.kt
@@ -73,7 +73,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log error W setEvaluationContext()`(forge: Forge) {
+    fun `M log warning W setEvaluationContext()`(forge: Forge) {
         // Given
         val fakeContext = EvaluationContext(forge.anAlphabeticalString(), emptyMap())
 
@@ -123,7 +123,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log error W resolveBooleanValue()`(forge: Forge) {
+    fun `M log warning W resolveBooleanValue()`(forge: Forge) {
         // Given
         val fakeFlagKey = forge.anAlphabeticalString()
         val fakeDefaultValue = forge.aBool()
@@ -173,7 +173,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log error W resolveStringValue()`(forge: Forge) {
+    fun `M log warning W resolveStringValue()`(forge: Forge) {
         // Given
         val fakeFlagKey = forge.anAlphabeticalString()
         val fakeDefaultValue = forge.anAlphabeticalString()
@@ -236,7 +236,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log error W resolveDoubleValue()`(forge: Forge) {
+    fun `M log warning W resolveDoubleValue()`(forge: Forge) {
         // Given
         val fakeFlagKey = forge.anAlphabeticalString()
         val fakeDefaultValue = forge.aDouble()
@@ -299,7 +299,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log error W resolveIntValue()`(forge: Forge) {
+    fun `M log warning W resolveIntValue()`(forge: Forge) {
         // Given
         val fakeFlagKey = forge.anAlphabeticalString()
         val fakeDefaultValue = forge.anInt()
@@ -350,7 +350,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log error W resolve()`(forge: Forge) {
+    fun `M log warning W resolve()`(forge: Forge) {
         // Given
         val fakeFlagKey = forge.anAlphabeticalString()
         val fakeDefaultValue = forge.anAlphabeticalString()
@@ -405,7 +405,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log error W resolveStructureValue()`(forge: Forge) {
+    fun `M log warning W resolveStructureValue()`(forge: Forge) {
         // Given
         val fakeFlagKey = forge.anAlphabeticalString()
 
--- a/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
+++ b/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
@@ -3,7 +3,9 @@
  * This product includes software developed at Datadog (https://www.datadoghq.com/).
  * Copyright 2016-Present Datadog, Inc.
  */
+
 package com.datadog.android.flags.internal
+
 import android.content.Context
 import android.content.pm.ApplicationInfo
 import com.datadog.android.api.InternalLogger
@@ -155,7 +157,7 @@ internal class FlagsFeatureTest {
     @Test
     fun `M log through internalLogger W logErrorWithPolicy() { release build }`() {
         // Given - release build (not debuggable)
-        val releaseAppInfo = ApplicationInfo().apply { flags = 0 }
+        val releaseAppInfo = ApplicationInfo()
         val releaseContext = mock<Context>()
         whenever(releaseContext.applicationInfo).thenReturn(releaseAppInfo)
         val config = FlagsConfiguration.Builder().build()
@@ -179,7 +181,7 @@ internal class FlagsFeatureTest {
     @Test
     fun `M log through internalLogger W logErrorWithPolicy() { release build, gracefulModeEnabled false }`() {
         // Given - release build should ignore gracefulModeEnabled setting
-        val releaseAppInfo = ApplicationInfo().apply { flags = 0 }
+        val releaseAppInfo = ApplicationInfo()
         val releaseContext = mock<Context>()
         whenever(releaseContext.applicationInfo).thenReturn(releaseAppInfo)
         val config = FlagsConfiguration.Builder()
@@ -205,9 +207,7 @@ internal class FlagsFeatureTest {
     @Test
     fun `M log to Android Logcat W logErrorWithPolicy() { debug build, graceful enabled }`() {
         // Given - debug build with gracefulModeEnabled=true
-        val debugAppInfo = ApplicationInfo().apply {
-            flags = ApplicationInfo.FLAG_DEBUGGABLE
-        }
+        val debugAppInfo = ApplicationInfo().apply { flags = ApplicationInfo.FLAG_DEBUGGABLE }
         val debugContext = mock<Context>()
         whenever(debugContext.applicationInfo).thenReturn(debugAppInfo)
         val config = FlagsConfiguration.Builder()
@@ -227,9 +227,7 @@ internal class FlagsFeatureTest {
     @Test
     fun `M crash W logErrorWithPolicy() { debug build, graceful disabled, shouldCrash true }`() {
         // Given - debug build with strict mode
-        val debugAppInfo = ApplicationInfo().apply {
-            flags = ApplicationInfo.FLAG_DEBUGGABLE
-        }
+        val debugAppInfo = ApplicationInfo().apply { flags = ApplicationInfo.FLAG_DEBUGGABLE }
         val debugContext = mock<Context>()
         whenever(debugContext.applicationInfo).thenReturn(debugAppInfo)
         val config = FlagsConfiguration.Builder()
@@ -251,9 +249,7 @@ internal class FlagsFeatureTest {
     @Test
     fun `M log to Android Logcat W logErrorWithPolicy() { debug build, graceful disabled, shouldCrash false }`() {
         // Given - debug build with strict mode but shouldCrashInStrict=false
-        val debugAppInfo = ApplicationInfo().apply {
-            flags = ApplicationInfo.FLAG_DEBUGGABLE
-        }
+        val debugAppInfo = ApplicationInfo().apply { flags = ApplicationInfo.FLAG_DEBUGGABLE }
         val debugContext = mock<Context>()
         whenever(debugContext.applicationInfo).thenReturn(debugAppInfo)
         val config = FlagsConfiguration.Builder()


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -201,10 +201,14 @@ interface FlagsClient {
          * 3. Returns the created client or [NoOpFlagsClient] on failure
          *
          * If a [FlagsClient] with the same name already exists for this [SdkCore]:
-         * - Logs a warning (WARN level, USER target)
+         * - Logs a warning according to the graceful mode policy
          * - Returns the existing [FlagsClient]
+         * - In strict mode (debug builds with gracefulModeEnabled=false), throws an exception
          *
          * @return the created [FlagsClient], existing client, or [NoOpFlagsClient].
+         * @throws IllegalStateException in strict mode (debug builds with gracefulModeEnabled=false)
```
typotter, line 218: thrown by Kotlin's builtin `error` function in the `logWithPolicy` call

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -71,6 +73,48 @@ data class FlagsConfiguration internal constructor(
...
+         *
+         * ### Development Environment
+         * - **Recommended:** `gracefulModeEnabled(false)` - Strict mode
+         * - **Behavior:** Crashes immediately on misuse (debug builds only)
+         * - **Use when:** Writing new code, running unit tests, debugging integration issues
+         * - **Benefit:** Catch configuration errors early, fail fast during development
+         *
+         * ### QA/Staging Environment
+         * - **Recommended:** `gracefulModeEnabled(true)` - Error mode
+         * - **Behavior:** Logs errors to System.err but doesn't crash (debug builds)
```
0xnm: There is no use of `System.err`, there is use of Logcat, but only if message targets `USER` and appropriate level is enabled with `Datadog.setVerbosity` call.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -71,6 +73,48 @@ data class FlagsConfiguration internal constructor(
...
+         * ### Production Environment
+         * - **Automatic:** Graceful mode (release builds always use this)
+         * - **Behavior:** Silent fallback with conditional SDK logging
+         * - **Use when:** Deployed to end users
+         * - **Benefit:** Maximum stability, SDK continues functioning with defaults
+         *
+         * ## Technical Details
+         *
+         * Policy selection logic:
+         * - **Release builds:** Always use graceful policy (safety override)
```
0xnm: does it mean it should be enabled or disabled?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -64,6 +77,85 @@ internal class FlagsFeature(
...
+                sdkCore.internalLogger.log(
+                    level,
+                    InternalLogger.Target.USER,
+                    { formattedMessage }
+                )
+            }
+
+            // Debug + gracefulMode enabled: Error policy (always log to console)
+            flagsConfiguration.gracefulModeEnabled -> {
+                System.err.println(formattedMessage)
```
0xnm: We shouldn't use `System.log` or `System.err` for logging. We should use `FeatureSdkCore.internalLogger` instead, because:

* it is bound to the particular SDK core
* it respects verbosity set by `Datadog.setVerbosity`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -64,6 +77,85 @@ internal class FlagsFeature(
...
+     * - Debug + gracefulMode disabled: Strict (crash immediately)
+     *
+     * @param message The error message
+     * @param level The log level for conditional logging (graceful policy)
+     * @param shouldCrashInStrict If true, crashes in strict policy
+     */
+    internal fun logErrorWithPolicy(
+        message: String,
+        level: InternalLogger.Level = InternalLogger.Level.ERROR,
+        shouldCrashInStrict: Boolean = false
```
0xnm: This is a controversial approach, because customers may choose strict mode by mistake in the release builds and get a crash in production.

Are we doing the same in iOS SDK?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -38,6 +38,19 @@ internal class FlagsFeature(
...
+
+    @Volatile
+    private var initialized = false
+
+    /**
+     * Cached debug build state to avoid repeated ApplicationInfo.flags checks.
+     * Primed during onInitialize() and cleared during onStop().
+     */
+    @Volatile
+    private var debugBuildCache: Boolean? = null
```
0xnm: I don't think we need `debugBuildCache` and `isDebugBuild` as computed property.

Under the hood it relies on `ctx.applicationInfo.flags and ApplicationInfo.FLAG_DEBUGGABLE` which is a static information and won't change for the current process, so you can just read it and store to the `isDebugBuild` property without making it computed.

This also removes the need to store `Context` as a property.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -64,6 +77,85 @@ internal class FlagsFeature(
...
+                    { formattedMessage }
+                )
+            }
+
+            // Debug + gracefulMode enabled: Error policy (always log to console)
+            flagsConfiguration.gracefulModeEnabled -> {
+                System.err.println(formattedMessage)
+            }
+
+            // Debug + gracefulMode disabled: Strict policy (crash if requested)
```
jonathanmos: It looks like we may be missing test coverage for gracefulmode

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -71,6 +73,48 @@ data class FlagsConfiguration internal constructor(
             return this
         }
 
+        /**
```
jonathanmos, line 79: We could think about shortening this documentation. It could be detailed in the public docs but in the code this seems like a lot

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: strict mode cannot be enabled in release builds - the graceful mode option is overridden by the build type below. When `!isDebugBuild`, we log with the internal logger only

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: We wanted to ensure the surfacing of these misuse errors.
@ncreated - would you mind weighing in?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: fair. shortened

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: that's a good question - this is worded ambiguously. I have updated the docs to clarify

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: right, of course. changed!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: added

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
ncreated: 👋 Yes — the goal here is indeed to bypass `Datadog.setVerbosity` and `internalLogger`. This supports _Graceful Mode_ - a setting specific to Datadog Flags library.

**Why:** Since feature flags are critical to the user experience (unlike other telemetry collected by the SDK), we want to ensure configuration errors are loud in development. Bypassing the internalLogger helps with that — otherwise, developers might miss configuration issues and ship a broken setup to production.

We’ve applied the same design in dd-sdk-ios.

That said, I don’t think `System.err` is the right tool for this - we should use Logcat’s `Log.e()` instead, no?

WDYT, @0xnm?

typotter: Thanks Maciek. Agree that I should be using `Log.e` here instead of system.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
0xnm: Yes, Logcat (`Log` class) should be used. Although Android using Java APIs, it doesn't have a concept of `stdout` vs `stderr`, so effectively under the hood everything is redirected to Logcat when calling `System.out / System.log / System.err`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt
```diff
@@ -71,6 +73,26 @@ data class FlagsConfiguration internal constructor(
...
+         * accessing non-existent clients.
+         *
+         * This setting has no impact on release builds. Release builds will always fail "gracefully".
+         *
+         * - **Debug (gracefulModeEnabled == false):** Crashes immediately to catch errors early
+         * - **Debug (gracefulModeEnabled == true):** Logs to Android Logcat at ERROR level
+         * - **Release:** Always uses graceful mode regardless of this setting
+         *
+         * @param enabled Whether to enable graceful mode in debug builds (default: true)
+         * @return this Builder instance for method chaining
```
0xnm:
```suggestion
         * @return this [Builder] instance for method chaining
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
```diff
@@ -338,10 +366,14 @@ interface FlagsClient {
                     { "Missing required context parameters: $missingParams" }
                 )
 
+                val logWithPolicy: (String, InternalLogger.Level) -> Unit = { message, level ->
```
0xnm: maybe it is worth to create typealias for `(String, InternalLogger.Level) -> Unit`

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
```diff
@@ -337,19 +337,77 @@ internal class FlagsClientTest {
         assertThat(client2).isInstanceOf(NoOpFlagsClient::class.java)
     }
 
+    @Test
+    fun `M throw IllegalArgumentException W Builder() {blank name}`() {
+        // When/Then
+        org.junit.jupiter.api.assertThrows<IllegalArgumentException> {
```
0xnm: this needs to go to import

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
```diff
@@ -337,19 +337,77 @@ internal class FlagsClientTest {
...
+        val client = FlagsClient.Builder(sdkCore = mockUninitializedCore).build()
 
         // Then - should return NoOpClient
         assertThat(client).isInstanceOf(NoOpFlagsClient::class.java)
+
+        // Verify error was logged
+        verify(mockLogger).log(
+            eq(InternalLogger.Level.ERROR),
+            eq(InternalLogger.Target.USER),
+            any(),
```
0xnm: It is better to verify the message as well, otherwise it can be an error for something else. Same for the test below.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -57,6 +58,12 @@ internal class FlagsFeatureTest {
         whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
         whenever(mockSdkCore.createSingleThreadExecutorService(any())) doReturn mockExecutorService
 
+        // Setup mockContext with default release build (flags = 0)
+        val applicationInfo = android.content.pm.ApplicationInfo().apply {
```
0xnm: also needs to go to `import`

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
 
     // endregion
 
+    // region Graceful Mode Policy
+
+    @Test
+    fun `M log through internalLogger W logErrorWithPolicy() { release build }`() {
+        // Given - release build (not debuggable)
+        val releaseAppInfo = android.content.pm.ApplicationInfo().apply { flags = 0 }
+        val releaseContext = org.mockito.kotlin.mock<Context>()
```
0xnm:
```suggestion
        val releaseAppInfo = ApplicationInfo().apply { flags = 0 }
        val releaseContext = mock<Context>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+        testedFeature.onInitialize(releaseContext)
+
+        // When
+        testedFeature.logErrorWithPolicy("test message", InternalLogger.Level.ERROR)
+
+        // Then
+        verify(mockInternalLogger).log(
+            eq(InternalLogger.Level.ERROR),
+            eq(InternalLogger.Target.USER),
+            any(),
```
0xnm: it is better to assert the message as well

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+            eq(false),
+            eq(null)
+        )
+    }
+
+    @Test
+    fun `M log through internalLogger W logErrorWithPolicy() { release build, gracefulModeEnabled false }`() {
+        // Given - release build should ignore gracefulModeEnabled setting
+        val releaseAppInfo = android.content.pm.ApplicationInfo().apply { flags = 0 }
+        val releaseContext = org.mockito.kotlin.mock<Context>()
```
0xnm:
```suggestion
        val releaseAppInfo = ApplicationInfo().apply { flags = 0 }
        val releaseContext = mock<Context>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+        )
+    }
+
+    @Test
+    fun `M log to Android Logcat W logErrorWithPolicy() { debug build, graceful enabled }`() {
+        // Given - debug build with gracefulModeEnabled=true
+        val debugAppInfo = android.content.pm.ApplicationInfo().apply {
+            flags = android.content.pm.ApplicationInfo.FLAG_DEBUGGABLE
+        }
+        val debugContext = org.mockito.kotlin.mock<Context>()
```
0xnm:
```suggestion
        val debugAppInfo = ApplicationInfo().apply {
            flags = ApplicationInfo.FLAG_DEBUGGABLE
        }
        val debugContext = mock<Context>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+            eq(null),
+            eq(false),
+            eq(null)
+        )
+    }
+
+    @Test
+    fun `M log through internalLogger W logErrorWithPolicy() { release build, gracefulModeEnabled false }`() {
+        // Given - release build should ignore gracefulModeEnabled setting
+        val releaseAppInfo = android.content.pm.ApplicationInfo().apply { flags = 0 }
```
0xnm: probably apply is redundant, because `ApplicationInfo` is created by default with `flags = 0`

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+        org.mockito.kotlin.verifyNoInteractions(mockInternalLogger)
+    }
+
+    @Test
+    fun `M crash W logErrorWithPolicy() { debug build, graceful disabled, shouldCrash true }`() {
+        // Given - debug build with strict mode
+        val debugAppInfo = android.content.pm.ApplicationInfo().apply {
+            flags = android.content.pm.ApplicationInfo.FLAG_DEBUGGABLE
+        }
+        val debugContext = org.mockito.kotlin.mock<Context>()
```
0xnm:
```suggestion
        val debugAppInfo = android.content.pm.ApplicationInfo().apply {
            flags = ApplicationInfo.FLAG_DEBUGGABLE
        }
        val debugContext = mock<Context>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+        }
+    }
+
+    @Test
+    fun `M log to Android Logcat W logErrorWithPolicy() { debug build, graceful disabled, shouldCrash false }`() {
+        // Given - debug build with strict mode but shouldCrashInStrict=false
+        val debugAppInfo = android.content.pm.ApplicationInfo().apply {
+            flags = android.content.pm.ApplicationInfo.FLAG_DEBUGGABLE
+        }
+        val debugContext = org.mockito.kotlin.mock<Context>()
```
0xnm:
```suggestion
        val debugAppInfo = ApplicationInfo().apply {
            flags = ApplicationInfo.FLAG_DEBUGGABLE
        }
        val debugContext = mock<Context>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+        // When
+        testedFeature.logErrorWithPolicy(
+            "test message",
+            InternalLogger.Level.ERROR,
+            shouldCrashInStrict = false
+        )
+
+        // Then - uses android.util.Log.e which can't be easily verified in unit tests
+        // Just verify it doesn't crash and doesn't use internalLogger
+        org.mockito.kotlin.verifyNoInteractions(mockInternalLogger)
```
0xnm:
```suggestion
        verifyNoInteractions(mockInternalLogger)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+
+        // When
+        testedFeature.logErrorWithPolicy(
+            "test message",
+            InternalLogger.Level.ERROR,
+            shouldCrashInStrict = false
+        )
+
+        // Then - uses android.util.Log.e which can't be easily verified in unit tests
+        // Just verify it doesn't crash and doesn't use internalLogger
```
0xnm: you can also wrap `logErrorWithPolicy` above in `assertNotThrows` to explicitly highlight the intention.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+            .build()
+        testedFeature = FlagsFeature(mockSdkCore, config)
+        testedFeature.onInitialize(debugContext)
+
+        // When
+        testedFeature.logErrorWithPolicy("test message", InternalLogger.Level.ERROR)
+
+        // Then - uses android.util.Log.e which can't be easily verified in unit tests
+        // Just verify it doesn't crash and doesn't use internalLogger
+        org.mockito.kotlin.verifyNoInteractions(mockInternalLogger)
```
0xnm:
```suggestion
        verifyNoInteractions(mockInternalLogger)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -141,6 +148,132 @@ internal class FlagsFeatureTest {
...
+        val debugContext = org.mockito.kotlin.mock<Context>()
+        whenever(debugContext.applicationInfo).thenReturn(debugAppInfo)
+        val config = FlagsConfiguration.Builder()
+            .gracefulModeEnabled(false)
+            .build()
+        testedFeature = FlagsFeature(mockSdkCore, config)
+        testedFeature.onInitialize(debugContext)
+
+        // When/Then
+        org.junit.jupiter.api.assertThrows<IllegalStateException> {
```
0xnm:
```suggestion
        assertThrows<IllegalStateException> {
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -38,6 +39,16 @@ internal class FlagsFeature(
 ) : StorageBackedFeature,
     FeatureContextUpdateReceiver {
 
+    @Volatile
+    private var initialized = false
```
0xnm:
```suggestion
    private var isInitialized = false
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -64,6 +75,52 @@ internal class FlagsFeature(
...
+     * Policy Selection:
+     * - Release builds: Graceful (log through SDK logger - conditional)
+     * - Debug + gracefulMode enabled: Error (log to Android Logcat)
+     * - Debug + gracefulMode disabled: Strict (crash immediately)
+     *
+     * @param message The error message
+     * @param level The log level for conditional logging (graceful policy)
+     * @param shouldCrashInStrict If true, crashes in strict policy
+     */
+    internal fun logErrorWithPolicy(
```
0xnm: this method should be moved below, after the `override` methods

0xnm (changes requested):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/featureflags/FlagsClient.kt
typotter: yes, agreed. helps to define the interaction.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: done

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
typotter: done

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/featureflags/FlagsClientTest.kt
typotter: done

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
typotter: done

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
typotter: done

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
typotter: thanks!

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
typotter: 👍

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -307,14 +343,14 @@ interface FlagsClient {
         internal fun createInternal(
             configuration: FlagsConfiguration,
             featureSdkCore: FeatureSdkCore,
-            flagsFeature: FlagsFeature
+            flagsFeature: FlagsFeature,
+            name: String = DEFAULT_CLIENT_NAME
```
0xnm: is default argument needed here, given it is already there for public API?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -330,19 +366,22 @@ interface FlagsClient {
...
+                    name = name,
                     reason = "Failed to create client - missing SDK context parameters: $missingParams",
-                    logger = internalLogger
+                    logWithPolicy = { message, level ->
+                        flagsFeature.logErrorWithPolicy(message, level, shouldCrashInStrict = false)
+                    }
                 )
             } else {
-                // Create FlagsContext combining core SDK context with feature configuration
+                // Build a [DatadogFlagsClient] instance from its dependencies.
```
0xnm: this comment is wrong

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/NoOpFlagsClientTest.kt
```diff
@@ -121,7 +123,7 @@ internal class NoOpFlagsClientTest {
     }
 
     @Test
-    fun `M log critical error W resolveBooleanValue()`(forge: Forge) {
+    fun `M log error W resolveBooleanValue()`(forge: Forge) {
```
0xnm: although here and for other tests in this file it is not error, but warning

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
```diff
@@ -3,10 +3,9 @@
  * This product includes software developed at Datadog (https://www.datadoghq.com/).
  * Copyright 2016-Present Datadog, Inc.
  */
-
 package com.datadog.android.flags.internal
-
```
0xnm, line 8: seems like redundant edits

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
typotter: negative

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
typotter: Comment is meant to apply to the whole block here.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsFeatureTest.kt
typotter: fixed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/NoOpFlagsClientTest.kt
typotter: fixed.

PR #2968, opened 2025-10-24
RUM-12225: Send TTID value as TTFD to rum intake if reportAppFullyDisplayed was called too early

aleksandr-gringauz:
### What does this PR do?

If `RumMonitor.reportAppFullyDisplayed` was called before TTID was computed we don't send this TTFD value to the intake, but rather send TTID value (computed later) as TTFD.

This aligns with current iOS implementation.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
@@ -63,8 +63,8 @@ internal class RumSessionScopeStartupManagerImpl(
 
     private var lastScenario: RumStartupScenario? = null
 
-    private var ttfdEventForScenario: RumRawEvent.AppStartTTFDEvent? = null
-    private var ttidEventForScenario: RumRawEvent.AppStartTTIDEvent? = null
+    private var ttfdReportedForScenario: Boolean = false
+    private var ttidReportedForScenario: Boolean = false
 
     private var ttfdReportedForSession = false
     private var ttidSentForSession = false
@@ -74,8 +74,8 @@ internal class RumSessionScopeStartupManagerImpl(
     override fun onAppStartEvent(event: RumRawEvent.AppStartEvent) {
         lastScenario = event.scenario
 
-        ttfdEventForScenario = null
-        ttidEventForScenario = null
+        ttfdReportedForScenario = false
+        ttidReportedForScenario = false
 
         appStartCount++
     }
@@ -88,7 +88,7 @@ internal class RumSessionScopeStartupManagerImpl(
         rumContext: RumContext,
         customAttributes: Map<String, Any?>
     ) {
-        ttidEventForScenario = event
+        ttidReportedForScenario = true
 
         rumAppStartupTelemetryReporter.reportTTID(
             info = event.info,
@@ -123,7 +123,7 @@ internal class RumSessionScopeStartupManagerImpl(
             )
         }.submit()
 
-        if (ttfdEventForScenario != null) {
+        if (ttfdReportedForScenario) {
             /**
              * [com.datadog.android.rum.RumMonitor.reportAppFullyDisplayed] was called earlier than TTID
              * was computed, reporting TTID value also as TTFD.
@@ -171,9 +171,9 @@ internal class RumSessionScopeStartupManagerImpl(
             return
         }
 
-        ttfdEventForScenario = event
+        ttfdReportedForScenario = true
 
-        if (ttidEventForScenario == null) {
+        if (!ttidReportedForScenario) {
             sdkCore.internalLogger.log(
                 level = InternalLogger.Level.WARN,
                 target = InternalLogger.Target.USER,


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -113,8 +122,25 @@ internal class RumSessionScopeStartupManagerImpl(
                 )
             )
         }.submit()
+
+        if (ttfdEventForScenario != null) {
+            /**
+             * [com.datadog.android.rum.RumMonitor.reportAppFullyDisplayed] was called earlier than TTID
+             * was computed, reporting TTID value also as TTFD.
+             */
+            sendTTFDEvent(
```
ambushwork, line 131: A little bit confused between the code here and the description in this PR, in the description it says "we don't report TTFD" in this case, but you are still reporting here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -145,6 +171,42 @@ internal class RumSessionScopeStartupManagerImpl(
...
+                    REPORT_APP_FULLY_DISPLAYED_CALLED_BEFORE_TTID_MESSAGE
+                },
+                throwable = null,
+                onlyOnce = false,
+                additionalProperties = null
+            )
+            return
+        }
+
+        sendTTFDEvent(
```
ambushwork, line 190: If `onTTFDEvent` happens before `onTTIDEvent`, this function will be called twice if I didn't miss anything?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -113,8 +122,25 @@ internal class RumSessionScopeStartupManagerImpl(
...
+                rumContext = rumContext,
+                customAttributes = customAttributes,
+                durationNs = event.info.durationNs,
+                scenario = event.info.scenario
+            )
+        }
     }
 
+    @Suppress("ReturnCount")
     override fun onTTFDEvent(
```
ambushwork, line 144: Should we consider concurrency risk between `onTTIDEvent` and `onTTFDEvent`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
aleksandr-gringauz: No, in this case we will return [here](https://github.com/DataDog/dd-sdk-android/blob/329adcbbd0d975966398490be6278d21aa3d7b17/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt#L187)

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
aleksandr-gringauz: all methods of `RumSessionScopeStartupManager` are called on the thread of rum pipeline [executor](https://github.com/DataDog/dd-sdk-android/blob/ee38a9a53bfc2a7f4b666533d7bc69e13dc6bbe3/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/Rum.kt#L140). So there shouldn't be any concurrency issues here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
aleksandr-gringauz: Yes, the description is a bit confusing, edited the PR name and the description a bit.

About this particular line. We send the computed TTID duration as TTFD to the intake. This is what this PR is all about.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -113,8 +122,25 @@ internal class RumSessionScopeStartupManagerImpl(
                 )
             )
         }.submit()
+
+        if (ttfdEventForScenario != null) {
```
ambushwork: should we use `ttfdReportedForSession` flag here? it seems more understandable here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -63,13 +63,20 @@ internal class RumSessionScopeStartupManagerImpl(
 
     private var lastScenario: RumStartupScenario? = null
 
-    private var ttfdSentForSession = false
+    private var ttfdEventForScenario: RumRawEvent.AppStartTTFDEvent? = null
+    private var ttidEventForScenario: RumRawEvent.AppStartTTIDEvent? = null
```
0xnm: it seems these can be simply boolean flags (if they are indeed needed on top of existing ones)? we write event object there, but the only thing we do is checking if it is `null` or not.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
aleksandr-gringauz: `ttfdReportedForSession` is a flag that is responsible for the restriction that there should be only 1 TTFD per RUM session.

`ttfdEventForScenario` - is a variable about "scenario". "scenario" is a single app launch process. There can be many of them during 1 session.

This particular `if` statement checks that if `reportAppFullyDisplayed` was called for this "scenario", we need to send TTID duration value as TTFD to the intake.

If we use `ttfdReportedForSession` here we will send TTID duration value as TTFD for all 2nd, 3rd, etc TTID in a session, which is not what we want.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
aleksandr-gringauz: changed to `bool`s

PR #2974, opened 2025-10-31
RUM-11897: Add scripts to set/get Vault secrets & Migrate CI

ambushwork:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/ci/pipelines/default-pipeline.yml
+++ b/ci/pipelines/default-pipeline.yml
@@ -13,6 +13,10 @@ stages:
   - notify
 
 .snippets:
+  source-secrets:
+    - source ./ci/scripts/vault_config.sh
+    - source ./ci/scripts/get-secret.sh
+
   # macOS AMI will already have cmdline-tools installed
   install-android-api-components:
     - echo y | ~/android_sdk/cmdline-tools/latest/bin/sdkmanager --install "emulator"
@@ -39,12 +43,12 @@ stages:
     - if [[ "$exit_code" -ne 0 ]]; then exit 1; fi
     - exit 0
   set-publishing-credentials:
-    - aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.gradle-properties --with-decryption --query "Parameter.Value" --out text >> ./gradle.properties
-    - export GPG_PRIVATE_KEY=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.signing.gpg_private_key --with-decryption --query "Parameter.Value" --out text)
-    - export GPG_PASSWORD=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.signing.gpg_passphrase --with-decryption --query "Parameter.Value" --out text)
-    - export CENTRAL_PUBLISHER_USERNAME=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.publishing.central_username --with-decryption --query "Parameter.Value" --out text)
-    - export CENTRAL_PUBLISHER_PASSWORD=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.publishing.central_password --with-decryption --query "Parameter.Value" --out text)
-    - export GPG_PUBLIC_FINGERPRINT=$(aws ssm get-parameter --region us-east-1 --name ci.dd-sdk-android.signing.gpg_public_key --with-decryption --query "Parameter.Value" --out text | gpg --import --import-options show-only | grep -E -o -e "[A-F0-9]{40}")
+    - get_secret $DD_ANDROID_SECRET__GRADLE_PROPERTIES > ./gradle.properties
+    - export GPG_PRIVATE_KEY=$(get_secret $DD_ANDROID_SECRET__SIGNING_GPG_PRIVATE_KEY)
+    - export GPG_PASSWORD=$(get_secret $DD_ANDROID_SECRET__SIGNING_GPG_PASSPHRASE)
+    - export CENTRAL_PUBLISHER_USERNAME=$(get_secret $DD_ANDROID_SECRET__PUBLISHING_CENTRAL_USERNAME)
+    - export CENTRAL_PUBLISHER_PASSWORD=$(get_secret $DD_ANDROID_SECRET__PUBLISHING_CENTRAL_PWD)
+    - export GPG_PUBLIC_FINGERPRINT=$(get_secret $DD_ANDROID_SECRET__SIGNING_GPG_PUBLIC_KEY | gpg --import --import-options show-only | grep -E -o -e "[A-F0-9]{40}")
 
 # CI IMAGE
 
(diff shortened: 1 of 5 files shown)


ci/pipelines/default-pipeline.yml
```diff
@@ -39,15 +44,27 @@ stages:
...
+    - export GPG_PASSWORD=$(cat ./ci/pipelines/secrets/gpg_passphrase)
+    - export CENTRAL_PUBLISHER_USERNAME=$(cat ./ci/pipelines/secrets/central_username)
+    - export CENTRAL_PUBLISHER_PASSWORD=$(cat ./ci/pipelines/secrets/central_password)
+    - export GPG_PUBLIC_FINGERPRINT=$(cat ./ci/pipelines/secrets/gpg_public_key | gpg --import --import-options show-only | grep -E -o -e "[A-F0-9]{40}")
 
 # CI IMAGE
 
+fetch-secrets:
+  stage: fetch-secrets
+  tags: ["macos:sonoma","specific:true"]
```
0xnm: Is there a benefit of using macOS runner for that?

Also if it indeed should be macOS runner, then `image` directive below has no effect.

ci/pipelines/default-pipeline.yml
```diff
@@ -39,15 +44,27 @@ stages:
...
+  image: $CI_IMAGE_DOCKER
+  script:
+    - !reference [.snippets, fetch-secrets]
+  artifacts:
+    paths:
+      - ./ci/pipelines/secrets/
+    expire_in: 1 hour
+    when: always
```
0xnm: Is it a safe thing to have? it means that anyone with just a basic Gitlab CI access can download this, not only RUM team members.

Moreover, it makes this artifact available for 1 hour, meaning it is still there even after pipeline completed, failed.

Should we maybe fetch secrets only in the boundaries of the particular job?

ci/pipelines/default-pipeline.yml
ambushwork: That's an obligation, only MacOS runner has the correct IAM role to access Vault of our project.

ci/pipelines/default-pipeline.yml
ambushwork: since the default behavior will be only uploading artifacts when success, do you think remove `when:always` will improve this? 
fetching secrets in each job may cause repeat `vault login` and `vault get`, it may increase the CI time, I can ask around if this is a safe way

ci/pipelines/default-pipeline.yml
0xnm: > only MacOS runner has the correct IAM role

Why so? AFAIK every repo has a unique service account associated, so maybe we can list it as well as account which can have this role?

The fleet of macOS runners is _much_ limited in size, while k8s runners are faster to get and run.

Not a blocker though.

ci/pipelines/default-pipeline.yml
0xnm: > fetching secrets in each job may cause repeat vault login and vault get

It should be fast. Anyway, by constraining secrets lifetime only to the particular job and removing the possibility to fetch them as an artifact we limit access only to the Repo Admins/Contributors (it is still possible to modify CI script to read them) compared to the Job artifacts access which is possible for everyone with dev access.

ambushwork:
Updated the PR, main changes:
* Now all the secrets are stored at path `kv/+/k8s/gitlab-runner/dd-sdk-android` for k8s gitlab-runners, path for macOS runners are no longer used.
* K8s gitlab-runners don't require vault login or setting VAULT_ADDR, so login process is removed for CI
* Every job retrieve its own key on demand, no artifacts are created for secrets

ci/scripts/fetch-secrets.sh
0xnm, line 1: Having this file is a bit dangerous, because secrets are not supposed to be fetched in bulk in local.

I would propose to delete it.

ci/scripts/list-secrets.sh
```diff
@@ -0,0 +1,38 @@
...
+    RESET="\e[0m"
+
+    echo "Available secrets:"
+    for key in ${(k)DD_ANDROID_SECRETS}; do
+        IFS=" | " read -r name description <<< "${DD_ANDROID_SECRETS[$key]}"
+        echo "$key) ${GREEN}$name${RESET} - $description"
+    done | sort -n
+
+    echo ""
+    echo "To add a new secret, first define it in 'tools/secrets/config.sh' and retry."
```
0xnm: I don't think we have this file

ci/scripts/vault_config.sh
```diff
@@ -0,0 +1,72 @@
...
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__PUBLISHING_CENTRAL_USERNAME | Username for publishing artifacts to Sonatype Maven repository."
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__PUBLISHING_CENTRAL_PWD | Password for publishing artifacts to Sonatype Maven repository."
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__API_KEY | API key for sending CI App reports to org2."
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__APP_KEY | Application key for sending CI App reports to org2."
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__CODECOV_TOKEN | CodeCov token for unit test jobs."
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__KEYSTORE | Android signing keystore for building all APKs for synthetics."
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__KEYSTORE_PWD | Android signing password for building all APKs for synthetics."
+DD_ANDROID_SECRETS[$((idx++))]="$DD_ANDROID_SECRET__E2E_CONFIG_JSON | config.json for uploading an end-to-end APK to the Mobile Integration org (529432, https://mobile-integration.datadoghq.com/)."
```
0xnm: I would suggest to remove Org IDs and URLs here and everywhere below, having a name is enough

ci/scripts/get-secret.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/zsh
```
0xnm, line 1: we assume that Z Shell is always installed? Do these scripts use extensions available only in zsh?

ambushwork: yes, it works on MacOS and for k8s runner so we can assume that. we use this one `declare -A DD_ANDROID_SECRETS` in `vault_config.sh` which is only in zsh

ci/scripts/fetch-secrets.sh
ambushwork: indeed, this is not needed since we don't fetch them into an artifact, removed

ci/scripts/list-secrets.sh
ambushwork: path updated

ci/scripts/vault_config.sh
ambushwork: updated

aleksandr-gringauz (approved):

PR #2975, opened 2025-10-31
RUM-12062: Refactor the configuration and integration of the RUM Debug Widget

kikoveiga:
### What does this PR do?

- Simplifies the configuration of the RUM Debug Widget. Customers can now simply use `enableRumDebugWidget()` in their `RumConfiguration.Builder`.
- Adds documentation for this configuration in the `README.md`.
- Replaced manual overlay lifecycle management with automatic `OverlayManager` using activity lifecycle callbacks.

### Motivation

Configuration was too complex for customers, where some unnecessary APIs had to be exposed. The `DefaultInsightsOverlay` had to be manually attached to every activity by customers.


features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
```diff
@@ -0,0 +1,231 @@
...
+            val overlayView = activity.layoutInflater.inflate(
+                R.layout.layout_dd_insights_overlay,
+                parent,
+                false
+            )
+
+            root = overlayView
+
+            viewName = overlayView.findViewById(R.id.view_name)
+            timelineView = overlayView.findViewById<TimelineView?>(R.id.timeline)?.apply {
```
aleksandr-gringauz: `overlayView.findViewById<TimelineView?>(R.id.timeline)?.apply` -> `overlayView.findViewById<TimelineView>(R.id.timeline)?.apply`

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
```diff
@@ -0,0 +1,231 @@
...
+                setOnClickListener {
+                    toggleOverlay(storage = storage, showWidget = true, widget = widgetView, fab = this)
+                }
+                setOnTouchListener(DragTouchListener(onUp = { storage.fabPosition = x to y }))
+                restoreVisibility(!storage.widgetDisplayed)
+            }
+        }
+
+        root?.let { v ->
+            if (v.parent != parent) {
```
aleksandr-gringauz, line 127: Why is this `if` needed? If I understand correctly it is always `false`, because we only set `root` using the code above:

```
val overlayView = activity.layoutInflater.inflate(
    R.layout.layout_dd_insights_overlay,
    parent,
    false
)

root = overlayView
```

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
```diff
@@ -0,0 +1,231 @@
...
+                R.layout.layout_dd_insights_overlay,
+                parent,
+                false
+            )
+
+            root = overlayView
+
+            viewName = overlayView.findViewById(R.id.view_name)
+            timelineView = overlayView.findViewById<TimelineView?>(R.id.timeline)?.apply {
+                setOnClickListener {
```
aleksandr-gringauz, line 68: question

Should we also call `onDataUpdated` in this listener to update the `TimelineView` to the most recent data?

Or we are kind of starting from scratch here?

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
```diff
@@ -0,0 +1,231 @@
...
+import com.datadog.android.rumdebugwidget.R
+
+/**
+ * Overlay displaying performance metrics collected by the [DefaultInsightsCollector].
+ * It shows as a floating widget that is attachable to any [Activity] and can be expanded to show more
+ * details.
+ *
+ * @param insightsCollector The [DefaultInsightsCollector] that provides the data to display.
+ */
+@ExperimentalRumApi
```
aleksandr-gringauz: `@ExperimentalRumApi` looks useless here, because `DefaultInsightsOverlay` is `internal class`.

We use `@ExperimentalRumApi` only for public API that is available to customers to show that this API isn't stable and can change. For our internal code we shouldn't use it.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -340,14 +340,13 @@ data class RumConfiguration internal constructor(
...
+         * @param insightsCollector the [InsightsCollector] implementation.
          * @return the [Builder] instance.
          */
+        @InternalApi
         @ExperimentalRumApi
-        fun setInsightsCollector(
-            insightsCollector: InsightsCollector?
-        ): Builder {
-            rumConfig = rumConfig.copy(insightsCollector = insightsCollector ?: NoOpInsightsCollector())
+        fun setInsightsCollector(insightsCollector: InsightsCollector): Builder {
```
aleksandr-gringauz: If this is an `@InternalApi` it should be placed in `_RumInternalProxy`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+fun RumConfiguration.Builder.enableRumDebugWidget(
+    application: Application,
+    enabled: Boolean = application.isDebuggable()
+): RumConfiguration.Builder = apply {
+    if (!enabled) return@apply
+
+    @Suppress("UnsafeThirdPartyFunctionCall")
+    runCatching {
+        val clazz = Class.forName("com.datadog.android.insights.internal.RumDebugWidget")
```
aleksandr-gringauz: What do you think of the following option to avoid reflection?

1. `setInsightsCollector` is in `_RumInternalProxy`. https://github.com/DataDog/dd-sdk-android/pull/2975/files#r2494746331. 
2. Then in this `enableRumDebugWidget` you call:
```
val insightsCollector = DefaultInsightsCollector()
_RumInternalProxy.setInsightsCollector(builder, insightsCollector)
OverlayManager(insightsCollector).start(application)
```

aleksandr-gringauz: **upd**

Sorry, forgot an important part. This file should be moved to `dd-sdk-android-rum-debug-widget`

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/OverlayManager.kt
```diff
@@ -0,0 +1,49 @@
...
+
+@ExperimentalRumApi
+internal class OverlayManager(
+    private val collector: DefaultInsightsCollector
+) : Application.ActivityLifecycleCallbacks {
+
+    private val overlays = IdentityHashMap<Activity, DefaultInsightsOverlay>()
+    private var registered = false
+
+    fun start(application: Application) {
```
aleksandr-gringauz: nit:

the contents of this `start` function can just be a part of a constructor (`init` block) of this class. This way you will not have to deal with `registered` checks.

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/DefaultInsightsCollector.kt
```diff
@@ -28,7 +28,7 @@ import java.util.concurrent.CopyOnWriteArraySet
  * @param platform Platform abstraction to access system information. Injected for test purposes.
  */
 @ExperimentalRumApi
```
aleksandr-gringauz: same as this https://github.com/DataDog/dd-sdk-android/pull/2975/files#r2494729335

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Enables the RUM Debug Widget for the given [application] if the [enabled] flag is set to `true`.
+ *
+ * @param application The application where to enable the RUM Debug Widget.
+ * @param enabled `true` to enable the RUM Debug Widget, `false` otherwise. By default, it is enabled only if the application is debuggable.
+ * @return The same [RumConfiguration.Builder] instance.
+ */
+fun RumConfiguration.Builder.enableRumDebugWidget(
+    application: Application,
+    enabled: Boolean = application.isDebuggable()
```
aleksandr-gringauz: IMHO this `enabled` parameter is useless. The method is called `enableRumDebugWidget`. If the customer calls it, they want to enable it.

maycmlee (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
kikoveiga: The idea here, and as dicussed with @satween, is that we wanted to safeguard the customer from accidentally enabling this in release builds. For this, they would have to explicitly pass `true` in the `enabled` parameter.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
kikoveiga: All this reflection inside the RUM module is because the `rum-debug-widget` module is supposed to only be included in debug builds, by using `debugImplementation` in the `build.gradle` file. Otherwise yes, we would put this extension to `RumConfiguration.Builder` in the `rum-debug-widget` module, but we want to safeguard the customer from crashing the app in production.

aleksandr-gringauz: > All this reflection inside the RUM module is because the rum-debug-widget module is supposed to only be included in debug builds, by using debugImplementation in the build.gradle file.

If the customers want to use `debugImplementation("dd-sdk-android-rum-debug-widget")`, they can create a sourceSet for debug and call `enableRumDebugWidget` from there.

satween: But in that case you will have to keep another copy of `Application` class there. Don you think it might be too complex in general for the customers so they gonna skip that option and add widget as a dependency directly in the app ending with it in a production build ?

aleksandr-gringauz: > But in that case you will have to keep another copy of Application class there.

Yes, it requires some custom setup with a separate `Application` class, but big apps in my experience often have this implemented.

> Don't you think it might be too complex in general for the customers so they gonna skip that option and add widget as a dependency directly in the app ending with it in a production build ?

If they don't want to deal with debug sourceSet for now, there is also an option for the customer to call the code from `dd-sdk-android-rum-debug-widget` inside `if (BuildConfig.DEBUG)`. In this case R8 will not include the code from `dd-sdk-android-rum-debug-widget` in the release APK (unless `dd-sdk-android-rum-debug-widget` has some services/contentproviders/etc in its manifest).

There are some more problems with the proposed approach with reflection:
1. Suppose I'm writing code in Android Studio and currently using only RUM and discover `enableRumDebugWidget` method. I decide to try it and it will not work, because I haven't included `dd-sdk-android-rum-debug-widget` module. I will have to read the documentation or source code to understand why. I just want it to work when I call the method.
2. If the customer has R8 enabled in the builds that have the widget (why not?), it will remove the code from `dd-sdk-android-rum-debug-widget` because it is currently used only through reflection. There is a way to fix it - we will have to add `consumer-rules.pro` into `dd-sdk-android-rum-debug-widget` module that says what methods to keep. But this becomes very error prone. If we add some code to the widget, we will forget to change `consumer-rules.pro` with very high probability.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
```diff
@@ -115,5 +116,13 @@ class _RumInternalProxy internal constructor(private val rumMonitor: AdvancedRum
         ): Builder {
             return builder.setRumSessionTypeOverride(rumSessionTypeOverride)
         }
+
+        @ExperimentalRumApi
```
aleksandr-gringauz: `@ExperimentalRumApi` is not needed here. Customers will not use this, they are warned [here](https://github.com/DataDog/dd-sdk-android/blob/6fc42d0e38f6d4ec2b95cad386725e2a0d2e593d/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt#L384)

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/OverlayManager.kt
```diff
@@ -0,0 +1,47 @@
...
+package com.datadog.android.insights.internal.overlay
+
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+import com.datadog.android.insights.internal.DefaultInsightsCollector
+import com.datadog.android.rum.ExperimentalRumApi
+import java.util.IdentityHashMap
+
+@ExperimentalRumApi
```
aleksandr-gringauz: `@ExperimentalRumApi` isn't needed here

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
```diff
@@ -0,0 +1,229 @@
...
+    private var threadsValue: ChartView? = null
+    private var gcValue: ChartView? = null
+    private var slowFrameRate: ChartView? = null
+    private var timelineLegend: TextView? = null
+
+    private var isPaused: Boolean = false
+    private var isTransitioning: Boolean = false
+
+    @Suppress("LongMethod")
+    internal fun attach(activity: Activity) {
```
aleksandr-gringauz, line 54: Not blocking for this PR, just some thoughts.

The current API for the widget integration allows either to enable it so that is visible everywhere in the app or not enable it at all. Also it is impossible to programmatically hide it.

Big apps might have some kind of Debug view that contains things related to debugging. For example it might contain toggles for feature flags, some debugging tools specific for the app. In this case it would be convenient to be able to enable and disable our widget by clicking on some button there. But currently it is impossible.

Also apps might want to attach the view of the widget not to the root view of activity, but somewhere else (maybe some debug panel that can be swiped to the side?).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
satween: The idea was to provide cunsumer-rules.pro yes, but I like this proposition with if (BuildConfig.DEBUG) and check if R8 could handle that with the release build 👍

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
satween: The idea behind this widget is that the customers could enable it while they work on SDK integration and then remove it completley. This is why I was thinking about adding it only as a `debugImplentation` dependency in order to prevent widget being available in release build of the app. Having an option to add this widget somewhere beside the root of activity a good idea in general, but would require much more effort to support (like the case with an app using only compose ui) and goes beside the scope of this initiative. 
We just want to provide faster feedback loop for the developers that struggle to integrate the SDK.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
satween: Well the idea was to use Build.DEBUG constant, not the application.isDebuggable(). 

Let's try to remove the `enabled` parameter and put whole logic into the `if(BuildConfig.DEBUG)` scope as suggested by @aleksandr-gringauz in the comment above

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
satween: Make sense, but need to check if it's works

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
aleksandr-gringauz: > if(BuildConfig.DEBUG)

I don't understand how this will work. Each gradle module has its own `BuildConfig` class and it is a statically generated thing. `BuildConfig.DEBUG` will be equal to `false` in the published artifacts of all modules.

What I was talking about is `if(BuildConfig.DEBUG)` where `BuildConfig` is from the app module of the customer.

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
aleksandr-gringauz: > The idea behind this widget is that the customers could enable it while they work on SDK integration and then remove it completley.

> We just want to provide faster feedback loop for the developers that struggle to integrate the SDK.

So it isn't a widget for investigating application performance (it's memory usage, skipped frames, other things) but rather a widget for debugging sdk integration?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
satween: > I don't understand how this will work. Each gradle module has its own BuildConfig class and it is a statically 
> generated thing. BuildConfig.DEBUG will be equal to false in the published artifacts of all modules.

I thought there is some R8 feature that I missed. 

Let me explain the problem we are trying to solve here: We want to let customer use `dd-android-sdk-rum-widget` module  for debug the Datadog SDK integration. We are expecting that the widget is going to be used only for the debug purposes, but in the other hand - want to make it handy and easy to use.  

From that perspective we want to avoid:
1. Accidentally adding widget into release build of the customer's app  (at the first glance it could be solved with the  `debugImplementation`)
2. Add the additional effort to manually remove usages of the widget from the release build. In case if customer will add widget as a `debugImplementation`  dependency, then use it in the code - it will break the compilation of the release build. 

So I see several options to do so: 
1. Create additional `debug` source on the customer's side - but this could be hard for some projects. Debug version of the Application has to be maintained and seems overkill for just a widget. 
2. Use reflection to check if widget is added as a dependency in the runtime. 

To reduce the impact of the reflection to the application launch time we could ask customer to provide `BuildConfig.DEBUG` value with the `enabled` parameter to not call reflection in  the production. 

WDYT ?

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
satween: Yes, widget is for:
1. Validating the integration (e.g see if the SDK identifies the user actions correctly) 
2. Debugging some issues like memory leaks or high CPU usage (but only in dev environment)

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
aleksandr-gringauz: > To reduce the impact of the reflection to the application launch time we could ask customer to provide BuildConfig.DEBUG value with the enabled parameter to not call reflection in the production.

IIUC you are proposing the following API:

```
fun RumConfiguration.Builder.enableRumDebugWidget(
    application: Application,
    enabled: Boolean
)
```

And in the documentation we tell the customers that they can pass `BuildConfig.DEBUG` as `enabled` parameter. 

Ok, this will work.

To me 2 "enable" words in one signature look confusing, I would probably call it `setRumDebugWidgetEnabled(enabled)` or if we want to call `enableRumDebugWidget` - just drop the `enabled` argument.

This is why I initially started this particular thread about `enabled` argument. 

Not blocking the PR.

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
aleksandr-gringauz: Not blocking the PR, just some ideas.

> Debugging some issues like memory leaks or high CPU usage (but only in dev environment)

It could useful to be able to do it in some builds that are intended for QA-engineers or the builds that are used internally in the customer's company. These are usually **release** builds and are often implemented as a separate Gradle flavor.

Also performance issues are often investigated in release, because when you enable compiler optimizations the picture usually changes.

> Having an option to add this widget somewhere beside the root of activity a good idea in general, but would require much more effort to support (like the case with an app using only compose ui) and goes beside the scope of this initiative.

The way it could be done is that we provide a custom android view (implemented in Android Views, not compose) as a public API. And inside it do something like `GlobalRumMonitor.getInstance(sdkName).getInsightsCollector.subscribe { data -> updateWidget(data) }` in this `CustomView.onAttachedToWindow`. `sdkName` should be provided by the customer's code either through a view's constructor or through a custom xml tag. Apps with Compose will be able to use it like [this](https://developer.android.com/develop/ui/compose/migrate/interoperability-apis/views-in-compose).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
kikoveiga: I changed the argument `enabled` to `allowInRelease` which defaults to `false`. This safeguards customers from accidentally enabling this in production, where the reflection would run with no success, as the `dd-sdk-android-rum-debug-widget` module should be added as `debugImplementation` in the `build.gradle` file. To really use the widget in release builds (e.g. local testing), customers should set `allowInRelease` to `true` and add the module as `implementation`. This is well explained in the function documentation.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
kikoveiga: As discussed with @satween, we can keep the reflection and don't need the consumer/proguard rules to clean this code. The refactoring on the [enableRumDebugWidget](https://github.com/DataDog/dd-sdk-android/blob/e131311de3df8aaf8df0df075d3191b4fa78a496/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt#L12-L30) function handles this well (as also explained in [this comment](https://github.com/DataDog/dd-sdk-android/pull/2975#discussion_r2518476032)).

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
kikoveiga: This is `false` for the first call to `attach()`, but for the next calls it could be `true` when moving the overlay between activities, so we should check to prevent reattaching, for example.

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
kikoveiga: From what I tested, there is no noticeable difference.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
aleksandr-gringauz: > and don't need the consumer/proguard rules to clean this code

`consumer-rules.pro` are needed not for cleaning, but so that the code of `dd-sdk-android-rum-debug-widget` isn't removed from the application by R8. See this example
https://github.com/DataDog/dd-sdk-android/commit/16215bab2cc684ac0f28d58a9f9f8346b998bdaa. You can run `sample:kotlin` in debug (I enabled R8 there to demonstrate) and see that `RumDebugWidget.enable` isn't called.

`RumDebugWidget` was removed from the APK of the sample app by R8.

kikoveiga: You are right, added the `consumer-rules.pro` file.

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/RumDebugWidget.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import android.app.Application
+import com.datadog.android.insights.internal.overlay.OverlayManager
+import com.datadog.android.rum.RumConfiguration
+import com.datadog.android.rum._RumInternalProxy
+
+@Suppress("OPT_IN_USAGE")
+internal object RumDebugWidget {
+
+    @JvmStatic
```
0xnm: Is this needed, given that class is `internal`, so this method can be only called from Kotlin? Or is it to simplify reflection somehow?

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/RumDebugWidget.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package com.datadog.android.insights.internal
+
+import android.app.Application
+import com.datadog.android.insights.internal.overlay.OverlayManager
+import com.datadog.android.rum.RumConfiguration
+import com.datadog.android.rum._RumInternalProxy
+
+@Suppress("OPT_IN_USAGE")
```
0xnm: is it needed?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumConfigurationBuilderTest.kt
```diff
@@ -687,10 +687,9 @@ internal class RumConfigurationBuilderTest {
     }
 
     @Test
-    fun `M use NoOpInsightsCollector W setInsightsCollector(null)`() {
+    fun `M use NoOpInsightsCollector W don't setInsightsCollector()`() {
```
0xnm, line 690:
```suggestion
    fun `M use NoOpInsightsCollector W build() { default configuration }`() {
```

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
0xnm: > Yes, widget is for:
>
> Validating the integration (e.g see if the SDK identifies the user actions correctly)

Then we have a to think what to do with `RumMonitor.debug`, because its purpose is also to debug SDK integration.

With this change we will have two "RUM debug" entrypoints.

Should we maybe unify things? WDYT?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
```diff
@@ -0,0 +1,58 @@
...
+ * To enable it in release builds (e.g., for local testing), set [allowInRelease] to `true` and
+ * add the `dd-sdk-android-rum-debug-widget` module as `implementation` (not `debugImplementation`)
+ * in your `build.gradle` file.
+ *
+ * @param application The application where to enable the RUM Debug Widget.
+ * @param allowInRelease Set to `true` to enable the widget in release builds. Defaults to `false`
+ * to prevent accidental exposure in production.
+ * @return The same [RumConfiguration.Builder] instance.
+ */
+fun RumConfiguration.Builder.enableRumDebugWidget(
```
0xnm, line 29: why it is an extension method rather than class method?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumDebugWidget.kt
0xnm: > the rum-debug-widget module is supposed to only be included in debug builds

but at the same time we have `allowInRelease` argument in this method.

Overall, it is a debug tool which _may_ be used in the release build to verify SDK integration (it makes sense if we merge the functionality of `RumMonitor.debug` here, because customers may want to check RUM view names in release builds if they are based on the class names).

I think we simply may want to take integration approach used by [LeakCanary](https://square.github.io/leakcanary/), I guess there it is closer to what @aleksandr-gringauz is proposing?

aleksandr-gringauz: > I think we simply may want to take integration approach used by [LeakCanary](https://square.github.io/leakcanary/), I guess there it is closer to what @aleksandr-gringauz is proposing?

I suppose yes, it is similar:
1. In the docs it suggests you use [debugImplementation](https://square.github.io/leakcanary/getting_started/). 
2. [Here](https://github.com/square/leakcanary/blob/22bef70ba5abdba5f65935c1a37227d01be598d3/samples/leakcanary-android-sample/src/debug/java/com/example/leakcanary/DebugExampleApplication.kt#L16-L24) there is an example of doing custom configuration in debug sourceSet.

One thing to mention is that custom configuration is optional, LeakCanary will start doing something just when you add the dependency. It does that with the help of multiple `ContentProvider`s [link](https://github.com/search?q=repo%3Asquare%2Fleakcanary%20%22ContentProvider()%22&type=code). I'm not suggesting doing the same thing in our case (we don't need it), just exploring what they are doing.

features/dd-sdk-android-rum-debug-widget/consumer-rules.pro
```diff
@@ -0,0 +1 @@
+-keep class com.datadog.android.insights.internal.RumDebugWidget
```
aleksandr-gringauz, line 1: You don't need this file anymore, it can be removed.

features/dd-sdk-android-rum-debug-widget/build.gradle.kts
```diff
@@ -39,6 +39,10 @@ plugins {
 }
 
 android {
+    defaultConfig {
+        consumerProguardFiles("consumer-rules.pro")
+    }
```
aleksandr-gringauz, line 44: This can be removed because of https://github.com/DataDog/dd-sdk-android/pull/2975/files#r2564723991

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/overlay/DefaultInsightsOverlay.kt
aleksandr-gringauz: There are 2 options:

1. If the new RUM widget also has all the features of `RumMonitor.debug`, then we could just deprecate `RumMonitor.debug` and tell the users to switch to `enableDebugRumWidget`. There is one important thing though. `RumMonitor.debug` allows you to disable the debug outline after it was previously enabled. `enableDebugRumWidget` proposed in this PR doesn't allow it.
2. Just put the new widget inside `RumMonitor.debug` as an improvement. In this case we need to put the code of the widget inside `dd-sdk-android-rum`, we will not have a separate module for rum widget.

For me currently it seems that option 2 is better.

aleksandr-gringauz (commented):
Nice work!

The only remaining issue is [this](https://github.com/DataDog/dd-sdk-android/pull/2975/files#r2564785445).

aleksandr-gringauz (approved):

PR #2976, opened 2025-10-31
Update RUM schema

aleksandr-gringauz:
### What does this PR do?

Updating RUM schema to include [this](https://github.com/DataDog/rum-events-format/pull/315) change. It split vital event into 3 different events:
1. `vital-duration-schema`
2. `vital-app-launch-schema`
3. `vital-operation-step-schema`

They don't have the same parent anymore. They are distinct event types.

There was also some refactoring arount `view` field [link](https://github.com/DataDog/rum-events-format/pull/315/files#diff-8c100040e5a58806d0081a5961e3884fa816e322abb91b2d97b7c4ac204b8de4).

The changes are so huge because a lot of classes classes in the `api-surface` files are now in a different order for some reason.

Similar PR on iOS https://github.com/DataDog/dd-sdk-ios/pull/2548

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumConfigurationBuilderTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumConfigurationBuilderTest.kt
@@ -449,6 +449,7 @@ internal class RumConfigurationBuilderTest {
         )
     }
 
+    @OptIn(ExperimentalRumApi::class)
     @Test
     fun `M build config with RUM Vital eventMapper W setVitalOperationStepEventMapper() & build()`() {
         // Given


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -220,13 +220,13 @@ data class RumConfiguration internal constructor(
...
-         * Sets the [EventMapper] for the RUM [VitalEvent]. You can use this interface implementation
-         * to modify the [VitalEvent] attributes before serialisation.
+         * Sets the [EventMapper] for the RUM [RumVitalOperationStepEvent]. You can use this interface implementation
+         * to modify the [RumVitalOperationStepEvent] attributes before serialisation.
          *
          * @param eventMapper the [EventMapper] implementation.
          */
-        fun setVitalEventMapper(eventMapper: EventMapper<VitalEvent>): Builder {
-            rumConfig = rumConfig.copy(vitalEventMapper = eventMapper)
+        fun setVitalOperationStepEventMapper(eventMapper: EventMapper<RumVitalOperationStepEvent>): Builder {
```
aleksandr-gringauz, line 241: **Important**

This is a change of the public API that was already released. 

Even if we don't change the name of the method, there is still a change in the function signature (because of the argument type).

It is probably ok, because feature operations are fairly recent are still under `ExperimentalRumAPI` (not this particular method though).

But I'm not entirely sure.

0xnm: Is it also a similar change for iOS SDK?

Maybe we can keep both methods for now: deprecate old one and make it alias for a new one. Old one will have a change of the argument type, but this is fine, because anyway it is functional interface (so no need to specify type explicitly) and new model has almost the same structure as the old one.

aleksandr-gringauz: https://github.com/DataDog/dd-sdk-ios/blob/7677cc0335f0993ffdec911a73c20546a44e014e/DatadogRUM/Sources/RUM%2Bobjc.swift#L434

Looks like they don't have a mapper for the vital events at all.

aleksandr-gringauz: > but this is fine, because anyway it is functional interface (so no need to specify type explicitly)

```
import com.datadog.android.event.EventMapper
import com.datadog.android.rum.model.RumVitalOperationStepEvent

object SomeMapper : EventMapper<RumVitalOperationStepEvent> {
    override fun map(event: RumVitalOperationStepEvent): RumVitalOperationStepEvent? {
        TODO("Not yet implemented")
    }
}

```

If you want to implement the mapper in a separate file, you have to. And IIUC binary compatibility is still broken.

0xnm: yes, totally. once it is not a functional SAM conversion change is not backward compatible.

0xnm: > It is probably ok, because feature operations are fairly recent are still under ExperimentalRumAPI (not this particular method though).

Feature Operations are indeed marked as experimental, but `setVitalEventMapper` is not.

I think we can deprecate it and keep it, making it either no-op (since once we add more Vital mappers it is not clear to which one it corresponds) or forward call to `setVitalOperationStepEventMapper` (since it is the only Vital we have right now).

This will be more compliant with semver semantics.

ambushwork (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
aleksandr-gringauz: OK

1. Returned `setVitalEventMapper` and made it `Deprecated`.
2. Made `setVitalEventMapper` call `setVitalOperationStepEventMapper`.
3. Annotated `setVitalOperationStepEventMapper` with `ExperimentalRumApi`. To be similar with other feature operations [methods](https://github.com/DataDog/dd-sdk-android/blob/fa9de09ca51c189b1929219608c7b45f54db7321/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt#L339).

0xnm (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
satween: The `setVitalEventMapper` wasn't planned as experimental as there was different types of `VitalEvents` and `@ExperimentalRumApi` was applied only for the Feature Operation. But  according to the latest changes it might make sense to add this annotation for the setting mapper method as well.

0xnm (approved):

PR #2987, opened 2025-11-05
RUM-13103: Introduce vitalAppLaunchEventMapper

aleksandr-gringauz:
### What does this PR do?

Adding `vitalAppLaunchEventMapper` similar to `vitalOperationStepEventMapper` now that app launch vital events have a separate type.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/api/apiSurface
+++ b/features/dd-sdk-android-rum/api/apiSurface
@@ -75,8 +75,7 @@ data class com.datadog.android.rum.RumConfiguration
     fun setErrorEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.ErrorEvent>): Builder
     fun setLongTaskEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.LongTaskEvent>): Builder
     DEPRECATED fun setVitalEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.RumVitalOperationStepEvent>): Builder
-    fun setVitalOperationStepEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.RumVitalOperationStepEvent>): Builder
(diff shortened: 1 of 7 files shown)


features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -76,6 +76,7 @@ data class com.datadog.android.rum.RumConfiguration
     fun setLongTaskEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.LongTaskEvent>): Builder
     DEPRECATED fun setVitalEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.RumVitalOperationStepEvent>): Builder
     fun setVitalOperationStepEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.RumVitalOperationStepEvent>): Builder
+    fun setVitalAppLaunchEventMapper(com.datadog.android.event.EventMapper<com.datadog.android.rum.model.RumVitalAppLaunchEvent>): Builder
```
0xnm: I'm wondering if we want to expose them in such shape. Maybe we can group them and introduce a single method here `withVitalEventMappers(operationsMapper = ..., appLaunchMapper = ..., ....)`.

With current layout these methods seems to be very verbose among the other methods in this `Builder`.

Maybe we need to ask iOS SDK what will be their approach regarding this as well.

satween: Or maybe we should revert to `setVitaEventMapper ` and use them for the all the `vital` events 🙂

aleksandr-gringauz: @satween What arguments should `setVitaEventMapper` have? Same as @0xnm suggested?

I was thinking about something like this: https://pl.kotl.in/nNeN0bmfx

Note that currently `VitalEvent` from this example doesn't exist in our code. `OperationStepVital` and `AppLaunchVital` don't have a common parent type. We could change it though.

But I still lean towards the solution with a separate method per vital type.

We started discussing it with @simaoseica-dd, but not finished yet.

PR #2990, opened 2025-11-06
RUM-10417: Handle potential StackOverflowError

kikoveiga:
### What does this PR do?

Wraps the `onRequestIntercepted` method in `TracingInterceptor.kt` inside a `try - catch` block that logs an error in case a `StackOverflowError` is caught.

### Motivation

If a customer implements a custom `TracedRequestListener` and keeps retrying a request with `onRequestIntercepted`, this will run forever, eventually leading to a `StackOverflowError`.

### DIFF
--- a/integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTest.kt
+++ b/integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTest.kt
@@ -1660,7 +1660,7 @@ internal open class TracingInterceptorTest {
     }
 
     @Test
-    fun `M log error W listener causes StackOverflowError through infinite recursion`(
+    fun `M log error and rethrow W listener causes StackOverflowError`(
         @IntForgery(min = 200, max = 600) statusCode: Int
     ) {
         // Given
@@ -1669,13 +1669,12 @@ internal open class TracingInterceptorTest {
 
         whenever(
             mockRequestListener.onRequestIntercepted(any(), any(), anyOrNull(), anyOrNull())
-        ).doAnswer {
-            testedInterceptor.intercept(mockChain)
-            return@doAnswer Unit
-        }
+        ).doAnswer { throw StackOverflowError() }
 
         // When
-        testedInterceptor.intercept(mockChain)
+        assertThrows<StackOverflowError> {
+            testedInterceptor.intercept(mockChain)
+        }
 
         // Then
         val expectedMessage = "${TracingInterceptor.ERROR_STACK_OVERFLOW}\n" +
@@ -1685,7 +1684,8 @@ internal open class TracingInterceptorTest {
             InternalLogger.Level.ERROR,
             InternalLogger.Target.USER,
             expectedMessage,
-            StackOverflowError::class.java
+            StackOverflowError::class.java,
+            mode = org.mockito.kotlin.atLeast(1)
         )
     }
 


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -168,11 +172,35 @@ internal constructor(
...
         throwable: Throwable?
     ) {
-        if (span != null) {
+        if (span != null && canNotifyListener(sdkCore, request)) {
             tracedRequestListener.onRequestIntercepted(request, span, response, throwable)
         }
     }
 
+    private fun canNotifyListener(sdkCore: FeatureSdkCore, request: Request): Boolean {
+        synchronized(requestInterceptionLock) {
```
satween: Why we need this synchronization?

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTest.kt
```diff
@@ -1659,6 +1659,117 @@ internal open class TracingInterceptorTest {
...
+        // When
+        testedInterceptor.intercept(mockChain)
+
+        // Then
+        verify(mockRequestListener, times(maxRetries)).onRequestIntercepted(any(), any(), anyOrNull(), anyOrNull())
+
+        verify(mockInternalLogger).log(
+            eq(InternalLogger.Level.WARN),
+            eq(InternalLogger.Target.USER),
+            any(),
```
satween: we need to check that the exact message is being printed. Look for `verifyLog` method across other tests

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTest.kt
```diff
@@ -1659,6 +1659,117 @@ internal open class TracingInterceptorTest {
...
+        val maxRetries = TracingInterceptor.MAX_REQUEST_INTERCEPTION_COUNT
+
+        whenever(mockResolver.isFirstPartyUrl(fakeUrl.toHttpUrl())).thenReturn(true)
+        stubChain(mockChain, statusCode)
+
+        whenever(
+            mockRequestListener.onRequestIntercepted(any(), any(), anyOrNull(), anyOrNull())
+        ).doAnswer {
+            interceptCallCount++
+            if (interceptCallCount <= maxRetries + 1) {
```
satween: this test looks wrong to me. This logic should be inside code, not in the test

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorTest.kt
```diff
@@ -1659,6 +1659,117 @@ internal open class TracingInterceptorTest {
...
+            eq(InternalLogger.Target.USER),
+            any(),
+            eq(null),
+            eq(false),
+            eq(null)
+        )
+    }
+
+    @Test
+    fun `M reset counter W listener succeeds without retry`(
```
satween: How this test verifies that counter is being reset after a success call? I see that this test basically verifies that the `onRequestIntercepted` is being called same amount of times as the `intercept`.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -168,11 +172,35 @@ internal constructor(
...
+            val currentCount = requestInterceptionCounts[request] ?: 0
+
+            if (currentCount >= MAX_REQUEST_INTERCEPTION_COUNT) {
+                val requestKey = "${request.method}:${request.url}"
```
satween: It's seems like you creating a key only for log message. This implementations uses `Request` object as a key. How you can be sure that two different `Request` objects in memory, targeting same url will end up with the same `hashCode`?

satween (changes requested):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -169,7 +169,18 @@ internal constructor(
         throwable: Throwable?
     ) {
         if (span != null) {
-            tracedRequestListener.onRequestIntercepted(request, span, response, throwable)
+            try {
+                tracedRequestListener.onRequestIntercepted(request, span, response, throwable)
+            } catch (e: StackOverflowError) {
```
0xnm, line 174: While this prevents crashing host app, it still allows having some loops before hitting stack overflow.

Maybe we can even prevent making loops by adding a tag to the `Request` with a `{Interceptor}.toString()` or `System.identityHashCode` (it should be a `Set` then since the same request may go through multiple interceptors) and reading it back to check if we are in the loop. WDYT?

0xnm: One more note: normally `Error` classes are non-recoverable JVM errors, it is not advised to continue execution when such is thrown.

https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.

satween: Not sure we want to make such changes. This PR is mostly to help customer to identify a problem. We want to explain why the stackoverflow  error is thrown, not preventing recursive calls. Some customers might like to have them. But I agree that the error should not be swallowed by `catch` block and should be re-thrown.

kikoveiga: I have added the re-throwing of the error.

0xnm: We can indeed re-throw it. My point is that after getting such error it is not possible to continue normal execution, because stack _may_ not recover to the state (free capacity) enough to continue reliably. Even if may continue here, if stack is not unrolled, similar error may pop up in random place further.

0xnm (dismissed):

satween (approved):

PR #3000, opened 2025-11-11
fix: Add internal disableJankStats config option

awforsythe:
refs: RUM-12763

### What does this PR do?

This PR adds a check to RUM feature initialization: if the new internal option `_dd.rum.disable_jank_stats` is explicitly set to true, then we refrain from calling `initializeFrameStatesAggregator()`. Otherwise, we call that function as normal.

### Motivation

This change allows the Unity SDK to configure the Android SDK in such a way that it will never call `JankStats.createAndTrack()`, which is necessary to avoid crashes caused by the older versions of `androidx.metrics` that the `dd-sdk-unity` is forced to use in older versions of Unity.

### Additional Notes

I notice that the generated `apiSurface` files are not up to date at latest in develop, so I expect that CI will fail. When I test locally with `./local-ci.sh --test`, everything passes. I can rebase once the unrelated API surface issue is fixed in develop.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
@@ -423,5 +423,14 @@ data class RumConfiguration internal constructor(
             rumConfig = rumConfig.copy(rumSessionTypeOverride = rumSessionTypeOverride)
             return this
         }
+
+        /**
+         * Disables JankStats tracking. This flag may be enabled by cross-platform SDKs where
+         * native frame metrics are not meaningful.
+         */
+        internal fun setDisableJankStats(disable: Boolean): Builder {
+            rumConfig = rumConfig.copy(disableJankStats = disable)
+            return this
+        }
     }
 }
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
@@ -239,13 +239,20 @@ internal class RumFeature(
             initializeVitalExecutorService(frequency)
             initializeCpuVitalMonitor(frequency)
             initializeMemoryVitalMonitor(frequency)
-            initializeFrameStatesAggregator(
-                application = appContext as? Application,
-                listeners = listOfNotNull(
-                    initializeSlowFrameListener(slowFrameListenerConfiguration),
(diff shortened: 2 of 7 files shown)


satween:
Looks good to me, but we need to port it to into release branch of v2

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -239,13 +239,20 @@ internal class RumFeature(
...
-            initializeFrameStatesAggregator(
-                application = appContext as? Application,
-                listeners = listOfNotNull(
-                    initializeSlowFrameListener(slowFrameListenerConfiguration),
-                    initializeFPSVitalMonitor(frequency)
+
+            // If "_dd.rum.disable_jank_stats" is explicitly set to true, refrain from registering
+            // any frame state listeners; if false or not set, proceed normally
+            val disableJankStats = configuration.additionalConfig[DD_RUM_DISABLE_JANK_STATS_TAG]
+            val allowJankStats = disableJankStats as? Boolean != true
```
0xnm: `RumConfiguration.additionalConfig` is something legacy I would say.

I would propose to add the explicit field to `RumConfiguration` and expose it through `internal` method which can be called via `_RumInternalProxy`.

awforsythe: Gotcha, that makes sense. I've switched to the approach you suggest.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -423,5 +423,14 @@ data class RumConfiguration internal constructor(
             rumConfig = rumConfig.copy(rumSessionTypeOverride = rumSessionTypeOverride)
             return this
         }
+
+        /**
+         * Disables JankStats tracking. This flag may be enabled by cross-platform SDKs where
+         * native frame metrics are not meaningful.
+         */
+        internal fun setDisableJankStats(disable: Boolean): Builder {
```
0xnm, line 431: minor: it can be just `disableJankStats`, but let's keep it like that, it is only for us anyway.

0xnm (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
0xnm: btw, probably PR title should be updated, since approach is different now

PR #3003, opened 2025-11-13
Fix RumWindowCallbacksRegistry for multiple SDK instances

aleksandr-gringauz:
### What does this PR do?

Fixing `RumWindowCallbacksRegistry` for the case when there are multiple Datadog SDK instances in the app.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/window/RumWindowCallbacksRegistry.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/window/RumWindowCallbacksRegistry.kt
@@ -44,26 +44,28 @@ internal class RumWindowCallbacksRegistryImpl : RumWindowCallbacksRegistry {
             }
         }
     }
-}
 
-private fun Window.wrapCallback(): RumWindowCallback {
-    val currentCallback = callback
-    val newCallback = RumWindowCallback(
-        wrapped = currentCallback
-    )
-    callback = newCallback
-    return newCallback
-}
+    private fun Window.wrapCallback(): RumWindowCallback {
+        val currentCallback = callback
+        val newCallback = RumWindowCallback(
+            wrapped = currentCallback,
+            registry = this@RumWindowCallbacksRegistryImpl
+        )
+        callback = newCallback
+        return newCallback
+    }
 
-private fun Window.tryToRemoveCallback() {
-    val currentCallback = callback
-    if (currentCallback is RumWindowCallback) {
-        callback = currentCallback.wrapped
+    private fun Window.tryToRemoveCallback() {
+        val currentCallback = callback
+        if (currentCallback is RumWindowCallback && currentCallback.registry === this@RumWindowCallbacksRegistryImpl) {
+            callback = currentCallback.wrapped
+        }
     }
 }
 
 private class RumWindowCallback(
-    val wrapped: Window.Callback
+    val wrapped: Window.Callback,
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/window/RumWindowCallbacksRegistry.kt
```diff
@@ -44,26 +44,28 @@ internal class RumWindowCallbacksRegistryImpl : RumWindowCallbacksRegistry {
...
+            wrapped = currentCallback,
+            registry = this@RumWindowCallbacksRegistryImpl
+        )
+        callback = newCallback
+        return newCallback
+    }
 
-private fun Window.tryToRemoveCallback() {
-    val currentCallback = callback
-    if (currentCallback is RumWindowCallback) {
```
aleksandr-gringauz, line 60: The problem is in this `if (currentCallback is RumWindowCallback)` check. Suppose we have 2 SDK instance and both installed their `RumWindowCallback`s, wrapping the existing ones.

When removing the callback we need to check that we are removing "our" `RumWindowCallback`, but not the `RumWindowCallback` of the `RumWindowCallbacksRegistry` from a different SDK instance.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/window/RumWindowCallbacksRegistry.kt
```diff
@@ -44,26 +44,28 @@ internal class RumWindowCallbacksRegistryImpl : RumWindowCallbacksRegistry {
...
-        callback = currentCallback.wrapped
+    private fun Window.tryToRemoveCallback() {
+        val currentCallback = callback
+        if (currentCallback is RumWindowCallback && currentCallback.registry === this@RumWindowCallbacksRegistryImpl) {
```
0xnm: Probably alternative here is to check if `RumWindowCallbacksRegistryImpl.callbacks.values()` contains this callback, but we avoid doing this, because `RumWindowCallbacksRegistryImpl.callbacks` is a `WeakHashMap`?

I'm okay with solution if it don't impose leaks due to the circular dependency.

0xnm (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/window/RumWindowCallbacksRegistry.kt
aleksandr-gringauz: > Probably alternative here is to check if RumWindowCallbacksRegistryImpl.callbacks.values() contains this callback, but we avoid doing this, because RumWindowCallbacksRegistryImpl.callbacks is a WeakHashMap?

This solution will also work, I just didn't think about it. It is actually a bit cleaner I think. Fixed.

> I'm okay with solution if it don't impose leaks due to the circular dependency.

Kotlin GC can deal with circular references. Unless this "circle" goes through some C++ code. But this isn't our case I think. Or are you talking about something else?

0xnm: It was just a raw assumption without going into the details.

Actually, we didn't check our SDK for leaks for a while, maybe we need to do a run of sample app with LeakCanary.

0xnm (approved):

PR #3005, opened 2025-11-14
RUM-10805: Remove resolveMeta & resolveMetrics methods in tests

kikoveiga:
### What does this PR do?

- Refactors trace integration tests to eliminate circular dependency to generate expected values.
- Improves the tests.

### Motivation

`resolveMeta` and `resolveMetrics` methods used production code to generate expected values, then compared those values against the actual output.


instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/trace/TracesTest.kt
```diff
@@ -140,11 +136,48 @@ internal abstract class TracesTest {
...
+            .hasField(VERSION_KEY, context.version)
+            .hasField(VARIANT_KEY, context.variant)
+
+        assertThat(metaObject).hasField(DD_KEY) {
+            hasField(SOURCE_KEY, context.source)
+        }
+
+        assertThat(metaObject).hasField(SPAN_KEY) {
+            hasField(SPAN_KIND_KEY, DEFAULT_SPAN_KIND)
+        }
```
0xnm: you can actually continue using the fluid syntax if you want:

```suggestion
            .hasField(VERSION_KEY, context.version)
            .hasField(VARIANT_KEY, context.variant)
            .hasField(DD_KEY) {
              hasField(SOURCE_KEY, context.source)
            }
            .hasField(SPAN_KEY) {
              hasField(SPAN_KIND_KEY, DEFAULT_SPAN_KIND)
            }
```

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/trace/TracesTest.kt
```diff
@@ -160,17 +193,73 @@ internal abstract class TracesTest {
...
+
+        // endregion
+
+        // region SpanEvent.Dd
+        const val SOURCE_KEY = "source"
+
+        // endregion
+
+        // region SpanEvent.Span
+        const val SPAN_KIND_KEY = "kind"
```
0xnm, line 231: nit: for a symmetry with `endregion`, it is better to add a newline

```suggestion
        // region SpanEvent.Span
        
        const val SPAN_KIND_KEY = "kind"
```

0xnm (dismissed):

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/trace/TracesTest.kt
kikoveiga: Fixed.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/trace/TracesTest.kt
kikoveiga: Fixed.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/trace/TracesTest.kt
```diff
@@ -140,11 +136,42 @@ internal abstract class TracesTest {
...
+            .hasField(DD_KEY) {
+                hasField(SOURCE_KEY, context.source)
+            }
+            .hasField(SPAN_KEY) {
+                hasField(SPAN_KIND_KEY, DEFAULT_SPAN_KIND)
+            }
+            .hasField(TRACER_KEY) {
+                hasField(TRACER_VERSION_KEY, context.sdkVersion)
+            }
+            .hasField(USR_KEY) {}
```
0xnm: I believe there should be something here?

kikoveiga: All the fields of `SpanEvent.Usr` are nullable and they are not used or set for these tests.

0xnm: Then this is something we shouldn't have. We should either set User object for the test or not add empty `UserInfo` object (like [we do](https://github.com/DataDog/dd-sdk-android/blob/499916cfba381a40f1ae8db7f4f0bc1e0bfca941/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L377) in RUM) to the trace event (or both), because sending object with empty property doesn't really make any sense.

Also I see that `UserInfo.anonymousId` is not set for the Span event, it is worth checking if it is on purpose or it is a miss.

0xnm (approved):

PR #3010, opened 2025-11-18
RUM-12891: Fix RUM resource duration breakdown

ambushwork:
### What does this PR do?

This PRs fixes the resource duration breakdown feature which is broken since the version 2.12.0, more precisely from this [commit](https://github.com/DataDog/dd-sdk-android/pull/2126).

#### The main cause of issue:

We have `DatadogInteceptor` generating the RUM event `startResource` with the `ResourceId(requestUrl, randomUUID)` as the key of the resource when intercepting a request, also we have `DatadogEventListener` to add `waitForResourceTiming` event and `addResourceTiming` to report the resource timing for this duration breakdown.

The issue is that, both `DatadogInteceptor` and `DatadogEventListener` are generating their own UUID for the same request, so in our event processing they won’t be able to match.

The fix is to remove the generation of UUID in `DatadogEventListener` since its key is only used for enhancing the current resource.

### Motivation

RUM-12891

### Demo

| Before | image  |
| --- | --- |
|After| image|

### DIFF
--- a/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogEventListener.kt
+++ b/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogEventListener.kt
@@ -9,7 +9,6 @@ package com.datadog.android.okhttp
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
 import com.datadog.android.core.SdkReference
-import com.datadog.android.okhttp.DatadogEventListener.Factory
 import com.datadog.android.okhttp.internal.rum.buildResourceId
 import com.datadog.android.rum.GlobalRumMonitor
 import com.datadog.android.rum.internal.domain.event.ResourceTiming
@@ -244,7 +243,7 @@ internal constructor(
 
         /** @inheritdoc */
         override fun create(call: Call): EventListener {
-            val resourceId = call.request().buildResourceId(generateUuid = true)
+            val resourceId = call.request().buildResourceId(generateUuid = false)
             val sdkCore = sdkCoreReference.get()
             return if (sdkCore != null) {
                 DatadogEventListener(sdkCore, resourceId)
--- a/integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/DatadogEventListenerFactoryTest.kt
+++ b/integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/DatadogEventListenerFactoryTest.kt
@@ -81,6 +81,16 @@ internal class DatadogEventListenerFactoryTest {
         assertThat(result).isSameAs(DatadogEventListener.Factory.NO_OP_EVENT_LISTENER)
     }
 
+    @Test
+    fun `M not generate UUID in the resource key W create()`() {
+        // When
+        val result = testedFactory.create(mockCall)
+
+        // Then
+        check(result is DatadogEventListener)
+        assertThat(result.key.uuid).isEqualTo(null)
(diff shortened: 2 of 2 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogEventListener.kt
```diff
@@ -244,7 +243,7 @@ internal constructor(
 
         /** @inheritdoc */
         override fun create(call: Call): EventListener {
-            val resourceId = call.request().buildResourceId(generateUuid = true)
+            val resourceId = call.request().buildResourceId(generateUuid = false)
```
0xnm, line 246: I don't see how this fixes the issue given [that under the hood](https://github.com/DataDog/dd-sdk-android/blob/494081b2b57b1345626201a3906999bc16dd08fa/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/rum/RequestExt.kt#L15) it will first try to pull existing UUID from the `Request.tag`, which should already be there if call to interceptor was made before.

Can you please give more details?

ambushwork: that's because the UUID we generate from `DatadogIntercptor` is not put inside `Request.tag`, so here if we try to retrieve `tag(UUID::class.java)` it will always return null. The UUID is only used as a RUM resource key.

0xnm: > that's because the UUID we generate from DatadogInterceptor is not put inside Request.tag

This is probably a real cause of the issue, it should be there and it is a miss from our side.

Basically we always should rely on the UUID, because the legacy way to identify the request is not reliable: say there are multiple `GET` calls in parallel with the same URL - they will be identified as the same call. Same for multiple `POST` calls with the same body length.

So when we generate UUID we should put it to the request as tag as well.

And we should add the tests to the `reliability/single-fit-okhttp` that for a call made with instrumentation setup `startResource` / `stopResource`/`waitForResourceTiming`/`addResourceTiming` all have the same `key` argument.

ambushwork: OK, I see the issue, then this is a bit tricky to handle with our current implementation, because `create` in `DatadogEventListener` will be called first, the `call` is immutable inside, meaning that we are not able to pass our resource id along with the request to `DatadogInterceptor`

0xnm: At the same time we have an access to the `Call` object in all the callbacks of `EventListener`, meaning we can read the actual state of `request` to get the ID in each callback instead of passing it at the `DatadogEventListener` constructor, probably it will still get consistent results for the same call.

I'm not sure at which point `callStart` is called, but if it is after `intercept`, then we should be good.

Overall, we shouldn't indeed mutate anything here as per [the following comment](https://github.com/square/okhttp/blob/4984568367caaf359b82c452bd28b5e192824d1c/okhttp/src/main/kotlin/okhttp3/EventListener.kt#L467-L468).

ambushwork: `callStart` also happens before `intercept`, and other functions are not systematically called after `intercept` neither, we will need to dig another way out.

0xnm (approved):

satween (approved):

PR #3011, opened 2025-11-18
RUM-10363: Replace System static calls with TimeProvider

kikoveiga:
### What does this PR do?

- Replaces `System.currentTimeMillis()` and `System.nanoTime()` calls with `TimeProvider` interface calls.
- Improves and optimizes some tests by replacing `Thread.sleep` calls with fake timestamps using a `mockTimeProvider`.

### Motivation

- Static calls are not mockable. Using an interface is better for injecting mock implementations and general testability.

### Additional Notes

These are the results of running debug unit tests locally on develop and current branch, where there are noticeable time improvements.

image


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
```diff
@@ -119,14 +134,15 @@ internal class RotatingDnsResolverTest {
...
+
+        val fakeExpiredTime = fakeElapsedTimeNs + TEST_TTL_MS.inWholeNanoseconds
+        whenever(mockTimeProvider.getDeviceElapsedTimeNs()).doReturn(fakeElapsedTimeNs, fakeExpiredTime)
+
         var exceptionThrown: Exception? = null
 
         // When
         List(100) {
             Thread {
-                Thread.sleep(forge.aLong(min = 0, max = 100))
```
aleksandr-gringauz, line 129: Are you sure we need to remove this sleep? It seems it is actually a part of the test, it is intended that different threads sleep for different amount of time.

aleksandr-gringauz (dismissed):
Nice work!

tools/benchmark/src/main/java/com/datadog/benchmark/DatadogVitalsMeter.kt
```diff
@@ -35,7 +37,9 @@ class DatadogVitalsMeter private constructor(private val meter: Meter) : Datadog
 
     private val cpuVitalReader: CPUVitalReader = CPUVitalReader()
     private val memoryVitalReader: MemoryVitalReader = MemoryVitalReader()
-    private val fpsVitalReader: FpsVitalReader = FpsVitalReader()
+    private val fpsVitalReader: FpsVitalReader = FpsVitalReader(
+        (Datadog.getInstance() as FeatureSdkCore).timeProvider
```
aleksandr-gringauz: It is better to provide `SdkCore` (or better `TimeProvider`) through constructor. Our SDK should support the case where there are multiple instances of the core currently running.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
```diff
@@ -23,6 +26,14 @@ interface TimeProvider {
      */
     fun getServerTimestamp(): Long
 
+    /**
+     * Returns the current device elapsed time in nanoseconds.
+     *
+     * Default implementation returns [System.nanoTime].
+     * This should not be overridden as device elapsed time always refers to the monotonic system clock.
+     */
+    fun getDeviceElapsedTimeNs(): Long = System.nanoTime()
```
0xnm: Probably we shouldn't call it `elapsed` time, because it can be a confusion with `elapsedRealTime`. Basically, documentation of `System.nanoTime` says:

> The value returned by this method does not account for elapsed time during deep sleep. For timekeeping facilities available on Android see android.os.SystemClock.

And then there is `SystemClock.elapsedRealtime`, which is different. And on top of that there is `Process.getStartElapsedRealtime`, etc.

With such name we can have misuse due to misunderstanding. Probably we need to add more detailed explanation to all methods actually, referencing the system methods they are backing.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
```diff
@@ -23,6 +26,14 @@ interface TimeProvider {
      */
     fun getServerTimestamp(): Long
 
+    /**
+     * Returns the current device elapsed time in nanoseconds.
+     *
+     * Default implementation returns [System.nanoTime].
+     * This should not be overridden as device elapsed time always refers to the monotonic system clock.
```
0xnm: but nothing prevents overriding it though

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProvider.kt
```diff
@@ -11,10 +11,13 @@ import android.os.Build
 import android.os.Process
 import android.os.SystemClock
 import com.datadog.android.core.internal.system.BuildSdkVersionProvider
+import com.datadog.android.internal.time.DefaultTimeProvider
+import com.datadog.android.internal.time.TimeProvider
 import com.datadog.android.rum.DdRumContentProvider
 import java.util.concurrent.TimeUnit
 
 internal class DefaultAppStartTimeProvider(
+    private val timeProvider: TimeProvider = DefaultTimeProvider(),
```
0xnm: I'm wondering if we can use the value from `CoreFeature`. Yes, this class is passed to the `CoreFeature` ctor, but we can have a lambda factory here maybe, which is called when things are really needed?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/MethodCalledTelemetryTest.kt
```diff
@@ -105,13 +114,16 @@ internal class MethodCalledTelemetryTest {
         whenever(mockDeviceInfo.osVersion).thenReturn(fakeOsVersion)
         whenever(mockDeviceInfo.deviceBuildId).thenReturn(fakeOsBuild)
 
-        fakeStartTime = System.nanoTime()
+        whenever(mockTimeProvider.getDeviceElapsedTimeNs())
+            .thenReturn(fakeStartTimeNs)
+            .thenReturn(fakeStartTimeNs + fakeElapsedTimeNs)
```
0xnm: I think it is better to create a stub actually which we can reuse everywhere. It would expose methods of `TimeProvider` and then also have additional methods to advance the clock.

This way it will be more granular control and better locality to the particular test.

WDYT?

0xnm (changes requested):
I very briefly went through some files.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
kikoveiga: I was thinking of adding them as extension functions so they aren't overridable, WDYT?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/MethodCalledTelemetryTest.kt
kikoveiga: I did create a `StubTimeProvider`, and use both depending on the situation. Sometimes it's enough to use the `mock` with a few `thenReturn` or `doReturn` statements, other times I use the `stub` for more granular and complex operations. I will double-check their use on this file and others.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
0xnm: can be the option as well I think

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/metrics/MethodCalledTelemetryTest.kt
0xnm: Then I didn't come to the Stub part then :)

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
kikoveiga: As dicussed, kept the name and improved the documentation.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
kikoveiga: You're right, added it back.

tools/benchmark/src/main/java/com/datadog/benchmark/DatadogVitalsMeter.kt
kikoveiga: Fixed.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
kikoveiga: Added a `BaseTimeProvider` abstract class to override these methods as `final`. Kept the interface as is to be able to override everything in `StubTimeProvider` for better testability.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProvider.kt
kikoveiga: It works :)

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
```diff
@@ -72,12 +82,13 @@ internal class RotatingDnsResolverTest {
         val result = mutableListOf<InetAddress>()
 
         // When
-        fakeInetAddresses.forEach {
+        fakeInetAddresses.forEach { _ ->
```
0xnm: redundant change

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
```diff
@@ -119,8 +132,7 @@ internal class RotatingDnsResolverTest {
         // the real use case where we have a small number of addresses to rotate
         fakeInetAddresses = forge.aList(size = forge.anInt(min = 1, max = 3)) { mock() }
         whenever(mockDelegate.lookup(fakeHostname)) doReturn fakeInetAddresses
-        // just wait the TTL time to make sure all threads are concurrently accessing the lookup
-        Thread.sleep(TEST_TTL_MS.inWholeMilliseconds)
```
0xnm, line 123: is this a right change?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/advanced/MoveDataMigrationOperationTest.kt
```diff
@@ -132,7 +150,7 @@ internal class MoveDataMigrationOperationTest {
     }
 
     @Test
-    fun `M retry with 500ms delay W run() {move fails once}`() {
+    fun `M not wait for real delay W run() {move fails once, time provider mocked}`() {
```
0xnm: I don't think we need to add `time provider mocked`, it is just a test setup details. Also the `retry with 500ms` should be put back.

What we are testing is that _there is a delay_ between retries, so that we don't overload I/O in a loop.

This means that we have to really imitate the real delay. I don't think you can avoid using `Thread.sleep` here, otherwise you test will be green even if the behavior of `MoveDataMigrationOperation` is broken and there is no expected delay anymore.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/advanced/MoveDataMigrationOperationTest.kt
```diff
@@ -161,7 +179,7 @@ internal class MoveDataMigrationOperationTest {
     }
 
     @Test
-    fun `M retry with 500ms delay W run() {move always fails}`() {
+    fun `M not wait for real delay W run() {move always fails, time provider mocked}`() {
```
0xnm: same as above

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/advanced/WipeDataMigrationOperationTest.kt
0xnm, line 1: same comment as above.

Maybe it is solvable by advancing the clock by say 400 ms, in this case you would expect 3 calls to the clock and 2 calls to move. But still, here it is probably better to use `Thread.sleep`.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/thread/ThreadPoolExecutorExtTest.kt
```diff
@@ -50,17 +61,21 @@ internal class ThreadPoolExecutorExtTest {
...
+        val isIdled = testedMockExecutor.waitToIdle(fakeTimeout, mockInternalLogger, mockTimeProvider)
         testedMockExecutor.isIdle()
 
         // THEN
         assertThat(isIdled).isFalse()
     }
 
     @Test
-    fun `M wait max timeout milliseconds W waitToIdle { executor not idled }`(
+    fun `M exit loop after timeout W waitToIdle { executor not idled }`(
```
0xnm: How can we be sure now that wait actually happened and `waitToIdle` didn't exit instantly?

Pretty much the same comment for the tests below.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/utils/MiscUtilsTest.kt
```diff
@@ -62,62 +62,52 @@ internal class MiscUtilsTest {
...
-        val wasSuccessful = retryWithDelay(mockedBlock, fakeTimes, fakeDelay, mockInternalLogger)
+        val wasSuccessful = retryWithDelay(mockedBlock, fakeTimes, fakeDelayNs, mockInternalLogger, mockTimeProvider)
 
         // Then
         assertThat(wasSuccessful).isFalse()
         verify(mockedBlock, times(fakeTimes)).invoke()
     }
 
-    @Test
-    fun `M execute the block in a delayed loop W retryWithDelay`(forge: Forge) {
```
0xnm, line 82: I think this shouldn't be removed.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/BatchClosedMetadataForgeryFactory.kt
```diff
@@ -14,7 +14,7 @@ internal class BatchClosedMetadataForgeryFactory : ForgeryFactory<BatchClosedMet
 
     override fun getForgery(forge: Forge): BatchClosedMetadata {
         return BatchClosedMetadata(
-            lastTimeWasUsedInMs = forge.aPositiveLong(),
+            lastTimeWasUsedInMs = forge.aLong(min = Long.MAX_VALUE / 2),
```
0xnm, line 17: are there any issues without this change?

features/dd-sdk-android-logs/src/test/kotlin/com/datadog/android/log/internal/logger/DatadogLogHandlerTest.kt
```diff
@@ -199,7 +206,7 @@ internal class DatadogLogHandlerTest {
                 .hasThreadName(Thread.currentThread().name)
                 .hasStatus(fakeLevel.asLogStatus())
                 .hasMessage(fakeMessage)
-                .hasDateAround(now)
+                .hasDateAround(fakeTimestamp)
```
0xnm: minor: probably we don't need `Around` part anymore, since we can assert exact timesteamp

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumContinuousActionScopeTest.kt
```diff
@@ -364,13 +382,17 @@ internal class RumContinuousActionScopeTest {
         val expectedAttributes = mutableMapOf<String, Any?>()
         expectedAttributes.putAll(fakeAttributes)
         expectedAttributes.putAll(attributes)
+        val stopActionTime = TEST_INACTIVITY_MS * 2
```
0xnm: I think it is better to call this delay rather than time, because I was thinking it is a timestamp, but it is a delay/duration in fact. Same for other tests here.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManagerTest.kt
```diff
@@ -323,14 +333,15 @@ internal class ResourceDataStoreManagerTest {
...
         forge: Forge,
-        isExpired: Boolean
+        isExpired: Boolean,
+        currentTime: Long
     ): DataStoreContent<ResourceHashesEntry> {
         val resourceHashes = forge.aList { aString() }.distinct()
         val fakeVersionCode = forge.anInt(min = 0)
         val entryTime = if (isExpired) {
-            System.nanoTime() - DATASTORE_EXPIRATION_NS
+            currentTime - DATASTORE_EXPIRATION_NS - 1
```
0xnm, line 342: why do we add `-1`?

tools/unit/build.gradle.kts
```diff
@@ -68,6 +68,7 @@ dependencies {
     implementation(libs.bundles.testTools)
     implementation(libs.gson)
     implementation(libs.mockitoKotlin)
+    implementation(project(":dd-sdk-android-internal"))
```
0xnm: Probably it is better to avoid having such dependency and to place `StubTimeProvider` here, because this module is pure JVM.

Maybe consider having `StubTimeProvider` in the test fixtures or `dd-sdk-android-internal`? WDYT?

kikoveiga: Makes sense, moved it there.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumContinuousActionScopeTest.kt
kikoveiga: Renamed all occurrences.

features/dd-sdk-android-logs/src/test/kotlin/com/datadog/android/log/internal/logger/DatadogLogHandlerTest.kt
kikoveiga: Done.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
kikoveiga: No, rolled back.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
kikoveiga: This was just to clean the IDE warnings, do we usually care about these? If not, I could rollback.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/file/advanced/MoveDataMigrationOperationTest.kt
kikoveiga: Added a `whenever` statement to answer with real time in these functions to imitate the real delay. Also rolled back the functions' names.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/BatchClosedMetadataForgeryFactory.kt
kikoveiga: This ensures `lastTimeWasUsedInMs` is always greater than file creation timestamps used in tests, e.g. `BatchMetricsDispatcherTest: M send metric W sendBatchClosedMetric`.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
0xnm: What is the warning? I guess you have some different checks profile in the IDE, because I don't have any warning with the line before.

Probably it warns you about unused argument?

If that is the case, then more semantically correct will be to use `repeat(fakeInetAddresses.size)`, because we don't use elements of `fakeInetAddresses` while iterating it.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProviderTest.kt
```diff
@@ -10,130 +10,129 @@ import android.os.Build
...
-
-        // THEN
-        assertThat(providedStartTime)
-            .isCloseTo(startTimeNs, Offset.offset(TimeUnit.MILLISECONDS.toNanos(100)))
-    }
+    @Mock
+    private lateinit var mockBuildSdkVersionProvider: BuildSdkVersionProvider
 
     @Test
-    fun `M fall back to DdRumContentProvider W appStartTime { N+ getStartElapsedRealtime returns buggy value }`(
```
0xnm, line 57: it seems that the test `M fall back to DdRumContentProvider W appStartTime { N+ getStartElapsedRealtime returns buggy value }` got removed?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/MainLooperLongTaskStrategyTest.kt
```diff
@@ -103,40 +112,31 @@ internal class MainLooperLongTaskStrategyTest : ObjectTest<MainLooperLongTaskStr
...
         // Then
-        argumentCaptor<Long> {
-            verify(rumMonitor.mockInstance as AdvancedRumMonitor)
-                .addLongTask(capture(), eq("$target $callback: $what"))
-            val capturedMs = TimeUnit.NANOSECONDS.toMillis(firstValue)
-            assertThat(capturedMs)
-                .isBetween(duration, duration + 16L)
-        }
+        verify(rumMonitor.mockInstance as AdvancedRumMonitor)
+            .addLongTask(eq(fakeDurationNs), eq("$target $callback: $what"))
```
0xnm: tip: when all matchers are `eq`, they can be omitted, e.g.:

```suggestion
            .addLongTask(fakeDurationNs, "$target $callback: $what")
```

kikoveiga: Thank you!

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProviderTest.kt
kikoveiga: My bad, added it back.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/data/upload/RotatingDnsResolverTest.kt
kikoveiga: Yes, it warned about that. Change to use `repeat` now, thanks!

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManagerTest.kt
kikoveiga: The `- 1` is required because `didDataStoreExpire()` uses strict greater `>` comparison.

aleksandr-gringauz (dismissed):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/thread/ThreadPoolExecutorExtTest.kt
kikoveiga: We could add `verify(mockTimeProvider, atLeast(2)).getDeviceElapsedTimeNs()` in each test. It guarantees `sleepSafe` was called implicitly, do you think it's enough?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/utils/MiscUtilsTest.kt
kikoveiga: Added it back.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/thread/ThreadPoolExecutorExtTest.kt
0xnm: I'm not quite sure this is enough and it relies on the internals knowledge. What if code is changes and method under the test adds one more invocation of `getDeviceElapsedTimeNs` somewhere inside (say, for the logging). In this case test will be green even if we didn't wait enough.

kikoveiga: Agreed. Changed to use `mockTimeProvider` answering real `System.nanoTime`, so there will be real delays on the tests.

0xnm (dismissed):

ambushwork (approved):

0xnm (approved):

PR #3013, opened 2025-11-18
RUM-10907: Support `process_uptime` attribute for telemetry debug/error events

satween:
### What does this PR do?

* Adds new method to the `TimeProvider` allowing to get the amount of time (in milliseconds) since process start. The api for that was added only for Android SDK 24, so for previous versions class load time gonna be used.
* Fixes some assert messages that was copy-pasted and contained wrong values.
* Reduces code duplications for `TimeInfo` instance creation.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/TimeInfo.kt
```diff
@@ -14,10 +14,19 @@ package com.datadog.android.api.context
...
 data class TimeInfo(
     val deviceTimeNs: Long,
     val serverTimeNs: Long,
     val serverTimeOffsetNs: Long,
-    val serverTimeOffsetMs: Long
-)
+    val serverTimeOffsetMs: Long,
+    val deviceTimeSinceProcessStartMs: Long
+) {
+    companion object {
```
0xnm:
```suggestion
    internal companion object {
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/TimeInfo.kt
```diff
@@ -14,10 +14,19 @@ package com.datadog.android.api.context
  * to the device time (nanoseconds)
  * @property serverTimeOffsetMs the difference between server time and device time, relative
  * to the device time (milliseconds)
+ * @property deviceTimeSinceProcessStartMs the time spend since application's process start (for Android SDK >= 24) or
+ * between the class being loaded for Android < N as a fallback.
+ *
+ * to the device time (milliseconds)
```
0xnm: seems like it is detached from somewhere?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/TimeInfo.kt
```diff
@@ -14,10 +14,19 @@ package com.datadog.android.api.context
  * to the device time (nanoseconds)
  * @property serverTimeOffsetMs the difference between server time and device time, relative
  * to the device time (milliseconds)
+ * @property deviceTimeSinceProcessStartMs the time spend since application's process start (for Android SDK >= 24) or
+ * between the class being loaded for Android < N as a fallback.
```
0xnm: I'm curious if it is fine to have a different behavior for the same property. Maybe it is worth to make this property nullable and return `null` for API 23? And then it will be up to the caller what to do.

Otherwise we may see inconsistent data if the result of this property is used in some data we send.

WDYT?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
```diff
@@ -32,4 +36,26 @@ interface TimeProvider {
...
+
+    /**
+     * Return the milliseconds from the moment the process starts.
+     * For Android SDK versions below 24, the milliseconds since class loading are returned as a fallback.
+     *
+     * Note that calculations made with SystemClock.uptimeMillis() that do not count deep sleep time.
+     *
+     * @return The number of milliseconds elapsed since the start of the process.
+     */
+    fun getTimeSinceProcessStartMillis(): Long {
```
aleksandr-gringauz: We already have [this](https://github.com/DataDog/dd-sdk-android/blob/499916cfba381a40f1ae8db7f4f0bc1e0bfca941/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProvider.kt#L21) code to determine the time when the process started.

Is there a benefit for having another solution here? Should we maybe reuse what we already have?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
```diff
@@ -32,4 +36,26 @@ interface TimeProvider {
...
+            Process.getStartUptimeMillis()
+        } else {
+            applicationClassLoadMs
+        }
+
+        return SystemClock.uptimeMillis() - processStartUptimeMillis
+    }
+
+    companion object {
+        private val applicationClassLoadMs = SystemClock.uptimeMillis()
```
aleksandr-gringauz: Companion object if I am not mistaken is created (and as a result `applicationClassLoadMs` is initialized) at the moment when the first instance of the class is used (here it is `TimeProvider`). 

You can check by placing a breakpoint like this:
```
companion object {
    private val applicationClassLoadMs = run {
        SystemClock.uptimeMillis() // <--- place breakpoint here
    }
}
```

In this particular case it will happen somewhere around `Rum.enable` I guess, which is a lot later than process start.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
satween: oh cool, i didn't know that, will use it instead, thanks 👍

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
satween: yes, `@JvmFiled` is missed here, but nevermind I will use existing solution instead

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/TimeInfo.kt
satween: Well I was trying to find a balance between support complexity and value it provides:
1. `getStartElapsedRealtime` is supported from the version 24 of Android SDK, current `minSdk` is 23. So the amount of such devices is relatively small.
2. This attribute represents the amount of time spend from some moment in the past and represents relativity, not the actual time, from that perspective seems like there no so big difference from which point it would be calculated as the point remains the same for each device.
3. After @aleksandr-gringauz's suggestion, I changed logic to use `AppStartTimeProvider` so the value will be more even precise

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/TimeProvider.kt
aleksandr-gringauz: >yes, @JvmFiled is missed here, but nevermind I will use existing solution instead

I see that you have another solution now. But FYI I don't think that JvmField helps here. If you run this  https://pl.kotl.in/KMS7zASqs, you'll see that the second timestamp is bigger than the first one.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/DefaultAppStartTimeProvider.kt
```diff
@@ -28,4 +28,15 @@ internal class DefaultAppStartTimeProvider(
             else -> DdRumContentProvider.createTimeNs
         }
     }
+
+    override val appUptimeNs: Long by lazy(LazyThreadSafetyMode.PUBLICATION) {
```
aleksandr-gringauz: Are you sure that this field should be lazy? IIUC you want each invocation of `appUptimeNs` to return a different value (duration from process start till now).

You can probably just do `get() = System.nanoTime() - appStartTimeNs` or not create this new `appUptimeNs` and just do the calculation where you need it.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/time/DefaultTimeProvider.kt
```diff
@@ -11,7 +11,6 @@ package com.datadog.android.internal.time
  * The offsets are always 0.
  */
 class DefaultTimeProvider : TimeProvider {
-    override fun getDeviceTimestamp(): Long = System.currentTimeMillis()
```
0xnm, line 14: is this change needed?

jonathanmos (approved):

PR #3014, opened 2025-11-19
RUM-11821: Fix crash in KronosTimeProvider

aleksandr-gringauz:
### What does this PR do?

See the ticket for a stacktrace.

I haven't managed to determine how widespread this crash is, I can't find it in the list of crashes on google play console, I only can see it through the link shared by a single customer though google play console.

I haven't managed to reproduce it either.

The crash happens most likely because (it is the most plausible explanation I can come up with):
1. We call `kronosClock.shutdown` [here](https://github.com/DataDog/dd-sdk-android/blob/fa39327ad28cfe5a1792db2a7f966526e75a1aec/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L272).
2. After that `kronosClock.getCurrentTimeMs` is called as a result of [this](https://github.com/DataDog/dd-sdk-android/blob/88d29bab0c5307e83f68c412551b5b20d8d0fd43/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogEventListener.kt#L90) call to `sdkCore.time.deviceTimeNs`.
3. [This](https://github.com/lyft/Kronos-Android/blob/b2774ab9458ffd0202e2c4880e2a181134953beb/kronos-java/src/main/java/com/lyft/kronos/internal/ntp/SntpService.kt#L98C9-L98C31) `ensureServiceIsRunning` call throws `IllegalStateException` [here](https://github.com/lyft/Kronos-Android/blob/b2774ab9458ffd0202e2c4880e2a181134953beb/kronos-java/src/main/java/com/lyft/kronos/internal/ntp/SntpService.kt#L142). 
4. We interact with already shutdown kronos clock, because [this](https://github.com/DataDog/dd-sdk-android/blob/88d29bab0c5307e83f68c412551b5b20d8d0fd43/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogEventListener.kt#L90) `sdkCore.time.deviceTimeNs` can get the old `timeProvider`, after that `kronosClock.shutdown` is called on another thread and finally `sdkCore.time.deviceTimeNs` interacts with this clock using the old 'timeProvider`.

This PR does the following:
Add try/catch in `KronosTimeProvider` so that it becomes physically impossible to crash if by some reason we interact with an already shutdown kronos clock.

We could make `timeProvider` volatile. But it will not help with the point 4. `sdkCore.time.deviceTimeNs` is not "atomic", clock can be shut down at any point during execution of this code, so the best we can do is try/catch.

### Motivation

### Additional Notes

### DIFF
--- a/detekt_custom_safe_calls.yml
+++ b/detekt_custom_safe_calls.yml
@@ -577,7 +577,7 @@ datadog:
       - "kotlin.ReplaceWith.constructor(kotlin.String, kotlin.Array)"
       - "kotlin.Result.failure(kotlin.Throwable)"
       - "kotlin.Result.fold(kotlin.Function1, kotlin.Function1)"
-      - "kotlin.Result.getOrDefault(kotlin.Int)"
+      - "kotlin.Result.getOrDefault(kotlin.Long)"
       - "kotlin.Result.getOrElse(kotlin.Function1)"
       - "kotlin.Result.map(kotlin.Function1)"
       - "kotlin.Result.onFailure(kotlin.Function1)"


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt
```diff
@@ -6,30 +6,57 @@
...
-    private val clock: Clock
+    private val clock: Clock,
+    private val internalLogger: InternalLogger
 ) : TimeProvider {
 
     override fun getDeviceTimestamp(): Long {
         return System.currentTimeMillis()
     }
 
+    @Suppress("TooGenericExceptionCaught")
```
0xnm: there is no need for this suppress here and below

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt
```diff
@@ -6,30 +6,57 @@
...
         return TimeUnit.MILLISECONDS.toNanos(getServerOffsetMillis())
     }
+
+    private fun Clock.safeGetCurrentTimeMs(): Result<Long> {
+        return runCatching {
+            getCurrentTimeMs()
+        }.onFailure { ex ->
+            internalLogger.log(
+                level = InternalLogger.Level.WARN,
+                targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
0xnm, line 48: I'm a bit afraid that adding `TELEMETRY` target here may cause a recursion loop. Yes, we have `onlyOnce = true`, but can you double-check that no time info is gathered between `log` call is made and `onlyOnce` check is actually hit?

detekt_custom_safe_calls.yml
```diff
@@ -1187,5 +1190,6 @@ datadog:
       - "com.squareup.sqldelight.android.AndroidSqliteDriver.Callback.onCorruption(androidx.sqlite.db.SupportSQLiteDatabase)"
       # endregion
       # region Kronos
+      - "com.lyft.kronos.Clock.runCatching(kotlin.Function1)"
```
0xnm, line 1194: is it how Detekt check sees it? `runCatching` definition has no receiver

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt
```diff
@@ -6,30 +6,57 @@
...
-        val device = System.currentTimeMillis()
-        val delta = server - device
-        return delta
+        return clock.safeGetCurrentTimeMs()
+            .map { server ->
+                val device = System.currentTimeMillis()
+                val delta = server - device
+                delta
+            }
+            .getOrElse { 0 }
```
jonathanmos: Minor: Here the fallback is a primitive. I think it would be more idiomatic to have an elvis operator

ambushwork (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt
aleksandr-gringauz: Thanks for mentioning it, I haven't thought about this situation.

I checked that:
1. Yes, recursion indeed happens because [here](https://github.com/DataDog/dd-sdk-android/blob/363c1aec9afd47948c244c0440f6fb8e681b71ac/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt#L74) when `DatadogContext` is being [created](https://github.com/DataDog/dd-sdk-android/blob/363c1aec9afd47948c244c0440f6fb8e681b71ac/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/SdkFeature.kt#L182) it calls `timeProvider.getServerTimestamp`.
2. But `onlyOnce` helps, I observe that [this](https://github.com/DataDog/dd-sdk-android/blob/363c1aec9afd47948c244c0440f6fb8e681b71ac/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/logger/SdkInternalLogger.kt#L234) line gets executed when the logging is called on the second iteration of "recursion".

So looks like everything is fine here.

detekt_custom_safe_calls.yml
aleksandr-gringauz: It is [this](https://github.com/JetBrains/kotlin/blob/0ada970e6d69ad05a006e778426aebf2c319495c/libraries/stdlib/src/kotlin/util/Result.kt#L155) function. It has a receiver. `getCurrentTimeMs` is a method of `Clock`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt
aleksandr-gringauz: Do you suggest this:

```
return clock.safeGetCurrentTimeMs()
    .map { server ->
        val device = System.currentTimeMillis()
        val delta = server - device
        delta
    }
    .getOrNull() ?: 0
```
?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt
aleksandr-gringauz: fixed

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt
aleksandr-gringauz: Changed to `getOrDefault(0)` as discussed in DMs.

PR #3017, opened 2025-11-19
RUM-12923: Attach RUM information on profiling event

ambushwork:
### What does this PR do?

This PR attaches following RUM context information in Profiling event so that it can be shown as tags in profiling page:
* session id
* vital id
* view name
* view id
Also it changes the timing of RUM feature send TTID event to Profiling feature in order to stop the recording, so that the profiling event can attach the real first view event instead of artificial "AppLaunch" view event.

### Motivation

RUM-12923

### Demo

image

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/rum/TTIDEvent.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/rum/TTIDEvent.kt
@@ -9,6 +9,18 @@ package com.datadog.android.rum
 /**
  * Internal event to pass the Time To Initial Display (TTID) value in nanoseconds.
  *
- * @param value The TTID value in nanoseconds.
+ * @param durationNs The TTID value in nanoseconds.
+ * @param applicationId The Id of the application of RUM.
+ * @param sessionId The Id of the RUM session where TTID is captured
+ * @param vitalId The Id of the TTID vital event
+ * @param viewId The Id of the view where TTID is captured
+ * @param viewName The name of the view where TTID is captured
  */
-data class TTIDEvent(val value: Long)
+data class TTIDEvent(
+    val durationNs: Long,
+    val applicationId: String,
+    val sessionId: String,
+    val vitalId: String,
+    val viewId: String?,
+    val viewName: String?
+)
--- a/dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/TTIDEventFactory.kt
+++ b/dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/TTIDEventFactory.kt
@@ -1,0 +1,24 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.tests.elmyr
+
+import com.datadog.android.rum.TTIDEvent
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.ForgeryFactory
+
+class TTIDEventFactory : ForgeryFactory<TTIDEvent> {
+    override fun getForgery(forge: Forge): TTIDEvent {
+        return TTIDEvent(
+            durationNs = forge.aLong(),
+            applicationId = forge.anAlphabeticalString(),
+            sessionId = forge.anAlphabeticalString(),
+            vitalId = forge.anAlphabeticalString(),
+            viewId = forge.aNullable { anAlphabeticalString() },
+            viewName = forge.aNullable { anAlphabeticalString() }
+        )
+    }
+}
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
@@ -15,20 +15,23 @@ import com.datadog.android.core.internal.persistence.file.readBytesSafe
 import com.datadog.android.internal.utils.formatIsoUtc
 import com.datadog.android.profiling.internal.perfetto.PerfettoResult
 import com.datadog.android.profiling.model.ProfileEvent
+import com.datadog.android.rum.TTIDEvent
 import java.io.File
 
 internal class ProfilingDataWriter(
     private val sdkCore: FeatureSdkCore
 ) : ProfilingWriter {
     override fun write(
-        profilingResult: PerfettoResult
+        profilingResult: PerfettoResult,
+        ttidEvent: TTIDEvent?
     ) {
         sdkCore.getFeature(Feature.PROFILING_FEATURE_NAME)
             ?.withWriteContext { context, writeScope ->
                 writeScope { writer ->
                     val rawBatchEvent = buildRawBatchEvent(
                         context = context,
-                        profilingResult = profilingResult
+                        profilingResult = profilingResult,
+                        ttidEvent = ttidEvent
                     )
                     if (rawBatchEvent != null) {
                         synchronized(this) {
@@ -45,13 +48,18 @@ internal class ProfilingDataWriter(
 
     private fun buildRawBatchEvent(
         context: DatadogContext,
-        profilingResult: PerfettoResult
+        profilingResult: PerfettoResult,
+        ttidEvent: TTIDEvent?
     ): RawBatchEvent? {
         val byteData = readProfilingData(profilingResult.resultFilePath)
         if (byteData == null || byteData.isEmpty()) {
             return null
         }
-        val profileEvent = createProfileEvent(context, profilingResult)
+        val profileEvent = createProfileEvent(
+            context,
+            profilingResult,
+            ttidEvent
+        )
         val serializedEvent =
             profileEvent.toJson().toString().toByteArray(Charsets.UTF_8)
         return RawBatchEvent(data = serializedEvent, metadata = byteData)
@@ -59,7 +67,8 @@ internal class ProfilingDataWriter(
 
     private fun createProfileEvent(
         context: DatadogContext,
-        profilingResult: PerfettoResult
+        profilingResult: PerfettoResult,
+        ttidEvent: TTIDEvent?
     ): ProfileEvent {
         return ProfileEvent(
             start = formatIsoUtc(profilingResult.start),
@@ -68,11 +77,14 @@ internal class ProfilingDataWriter(
             family = ANDROID_FAMILY_NAME,
             runtime = ANDROID_RUNTIME_NAME,
             version = VERSION_NUMBER,
-            tagsProfiler = buildTags(context)
+            tagsProfiler = buildTags(context, ttidEvent)
         )
     }
 
-    private fun buildTags(context: DatadogContext): String = buildString {
+    private fun buildTags(
+        context: DatadogContext,
+        ttidEvent: TTIDEvent?
+    ): String = buildString {
         append("$TAG_KEY_SERVICE:${context.service}")
         append(",")
         append("$TAG_KEY_ENV:${context.env}")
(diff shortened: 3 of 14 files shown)


ambushwork:
application id + view id

dd-sdk-android-internal/src/main/java/com/datadog/android/rum/TTIDEvent.kt
```diff
@@ -10,5 +10,15 @@ package com.datadog.android.rum
...
+ * @param viewName The name of the view where TTID is captured
+ * @param vitalId The Id of the TTID vital event
  */
-data class TTIDEvent(val value: Long)
+data class TTIDEvent(
+    val value: Long,
+    val sessionId: String,
+    val viewId: String?,
+    val viewName: String?,
+    val vitalId: String?
```
0xnm: TTID event should always come with Vital ID

dd-sdk-android-internal/src/main/java/com/datadog/android/rum/TTIDEvent.kt
```diff
@@ -10,5 +10,15 @@ package com.datadog.android.rum
...
  * @param value The TTID value in nanoseconds.
+ * @param sessionId The Id of the RUM session where TTID is captured
+ * @param viewId The Id of the view where TTID is captured
+ * @param viewName The name of the view where TTID is captured
+ * @param vitalId The Id of the TTID vital event
  */
-data class TTIDEvent(val value: Long)
+data class TTIDEvent(
+    val value: Long,
+    val sessionId: String,
```
0xnm, line 22: we should have RUM application ID as well

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -29,6 +29,8 @@ internal class ProfilingFeature(
 
     private var dataWriter: ProfilingWriter = NoOpProfilingWriter()
 
+    private var profilingContext: ProfilingContext = ProfilingContext()
```
0xnm: Maybe we can simply has a property `ttidEvent: TTIDEvent?`? `ProfilingContext` is too generic and have all properties nullable, and also having dedicated property for a particular event will allow us better associate profiling result with a corresponding vital in the future.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumVitalAppLaunchEventHelper.kt
```diff
@@ -29,6 +29,7 @@ internal class RumVitalAppLaunchEventHelper(
 ) {
     @Suppress("LongMethod")
     fun newVitalAppLaunchEvent(
+        vitalId: String?,
```
0xnm: it is better to make it non-nullable and simply generate each time at the call site: this will avoid nullability and usage of elvis operator.

0xnm (changes requested):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumVitalAppLaunchEventHelper.kt
0xnm: It is also worth to make the type here `UUID` and convert it to `String` only at the model creation moment.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -99,8 +102,19 @@ internal class RumSessionScopeStartupManagerImpl(
 
         ttidSentForSession = true
 
+        val ttidVitalId = UUID.randomUUID().toString()
```
aleksandr-gringauz: You could do it in an easier way I think:

```
val ttidEvent = rumVitalAppLaunchEventHelper.newVitalAppLaunchEvent(
    timestampMs = event.info.scenario.initialTime.timestamp + sdkCore.time.serverTimeOffsetMs,
    datadogContext = datadogContext,
    eventAttributes = emptyMap(),
    customAttributes = customAttributes,
    hasReplay = null,
    rumContext = rumContext,
    durationNs = event.info.durationNs,
    appLaunchMetric = RumVitalAppLaunchEvent.AppLaunchMetric.TTID,
    scenario = event.info.scenario
)

sdkCore.getFeature(Feature.PROFILING_FEATURE_NAME)?.sendEvent(
    TTIDEvent(
        value = event.info.durationNs,
        sessionId = rumContext.sessionId,
        viewId = rumContext.viewId,
        viewName = rumContext.viewName,
        vitalId = ttidEvent.vital.id
    )
)

sdkCore.newRumEventWriteOperation(datadogContext, writeScope, writer) { ttidEvent }.submit()

```

This way no changes in `RumVitalAppLaunchEventHelper` are needed. You don't need to pass `vitalId` from outside of `RumVitalAppLaunchEventHelper`, just use the one that is generated there.

ambushwork: yes this is the best, updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumVitalAppLaunchEventHelper.kt
ambushwork: accept Sasha's suggestion here

aleksandr-gringauz (dismissed):

dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/TTIDEventFactory.kt
```diff
@@ -0,0 +1,24 @@
...
+
+class TTIDEventFactory : ForgeryFactory<TTIDEvent> {
+    override fun getForgery(forge: Forge): TTIDEvent {
+        return TTIDEvent(
+            value = forge.aLong(),
+            applicationId = forge.anAlphabeticalString(),
+            sessionId = forge.anAlphabeticalString(),
+            vitalId = forge.anAlphabeticalString(),
+            viewId = forge.anAlphabeticalString(),
+            viewName = forge.anAlphabeticalString()
```
0xnm: these are nullables, so it should be:

```suggestion
            viewId = forge.aNullable { anAlphabeticalString() },
            viewName = forge.aNullable { anAlphabeticalString() }
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -15,20 +15,23 @@ import com.datadog.android.core.internal.persistence.file.readBytesSafe
...
+import com.datadog.android.rum.TTIDEvent
 import java.io.File
 
 internal class ProfilingDataWriter(
     private val sdkCore: FeatureSdkCore
 ) : ProfilingWriter {
     override fun write(
-        profilingResult: PerfettoResult
+        profilingResult: PerfettoResult,
+        ttidEvent: TTIDEvent?
```
0xnm, line 26: For now, if we don't have `TTIDEvent`, we shouldn't even write profile and upload it.

I guess the goal is to have only profiles with RUM context associated.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -99,18 +101,30 @@ internal class RumSessionScopeStartupManagerImpl(
...
-                timestampMs = event.info.scenario.initialTime.timestamp + sdkCore.time.serverTimeOffsetMs,
-                datadogContext = datadogContext,
-                eventAttributes = emptyMap(),
-                customAttributes = customAttributes,
-                hasReplay = null,
-                rumContext = rumContext,
-                durationNs = event.info.durationNs,
-                appLaunchMetric = RumVitalAppLaunchEvent.AppLaunchMetric.TTID,
-                scenario = event.info.scenario
+        val ttidEvent = rumVitalAppLaunchEventHelper.newVitalAppLaunchEvent(
```
aleksandr-gringauz, line 104: Sorry, just remembered one thing.

Don't you need to add [this](https://github.com/DataDog/rum-events-format/blob/834392ddf77531ed3f383e0808192879490c221d/schemas/rum/vital-app-launch-schema.json#L60) to vital event?

I'm not in full context for profiling project, but this field exists for some reason in the schema. Or should it be removed?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -67,6 +72,7 @@ internal class ProfilingFeature(
             )
             return
         }
+        this.ttidEvent = event
         profiler.stop()
```
0xnm, line 76: minor: for the future, since we may have different profile starts reasons, it may be better to simply pass `ttidEvent` as an argument of `stop` (obviously, it will have multiple overloads).

And then, inside the stop we can bundle result with event by checking the result tag and pass both in `ProfilingSuccess`, this allows better result bundling.

not blocking for now.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -99,18 +101,30 @@ internal class RumSessionScopeStartupManagerImpl(
...
+            customAttributes = customAttributes,
+            hasReplay = null,
+            rumContext = rumContext,
+            durationNs = event.info.durationNs,
+            appLaunchMetric = RumVitalAppLaunchEvent.AppLaunchMetric.TTID,
+            scenario = event.info.scenario
+        )
+        sdkCore.getFeature(Feature.PROFILING_FEATURE_NAME)?.sendEvent(
+            TTIDEvent(
+                value = event.info.durationNs,
```
0xnm: should we rename `TTIDEvent.value` to `TTIDEvent.duration`? because currently it is unclear on the consuming side if it is duration or timestamp, `value` is way too generic term.

0xnm (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
ambushwork: good point, this will be my next PR

aleksandr-gringauz: ok

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: in previous dogfooding, all the profiles > 1min are probably the profiles without correct TTID signal, if we stop sending them, we won't be able to see these kinds of abnormal profiles in next dogfooding.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
ambushwork: renamed to `durationNs`

aleksandr-gringauz (approved):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
0xnm: > all the profiles > 1min are probably the profiles without correct TTID signal

I don't think it is a problem of TTID signal, but rather a problem of Perfetto callback result delayed.

But we can keep it if you suggest it is valuable to have such, we just need to remove that before the GA.

0xnm (approved):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: yes I agree, my point is that we need to keep it for exposing the perfetto callback issue

PR #3022, opened 2025-11-21
RUM-5387: Override Metrics telemetry sample rate

kikoveiga:
### What does this PR do?

Creates a `metricTelemetrySampleRateBypass` field in `CoreFeature` and respective internal API setter, that will bypass any sample rates previously set throughout the codebase for metrics. Setting this and `telemetrySampleRate` to `100%` will allow the Shopist app to send all metrics.

### Motivation

Shopist metric telemetry was sampled the same way as our customers (low sample rate). We want to be able to customize this in order to have more events for dogfooding.


dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
```diff
@@ -518,6 +504,86 @@ internal class SdkInternalLoggerTest {
...
+    @Test
+    fun `M use bypass metric telemetry sample rate W logMetric() {bypass is set}`(
+        forge: Forge,
+        @FloatForgery(min = 0.0f, max = 100.0f) fakeSampleRate: Float
+    ) {
+        // Given
+        val bypassSampleRate = 100.0f
+        givenLoggerWithMetricTelemetrySampleRateBypass(bypassSampleRate)
+
+        val mockLambda = mockMessageLambda(forge.aString())
```
ambushwork: nip: to be consistent, this can be a @StringForgery param,

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
```diff
@@ -518,6 +504,86 @@ internal class SdkInternalLoggerTest {
         }
     }
 
+    @Test
+    fun `M use bypass metric telemetry sample rate W logMetric() {bypass is set}`(
+        forge: Forge,
+        @FloatForgery(min = 0.0f, max = 100.0f) fakeSampleRate: Float
+    ) {
+        // Given
+        val bypassSampleRate = 100.0f
```
ambushwork, line 513: This can be faked as a random value by forge as well I think

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
```diff
@@ -518,6 +504,86 @@ internal class SdkInternalLoggerTest {
...
+            ).isEqualTo(bypassSampleRate)
+        }
+    }
+
+    @Test
+    fun `M use hardcoded sample rate W logMetric() {bypass is null}`(
+        forge: Forge
+    ) {
+        // Given
+        val hardcodedSampleRate = 100.0f
```
ambushwork, line 542: same as above this can be a random value

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
```diff
@@ -518,6 +504,86 @@ internal class SdkInternalLoggerTest {
...
+            val metricEvent = firstValue as InternalTelemetryEvent.Metric
+
+            assertThat(
+                metricEvent.additionalProperties?.get(LocalAttribute.Key.REPORTING_SAMPLING_RATE.toString())
+            ).isEqualTo(hardcodedSampleRate)
+        }
+    }
+
+    @Test
+    fun `M not send metric W logMetric() {bypass rate is 0}`(
```
ambushwork, line 567: If I don't miss anything, there should be a test like  `M send metric W logMetric() {bypass rate is 100}` ?

ambushwork (dismissed):
Well done, just a few comments in the unit test

dd-sdk-android-core/src/main/kotlin/com/datadog/android/_InternalProxy.kt
```diff
@@ -74,6 +75,12 @@ class _InternalProxy internal constructor(
     @Suppress("PropertyName")
     val _telemetry: _TelemetryProxy = _TelemetryProxy(sdkCore)
 
+    fun setMetricTelemetrySampleRateBypass(
+        @FloatRange(from = 0.0, to = 100.0) sampleRate: Float
+    ) {
+        (sdkCore as? DatadogCore)?.coreFeature?.metricTelemetrySampleRateBypass = sampleRate
```
0xnm, line 81: I'm not quite sure it is a safe thing to do, given that here we apply the same rate for all the metrics logged and right now the rate used for metrics call vary from `0.75f` to `100f`, but at worst case Shopist shouldn't bring a huge amount of data.

Also by setting it to non-100 we increase the rate for some metrics and decrease for some others.

Although I would question if we still need this change.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
```diff
@@ -150,6 +150,8 @@ class SampleApplication : Application() {
             preferences.getTrackingConsent()
         )
 
+        Datadog._internalProxy().setMetricTelemetrySampleRateBypass(100f)
```
0xnm: Is it useful in the sample app? I was thinking it is rather for Shopist.

I would suggest to remove this, given it is an example on how to setup SDK for customers and some customers may copy-paste this by mistake, making unwanted consequences.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
kikoveiga: The idea of using `100.0f` is to guarantee that the metric is always sent. Having a forged rate will make the tests flaky.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/logger/SdkInternalLoggerTest.kt
kikoveiga: Please read my comment above.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/_InternalProxy.kt
kikoveiga: It's true that some metrics will go from a sample rate of `0.75f` to `100f`, representing a `133x` increase. I estimated that, in the Shopist Android app, there will be a `40x` overall increase throughout all the metrics. Still, this will be a very small number when taking into account the intake from all the customers.

0xnm: Maybe a better API would be a sampling callback which can decide if particular metric can be sampled or not. And if decision is not made by this callback, then default sampling mechanism should be applied.

This would allow a granular control, because with the current change we apply the same rate for all metrics, meaning some will go down, some will go up.

But this API is not blocking for me.

0xnm (approved):

PR #3024, opened 2025-11-24
RUM-13468: Add `totalRam`, `logicalCpuCount` and `isLowRam` to DeviceInfo

marco-saia-datadog:
### What does this PR do?

Adds new attributes to the Device Info:

- **totalRam**: The total RAM in megabytes
- **processorCount**: Number of logical CPU cores available for scheduling on the device at runtime, as reported by the operating system.
- **isLowRam**: Whether the device is considered a low RAM device (from `ActivityManager.isLowRamDevice`)

### Additional Notes

- The relevant schemas have been modified locally
- **PR in rum-events-format**: https://github.com/DataDog/rum-events-format/pull/323
- **PR in rum-events-format**: https://github.com/DataDog/rum-events-format/pull/337 (refactoring)


features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
```diff
@@ -355,6 +355,21 @@
           "type": "number",
           "description": "Current screen brightness level (0.0 to 1.0).",
           "readOnly": true
+        },
+        "processor_count": {
+          "type" : "number",
+          "description": "Number of device processors",
+          "readOnly": true
+        },
```
0xnm: Probably the naming is not the best one, because processor is a chip containing multiple cores. We probably want to send the number of _physical_ cores, right?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/system/DefaultAndroidInfoProvider.kt
```diff
@@ -127,10 +128,37 @@ internal class DefaultAndroidInfoProvider(
         }
     }
 
+    override val processorCount: Int by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        Runtime.getRuntime().availableProcessors()
+    }
```
0xnm: not sure if we want to send the number of processors (chips) or cores, but this method returns the number of _physical_ cores and not the number of chips.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/system/DefaultAndroidInfoProvider.kt
```diff
@@ -127,10 +128,37 @@ internal class DefaultAndroidInfoProvider(
         }
     }
 
+    override val processorCount: Int by lazy(LazyThreadSafetyMode.PUBLICATION) {
```
0xnm: unless we have any performance concerns, we can use default mode

```suggestion
    override val processorCount: Int by lazy {
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/system/DefaultAndroidInfoProvider.kt
```diff
@@ -127,10 +128,37 @@ internal class DefaultAndroidInfoProvider(
...
+        Runtime.getRuntime().availableProcessors()
+    }
+
+    @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
+    override val totalRam: Int? by lazy(LazyThreadSafetyMode.PUBLICATION) {
```
0xnm:
```suggestion
    override val totalRam: Int? by lazy {
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/system/DefaultAndroidInfoProvider.kt
```diff
@@ -127,10 +128,37 @@ internal class DefaultAndroidInfoProvider(
...
+            val info = ActivityManager.MemoryInfo()
+            activityManager.getMemoryInfo(info)
+            (info.totalMem / MB_IN_BYTES).toInt()
+        } catch (_: Exception) {
+            null
+        }
+    }
+
+    @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
+    override val isLowRamDevice: Boolean? by lazy(LazyThreadSafetyMode.PUBLICATION) {
```
0xnm:
```suggestion
    override val isLowRamDevice: Boolean? by lazy {
```

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/system/DefaultAndroidInfoProvider.kt
```diff
@@ -127,10 +128,37 @@ internal class DefaultAndroidInfoProvider(
         }
     }
 
+    override val processorCount: Int by lazy(LazyThreadSafetyMode.PUBLICATION) {
+        Runtime.getRuntime().availableProcessors()
+    }
+
+    @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
```
0xnm: minor: it is better to move this annotation right to the call site rather than to keep it for the whole method. I guess here it will be `getMemoryInfo` and `ActivityManager.MemoryInfo()` constructor can be declared as safe in YAML config.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/AndroidInfoProviderForgeryFactory.kt
```diff
@@ -27,6 +27,9 @@ internal class AndroidInfoProviderForgeryFactory : ForgeryFactory<AndroidInfoPro
         val locales = forge.aList { forge.aString() }
         val currentLocale = forge.aString()
         val timeZone = forge.aString()
+        val processorCount: Int = forge.anInt()
+        val totalRam: Int = forge.anInt()
+        val isLowRamDevice: Boolean = forge.aBool()
```
0xnm:
```suggestion
        val processorCount = forge.anInt()
        val totalRam = forge.aNullable { anInt() }
        val isLowRamDevice = forge.aNullable { aBool() }
```

dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/DeviceInfoForgeryFactory.kt
```diff
@@ -24,7 +24,10 @@ class DeviceInfoForgeryFactory : ForgeryFactory<DeviceInfo> {
             osMajorVersion = forge.aString(),
             architecture = forge.aString(),
             numberOfDisplays = forge.aNullable { forge.anInt() },
-            localeInfo = forge.getForgery()
+            localeInfo = forge.getForgery(),
+            processorCount = forge.anInt(),
+            totalRam = forge.anInt(),
+            isLowRamDevice = forge.aBool()
```
0xnm:
```suggestion
            totalRam = forge.aNullable { anInt() },
            isLowRamDevice = forge.aNullable { aBool() }
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -395,7 +395,10 @@ internal open class RumViewScope(
                 timeZone = datadogContext.deviceInfo.localeInfo.timeZone,
                 batteryLevel = batteryInfo.batteryLevel,
                 powerSavingMode = batteryInfo.lowPowerMode,
-                brightnessLevel = displayInfo.screenBrightness
+                brightnessLevel = displayInfo.screenBrightness,
+                processorCount = datadogContext.deviceInfo.processorCount,
+                totalRam = datadogContext.deviceInfo.totalRam,
+                isLowRamDevice = datadogContext.deviceInfo.isLowRamDevice
```
0xnm: we need to add assertions for these new properties in tests for `RumViewScope`, `RumResourceScope`, etc.

features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
marco-saia-datadog: Hey @0xnm  thanks for the review! I’m picking this PR up again, sorry for the delay.

I initially went with `processor_count` rather than something more precise is because both the Android and iOS APIs use the term processor, even though they effectively return the number of CPU cores:

- Android: `Runtime.getRuntime().availableProcessors()`
- iOS: `ProcessInfo.processInfo.processorCount`

That said, I'm not particularly happy with this name either, as it could be clearer. I was thinking `cpu_core_count`.

WDYT?

0xnm: Apple [docs](https://developer.apple.com/documentation/foundation/processinfo/processorcount?language=objc) say:

```
processorCount
The number of processing cores available on the computer.
```

Android [docs](https://developer.android.com/reference/java/lang/Runtime#availableProcessors()) say:

```
Returns the number of processors available to the Java virtual machine.

This value may change during a particular invocation of the virtual machine. Applications that are sensitive to the number of available processors should therefore occasionally poll this property and adjust their resource usage appropriately.
```

Apple docs are more explicit about the cores meaning behind. But given that both are using processors <-> cores lingvo interchangeably, I don't have a strong no against it, so I'm okay with any of `cpu_core_count` or `processor_count`. Maybe the former is a bit better for understanding (not sure if we can find any mobile devices with multiple CPU units).

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/context/DeviceInfo.kt
```diff
@@ -32,5 +36,8 @@ data class DeviceInfo(
     val osVersion: String,
     val architecture: String,
     val numberOfDisplays: Int?,
-    val localeInfo: LocaleInfo
+    val localeInfo: LocaleInfo,
+    val logicalCpuCount: Int,
+    val totalRam: Int?,
+    val isLowRam: Boolean?
```
0xnm, line 42: I believe these two are nullable because the API we use to get this info may not return the data?

UPD: I see, it is due to the exception handling.

detekt_custom_safe_calls.yml
```diff
@@ -22,6 +22,7 @@ datadog:
       - "android.content.Context.getSharedPreferences(kotlin.String?, kotlin.Int)"
       - "android.content.Context.getSystemService(kotlin.String)"
       - "android.content.Context.getSharedPreferences(kotlin.String?, kotlin.Int)"
+      - "android.content.Context.getSystemService(java.lang.Class)"
```
0xnm, line 25: neat! we indeed can use it now since min API is 23.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventDeserializerTest.kt
```diff
@@ -41,7 +41,7 @@ import org.mockito.quality.Strictness
     ExtendWith(ForgeExtension::class)
 )
 @MockitoSettings(strictness = Strictness.LENIENT)
-@ForgeConfiguration(Configurator::class)
+@ForgeConfiguration(value = Configurator::class, seed = 0xd1285cd0106L)
```
0xnm: this is a debugging leftover and should be removed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedActionEventAssert.kt
```diff
@@ -24,7 +24,7 @@ internal class DeserializedActionEventAssert(actual: ActionEvent) :
             .isEqualTo(expected)
         assertThat(actual.device)
             .usingRecursiveComparison()
-            .ignoringFields("batteryLevel", "brightnessLevel")
+            .ignoringFields("batteryLevel", "brightnessLevel", "totalRam", "logicalCpuCount")
```
0xnm: Afaik this was to avoid losing precision in the floating point values between serialization/deserialization roundtrip, so it is not needed for the integer values. Same for other deserialized assertions below.

0xnm (commented):
LGTM, just some minor cleanup is needed.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedActionEventAssert.kt
marco-saia-datadog: Actually, I added this because the recursive comparison seems to break with integer values, but I am happy to revisit it. Maybe I'll clue you in with the error stacktrace, you might have some useful insight to share. I'll follow up soon.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventDeserializerTest.kt
marco-saia-datadog: Nice catch 🙏

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedActionEventAssert.kt
marco-saia-datadog: In `RumEventDeserializerTest.kt` using:

```
@ForgeConfiguration(value = Configurator::class, seed = 0xd1285cd0106L)
```

The forged event has:

```
logicalCpuCount = 1181136908
totalRam = -479744429
 
// Also, would it make sense to only forge positive values?
// I kept it like this to cover all possible values and make sure 
// we never get runtime errors as long as these values are Integers
```

And I get this error:

```
Can't find any field or property with name 'value'.
Error when introspecting properties was :
- No getter for property 'value' in java.lang.Integer 
Error when introspecting fields was :
- Unable to obtain the value of the field <'value'> from <-479744429>
org.assertj.core.util.introspection.IntrospectionError: 
Can't find any field or property with name 'value'.
Error when introspecting properties was :
- No getter for property 'value' in java.lang.Integer 
Error when introspecting fields was :
- Unable to obtain the value of the field <'value'> from <-479744429>
	at app//org.assertj.core.util.introspection.PropertyOrFieldSupport.getSimpleValue(PropertyOrFieldSupport.java:88)
	at app//org.assertj.core.api.recursive.comparison.RecursiveComparisonDifferenceCalculator.determineDifferences(RecursiveComparisonDifferenceCalculator.java:295)
	at app//org.assertj.core.api.recursive.comparison.RecursiveComparisonDifferenceCalculator.determineDifferences(RecursiveComparisonDifferenceCalculator.java:186)
	at app//org.assertj.core.api.RecursiveComparisonAssert.determineDifferencesWith(RecursiveComparisonAssert.java:1199)
	at app//org.assertj.core.api.RecursiveComparisonAssert.isEqualTo(RecursiveComparisonAssert.java:160)
	at app//com.datadog.android.rum.utils.assertj.DeserializedLongTaskEventAssert.isEqualTo(DeserializedLongTaskEventAssert.kt:28)
	at app//com.datadog.android.rum.internal.domain.event.RumEventDeserializerTest.M deserialize a serialized RUM LongTaskEvent W deserialize()(RumEventDeserializerTest.kt:156)
	at java.base@21.0.8/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)

Caused by: org.assertj.core.util.introspection.IntrospectionError: Unable to obtain the value of the field <'value'> from <-479744429>
	at app//org.assertj.core.util.introspection.FieldSupport.readSimpleField(FieldSupport.java:248)
	at app//org.assertj.core.util.introspection.FieldSupport.fieldValue(FieldSupport.java:202)
	at app//org.assertj.core.util.introspection.PropertyOrFieldSupport.getSimpleValue(PropertyOrFieldSupport.java:70)
	... 87 more

Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field private final int java.lang.Integer.value accessible: module java.base does not "opens java.lang" to unnamed module @72a34537
	at java.base/java.lang.reflect.AccessibleObject.throwInaccessibleObjectException(Unknown Source)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(Unknown Source)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(Unknown Source)
	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Unknown Source)
	at java.base/java.lang.reflect.Field.setAccessible(Unknown Source)
	at org.assertj.core.util.introspection.FieldUtils.getField(FieldUtils.java:67)
	at org.assertj.core.util.introspection.FieldUtils.readField(FieldUtils.java:141)
	at org.assertj.core.util.introspection.FieldSupport.readSimpleField(FieldSupport.java:208)
```

I'll debug this further and get back to you, but please let me know if this is a known issue or if you have any insight! :)

marco-saia-datadog: I think I have found a fix!

When serializing and deserializing the forged event with JsonParser, the `Integer` values are converted to `Number` values, specifically to `LazilyParsedNumber`.

Internally, `assertj` encounters the attribute (e.g `totalRam`), and calls:

```java
Set<String> actualFieldsNames = Objects.getFieldsNames(dualValue.actual.getClass());
```

Which returns `["value"]`.

It then tries to access the `value` field on both the actual and expected event, causing the Runtime Error above, since `Integer.value` is private.

Here is the quick fix:

```kt
val lazilyParsedIntPredicate = BiPredicate { v1: LazilyParsedNumber, v2: Integer -> v1.toInt() == v2.toInt() }

assertThat(actual.device)
  .usingRecursiveComparison()
  .withEqualsForFields(lazilyParsedIntPredicate, "totalRam", "logicalCpuCount")
  .ignoringFields("batteryLevel", "brightnessLevel")
  .isEqualTo(expected.device)
```

I'll refine this and update the PR :)

marco-saia-datadog: I have pushed [new changes](https://github.com/DataDog/dd-sdk-android/compare/a95c0f6578436b0d94f282e9cdebbe34c20aa4df..cf9f086ffbe358a04e2ca731019ecbe5c44d9cac), and added a common extension:

```kt
/**
 * Custom predicate for comparing Gson's LazilyParsedNumber with Integer fields.
 *
 * See [withGsonIntEqualsForFields].
 */
private val lazilyParsedIntPredicate = BiPredicate { v1: LazilyParsedNumber, v2: Int -> v1.toInt() == v2 }

/**
 * Adds support for comparing Gson's LazilyParsedNumber with Integer fields
 * during AssertJ recursive comparison.
 *
 * Gson deserializes Integer values as LazilyParsedNumber. During recursive
 * comparison, AssertJ may attempt reflective field comparison (e.g. Integer.value),
 * which fails because the field is private. This forces value-based comparison instead.
 */
fun<T : RecursiveComparisonAssert<T>> RecursiveComparisonAssert<T>.withGsonIntEqualsForFields(
    vararg fieldNames: String
): RecursiveComparisonAssert<T> {
    return this.withEqualsForFields(lazilyParsedIntPredicate, *fieldNames)
}
```

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/event/RumEventDeserializerTest.kt
```diff
@@ -151,7 +151,7 @@ internal class RumEventDeserializerTest {
 
         // WHEN
         val deserializedEvent = testedDeserializer.deserialize(serializedEvent) as LongTaskEvent
-
+        deserializedEvent.device?.totalRam
```
0xnm: seems like it is some debugging leftover

0xnm (commented):
lgtm, there is just one line leftover from debugging

0xnm (approved):

marco-saia-datadog:
@0xnm The checks are failing with:
```
e: DatadogLogGenerator.kt:228:9 Cannot find a parameter with this name: logicalCpuCount
e: DatadogLogGenerator.kt:229:9 Cannot find a parameter with this name: totalRam
e: DatadogLogGenerator.kt:230:9 Cannot find a parameter with this name: isLowRam
```
Seems like a schema issue, but shouldn't it be updated automatically?

**Update**

I have checked the pipeline configuration, and it seems like we rely on caching using `$CI_COMMIT_REF_SLUG` as the key.

Is it possible that the CI is still picking up the old schema, since the branch hasn't changed after I have updated `rum-events-format`?

Also, shouldn't we be using `$CI_COMMIT_SHA` as the cache key?

PR #3025, opened 2025-11-25
feat: State change notification for flags client

typotter:
### What does this PR do?

Implements state management for `FlagsClient` in the `dd-sdk-android-flags` module, enabling applications and OpenFeature providers to observe client lifecycle states.

**Key Changes:**
- Adds `FlagsClientState` enum with states: `NOT_READY`, `READY`, `RECONCILING`, `ERROR`
- Adds `FlagsStateListener` interface for receiving state change notifications
- Extends `FlagsClient` interface with state management methods: `getCurrentState()`, `addStateListener()`, `removeStateListener()`
- Implements `FlagsStateChannel` wrapper over `DDCoreSubscription` with semantic notification methods
- Wires state transitions through `EvaluationsManager` to `DatadogFlagsClient`
- Provides comprehensive unit test coverage for all state management components

### Motivation

This change is a prerequisite for implementing the OpenFeature provider's `observe()` method. The OpenFeature specification requires providers to emit events (PROVIDER_READY, PROVIDER_RECONCILING, PROVIDER_ERROR) when their state changes.

Without state tracking in the underlying `FlagsClient`, the OpenFeature provider has no mechanism to know when flags are loaded, when context changes begin/complete, or when errors occur.

**Benefits:**
- Enables OpenFeature provider compliance with the `observe()` specification
- Provides visibility into flag loading lifecycle for debugging and monitoring
- Allows applications to react to flag availability changes
- Fully backwards compatible - existing clients continue to work unchanged

### Additional Notes

**Design Decisions:**

1. **Left out STALE state:** The OpenFeature spec includes a `STALE` state for "cached flags may be outdated", but we removed it because there's no current mechanism to determine staleness. Added a note in `FlagsClientState` docs that this may be added in a future release.

2. **DDCoreSubscription with synchronization:** Used the existing `DDCoreSubscription` utility for listener management, wrapped with `synchronized` blocks to guarantee ordered delivery of state changes (critical for `NOT_READY → RECONCILING → READY` flows).

3. **FlagsStateChannel abstraction:** Created a wrapper class with semantic methods (`notifyReady()`, `notifyReconciling()`, etc.) to abstract DDCoreSubscription internals and improve code readability.

**State Transition Flow:**
```
setEvaluationContext()
    ↓
EvaluationsManager.notifyReconciling()
    ↓
[Background fetch...]
    ↓
Success: EvaluationsManager.notifyReady()
Failure: EvaluationsManager.notifyError()
    ↓
FlagsStateChannel → All registered listeners notified
```

**Thread Safety:**
- `getCurrentState()`: Lock-free atomic read
- `addStateListener()`/`removeStateListener()`: Thread-safe via `DDCoreSubscription`
- State notifications: Synchronized to guarantee ordered delivery

### DIFF
--- a/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
+++ b/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
@@ -24,6 +24,7 @@ import org.junit.jupiter.api.extension.ExtendWith
 import org.mockito.Mock
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.kotlin.any
 import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.doReturn
 import org.mockito.kotlin.mock
@@ -58,7 +59,7 @@ internal class FlagsTest {
     @BeforeEach
     fun `set up`() {
         whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
-        whenever(mockSdkCore.createSingleThreadExecutorService(org.mockito.kotlin.any())) doReturn mockExecutorService
+        whenever(mockSdkCore.createSingleThreadExecutorService(any())) doReturn mockExecutorService
 
         whenever(mockDatadogContext.clientToken) doReturn fakeClientToken
         whenever(mockDatadogContext.site) doReturn DatadogSite.US1


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -148,6 +151,35 @@ interface FlagsClient {
...
 
+    /**
+     * Gets the current state of this [FlagsClient].
+     *
+     * The state indicates whether the client is ready to evaluate flags, currently loading
+     * new flag values, or in an error state.
+     *
+     * @return The current [FlagsClientState].
+     */
+    fun getCurrentState(): FlagsClientState
```
0xnm: is this API supposed to be used by the customer or it should be internal API for the SDK needs?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -148,6 +151,35 @@ interface FlagsClient {
...
+
+    /**
+     * Registers a listener to receive state change notifications.
+     *
+     * The listener will be notified whenever the client's state changes (e.g., from NOT_READY
+     * to READY, or from READY to RECONCILING when the evaluation context changes).
```
0xnm:
```suggestion
     * The listener will be notified whenever the client's state changes (e.g., from [NOT_READY]
     * to [READY], or from [READY] to [RECONCILING] when the evaluation context changes).
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsStateListener.kt
```diff
@@ -0,0 +1,27 @@
...
+package com.datadog.android.flags
+
+import com.datadog.android.flags.model.FlagsClientState
+
+/**
+ * Listener interface for receiving state change notifications from a [FlagsClient].
+ *
+ * Implementations of this interface can be registered with a [FlagsClient] to receive
+ * callbacks whenever the client's state changes (e.g., from NOT_READY to READY, or
+ * from READY to RECONCILING when the evaluation context changes).
```
0xnm:
```suggestion
 * callbacks whenever the client's state changes (e.g., from [NOT_READY] to [READY], or
 * from [READY] to [RECONCILING] when the evaluation context changes).
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsStateListener.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+interface FlagsStateListener {
+    /**
+     * Called when the state of the [FlagsClient] changes.
+     *
+     * @param newState The new state of the client.
+     * @param error Optional error that caused the state change. This is typically provided
+     *              when transitioning to the [FlagsClientState.ERROR] state.
+     */
+    fun onStateChanged(newState: FlagsClientState, error: Throwable? = null)
```
0xnm: If error can come only with `FlagsClientState.ERROR` maybe it makes sense to make `FlagsClientState` a sealed class and have `error` as a property of `FlagsClientState.ERROR`?

Although if it can come with any state, we can keep it like that.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateChannel.kt
```diff
@@ -0,0 +1,90 @@
...
+     * Notifies all registered listeners of a state change.
+     *
+     * This method is thread-safe and guarantees ordered delivery. Multiple concurrent
+     * calls will be serialized to prevent out-of-order notifications.
+     *
+     * @param newState the new state
+     * @param error optional error associated with the state change
+     */
+    private fun notifyState(newState: FlagsClientState, error: Throwable?) {
+        synchronized(notificationLock) {
```
0xnm: I don't think it is needed, because `DDCoreSubscription` is using `CopyOnWriteArraySet` under the hood to manage listeners, but it is better to get an input from @aleksandr-gringauz.

Or maybe I don't understand right the problem of `out-of-order notifications` mentioned in the doc above.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/NoOpFlagsClientTest.kt
```diff
@@ -425,4 +427,43 @@ internal class NoOpFlagsClientTest {
...
+        val result = testedClient.getCurrentState()
+
+        // Then
+        assertThat(result).isEqualTo(FlagsClientState.ERROR)
+    }
+
+    @Test
+    fun `M do nothing W addStateListener()`() {
+        // Given
+        val mockListener = org.mockito.Mockito.mock(FlagsStateListener::class.java)
```
0xnm:
```suggestion
        val mockListener = mock(FlagsStateListener::class.java)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/NoOpFlagsClientTest.kt
```diff
@@ -425,4 +427,43 @@ internal class NoOpFlagsClientTest {
...
+
+        // Then
+        // No exception should be thrown, method should be no-op
+        verifyNoInteractions(mockListener)
+    }
+
+    @Test
+    fun `M do nothing W removeStateListener()`() {
+        // Given
+        val mockListener = org.mockito.Mockito.mock(FlagsStateListener::class.java)
```
0xnm:
```suggestion
        val mockListener = mock(FlagsStateListener::class.java)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
```diff
@@ -97,7 +100,8 @@ internal class DatadogFlagsClientTest {
                 rumIntegrationEnabled = true
             ),
             rumEvaluationLogger = mockRumEvaluationLogger,
-            processor = mockProcessor
+            processor = mockProcessor,
+            flagStateChannel = FlagsStateChannel(DDCoreSubscription.create())
```
0xnm: it is better to introduce `mockFlagsStateChannel` here and verify `addListener`/`removeListener` against it instead of having only `verifyNoInteractions(mockListener)`.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateChannelTest.kt
```diff
@@ -0,0 +1,179 @@
...
+        testedChannel.notifyReady()
+
+        // Then
+        verifyNoInteractions(mockListener)
+    }
+
+    @Test
+    fun `M notify all listeners W multiple listeners registered`() {
+        // Given
+        val mockListener2 = org.mockito.Mockito.mock(FlagsStateListener::class.java)
```
0xnm:
```suggestion
        val mockListener2 = mock(FlagsStateListener::class.java)
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateChannelTest.kt
```diff
@@ -0,0 +1,179 @@
...
+        testedChannel.addListener(mockListener)
+
+        // When
+        testedChannel.notifyNotReady()
+        testedChannel.notifyReconciling()
+        testedChannel.notifyReady()
+
+        // Then
+        val captor = argumentCaptor<FlagsClientState>()
+        verify(mockListener, org.mockito.kotlin.times(3)).onStateChanged(captor.capture(), org.mockito.kotlin.any())
```
0xnm:
```suggestion
        verify(mockListener, times(3)).onStateChanged(captor.capture(), any())
```

Although it is better to use `inOrder { verify(...) }` verification, this way no need to use the captor

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/evaluation/EvaluationsManagerTest.kt
```diff
@@ -218,4 +223,43 @@ internal class EvaluationsManagerTest {
...
+        whenever(mockAssignmentsDownloader.readPrecomputedFlags(publicContext)).thenReturn(mockResponse)
+        whenever(mockPrecomputeMapper.map(mockResponse)).thenReturn(expectedFlags)
+
+        // When
+        evaluationsManager.updateEvaluationsForContext(publicContext)
+
+        // Then
+        val inOrder = org.mockito.kotlin.inOrder(mockFlagsStateChannel)
+        inOrder.verify(mockFlagsStateChannel).notifyReconciling()
+        inOrder.verify(mockFlagsStateChannel).notifyReady()
```
0xnm:
```suggestion
        inOrder {
          verify(mockFlagsStateChannel).notifyReconciling()
          verify(mockFlagsStateChannel).notifyReady()
        }

```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/evaluation/EvaluationsManagerTest.kt
```diff
@@ -218,4 +223,43 @@ internal class EvaluationsManagerTest {
...
+
+        whenever(mockAssignmentsDownloader.readPrecomputedFlags(publicContext)).thenReturn(null)
+
+        // When
+        evaluationsManager.updateEvaluationsForContext(publicContext)
+
+        // Then
+        val inOrder = org.mockito.kotlin.inOrder(mockFlagsStateChannel)
+        inOrder.verify(mockFlagsStateChannel).notifyReconciling()
+        inOrder.verify(mockFlagsStateChannel).notifyError()
```
0xnm:
```suggestion
        inOrder {
          verify(mockFlagsStateChannel).notifyReconciling()
          verify(mockFlagsStateChannel).notifyError()
        }
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -43,9 +47,39 @@ internal class DatadogFlagsClient(
...
+    /**
+     * Updates the client state and notifies all registered listeners.
+     *
+     * This method is thread-safe and guarantees that listeners receive state changes in order.
+     * The notification is delegated to [flagStateChannel] which handles synchronization.
+     *
+     * @param newState The new state to transition to.
+     * @param error Optional error that caused the state change.
+     */
+    internal fun updateState(newState: FlagsClientState, error: Throwable? = null) {
```
jonathanmos: who will call this method (at the moment I only see it called from the tests)? Also, should the client modify it's internal state when flagStateChannel sends it a notification? eg evaluationsManager can set the state to ready/error, but this doesn't look like it will affect currentState.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateChannel.kt
aleksandr-gringauz: > Or maybe I don't understand right the problem of out-of-order notifications mentioned in the doc above.

Probably it means the following. Actually a good point.

1. Suppose we currently have 2 registered listeners `listener_1` and `listener_2`.
2. Suppose 2 threads call `notifyState(state_1)` and `notifyState(state_2)` concurrently.
3. Without the lock it is possible that the order of notifications will be like this `listener_1(state_2)`, `listener_1(state_1)` and `listener_2(state_1), listener_2(state_2)`. So `listener_1` will have `state_1` as the final state and `listener_2` will have `state_2`. Not good.
4. Introducing `synchronized` fixes this problem.

Do we need `synchronized` in this particular case? Currently all `notify<Something>` methods are called from a thread of a single-threaded executor. So - probably not. But I'm not sure what the plan for this code is. Maybe the plan is to call it from multiple threads.

We could add this `synchronized` directly into `DDCoreSubscriptionImpl.notifyListeners`. This should be the default behavior.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateChannel.kt
```diff
@@ -0,0 +1,90 @@
...
+ * Channel for managing and notifying state change listeners.
+ *
+ * This class wraps [DDCoreSubscription] and provides semantic methods for each state
+ * transition, abstracting the underlying notification mechanism. All notification methods
+ * are thread-safe and guarantee ordered delivery to listeners.
+ *
+ * @param subscription the underlying subscription for managing listeners
+ * @param notificationLock lock for ensuring ordered delivery of state notifications
+ */
+internal class FlagsStateChannel(
```
aleksandr-gringauz: I suggest we put the state itself inside this class and rename `FlagsStateChannel` to something like `FlagsStateManager`.

It should have the following API.
```
class FlagsStateManager {
    fun getCurrentState(): State
    fun setState(newState)
    fun addListener(listener)
    fun removeListener(listener)
}
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
typotter: Thanks Jonathan. I've wrapped up all the state mechanisms into the `FlagsClientStateManager` and tidied any direct state handling out of the client.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateChannel.kt
typotter: You are correct in your assessment that we want the state changes to be delivered in the same order across all listeners. Using the executor here (instead of synchronized) ensures all listeners are notified for each state change before the next state change is processed. Since we're also writing to the current state from the executor, we're guaranteed that the state can't be changed until the previous state change has completed notifying all listeners

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
typotter: Current state is exposed to the user. In the new observable state, the user can either collect from a `Flow` or use listeners/`getCurrentState()`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsStateListener.kt
typotter: sealed class makes sense here, thanks!

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
typotter: done.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateChannel.kt
typotter: agreed. I've also added `Flow` for a more idiomatic observable interface.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/StateObservable.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Returns the current state synchronously.
+     *
+     * This method is safe to call from any thread and does not require coroutines.
+     * Suitable for Java users or quick state checks.
+     *
+     * @return The current [FlagsClientState].
+     */
+    fun getCurrentState(): FlagsClientState
```
aleksandr-gringauz, line 43: If you expose `StateFlow`, there is no need to expose a getter, because there is `StateFlow.value` that returns the current value.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -148,6 +150,28 @@ interface FlagsClient {
...
+     * val current = client.state.getCurrentState()
+     *
+     * // Reactive Flow
+     * client.state.flow.collect { state -> /* ... */ }
+     *
+     * // Callback
+     * client.state.addListener(listener)
+     * ```
+     */
+    val state: StateObservable
```
aleksandr-gringauz, line 169: Up until now in our SDK we have never used coroutines anywhere except in modules dedicated specifically to integrating with coroutines (`dd-sdk-android-rum-coroutines`).

So this change should be discussed in detail with the Android team.

Also if we were to introduce coroutines in `dd-sdk-android-flags`, we could have written the code a lot differently, without using java executors for multi-threading, but rather directly using coroutines.

But again - this is a big thing, we should discuss it before moving forward.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -0,0 +1,103 @@
...
+     * MutableStateFlow itself is thread-safe.
+     */
+    private val _stateFlow = MutableStateFlow<FlagsClientState>(FlagsClientState.NotReady)
+
+    /**
+     * The current state of the client.
+     * Thread-safe: synchronized to ensure atomicity with flow updates.
+     */
+    @Volatile
+    private var currentState: FlagsClientState = FlagsClientState.NotReady
```
aleksandr-gringauz, line 41: this field is useless, current state can be read from `_stateFlow` using `.value`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/StateObservable.kt
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Registers a listener to receive state change notifications.
+     *
+     * The listener will immediately receive the current state upon registration,
+     * then be notified of all future state changes. Suitable for Java users or
+     * when Flow is not available.
+     *
+     * @param listener The [FlagsStateListener] to register.
+     */
+    fun addListener(listener: FlagsStateListener)
```
aleksandr-gringauz, line 53: If you are exposing `StateFlow` the listener-style API is redundant, the users can subscribe to `StateFlow` directly using `stateFlow.collect`.

aleksandr-gringauz (changes requested):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
0xnm: Yes, adding coroutines adds significant overhead to the transitive dependencies:

> Adding coroutines has increased the dependencies to 5mb from 3

I don't have a strong opinion against them _in this module_, because it is almost 2026 and it is a new module. However, it can be still that flags module is used by the customers who don't have coroutines in their codebase yet, then for them it will be quite a significant increase of the APK size.

But I agree that the implementation code should be consistent: if we use coroutines, it should be idiomatic.

typotter: Thanks for weighing in! I am going to remove coroutines from this module. The OpenFeature SDK includes coroutines so it makes more sense to add coroutines in our `flags-openfeature` module and omit them here for now. We can revisit in the future if coroutines are desirable in the core flags module with something like `dd-sdk-android-flags-coroutines`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/StateObservable.kt
typotter: Yes. this was built redundant. Removing Flow

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/StateObservable.kt
typotter: Built this redundant to avoid use of coroutines/Flow where desired, but I'm pulling out coroutines now.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
typotter: ack

detekt_custom_safe_calls.yml
```diff
@@ -1081,6 +1083,8 @@ datadog:
       - "kotlinx.coroutines.flow.FlowCollector.emit(kotlin.Any?)"
       - "kotlinx.coroutines.flow.FlowCollector(kotlin.coroutines.SuspendFunction1)"
       - "kotlinx.coroutines.flow.flow(kotlin.coroutines.SuspendFunction1)"
+      - "kotlinx.coroutines.flow.MutableStateFlow(com.datadog.android.flags.model.FlagsClientState)"
+      - "kotlinx.coroutines.flow.MutableStateFlow.asStateFlow()"
```
0xnm: since coroutines are now removed from this module, this is not relevant

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -148,6 +150,28 @@ interface FlagsClient {
...
+     * - Reactive: [StateObservable.flow] for coroutine-based updates (Kotlin)
+     * - Callback: [StateObservable.addListener] for traditional observers (Java-friendly)
+     *
+     * Example:
+     * ```kotlin
+     * // Synchronous
+     * val current = client.state.getCurrentState()
+     *
+     * // Reactive Flow
+     * client.state.flow.collect { state -> /* ... */ }
```
0xnm: this comment is outdated, there is no `Flow` option

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/NoOpFlagsClient.kt
```diff
@@ -32,6 +35,12 @@ internal class NoOpFlagsClient(
     private val logWithPolicy: LogWithPolicy
 ) : FlagsClient {
 
+    override val state: StateObservable = object : StateObservable {
+        override fun getCurrentState(): FlagsClientState = FlagsClientState.Error(null)
+        override fun addListener(listener: FlagsStateListener) { /* no-op */ }
+        override fun removeListener(listener: FlagsStateListener) { /* no-op */ }
```
0xnm:
```suggestion
        override fun addListener(listener: FlagsStateListener) = Unit
        override fun removeListener(listener: FlagsStateListener) = Unit
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/repository/DefaultFlagsRepositoryTest.kt
```diff
@@ -184,4 +184,116 @@ internal class DefaultFlagsRepositoryTest {
...
+    @Test
+    fun `M return expected value W hasFlags() { for various states }`(forge: Forge) {
+        data class TestCase(val given: () -> Unit, val then: Boolean)
+
+        val testCases = listOf(
+            TestCase(
+                given = { /* no state set */ },
+                then = false
+            ),
+            TestCase(
```
0xnm: it is better to split this into different standalone tests

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/repository/DefaultFlagsRepositoryTest.kt
```diff
@@ -184,4 +184,116 @@ internal class DefaultFlagsRepositoryTest {
...
+
+            // Then
+            assertThat(result).isEqualTo(testCase.then)
+        }
+    }
+
+    @Test
+    fun `M not block W hasFlags() { persistence still loading }`() {
+        // Given
+        val startTime = System.currentTimeMillis()
```
0xnm: probably it is not a right place to start counting time, it is better to move it right before `slowRepository.hasFlags()` call

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
```diff
@@ -59,7 +58,7 @@ internal class FlagsTest {
     @BeforeEach
     fun `set up`() {
         whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
-        whenever(mockSdkCore.createSingleThreadExecutorService(FLAGS_CLIENT_EXECUTOR_NAME)) doReturn
+        whenever(mockSdkCore.createSingleThreadExecutorService(org.mockito.kotlin.any())) doReturn
```
0xnm:
```suggestion
        whenever(mockSdkCore.createSingleThreadExecutorService(any())) doReturn
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
```diff
@@ -1292,4 +1310,37 @@ internal class DatadogFlagsClientTest {
     }
 
     // endregion
+
+    // region State Management
+
+    @Test
+    fun `M delegate to state manager W state_addListener()`() {
+        // Given
+        val mockListener = mock(FlagsStateListener::class.java)
```
0xnm:
```suggestion
        val mockListener = mock<FlagsStateListener>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
```diff
@@ -1292,4 +1310,37 @@ internal class DatadogFlagsClientTest {
...
+
+        // Then
+        // Verify delegation to state manager
+        verify(mockFlagsStateManager).addListener(mockListener)
+    }
+
+    @Test
+    fun `M delegate to state manager W state_removeListener()`() {
+        // Given
+        val mockListener = mock(FlagsStateListener::class.java)
```
0xnm:
```suggestion
        val mockListener = mock<FlagsStateListener>()
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+    }
+
+    // endregion
+
+    // region addListener / removeListener
+
+    @Test
+    fun `M notify listener W addListener() and notify`() {
```
0xnm: this test is already included by the parametrized test above

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -0,0 +1,169 @@
...
+        // Verify initial state was emitted
+        verify(mockListener).onStateChanged(FlagsClientState.NotReady)
+
+        testedManager.removeListener(mockListener)
+
+        // When
+        testedManager.updateState(FlagsClientState.Ready)
+
+        // Then - no further notifications after removal
+        org.mockito.kotlin.verifyNoMoreInteractions(mockListener)
```
0xnm:
```suggestion
        verifyNoMoreInteractions(mockListener)
```

typotter: applied

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
typotter: applied

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
typotter: applied

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/NoOpFlagsClient.kt
typotter: applied

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
typotter: fixed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
typotter: applied

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/repository/DefaultFlagsRepositoryTest.kt
typotter: thanks. good catch

detekt_custom_safe_calls.yml
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
typotter: fixed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
typotter: fixed

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/FlagsTest.kt
```diff
@@ -59,8 +58,7 @@ internal class FlagsTest {
     @BeforeEach
     fun `set up`() {
         whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
-        whenever(mockSdkCore.createSingleThreadExecutorService(FLAGS_CLIENT_EXECUTOR_NAME)) doReturn
-            mockExecutorService
+        whenever(mockSdkCore.createSingleThreadExecutorService(org.mockito.kotlin.any())) doReturn mockExecutorService
```
0xnm:
```suggestion
        whenever(mockSdkCore.createSingleThreadExecutorService(any())) doReturn mockExecutorService
```

0xnm (dismissed):

PR #3026, opened 2025-11-25
RUM-12330: Handle multi SDK instance in profiling

ambushwork:
### What does this PR do?

This PR handles multi SDK instance in profiling by bringing following changes:
* `PerfettoProfiler` hold a map of <InstanceName, ProfilerCallback> to make sure the result can be sent back to the correct instance.
* `ProfilingStorage` save the name of the instance on which the profiling feature is enabled, and when profiler is started in content provider, the instance name can be used to indicate the profiling is launched by which feature.

### Motivation

RUM-12330


features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -219,28 +248,162 @@ class PerfettoProfilerTest {
...
+            )
+
+        val successResult = mock<ProfilingResult> {
+            on { errorCode } doReturn ProfilingResult.ERROR_NONE
+            on { resultFilePath } doReturn fakePath
+        }
+        callbackCaptor.firstValue.accept(successResult)
+        val captor = argumentCaptor<PerfettoResult>()
+        verifyNoInteractions(mockOtherProfilerCallback)
+        verify(mockProfilerCallback).onSuccess(captor.capture())
```
0xnm: there is no need for the `captor` here, because we don't read what was captured

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -63,35 +66,52 @@ internal class PerfettoProfiler(
             val duration = endTime - profilingStartTime
             if (result.errorCode == ProfilingResult.ERROR_NONE) {
                 result.resultFilePath?.let {
-                    onProfilingSuccess?.invoke(
-                        PerfettoResult(
-                            start = profilingStartTime,
-                            end = endTime,
-                            resultFilePath = it
+                    runningInstance.get()?.let { instanceName ->
```
aleksandr-gringauz: Question.

Suppose there are 2 instances (`sdk_1` and `sdk_2`) of the SDK in the app and they both call `Profiling.enable`. `ProfilingStorage` will have `sdk_2` string stored in shared preferences.

On the next launch profiling will be started for `sdk_2` and only `ProfilingFeature` of `sdk_2` will be notified with the profiling result even though `Profiling.enable` was called on `sdk_1` too.

Is it ok? Shouldn't we notify all callbacks in `callbackMap`?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -42,13 +43,15 @@ internal class PerfettoProfiler(
...
+    // This flag represents which instance of this class is working for.
+    private val runningInstance: AtomicReference<String?> = AtomicReference(null)
 
     private var profilingStartTime = 0L
 
     override var internalLogger: InternalLogger? = null
 
-    override var onProfilingSuccess: ((PerfettoResult) -> Unit)? = null
+    // Map of <InstanceName, ProfilerCallback>
+    private val callbackMap: MutableMap<String, ProfilerCallback> = mutableMapOf()
```
0xnm: we probably can assume that it can be mutated from different threads, so it needs to be thread-safe

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -63,35 +66,52 @@ internal class PerfettoProfiler(
             val duration = endTime - profilingStartTime
             if (result.errorCode == ProfilingResult.ERROR_NONE) {
                 result.resultFilePath?.let {
-                    onProfilingSuccess?.invoke(
-                        PerfettoResult(
-                            start = profilingStartTime,
-                            end = endTime,
-                            resultFilePath = it
+                    runningInstance.get()?.let { instanceName ->
+                        val callback = callbackMap[instanceName]
```
0xnm: I'm not sure this approach scales, because we effectively deliver result only to the single instance.

Let's imagine 2 profiling situations: application launch and custom profiling.

In application launch profiling it is not the customer who controls start and stop, start is the same for all SDK instances, same for the stop (TTID is not bound to the SDK setup). It means that application launch profiling result can be delivered to _all_ instances running profiling and who requested it before, so it is one-to-many (we run a single profiling session).

The situation is different for custom profiling: start/stop calls are specific to the SDK instance and it means that in this case we need to spin off profiling session which is unique for the particular SDK instance and we should deliver result only to this instance, so it is one-to-one.

The API above also creates concurrency issue, because different SDK instances may request profiling and it is not clear which one will effectively get it.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
ambushwork: Yes, it will happen as you described, and that's this PR is designed for.
if we notify all the callbacks, then both `sdk_1` and `sdk_2` will generate a profiling event with the same profiling results, upload to the backend, which is not we want.
The idea is that if multi instances enable Profiling, then the last one will count. (although we should suggest not to do that).

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
ambushwork: This PR is implemented based one-to-one relationship for both of these scenarios that you described.
So the main disagreement is that if the profiling for application launch should be broadcast to all the SDK instances,  my thought is that we should avoid uploading single profile as multiple events, because the content of profile itself is exactly the same. And here the concurrency is expected, the last who overrides the SP will be counted for next profiling. I am not sure how much this will affect the client (I don't see a lot of reason to enable profiling besides the app level), we can make the doc clear.

0xnm: > because the content of profile itself is exactly the same

Yes, that is why we can allow to broadcast it, different SDK instances not necessarily belong to the same org.

> And here the concurrency is expected, the last who overrides the SP will be counted for next profiling.

I don't think it is a right approach, but rather a confusing one. Nothing prevents us to remove this limitation for the application launch profiling.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStorage.kt
```diff
@@ -15,16 +15,30 @@ internal object ProfilingStorage {
...
 
-    internal fun setProfilingFlag(appContext: Context) {
-        getStorage(appContext).putBoolean(KEY_PROFILING_ENABLED, true)
+    internal fun addProfilingFlag(appContext: Context, sdkInstanceName: String) {
+        getStorage(appContext).apply {
+            val oldValue = getStringSet(KEY_PROFILING_ENABLED, emptySet()) ?: emptySet()
+            val newSet = mutableSetOf<String>()
+            newSet.addAll(oldValue)
+            newSet.add(sdkInstanceName)
+            putStringSet(KEY_PROFILING_ENABLED, newSet)
```
aleksandr-gringauz: very nit:

```
val oldValue = getStringSet(KEY_PROFILING_ENABLED, emptySet()).orEmpty()
val newSet = oldValue + sdkInstanceName
putStringSet(KEY_PROFILING_ENABLED, newSet)
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStorage.kt
```diff
@@ -15,16 +15,30 @@ internal object ProfilingStorage {
...
+        }
     }
 
-    internal fun isProfilingEnabled(appContext: Context): Boolean {
-        return getStorage(appContext).getBoolean(KEY_PROFILING_ENABLED, false)
+    internal fun getProfilingEnabledInstanceName(appContext: Context): Set<String> {
```
aleksandr-gringauz: `getProfilingEnabledInstanceName` -> `getProfilingEnabledInstanceNames`

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -45,24 +45,27 @@ internal class ProfilingFeature(
     override fun onInitialize(appContext: Context) {
         profiler.apply {
             this.internalLogger = sdkCore.internalLogger
-            this.onProfilingSuccess = { result ->
+            registerProfilingCallback(sdkCore.name) { result ->
                 dataWriter.write(
```
aleksandr-gringauz: IIUC If this `data.write` occurs for the profile that ended because TTID happened in a **different** sdk instance, `ttidEvent` here might still be `null`.

ambushwork: In theory TTID for each sdk instance should arrive nearly the same time, so when profiling is stopped by another instance, probably the others will also have TTID event. But if we have null TTID event for some cases, the profile will still be sent, but it will miss some info in the tag

aleksandr-gringauz: > In theory TTID for each sdk instance should arrive nearly the same time, so when profiling is stopped by another instance, probably the others will also have TTID event.

Yes, but it all depends on how fast ProfilingManager calls its callback with the result. If it is really fast, the TTID for "another" instance might not have arrived yet.

https://github.com/DataDog/dd-sdk-android/commit/48952f00419b2c0bbbeb2f682d6e174b9b71b532 - you could do something like this.

ambushwork: updated.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/data/PreferencesStorage.kt
```diff
@@ -34,6 +34,16 @@ interface PreferencesStorage {
      */
     fun putString(key: String, value: String)
 
+    /**
+     * Get a string set value from the storage.
+     */
+    fun getStringSet(key: String, defaultValue: Set<String>? = null): Set<String>?
```
0xnm: in case of collections it is better to avoid having `null` value and use empty collection instead

```suggestion
    fun getStringSet(key: String, defaultValue: Set<String> = emptySet()): Set<String>
```

dd-sdk-android-internal/src/test/java/com/datadog/android/internal/data/SharedPreferencesStorageTest.kt
```diff
@@ -149,6 +150,34 @@ internal class SharedPreferencesStorageTest {
         assertThat(result).isEqualTo(fakeValue)
     }
 
+    @Test
+    fun `M persist data W putStringSet`(
```
0xnm, line 157: we added `ClassCastException` handling behaviour, so we need to add tests as well

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStorage.kt
```diff
@@ -15,16 +15,28 @@ internal object ProfilingStorage {
...
     }
 
-    internal fun removeProfilingFlag(appContext: Context) {
-        getStorage(appContext).remove(KEY_PROFILING_ENABLED)
+    internal fun removeProfilingFlag(appContext: Context, sdkInstanceNames: Set<String>) {
+        getStorage(appContext).apply {
+            val oldValue = getStringSet(KEY_PROFILING_ENABLED) ?: emptySet()
+            val newSet = mutableSetOf<String>()
+            newSet.addAll(oldValue)
+            val removed = newSet.removeAll(sdkInstanceNames)
```
0xnm: nit: can be even shorter

```suggestion
            val value = getStringSet(KEY_PROFILING_ENABLED).orEmpty().toMutableSet()
            val removed = value.removeAll(sdkInstanceNames)
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -45,24 +45,29 @@ internal class ProfilingFeature(
     override fun onInitialize(appContext: Context) {
         profiler.apply {
             this.internalLogger = sdkCore.internalLogger
-            this.onProfilingSuccess = { result ->
-                dataWriter.write(
-                    profilingResult = result,
-                    ttidEvent = ttidEvent
-                )
+            registerProfilingCallback(sdkCore.name) { result ->
```
0xnm, line 51: This approach will miss some TTID events.

Let's imagine we have 2 SDK instances running: A and B. Instance A sends TTID event to Profiling feature resulting in Profiler stop and this callback invocation for the Feature A **and** for the Feature B. We have TTID event A which will be bundled with profiling recording.

Now, instance B also sends TTID event, but because of concurrency (it is send from another thread, from the pool of instance B), it arrives _after_ profiler is stopped and callback is called (it was called for TTID event A), meaning TTID event B will never be bundled with profiling result. Unless multiple `stopSignal.cancel()` calls lead to the multiple callback invocations, is it the case?

I think better approach would be to store app launch file path recording in the Profiler once we get it, and when TTID event arrives just bundle it. This way we don't store TTID in a property, but app launch recording file path and don't lose anything.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/Profiler.kt
```diff
@@ -8,19 +8,20 @@ package com.datadog.android.profiling.internal
...
     var internalLogger: InternalLogger?
 
-    var onProfilingSuccess: ((PerfettoResult) -> Unit)?
+    fun start(appContext: Context, sdkInstanceNames: Set<String>)
 
-    fun start(appContext: Context)
+    fun stop(sdkInstanceName: String)
 
-    fun stop()
+    fun isRunning(sdkInstanceName: String): Boolean
```
0xnm, line 22: Why do we stop and check for `isRunning` for a given SDK instance name? Given that we have a single profiler instance and single profiling session for the application launch.

I think that at least for the application launch we should have a set of SDK instances only for the `start` call (maybe we should rename it to the `startApplicationLaunch` profiling for the clarity? because once we have custom profiling the approach will be different).

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -76,22 +80,41 @@ internal class PerfettoProfiler(
...
             requestProfiling(appContext, requestBuilder.build(), profilingExecutor, resultCallback)
         }
     }
 
-    override fun stop() {
-        profilingStarted.set(false)
-        stopSignal.cancel()
+    override fun stop(sdkInstanceName: String) {
+        if (runningInstances.get().contains(sdkInstanceName)) {
+            stopSignal.cancel()
```
0xnm, line 101: Is it safe to call `stopSignal.cancel()` multiple times? shouldn't we just call it only once? Normally we should stop profiling session only once, so I'm not sure what is the point of having `sdkInstanceName` in this call.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
aleksandr-gringauz: @ambushwork Did you change anything in `ProflingFeature.kt`? I don't see how the problem is fixed.

aleksandr-gringauz: https://github.com/DataDog/dd-sdk-android/pull/3026#discussion_r2567540883 - I was describing the same problem as @0xnm is describing in this thread.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
ambushwork: it doesn't harm to call `stopSignal.cancel()` multiple times, `CancellationSignal` has state check underneath.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/Profiler.kt
ambushwork: Imagine we have instance A and instance B, instance A has set flag for profiling in previous session but B not, and in current session, both A & B enable Profiling, then they will both receive TTID and try to send stop signal and try to update the feature context with the profiling running status. In this case we need to prevent instance B to stop profiler and seeing the profiler is running for B

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
ambushwork: updated with @aleksandr-gringauz 's proposal

dd-sdk-android-internal/src/test/java/com/datadog/android/internal/data/SharedPreferencesStorageTest.kt
ambushwork: good catch, updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
ambushwork: updated now

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -42,13 +44,15 @@ internal class PerfettoProfiler(
 
     private val resultCallback: Consumer<ProfilingResult>
 
-    private val profilingStarted = AtomicBoolean(false)
+    // This flag represents which instance of this class is working for.
+    private val runningInstances: AtomicReference<Set<String>> = AtomicReference(emptySet())
```
0xnm, line 48: minor: this doesn't prevent concurrent access to the same instance `Set<String>`. If it is possible to have such access, we need to use `Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())`.

`AtomicReference` is used below mainly for `compareAndSet`, but if we use concurrent set we probably don't need `AtomicReference` (not quite sure here though without more precise checks).

ambushwork: Currently this `set` is only modified integrately, not partially (maybe it's the case in the future), so it doesn't have a concurrent issue of the collection

PR #3027, opened 2025-11-26
PANA-5027: Support svgs with navigation3

jonathanmos:
### What does this PR do?
Navigation3 bumps the compose-ui dependency to 1.9. Whereas previously we could get argb8888 from vectorDrawables with version 1.5, with 1.9 we get alpha8. In this pr, we convert the alpha8 mask into a grayscale bitmap, using the alpha transparency to set the level of gray. This means we lose color information but we have a performant method to keep supporting svgs.

To improve performance, we don't want to generate bitmaps from alpha8 on every recomposition. We therefore sample 16 horizontal and vertical or 256 alpha value points from every alpha8 in order to create a fast and accurate signature and use that to determine whether we have seen that svg before and whether a bitmap should be generated. This method is resistant to instance or id change of the svg, doesn't rely on unreliable internal metadata, and should have a minimal risk of collisions. The signature to resourceId is maintained in an lrucache similar to what exists for native bitmaps.

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
@@ -248,6 +248,7 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
 
     override fun registerCallbacks() {
         appContext.registerActivityLifecycleCallbacks(sessionReplayLifecycleCallback)
+        resourceResolver.registerCallbacks()
     }
 
     override fun unregisterCallbacks() {
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/Alpha8ResourceCache.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/Alpha8ResourceCache.kt
@@ -32,10 +32,10 @@ internal data class Alpha8CacheKey(
 
 internal class DefaultAlpha8ResourceCache(
     private val signatureGenerator: BitmapSignatureGenerator,
-    private val cache: LruCache<Alpha8CacheKey, ByteArray> =
-        object : LruCache<Alpha8CacheKey, ByteArray>(MAX_CACHE_MEMORY_SIZE_BYTES) {
-            override fun sizeOf(key: Alpha8CacheKey, value: ByteArray): Int {
-                return value.size
+    private val cache: LruCache<Alpha8CacheKey, String> =
+        object : LruCache<Alpha8CacheKey, String>(MAX_CACHE_MEMORY_SIZE_BYTES) {
+            override fun sizeOf(key: Alpha8CacheKey, value: String): Int {
+                return (value.length * 2) + STRING_OBJECT_OVERHEAD_BYTES
             }
         }
 ) : Alpha8ResourceCache {
@@ -46,11 +46,11 @@ internal class DefaultAlpha8ResourceCache(
     }
 
     override fun get(key: Alpha8CacheKey): String? {
-        return cache[key]?.toString(Charsets.UTF_8)
+        return cache[key]
     }
 
     override fun put(key: Alpha8CacheKey, resourceId: String) {
-        cache.put(key, resourceId.toByteArray(Charsets.UTF_8))
+        cache.put(key, resourceId)
     }
 
     @Suppress("DEPRECATION")
@@ -90,5 +90,6 @@ internal class DefaultAlpha8ResourceCache(
 
     internal companion object {
         const val MAX_CACHE_MEMORY_SIZE_BYTES = 4 * 1024 * 1024 // 4MB
+        const val STRING_OBJECT_OVERHEAD_BYTES = 40
     }
 }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
@@ -46,7 +46,7 @@ internal class ResourceResolver(
     )
 ) {
 
-    init {
+    internal fun registerCallbacks() {
         applicationContext.registerComponentCallbacks(alpha8ResourceCache)
     }
 
--- a/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/DefaultAlpha8ResourceCacheTest.kt
+++ b/features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/DefaultAlpha8ResourceCacheTest.kt
@@ -21,6 +21,7 @@ import org.mockito.Mock
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
 import org.mockito.kotlin.any
+import org.mockito.kotlin.never
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.mockito.quality.Strictness
@@ -36,7 +37,7 @@ internal class DefaultAlpha8ResourceCacheTest {
     private lateinit var mockSignatureGenerator: BitmapSignatureGenerator
 
     @Mock
-    private lateinit var mockCache: LruCache<Alpha8CacheKey, ByteArray>
+    private lateinit var mockCache: LruCache<Alpha8CacheKey, String>
 
     @Mock
     private lateinit var mockBitmap: Bitmap
@@ -92,7 +93,7 @@ internal class DefaultAlpha8ResourceCacheTest {
     fun `M return cached resourceId W get { key found in cache }`() {
         // Given
         val key = Alpha8CacheKey(100, 100, fakeSignature)
-        whenever(mockCache[key]).thenReturn(fakeResourceId.toByteArray(Charsets.UTF_8))
+        whenever(mockCache[key]).thenReturn(fakeResourceId)
 
         // When
         val result = testedCache.get(key)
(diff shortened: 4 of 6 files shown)


features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/BitmapCreationHelper.kt
```diff
@@ -0,0 +1,87 @@
...
+
+internal interface BitmapCreationHelper {
+    fun createBitmap(width: Int, height: Int, config: Bitmap.Config): Bitmap?
+    fun drawBitmap(destination: Bitmap, source: Bitmap): Bitmap?
+}
+
+internal class DefaultBitmapCreationHelper(
+    private val logger: InternalLogger =
+        (Datadog.getInstance() as? FeatureSdkCore)?.internalLogger
+            ?: InternalLogger.UNBOUND
```
0xnm: it is possible to have `internalLogger` from the particular SDK instance?

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/BitmapCreationHelper.kt
```diff
@@ -0,0 +1,87 @@
...
+        // IllegalStateException shouldn't happen - we check for mutability
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        val canvas = Canvas(destination)
+
+        // Fill with Black (Transparency -> Black pixel)
+        canvas.drawColor(android.graphics.Color.BLACK)
+        val paint = Paint()
+        // Draw content in White (Opacity -> White pixel)
+        // Intermediate alpha levels will blend to create shades of gray (Grayscale)
+        paint.color = android.graphics.Color.WHITE
```
0xnm: I don't quite get this and comments. As I understand, drawing black and white for `Bitmap.Config.ALPHA_8` somehow creates grayscale since this config contains only alpha information.

But this method doesn't check that we get `Bitmap.Config.ALPHA_8` config for the input.

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/BitmapCreationHelper.kt
```diff
@@ -0,0 +1,87 @@
...
+            null
+        } catch (@Suppress("TooGenericExceptionCaught") e: RuntimeException) {
+            logger.log(
+                target = InternalLogger.Target.MAINTAINER,
+                level = InternalLogger.Level.ERROR,
+                messageBuilder = { BITMAP_DRAWING_FAILURE },
+                throwable = e
+            )
+            null
+        } catch (e: OutOfMemoryError) {
```
ambushwork: Why we catch OOM exception specifically here?

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/BitmapCreationHelper.kt
```diff
@@ -0,0 +1,87 @@
...
+    fun createBitmap(width: Int, height: Int, config: Bitmap.Config): Bitmap?
+    fun drawBitmap(destination: Bitmap, source: Bitmap): Bitmap?
+}
+
+internal class DefaultBitmapCreationHelper(
+    private val logger: InternalLogger =
+        (Datadog.getInstance() as? FeatureSdkCore)?.internalLogger
+            ?: InternalLogger.UNBOUND
+) : BitmapCreationHelper {
+    override fun createBitmap(width: Int, height: Int, config: Bitmap.Config): Bitmap {
```
ambushwork: It seems this function only calls `Bitmap.createBitmap` and return it, and in the callsite it directly chains with `drawBitmap`, so why can't we have a single function create inline this one?

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/BitmapCreationHelper.kt
jonathanmos: the key is in this line:
```
canvas.drawBitmap(source, 0f, 0f, paint)
```
because `source` is the alpha8 mask. When the paint is applied to this mask the alpha channel will cause the shades. We only reach this method through createBitmapFromAlpha8 in semanticsUtil, so we know that the source will necessarily be an alpha8 mask. I'll change the name of the class or method to make it clear that it is only for alpha8 processing.

features/dd-sdk-android-session-replay-compose/src/main/kotlin/com/datadog/android/sessionreplay/compose/internal/utils/BitmapCreationHelper.kt
jonathanmos: 👍  we generally don't catch this type of error. I'll remove it

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
```diff
@@ -248,6 +252,7 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
 
     override fun unregisterCallbacks() {
         appContext.unregisterActivityLifecycleCallbacks(sessionReplayLifecycleCallback)
+        resourceResolver.unregisterCallbacks()
```
0xnm, line 256: design nit: this would imply that somewhere in this class it should be `registerCallbacks()` call, but I guess it is done in the `ResourceResolver` constructor call?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/Alpha8ResourceCache.kt
```diff
@@ -0,0 +1,94 @@
...
+
+internal data class Alpha8CacheKey(
+    val width: Int,
+    val height: Int,
+    val signature: Long
+)
+
+internal class DefaultAlpha8ResourceCache(
+    private val signatureGenerator: BitmapSignatureGenerator,
+    private val cache: LruCache<Alpha8CacheKey, ByteArray> =
```
0xnm: cannot it be just `LruCache<Alpha8CacheKey, ByteArray>`? this way we would avoid making `toString` / `toByteArray` conversions.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
```diff
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit
 
 @Suppress("TooManyFunctions")
 internal class ResourceResolver(
```
0xnm, line 30: minor: It could also implement `Closeable` interface, where `close` method is used to release allocated resources. This way you don't need to call `unregisterCallbacks` from `SessionReplayRecorder`, but `close`, which is a bit more idiomatic.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/DefaultAlpha8ResourceCacheTest.kt
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Test
+    fun `M do nothing W onTrimMemory { TRIM_MEMORY_UI_HIDDEN }`() {
+        // When
+        testedCache.onTrimMemory(ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN)
+
+        // Then
+        verify(mockCache, org.mockito.kotlin.never()).evictAll()
+        verify(mockCache, org.mockito.kotlin.never()).trimToSize(any())
```
0xnm:
```suggestion
        verify(mockCache, never()).evictAll()
        verify(mockCache, never()).trimToSize(any())
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourceResolver.kt
jonathanmos: I've moved the registration out of init so that it is called by SessionReplayRecorder and is now symmetrical

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/Alpha8ResourceCache.kt
```diff
@@ -0,0 +1,94 @@
...
+    val height: Int,
+    val signature: Long
+)
+
+internal class DefaultAlpha8ResourceCache(
+    private val signatureGenerator: BitmapSignatureGenerator,
+    private val cache: LruCache<Alpha8CacheKey, String> =
+        object : LruCache<Alpha8CacheKey, String>(MAX_CACHE_MEMORY_SIZE_BYTES) {
+            override fun sizeOf(key: Alpha8CacheKey, value: String): Int {
+                return value.length
```
0xnm:
```suggestion
                return value.length * 2
```

this needs to be multiplied by 2, because in UTF-16 you one or two 16-bits units (so 2 or 4 bytes).

0xnm: Maybe it is worth also adding object header size, which is around ~40 bytes

jonathanmos: good catch - I'll add this

PR #3049, opened 2025-12-11
FFL-1460 FlagsClient changes to accommodate Flags RN SDK sync flag evals

yugisu-flux:
### What does this PR do?

This PR introduces new exposed "internal" APIs supporting the upcoming Flags SDK for React Native.

Changes:
- New "internal" method for retrieving a snapshot of all precomputed flags to FlagsClient
- New "internal" method for tracking a flag evaluation within a certain EvaluationContext
- Add a `_FlagsInternalProxy` class for `FlagsClient`
- Add an `UnparsedFlag` interface to prevent `PrecomputedFlag` from being exposed to the public

"internal" APIs = exposed APIs that are added purely for consumption within the RN SDK. This implementation follows the example of `_RumInternalProxy`.

### Motivation

The motivation for this work is to allow the upcoming Flags RN SDK to retrieve the complete feature flags state snapshot from the JS side at SDK init time ([FFL-1460](https://datadoghq.atlassian.net/browse/FFL-1460)).

### Additional Notes

I've been developing these changes using the example application in the dd-sdk-reactnative repo (by leveraging composite builds)

Sibling PRs in the React Native and iOS repos:
- https://github.com/DataDog/dd-sdk-reactnative/pull/1064
- https://github.com/DataDog/dd-sdk-ios/pull/2604

[FFL-1460]: https://datadoghq.atlassian.net/browse/FFL-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagValueConverter.kt
```diff
@@ -52,13 +52,18 @@ internal object FlagValueConverter {
             @Suppress("UNCHECKED_CAST")
             val result: T? = when (targetType) {
                 Boolean::class -> variationValue.lowercase(Locale.US).toBooleanStrictOrNull() as? T
+
```
0xnm: this is a redundant change, you need to run KtLint version `0.50.0`:

```
ktlint-0.50.0 -F "**/*.kt" "**/*.kts" '!**/build/generated/**' '!**/build/kspCaches/**'
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
```diff
@@ -150,6 +150,15 @@ interface FlagsClient {
...
     fun <T : Any> resolve(flagKey: String, defaultValue: T): ResolutionDetails<T>
 
+    /**
+     * For Datadog internal use only.
+     *
+     * @see _FlagsInternalProxy
+     */
+    @Suppress("FunctionName")
+    @JvmSynthetic
+    fun _getInternal(): _FlagsInternalProxy?
```
0xnm: I would propose to remove this method: it will still show up in the suggestions dropdown in Kotlin.

One can create an instance of `_FlagsInternalProxy` by themselves, the constructor is public there.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,14 +407,75 @@ internal class DatadogFlagsClient(
...
+        }
+    }
+
+    // endregion
+
+    // region Internal APIs exposed through _FlagsInternalProxy
+
+    private val internalProxy = _FlagsInternalProxy(this)
+
+    override fun _getInternal(): _FlagsInternalProxy = internalProxy
```
0xnm: both can be removed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/PrecomputedFlagConstants.kt
0xnm, line 1: Seems like moving this class is not necessary? It still stays `internal`. Can you please rollback this change?

0xnm (changes requested):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,14 +407,75 @@ internal class DatadogFlagsClient(
     }
 
     private fun <T : Any> trackResolution(resolution: InternalResolution.Success<T>) {
-        if (resolution.flag.doLog) {
+        trackResolution(resolution.flagKey, resolution.flag, resolution.value, resolution.context)
+    }
+
+    private fun <T : Any> trackResolution(
+        flagKey: String,
```
typotter, line 412: Flag Value is a member of the precomputed flag.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,14 +407,75 @@ internal class DatadogFlagsClient(
...
+    override fun _getInternal(): _FlagsInternalProxy = internalProxy
+
+    /**
+     * Retrieves a snapshot of all flag assignments.
+     *
+     * Supposed to be used by internal Datadog packages to get flags state snapshot for a given evaluation context.
+     *
+     * @return A map of flag key to precomputed flag, or null if no flags are available.
+     */
+    internal fun getFlagAssignmentsSnapshot(): Map<String, PrecomputedFlag>? = flagsRepository.getFlagsSnapshot()
```
typotter: Where does the caller get the `EvaluationContext` from?
What do you think of 

```kotlin
interface FlagAssignmentSnapshot {
  Map<String, ResolutionDetails> flagAssignments
  EvaluationContext: context
}
```

Then, the track method would be

```kotlin
trackResolution(flagKey, flagAssignmentsSnapshot)
```
The `trackResolution` method would look up the resolution for that flag from the snapshot and the context for tracking.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/PrecomputedFlag.kt
```diff
@@ -0,0 +1,34 @@
...
+ * @param variationType The variation type.
+ * @param variationValue The variation value.
+ * @param doLog Whether to log.
+ * @param allocationKey The allocation key.
+ * @param variationKey The variation key.
+ * @param extraLogging Extra logging metadata.
+ * @param reason The evaluation reason.
+ */
+@InternalApi
+data class PrecomputedFlag(
```
typotter: Which of these fields are needed for including in the Flag Assignments snapshot? I'd prefer to keep an internal model as this specifically represents when we are sending "over the wire" from the edge assignments server to the SDK.
I think `ResolutionDetails` should have what you need and it's already public.

yugisu-flux: `ResolutionDetails` doesn't have the `doLog` property from `PrecomputedFlag`, which is required in the `trackResolution` method

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
yugisu-flux: The context is stored on JavaScript side alongside with flags cache, and is being sent from there.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
yugisu-flux: RUM subpackage does the same in [RumMonitor](https://github.com/DataDog/dd-sdk-android/blob/187cbd95afe4d84122e5fa500144950fc8a35a89/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumMonitor.kt#L390). This seemed like an established pattern in the codebase so I mimicked it in the flags package

0xnm: This was a first try to expose private APIs, but this still leaks it to the suggestions dropdown in case of Kotlin (not in Java though), so it is better simply to remove this and go through the manual call of `_FlagsInternalProxy` constructor.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/PrecomputedFlag.kt
typotter: Synced on this and agreed to make an external Interface that we implement with our internal data class. This avoids a lot of unnecessary fieldmapping code

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/_FlagsInternalProxy.kt
```diff
@@ -0,0 +1,45 @@
...
+ * are not considered part of the public interface of the Datadog SDK.
+ */
+@InternalApi
+@Suppress(
+    "UndocumentedPublicClass",
+    "UndocumentedPublicFunction",
+    "UndocumentedPublicProperty",
+    "ClassName",
+    "ClassNaming",
+    "VariableNaming"
```
0xnm: variable naming suppression can be removed, there is no violation here it seems

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,14 +404,66 @@ internal class DatadogFlagsClient(
...
+            VariationType.BOOLEAN.value -> flag.variationValue.lowercase(Locale.US).toBooleanStrictOrNull()
+                ?: flag.variationValue
+            VariationType.STRING.value -> flag.variationValue
+            VariationType.INTEGER.value -> flag.variationValue.toIntOrNull() ?: flag.variationValue
+            VariationType.NUMBER.value, VariationType.FLOAT.value -> flag.variationValue.toDoubleOrNull()
+                ?: flag.variationValue
+            VariationType.OBJECT.value -> try {
+                JSONObject(flag.variationValue)
+            } catch (_: JSONException) {
+                flag.variationValue
```
0xnm: would it be useful to make `featureSdkCore.internalLogger.log` call here to log exception instead of swallowing it?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/model/PrecomputedFlag.kt
```diff
@@ -6,14 +6,15 @@
...
-internal data class PrecomputedFlag(
-    val variationType: String,
-    val variationValue: String,
-    val doLog: Boolean,
-    val allocationKey: String,
-    val variationKey: String,
-    val extraLogging: JSONObject,
-    val reason: String
-)
+data class PrecomputedFlag(
```
0xnm: blocking: It seems that this class can get back the `internal` modifier. If I'm not wrong, it is never exposed in the public API and only `UnparsedFlag` is exposed as a part of private API.

0xnm (changes requested):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/model/PrecomputedFlag.kt
yugisu-flux: That's a pretty good catch, thank you

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/UnparsedFlag.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.json.JSONObject
+
+/**
+ * Represents unparsed flag data kept in the flags state.
+ *
+ * This is an internal API and should not be used by public consumers.
+ */
+@InternalApi
+@Suppress("UndocumentedPublicProperty")
+interface UnparsedFlag {
```
yugisu-flux, line 19: TBH I don't really like this naming, open to better name ideas 🙏

features/dd-sdk-android-flags/api/apiSurface
```diff
@@ -29,6 +29,10 @@ interface com.datadog.android.flags.StateObservable
   fun getCurrentState(): com.datadog.android.flags.model.FlagsClientState
   fun addListener(FlagsStateListener)
   fun removeListener(FlagsStateListener)
+class com.datadog.android.flags._FlagsInternalProxy
+  constructor(FlagsClient)
+  fun getFlagAssignmentsSnapshot(): Map<String, com.datadog.android.flags.model.UnparsedFlag>?
```
0xnm: Minor: normally returning nullable collection should be avoided, it is better to return empty collection if there is nothing to return. This way there is no `null` case to handle at the call site.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/_FlagsInternalProxy.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Methods, members, and functionality of this class are subject to change without notice, as they
+ * are not considered part of the public interface of the Datadog SDK.
+ */
+@InternalApi
+@Suppress("ClassName", "UndocumentedPublicFunction")
+class _FlagsInternalProxy(private val client: FlagsClient) {
+    fun getFlagAssignmentsSnapshot(): Map<String, UnparsedFlag>? = if (client is DatadogFlagsClient) {
+        client.getFlagAssignmentsSnapshot()
+    } else {
+        null
```
0xnm:
```suggestion
        emptyMap()
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,15 +404,81 @@ internal class DatadogFlagsClient(
...
+            VariationType.NUMBER.value, VariationType.FLOAT.value -> flag.variationValue.toDoubleOrNull()
+            VariationType.OBJECT.value -> try {
+                JSONObject(flag.variationValue)
+            } catch (exception: JSONException) {
+                val errorMessage = "Failed to parse value '${flag.variationValue}' as JSONObject - ${exception.message}"
+                featureSdkCore.internalLogger.log(
+                    InternalLogger.Level.WARN,
+                    InternalLogger.Target.MAINTAINER,
+                    { "Flag '$flagKey': $errorMessage" }
+                )
```
0xnm: it is better to attach exception to the `log` call

```suggestion
                val errorMessage = "Failed to parse value '${flag.variationValue}' as JSONObject"
                featureSdkCore.internalLogger.log(
                    InternalLogger.Level.WARN,
                    InternalLogger.Target.MAINTAINER,
                    { "Flag '$flagKey': $errorMessage" },
                    exception
                )
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,15 +404,81 @@ internal class DatadogFlagsClient(
...
+            VariationType.STRING.value -> flag.variationValue
+            VariationType.INTEGER.value -> flag.variationValue.toIntOrNull()
+            VariationType.NUMBER.value, VariationType.FLOAT.value -> flag.variationValue.toDoubleOrNull()
+            VariationType.OBJECT.value -> try {
+                JSONObject(flag.variationValue)
+            } catch (exception: JSONException) {
+                val errorMessage = "Failed to parse value '${flag.variationValue}' as JSONObject - ${exception.message}"
+                featureSdkCore.internalLogger.log(
+                    InternalLogger.Level.WARN,
+                    InternalLogger.Target.MAINTAINER,
```
0xnm: Should the target be the `USER`? Since I guess it is about user data parsing

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,15 +404,81 @@ internal class DatadogFlagsClient(
...
+            }
+            else -> null
+        }
+
+        if (value == null) {
+            featureSdkCore.internalLogger.log(
+                InternalLogger.Level.WARN,
+                InternalLogger.Target.MAINTAINER,
+                { "Flag '$flagKey': Failed to parse value '${flag.variationValue}' as '${flag.variationType}'" }
+            )
```
0xnm, line 461: not sure if it is useful, because we already have a log entry above in case of exception, so if exception happens we will have two log entries about the same

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/repository/DefaultFlagsRepository.kt
```diff
@@ -82,6 +82,20 @@ internal class DefaultFlagsRepository(
...
+        val state = atomicState.get()
+        if (state != null) {
+            return state.flags
+        }
+        internalLogger.log(
+            InternalLogger.Level.WARN,
+            InternalLogger.Target.USER,
+            { WARN_CONTEXT_NOT_SET }
+        )
+        return null
```
0xnm:
```suggestion
        return emptyMap()
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/repository/FlagsRepository.kt
```diff
@@ -17,4 +17,5 @@ internal interface FlagsRepository {
     fun setFlagsAndContext(context: EvaluationContext, flags: Map<String, PrecomputedFlag>)
     fun getPrecomputedFlagWithContext(key: String): Pair<PrecomputedFlag, EvaluationContext>?
     fun hasFlags(): Boolean
+    fun getFlagsSnapshot(): Map<String, PrecomputedFlag>?
```
0xnm:
```suggestion
    fun getFlagsSnapshot(): Map<String, PrecomputedFlag>
```

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
```diff
@@ -1343,4 +1343,369 @@ internal class DatadogFlagsClientTest {
...
+
+        // When
+        testedClient.trackFlagSnapshotEvaluation(fakeFlagKey, fakeFlag, fakeContext)
+
+        // Then
+        verify(mockProcessor).processEvent(
+            flagName = eq(fakeFlagKey),
+            context = eq(fakeContext),
+            data = eq(fakeFlag)
+        )
```
0xnm, line 1746: tip: when all matchers are `eq`, `eq` usage can be omitted

```suggestion
        verify(mockProcessor).processEvent(
            flagName = fakeFlagKey,
            context = fakeContext,
            data = fakeFlag
        )
```

0xnm (commented):
I added some suggestions, nothing blocking, mostly about API/logging. I think it is a final round and we are good to go after.

features/dd-sdk-android-flags/api/apiSurface
yugisu-flux: If the returned collection is `null`, that means the client initialization has failed at some step. Do you have an alternative to handle the case of the clear failure vs the flags cache is empty so we return an empty collection? 

I guess the alternatives would be an pattern Optional (feels better than returning `null`?) or throwing an exception from this method (personally I don't like this pattern).

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
yugisu-flux: This log triggers for all variation types other than JSONObject. This is made this way to avoid swallowing JSONException details and provider them in the corresponding branch.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
yugisu-flux: This log is corresponding to the flags data parsing logic in `readAndParseAssignment` (line 289).

features/dd-sdk-android-flags/api/apiSurface
0xnm: I think for the client initialization we now have `FlagsClientState` and `FlagsClient.state.getCurrentState` API, so I wouldn't rely on the `null` value here to make a state judgement. If RN needs to know the state it is better to use that API.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
0xnm: It seems there is a request to do another round of review, but I don't see a change related to this comment, there is only a merge of `develop` since last review.

yugisu-flux: Whops forgot to hit the push button

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
0xnm: My point here is that `MAINTAINER` target logs are visible only in the SDK debug builds, they won't be shown in the release build. So if user _needs_ to know about this error, it should be `USER` target.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,15 +404,82 @@ internal class DatadogFlagsClient(
...
+    private fun parseFlagValueString(flagKey: String, flag: UnparsedFlag): Any {
+        val value: Any? = when (flag.variationType) {
+            VariationType.BOOLEAN.value -> flag.variationValue.lowercase(Locale.US).toBooleanStrictOrNull()
+            VariationType.STRING.value -> flag.variationValue
+            VariationType.INTEGER.value -> flag.variationValue.toIntOrNull()
+            VariationType.NUMBER.value, VariationType.FLOAT.value -> flag.variationValue.toDoubleOrNull()
+            VariationType.OBJECT.value -> try {
+                JSONObject(flag.variationValue)
+            } catch (exception: JSONException) {
+                val errorMessage = "Failed to parse value '${flag.variationValue}' as JSONObject - ${exception.message}"
```
0xnm:
```suggestion
                val errorMessage = "Failed to parse value '${flag.variationValue}' as JSONObject"
```

There is no need to add exception message here, it will already be shown due to passing `exception` object to the `log` call.

0xnm (dismissed):
lgtm, added some final remarks.

And I would still strongly recommend to change `fun getFlagAssignmentsSnapshot(): Map<String, com.datadog.android.flags.model.UnparsedFlag>?` to `fun getFlagAssignmentsSnapshot(): Map<String, com.datadog.android.flags.model.UnparsedFlag>`, because having a nullable collection is anti-pattern. Once it is out it will be public ABI-breaking change if type should be changed to non-nullable.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
yugisu-flux: @typotter FYI the target has been changed to USER in other related flag value parsing logic as well. Does that work for you?

0xnm (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/UnparsedFlag.kt
typotter: I think PrecomputedFlag or FlagAssignment would work here. We don't really need to say that it's "unparsed" because the flag assignment itself has been parsed from the server response. It's just that the value has not been typed yet and that's more of an internal concern.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,14 +404,81 @@ internal class DatadogFlagsClient(
...
+    }
+
+    // endregion
+
+    // region Internal APIs exposed through _FlagsInternalProxy
+
+    /**
+     * Retrieves a snapshot of all flag assignments.
+     *
+     * Supposed to be used by internal Datadog packages to get flags state snapshot for a given evaluation context.
```
typotter:
```suggestion
     * Explicitly for use by the Datadog React Native SDK to get flags state snapshot for a given evaluation context.
```

typotter (dismissed):
LGTM for FFE SDK

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -400,14 +404,81 @@ internal class DatadogFlagsClient(
...
+     *
+     * Supposed to be used by internal Datadog packages to track flag evaluations from an exact flags state snapshot.
+     */
+    internal fun trackFlagSnapshotEvaluation(flagKey: String, flag: UnparsedFlag, context: EvaluationContext) {
+        val flagValue = parseFlagValueString(flagKey, flag)
+
+        trackResolution(flagKey, flag, flagValue, context)
+    }
+
+    private fun parseFlagValueString(flagKey: String, flag: UnparsedFlag): Any {
```
typotter, line 451: Can you use the `FlagValueConverter` here?

0xnm (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -451,37 +451,39 @@ internal class DatadogFlagsClient(
...
-                featureSdkCore.internalLogger.log(
-                    InternalLogger.Level.WARN,
-                    InternalLogger.Target.USER,
-                    { "Flag '$flagKey': $errorMessage" },
-                    exception
-                )
-                flag.variationValue
-            }
-            else -> null
+        val variationTypeToKClass = mapOf(
```
typotter: nit: this could be a static constant instead of being redefined everytime the function runs.

typotter (dismissed):
lgtm, approved.
Just a suggestion but nothing blocking from ffe. 🚢

typotter (dismissed):

0xnm (dismissed):

0xnm (dismissed):

0xnm (approved):

PR #3052, opened 2025-12-12
RUM-12069: Fix service handling in `ddtags` of `LogEvent`

0xnm:
### What does this PR do?

Follow up for https://github.com/DataDog/dd-sdk-android/pull/3006

Apparently `service` in `ddtags` was dropped due to the following legacy verification https://github.com/DataDog/dd-sdk-android/blob/ed7f4e16f94359d32ad1a7e16fdd3121ad5d87b5/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/constraints/DatadogDataConstraints.kt#L163

It was added before reserved tags were added by the SDK, and didn't account for `variant`, `version`, `env` tags we add, but hit `service` tag.

And since `ddtags` don't show up in the Log event attributes in UI and `service` exists also in attributes, the only way to spot this was in the Logcat.

This was also missed due to the missing test coverage in `single-fit/logs`, so this PR adds the necessary assertions and also the logic to drop user-provided tags if they clash with what SDK adds.


features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/domain/DatadogLogGenerator.kt
```diff
@@ -316,7 +319,24 @@ internal class DatadogLogGenerator(
...
     ): MutableSet<String> {
-        val combinedTags = mutableSetOf<String>().apply { addAll(tags) }
+        val combinedTags = tags
+            .filter {
+                val key = it.split(":").firstOrNull()
+                if (key != null && key in RESERVED_TAG_KEYS) {
+                    internalLogger.log(
+                        InternalLogger.Level.WARN,
+                        InternalLogger.Target.USER,
+                        { RESERVED_TAG_DROPPED_WARNING.format(Locale.US, it) },
```
satween, line 329: nit: shouldn't we send this event for Maintainer or as the Telemetry as well?

satween (dismissed):

features/dd-sdk-android-logs/src/main/kotlin/com/datadog/android/log/internal/domain/DatadogLogGenerator.kt
0xnm: I don't think so, it is about user tags, so intended audience is only a user.

PR #2998, opened 2025-11-10
feat: Openfeature Provider implementation powered by DD Flags

typotter:
What does this PR do?

  This PR adds a new OpenFeature provider module (dd-sdk-android-flags-openfeature) that integrates the Datadog Feature Flags SDK with the
  OpenFeature API, enabling developers to use standardized feature flag management while leveraging Datadog's observability platform.

  Key components:
  - DatadogFlagsProvider: OpenFeature FeatureProvider implementation that adapts the OpenFeature API to Datadog's FlagsClient
  - Blocking Lifecycle: Implements spec-compliant blocking for initialize() and onContextSet() methods
  - State Observation: observe() method emitting provider state change events via Kotlin Flow
  - Extension Function: FlagsClient.asOpenFeatureProvider() for convenient wrapping of existing clients
  - Static-Context Paradigm: Implements OpenFeature's static-context model with global context management

  Motivation

  Why OpenFeature Integration?
  1. Standardization: OpenFeature provides a vendor-neutral, community-driven specification for feature flagging
  2. Ecosystem: Enables integration with OpenFeature-compatible tools and libraries
  3. Migration Path: Allows customers to adopt OpenFeature API while using Datadog as the backend
  4. Multi-Vendor: Customers can switch between feature flag providers without code changes

  Additional Notes

**  Provider Lifecycle & Blocking Behavior (OpenFeature Spec 1.7)**

  Per the OpenFeature specification, provider lifecycle methods must block until ready:

  initialize(initialContext)
  - Blocks until provider reaches Ready or Error state
  - Uses suspendCoroutine + StateObservable.addListener() to wait for terminal state
  - Throws exception if initialization fails (Error state)
  - Implementation: DatadogFlagsProvider.kt:99-124

  onContextSet(oldContext, newContext)
  - Blocks during context reconciliation until Ready, Stale, or Error state
  - Enables OpenFeature SDK to emit PROVIDER_RECONCILING events while method executes
  - Returns normally for Ready/Stale, throws for Error
  - Implementation: DatadogFlagsProvider.kt:143-169

  This blocking design ensures:
  - SDK knows when reconciliation is happening → emits PROVIDER_RECONCILING automatically
  - SDK emits PROVIDER_CONTEXT_CHANGED when onContextSet() completes
  - Proper state visibility for OpenFeature clients

  Event Emission via observe() (OpenFeature Spec Section 5)

  Per the OpenFeature Events specification, providers emit only certain events while the SDK handles others:

[...]

PR #3068, opened 2025-12-19
FFL-1621 Fix: set currentState in a blocking call

typotter:
### What does this PR do?

This PR changes the way that the current state's access and mutation is handled to ensure the ordering of state changes delivered to listeners. It also emphasizes the need for usercode listeners to catch their own exceptions lest they bubble and crash the app.

### Motivation

As noted in #2998, the `currentState` was updated synchronously erroneously and prevented other components from realizing that state directly after updating
 
The decision was made to bubble usercode exceptions in order to _fail fast_ and surface those errors to the dev rather than catching and logging them in our dispatch handler.

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
+++ b/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
@@ -24,6 +24,7 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.verifyNoMoreInteractions
 import org.mockito.quality.Strictness
+import java.util.concurrent.CountDownLatch
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.stream.Stream
 
@@ -214,9 +215,9 @@ internal class FlagsStateManagerTest {
         val stateNew = FlagsClientState.Ready
 
         val receivedStates = mutableListOf<Pair<FlagsClientState, Long>>()
-        val addListenerStarted = java.util.concurrent.CountDownLatch(1)
-        val addListenerSlowCallbackStarted = java.util.concurrent.CountDownLatch(1)
-        val updateStateCanProceed = java.util.concurrent.CountDownLatch(1)
+        val addListenerStarted = CountDownLatch(1)
+        val addListenerSlowCallbackStarted = CountDownLatch(1)
+        val updateStateCanProceed = CountDownLatch(1)
 
         // Listener that is slow to process the initial state notification
         val slowListener = object : FlagsStateListener {


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -56,27 +56,34 @@ internal class FlagsStateManager(
...
-        executorService.executeSafe(
-            operationName = UPDATE_STATE_OPERATION_NAME,
-            internalLogger = internalLogger
-        ) {
+        // lock the currentState until listener notifications are queued to prevent new listeners during state change.
+        synchronized(currentState) {
             currentState = newState
             subscription.notifyListeners {
-                onStateChanged(newState)
+                executorService.executeSafe(
```
typotter: enqueue safe executions of the listeners instead of safe execution of the queue. Now listeners cannot not crash the rest

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -56,27 +56,34 @@ internal class FlagsStateManager(
...
+                ) {
+                    onStateChanged(newState)
+                }
             }
         }
     }
 
     override fun addListener(listener: FlagsStateListener) {
-        subscription.addListener(listener)
+        synchronized(currentState) {
```
typotter: This sync block is so we can notify the listener of the current state and add it to the subscriber list without the state changing underneath

ambushwork: Question, should `removeListener` be synchronized as well same as here?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
ambushwork: Should we move `synchronized` protection inside `DDCoreSubscriptionImpl`?, it seems to me the job of `DDCoreSubscriptionImpl` to guarantee the thread-safety no matter where it is used.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
typotter: yes, fixed.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
typotter: The `DDCoreSubscription` ensures atomicity wrt modifying the subscriber list during a notification loop. Our `synchronized` here is layering atomicity linking the notification mechanics of `DDCoreSubscription` with mutation of the `currentState`. This coupling is a specific use case of the `DDCoreSubscription` or another layer on top.

`DDCoreSubscription` is still susceptible to listeners crashing the notification loop. We could put a try-catch there, but it might be better to keep the onus on the caller (which easy to do via the `block` argument passed to `notifyListeners`.

wdyt?

aleksandr-gringauz: why do we need do `executorService.executeSafe` instead of simply calling `onStateChanged(newState)`?

typotter: originally, it was to ensure fairness of state changes - which we've actually lost here with the synchronized block (switched to a `ReentrantLock` now)

The executor lets us run the subscription consumer off-thread and not block the caller to `setEvaluationContext` on potential long running consumers. Would it be better to leave the thread mgmt to the caller here? Would the caller need to run their consumer specifically in the UI thread (or some other specific thread?)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -33,19 +35,26 @@ internal class FlagsStateManager(
     private val executorService: ExecutorService,
     private val internalLogger: InternalLogger
 ) : StateObservable {
+    /**
+     * Fair lock to ensure FIFO ordering of state mutations and listener operations.
+     * The fair parameter ensures that threads acquire the lock in the order they requested it.
+     */
```
0xnm, line 39: > threads acquire the lock in the order they requested it

Does it really matter actually? I'm wondering if it is better to use `ReadWriteLock` since probably the number of read operations will dominate over the number of write operations, or is it a wrong assumption?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
0xnm: > Would it be better to leave the thread mgmt to the caller here?

I would say yes, it is up to the caller to do it unless we are picky about listener performance and for us it is critical for it to be fast. And then we can wrap simply `onStateChanged` calls in `try-catch` if we want to continue work after consumer threw exception. But then this exception will be handled, so maybe it is actually better to let it propagate and crash the app (since it is exception from the customer code).

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
typotter: Good point. We could use a `ReentrantReadWriteLock` to allow reads and still provide ordering for writes (to maintain the ordering of state change notifications sent to listeners).

In practice, I think the direct reads will not be that high, as the use case will typically be to, once on app load:
1. check the current state
2. listen to state changes (typically, to then load flag-critical parts of the app when the provider is READY)

At any rate, it makes sense to allow the current state to be read concurrently with notification operations. Thanks!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
typotter: I am in favour of the more straightforward approach here where we put onus for threading and exception handling on the user. Fail-fast prevents us from hiding exceptions from the dev and clear docs should point them in the right direction.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -6,60 +6,64 @@
...
+import kotlin.concurrent.read
+import kotlin.concurrent.write
 
 /**
  * Manages state transitions and notifications for a [com.datadog.android.flags.FlagsClient].
  *
  * This class handles state change notifications to registered listeners. All notification
  * methods are thread-safe and guarantee ordered delivery to listeners by using a
- * single-threaded executor service.
+ * fair ReentrantReadWriteLock.
```
0xnm:
```suggestion
 * fair [ReentrantReadWriteLock].
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -6,60 +6,64 @@
...
+     * Fair read-write lock to ensure FIFO ordering of state mutations and allow concurrent reads.
+     * The fair parameter ensures that threads acquire the lock in the order they requested it.
+     * Read operations can proceed concurrently, while write operations are exclusive.
+     */
+    private val stateLock = ReentrantReadWriteLock(true)
+
     /**
      * The current state of the client.
-     * Thread-safe: uses volatile for visibility across threads.
+     * Thread-safe: access is protected by stateLock.
```
0xnm:
```suggestion
     * Thread-safe: access is protected by [stateLock].
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -68,24 +72,15 @@ internal class FlagsStateManager(
...
-        subscription.addListener(listener)
-
-        // Emit current state to new listener
-        executorService.executeSafe(
-            operationName = NOTIFY_NEW_LISTENER_OPERATION_NAME,
-            internalLogger = internalLogger
-        ) {
-            val state = currentState
-            listener.onStateChanged(state)
+        stateLock.write {
```
0xnm: in fact both `addListener` and `removeListener` don't need to be wrapped in the lock, because `DDCoreSubscription` is using copy-on-write under the hood, so changes to the subscription won't affect ongoing `subscription.notifyListeners` call

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -139,6 +122,90 @@ internal class FlagsStateManagerTest {
...
+            "listener1",
+            "listener2"
+            // listener3 should NOT be in the list
+        )
+    }
+
+    @Test
+    fun `M notify all listeners W multiple listeners registered and state updated`() {
+        // Given
+        val notificationCount = java.util.concurrent.atomic.AtomicInteger(0)
```
0xnm:
```suggestion
        val notificationCount = AtomicInteger(0)
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
typotter: We want to ensure that the listener gets the current state here before any other state is broadcast, so we lock the current state until done. However, if we notify `listener` with the current state before adding it to the listeners, we can get by without the lock.

dropped the lock for `removeListener` as well - `DDCoreSubscription` is indeed sufficient here

0xnm (dismissed):
lgtm. Maybe you would like to get also a look from @aleksandr-gringauz before merging.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -68,24 +72,11 @@ internal class FlagsStateManager(
     }
 
     override fun addListener(listener: FlagsStateListener) {
+        listener.onStateChanged(currentState)
```
0xnm: indeed, it should be probably `getCurrentState` here or `currentState` guarded by the read lock

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsStateManager.kt
```diff
@@ -68,24 +72,11 @@ internal class FlagsStateManager(
...
         subscription.addListener(listener)
```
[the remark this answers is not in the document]
aleksandr-gringauz: Yes. This is it.

I think it is better to put these 2 lines under the read-lock:

```
stateLock.read {
    listener.onStateChanged(currentState)
    subscription.addListener(listener)
}
```

Because otherwise in case of concurrent `addListener(listener_1)` and `updateState(new_state)` `listener_1` might end up not with the latest state in the end.

The order could be the following:
1. `addListener(listener_1)` and `updateState(state_new)` are called concurrently from different threads.
2. read of `currentState` occurs in `addListener` and returns `state_old`.
3. `updateState` does its thing, updating `currentState=state_new` and notifying listeners. But not `listener_1`, because it hasn't been added yet.
4. `listener.onStateChanged(state_old)` inside `addListener` is called.
5. `listener_1` ends up with `state_old`.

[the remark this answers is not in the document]
typotter: Reverting back to the read lock here!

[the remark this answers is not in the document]
typotter: rewrapped the state access + addListener in the read lock plus added a test. ptal

aleksandr-gringauz (dismissed):

PR #3076, opened 2025-12-24
Rebuild profiling request for every profiling start

ambushwork:
### What does this PR do?

Prior to this PR, in `PerfettoProfile` the cancel signal and profiling request are created only when the class is initialized (not recreated if the `start` is called for another time), which causes the following request can't be cancelled, because the original cancel signal is already cancelled before thus in a "cancelled" state.

### Motivation

This changes will be necessary to support multiple profiling request in the future.

### DIFF
--- a/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
+++ b/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
@@ -30,6 +30,7 @@ import org.mockito.kotlin.any
 import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.doReturn
 import org.mockito.kotlin.eq
+import org.mockito.kotlin.inOrder
 import org.mockito.kotlin.isNull
 import org.mockito.kotlin.mock
 import org.mockito.kotlin.verify
@@ -393,6 +394,44 @@ class PerfettoProfilerTest {
 
         callbackCaptor.firstValue.accept(mockResult)
         verifyNoInteractions(mockProfilerCallback)
+    }
+
+    @Test
+    fun `M not have the same stop signal W start&stop multiple times`() {
+        inOrder(mockService) {
+            // When
+            // First request
+            testedProfiler.start(mockContext, setOf(fakeInstanceName))
+            val stopSignalCaptor = argumentCaptor<CancellationSignal>()
+            testedProfiler.stop(fakeInstanceName)
+
+            verify(mockService).requestProfiling(
+                eq(ProfilingManager.PROFILING_TYPE_STACK_SAMPLING),
+                any<Bundle>(),
+                any<String>(),
+                stopSignalCaptor.capture(),
+                any(),
+                callbackCaptor.capture()
+            )
+            // Then
+            val successResult = mock<ProfilingResult> {
+                on { errorCode } doReturn ProfilingResult.ERROR_NONE
+                on { resultFilePath } doReturn fakePath
+            }
+            callbackCaptor.firstValue.accept(successResult)
+
+            // Second request
+            testedProfiler.start(mockContext, setOf(fakeInstanceName))
+            verify(mockService).requestProfiling(
+                eq(ProfilingManager.PROFILING_TYPE_STACK_SAMPLING),
+                any<Bundle>(),
+                any<String>(),
+                stopSignalCaptor.capture(),
+                any(),
+                callbackCaptor.capture()
+            )
+            assertThat(stopSignalCaptor.firstValue).isNotSameAs(stopSignalCaptor.secondValue)
+        }
     }
 
     private class StubTimeProvider : TimeProvider {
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -395,6 +396,49 @@ class PerfettoProfilerTest {
...
+        val inOrder = inOrder(mockService)
+
+        // When
+        // First request
+        testedProfiler.start(mockContext, setOf(fakeInstanceName))
+        val firstStopSignalCaptor = argumentCaptor<CancellationSignal>()
+        testedProfiler.stop(fakeInstanceName)
+
+        // Then
+        inOrder.verify(mockService).requestProfiling(
```
0xnm: there is a bit better API which will allow to get rid of `val inOrder = inOrder(mockService)` above:

```suggestion
        inOrder(mockService) {
          verify(mockService).requestProfiling(...)
        }
```

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -395,6 +396,49 @@ class PerfettoProfilerTest {
...
+        val successResult = mock<ProfilingResult> {
+            on { errorCode } doReturn ProfilingResult.ERROR_NONE
+            on { resultFilePath } doReturn fakePath
+        }
+        callbackCaptor.firstValue.accept(successResult)
+        val captor = argumentCaptor<PerfettoResult>()
+        verify(mockProfilerCallback).onSuccess(captor.capture())
+
+        // Second request
+        val secondStopSignalCaptor = argumentCaptor<CancellationSignal>()
```
0xnm: you can actually reuse the same captor, all captures are recorded and accessible through `allValue`

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -395,6 +396,49 @@ class PerfettoProfilerTest {
...
+        testedProfiler.start(mockContext, setOf(fakeInstanceName))
+        inOrder.verify(mockService).requestProfiling(
+            eq(ProfilingManager.PROFILING_TYPE_STACK_SAMPLING),
+            any<Bundle>(),
+            any<String>(),
+            secondStopSignalCaptor.capture(),
+            any(),
+            callbackCaptor.capture()
```
0xnm: seems like this capture is useless, we don't read it back

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -395,6 +396,44 @@ class PerfettoProfilerTest {
         verifyNoInteractions(mockProfilerCallback)
     }
 
+    @Test
+    fun `M not have the same stop signal W start&stop multiple times`() {
+        inOrder(mockService) {
```
0xnm, line 401: I think it is better just to have 2 `inOrder(...) { }` invocations local to the corresponding `// Then` sections, rather than wrap the whole test in it.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -395,6 +396,44 @@ class PerfettoProfilerTest {
...
+
+            // Second request
+            testedProfiler.start(mockContext, setOf(fakeInstanceName))
+            verify(mockService).requestProfiling(
+                eq(ProfilingManager.PROFILING_TYPE_STACK_SAMPLING),
+                any<Bundle>(),
+                any<String>(),
+                stopSignalCaptor.capture(),
+                any(),
+                callbackCaptor.capture()
```
0xnm, line 431:
```suggestion
                any()
```

captured value is not read anyway

0xnm (approved):

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
ambushwork: It doesn't seem to work, if I break it down into two `inOrder` block, the second block in not able to know the previous call is verified, and throw error "Wanted 1 time, But was 2 times"

PR #3079, opened 2025-12-29
RUM-10770: Remove KeepAlive event in RUM

ambushwork:
### What does this PR do?

Main changes for this PR:

* Remove `KeepAlive` in `RumRawEvent`
* Remove keep alive callback in `DatadogRumMonitor`

Due to the removal of `KeepAlive` event, there are some side-effect changes:

* `RumViewScope.sendViewUpdate` becomes `internal` so `RumViewManagerScope` can call it directly to update the view.
* `startForegroundView` used to dispatch a keep alive to update the view, now we call sendViewUpdate directly for the same purpose
* Many unit tests relied on `KeepAlive` event to flush the Rum Event into the writer,  now we use either `StopView` to flush them, either we call `sendViewUpdate` directly from `RumViewManagerScope`

### Motivation

RUM-10770


ambushwork:
@codex review

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -176,6 +177,15 @@ internal class RumViewManagerScope(
...
     }
 
+    private fun sendViewUpdateToChildren(
+        event: RumRawEvent,
+        datadogContext: DatadogContext,
+        writeScope: EventWriteScope,
+        writer: DataWriter<Any>
+    ) {
+        childrenScopes.forEach { it.sendViewUpdate(event, datadogContext, writeScope, writer) }
+    }
```
0xnm, line 189: suggestion: let's avoid having this method, because `sendViewUpdate` was made `internal` only for tests and we probably don't want to advertise calling this, the main entrypoint is still `handleEvent`. It is better to inline this at the call site.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
```diff
@@ -108,7 +108,8 @@ internal class RumViewManagerScope(
         delegateToChildren(event, datadogContext, writeScope, writer)
 
         if (event is RumRawEvent.StartView && !stopped) {
-            startForegroundView(event, datadogContext, writeScope, writer)
+            startForegroundView(event)
+            sendViewUpdateToChildren(event, datadogContext, writeScope, writer)
```
0xnm, line 114: I guess this change will trigger zero view duration telemetry https://github.com/DataDog/dd-sdk-android/blob/f7fffa7ba8bba5734294722b8ed6c95a08074e06/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L1361-L1371

This is because the time of the scope creation matches event for view update time. And then we will set view duration to 1ns, which may be very puzzling when exploring the telemetry.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
ambushwork: this is not only for tests right now, it is also called to update the view event when creating a new foreground view (before it was updated by keep alive).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
ambushwork: is it the case already before the PR? here we try to align with the existing logic

0xnm: No, here is a minor change of behavior: before it was `KeepAlive` event which was used to send view update and which is created _after_ the original event used to create view scope, so the duration was always positive. Now it is the same event as it was used to create view scope, so the duration is now zero for such initial view update.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
0xnm: It is still mainly for tests, because such usage in production code is a small workaround which shouldn't be used in general: there may be some logic between `handleEvent` and sending view update.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewManagerScope.kt
ambushwork: If we remove `KeepAlive` event and call `sendViewUpdate` directly, indeed, the view duration becomes 1ns after the creation. But this is somehow aligned to the existing IOS logic after checking with IOS team, because if the view is the last view in the session and it doesn't have any sub events which can update it, we still need to report this view event, and we will use this duration as the `time_spent`.

0xnm: If 0 -> 1ns is a legitimate view duration for such case you still have to deal with the logging I mentioned above, because it shouldn't log in this case (especially taking into account that it also logs to telemetry, it will blow up the telemetry volume).

ambushwork: As we discussed since IOS has 1ns view as a normal case if there is no update for it, we should align with them, so I add a exempting condition to check if the event is `RumRawEvent.StartView` before sending the telemetry

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1360,12 +1346,16 @@ internal open class RumViewScope(
...
     private fun resolveViewDuration(event: RumRawEvent) {
         stoppedNanos = event.eventTime.nanoTime
         val duration = stoppedNanos - startedNanos
         viewEndedMetricDispatcher.onDurationResolved(duration)
         if (duration == 0L) {
-            if (type == RumViewType.BACKGROUND && event is RumRawEvent.AddError && event.isFatal) {
+            if (type == RumViewType.BACKGROUND &&
+                event is RumRawEvent.AddError &&
+                event.isFatal || event is RumRawEvent.StartView
+            ) {
```
0xnm: minor: it is worth to extract complex condition instead of suppressing a warning, because order of the validation here is not obvious(is it `(a && b && c) || d` or `a && b && (c || d)`?):

```suggestion
            val isBackgroundFatalError = type == RumViewType.BACKGROUND &&
                event is RumRawEvent.AddError &&
                event.isFatal
            if (isBackgroundFatalError || event is RumRawEvent.StartView) {
```

Also it is worth adding a test to verify that we indeed don't have logging/telemetry for such case

ambushwork: moved part of the condition into a dedicated function

ambushwork: a unit test is also added to verify that when the event type is a background fatal error, we don't send any telemetry for 1ns duration

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1365,7 +1351,8 @@ internal open class RumViewScope(
         val duration = stoppedNanos - startedNanos
         viewEndedMetricDispatcher.onDurationResolved(duration)
         if (duration == 0L) {
-            if (type == RumViewType.BACKGROUND && event is RumRawEvent.AddError && event.isFatal) {
+            if (isFatalBackgroundEvent(event) || event is RumRawEvent.StartView
+            ) {
```
0xnm: formatting looks off

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1401,6 +1388,10 @@ internal open class RumViewScope(
         }
     }
 
+    private fun isFatalBackgroundEvent(event: RumRawEvent): Boolean {
```
0xnm: nit: while it is okay to have it as a function, probably there will be only a single call side, so it could be just a `val` inside the method. no changes needed though

ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: updated

aleksandr-gringauz (approved):

0xnm (approved):

PR #3080, opened 2025-12-29
RUM-13514: Pass original resource id to stopResource in DatadogInterceptor

aleksandr-gringauz:
### What does this PR do?

See description of the bug in the ticket (RUM-13514).

The fix is about passing the RUM resource id generated for `startResource` call to the `stopResource` call through different places in `DatadogInterceptor` and `TracingInterceptor`.

### Motivation

### Additional Notes

### DIFF
--- a/integrations/dd-sdk-android-okhttp/api/apiSurface
+++ b/integrations/dd-sdk-android-okhttp/api/apiSurface
@@ -17,7 +17,7 @@ class com.datadog.android.okhttp.DatadogEventListener : okhttp3.EventListener
     override fun create(okhttp3.Call): okhttp3.EventListener
 open class com.datadog.android.okhttp.DatadogInterceptor : com.datadog.android.okhttp.trace.TracingInterceptor
   override fun intercept(okhttp3.Interceptor.Chain): okhttp3.Response
-  override fun onRequestIntercepted(com.datadog.android.api.feature.FeatureSdkCore, okhttp3.Request, com.datadog.android.trace.api.span.DatadogSpan?, okhttp3.Response?, Throwable?)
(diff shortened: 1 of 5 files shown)


integrations/dd-sdk-android-okhttp/api/apiSurface
```diff
@@ -17,7 +17,7 @@ class com.datadog.android.okhttp.DatadogEventListener : okhttp3.EventListener
     override fun create(okhttp3.Call): okhttp3.EventListener
 open class com.datadog.android.okhttp.DatadogInterceptor : com.datadog.android.okhttp.trace.TracingInterceptor
   override fun intercept(okhttp3.Interceptor.Chain): okhttp3.Response
-  override fun onRequestIntercepted(com.datadog.android.api.feature.FeatureSdkCore, okhttp3.Request, com.datadog.android.trace.api.span.DatadogSpan?, okhttp3.Response?, Throwable?)
+  override fun onRequestIntercepted(com.datadog.android.api.feature.FeatureSdkCore, okhttp3.Request, com.datadog.android.trace.api.span.DatadogSpan?, okhttp3.Response?, Throwable?, com.datadog.android.rum.resource.ResourceId?)
```
0xnm, line 20: This makes it backward-incompatible change. I don't think it is a big deal, but still, customer code may override this  method.

Is there any way to avoid such change?

aleksandr-gringauz: Fair point.

I think it is possible to fix this problem (updated the current PR). But I don't like the solution, seems too complicated, error-prone and hard to maintain.

I created a different PR with a different approach to fix the bug https://github.com/DataDog/dd-sdk-android/pull/3086. I think it is better and simpler.

PR #3081, opened 2025-12-29
Initialize Profiler only once for multiple start calls

ambushwork:
### What does this PR do?

This PR takes one step further to support custom profiling.

The current implementation of `Profiling.kt` is based on application launch profiling, which presumes the `Profiling.start` is called only once and always called before `Profiling.enable`. 

Which brings following consequences for custom profiling:
* If `Profiling.enable` is called before `Profiling.start`, then `ProfilingFeature` will be initialized with `NoOpProfiler`
* If `Profiling.start` is called several times, the profiler will be recreated each times, which will be different from the instance held inside `ProfilingFeature`

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
@@ -21,13 +21,15 @@ import com.datadog.android.profiling.internal.ProfilingFeature
 import com.datadog.android.profiling.internal.ProfilingStorage
 import com.datadog.android.profiling.internal.perfetto.PerfettoProfiler
 import java.util.concurrent.Executors
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * An entry point to Datadog Profiling feature.
  */
 object Profiling {
 
-    private var profiler: Profiler = NoOpProfiler()
+    internal var profiler: Profiler = NoOpProfiler()
+    internal val isProfilerInitialized = AtomicBoolean(false)
 
     /**
      * Enables the profiling feature.
@@ -44,6 +46,7 @@ object Profiling {
         sdkCore: SdkCore = Datadog.getInstance()
     ) {
         val featureSdkCore = sdkCore as FeatureSdkCore
+        initializeProfiler()
         val profilingFeature = ProfilingFeature(
             sdkCore = featureSdkCore,
             configuration = configuration,
@@ -81,12 +84,18 @@ object Profiling {
 
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
     internal fun start(context: Context, sdkInstanceNames: Set<String>) {
-        profiler = PerfettoProfiler(
-            timeProvider = DefaultTimeProvider(),
-            profilingExecutor = Executors.newSingleThreadExecutor()
-        ).apply {
-            this.start(context, sdkInstanceNames)
-        }
+        initializeProfiler()
+        profiler.start(context, sdkInstanceNames)
         ProfilingStorage.removeProfilingFlag(context, sdkInstanceNames)
+    }
+
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    private fun initializeProfiler() {
+        if (!isProfilerInitialized.getAndSet(true)) {
+            profiler = PerfettoProfiler(
+                timeProvider = DefaultTimeProvider(),
+                profilingExecutor = Executors.newSingleThreadExecutor()
+            )
+        }
     }
 }
--- a/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
+++ b/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
@@ -1,0 +1,154 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.profiling
+
+import android.content.Context
+import android.os.ProfilingManager
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.core.InternalSdkCore
+import com.datadog.android.internal.data.SharedPreferencesStorage
+import com.datadog.android.profiling.forge.Configurator
+import com.datadog.android.profiling.internal.NoOpProfiler
+import com.datadog.android.profiling.internal.ProfilingFeature
+import com.datadog.android.profiling.internal.ProfilingStorage
+import com.datadog.android.profiling.internal.perfetto.PerfettoProfiler
+import fr.xgouchet.elmyr.annotation.Forgery
+import fr.xgouchet.elmyr.annotation.StringForgery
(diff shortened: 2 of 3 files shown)


aleksandr-gringauz (approved):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -81,12 +84,19 @@ object Profiling {
...
-            this.start(context, sdkInstanceNames)
-        }
+        initializeProfiler()
+        profiler.start(context, sdkInstanceNames)
         ProfilingStorage.removeProfilingFlag(context, sdkInstanceNames)
     }
+
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    private fun initializeProfiler() {
+        if (!isProfilerInitialized.get()) {
```
0xnm: nit: probably it should be `getAndSet(true)` here, because in the current shape it solves change visibility between the threads, but doesn't solve concurrency (there is a small chance that multiple threads enter this block at the same time -> one thread may yet not reach `isProfilerInitialized.set(true)` while another is doing `isProfilerInitialized.get()` check).

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
```diff
@@ -0,0 +1,183 @@
...
+
+        resetProfilerField()
+    }
+
+    @Test
+    fun `M use PerfettoProfiler W enable called before start`() {
+        // Given
+        val sdkInstanceNames = setOf(fakeInstanceName)
+
+        Mockito.mockStatic(ProfilingStorage::class.java)
```
0xnm: Is this needed? There is no work with `ProfilingStorage` inside the block.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
```diff
@@ -0,0 +1,183 @@
...
+                // When
+                Profiling.enable(fakeConfiguration, mockSdkCore)
+                Profiling.start(mockContext, sdkInstanceNames)
+
+                // Then
+                verify(mockSdkCore).registerFeature(any<ProfilingFeature>())
+
+                val profilerField = Profiling::class.java.getDeclaredField("profiler")
+                profilerField.isAccessible = true
+                val profiler = profilerField.get(Profiling) as Profiler
```
0xnm: maybe it is just easier to make `profiler` field `internal` to avoid reflection

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
```diff
@@ -0,0 +1,183 @@
...
+                val secondProfiler = profilerField.get(Profiling) as Profiler
+
+                // Then
+                assertThat(firstProfiler).isNotInstanceOf(NoOpProfiler::class.java)
+                assertThat(secondProfiler).isNotInstanceOf(NoOpProfiler::class.java)
+                assertThat(firstProfiler).isSameAs(secondProfiler)
+            }
+    }
+
+    private fun resetProfilerField() {
```
0xnm, line 150: same here, we can simply create `internal` method to reset the state if we want.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
ambushwork: yes, `Profiling.start` will call `ProfilingStorage.removeProfilingFlag(context, sdkInstanceNames)` inside.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
ambushwork: I think it will be nice to keep the minimum visibility for this field in production code, since we have a workaround in the test

0xnm: I would disagree here, because test code is reflection-heavy for no reason, we control the source code and can adjust it as needed for the test visibility. It is `internal` modifier, it won't be a part of public API.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
0xnm: This can also be avoided by simply making `ProfilingStorage.sharedPreferencesStorage` as `internal` instead of `private` and simply setting a mock there.

And by doing this we don't need to mock static.

ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
```diff
@@ -0,0 +1,157 @@
...
+        // Then
+        verify(mockSdkCore).registerFeature(any<ProfilingFeature>())
+
+        assertThat(Profiling.profiler).isNotInstanceOf(NoOpProfiler::class.java)
+        assertThat(Profiling.profiler).isInstanceOf(PerfettoProfiler::class.java)
+    }
+
+    @Test
+    fun `M use PerfettoProfiler W enable called but start not called`() {
+        Mockito.mockStatic(ProfilingStorage::class.java)
```
0xnm: `Mockito.mockStatic(ProfilingStorage::class.java)` calls here and below are not needed due to the `ProfilingStorage.sharedPreferencesStorage = mockSharedPreferencesStorage` call in the setup.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
```diff
@@ -0,0 +1,157 @@
...
+    @StringForgery
+    private lateinit var fakeInstanceName: String
+
+    @BeforeEach
+    fun `set up`() {
+        whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
+        whenever(mockSdkCore.name) doReturn fakeInstanceName
+        whenever(mockSdkCore.createSingleThreadExecutorService(any())) doReturn mockProfilingExecutor
+        whenever(mockContext.getSystemService(ProfilingManager::class.java)) doReturn mockProfilingManager
+        ProfilingStorage.sharedPreferencesStorage = mockSharedPreferencesStorage
```
0xnm, line 77: we need to remove this mock in the tear down, otherwise it stays there and may hit back in some other test

ambushwork: Good point, updated

0xnm (approved):

PR #3086, opened 2025-12-30
RUM-13514: Put Request tag with UUID in DatadogInterceptor

aleksandr-gringauz:
### What does this PR do?

See description of the bug in the ticket (RUM-13514).

This fix is about putting UUID into the Request tags, so that `buildResourceId` function can use it when it is called before `stopResource`.

### Motivation

### Additional Notes

### DIFF
--- a/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
+++ b/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
@@ -39,7 +39,6 @@ import okhttp3.Request
 import okhttp3.Response
 import okhttp3.ResponseBody
 import java.io.IOException
-import java.lang.ClassCastException
 import java.util.Locale
 import java.util.UUID
 
@@ -112,7 +111,10 @@ open class DatadogInterceptor internal constructor(
 
         val request = chain.request()
             .newBuilder()
-            .safeAddUUIDTag(UUID.randomUUID())
+            .apply {
+                @Suppress("UnsafeThirdPartyFunctionCall") // ClassCastException can't happen here.
+                tag(UUID::class.java, UUID.randomUUID())
+            }
             .safeBuild() ?: chain.request()
 
         if (rumFeature != null) {
@@ -276,14 +278,6 @@ open class DatadogInterceptor internal constructor(
             build()
         } catch (_: IllegalStateException) {
             null
-        }
-    }
-
-    private fun Request.Builder.safeAddUUIDTag(uuid: UUID): Request.Builder {
-        return try {
-            tag(UUID::class.java, uuid)
-        } catch (_: ClassCastException) {
-            this
         }
     }
 
--- a/detekt_custom_safe_calls.yml
+++ b/detekt_custom_safe_calls.yml
@@ -1220,7 +1220,6 @@ datadog:
       - "okhttp3.Request.method()"
       - "okhttp3.Request.newBuilder()"
       - "okhttp3.Request.tag(java.lang.Class)"
-      - "okhttp3.Request.Builder.tag(java.lang.Class, java.util.UUID?)"
       - "okhttp3.Request.url()"
       - "okhttp3.RequestBody.contentLength()"
       - "okhttp3.RequestBody.contentType()"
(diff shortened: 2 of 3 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -135,7 +140,7 @@ open class DatadogInterceptor internal constructor(
             originalChain = chain
         )
 
-        return super.intercept(localChain)
+        return doIntercept(localChain, request)
```
ambushwork, line 147: Is it possible to add a unit test to make sure that we pass down the modified request instead of request from the original chain?

aleksandr-gringauz: Added a new test `M call chain proceed with a different Request that contains UUID tag W intercept()`.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -264,6 +271,22 @@ open class DatadogInterceptor internal constructor(
...
 
+    private fun Request.Builder.safeBuild(): Request? {
+        return try {
+            build()
+        } catch (_: IllegalStateException) {
+            null
+        }
+    }
+
+    private fun Request.Builder.safeAddUUIDTag(uuid: UUID): Request.Builder {
```
ambushwork: s this needed? although it may throw ClassCastException here in theory, but we know that the class will match in our code.

ambushwork (dismissed):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
aleksandr-gringauz: I added [this](https://github.com/DataDog/dd-sdk-android/blob/cc54e1bb4d516e9e0594dbd1657ba0af11ace738/detekt_custom_unsafe_calls.yml#L221) line in `detekt_custom_unsafe_calls.yml` right after a similar line but for `TraceContext`.

This is now "know unsafe call" and you need to use try-catch in code if you want lint check to pass.

Should it be "know unsafe call"? I think - yes, it can throw `ClassCastException` [here](https://github.com/square/okhttp/blob/acf873543f197283995232446a35fd8a83fad77e/okhttp/src/main/kotlin/okhttp3/Request.kt#L283).

ambushwork: in this case we usually do [this](https://github.com/DataDog/dd-sdk-android/blob/develop/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/OkHttpRequestExt.kt#L17)

0xnm (dismissed):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
aleksandr-gringauz: Actually, why not. Done.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -107,13 +109,20 @@ open class DatadogInterceptor internal constructor(
...
+            .safeBuild() ?: chain.request()
+
         if (rumFeature != null) {
-            val request = chain.request()
             val url = request.url.toString()
             val method = toHttpMethod(request.method, sdkCore.internalLogger)
 
             @Suppress("DEPRECATION")
-            val requestId = request.buildResourceId(generateUuid = true)
+            val requestId = request.buildResourceId(generateUuid = false)
```
0xnm, line 125: In theory we can still keep it as `true`, right? Since `UUID` tag is already attached to the request.

aleksandr-gringauz: yes, we can leave generateUuid as true. I changed to false because it looks a bit clearer. We put UUID into tags a few lines above and it would look weird if we had `generateUuid = true` here.

PR #3092, opened 2025-12-31
Expose Profiling start/stop as public API

ambushwork:
### What does this PR do?

This PR exposes start/stop profiling as public API, make it possible to send some custom profiling from sample app.
Note: this is not the final form of custom profiling.

### DIFF
--- a/features/dd-sdk-android-profiling/api/apiSurface
+++ b/features/dd-sdk-android-profiling/api/apiSurface
@@ -8,6 +8,8 @@ class com.datadog.android.profiling.DdProfilingContentProvider : android.content
 object com.datadog.android.profiling.Profiling
   fun enable(ProfilingConfiguration = ProfilingConfiguration.DEFAULT, com.datadog.android.api.SdkCore = Datadog.getInstance())
   fun profileNextAppStartup(com.datadog.android.api.SdkCore = Datadog.getInstance(), Boolean)
+  fun start(android.content.Context, com.datadog.android.api.SdkCore = Datadog.getInstance())
+  fun stop(com.datadog.android.api.SdkCore = Datadog.getInstance())
 data class com.datadog.android.profiling.ProfilingConfiguration
   class Builder
     fun useCustomEndpoint(String): Builder
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
@@ -82,11 +82,37 @@ object Profiling {
         }
     }
 
+    /**
+     * Start profiling with given SDK instances names.
+     *
+     * @param context application context
+     * @param sdkInstanceNames the set of the SDK instances name
+     */
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
     internal fun start(context: Context, sdkInstanceNames: Set<String>) {
         initializeProfiler()
         profiler.start(context, sdkInstanceNames)
         ProfilingStorage.removeProfilingFlag(context, sdkInstanceNames)
+    }
+
+    /**
+     * Start profiling for a given SDK instance.
+     *
+     * @param context application context
+     * @param sdkCore SDK instance to start profiling with. If not provided, default SDK instance.
+     */
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    fun start(context: Context, sdkCore: SdkCore = Datadog.getInstance()) {
+        start(context, setOf(sdkCore.name))
+    }
+
+    /**
+     * Stop profiling for a given SDK instance.
+     *
+     * @param sdkCore SDK instance to stop profiling. If not provided, default SDK instance.
+     */
+    fun stop(sdkCore: SdkCore = Datadog.getInstance()) {
+        profiler.stop(sdkCore.name)
     }
 
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
--- a/sample/kotlin/src/main/kotlin/com/datadog/android/sample/profiling/ProfilingFragment.kt
+++ b/sample/kotlin/src/main/kotlin/com/datadog/android/sample/profiling/ProfilingFragment.kt
@@ -6,6 +6,8 @@
 
 package com.datadog.android.sample.profiling
 
+import android.content.Context
+import android.os.Build
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
@@ -14,7 +16,10 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.Button
 import android.widget.ProgressBar
+import android.widget.Toast
+import androidx.annotation.RequiresApi
 import androidx.fragment.app.Fragment
+import com.datadog.android.profiling.Profiling
 import com.datadog.android.sample.R
 
 @Suppress("MagicNumber")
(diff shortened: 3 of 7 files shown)


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -82,11 +83,34 @@ object Profiling {
         }
     }
 
+    /**
+     * Start profiling with given sdk instances.
+     *
+     * Note: This is currently internal usage for demo purpose not the final public API.
```
0xnm: I don't think we should put such note and `InternalApi` as well, because we have only internal builds anyway, we don't have an official release.

The whole profiling API is a subject to change.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -82,11 +83,34 @@ object Profiling {
...
+     *
+     * @param context application context
+     * @param sdkInstanceNames the set of the sdk instances name
+     */
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
-    internal fun start(context: Context, sdkInstanceNames: Set<String>) {
+    @InternalApi
+    fun start(context: Context, sdkInstanceNames: Set<String> = emptySet()) {
```
0xnm: What is the reason behind having `sdkInstanceNames: Set<String> = emptySet()` instead of the **single** `SdkCore` argument as for other features?

We used SDK instance names just for the app launch profiling because it started when SDK is not yet initialized, but here we need to make it strict: we expect user `start/stop` methods only when SDK is initialized.

Also we need to discuss if we allow/if it is possible to run multiple profiler instances in parallel, maybe we should allow only one profiler instance at the time.

ambushwork: I separate the `start` API into two for app launch and custom profiling.

For the multiple profiler instances support I guess we shouldn't support it, but will discuss properly in the RFC later.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
ambushwork: My thought was that although it's in feature branch, we build snapshot version and dogfood it, it's still visible to others.

removed anyway.

0xnm: Yes, it is visible, but it is published in a snapshots repo, so it is not considered as a stable subject.

features/dd-sdk-android-profiling/api/apiSurface
```diff
@@ -8,6 +8,9 @@ class com.datadog.android.profiling.DdProfilingContentProvider : android.content
 object com.datadog.android.profiling.Profiling
   fun enable(ProfilingConfiguration = ProfilingConfiguration.DEFAULT, com.datadog.android.api.SdkCore = Datadog.getInstance())
   fun profileNextAppStartup(com.datadog.android.api.SdkCore = Datadog.getInstance(), Boolean)
+  fun start(android.content.Context, Set<String>)
```
0xnm: this method shouldn't be `public`, it should be `internal`, it is not a part of public API

features/dd-sdk-android-profiling/api/apiSurface
```diff
@@ -8,6 +8,9 @@ class com.datadog.android.profiling.DdProfilingContentProvider : android.content
...
+  fun start(android.content.Context, com.datadog.android.api.SdkCore = Datadog.getInstance())
+  fun stop(String? = null)
```
0xnm: same here, we don't want to stop by the instance name, but by the SDK instance itself.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -82,13 +82,39 @@ object Profiling {
         }
     }
 
+    /**
+     * Start profiling with given sdk instances names.
+     *
+     * @param context application context
+     * @param sdkInstanceNames the set of the sdk instances name
```
0xnm:
```suggestion
     * Start profiling with given SDK instances names.
     *
     * @param context application context
     * @param sdkInstanceNames the set of the SDK instances name
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -82,13 +82,39 @@ object Profiling {
...
+     */
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
     internal fun start(context: Context, sdkInstanceNames: Set<String>) {
         initializeProfiler()
         profiler.start(context, sdkInstanceNames)
         ProfilingStorage.removeProfilingFlag(context, sdkInstanceNames)
     }
 
+    /**
+     * Start profiling with given sdk instances names.
```
0xnm:
```suggestion
     * Start profiling for a given SDK instance.
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -82,13 +82,39 @@ object Profiling {
...
+     *
+     * @param context application context
+     * @param sdkCore SDK instance to start profiling with. If not provided, default SDK instance
```
0xnm:
```suggestion
     * @param sdkCore SDK instance to start profiling with. If not provided, default SDK instance.
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -82,13 +82,39 @@ object Profiling {
...
+     */
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    fun start(context: Context, sdkCore: SdkCore = Datadog.getInstance()) {
+        start(context, setOf(sdkCore.name))
+    }
+
+    /**
+     * Stop profiling on given sdk instance.
```
0xnm:
```suggestion
     * Stop profiling for a given SDK instance.
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -82,13 +82,39 @@ object Profiling {
...
+     *
+     * @param sdkCore SDK instance to stop profiling. If not provided, default SDK instance
```
0xnm:
```suggestion
     * @param sdkCore SDK instance to stop profiling. If not provided, default SDK instance.
```

sample/benchmark/transitiveDependencies
0xnm, line 1: Doesn't seem like it is related to the PR

sample/kotlin/src/main/kotlin/com/datadog/android/sample/profiling/ProfilingFragment.kt
```diff
@@ -54,20 +58,23 @@ internal class ProfilingFragment : Fragment() {
...
-        }
-
-        button10s.setOnClickListener {
-            startProfiling(duration = 10, progressBar = progressBar10s)
-        }
-
-        button60s.setOnClickListener {
-            startProfiling(duration = 60, progressBar = progressBar60s)
+        context?.let { context ->
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
```
0xnm: It will be confusing if someone runs sample app on API below `VANILLA_ICE_CREAM` and has no any effect after clicking these buttons. Maybe consider showing toast with a message?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/profiling/ProfilingFragment.kt
```diff
@@ -54,20 +58,23 @@ internal class ProfilingFragment : Fragment() {
...
-            startProfiling(duration = 1, progressBar = progressBar1s)
-        }
-
-        button10s.setOnClickListener {
-            startProfiling(duration = 10, progressBar = progressBar10s)
-        }
-
-        button60s.setOnClickListener {
-            startProfiling(duration = 60, progressBar = progressBar60s)
+        context?.let { context ->
```
0xnm: we can use `requireContext` here, it shouldn't be null at this point

sample/benchmark/transitiveDependencies
ambushwork: reverted

sample/kotlin/src/main/kotlin/com/datadog/android/sample/profiling/ProfilingFragment.kt
ambushwork: Good idea, updated

sample/kotlin/src/main/kotlin/com/datadog/android/sample/profiling/ProfilingFragment.kt
```diff
@@ -54,20 +59,27 @@ internal class ProfilingFragment : Fragment() {
...
+            button10s.setOnClickListener {
+                startProfiling(requireContext(), duration = 10, progressBar = progressBar10s)
+            }
+            button60s.setOnClickListener {
+                startProfiling(requireContext(), duration = 60, progressBar = progressBar60s)
+            }
+        } else {
+            Toast.makeText(
+                requireContext(),
+                "Profiling is only working above API 35!",
```
0xnm:
```suggestion
                "Profiling is only working for API 35 and above",
```

PR #3097, opened 2026-01-04
RUM-13564: Add GraphQL errors support to RUM resource events

cdn34dd:
### What does this PR do?

Adds GraphQL errors support to RUM resource events. Extracts and parses GraphQL errors from the `_dd.graphql.errors` attribute, and populates `resource.graphql.errors` and `resource.graphql.error_count` fields in RUM event data.

### Motivation

Achieve cross-platform parity with iOS SDK. React Native SDK sends GraphQL errors from JS, but Android wasn't parsing them, resulting in missing error data.

### Additional Notes

This supports both wrapped errors format (e.g., `_dd.graphql.errors: {"errors": [...]}`) and direct array format (e.g., `_dd.graphql.errors: [...]` ) for flexibility. iOS only accepts the wrapped format, which was surprising given the the shape of the data, so here both work. 

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
@@ -253,6 +253,9 @@ internal class RumResourceScope(
         val graphqlOperationType = resourceAttributes.remove(RumAttributes.GRAPHQL_OPERATION_TYPE) as? String
         val graphqlVariables = resourceAttributes.remove(RumAttributes.GRAPHQL_VARIABLES) as? String
 
+        val graphqlErrors = (resourceAttributes.remove(RumAttributes.GRAPHQL_ERRORS) as? List<*>)
+            ?.mapNotNull { it as? ResourceEvent.Error }
+
         // The decision whether to send payloads is determined by a DatadogApolloInterceptor parameter
         val rawPayload = resourceAttributes.remove(RumAttributes.GRAPHQL_PAYLOAD) as? String
         val graphqlPayload = rawPayload?.truncateToUtf8Bytes(MAX_GRAPHQL_PAYLOAD_SIZE_BYTES)
@@ -261,7 +264,8 @@ internal class RumResourceScope(
             operationType = graphqlOperationType,
             operationName = graphqlOperationName,
             variables = graphqlVariables,
-            payload = graphqlPayload
+            payload = graphqlPayload,
+            errors = graphqlErrors
         )
 
         sdkCore.newRumEventWriteOperation(datadogContext, writeScope, writer) {
@@ -554,14 +558,17 @@ internal class RumResourceScope(
         operationType: String?,
         operationName: String?,
         variables: String?,
-        payload: String?
+        payload: String?,
+        errors: List<ResourceEvent.Error>?
     ): ResourceEvent.Graphql? {
         operationType?.toOperationType(sdkCore.internalLogger)?.let {
             return ResourceEvent.Graphql(
                 operationType = it,
                 operationName = operationName,
                 variables = variables,
-                payload = payload
+                payload = payload,
+                errorCount = errors?.takeIf { errors -> errors.isNotEmpty() }?.size?.toLong(),
+                errors = errors?.takeIf { errors -> errors.isNotEmpty() }
             )
         }
 
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScopeTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScopeTest.kt
@@ -3668,6 +3668,211 @@ internal class RumResourceScopeTest {
         }
     }
 
+    @Test
+    fun `M parse GraphQL errors W handleEvent { multiple errors with locations and path }`(
+        @Forgery kind: RumResourceKind,
+        @LongForgery(200, 600) statusCode: Long,
+        @LongForgery(0, 1024) size: Long,
+        forge: Forge
+    ) {
+        // Given
+        val operationType = forge.aValueFrom(ResourceEvent.OperationType::class.java)
+        val operationName = forge.aNullable { aString() }
+        val variables = forge.aNullable { aString() }
+
+        val errors = listOf(
+            ResourceEvent.Error(
+                message = "User not found",
+                code = "NOT_FOUND",
+                locations = listOf(ResourceEvent.Location(line = 3, column = 5)),
+                path = listOf(ResourceEvent.Path.String("user"), ResourceEvent.Path.String("profile"))
+            ),
+            ResourceEvent.Error(
+                message = "Validation failed",
+                code = null,
+                locations = null,
+                path = null
+            )
+        )
+
+        val attributes = forge.exhaustiveAttributes(excludedKeys = fakeResourceAttributes.keys) +
+            mapOf(
+                RumAttributes.GRAPHQL_OPERATION_TYPE to operationType.toString(),
+                RumAttributes.GRAPHQL_OPERATION_NAME to operationName,
+                RumAttributes.GRAPHQL_VARIABLES to variables,
+                RumAttributes.GRAPHQL_ERRORS to errors
+            )
+
+        mockEvent = RumRawEvent.StopResource(fakeKey, statusCode, size, kind, attributes)
+
+        // When
+        testedScope.handleEvent(mockEvent, fakeDatadogContext, mockEventWriteScope, mockWriter)
+
+        // Then
+        argumentCaptor<ResourceEvent> {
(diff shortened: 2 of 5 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -551,20 +553,84 @@ internal class RumResourceScope(
...
+                payload = payload,
+                errorCount = parsedErrors?.size?.toLong(),
+                errors = parsedErrors
             )
         }
 
         return null
     }
 
+    @Suppress("ReturnCount", "TooGenericExceptionCaught", "SwallowedException", "UnsafeThirdPartyFunctionCall")
```
0xnm: I don't see any swallowed exception. Also `TooGenericExceptionCaught` and `UnsafeThirdPartyFunctionCall` are better applied directly at the violation site.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -551,20 +553,84 @@ internal class RumResourceScope(
...
+    private fun parseGraphQLErrors(errorsJson: String): List<ResourceEvent.Error>? {
+        try {
+            val jsonElement = com.google.gson.JsonParser.parseString(errorsJson)
+
+            // Handle both formats:
+            // 1. Direct array: [{"message": "...", ...}]
+            // 2. Wrapped in object: {"errors": [{"message": "...", ...}]}
```
0xnm: if it is a cross-platform sending this, why it cannot be sent in a single format?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -551,20 +553,84 @@ internal class RumResourceScope(
...
             )
         }
 
         return null
     }
 
+    @Suppress("ReturnCount", "TooGenericExceptionCaught", "SwallowedException", "UnsafeThirdPartyFunctionCall")
+    private fun parseGraphQLErrors(errorsJson: String): List<ResourceEvent.Error>? {
+        try {
+            val jsonElement = com.google.gson.JsonParser.parseString(errorsJson)
```
0xnm:
```suggestion
            val jsonElement = JsonParser.parseString(errorsJson)
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -551,20 +553,84 @@ internal class RumResourceScope(
...
+                            messageBuilder = { "GraphQL errors JSON is an object but has no 'errors' array" }
+                        )
+                        return null
+                    }
+                }
+                else -> {
+                    sdkCore.internalLogger.log(
+                        level = InternalLogger.Level.WARN,
+                        target = InternalLogger.Target.USER,
+                        messageBuilder = { "GraphQL errors JSON is neither an array nor an object" }
```
0xnm: is it actionable somehow from the user standpoint?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -551,20 +553,84 @@ internal class RumResourceScope(
...
+                    )
+                }
+            }
+
+            return errors.ifEmpty { null }
+        } catch (e: Exception) {
+            sdkCore.internalLogger.log(
+                level = InternalLogger.Level.WARN,
+                target = InternalLogger.Target.USER,
+                messageBuilder = { "Failed to parse GraphQL errors JSON: ${e.message}" },
```
0xnm: exception message will already be printed by the logger

```suggestion
                messageBuilder = { "Failed to parse GraphQL errors JSON" },
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -551,20 +553,84 @@ internal class RumResourceScope(
...
+                errorCount = parsedErrors?.size?.toLong(),
+                errors = parsedErrors
             )
         }
 
         return null
     }
 
+    @Suppress("ReturnCount", "TooGenericExceptionCaught", "SwallowedException", "UnsafeThirdPartyFunctionCall")
+    private fun parseGraphQLErrors(errorsJson: String): List<ResourceEvent.Error>? {
```
0xnm: normally it is better to return empty collection rather than null

```suggestion
    private fun parseGraphQLErrors(errorsJson: String): List<ResourceEvent.Error> {
```

0xnm: what is going to send these errors? cannot we simply send a typed object instead of trying to parse the things?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScopeTest.kt
```diff
@@ -3631,6 +3631,212 @@ internal class RumResourceScopeTest {
...
+        // Then
+        argumentCaptor<ResourceEvent> {
+            verify(mockWriter).write(eq(mockEventBatchWriter), capture(), eq(EventType.DEFAULT))
+            val graphql = firstValue.resource.graphql
+            assertThat(graphql).isNotNull()
+            assertThat(graphql?.errorCount).isEqualTo(2)
+            assertThat(graphql?.errors).hasSize(2)
+            assertThat(graphql?.errors?.get(0)?.message).isEqualTo("User not found")
+            assertThat(graphql?.errors?.get(0)?.code).isEqualTo("NOT_FOUND")
+            assertThat(graphql?.errors?.get(1)?.message).isEqualTo("Validation failed")
```
0xnm:
```suggestion
            checkNotNull(graphql)
            assertThat(graphql.errorCount).isEqualTo(2)
            assertThat(graphql.errors).hasSize(2)
            assertThat(graphql.errors?.get(0)?.message).isEqualTo("User not found")
            assertThat(graphql.errors?.get(0)?.code).isEqualTo("NOT_FOUND")
            assertThat(graphql.errors?.get(1)?.message).isEqualTo("Validation failed")
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
```diff
@@ -250,6 +250,7 @@ internal class RumResourceScope(
         val graphqlOperationName = resourceAttributes.remove(RumAttributes.GRAPHQL_OPERATION_NAME) as? String
         val graphqlOperationType = resourceAttributes.remove(RumAttributes.GRAPHQL_OPERATION_TYPE) as? String
         val graphqlVariables = resourceAttributes.remove(RumAttributes.GRAPHQL_VARIABLES) as? String
+        val graphqlErrors = resourceAttributes.remove(RumAttributes.GRAPHQL_ERRORS) as? String
```
aleksandr-gringauz: Question. Where is this attribute set? IIUC Android SDK currently doesn't process errors from GraphQL response. 

Or will this PR only work for React Native SDK, because it sets these attributes?

cdn34dd: We're setting these attributes in react-native yes and we need the android SDK to also parse the errors we set in those attributes similar to what's currently done on iOS.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
cdn34dd: The error data comes from react-native (JS side), and due to a bug we just found (that was flattening object properties), I was forced to send the errors object as a stringified JSON which then had to be parsed natively, so I'll double-check this once the fix is merged.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt
cdn34dd: We could, I just left it in case we changed the shape of the data in the future, so I could simply change the RN implementation and not the android-sdk, as in my view it makes more sense to not have a wrapped object `{errors: [...]}` inside a key already named `_dd.graphql.errors`.

PR #3102, opened 2026-01-05
RUM-8949: Enabling UI slow frames by default

satween:
### What does this PR do?
* Enables `UI Slowness` feature by default. 
* Removes `@ExperimentalRumApi` annotation.

### Motivation

Back-end (and web frontend) teams has supported required parts for the slowness feature and ask to start rolling out this feature from mobile SDK side.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
@@ -349,9 +349,10 @@ data class RumConfiguration internal constructor(
          * Assigning a null value to this property will disable the [SlowFramesListener] and stop the computation of the
          * associated rates.
          *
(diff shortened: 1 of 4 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -351,7 +351,6 @@ data class RumConfiguration internal constructor(
          *
          * @param slowFramesConfiguration The configuration to be applied to the [SlowFramesListener].
          */
```
ambushwork, line 355: we have a behavior change but the documentation doesn't imply anything, maybe we should enhance the documentation by saying if the method is not called, the default configuration is applied?

satween: The whole feature was experimental, so I think it's okay, but I just added a line with the clarification for the default behaviour 👍

PR #3109, opened 2026-01-07
RUM-13627: Add sample rate for App launch profiling

ambushwork:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
@@ -15,6 +15,7 @@ import android.database.Cursor
 import android.net.Uri
 import android.os.Build
 import androidx.annotation.RequiresApi
+import com.datadog.android.core.sampling.RateBasedSampler
 import com.datadog.android.profiling.internal.ProfilingStorage
 
 /**
@@ -28,11 +29,20 @@ class DdProfilingContentProvider : ContentProvider() {
             context?.let {
                 val instanceNames = ProfilingStorage.getProfilingEnabledInstanceNames(it)
                 if (instanceNames.isNotEmpty() && isProcessFromLauncher()) {
-                    Profiling.start(it, instanceNames)
+                    sampleProfiling(it, instanceNames)
                 }
             }
         }
         return true
+    }
+
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    private fun sampleProfiling(context: Context, instanceNames: Set<String>) {
+        val sampleRate = ProfilingStorage.getSampleRate(context)
+        if (RateBasedSampler<Unit>(sampleRate).sample(Unit)) {
+            Profiling.start(context, instanceNames)
+        }
+        ProfilingStorage.removeSampleRate(context)
     }
 
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
@@ -6,11 +6,14 @@
 
 package com.datadog.android.profiling
 
+import androidx.annotation.FloatRange
+
 /**
  * Describes configuration to be used for the Profiling feature.
  */
 data class ProfilingConfiguration internal constructor(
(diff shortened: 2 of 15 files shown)


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStorage.kt
```diff
@@ -10,11 +10,22 @@ import android.content.Context
...
     internal var sharedPreferencesStorage: SharedPreferencesStorage? = null
 
+    internal fun setSampleRate(appContext: Context, sampleRate: Float) {
+        getStorage(appContext).apply {
+            putFloat(KEY_PROFILING_SAMPLE_RATE, sampleRate)
+        }
+    }
+
+    internal fun getSampleRate(appContext: Context): Float {
+        return getStorage(appContext).getFloat(KEY_PROFILING_SAMPLE_RATE, 0f)
```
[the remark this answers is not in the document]
ambushwork: won't be an issue since we always profile for next app launch

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
```diff
@@ -28,13 +29,21 @@ class DdProfilingContentProvider : ContentProvider() {
...
+                    sampleProfiling(it, instanceNames)
                 }
             }
         }
         return true
     }
 
+    @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    private fun sampleProfiling(context: Context, instanceNames: Set<String>) {
+        val sampleRate = ProfilingStorage.getSampleRate(context)
```
0xnm, line 41: Storing sample rate disconnected from the SDK instance it set creates the following problem: suppose there is instance A with sample rate 10%, and instance B with sample rate 40%. Instance B is initialized after instance A, so the sample rate recorded will be 40%.

It means that instance A will have 40% sample rate, increasing the billing and exploding the profile volume for this org.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
```diff
@@ -34,13 +50,16 @@ data class ProfilingConfiguration internal constructor(
...
-                customEndpointUrl = customEndpointUrl
+                customEndpointUrl = customEndpointUrl,
+                sampleRate = sampleRate
             )
         }
     }
 
     companion object {
 
+        private const val DEFAULT_SAMPLE_RATE = 100f
```
0xnm, line 61: I don't think it should be 100% by default.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
ambushwork: I updated the logic of persisting the sample rate, now we set sample rate of configuration only if the sample rate is absent in sp or the old value is bigger than the new value.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/ProfilingConfiguration.kt
ambushwork: For me 100f means let rate limiter handle the sampling, which is kind of OK, what value do you suggest?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
0xnm: But this new logic also means if user wants to set sampling rate higher, nothing will be done, isn't it?

ambushwork: Sample rate is also an one-time flag, once it's consumed, it will be removed, see [here](https://github.com/DataDog/dd-sdk-android/blob/e699abe1a47d5f30538d336e8207572d18c6c25c/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt#L67). then user can set another sample rate, lower or higher. one thing may miss here is that, if the session is sampled out, then set a lower sample rate will be ignored

ambushwork: I updated the logic when the profiling is sampled out, the sample rate will be removed

PR #3117, opened 2026-01-08
Address concurrency and app start detection issues in Profiling

0xnm:
### What does this PR do?

This PR does the following:

* Fixes some threading/concurrency issues due to the properties accessed from different threads: profiler callback thread vs SDK thread
* Fixes app start detection logic: we were checking if any historical reason is from the launcher, we need to check only the current (most recent one).

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
@@ -45,6 +45,7 @@ internal class PerfettoProfiler(
     // This flag represents which instance of this class is working for.
     private val runningInstances: AtomicReference<Set<String>> = AtomicReference(emptySet())
 
+    @Volatile
     private var profilingStartTime = 0L
 
     override var internalLogger: InternalLogger? = null
@@ -58,6 +59,7 @@ internal class PerfettoProfiler(
             val endTime = timeProvider.getDeviceTimestampMillis()
             val duration = endTime - profilingStartTime
             if (result.errorCode == ProfilingResult.ERROR_NONE) {
+                // TODO RUM-13679: need to delete the file after it is no longer needed
                 result.resultFilePath?.let {
                     notifyAllCallbacks(
                         PerfettoResult(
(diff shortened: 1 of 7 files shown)


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -58,6 +59,7 @@ internal class PerfettoProfiler(
             val endTime = timeProvider.getDeviceTimestampMillis()
             val duration = endTime - profilingStartTime
             if (result.errorCode == ProfilingResult.ERROR_NONE) {
+                // TODO RUM-13679: need to delete the file after it is no longer needed
```
ambushwork, line 62: ProfilingManager stores the result as a temp file and it will be cleaned up automatically, I don't think we need to handle it by us.

0xnm: > it will be cleaned up automatically

do you know the cleanup logic? because we store the path to the file, so is there any risk that by the time we try to read from this path file is not there already?

the location given is not in the cache folder

ambushwork: I don't know how it is cleaned under the hood exactly, but I guess we will probably not hit the this risk since from the callback to writing batch, it has very small delay (although we switch the thread).

 I think we can add a telemetry to monitor if we can actually hit the case where we fail the read the result with the this path.

0xnm: From my runs I see that profiling results are stored on the device like forever. Maybe they are cleaned up though once `/data/data/<app>/files/profiling` folder hits certain size.

ambushwork: Google claims that to keep the results forever we need to enable a flag through adb, let's confirm with them about this.

PR #3131, opened 2026-01-16
[Flags] fix: add STATIC resolution reason

typotter:
### What does this PR do?

This PR adds the STATIC resolution reason to the `flags` and `flags-openfeature` modules

### Motivation

This resolution reason is returned from the flagging service when a flag's value is determined by the platform default rather than any targeting rules

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ResolutionReason.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ResolutionReason.kt
@@ -14,12 +14,14 @@ package com.datadog.android.flags.model
  */
 enum class ResolutionReason {
     /**
-     * The resolved value is a static value configured for the flag.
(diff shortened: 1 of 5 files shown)


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/ResolutionReason.kt
```diff
@@ -14,7 +14,14 @@ package com.datadog.android.flags.model
  */
 enum class ResolutionReason {
     /**
-     * The resolved value is the default value for the flag.
+     * No targeting rules matched; the fallthrough/default allocation was used.
```
leoromanovsky, line 17: another reason is the "allocation override" that forces a variant value

leoromanovsky (approved):

0xnm (approved):

jonathanmos (approved):

PR #3132, opened 2026-01-19
[flags] fix: use read lock for removeListener

typotter:
### What does this PR do?

Uses a `read` lock in the `removeListener` method.
This locking fixes a race condition between `removeListener` and `updateState`

### Motivation

Removing a listener needs to block state changes so that removed listeners do not receive notifications after removal.

### Additional Notes


features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -125,21 +125,21 @@ internal class FlagsStateManagerTest {
...
     @Test
-    fun `M stop notifying subsequent listeners W updateState() { if listener throws }`() {
+    fun `M stop notifying subsequent listeners W updateState() { listener throws }`() {
         // Given
         val executionOrder = mutableListOf<String>()
 
         val listener1 = object : FlagsStateListener {
             override fun onStateChanged(newState: FlagsClientState) {
-                if (newState is FlagsClientState.Ready) {
+                if (newState == FlagsClientState.Ready) {
```
typotter, line 135: `FlagClientState` is a sealed class

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -125,21 +125,21 @@ internal class FlagsStateManagerTest {
     }
 
     @Test
-    fun `M stop notifying subsequent listeners W updateState() { if listener throws }`() {
+    fun `M stop notifying subsequent listeners W updateState() { listener throws }`() {
```
typotter, line 129: polishing some names/comments while here

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -229,65 +221,145 @@ internal class FlagsStateManagerTest {
...
         }
 
-        // Verify final state
         assertThat(testedManager.getCurrentState()).isEqualTo(stateNew)
     }
 
+    @Test
+    fun `M block getCurrentState calls W addListener() { slow listener notification }`() {
+        // Given
+        val stateOld = FlagsClientState.NotReady
```
0xnm: nit: probably should be `oldState` to be consistent with naming convention used in `newState`.

Or maybe for the clarity we can just inline `FlagsClientState.NotReady` where needed, because `stateOld` is `val` anyway and cannot be changed.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -229,65 +221,145 @@ internal class FlagsStateManagerTest {
...
+            testedManager.getCurrentState()
+            synchronized(operationTimestamps) {
+                operationTimestamps.add("getCurrentState completed" to System.nanoTime())
+            }
+        }
+
+        addListenerThread.start()
+        addListenerStarted.await()
+        getCurrentStateThread.start()
+        getCurrentStateAttempted.await()
```
0xnm: `getCurrentStateAttempted` is not actually needed I think,  the necessary wait for the `getCurrentStateThread` completion is already implemented by the `getCurrentStateThread.join()` call below.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -229,65 +221,145 @@ internal class FlagsStateManagerTest {
...
+    @Test
+    fun `M block getCurrentState calls W addListener() { slow listener notification }`() {
+        // Given
+        val stateOld = FlagsClientState.NotReady
+
+        val addListenerStarted = CountDownLatch(1)
+        val addListenerSlowCallbackStarted = CountDownLatch(1)
+        val getCurrentStateAttempted = CountDownLatch(1)
+
+        val operationTimestamps = mutableListOf<Pair<String, Long>>()
```
0xnm: `synchronized` calls can be removed if type here is `CopyOnWriteArrayList`

aleksandr-gringauz (dismissed):

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -229,65 +221,145 @@ internal class FlagsStateManagerTest {
...
-            // The states should be received in chronological order
             assertThat(oldStateTime).isLessThan(newStateTime)
         }
 
-        // Verify final state
         assertThat(testedManager.getCurrentState()).isEqualTo(stateNew)
     }
 
+    @Test
+    fun `M block getCurrentState calls W addListener() { slow listener notification }`() {
```
0xnm: Why do we want to block `getCurrentState` when `addListener` is called, if `addListener` is not changing the state?

typotter: 🤔  We may have had the right lock here in the first place...

If `addListener` holds the `read` lock, then, ostensibly, nothing can write to the state (current state or the listeners). While `addListener` is technically mutating the protected state by adding a listener, the underlying mechanism itself is theadsafe/synchronized via `CopyOnWriteArrayList` so it's fine if parallel calls to `addListener` attempt to mutate the set of listeners. The order of listener subscription is not important to maintain here.

We should just be able to rely on the underlying thread-safety of `CopyOnWriteArrayList` to avoid clobbering of the listener list, right?

satween: Maybe we shouldn't wrap adding listener with lock then?

typotter: We need to block writing to the current state until the first notification and `addListener` are complete. We don't get that atomicity with `DDCoreSubscription`, only that the calls underlying calls to add the listener to the list are thread safe and synchronized so no listener will be lost when parallel calls are made.

In effect, I believe we can drop this change altogether and rely on the `read` lock in conjunction with the underlying lock in `CopyOnWriteArrayList`

0xnm: I think yes, we should. My thinking is the following:

* `getCurrentState` covers only `FlagsStateManager.currentState`. Listeners is not a part of `FlagsStateManager.currentState`, so they are not related to the `getCurrentState` call.
* `FlagsStateManager.addListener` call doesn't modify `FlagsStateManager.currentState`, it only reads. So that us why we added a read lock. If something is in the process of modifying `FlagsStateManager.currentState`, then there will be a wait to acquire `read` lock.
* `FlagsStateManager.addListener` mutates underlying listeners collection, but anyway this is atomic and thread-safe, `currentState` value passed down is guarded by the read lock and if there is iteration already happening in `updateState -> subscription.notifyListeners`, then this new listener won't be a part of ongoing iteration (due to the usage of `CopyOnWriteArrayList` in `DDSubscription`).

So I'm curious why should we block `getCurrentState` call during `FlagsStateManager.addListener` invocation.

> The race condition could otherwise result in a listener missing a state change

Is it about `FlagsStateManager.currentState` property here? From the code I see, listener shouldn't miss any state change, since read lock cannot be acquired if write lock is active, and the opposite.

satween: Well, having read it more carefully, I agree with @nikita. The only thing that still makes me curious is the fact that we adding listener with a lock, but removing without. Yes, underlying `CopyOnWriteArrayList` will protect us from the `ConcurrentModificationException`, however if `updateState` and `removeListener` will be executed at the same time, removed listener could get an undesired update (from  subscription.notifyListeners ) and could lead to undefined behavior. Not sure if this is a critical issue, but I'd prefer to maintain consistency here.

typotter: Thanks @satween. Agreed that there's an apparent lack of synchronicity on removing a listener.

We are not just "Adding a listener", however.
We need to 
- get the current state
- notify the listener
- add the listener to the list
All without the `currentState` changing, so we block changes with a `read` lock.

When we remove the listener, have only that operation to complete so there isn't a risk of missing data, only, as you noted, receiving extra data. We would need to add an extra synchronization shared between the `removeListener` method and the `updateState` method. I think you're correct here though - a long-running listener callback could cause a removed listener to be notified _after_ the call to _removeListener_ completes

The scale of adding/removing listeners will be very low (0, or 1 or maybe a few) so there is a very small risk here of a superfluous event.

I'll come back to this with an updated lock after a couple of other tasks

typotter: Updated to use a `read` lock on both add and remove listener to ensure that state changes are not sent to removed listeners

typotter:
Updated the locks. PTAL

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -227,64 +219,123 @@ internal class FlagsStateManagerTest {
...
+        updateStateThread.start()
+        updateStateStarted.await()
+        removeListenerThread.start()
+
+        updateStateThread.join(5000)
+        removeListenerThread.join(5000)
+
+        // Then
+        // After removeListener returns, no more notifications should be received
+        synchronized(notificationsAfterRemove) {
```
0xnm: I don't quite understand why do we need `synchronized` in the assertions block. Assertions are supposed to be executed only when everything is completed and `updateStateThread.join()` should guarantee safe read without this block anyway (and it is not needed if `CopyOnWriteArrayList` is used as well).

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -227,64 +219,123 @@ internal class FlagsStateManagerTest {
...
+                }
+            }
+        }
+
+        // Another listener to track notifications after removal
+        val trackingListener = object : FlagsStateListener {
+            override fun onStateChanged(newState: FlagsClientState) {
+                // Only track notifications that happen after removeListener completed
+                if (removeListenerCompleted.count == 0L) {
+                    synchronized(notificationsAfterRemove) {
```
0xnm: A bit more fair will be to use `CopyOnWriteArrayList` for `notificationsAfterRemove` instead of using  `synchronized` here.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -227,64 +219,123 @@ internal class FlagsStateManagerTest {
...
+        val updateStateStarted = CountDownLatch(1)
+        val slowCallbackStarted = CountDownLatch(1)
+        val removeListenerCompleted = CountDownLatch(1)
+
+        val notificationsAfterRemove = mutableListOf<FlagsClientState>()
+
+        // Listener that is slow during updateState notification
+        val slowListener = object : FlagsStateListener {
+            override fun onStateChanged(newState: FlagsClientState) {
+                if (newState == stateNew) {
```
0xnm: nit: having `newState` and `stateNew` is a bit confusing. Maybe we should rename `stateNew` to `readyState`?

0xnm (dismissed):

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/FlagsStateManagerTest.kt
```diff
@@ -209,83 +203,128 @@ internal class FlagsStateManagerTest {
...
+            initialState, // From addListener's initial notification (slow)
+            updatedState // From updateState (blocked until addListener completes)
+        )
+
+        val initialStateTime = receivedStates[0].second
+        val updatedStateTime = receivedStates[1].second
+
+        // updateState should have been delayed by at least the sleep time
+        assertThat(updatedStateTime - initialStateTime).isGreaterThan(10_000_000) // ~10ms in nanoseconds
+        assertThat(initialStateTime).isLessThan(updatedStateTime)
```
0xnm: nit: the line above already imposes that `initialStateTime` is less than `updatedStateTime`. If it is not the case, this line won't be even executed.

0xnm (dismissed):

0xnm (approved):

PR #3136, opened 2026-01-20
RUM-13568: [Cronet]: Supporting RUM <> APM integration for Cronet

satween:
### What does this PR do?                                                                                                                                                    
                                                                                                                                                                               
  In this PR we are copying the main logic from `TracingInterceptor` / `DatadogInterceptor` into the `NetworkTracingInstrumentation` in order to:                              
  1. Support RUM <> APM integration for `Cronet`                                                                                                                               
  2. Create a foundation for single network instrumentation logic that should be shared across several networking libraries in the future.                                     
                                                                                                                                                                               
  ### Motivation                                                                                                                                                               
                                                                                                                                                                               
  Continuing to add an integration for `Cronet` library meanwhile preparing infrastructure to migrate `OkHttp` onto the same logic.                                            
                                                                                                                                                                               
  ### Additional Notes                                                                                                                                                         
                                                                                                                                                                               
  Note that in this PR we are not supporting deep network instrumentation that is being done by `TracingInterceptor` in `OkHttp`. This logic will be supported in a further PR as this one is already pretty big.                                                                                                                                           
                                                                                                                                                              
**NOTE:** In this PR as an experiment an AI agent helped to write tests and some docs. I've made few iterations of a fixes for that but if you spot any issues - let me know. AI wasn't involved into production code development.
                

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogUrlRequestBuilder.kt
```diff
@@ -32,8 +34,17 @@ internal class DatadogUrlRequestBuilder(
     }
 
     override fun addHeader(header: String, value: String?): UrlRequest.Builder = apply {
-        headers[header] = listOfNotNull(value)
-        delegate.addHeader(header, value)
+        val currentValues = headers[header]?.toMutableList() ?: mutableListOf()
+        if (value != null) {
+            currentValues.add(value)
+        }
+        headers[header] = currentValues
```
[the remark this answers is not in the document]
satween: Interesting catch, but the problem was bit deeper, fixed

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/LangExt.kt
```diff
@@ -0,0 +1,14 @@
...
+ */
+package com.datadog.android.internal.utils
+
+/**
+ * Executes the given [block] if this value is null.
+ * @param T the type of the value.
+ * @param block the action to execute if this is null.
+ * @return this value unchanged.
+ */
+fun <T> T?.runIfNull(block: () -> Unit): T? = apply { if (this == null) block() }
```
ambushwork: Can we just use simple if condition for this?
```
if(T == null) {
     block()
}
```
You create a extension function and a lot of unit tests around it, seems a bit overkill

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumResourceInstrumentation.kt
```diff
@@ -166,6 +172,14 @@ class RumResourceInstrumentation(
             return ResourceId(key, uuid?.toString())
         }
 
+        /**
+         * Creates a new [com.datadog.android.rum.configuration.RumResourceInstrumentationConfiguration] for configuring RUM resource tracking.
+         * @return a new builder instance.
+         */
+        @JvmStatic
+        @Suppress("FunctionName")
+        fun Configuration() = RumResourceInstrumentationConfiguration()
```
ambushwork: Question, why we need such a function instead of just building `RumResourceInstrumentationConfiguration` directly?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,286 @@
...
+import com.datadog.android.trace.internal.net.applyPriority
+import com.datadog.android.trace.internal.net.buildSpan
+import com.datadog.android.trace.internal.net.finishRumAware
+import com.datadog.android.trace.internal.net.isRumEnabled
+import com.datadog.android.trace.internal.net.sample
+import java.net.HttpURLConnection
+import java.util.Locale
+
+/**
+ * For internal usage only.
```
ambushwork, line 40: should we mark with @InternalApi in this case?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/cronet/CronetImageFragment.kt
```diff
@@ -41,6 +43,12 @@ internal class CronetImageFragment : Fragment() {
         "https://storage.googleapis.com/cronet/walnut.jpg"
     )
 
+    private val tracedHosts = listOf(
+        "datadoghq.com",
+        "127.0.0.1",
+        "storage.googleapis.com"
```
0xnm, line 48: is it added for some validation? we don't own this host, so sending trace context there is useless

dd-sdk-android-core/api/apiSurface
```diff
@@ -158,6 +159,14 @@ interface com.datadog.android.api.instrumentation.network.HttpRequestInfo
   val contentType: String?
   val method: String
   fun contentLength(): Long?
+  fun modify(): HttpRequestInfoModifier
```
0xnm: is this one supposed to be called by the user? probably it is for the internal usage only?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/instrumentation/network/HttpRequestInfoModifier.kt
```diff
@@ -0,0 +1,67 @@
...
+     * @param tag the tag value, or null to remove the tag.
+     * @return this modifier for chaining.
+     */
+    fun <T> addTag(type: Class<in T>, tag: T?): HttpRequestInfoModifier
+
+    /**
+     * Builds and returns the modified [HttpRequestInfo].
+     * @return the resulting [HttpRequestInfo] with all modifications applied.
+     */
+    fun result(): HttpRequestInfo
```
0xnm: maybe canonically it should be called `build`?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/LangExt.kt
0xnm: agree on that, seems unnecessary

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
```diff
@@ -130,5 +132,8 @@ class _RumInternalProxy internal constructor(private val rumMonitor: AdvancedRum
         ): Builder {
             return builder.setInsightsCollector(insightsCollector)
         }
+
+        fun RumResourceInstrumentationConfiguration?.build(name: String): RumResourceInstrumentation? =
```
0xnm: why is it a part of `_RumInternalProxy`? how it is supposed to be called?

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -224,11 +228,12 @@ class com.datadog.android.rum.internal.net.RumResourceInstrumentation
   fun sendWaitForResourceTimingEvent(com.datadog.android.api.instrumentation.network.HttpRequestInfo)
   fun sendTiming(com.datadog.android.api.instrumentation.network.HttpRequestInfo, com.datadog.android.rum.internal.domain.event.ResourceTiming)
   fun startResource(com.datadog.android.api.instrumentation.network.HttpRequestInfo)
-  fun stopResource(com.datadog.android.api.instrumentation.network.HttpRequestInfo, com.datadog.android.api.instrumentation.network.HttpResponseInfo)
+  fun stopResource(com.datadog.android.api.instrumentation.network.HttpRequestInfo, com.datadog.android.api.instrumentation.network.HttpResponseInfo, Map<String, Any?> = emptyMap())
   fun stopResourceWithError(com.datadog.android.api.instrumentation.network.HttpRequestInfo, Throwable)
   fun reportInstrumentationError(String)
   companion object 
     fun buildResourceId(com.datadog.android.api.instrumentation.network.HttpRequestInfo, Boolean): com.datadog.android.rum.resource.ResourceId
+    fun Configuration()
```
0xnm: doesn't seem right

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/RumResourceInstrumentationConfiguration.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.datadog.android.rum.NoOpRumResourceAttributesProvider
+import com.datadog.android.rum.RumResourceAttributesProvider
+import com.datadog.android.rum.internal.net.RumResourceInstrumentation
+
+/**
+ * Builder for creating [com.datadog.android.rum.internal.net.RumResourceInstrumentation] instances.
+ * This builder allows configuring RUM resource tracking for network requests.
+ *
+ * Use [com.datadog.android.rum.internal.net.RumResourceInstrumentation.Companion.Configuration] to create instances of this builder.
+ */
```
0xnm, line 15: this description seems wrong, it is not a builder class

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/net/RumInstrumentationConfigurationTest.kt
```diff
@@ -0,0 +1,91 @@
...
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class RumResourceInstrumentationBuilderTest {
+
+    private lateinit var testedBuilder: RumResourceInstrumentationConfiguration
```
0xnm: this is not a builder class

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/B3HttpCodec.java
```diff
@@ -23,17 +23,17 @@
 import com.datadog.android.trace.internal.compat.function.Supplier;
 
 /** A codec designed for HTTP transport via headers using B3 headers */
-class B3HttpCodec {
+public class B3HttpCodec {
```
0xnm, line 26: do we need to expose the whole class? we can create a Kotlin class e.g. `B3Headers` and reference there the neccary:

```
class B3Headers {
  const val TRACE_ID_KEY = B3Codec.TRACE_ID_KEY
}
```

Or maybe we need to push more things to `dd-sdk-android-trace-internal`, so that we don't need to make things public?

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -1,12 +1,36 @@
...
   fun getCoreTracerWriter(): com.datadog.android.trace.api.span.DatadogSpanWriter
+interface com.datadog.android.trace.NetworkTracedRequestListener
+  fun onRequestIntercepted(com.datadog.android.api.instrumentation.network.HttpRequestInfo, com.datadog.android.trace.api.span.DatadogSpan, com.datadog.android.api.instrumentation.network.HttpResponseInfo?, Throwable?)
+class com.datadog.android.trace.NetworkTracingInstrumentation
+  fun onRequest(com.datadog.android.api.instrumentation.network.HttpRequestInfo): com.datadog.android.trace.internal.net.RequestTraceState
+  fun onResponseSucceed(com.datadog.android.trace.internal.net.RequestTraceState, com.datadog.android.api.instrumentation.network.HttpResponseInfo)
+  fun onResponseFailed(com.datadog.android.trace.internal.net.RequestTraceState, Throwable)
+  companion object 
+    fun Configuration(Map<String, Set<TracingHeaderType>>)
+    fun Configuration(List<String>)
```
0xnm: seems not right as well

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/OkHttpHttpRequestInfo.kt
```diff
@@ -37,4 +39,33 @@ internal class OkHttpHttpRequestInfo(internal val request: Request) : HttpReques
     } catch (@Suppress("SwallowedException") _: IOException) {
         null
     }
+
+    override fun modify(): HttpRequestInfoModifier = OkHttpHttpRequestInfoModifier(request.newBuilder())
```
0xnm: We need to find a way to not expose this in the user-facing API. Maybe some cast? Like `HttpRequestInfo` and `MutableHttpRequestInfo`? Then SDK can make a cast considering that `MutableHttpRequestInfo` is always implemented.

integrations/dd-sdk-android-cronet/api/apiSurface
```diff
@@ -24,9 +24,9 @@ class com.datadog.android.cronet.DatadogCronetEngine : org.chromium.net.CronetEn
   override fun removeThroughputListener(org.chromium.net.NetworkQualityThroughputListener?)
   class Builder : org.chromium.net.CronetEngine.Builder
     constructor(android.content.Context, org.chromium.net.CronetEngine.Builder = CronetEngine.Builder(context))
-    fun setRumResourceAttributesProvider(com.datadog.android.rum.RumResourceAttributesProvider)
-    fun setSdkInstanceName(String)
```
0xnm, line 28: this makes it a breaking change, are we ok with that?

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetRequestContext.kt
```diff
@@ -0,0 +1,219 @@
...
+@Suppress("TooManyFunctions")
+internal class DatadogCronetRequestContext private constructor(
+    internal var url: String,
+    private val executor: Executor,
+    private val engine: DatadogCronetEngine,
+    private val datadogRequestCallback: DatadogRequestCallback,
+    private val requestParams: CronetRequestParams,
+    private val additionalAnnotations: MutableMap<Class<*>, Any>
+) {
+    internal constructor(
```
0xnm: why do we need a private constructor as well?

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetRequestContext.kt
```diff
@@ -0,0 +1,219 @@
...
+
+    internal fun buildRequestInfo() = CronetHttpRequestInfo(this)
+}
+
+private fun UrlRequest.Builder.applyRequestParams(params: CronetRequestParams) = apply {
+    setHttpMethod(params.method)
+    addHeaders(params.headers)
+    if (params.disableCache) disableCache()
+    if (params.allowDirectExecutor) allowDirectExecutor()
+    params.priority?.let(::setPriority)
```
0xnm, line 152: how do we keep all this intact if there is new API?

sample/kotlin/src/main/kotlin/com/datadog/android/sample/cronet/CronetImageFragment.kt
satween: it's added in order to check that cronet is working with QUIC, RUM <> APM integration will not work for this host, but I want to keep this for sample app in order to send spans for it ( and check that APM is working with QUICK)

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/propagation/B3HttpCodec.java
satween: TBH I was aiming the opposite goal - "stop repeating constants everywhere".
It's becomes harder to support logic across different modules and not forgetting to add constant values into all required places. Another point - it's hard to bound components and understand what class is responsible for the specific header. 

Making this class public should not affect user too much because: 
1. This constant only exposed to the `-trace` module that already has an access to the `-internal` module.
2. Package-private classes are still possible to access from java (by creating same package and put public class/method there) 
3. We don't add this module directly to the user's codebase

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/OkHttpHttpRequestInfo.kt
satween: Hmm, maybe it's a good step, changed

integrations/dd-sdk-android-cronet/api/apiSurface
satween: yes, as cronet support is experimental for now

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumResourceInstrumentation.kt
satween: I was trying to minimize the amount of letters, required to create an instance of instrumentation, but such approach worser integrates with the IDE, fixed

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/instrumentation/network/HttpRequestInfoModifier.kt
satween: Renamed to HttpInfoRequestBuilder and build method respectively

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
satween: I want to hide `build` method from users. This made for several reasons: 
1. build() - basically useless method, it does not add any information to the builder
2. Integrations (`Cronet` and `OkHttp`) should set their names into the instrumentation. 

>  how it is supposed to be called?
```kotlin
val rumResourceInstrumentation = with(_RumInternalProxy) {       
  rumInstrumentationConfiguration.build(CRONET_NETWORK_INSTRUMENTATION_NAME)
}
```

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetRequestContext.kt
satween: it was needed because I wanted to use private classes inside the constructor. But seems like it confuses, so switched back to internal

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetRequestContext.kt
satween: [CronetApiInstrumentationTest](https://github.com/DataDog/dd-sdk-android/blob/cd6029be95755b3d342808bcaa63021beb73e6ee/integrations/dd-sdk-android-cronet/src/test/kotlin/com/datadog/android/cronet/internal/CronetApiInstrumentationTest.kt#L16) will catch api changes for cronet

features/dd-sdk-android-rum/api/apiSurface
satween: fixed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/net/RumInstrumentationConfigurationTest.kt
satween: Android studio not always catches renaming for some reason

dd-sdk-android-core/api/apiSurface
satween: marked it for internal usage, but it should remain public in order to access from different modules

0xnm: Does this comment consider the thread here? https://github.com/DataDog/dd-sdk-android/pull/3136#discussion_r2725942000

It shouldn't be a part of public API by default. If we want to access this method for the instrumentation purpose, we need to downcast explicitly then.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/instrumentation/network/HttpRequestInfoBuilder.kt
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package com.datadog.android.api.instrumentation.network
+
+/**
+ * A builder interface for modifying [com.datadog.android.api.instrumentation.network.HttpRequestInfo] instances.
+ * This interface allows modifying HTTP request properties such as URL, headers, and tags
+ * before the request is processed by the instrumentation.
+ *
+ * Use [com.datadog.android.api.instrumentation.network.HttpRequestInfo.modify] to obtain a modifier instance.
```
0xnm: Need to update this doc, there is no `modify` method anymore.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/instrumentation/network/HttpRequestInfoBuilder.kt
```diff
@@ -0,0 +1,68 @@
...
+ */
+interface HttpRequestInfoBuilder {
```
0xnm, line 21: should this be annotated with `@InternalApi` as well?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
0xnm: I'm wondering how it looks like in a java bytecode. Are you sure that `rumInstrumentationConfiguration.build(CRONET_NETWORK_INSTRUMENTATION_NAME)` call is not possible outside of `_RumInternalProxy` receiver scope?

First time I see such trick given that visibility of extension function is public.

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -1,12 +1,30 @@
+class com.datadog.android.trace.ApmNetworkInstrumentationConfiguration
+  constructor(List<String>)
+  fun setTraceOrigin(String)
+  fun setSdkInstanceName(String)
+  fun setTracedRequestListener(NetworkTracedRequestListener)
+  fun setTraceSampleRate(Float)
+  fun setTraceSampler(com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>)
+  fun setTraceContextInjection(TraceContextInjection)
+  fun set404ResourcesRedacted(Boolean)
+  fun disableNetworkLayerInstrumentation()
```
0xnm: This method is a bit confusing, given that the name of the class `ApmNetworkInstrumentationConfiguration` is already about network.

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -1,12 +1,30 @@
...
+  fun disableRumToTracesLinking()
+  companion object
```
0xnm: nit: can be made internal, same applies for other 2 similar instance below.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -0,0 +1,197 @@
...
+import com.datadog.android.core.configuration.HostsSanitizer
+import com.datadog.android.core.internal.net.DefaultFirstPartyHostHeaderTypeResolver
+import com.datadog.android.core.sampling.Sampler
+import com.datadog.android.trace.api.span.DatadogSpan
+import com.datadog.android.trace.api.tracer.DatadogTracer
+import com.datadog.android.trace.internal.ApmNetworkInstrumentation
+import com.datadog.android.trace.internal.net.TracerProvider
+
+/**
+ * Configuration for creating [com.datadog.android.trace.internal.ApmNetworkInstrumentation] instances.
```
0xnm: should we reference internal classes in the description of the public class? similar question is for `RumInstrumentationConfiguration`

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -0,0 +1,197 @@
...
+    internal var redacted404ResourceName = true
+    internal var sdkInstanceName: String? = null
+
+    internal var localTracerFactory = DEFAULT_LOCAL_TRACER_FACTORY
+    internal var traceContextInjection = TraceContextInjection.SAMPLED
+    internal var tracedRequestListener: NetworkTracedRequestListener = NoOpNetworkTracedRequestListener()
+    internal var traceSampler: Sampler<DatadogSpan> =
+        DeterministicTraceSampler(DEFAULT_TRACE_SAMPLE_RATE)
+    internal var globalTracerProvider: () -> DatadogTracer? = { GlobalDatadogTracer.getOrNull() }
+    internal var rumApmLinkingEnabled = true
```
0xnm: nit: it is called `bundle` in the existing API

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -0,0 +1,197 @@
...
+    internal var networkInstrumentationEnabled: Boolean = true
+
+    constructor(
```
0xnm, line 42: when do we expect this constructor to be called?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,290 @@
...
+        internal const val SPAN_NAME = "%s.request"
+        internal const val RESOURCE_NAME_404 = "404"
+        internal const val AGENT_PSR_ATTRIBUTE = "_dd.agent_psr"
+        internal const val URL_QUERY_PARAMS_BLOCK_SEPARATOR = '?'
+        internal const val WARNING_TRACING_NO_HOSTS =
+            "You added a TraceInstrumentation to your %s instrumentation, " +
+                "but you did not specify any first party hosts. " +
+                "Your requests won't be traced.\n" +
+                "To set a list of known hosts, you can use the " +
+                "Configuration.Builder::setFirstPartyHosts() method."
```
0xnm:
```suggestion
                "Configuration.Builder.setFirstPartyHosts() method."
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,290 @@
...
+
+    companion object {
+        internal const val ZERO_SAMPLE_RATE: Float = 0.0f
+        internal const val ALL_IN_SAMPLE_RATE: Double = 100.0
+        internal const val SPAN_NAME = "%s.request"
+        internal const val RESOURCE_NAME_404 = "404"
+        internal const val AGENT_PSR_ATTRIBUTE = "_dd.agent_psr"
+        internal const val URL_QUERY_PARAMS_BLOCK_SEPARATOR = '?'
+        internal const val WARNING_TRACING_NO_HOSTS =
+            "You added a TraceInstrumentation to your %s instrumentation, " +
```
0xnm: Seems there is no class called `TraceInstrumentation`

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,290 @@
...
+                "onRequestIntercepted callback, leading to infinite recursion."
+
+        internal const val ERROR_REQUEST_INFO_IS_NOT_MUTABLE =
+            "RequestInfo is not mutable. Your requests won't be traced."
+
+        private fun HttpRequestInfo.newBuilder(internalLogger: InternalLogger): HttpRequestInfoBuilder? {
+            return if (this is MutableHttpRequestInfo) {
+                newBuilder()
+            } else {
+                internalLogger.logToUser(InternalLogger.Level.ERROR) { ERROR_REQUEST_INFO_IS_NOT_MUTABLE }
```
0xnm: I don't think user can do anything with this. This is rather maintainer log, so that classes should conform to `MutableHttpRequestInfo` at the SDK build time.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,290 @@
...
+    private fun RequestTraceState.onRequestCompleted(
+        response: HttpResponseInfo?,
+        throwable: Throwable?
+    ) {
+        if (!isSampled || span == null) return
+        val request = requestBuilder.build()
+        try {
+            tracedRequestListener.onRequestIntercepted(request, span, response, throwable)
+        } catch (e: StackOverflowError) {
+            getSdkCoreOrNull(request.url)?.internalLogger?.log(
```
0xnm: this will print `skipping tracking of request...`, which is not true, we started tracking already, but this doesn't really matter since it is `Error` propagated further.

0xnm (commented):
I partially covered this PR, but will continue review later.

integrations/dd-sdk-android-cronet/api/apiSurface
```diff
@@ -24,9 +24,9 @@ class com.datadog.android.cronet.DatadogCronetEngine : org.chromium.net.CronetEn
...
+    fun setCustomRumInstrumentation(com.datadog.android.rum.configuration.RumResourceInstrumentationConfiguration?)
     fun setListenerExecutor(java.util.concurrent.Executor)
+    fun enableApmInstrumentation(com.datadog.android.trace.ApmNetworkInstrumentationConfiguration)
```
0xnm: There is inconsistency in naming between the two: `setCustomRumInstrumentation` vs `enableApmInstrumentation`

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetRequestContext.kt
```diff
@@ -0,0 +1,214 @@
...
+    data class UploadDataProviderParams(
+        val uploadDataProvider: UploadDataProvider?,
+        val executor: Executor?
+    )
+
+    data class RawCompressionDictionary(
+        val dictionarySha256Hash: ByteArray?,
+        val dictionary: ByteBuffer?,
+        val dictionaryId: String?
+    ) {
```
0xnm: should this be a data class? it involves byte comparison which may be quite heavy

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,290 @@
...
+            return RequestTraceState(requestInfoBuilder)
+        }
+
+        val tracer = tracerProvider.provideTracer(
+            sdkCore,
+            localFirstPartyHostHeaderTypeResolver.getAllHeaderTypes(),
+            networkingLibraryName
+        )
+
+        if (tracer == null || !networkInstrumentationEnabled || !request.isTraceable(sdkCore)) {
```
0xnm: nit: so maybe we can lift `!networkInstrumentationEnabled || !request.isTraceable(sdkCore)` just above, next to the `sdkCore == null` check

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,290 @@
...
+            .sample(request, traceSampler)
+
+    /**
+     * Called when a network request succeeds.
+     * This method updates the span with response information and finishes it.
+     *
+     * @param tracingState the tracing state from [onRequest].
+     * @param response the HTTP response information.
+     */
+    fun onResponseSucceed(tracingState: RequestTraceState, response: HttpResponseInfo) {
```
0xnm: nit: `onResponseSucceeded ` (for consistency with `Failed`)

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Locale
+import java.util.concurrent.atomic.AtomicReference
+
+internal class TracerProvider internal constructor(
+    private val localTracerFactory: (SdkCore, Set<TracingHeaderType>) -> DatadogTracer,
+    private val globalTracerProvider: () -> DatadogTracer?
+) {
+    private val localTracerReference: AtomicReference<DatadogTracer> = AtomicReference()
+
+    @Synchronized
```
0xnm, line 24: is this annotation needed?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+    @Suppress("UnsafeThirdPartyFunctionCall") // updateAndGet is safe
```
0xnm: there is no `updateAndGet` usage though

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+                        WARNING_DEFAULT_TRACER.format(Locale.US, networkingLibraryName)
+                    }
+                }
+                return localTracerReference.get()
+            }
+        }
+    }
+
+    internal companion object {
+        const val WARNING_TRACING_DISABLED = "You added a TracingInstrumentation to your %s, " +
```
0xnm: no such class

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+            "but you did not enable the TracingFeature. " +
+            "Your requests won't be traced."
+
+        const val WARNING_DEFAULT_TRACER =
+            "You added a TracingInstrumentation to your %s instrumentation, " +
```
0xnm: no such class

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
```diff
@@ -0,0 +1,74 @@
...
+                "but you didn't register any AgentTracer.TracerAPI. " +
```
0xnm: Should we reference this interface? Normally users don't deal with it.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/_RumInternalProxy.kt
satween: After several experiments with api calls - I decided to move back to more classic approach, as in current state `build` method is still visible for customers but they cannot call it which could be confusing

features/dd-sdk-android-trace/api/apiSurface
satween: changed to `setTracingScope` with more readable enum constants ( as combination of boolean flags could create useless configuration)

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
satween: this is the same signature as used in `TracingInterceptor` and created in order to minimize migration steps in the future

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
satween: agreed, changed to MAINTAINER

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
satween: make sense, but we want to hint the developer about potential issue on their side, so changed to `internalSdkCore?....`

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetRequestContext.kt
satween: well hashCode/equals methods are overridden already, but we can get rid from `data` keyword here

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
satween: `networkInstrumentationEnabled` was an intermediate solution to not reporting the traces which are not fully supported in this ticket, but it changed to the `ApmNetworkTracingScope` enum as described above, so can be completely removed

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
satween: yes, as otherwise the tracer provider switching is not atomic. same logic presents in `TracingInterceptor`

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
satween: SDK v3 migration leftover, also fixed in `TracingInterceptor`

integrations/dd-sdk-android-cronet/api/apiSurface
satween: yes, changed to implicit enabling rum resources for the better consistency

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
satween: changed to ApmTracingScope, so not the case anymore

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/RumInternalProxyTest.kt
```diff
@@ -93,4 +95,21 @@ internal class RumInternalProxyTest {
...
+    fun `M return RumResourceInstrumentation W build() {non-null builder}`(
+        @StringForgery fakeInstrumentationName: String
+    ) {
+        // Given
+        val builder = RumResourceInstrumentationConfiguration()
+
+        // When
+        val result = with(_RumInternalProxy.Companion) {
+            builder.createInstrumentation(fakeInstrumentationName)
+        }
```
0xnm, line 109: need to update this

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -1,12 +1,31 @@
+class com.datadog.android.trace.ApmNetworkInstrumentationConfiguration
+  constructor(List<String>)
+  fun setTraceOrigin(String)
+  fun setSdkInstanceName(String)
+  fun setTracedRequestListener(NetworkTracedRequestListener)
+  fun setTraceSampleRate(Float)
+  fun setTraceSampler(com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>)
+  fun setTraceContextInjection(TraceContextInjection)
+  fun set404ResourcesRedacted(Boolean)
+  fun setTracingScope(ApmNetworkTracingScope)
```
0xnm: very nit: `tracing` vs `trace` above, but maybe it is okay if we think of tracing as a process, not as of trace as item

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -0,0 +1,196 @@
...
+     */
+    fun set404ResourcesRedacted(redacted: Boolean) = apply {
+        redacted404ResourceName = redacted
+    }
+
+    /**
+     * Sets the tracing scope for network instrumentation.
+     *
+     * This controls how detailed the tracing will be:
+     * - [ApmNetworkTracingScope.DETAILED]: Traces both application-level requests and internal
```
0xnm: maybe `ALL` is a better name? Also this separation applies to OkHttp only, does Cronet also have similar?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkTracingScope.kt
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Default one.
+     * With this scope the Datadog SDK will trace both the application level requests and the network
+     * layer requests (redirect, retries).
+     */
+    DETAILED,
+
+    /**
+     * Only application level request is gonna be traced.
+     * In this mode the Datadog SDK still able to link trace spans to Rum.Resources making possible to navigate
```
0xnm:
```suggestion
     * In this mode the Datadog SDK still able to link trace spans to RUM Resources making possible to navigate
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,292 @@
...
+                "Your requests won't be traced.\n" +
+                "To set a list of known hosts, you can use the " +
+                "Configuration.Builder.setFirstPartyHosts() method."
+        internal const val ERROR_STACK_OVERFLOW =
+            "StackOverflowError detected in TracedRequestListener. " +
+                "This is likely caused by retrying the same request within the " +
+                "onRequestIntercepted callback, leading to infinite recursion."
+
+        internal const val ERROR_REQUEST_INFO_IS_NOT_MUTABLE =
+            "RequestInfo is not mutable. Your requests won't be traced."
```
0xnm: should it be `HttpRequestInfo`?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/DatadogTracingToolkitTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    @Test
+    fun `M build TracingInstrumentation W build() extension`() {
+        // Given
+        val builder = ApmNetworkInstrumentationConfiguration(fakeTracedHosts)
+
+        // When
+        val result: ApmNetworkInstrumentation = with(DatadogTracingToolkit) {
```
0xnm: `with` is redundant here

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/DatadogPropagationHelperTest.kt
```diff
@@ -0,0 +1,912 @@
...
+        val result = testedHelper.isExtractedContext(context)
+
+        // Then
+        assertThat(result).isTrue()
+    }
+
+    @Test
+    fun `M return false W isExtractedContext() {non-extracted DatadogSpanContextAdapter}`() {
+        // Given
+        val mockAgentSpanContext: com.datadog.trace.bootstrap.instrumentation.api.AgentSpan.Context = mock()
```
0xnm:
```suggestion
        val mockAgentSpanContext = mock<AgentSpan.Context>()
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/DatadogPropagationHelperTest.kt
```diff
@@ -0,0 +1,912 @@
...
+        val mockRequest = createMockRequestWithTags(datadogSpan = mockSpan)
+        whenever(
+            mockPropagation.extract(
+                eq(mockRequest),
+                any<
+                    (
+                        HttpRequestInfo,
+                        (String, String) -> Boolean
+                    ) -> Unit
+                    >()
```
0xnm: nit: seems like formatting gone wrong

integrations/dd-sdk-android-cronet/src/test/kotlin/com/datadog/android/cronet/DatadogCronetEngineBuilderTest.kt
```diff
@@ -338,47 +339,55 @@ internal class DatadogCronetEngineBuilderTest {
 
     @OptIn(ExperimentalRumApi::class)
     @Test
-    fun `M propagate RumResourceAttributesProvider W setRumResourceAttributesProvider()`() {
+    fun `M propagate RumResourceAttributesProvider W setCustomRumInstrumentation()`() {
```
0xnm: there is no such method

integrations/dd-sdk-android-cronet/src/test/kotlin/com/datadog/android/cronet/DatadogCronetEngineBuilderTest.kt
```diff
@@ -338,47 +339,55 @@ internal class DatadogCronetEngineBuilderTest {
...
         check(engine is DatadogCronetEngine)
-        RumResourceInstrumentationAssert.assertThat(engine.rumResourceInstrumentation)
+        RumResourceInstrumentationAssert.assertThat(engine.rumResourceInstrumentation!!)
             .hasRumResourceAttributesProvider(customProvider)
     }
 
     @OptIn(ExperimentalRumApi::class)
     @Test
-    fun `M propagate sdkInstanceName W setSdkInstanceName()`(@StringForgery sdkInstanceName: String) {
+    fun `M propagate sdkInstanceName W setCustomRumInstrumentation()`(@StringForgery sdkInstanceName: String) {
```
0xnm: same here

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/TracerProvider.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    internal companion object {
+        const val WARNING_TRACING_DISABLED = "You added a ApmNetworkInstrumentation to your %s, " +
+            "but you did not enable the TracingFeature. " +
+            "Your requests won't be traced."
+
+        const val WARNING_DEFAULT_TRACER =
+            "You added a ApmNetworkInstrumentation to your %s instrumentation, " +
+                "but you didn't register any DatadogTracer in . " +
```
kikoveiga: Is there something missing? In what?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -0,0 +1,292 @@
...
+        if (!tracingState.isSampled) {
+            tracingState.onRequestCompleted(null, throwable)
+        } else {
+            tracingState.span?.isError = true
+            tracingState.span?.setTag(DatadogTracingConstants.Tags.KEY_ERROR_MSG, throwable.message)
+            tracingState.span?.setTag(DatadogTracingConstants.Tags.KEY_ERROR_TYPE, throwable.javaClass.name)
+            tracingState.span?.setTag(DatadogTracingConstants.Tags.KEY_ERROR_STACK, throwable.loggableStackTrace())
+            tracingState.onRequestCompleted(null, throwable)
+        }
+        tracingState.span?.finishRumAware(tracingState.isSampled, !internalSdkCore.isRumEnabled)
```
kikoveiga: Can this be simplified as `tracingState.onRequestCompleted(null, throwable)` is common in both branches?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
```diff
@@ -0,0 +1,78 @@
...
+        propagationHelper.extractSamplingDecision(request) ?: traceSampler.sample(this)
+    }
+}
+
+internal fun DatadogSpan.finishRumAware(isSampled: Boolean, canSendSpan: Boolean) {
+    if (canSendSpan) {
+        if (isSampled) finish() else drop()
+    } else {
+        drop()
+    }
```
kikoveiga:
```suggestion
    if (canSendSpan && isSampled) finish()
    else {
        drop()
    }
```

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
satween: Prefer to keep this methods separate if you don't mind, otherwise we would need to add several if-else blocks inside method's body (aka `if (throwabel == null)` else `if (requestIInfo == null`) and so on), and for the caller will be less obvious to decide what to pass as the parameters.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
satween: Cronet will have similar in further ticket. DETAILED -> ALL renamed

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -220,15 +224,14 @@ interface com.datadog.android.rum.internal.monitor.AdvancedNetworkRumMonitor : c
...
+class com.datadog.android.rum.internal.net.RumNetworkInstrumentation
   fun sendWaitForResourceTimingEvent(com.datadog.android.api.instrumentation.network.HttpRequestInfo)
   fun sendTiming(com.datadog.android.api.instrumentation.network.HttpRequestInfo, com.datadog.android.rum.internal.domain.event.ResourceTiming)
   fun startResource(com.datadog.android.api.instrumentation.network.HttpRequestInfo)
-  fun stopResource(com.datadog.android.api.instrumentation.network.HttpRequestInfo, com.datadog.android.api.instrumentation.network.HttpResponseInfo)
+  fun stopResource(com.datadog.android.api.instrumentation.network.HttpRequestInfo, com.datadog.android.api.instrumentation.network.HttpResponseInfo, Map<String, Any?> = emptyMap())
   fun stopResourceWithError(com.datadog.android.api.instrumentation.network.HttpRequestInfo, Throwable)
   fun reportInstrumentationError(String)
-  companion object 
+  companion object Companion
```
0xnm: seems like unnecessary

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumNetworkInstrumentation.kt
```diff
@@ -142,7 +147,7 @@ class RumResourceInstrumentation(
         }
     }
 
-    companion object {
+    companion object Companion {
```
0xnm:
```suggestion
    companion object {
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/DatadogPropagationHelperTest.kt
```diff
@@ -0,0 +1,907 @@
...
+        val mockRequest = createMockRequestWithTags()
+        whenever(
+            mockPropagation.extract(
+                eq(mockRequest),
+                any<
+                    (
+                        HttpRequestInfo,
+                        (String, String) -> Boolean
+                    ) -> Unit
+                    >()
```
0xnm: formatting

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmInstrumentationConfigurationTest.kt
```diff
@@ -0,0 +1,254 @@
...
+        // When
+        val result = testedBuilder.setTraceScope(ApmNetworkTracingScope.APPLICATION_LEVEL_REQUESTS_ONLY)
+
+        // Then
+        assertThat(result).isSameAs(testedBuilder)
+        assertThat(testedBuilder.networkTracingScope).isEqualTo(ApmNetworkTracingScope.APPLICATION_LEVEL_REQUESTS_ONLY)
+    }
+
+    @Test
+    fun `M set DETAILED scope W setTracingScope()`() {
```
0xnm:
```suggestion
    fun `M set DETAILED scope W setTraceScope()`() {
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmInstrumentationConfigurationTest.kt
```diff
@@ -0,0 +1,254 @@
...
+
+        // When
+        val result = testedBuilder.createInstrumentation(fakeNetworkLibraryName)
+
+        // Then
+        assertThat(result.localFirstPartyHostHeaderTypeResolver.isFirstPartyUrl("https://$validHost/path")).isTrue()
+    }
+
+    @Test
+    fun `M set APPLICATION_LEVEL_REQUESTS_ONLY scope W setTracingScope()`() {
```
0xnm:
```suggestion
    fun `M set APPLICATION_LEVEL_REQUESTS_ONLY scope W setTraceScope()`() {
```

integrations/dd-sdk-android-okhttp/api/apiSurface
```diff
@@ -25,15 +25,15 @@ open class com.datadog.android.okhttp.DatadogInterceptor : com.datadog.android.o
...
-  constructor(String, String, Int)
-enum com.datadog.android.okhttp.TraceContextInjection
-  - ALL
-  - SAMPLED
-open class com.datadog.android.okhttp.trace.DeterministicTraceSampler : com.datadog.android.core.sampling.DeterministicSampler<com.datadog.android.trace.api.span.DatadogSpan>
-  constructor(() -> Float)
-  constructor(Float)
-  constructor(Double)
+typealias TraceContextInjection = com.datadog.android.trace.TraceContextInjection
+class com.datadog.android.okhttp.internal.OkHttpRequestInfoBuilder : com.datadog.android.api.instrumentation.network.HttpRequestInfoBuilder
```
0xnm, line 29: should this class be public?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
kikoveiga: I was just suggesting this:

```suggestion
        if (tracingState.isSampled) {
            tracingState.span?.isError = true
            tracingState.span?.setTag(DatadogTracingConstants.Tags.KEY_ERROR_MSG, throwable.message)
            tracingState.span?.setTag(DatadogTracingConstants.Tags.KEY_ERROR_TYPE, throwable.javaClass.name)
            tracingState.span?.setTag(DatadogTracingConstants.Tags.KEY_ERROR_STACK, throwable.loggableStackTrace())
        }
        tracingState.onRequestCompleted(null, throwable)
        tracingState.span?.finishRumAware(tracingState.isSampled, !internalSdkCore.isRumEnabled)
```

integrations/dd-sdk-android-okhttp/api/apiSurface
satween: yes, for `-otel` module, it's marked with `@InternalApi`

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
satween: Ah, make sense, didn't understand from the first attempt,  applied 👍

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/DatadogCronetEngine.kt
```diff
@@ -325,24 +369,30 @@ class DatadogCronetEngine internal constructor(
...
+                    rumNetworkInstrumentation = instrumentation,
+                    executor = listenerExecutor ?: newListenerExecutor()
                 )
-            )
-            return engine
+            }
+
+            return if (rumResourceInstrumentation == null && tracingInstrumentation == null) {
+                delegate.build()
+            } else {
```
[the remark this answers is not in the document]
satween: this is okay - as if user didn't enabled rum or apm features - there should not be any events

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetRequestContext.kt
```diff
@@ -0,0 +1,214 @@
...
+            dictionaryId
+        )
+    }
+
+    internal fun setRequestFinishedListener(listener: RequestFinishedInfo.Listener?) {
+        requestParams.listener = listener
+    }
+
+    internal fun addRequestAnnotation(annotation: Any) {
+        additionalAnnotations[annotation::class.java] = annotation
```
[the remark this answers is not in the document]
satween: make sense

Issue #1550, opened 2023-07-19
Allow client SDK to provide a custom version number


ankushg:
Currently, the datadog SDK seems to be looking up our version number from the PackageManager: https://github.com/DataDog/dd-sdk-android/blob/4be13fb69720ba54a7d61284816fb8a91ed75d19/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L314C34-L314C59

We reuse the same user-visible number if we discover bugs before pushing to production, but this makes all of our datadog dashboards extremely confusing to use.

We'd like to be able to provide our version **code** as the datadog version, rather than the user-visible version string.

xgouchet:
Hi @ankushg , this is a topic we're considering supporting in the future, we'll keep you posted when we have a plan for such use case

ankushg:
Thanks -- from the code, it looked like we should have been able to pass in an attribute called `"_dd.version"` and have that used instead of what the PackageManager reports

Is that accurate?

xgouchet:
Yes indeed that workaround will work, you just need to make sure you change also the version in the gradle plugin when uploading your Proguard/R8 mapping files, with the following snippet :

```kotlin
datadog {
    versionName = "42"
}
```

PR #3137, opened 2026-01-21
RUM-13018 - Add 'setVersion' method to the configuration

hamorillo:
### What does this PR do?

Similar to what https://github.com/DataDog/dd-sdk-ios/pull/2586 applies for the iOS SDK, this PR adds a `setVersion` method in the `Configuration.Builder` that allows applications to use a custom version that will be used by the SDK.

### How

<details><summary>First Approach - Discarded</summary>

I noticed that there was already a way to set a custom version using `additionalConfiguration`:~~

```kotlin
.setAdditionalConfiguration(
    mapOf(Datadog.DD_APP_VERSION_TAG to "customVersion")
)
```

~~However `Datadog.DD_APP_VERSION_TAG` is internal an cannot be used directly by applications. (although it would be possible to rely on the raw string). In [DatadogCore.applyAdditionalConfiguration](https://github.com/DataDog/dd-sdk-android/blob/6c11581261c7cf66e900e1ae5faf6eb331880084/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt#L521) if a value for the version is provided, it override the default value.
</details>

If a custom version is provided will be saved in the Configuration and used when initializing the CoreFeature. The priority order is:

**Custom Version** -> **Package Version Name** -> **Package Version Code** -> **Default App Version**

Apart from that, the version can be overridden via AdditionalConfig when used by cross-platform SDKs.

### Additional Notes

~~The previous information raises the following question:
– I haven’t found any uses of `Datadog.DD_APP_VERSION_TAG` outside of that method. Can we remove it and go with a different implementation, or is there another reason to keep that code?~~

**Edited:** `additionalConfig` is used by cross-platforms for [example](https://github.com/DataDog/dd-sdk-reactnative/blob/ceddd4c317a90a7b19409146690022cd6889cb26/packages/core/android/src/main/kotlin/com/datadog/reactnative/DdSdkNativeInitialization.kt#L253) so we should keep it there.

### DIFF
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreInitializationTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreInitializationTest.kt
@@ -626,7 +626,7 @@ internal class DatadogCoreInitializationTest {
     }
 
     @Test
-    fun `M apply app version W setVersion(appVersion) { with app version }`(
+    fun `M apply app version from config W initialize { version in config is not null }`(
         @StringForgery appVersion: String
     ) {
         // When


dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/ConfigurationForgeryFactory.kt
```diff
@@ -20,7 +20,8 @@ internal class ConfigurationForgeryFactory :
             variant = forge.anElementFrom(forge.anAlphabeticalString(), ""),
             service = forge.aStringMatching("[a-z]+(\\.[a-z]+)+"),
             crashReportsEnabled = forge.aBool(),
-            additionalConfig = forge.aMap { aString() to aString() }
+            additionalConfig = forge.aMap { aString() to aString() },
+            version = forge.anAlphaNumericalString()
```
aleksandr-gringauz: It is better to do forge.aNullable { anAlphaNumericalString() } here. Because version can be null.

aleksandr-gringauz: And in CoreFeatureTest probably create separate fakeConfigs with version and without. WDYT?

hamorillo: > It is better to do forge.aNullable { anAlphaNumericalString() } here. Because version can be null.

Agree! Updated! 

> And in CoreFeatureTest probably create separate fakeConfigs with version and without. WDYT?

I'm not entirely sure that having two separate fakeConfigs is better than preparing it in the when part of the test, as we do for other parameters. I would prefer to keep a single fakeConfig and set it up when necessary in the test. I’ve added this in the last commit as well. Let me know what you think.

aleksandr-gringauz:
> I think on Android we didn't have much demand to add such configuration option, however, once merged we can close the following issue #1550.
> 
> Worth also notify Cross-Platform teams that they can switch to this API.

Thanks for mentioning this. There is actually one important [comment](https://github.com/DataDog/dd-sdk-android/issues/1550#issuecomment-1645173044) there. @hamorillo could you please check that this comment is relevant and if yes - I suggest adding a comment about this for the new `setVersion` function.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/forge/ConfigurationForgeryFactory.kt
aleksandr-gringauz: looks fine to me!

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/DatadogCoreInitializationTest.kt
```diff
@@ -623,6 +625,28 @@ internal class DatadogCoreInitializationTest {
         )
     }
 
+    @Test
+    fun `M apply app version W setVersion(appVersion) { with app version }`(
```
aleksandr-gringauz: This should probably be called:

`M apply app version from config W initialize { version in config is not null }`.

And the test itself:

```
// Given
testedCore = DatadogCore(...)

// When
testedCore.initialize(...)

// Then
assertThat(testedCore.coreFeature.packageVersionProvider.version).isEqualTo(appVersion)
```

hamorillo:
> Thanks for mentioning this. There is actually one important https://github.com/DataDog/dd-sdk-android/issues/1550#issuecomment-1645173044 there.

Just to keep track of what we've discovered. 
Looks like that comment don't apply anymore as we are [no longer using](https://docs.datadoghq.com/error_tracking/frontend/mobile/android/?tab=kotlin#upload-your-mapping-file) `version` to match with the correct mapping file.

image

That change was merged in the SDK [here](https://github.com/DataDog/dd-sdk-android-gradle-plugin/pull/211).

PR #3143, opened 2026-01-22
Add Experimental annotation on Profiling APIs & Bug fix

ambushwork:
### What does this PR do?

* Add Experimental annotation on Profiling APIs
* Fix the bug when profiling manager callback is called with error, `runningInstances` collection is not cleaned, leading `isRunning` return true even if the profiler has stopped.


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -40,6 +40,7 @@ object Profiling {
     @JvmStatic
     @JvmOverloads
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+    @ExperimentalProfilingApi
```
0xnm: it is better to put it on `Profiling` object as a whole and also annotate whole `ProfilingConfiguration` class as well.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingConfigurationBuilderTest.kt
```diff
@@ -4,8 +4,11 @@
  * Copyright 2016-Present Datadog, Inc.
  */
 
+@file:OptIn(ExperimentalProfilingApi::class)
```
0xnm, line 7: probably should be at the class level? alternatively we can opt in for the whole module https://kotlinlang.org/docs/opt-in-requirements.html#opt-in-a-module

0xnm (approved):

PR #3144, opened 2026-01-22
[FFL-1720] Evaluation Logging: Event Schema & Data Models

typotter:
### 📚 Evaluation Logging Stacked Pull Requests 📚
🔲 Integration & Configuration (PR #3147)
🔲 Storage & Network Infrastructure (PR #3146)
🔲 Aggregation Engine & Test Utilities (PR #3145)
👉 **Event Schema & Data Models** (PR #3144)
☑️ Evaluations Subfeature (PR #3159)
⎿ `typo/FFL-1720-android-kotlin-sdk-emits` (feature branch)

_Datadog Internal_
🎟️ **Ticket:** [FFL-1720 - Implement Evaluation Logging for Android SDK](https://datadoghq.atlassian.net/browse/FFL-1720)

### What does this PR do?

Introduces the `BatchedEvaluationEvent` data model and JSON schema for aggregated feature flag evaluation events. This establishes the contract for evaluation logging that will be sent to Datadog's `/api/v2/flagevaluations` intake endpoint.

### Motivation

We need to implement **[Evaluation Logging](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/5906009399/FFE+SDK+requirements#Evaluation-logging)** to provide comprehensive visibility into all feature flag evaluations, including defaults, errors, and successful matches. This goes beyond exposure logging by capturing aggregated metrics about evaluation frequency, error rates, and runtime default usage across all flags.

### Description

This PR adds:
- **`batched-evaluation-event-schema.json`**: JSON schema defining the structure of aggregated evaluation events
- **`BatchedEvaluationEvent` model**: Generated Kotlin data class with nested types (Identifier, Context, Error, etc.)

The schema follows the [Evaluation Logging specification](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/5906009399/FFE+SDK+requirements#Evaluation-logging) requirements for event structure, field naming, optional field handling, and timestamp conventions.

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

PR #3145, opened 2026-01-22
[FFL-1720] Evaluation Logging: Aggregation Engine & Test Utilities

typotter:
🔲 Integration & Configuration (PR #3147)
🔲 Storage & Network Infrastructure (PR #3146)
👉 **Aggregation Engine & Test Utilities** (this PR)
☑️ FlagEvaluation Schema (PR #3166)

_Datadog Internal_

### What does this PR do?

Implements the core evaluations aggregation engine that groups feature flag evaluations by key characteristics and manages periodic flushing from memory to persistent storage (via `StorageBackedFeature`/`EvaluationsFeature`).

The `EvaluationEventProcessor` manages concurrent access and mutation of `AggregationStats` instances in memory. Each Stats object is given an `AggregationKey` - a composite key made of the set of identifiers across the evaluation (targetingKey, variationKey, etc.) and SDK context (rum View ID). This ensures proper evaluation counting under the "same" circumstances.

The processor also manages the periodic and size-based triggers for flushing the in-memory aggregations to persistent storage.

### Motivation

### Description

This PR adds the core business logic for evaluation logging:

**Implementation:**
- **`AggregationKey`**: Defines how evaluations are grouped (by flag, variant, allocation, targeting key, and error code)
- **`AggregationStats`**: Tracks count, first/last timestamps, and last error message for each aggregation
- **`EvaluationEventsProcessor`**: Orchestrates aggregation, manages flush triggers (time-based, size-based, shutdown)
- **`EvaluationEventWriter`**: Interface for persisting evaluation events (implementation in PR #3146)

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
@@ -80,7 +80,7 @@ internal class EvaluationAggregator(private val maxAggregations: Int) {
         if (mapSize < maxAggregations) {
             return emptyList()
         }
-    
+
         // Re-check inside lock to ensure only one thread drains
         val drained = mapLock.withLock {
             if (aggregationMap.size < maxAggregations) emptyMap() else drainAggregationStats()


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,104 @@
...
+     * Updates the last error message to the most recent one.
+     * Thread-safe: can be called concurrently from multiple threads.
+     *
+     * @param timestamp the timestamp of the evaluation
+     * @param errorMessage the error message for this evaluation (updates to latest)
+     */
+    fun recordEvaluation(timestamp: Long, errorMessage: String?) {
+        synchronized(this) {
+            count++
+            lastEvaluation = timestamp
```
dd-oleksii: major: this assumes monotonically increasing timestamps, which is generally not the case for computers (OS time synchronization may make the time go backwards). This is doubly not the case for concurrent applications, where there's arbitrary delay between the threads, so they are not guaranteed to call this function in the proper order.

We need to compare `timestamp` against both `firstEvaluation` and `lastEvaluation`, and it might already fall in-between both.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+
+        // Update existing stats or create new ones (thread-safe for API 21+)
+        synchronized(aggregationMap) {
+            val existing = aggregationMap[key]
+            if (existing != null) {
+                existing.recordEvaluation(timestamp, errorMessage)
+            } else {
+                aggregationMap[key] = AggregationStats(timestamp, context, data, errorMessage)
+            }
+        }
```
dd-oleksii: minor: synchronizing on `aggregationMap` is killing the concurrency benefit that `ConcurrentHashMap` provides. If we want to benefit from concurrent hash map, we should use `.compute()` API. However, this makes the synchronization with `flush` more complicated, so perhaps a regular `HashMap` is a better choice here.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+     * The aggregation map is cleared after flushing.
+     *
+     * Thread-safe: synchronized to prevent concurrent flush operations.
+     */
+    fun flush() {
+        // Take atomic snapshot and clear map
+        val snapshot = synchronized(aggregationMap) {
+            val copy = aggregationMap.toMap()
+            aggregationMap.clear()
+            copy
```
dd-oleksii: minor: the copy here seems unnecessary as we can "take out" the previous map and assign a fresh map to `aggregationMap` instead.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+            } else {
+                aggregationMap[key] = AggregationStats(timestamp, context, data, errorMessage)
+            }
+        }
+
+        // Size-based flush
+        if (aggregationMap.size >= maxAggregations) {
+            flush()
+        }
+    }
```
dd-oleksii, line 76: minor: given that this is outside the synchronized block, this may trigger multiple flushes. Depending on the timing of new inserts/flushes, this may trigger a couple of very small flushes in a short time frame.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+        try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // exception caught below
+            scheduledFlushFuture = scheduledExecutor.schedule(
+                {
+                    flush()
+                    schedulePeriodicFlush() // Self-reschedule
+                },
+                flushIntervalMs,
+                TimeUnit.MILLISECONDS
+            )
```
dd-oleksii: minor: periodic flush does not seem to cooperate with the size-based flush. A size-based flush does not prevent/postpone a periodic flush, so periodic flush may trigger very shortly after a size-based flush and send a very small batch.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * Stops the processor and flushes remaining evaluations.
+     *
+     * Performs final flush on shutdown.
+     * Cancels any scheduled flush tasks.
+     */
+    fun stop() {
+        // Cancel scheduled flush
+        scheduledFlushFuture?.cancel(false)
```
dd-oleksii: major: this code may fail to stop the scheduled future. Given that this block is not synchronized, it is possible for it to fire exactly at the same time as the scheduled task is executing. This may lead to `stop()` reading and canceling the previous scheduled future and the scheduled task to continue scheduling a new one. To prevent this from happening, we should shutdown the `scheduledExecutor` service.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    /**
+     * Flushes all aggregated evaluations.
+     *
+     * Called on: time interval, size limit, or shutdown.
+     * The aggregation map is cleared after flushing.
+     *
+     * Thread-safe: synchronized to prevent concurrent flush operations.
```
dd-oleksii: nitpick: it does not prevent concurrent flushes (but tries to make them work correctly).

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,104 @@
...
+     * Records an additional evaluation at the given timestamp.
+     *
+     * Updates the last error message to the most recent one.
+     * Thread-safe: can be called concurrently from multiple threads.
+     *
+     * @param timestamp the timestamp of the evaluation
+     * @param errorMessage the error message for this evaluation (updates to latest)
+     */
+    fun recordEvaluation(timestamp: Long, errorMessage: String?) {
+        synchronized(this) {
```
dd-oleksii: nitpick: due to synchronization in the events processor, `AggregationStats.recordEvaluation()` is never called concurrently, so perhaps we don't need to make it concurrent? (Even with proper use of `ConcurrentHashMap`, this class would still be externally synchronized.)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,104 @@
...
+        val snapshotCount: Int
+        val snapshotFirst: Long
+        val snapshotLast: Long
+        val snapshotMessage: String?
+        synchronized(this) {
+            snapshotCount = count
+            snapshotFirst = firstEvaluation
+            snapshotLast = lastEvaluation
+            snapshotMessage = lastErrorMessage
+        }
```
dd-oleksii: minor: this piece is confusing. These fields are volatile, so torn read is not possible. Otherwise, this class is not designed to handle concurrent calls to `recordEvaluation` and `toEvaluationEvent` as that may leave some events uncounted.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationKey.kt
```diff
@@ -0,0 +1,63 @@
...
+            errorCode: String?
+        ): AggregationKey {
+            // Omit variant/allocation for DEFAULT/ERROR reasons
+            // Use string comparison to be forward-compatible with new reason values
+            val isDefaultOrError = data.reason == "DEFAULT" || data.reason == "ERROR"
+
+            return AggregationKey(
+                flagKey = flagName,
+                variantKey = if (isDefaultOrError) null else data.variationKey,
+                allocationKey = if (isDefaultOrError) null else data.allocationKey,
```
dd-oleksii: minor: the handling of `isDefaultOrError` is confusing and I would argue it's not a job of `AggregationKey`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventWriter.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Writer interface for persisting evaluation events.
+ */
+@NoOpImplementation
+internal interface EvaluationEventWriter {
+    /**
+     * Writes an evaluation event.
+     *
+     * @param event the evaluation event to write
+     */
+    fun write(event: BatchedFlagEvaluations.FlagEvaluation)
```
dd-oleksii: minor: it's worth documenting that it _must_ be thread-safe (based on how it's used in event processor)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+
+        if (snapshot.isEmpty()) {
+            return
+        }
+
+        // Convert each aggregation to an event and write
+        snapshot.forEach { (key, stats) ->
+            val event = stats.toEvaluationEvent(key.flagKey, key)
+            writer.write(event)
+        }
```
dd-oleksii: Q: given that we send each event individually, how does the writer knows when we're finished and it may flush? How does the writer knows that the app is shutting down and we _need_ to flush?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: updated the logic to check each timestamp and stretch the range as necessary

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: unfortunately, `.compute()` isn't available in the min API we support (21). I have revamped the locking across the processor to be more precise.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: I have changed the flush logic:
- all calls to `flush()` will stop any scheduled flush from being executed (doesn't interrupt any _currently running_)
- `flush()` will schedule a periodic flush upon completion (or exception).

My thinking is that the flush timer should just restart on every flush.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: updated to cancel and reschedule flushes when a flush starts/finishes, respectively.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: Thanks for the solve. implemented

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: locking has been restructured, so the call to `recordEvaluation` is not protected by that synchronization umbrella any longer.
Internally, we're making a compound update, so if `recordEvaluation` is called concurrently, we properly compute the net values by processing serially

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: perhaps this is being overly defensive. Because of the CAS of the map of aggregation stats for flushing, _no other thread should be calling `toEvaluationEvent` or `recordEvaluation` on this `AggregationStats` instance.

Without this `synchronized`, `recordEvaluation` could change the computed values while we're getting them for constructing the `FlagEvaluation`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: It is possible now to call this concurrently so we `synchronize` to complete the compound update.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: With the synchronized CAS snapshotting the `aggregationMap` for flushing, this method should be protected from another thread calling `recordEvaluation` which could change the values as they're being grabbed for the `FlagEvaluation` constructor
Is this too defensive?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: for the writer, it is writing each event to storage. The SDK core manages the lifecycle from there, flushing the batch data from persistent storage to the intake.
I did, however, modify the interface to support writing a list of events, this saves the overhead of the write context, scope and lock in the implementation in #3146

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Converts the aggregated statistics to an EvaluationEvent.
+     *
+     * Thread-safe: synchronizes to create consistent snapshot of statistics.
+     *
+     * @return the evaluation event
+     */
+    fun toEvaluationEvent(): BatchedFlagEvaluations.FlagEvaluation {
+        // Take atomic snapshot of statistics
```
typotter: atomic snapshot here is probably overkill considering `EvaluationEventsProcessor` only calls `toEvaluaition` after it has been removed from the aggregation map (behind a lock), so nothing should have a handle to this object nor calling `recordEvaluation`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/DDContext.kt
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * This context is used for aggregating evaluations by RUM view and for
+ * enriching evaluation events with RUM session information.
+ *
+ * @param service Service name, or null if not available
+ * @param applicationId RUM application ID, or null if RUM is not active
+ * @param viewId RUM view ID, or null if no active view
+ * @param viewName RUM view name, or null if no active view
+ */
+internal data class DDContext(
```
typotter: These fields are used to populate both the top level context of the BatchedFlagEvaluations and the "snapshot" context encoded in each `FlagEvaluation`.
Using the context update mechanism in the `EvaluationsFeature` provides a best effort to keep these fields current for when evaluation events are processed.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,159 @@
...
+                context = context,
+                service = service,
+                rumApplicationId = rumApplicationId,
+                rumViewName = rumViewName,
+                reason = reason,
+                errorMessage = errorMessage
+            )
+        )
+
+        existing?.recordEvaluation(timestamp, errorMessage)
```
dd-oleksii: major: ok, this is now not thread-safe with respect to concurrent flushes.

What may happen is:
1. Thread A calls `putIfAbsent`, puts a new entry or returns an existing one which is saved to `existing`.
2. Thread B does the flush, replacing `aggregationMap` with a new map, and calling `toEvaluationEvent()` on all stats.
3. Thread A now proceeds to record evaluation event on a dangling `AggregationStats` (saved in now-discarded aggregation map)

A couple of ways to fix this:
1. Use exclusive lock for any aggregation map changes. It needs to be held in both `processEvaluation` and in `flush` (during the swap). The downside is that it kills concurrency of the `ConcurrentHashMap` (so it would be better to use a lock + a regular hash map).
2. Perhaps counter-intuitively, we can use a read-write lock here with `processEvaluation` acquiring a shared *read* lock, and `flush` acquiring an exclusive *write* lock (for the swap only). The reason is that the lock is protecting `aggregationMap` field itself but not the data inside (which is protected by concurrent hash map and `AggregationStats` internal locking).

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    /**
+     * Flushes aggregated evaluations. Non-blocking if flush already in progress.
+     */
+    fun flush(rescheduleFlush: Boolean = false) {
+        if (!flushInProgress.compareAndSet(false, true)) {
+            return
+        }
```
dd-oleksii: major: because there's no synchronization between processEvent and flush, one scenario that might happen:

1. Thread A starts a flush, swaps the maps for the new one.
2. While it's sending events to the writer, other threads submit more events that trigger size check. However, because `flushInProgress` is still true, it's going to be discarded here and the next flush will only happen after period.

We can fix that by either exclusive locking or by adding a loop in flush: at the end of the flush, check if new `aggregationMap` exceeds the limit and flush immediately if so.

Alternatively, maybe we can do the actual flush outside of holding `flushInProgress`. So the sequence would be:
1. Acquire flush lock (if we do read-write lock, this can be the same lock that we use for `processEvaluation`)
3. Re-arm periodic flush
5. Swap aggregation map for the new one
6. Release the flush lock
7. Serialize snapshot and write to writer

In `processEvaluation`:
1. Acquire read lock
2. `putIfAbsent` + `recordEvaluation`
3. `val tooBig = aggregationMap.size >= maxAggregations`
4. Release read lock
5. *If* tooBig:
   1. Acquire write lock (or flush lock)
   2. Re-check `aggregationMap.size >= maxAggregations` and run flush if needed

Acquiring the write lock in `processEvaluation` is needed to make sure we don't trigger too many or too few flushes.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,159 @@
...
+                e
+            )
+        }
+    }
+
+    fun stop() {
+        @Suppress("UnsafeThirdPartyFunctionCall") // safe - does not throw in Android
+        scheduledExecutor.shutdown()
+        scheduledFlushFuture?.cancel(false)
+        flush(rescheduleFlush = false)
```
dd-oleksii: major: because of `flushInProgress` effectively canceling the flush call, calling `flush()` is not guaranteed to flush all events up to this point.

Example:
1. Thread A is running flush, holding `flushInProgress`
2. There are more events written to the map, since it was swapped
3. Thread B calls `stop()`, calling `flush()` here. However, because Thread A is still serializing/flushing entries, this call to `flush()` returns early without flushing anything. All the events written in (2) are lost.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
dd-oleksii: I agree that this is how it should work and atomic snapshot should not be needed here. Moreover, if it's ever possible for `recordEvaluation` and `toEvaluationEvent` to be called concurrently, that will lead to the undercounting bug / lost events.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: Thanks for the insight and solutions. I pulled out the CHM and lock into `EvaluationAggregator` to help me keep a mental model of all the locks and operations. I believe that we now have pretty tight locking in place, avoiding un-necessary contention and enabling high throughput.

typotter (commented):
Thanks Oleksii, as always for your 👀 . ptal

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,128 @@
...
+ * @param reason the resolution reason (null for error evaluations, non-null for success evaluations)
+ * @param errorMessage optional error message (detailed, for logging)
+ */
+internal class AggregationStats(
+    private val aggregationKey: AggregationKey,
+    firstTimestamp: Long,
+    private val context: EvaluationContext,
+    private val service: String?,
+    private val rumApplicationId: String?,
+    private val rumViewName: String?,
```
dd-oleksii: nit: view name is duplicated in aggregation key. do we need it here?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationKey.kt
```diff
@@ -0,0 +1,32 @@
...
+ * - targeting key
+ * - RUM view name (or null when RUM is not active or no view available)
+ * - error code (ErrorCode enum name for aggregation, e.g., "FLAG_NOT_FOUND")
+ *
+ * Note: Error messages are stored in AggregationStats but not used for aggregation
+ * to keep the number of unique keys bounded while preserving debugging information.
+ *
+ * Context aggregation is reserved for future use and not implemented.
+ */
+internal data class AggregationKey(
```
dd-oleksii, line 25: do we need `reason` in the aggregation key?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,128 @@
...
+            targetingRule = null, // Not applicable
+            targetingKey = aggregationKey.targetingKey,
+            context = eventContext,
+            error = snapshotMessage?.let { FlagEvaluation.Error(message = it) },
+            evaluationCount = snapshotCount.toLong(),
+            firstEvaluation = snapshotFirst,
+            lastEvaluation = snapshotLast,
+            runtimeDefaultUsed = reason == null ||
+                reason == ResolutionReason.DEFAULT.name ||
+                reason == ResolutionReason.ERROR.name
```
dd-oleksii: minor: the condition I would use here is `aggregationKey.variantKey == null`. Reasons are somewhat less reliable and it's valid to return runtime default with reasons besides "default" or "error"

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,143 @@
...
+            variantKey = variantKey,
+            allocationKey = allocationKey,
+            reason = reason,
+            errorCode = errorCode,
+            errorMessage = errorMessage
+        )
+
+        if (shouldFlush) {
+            flush()
+        }
```
dd-oleksii: minor: if there are many threads recording evaluations, it's still possible to flush too frequently:

1. Thread A records an event, `shouldFlush` is true
2. Thread B records an event, `shouldFlush` is true
3. Thread A flushes all events, aggregation map is now empty
4. Thread C records an event, `shouldFlush` is false
5. Thread B tries to flush, flushes a single event (written by C)

Note in my previous message, the size limit is re-checked after acquiring a _write_ lock and holding it through to the flush.

Two ways we can implement this in the current structure: either add a parameter to `flush()` for it to do an extra check and only flush if the size exceeds the limit. Alternatively, we can make `aggregator.record()` return drained records instead of a flag—then we can keep this synchronization entirely inside aggregator.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,94 @@
...
+            reason = reason,
+            errorMessage = errorMessage
+        )
+
+        return mapLock.read {
+            @Suppress("UnsafeThirdPartyFunctionCall") // Only throws if null is passed
+            val existing = aggregationMap.putIfAbsent(key, stats)
+            existing?.recordEvaluation(timestamp, errorMessage)
+            aggregationMap.size >= maxAggregations
+        }
```
dd-oleksii: Example of how to avoid spurious flushes:
```suggestion
    ): List<FlagEvaluation>? {
        val key = AggregationKey(
            flagKey = flagKey,
            variantKey = variantKey,
            allocationKey = allocationKey,
            targetingKey = context.targetingKey,
            viewName = rumViewName,
            errorCode = errorCode
        )

        val stats = AggregationStats(
            aggregationKey = key,
            firstTimestamp = timestamp,
            context = context,
            service = service,
            rumApplicationId = rumApplicationId,
            rumViewName = rumViewName,
            reason = reason,
            errorMessage = errorMessage
        )

        mapLock.read {
            @Suppress("UnsafeThirdPartyFunctionCall") // Only throws if null is passed
            val existing = aggregationMap.putIfAbsent(key, stats)
            existing?.recordEvaluation(timestamp, errorMessage)
        }
        
        // note: it's safe to check the size without holding a lock,
        // we will re-check it below to avoid spurious flushes.
        // We use this double-check to avoid acquiring a costly write lock
        // for every record.
        if (aggregationMap.size < maxAggregations) {
          // No flush needed
          return null
        }
        
        // Re-check if we need to flush, holding an exclusive lock.
        // This is required to make sure that only a single thread flushes
        // on exceeding the size limit.
        val snapshot = mapLock.write {
          if (aggregationMap.size < maxAggregations) {
            return null
          }
          val toDrain = aggregationMap
          aggregationMap = ConcurrentHashMap()
          toDrain
        }
        return snapshot.map { (_, stats) -> stats.toEvaluationEvent() }
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,94 @@
...
+import kotlin.concurrent.read
+import kotlin.concurrent.write
+
+/**
+ * Thread-safe aggregator for flag evaluation events.
+ *
+ * Uses a [ReentrantReadWriteLock] to allow concurrent [record] calls while
+ * ensuring exclusive access during [drain].
+ */
+internal class EvaluationAggregator(
```
dd-oleksii: 👍 nice idea moving the locking into its own class

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: > return drained records instead of a flag—then we can keep this synchronization entirely inside aggregator.

Interesting approach

dd-oleksii: yeah, the only downside is that it doesn't cancel scheduled flush holding the same lock, so it's possible for:
1. Thread A calls `aggregator.record()` which triggers size limit and drains records
2. Thread B writes more records
4. Thread C — scheduled flush triggers
5. Thread A flushes returned records
6. Thread C continues with scheduled flush and writes just a few records

This is probably less severe than the previous situation with extra flushes due to size limit because there's at most one scheduled flush whereas the number of threads producing evaluations can be much higher.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationKey.kt
typotter: no, fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: good idea. fixed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: agreed

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
typotter: 👌 nice

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,105 @@
...
+     * Swaps the aggregation map with a fresh one and converts entries to events.
+     * Must be called while holding the write lock.
+     */
+    private fun drainInternal(): List<FlagEvaluation> {
+        if (aggregationMap.isEmpty()) {
+            return emptyList()
+        }
+        val toDrain = aggregationMap
+        aggregationMap = ConcurrentHashMap()
+        return toDrain.map { (_, stats) -> stats.toEvaluationEvent() }
```
dd-oleksii: minor: this mapping is best done outside of write lock

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,141 @@
...
+            rumApplicationId = rumApplicationId,
+            rumViewName = rumViewName,
+            variantKey = variantKey,
+            allocationKey = allocationKey,
+            errorCode = errorCode,
+            errorMessage = errorMessage
+        )
+
+        if (drainedEvents != null) {
+            writer.writeAll(drainedEvents)
```
dd-oleksii, line 74: re-schedule periodic flush here? (don't forget `flushMutex`, so we don't schedule multiple futures simultaneously)

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventWriter.kt
```diff
@@ -0,0 +1,36 @@
...
+     * Writes flag evaluation events to persistent storage.
+     *
+     * The events are serialized and stored as individual records. Multiple events
+     * will be collected and batched together at upload time.
+     *
+     * Thread Safety: Implementations must be thread-safe and not rely on external synchronization.
+     *
+     * @param events the evaluation events to write to storage
+     */
+    fun writeAll(events: List<FlagEvaluation>)
```
0xnm, line 35: nit: as I noted in another review, maybe it can be just `write`, especially if we don't expect another methods in this interface

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationKey.kt
```diff
@@ -0,0 +1,32 @@
...
+    val flagKey: String,
+    val variantKey: String? = null,
+    val allocationKey: String? = null,
+    val targetingKey: String?,
```
0xnm, line 29: Is there specific reason why `targetingKey` has no default value as other nullable properties of this class?

Should this property be nullable?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,141 @@
...
+import com.datadog.android.internal.time.TimeProvider
+import java.util.concurrent.RejectedExecutionException
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
+
+/**
+ * Aggregates flag evaluation events before writing them to storage.
```
0xnm, line 22: It seems that aggregation logic can be done on the backend side, is there a reason we do this on the client side? Seems like doing it on the client side adds some complexity.

Or is it because otherwise we have to send too many events to the intake?

dd-oleksii: Yep, the client-side aggregation is to reduce network usage. Flag evaluations can be on the hot path in users' code, and would generate tons of events without aggregation.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationKey.kt
dd-oleksii: It must be nullable because the user might not pass the targeting key. But `null` is not a good default value per se as that might hide issues/bugs.

I would suggest removing default values from the rest of the properties. If they are null, it's better to pass that explicitly

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Swaps the aggregation map with a fresh one and converts entries to events.
+     * Must be called while holding the write lock.
+     */
+    private fun drainInternal(): List<FlagEvaluation> {
+        if (aggregationMap.isEmpty()) {
+            return emptyList()
+        }
+        val toDrain = aggregationMap
+        aggregationMap = ConcurrentHashMap()
```
0xnm: this can be simpler:

```
val toDrain = aggregationMap.toMap()
aggregationMap.clear()
```

and then `aggregationMap` can be simply `val` without `@Volatile`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,105 @@
...
+        val stats = AggregationStats(
+            aggregationKey = key,
+            firstTimestamp = timestamp,
+            context = context,
+            service = service,
+            rumApplicationId = rumApplicationId,
+            errorMessage = errorMessage
+        )
+
+        mapLock.read {
```
0xnm: although `aggregationMap.putIfAbsent` below is a write operation

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,105 @@
...
+            @Suppress("UnsafeThirdPartyFunctionCall") // Only throws if null is passed
+            val existing = aggregationMap.putIfAbsent(key, stats)
+            existing?.recordEvaluation(timestamp, errorMessage)
+        }
+
+        if (aggregationMap.size < maxAggregations) {
```
0xnm: note:

```
* Bear in mind that the results of aggregate status methods including
 * {@code size}, {@code isEmpty}, and {@code containsValue} are typically
 * useful only when a map is not undergoing concurrent updates in other threads.
```

but probably it is fine is `.size` call is off by a bit

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,121 @@
...
+    private val context: EvaluationContext,
+    private val service: String?,
+    private val rumApplicationId: String?,
+    errorMessage: String?
+) {
+    // All field access is synchronized - see recordEvaluation() and toEvaluationEvent()
+    private var count: Int = 1
+    private var firstEvaluation: Long = firstTimestamp
+    private var lastEvaluation: Long = firstTimestamp
+    private var lastErrorMessage: String? = errorMessage
```
0xnm: I think this can be done simpler, without mutability and synchronisation primitives.

`AggregationStats` can be just a `data` class, meaning when update is needed `.copy` method is called at the usage side.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
```diff
@@ -0,0 +1,121 @@
...
+        // Build context with Datadog-specific information
+        val eventContext = FlagEvaluation.Context(
+            evaluation = null, // Evaluation context reserved for future use
+            dd = FlagEvaluation.Dd(
+                service = service,
+                rum = rumApplicationId?.let { appId ->
+                    FlagEvaluation.Rum(
+                        application = FlagEvaluation.Application(id = appId),
+                        view = aggregationKey.viewName?.let { viewName ->
+                            FlagEvaluation.View(url = viewName)
```
0xnm, line 47: Generally speaking View URL != View Name. Is there a mistake here?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,105 @@
...
+ * Uses a [ReentrantReadWriteLock] to allow concurrent [record] calls while
+ * ensuring exclusive access during [drain].
+ */
+internal class EvaluationAggregator(
+    private val maxAggregations: Int
+) {
+    @Volatile
+    private var aggregationMap = ConcurrentHashMap<AggregationKey, AggregationStats>()
+
+    private val mapLock = ReentrantReadWriteLock()
```
0xnm: if we guard concurrent access to this class from the processor side, why do we need another guard here?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
dd-oleksii: This is making an unnecessary copy of the map

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
dd-oleksii: Yes, that's intended. The lock is protecting the `aggregationMap` field (i.e. reference) and `putIfAbsent` does not modify it. `ConcurrentHashMap` is protecting its own writes

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
dd-oleksii: That's OK/expected. We re-check the size below while holding an exclusive lock

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
dd-oleksii: This would require outside synchronization via `ConcurrentHashMap.compute()` which is sadly not available in our min API level

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
dd-oleksii: The processor does not prevent concurrent access to the aggregator.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
0xnm: It will, but it is on purpose. Advantage is removing the `@Volatile` and making property itself immitable.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
0xnm: `compute` is just a higher-level abstraction on top of simpler operations, the necessary logic can be implemented manually.

Overall, this PR has concurrency primitives at 3 different levels, which is a bit too much in my opinion. It is better to have immutable structures instead to reduce that.

dd-oleksii: > `compute` is just a higher-level abstraction on top of simpler operations, the necessary logic can be implemented manually.

`compute` has different atomicity guarantees compared to combining multiple other functions, so we can't implement it manually from simpler operations

I agree that concurrency is complicated here and I too would prefer immutable structures. It is possible to make stats class immutable but this would push the locking outwards, and I don't think it would make for a simpler code. Again, `compute()` is exactly the locking pattern that we need but unfortunately we can't replicate it

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
dd-oleksii: How does this help our users?

`.toMap()` + `.clear()` isn't much simpler than two assignments (and arguably not simpler at all) but it is slower.

0xnm: It is a trade-off allowing to remove concurrency-related code and mutability. Immutability has some negligible performance penalty, but it removes concurrency-related issues.

I don't think the performance here is a concern given expected call frequency of this code block.

dd-oleksii: That's true in general. However, in this specific case, copying doesn't allow us to remove any concurrency-related code. Even with copying, we still need the lock here, so that the drain is atomic and we don't under-count events.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventWriter.kt
typotter: Since we are buffering and aggregating events in memory, we are always dealing with them in a batch, so I don't anticipate a need to write a single record (outside of the batch coincidentally being 1).

otoh, `writeAll` feels more in line with Kotlin convention for writing a list of items.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: On `ios`, we don't have `view_url` in the RUM context, so we opted to use the human-friendly name on both platforms for consistency.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
typotter: For a separation of concerns, I encapsulated the aggregation work in its own class with its own internal consistency/thread safety. Its operations don't directly interleave with the locking required around flushing

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
typotter: done

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: done.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
dd-oleksii: Checked `ConcurrentHashMap` again. [`merge()`](https://developer.android.com/reference/java/util/concurrent/ConcurrentHashMap#merge(K,%20V,%20java.util.function.BiFunction%3C?%20super%20V,?%20super%20V,?%20extends%20V%3E)) could be useful instead of `compute` but it's also API level 24, so doesn't help.

We could use [`replace(key, oldValue, newValue)`](https://developer.android.com/reference/java/util/concurrent/ConcurrentHashMap#replace(K,%20V,%20V)) in a compare-and-swap loop but that's more complicated than a synchronized block and would perform worse under contention.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationKey.kt
typotter: removed defaults

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
typotter: > Overall, this PR has concurrency primitives at 3 different levels, which is a bit too much in my opinion. It is better to have immutable structures instead to reduce that.

I agree that the complexity is high in this PR and I could certainly split it into even smaller chunks. Different locking is needed for the different resources (or groups of resources) under contention. Separating the locks as much as possible seemed a good way to encapsulate each resource's access under contention and simplify the callsites.

What should we optimize for here? Centrality of lock logic? Simplicity/cardinality of locks/lock types? Precision of contention control? Performance/performance under contention? Easy to digest architecture?

dd-oleksii (dismissed):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/AggregationStats.kt
0xnm: I drafted a small proposal to reduce the number of concurrency primitives usage https://github.com/DataDog/dd-sdk-android/commit/33544d746f8ed04d35bb36d986dff4d39537af5f#r177031080 by making a trade-off for the immutability (which is more Kotlin-idiomatic). Feel free to adopt it or to reject it, whatever works for you. Performance penalty is negligible, but in exchange we reduce the number of code sites where we can run into concurrency bugs (which are hard to debug).

https://github.com/DataDog/dd-sdk-android/blob/117049fb3a739a0089a90bda12f67b2264921b1f/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestFactory.kt#L36

There we leverage event metadata to do the lightweight deserialization before the upload to drop events we are not interested in.

As of aggregation logic itself: we have a bit similar place in the SDK, where we reduce views, but there is it is in the boundaries of a single batch file and also the same model is used as a result (while here models are different).

0xnm: @dd-oleksii Cannot answer your comments on the commit directly, because new Github UI doesn't show them to me.

> 🐛 it's possible that timestamp is less than lastEvaluation, in which case lastEvaluation should not be updated. Similarly for firstEvaluation — it's possible that timestamp < firstEvlauation

Yes, it is a draft commit, I didn't 100% port this logic, but it is easy to fix with `min` / `max` functions.

> 🐛 Splitting size check and drain into different lock sections introduces a race condition and allows draining multiple times when size limit is exceeded (the number of drains is not limited to two).

Yes, a small sacrifice to have lighter locking. Worst case we just drain preliminary. Otherwise can be addressed by doing the size check again in the drain function inside the lock.

typotter: Thanks @0xnm .
I have applied the proposed change, fixed the timestamp bug and shifted the locking to allow a second check of the map size within the lock in order to limit to 1 drain. PTAL

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
```diff
@@ -0,0 +1,113 @@
...
+            )
+            aggregationMap.size
+        }
+
+        if (mapSize < maxAggregations) {
+            return emptyList()
+        }
+
+        // Re-check inside lock to ensure only one thread drains
+        val drained = mapLock.withLock {
```
dd-oleksii, line 85: minor: given this is the same lock we re-acquire, we don't need to release it and can simply proceed to draining.
```suggestion
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -0,0 +1,152 @@
...
+            rumApplicationId = rumApplicationId,
+            rumViewName = rumViewName,
+            variantKey = variantKey,
+            allocationKey = allocationKey,
+            errorCode = errorCode,
+            errorMessage = errorMessage
+        )
+
+        if (drainedEvents.isNotEmpty()) {
+            writer.writeAll(drainedEvents)
```
dd-oleksii, line 74: re-schedule periodic flush?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
typotter: 🧑‍🍳

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: #3178

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/aggregation/EvaluationAggregator.kt
typotter: #3178

PR #3146, opened 2026-01-22
[FFL-1720] Evaluation Logging: Storage & Network Infrastructure

typotter:
🔲 Integration & Configuration (PR #3147)
👉 **Storage & Network Infrastructure** (this PR)
☑️ Aggregation Engine & Test Utilities (PR #3145)
☑️FlagEvaluation Schema (PR #3166)

_Datadog Internal_

### What does this PR do?

Implements the infrastructure for persisting evaluation events to disk and uploading them to Datadog's intake endpoint. Connects the aggregation engine (PR #3145) to the SDK Core's storage and network layer.

### Motivation

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt
@@ -9,12 +9,14 @@ package com.datadog.android.flags
 import com.datadog.android.Datadog
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.SdkCore
+import com.datadog.android.api.feature.Feature.Companion.FLAGS_EVALUATIONS_FEATURE_NAME
 import com.datadog.android.api.feature.Feature.Companion.FLAGS_FEATURE_NAME
 import com.datadog.android.api.feature.Feature.Companion.RUM_FEATURE_NAME
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.core.InternalSdkCore
 import com.datadog.android.flags.internal.DatadogFlagsClient
 import com.datadog.android.flags.internal.DefaultRumEvaluationLogger
+import com.datadog.android.flags.internal.EvaluationsFeature
 import com.datadog.android.flags.internal.FlagsFeature
 import com.datadog.android.flags.internal.FlagsStateManager
 import com.datadog.android.flags.internal.LogWithPolicy
@@ -286,12 +288,16 @@ interface FlagsClient {
                     logWithPolicy = logWithPolicy
                 )
             }
+            val evaluationsFeature = sdkCore
+                .getFeature(FLAGS_EVALUATIONS_FEATURE_NAME)
+                ?.unwrap<EvaluationsFeature>()
 
             return flagsFeature.getOrRegisterNewClient(name) {
                 createInternal(
                     configuration = flagsFeature.flagsConfiguration,
                     featureSdkCore = sdkCore,
                     flagsFeature = flagsFeature,
+                    evaluationsFeature = evaluationsFeature,
                     name = name
                 )
             }
@@ -385,6 +391,7 @@ interface FlagsClient {
             configuration: FlagsConfiguration,
             featureSdkCore: FeatureSdkCore,
             flagsFeature: FlagsFeature,
+            evaluationsFeature: EvaluationsFeature?,
             name: String
         ): FlagsClient {
             val networkExecutorService = featureSdkCore.createSingleThreadExecutorService(
@@ -461,7 +468,8 @@ interface FlagsClient {
                     flagsRepository = flagsRepository,
                     flagsConfiguration = configuration,
                     rumEvaluationLogger = rumEvaluationLogger,
-                    processor = flagsFeature.processor,
+                    exposureProcessor = flagsFeature.processor,
+                    evaluationsFeature = evaluationsFeature,
                     flagStateManager = flagStateManager
                 )
             }
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
@@ -38,7 +38,8 @@ import org.json.JSONObject
  * @param flagsRepository local storage for precomputed flag values
  * @param flagsConfiguration configuration for the flags feature
  * @param rumEvaluationLogger responsible for sending flag evaluations to RUM.
- * @param processor responsible for writing exposure batches to be sent to flags backend.
+ * @param exposureProcessor responsible for writing exposure batches to be sent to flags backend.
+ * @param evaluationsFeature the evaluations subfeature for accessing processor and context (optional).
  * @param flagStateManager channel for managing state change listeners
  */
 @Suppress("TooManyFunctions") // All functions are necessary for flag evaluation lifecycle
@@ -48,7 +49,8 @@ internal class DatadogFlagsClient(
     private val flagsRepository: FlagsRepository,
     private val flagsConfiguration: FlagsConfiguration,
     private val rumEvaluationLogger: RumEvaluationLogger,
-    private val processor: EventsProcessor,
+    private val exposureProcessor: EventsProcessor,
+    private val evaluationsFeature: EvaluationsFeature?,
     private val flagStateManager: FlagsStateManager
 ) : FlagsClient {
 
@@ -196,10 +198,28 @@ internal class DatadogFlagsClient(
         resolveTracked(readAndParseAssignment(flagKey, defaultValue))
 
     private fun writeExposureEvent(name: String, data: UnparsedFlag, context: EvaluationContext) {
-        processor.processEvent(
+        exposureProcessor.processEvent(
             flagName = name,
             context = context,
             data = data
+        )
+    }
+
+    private fun writeEvaluationEvent(
+        flagKey: String,
+        context: EvaluationContext,
+        variantKey: String?,
+        allocationKey: String?,
+        errorCode: String?,
+        errorMessage: String?
+    ) {
+        evaluationsFeature?.processEvaluation(
+            flagKey = flagKey,
+            context = context,
+            variantKey = variantKey,
+            allocationKey = allocationKey,
+            errorCode = errorCode,
+            errorMessage = errorMessage
         )
     }
 
@@ -356,6 +376,10 @@ internal class DatadogFlagsClient(
                     { "Flag '${resolution.flagKey}': ${resolution.errorMessage}" }
                 )
             }
+
+            // Log evaluation events for errors
+            trackErrorResolution(resolution)
+
             resolution.defaultValue
         }
     }
@@ -431,6 +455,7 @@ internal class DatadogFlagsClient(
         flagValue: T,
         context: EvaluationContext
     ) {
+        // Exposure logging (only when doLog=true)
         if (flag.doLog) {
             if (flagsConfiguration.trackExposures) {
                 writeExposureEvent(flagKey, flag, context)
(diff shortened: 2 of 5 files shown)


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * Request format:
+ * - Endpoint: /api/v2/flagevaluations
+ * - Content-Type: application/json
+ * - Payload: BatchedFlagEvaluations with context wrapper
+ *
+ * @param internalLogger logger for errors
+ * @param customEvaluationEndpoint optional custom endpoint override
+ */
+internal class EvaluationsRequestFactory(
```
dd-oleksii, line 29: how is this class used? it doesn't seem to be used in this PR and I can't find how it is used in the next PR

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+    private fun buildTopLevelContext(context: DatadogContext): BatchedFlagEvaluations.Context =
+        BatchedFlagEvaluations.Context(
+            service = context.service,
+            version = context.version,
+            env = context.env,
+            device = buildDeviceContext(context),
+            os = buildOsContext(context),
+            geo = null, // Geo info not currently tracked
+            rum = buildRumContext(context)
+        )
```
dd-oleksii: major: constructing this context independently from aggregating events may (and will) misattribute events to wrong context. An obvious example of this is RUM view — the user may navigate to another view between flag evaluation and flush.

I couldn't find how this class is used but my understanding is that events are persistently stored to survive network outages and app restarts. This makes the problem even worse as now even app version / OS version might have changed between restarts.

We have to make this context part of the aggregation key (at least the rum view) and persist it along the events.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+            name = deviceInfo.osName,
+            version = deviceInfo.osVersion
+        )
+    }
+
+    private fun buildRumContext(context: DatadogContext): BatchedFlagEvaluations.Rum? {
+        // Extract RUM context from features context if available
+        val rumContext = context.featuresContext["rum"]
+        val applicationId = rumContext?.get("application_id") as? String
+        val viewUrl = rumContext?.get("view_url") as? String
```
sameerank: Is `viewUrl` required? Asking because `viewURL` doesn't appear to be among the supported attributes in the iOS SDK's RUM context: https://github.com/DataDog/dd-sdk-ios/blob/54674154872a53bf5fdd84511fb807e771435662/DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift#L10-L22

I think I'd need help with understanding why it's needed before proposing to add it for parity since a change to this model would be treading outside flagging

sameerank: Actually it looks optional in https://github.com/DataDog/dd-source/blob/main/domains/evp-workers/apps/flageval-worker/schemas/batchedflagevaluations.json#L32 so maybe it's okay if we don't quite match up here

typotter: Hmm. that's odd. It is in [Android](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumContext.kt#L40)
started a🧵 with Petzel et al for clarification

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt
typotter: It is correctly wired up now, in the `flags-evaluations` sub feature.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt
typotter: There is a RUM view block within the repeated `FlagEvaluation`. I do agree that we should include the rum view in the aggregation key for sure. updated.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt
typotter: changing to view name

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt
0xnm: I agree, that it can be misattribution here, because at the time of upload RUM view/session ID can (and probably will be) be different from the moment evaluation was written.

Ideally, we should associate this information at the time of writing event to the disk.

Also the same batch may contain events happening at the different RUM views, so I'm not sure that data model proposed fits the RUM model. The current data model also forces to deserialize the data written, and this is something we would like to avoid.

Probably `BatchedFlagEvaluations.Context` should be a part of `BatchedFlagEvaluations.FlagEvaluation` instead (`BatchedFlagEvaluations.FlagEvaluation.Context`?) and this will solve both the issue of misattribution and also the need to deserialize.

typotter: So we do log the context at evaluation time to the FlagEvaluation record.
It is not clear to me yet what the top level context is doing, but I have reached out to the EvP owner for clarification

I am updating the stack to use a different schema - one that does not have the top level fields (saves us the redundant field) and also uses NDJson format for uploading, so we can simply join the raw batch entries with newlines and avoid the deserialize/reserialize performance hit.

typotter: PR here: #3166

typotter: Updated to use the new schema

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactoryTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        // Then
+        assertThat(request1.id).isNotEqualTo(request2.id)
+        assertThat(request1.headers[RequestFactory.HEADER_REQUEST_ID])
+            .isNotEqualTo(request2.headers[RequestFactory.HEADER_REQUEST_ID])
+        assertThat(request1.id).isEqualTo(request1.headers[RequestFactory.HEADER_REQUEST_ID])
+        assertThat(request2.id).isEqualTo(request2.headers[RequestFactory.HEADER_REQUEST_ID])
+    }
+
+    @Test
+    fun `M preserve batch data order W create() { multiple batch events }`(
```
0xnm: does it actually matter? should we preserve the order?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/storage/EvaluationEventRecordWriter.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.datadog.android.flags.model.FlagEvaluation
+
+/**
+ * Persists serialized flag evaluation events to SDK Core storage.
+ *
+ * Events are written to the EVALUATIONS feature storage and will be uploaded
+ * to the EVP intake endpoint by the SDK Core's upload mechanism.
+ */
+internal class EvaluationEventRecordWriter(private val sdkCore: FeatureSdkCore) : EvaluationEventWriter {
+    override fun writeAll(events: List<FlagEvaluation>) {
```
0xnm, line 23: nit: maybe can be just `write`

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/EvaluationEventRecordWriterTest.kt
```diff
@@ -0,0 +1,230 @@
...
+                    writerScope.invoke(mockEventBatchWriter)
+                }
+            }
+
+        // When
+        testedWriter.writeAll(listOf(event))
+
+        // Then - should use withWriteContext
+        verify(mockFeature).withWriteContext(any(), any())
+    }
```
0xnm: I'm not sure this test makes sense, we already assert that events are written in the tests above

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/EvaluationEventRecordWriterTest.kt
```diff
@@ -0,0 +1,230 @@
...
+        // When
+        testedWriter.writeAll(events)
+
+        // Then - should not write anything
+        verify(mockSdkCore).getFeature(Feature.FLAGS_EVALUATIONS_FEATURE_NAME)
+        verifyNoInteractions(mockEventBatchWriter)
+    }
+
+    @Test
+    fun `M serialize events correctly W writeAll() { events with all fields }`() {
```
0xnm: I would suggest to start creating tests in the `reliability/single-fit` for that kind, it is a better place to verify E2E functionality.

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/storage/EvaluationEventRecordWriterTest.kt
```diff
@@ -0,0 +1,230 @@
...
+            eventType = eq(EventType.DEFAULT)
+        )
+
+        val capturedJson = String(eventCaptor.firstValue.data, Charsets.UTF_8)
+        val expectedJson = event.toJson().toString()
+        assertThat(capturedJson).isEqualTo(expectedJson)
+    }
+
+    @Test
+    fun `M write event with UTF-8 encoding W writeAll() { event provided }`(
```
0xnm: same for this test

0xnm (approved):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/storage/EvaluationEventRecordWriter.kt
```diff
@@ -0,0 +1,41 @@
...
+        if (events.isEmpty()) return
+
```
0xnm, line 25: one more thing: it is better to wrap a code block below `synchronized(this)` to avoid race condition when the same batch is accessed

dd-oleksii: What do you mean by "the same batch is accessed?" Where is it accessed from and how `synchronized(this)` helps?

0xnm: So far we have a single thread in the SDK which writes the data to the batch by appending to it (single batch can contain multiple events).

But this may change at some point and the _same_ writer can be called from multiple threads trying to access the single batch file. To avoid concurrent access to the single file it is better to add `synchronized(this)` (you can check other data writers in the SDK for the reference), this does safeguarding because there is a single instance of the writer attached to the particular feature.

dd-oleksii: `AsyncEventWriteScope` seems to already be synchronizing the writer block per-feature?

https://github.com/DataDog/dd-sdk-android/blob/19c66239fe54d6c1a962711f3b1b8cd7471301f8/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/AsyncEventWriteScope.kt#L24-L28

most data writers also don't have external synchronization (e.g., CoreTraceWriter, TelemetryEventHandler, DatadogLogHandler, WebViewRumEventConsumer, WebViewReplayEventConsumer, WebViewLogEventConsumer)

am I missing something?

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactoryTest.kt
typotter: doesn't matter, no. I suppose this is more or less just validating the request body construction

typotter: updated tests

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/storage/EvaluationEventRecordWriter.kt
typotter: will defer to thread in #3145

PR #3147, opened 2026-01-22
[FFL-1720] Evaluation Logging: Integration

typotter:
👉 **Integration & Configuration** (this PR)
☑️  Storage & Network Infrastructure (PR #3146)
☑️ Aggregation Engine & Test Utilities (PR #3145)
☑️ FlagEvaluation Schema (PR #3166)

_Datadog Internal_

### What does this PR do?

Wires the evaluation logging feature end-to-end by connecting the `EvaluationsFeature` to the flag evaluation flow. This is the final PR that enables evaluation logging in the Flags SDK.

### Motivation


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -439,6 +462,37 @@ internal class DatadogFlagsClient(
...
+        val context = flagsRepository.getEvaluationContext() ?: EvaluationContext.EMPTY
+        val errorFlag = PrecomputedFlag(
+            variationType = "string",
+            variationValue = resolution.defaultValue.toString(),
+            doLog = false,
+            allocationKey = "",
+            variationKey = "",
+            extraLogging = org.json.JSONObject(),
+            reason = ResolutionReason.ERROR.name
+        )
```
dd-oleksii: this seems hacky and confusing. Especially the obviously-wrong `string` variation type and variation value

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
```diff
@@ -32,8 +32,11 @@ import com.datadog.android.log.LogAttributes.RUM_APPLICATION_ID
  */
 internal typealias LogWithPolicy = (String, InternalLogger.Level) -> Unit
 
-internal class FlagsFeature(private val sdkCore: FeatureSdkCore, internal val flagsConfiguration: FlagsConfiguration) :
-    StorageBackedFeature,
+internal class FlagsFeature(
+    private val sdkCore: FeatureSdkCore,
+    internal val flagsConfiguration: FlagsConfiguration,
+    internal val evaluationsFeature: EvaluationsFeature? = null
```
0xnm: Having it as a property just to be read in the `FlagsClient` setup without a real usage inside `FlagsFeature` may look a bit confusing. Maybe simply get it in the `FlagsClient.Builder.build` as it is done for the `FlagsFeature`?

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -439,6 +464,37 @@ internal class DatadogFlagsClient(
...
+
+        // Create synthetic flag data for error evaluation
+        val context = flagsRepository.getEvaluationContext() ?: EvaluationContext.EMPTY
+        val errorFlag = PrecomputedFlag(
+            variationType = "string",
+            variationValue = resolution.defaultValue.toString(),
+            doLog = false,
+            allocationKey = "",
+            variationKey = "",
+            extraLogging = org.json.JSONObject(),
```
0xnm:
```suggestion
            extraLogging = JSONObject(),
```

We then need to have it as a static object instead of creating the same object all the time. But only if it is not supposed to be mutated during the downstream processing.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationsFeature.kt
```diff
@@ -44,12 +72,101 @@ internal class EvaluationsFeature(
             return
         }
 
+        // Register for context updates
+        sdkCore.setContextUpdateReceiver(this)
+
+        // Get initial service from DatadogContext
+        val internalSdkCore = sdkCore as? InternalSdkCore
+        val initialService = internalSdkCore?.getDatadogContext()?.service
```
0xnm: We should avoid doing this since it is a blocking call and may slow down SDK initialization.

What prevents from reading RUM context and SDK context during the event write stage? I believe it is evaluation aggregation where we write data to the disk only after some time.

Normally, service doesn't change during the SDK core lifetime, because there is no way to set if after SDK is initialized, so it can be read actually during the event write phase, not here.

0xnm (changes requested):

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationsFeature.kt
typotter: good catch.
Since we are mutating the `ddContext` onContextUpdate, I've moved the `internalSdkCore?.getDatadogContext()?.service` call to happen once there

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
typotter: this is fixed by removal of the unnecessary "synthetic flag data"

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/FlagsFeature.kt
typotter: done!

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationsFeature.kt
```diff
@@ -44,12 +72,97 @@ internal class EvaluationsFeature(
...
 
     // endregion
+
+    // region FeatureContextUpdateReceiver
+
+    override fun onContextUpdate(featureName: String, context: Map<String, Any?>) {
+        if (featureName == Feature.RUM_FEATURE_NAME) {
+            val currentContext = ddContext
+            val service = currentContext?.service
+                ?: (sdkCore as? InternalSdkCore)?.getDatadogContext()?.service
```
0xnm, line 119: This operation is blocking and may be expensive (and block upstream). Since `service` doesn't change over SDK lifetime, it is better to read it once during the feature initialization.

dd-oleksii: 🤔 You have [suggested](https://github.com/DataDog/dd-sdk-android/pull/3147#discussion_r2761639697) removing it from feature initialization before. Do you suggest reverting that or do something differently? wdyt about lazy init?

0xnm: I was suggesting reading it at the event write stage, this is not implemented. In this case it wouldn't be a blocking call, but an argument received at the `withWriteContext` callback.

And choosing between making such blocking call multiple times (at the `onContextUpdate`) vs once (at the feature initialization), the latter is a lighter approach.

dd-oleksii: Oh, I see, that makes sense. I guess one caveat is rum view name which is dynamic, so on write is too late as it might have changed. We need up-to-date context in `processEvaluation()` step

0xnm: > so on write is too late as it might have changed

SDK handles this case. When `withWriteContext` is called, it is processed on a single thread for the whole SDK, which guarantees sequential processing and also `DatadogContext` is captured at the moment of the call (it is immutable data class).

This means that if evaluation flow opens `withWriteContext` as early as possible, and even if there is a time-arranged sequence of client calls (may be done from different threads, so actions may happen in parallel) like `startView(A) -> evaluation -> startView(B)`, `evaluation` is guaranteed to receive a context _at the time_ once `startView(A)` finished its processing and it won't be affected by the `startView(B)` context change.

But I guess since for the evaluation there is an aggregation, and write happens not the moment event is captured, the flow can be different. If write isn't not necessary there is a `withContext` callback (but there is no callback which would provide only `WriteScope` though, although it is not a big deal).

dd-oleksii: Cool. It looks like `withContext` is what we need here

typotter: We are only getting the `ddContext` once as we are using the `currentContext.service` and falling through to `sdkCore.getDatadogContext()` only when the value is null. Also, this first call to `onContextUpdate` will happen _when the Feature is registered_ so it's not part of a regular queue of `onContextUpdate` calls (when the context actually is updated). It is, for all intents and purposes, tied into the initialization lifecycle already.

PR #3168, opened 2026-02-04
RUM-6171: Use FileObserver

kikoveiga:
### What does this PR do?

This is the follow-up on #2720, which was rolled back in #3124.
The fixes can be found in [this commit](https://github.com/DataDog/dd-sdk-android/pull/3168/changes/df19652ba0a6296be02ed1d1f2c8f27a180161d1), which include:
- Add `existsSafe` call inside `getReadableFile()`, so that we don't return a stale file that no longer exists on disk.
- Improve thread safety, by using `synchronized` and grouping the state into `currentBatchState: AtomicReference<CurrentBatchState>`.
- Instead of using direct equality between `event` and `FILE_OBSERVER_MASK`, use bitmask logic to guarantee matches when some additional bits might have been set.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -146,106 +168,117 @@ internal class BatchFileOrchestrator(
...
                 )
             )
         }
-        previousFile = newFile
-        previousFileItemCount = 1
-        lastFileAccessTimestamp = timeProvider.getDeviceTimestampMillis()
+        currentBatchState.set(CurrentBatchState(newFile, 1L, now))
         pendingFiles.incrementAndGet()
-        return newFile
+        synchronized(knownFiles) {
```
aleksandr-gringauz, line 241: Between `currentBatchState.set()` and `knownFiles.add()`, there's a brief window where:
1. `currentBatchState` points to the new file
2. `knownFiles` doesn't contain it yet

If `listBatchFiles()` is called during this window, it won't include the new file, causing `getReusableWritableFile()` to potentially return null when it shouldn't.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt
```diff
@@ -146,106 +168,117 @@ internal class BatchFileOrchestrator(
...
-            // 3. `lastKnownFile` was deleted
-            // In any case, we don't know the item count, so to be safe, we create a new file
+        if (!lastFile.existsSafe(internalLogger)) {
             return null
         }
 
-        val isRecentEnough = isFileRecent(lastFile, recentWriteDelayMs)
-        val hasRoomForMore = lastFile.lengthSafe(internalLogger) < config.maxBatchSize
-        val hasSlotForMore = (lastKnownFileItemCount < config.maxItemsPerBatch)
+        synchronized(currentBatchState) {
```
aleksandr-gringauz, line 256: IIUC `currentBatchState` is only used in `getWritableFile` which is called always from a single thread (`persistanceExecutorService`). So there is no need for `synchronized` here.

PR #3171, opened 2026-02-10
RUM-13974: [Cronet] Supporting APM on network layer for Cronet

satween:
### What does this PR do?                                                                                             
                                                                                                                        
This PR adds APM support for `Cronet` network layer (i.e., automatic tracing for all redirects that happen during the application-level request).

### Additional Notes

Note that `Cronet` does not provide any info for retries, so retry requests cannot be traced as it is done in
`OkHttp`.


dd-sdk-android-core/api/apiSurface
```diff
@@ -332,11 +333,35 @@ object com.datadog.android.core.internal.net.HttpSpec
     const val OPTIONS: String
     const val CONNECT: String
     fun values()
+    fun isMethodWithBody(String)
```
0xnm: It seems something got wrong, since we have now `HttpSpec` in both `dd-sdk-android-core` and `dd-sdk-android-internal`. Shouldn't it be only in the internal module?

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogRequestCallback.kt
```diff
@@ -5,27 +5,73 @@
...
+
+        finishTracingForSuccessfulRequest(info)
+
+        val redirectRequest = RedirectTrackingUrlRequest(request).also {
+            delegate.onRedirectReceived(it, info, newLocationUrl)
+        }
+
+        if (redirectRequest.wasFollowRedirectCalled) {
+            requestTracingSnapshot.get().requestInfo?.instrumentRedirect(newLocationUrl)
+        } else {
```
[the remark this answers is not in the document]
satween: Hmm, i think that chance for that way of usage is quite low as untill this callback isn't called - there will be no redirect. But I refactored a bit anyway 👍

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogRequestCallback.kt
```diff
@@ -38,37 +84,47 @@ internal class DatadogRequestCallback(
...
-                CronetHttpResponseInfo(info)
-            )
-        }
-        delegate.onSucceeded(request, info)
+    private fun HttpRequestInfo.instrumentRedirect(redirectUrl: String?) {
+        if (this !is MutableHttpRequestInfo || redirectUrl == null || apmNetworkInstrumentation == null) return
+
+        val redirectRequestInfo = newBuilder()
+            .setUrl(redirectUrl)
+            // Clear existing tracing headers so that each redirect creates a new independent span/trace
```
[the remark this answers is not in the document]
satween: Good catch, fixed

dd-sdk-android-core/api/apiSurface
satween: Yes, there was an issue with rebasing, but it should be fixed with one of the previous PR's

0xnm: But here it is in the `core` module, shouldn't it be in the `internal` module? It seems it doesn't make it to the public API.

satween: HttpSpec was moved to core in order to use constants from it inside `RequestInfoForgeryFactory` but we could handle this. Moved to internal.

dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/RequestInfoForgeryFactory.kt
```diff
@@ -20,7 +20,7 @@ class RequestInfoForgeryFactory : ForgeryFactory<HttpRequestInfo> {
             url = forge.aStringMatching("https://[a-z0-9]+\\.com"),
             headers = forge.aMap { aString() to aList { aString() } },
             contentType = forge.anAlphabeticalString(),
-            method = forge.anElementFrom(HttpSpec.Method.values()),
+            method = forge.anElementFrom("GET", "POST", "PUT", "HEAD", "DELETE", "TRACE", "OPTIONS", "CONNECT"),
```
0xnm: We probably still can reference it if we want by adding `dd-sdk-android-internal` to the dependencies of the `testFixtures` configuration.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationTest.kt
```diff
@@ -218,6 +219,7 @@ internal class ApmNetworkInstrumentationTest {
             assertThat(result.span).isEqualTo(mockSpan)
             assertThat(result.isSampled).isTrue()
             assertThat(result.sampleRate).isEqualTo(fakeSampleRate)
+            assertThat(result.sampleRate).isEqualTo(fakeSampleRate)
```
0xnm: seems like redundant change

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/tests/NetworkInstrumentationComparisonTest.kt
```diff
@@ -0,0 +1,192 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Integration tests comparing network instrumentation behavior across HTTP clients.
+ *
+ * Uses [CompositeHttpClientWrapper] to execute requests through all available clients
+ * and verify that they produce equivalent spans.
```
0xnm, line 32: that is nice validation approach

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/utils/ExecutionResultComparisonAssert.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.assertj.core.api.AbstractObjectAssert
+import org.assertj.core.api.Assertions.assertThat
+
+internal class ExecutionResultComparisonAssert(actual: Map<String, ClientExecutionResult>) :
+    AbstractObjectAssert<ExecutionResultComparisonAssert, Map<String, ClientExecutionResult>>(
+        actual,
+        ExecutionResultComparisonAssert::class.java
+    ) {
+
+    fun haveRequestMethod(expected: String) = apply {
```
0xnm, line 20: very nit: existing asserts are using `hasXXX`, not `haveXXX`. Doesn't matter much though, other than when searching.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/utils/ExecutionResultComparisonAssert.kt
```diff
@@ -0,0 +1,145 @@
...
+        actual.values.forEach {
+            assertThat(it.request?.method)
+                .overridingErrorMessage {
+                    "Expected client ${it.name} has request method: $expected but was ${it.request?.method}"
```
0xnm: seems wording here and for assertions below is wrong. Right now it is like

```
Expected client xyz has request method: _expected_ but was _actual_
```

Probably should be:

```
Expected client xyz to have request method: _expected_ but was _actual_
```

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/utils/NetworkTestConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+    val ALLOWED_METHODS = setOf(
+        HttpSpec.Method.HEAD,
+        HttpSpec.Method.GET,
+        HttpSpec.Method.PATCH,
+        HttpSpec.Method.PUT,
+        HttpSpec.Method.POST,
+        HttpSpec.Method.DELETE
+    )
+
+    fun asyncTest(block: suspend CoroutineScope.() -> Unit): Unit = runBlocking(block = block)
```
0xnm, line 47: is it worth to have a method which just calls another method without any modifications?

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/wrappers/HttpTestClientWrapper.kt
```diff
@@ -0,0 +1,27 @@
...
+internal interface HttpTestClientWrapper {
+    val name: String
+
+    fun shutdown()
+
+    suspend fun execute(request: TestRequest): ClientExecutionResult
+
+    companion object {
+        val tracedHosts = mapOf<String, Set<TracingHeaderType>>(
+            "10.0.2.2" to setOf(TracingHeaderType.DATADOG, TracingHeaderType.TRACECONTEXT),
```
0xnm: TIL about this loopback address to a host machine in case of Android Emulator. Worth adding a comment probably.

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRedirectTracingRequestWrapper.kt
```diff
@@ -0,0 +1,59 @@
...
+        if (this !is MutableHttpRequestInfo || redirectUrl == null || apmNetworkInstrumentation == null) return
+
+        val redirectRequestInfo = newBuilder()
+            .setUrl(redirectUrl)
+            // For 301, 302, 303 redirects the HTTP client rewrites POST/PUT/PATCH to GET and drops the body.
+            // We mirror that here so the span metadata (http.method) stays accurate.
+            .applyMethodRewrite(method, redirectStatusCode)
+            // Clear existing tracing headers so that each redirect creates a new independent span/trace
+            // This matches OkHttp behavior where each redirect is a separate root span
+            .also(apmNetworkInstrumentation::removeTracingHeaders)
```
0xnm, line 46: although maybe we shouldn't do this, but let's keep it

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRedirectTracingRequest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    private fun HttpRequestInfoBuilder.applyMethodRewrite(
+        originalMethod: String,
+        redirectStatusCode: Int?
+    ): HttpRequestInfoBuilder = apply {
+        if (redirectStatusCode != null &&
+            HttpSpec.StatusCode.redirectRequiresMethodRewrite(redirectStatusCode) &&
+            HttpSpec.Method.redirectAllowed(originalMethod)
+        ) {
+            setMethod(HttpSpec.Method.GET)
```
0xnm: why do we set `GET` here instead of original method? Is it a part of HTTP spec?

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogRequestCallback.kt
```diff
@@ -38,37 +83,26 @@ internal class DatadogRequestCallback(
...
-        traceState?.let {
-            apmNetworkInstrumentation?.onResponseFailed(
-                it,
-                error ?: IOException("Response failed")
-            )
-        }
-        delegate.onFailed(request, info, error)
+    internal companion object {
+        const val FOLLOW_REDIRECT_IS_NOT_DETECTED_MESSAGE =
+            "Redirect not followed - APM trace chain may be broken"
```
0xnm: maybe we need to improve this message, it is a bit vague imo: it is not clear what does it mean by "broken".

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/utils/ExecutionResultComparisonAssert.kt
satween: have here in order to highlight that not only one result is checked

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/utils/NetworkTestConfig.kt
satween: this is just a syntatic sugar in order to declare methods with `=`: 
```kotlin
  fun executionResultsMustBeSimilarWhen_clientError() = asyncTest { ... }
```

Because this code won't compile as test method should return `Unit`:
```kotlin
  fun executionResultsMustBeSimilarWhen_clientError() = runBlocking { ... }
```

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/wrappers/HttpTestClientWrapper.kt
satween: It's just a leftover from QUIC test server approach, removed

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRedirectTracingRequest.kt
satween: Yes, as far as I understand redirect forces clients to switch method to GET for historical reasons

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
```diff
@@ -0,0 +1,108 @@
...
+        } else {
+            DatadogCronetEngine(
+                delegate = delegate.build(),
+                apmNetworkInstrumentation = apmInstrumentation,
+                rumNetworkInstrumentation = rumNetworkInstrumentation
+            ).also { it.addRequestFinishedListener(requestFinishedListener) }
+        }
+    }
+
+    internal companion object Companion {
```
0xnm:
```suggestion
    internal companion object {
```

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRedirectTracingRequestWrapper.kt
```diff
@@ -0,0 +1,59 @@
...
+        super.followRedirect()
+    }
+
+    private fun HttpRequestInfo.instrumentRedirect(redirectUrl: String?) {
+        if (this !is MutableHttpRequestInfo || redirectUrl == null || apmNetworkInstrumentation == null) return
+
+        val redirectRequestInfo = newBuilder()
+            .setUrl(redirectUrl)
+            // For 301, 302, 303 redirects the HTTP client rewrites POST/PUT/PATCH to GET and drops the body.
+            // We mirror that here so the span metadata (http.method) stays accurate.
```
0xnm, line 41: although it is worth checking if `CronetEngine` indeed follows this

integrations/dd-sdk-android-cronet/src/test/kotlin/com/datadog/android/cronet/internal/CronetRequestCallbackTest.kt
```diff
@@ -0,0 +1,603 @@
...
+        verify(mockDelegate).onRedirectReceived(
+            isA<CronetRedirectTracingRequest>(),
+            eq(mockUrlResponseInfo),
+            eq(fakeNewLocationUrl)
+        )
+        verify(mockApmNetworkInstrumentation, never()).onResponseSucceeded(any(), any())
+    }
+
+    @Test
+    fun `M create redirect span W onRedirectReceived() {ALL, MutableHttpRequestInfo, follows redirect}`(
```
0xnm, line 229: Some tests, like this one, require a heavy setup. I'm wondering if we should keep unit tests like this or simply prefer integration tests instead.

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
```diff
@@ -0,0 +1,108 @@
...
+ * will be created for HTTP requests and tracing headers will be injected for first-party hosts.
+ * @param rumInstrumentationConfiguration optional RUM configuration. When provided, HTTP requests
+ * will be automatically tracked as RUM resources with timing information.
+ */
+@ExperimentalTracingApi
+@ExperimentalRumApi
+fun CronetEngine.Builder.configureDatadogInstrumentation(
+    apmInstrumentationConfiguration: ApmNetworkInstrumentationConfiguration? = null,
+    rumInstrumentationConfiguration: RumNetworkInstrumentationConfiguration? = null
+) = CronetIntegrationPlugin(this, rumInstrumentationConfiguration, apmInstrumentationConfiguration)
```
0xnm: I'm wondering if we can slightly change the API here:

```
fun CronetEngine.Builder.configureDatadogInstrumentation(
    apmInstrumentationConfiguration: ApmNetworkInstrumentationConfiguration? = null,
    rumInstrumentationConfiguration: RumNetworkInstrumentationConfiguration? = null,
    configure: CronetIntegrationPluginConfigure.() -> Unit
)
```

This way after calling `configureDatadogInstrumentation` we don't switch to the `CronetIntegrationPlugin`-specific methods in the fluent flow, but have only `.build` method then.

Ideally return type can be `CronetEngine.Builder` as well, but I guess it is not possible.

Not a call for the action, just thoughts.

0xnm (commented):
LGTM, but there is a lint failure. Also there are some comments from Codex above.

integrations/dd-sdk-android-cronet/src/test/kotlin/com/datadog/android/cronet/internal/CronetRequestCallbackTest.kt
satween: I will think on how to simplify them(or remove completely), but maybe in a separate PR

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
satween: The idea behind this API is to make Datadog's configuration applicable as the last chain to any networking library.

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRedirectTracingRequest.kt
```diff
@@ -0,0 +1,59 @@
...
+            // We mirror that here so the span metadata (http.method) stays accurate.
+            .applyMethodRewrite(method, redirectStatusCode)
+            // Clear existing tracing headers so that each redirect creates a new independent span/trace
+            // This matches OkHttp behavior where each redirect is a separate root span
+            .also(apmNetworkInstrumentation::removeTracingHeaders)
+            .build()
+
+        val newTraceState = apmNetworkInstrumentation.onRequest(redirectRequestInfo)
+
+        requestTracingState.set(newTraceState)
```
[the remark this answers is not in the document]
satween: That's true. The solution for now - for distributed tracing we supporting only application layer requests, for apm - networking layer instrumentation is also supported ( but would be visible only in apm explorer)

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
```diff
@@ -0,0 +1,108 @@
...
+     * @return an instrumented [CronetEngine] instance, or a plain one if no instrumentation was configured.
+     */
+    fun build(): CronetEngine {
+        val rumNetworkInstrumentation = rumInstrumentationConfiguration?.let {
+            _RumInternalProxy.createRumNetworkInstrumentation(CRONET_NETWORK_INSTRUMENTATION_NAME, it)
+        }
+
+        val apmInstrumentation = apmInstrumentationConfiguration?.let {
+            DatadogTracingToolkit.setRumInstrumentationActive(rumInstrumentationConfiguration != null, it)
+            DatadogTracingToolkit.setRumOriginIfPossible(it)
```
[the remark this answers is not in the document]
satween: reasonable, fixed with separate apm configuration only for distributed tracing purposes

0xnm (commented):
One thing I forgot to mention is `ApmNetworkTracingScope` members naming. I have a concern about `APPLICATION_LEVEL_REQUESTS_ONLY` - one may think that this is application requests vs application+libraries (all) requests tracking, because some observability SDKs have a possibility to instrument HTTP clients implicitly, without having direct access to them.

Maybe we should come with better name, especially considering that we are trying to bring OkHttp-native concept to the generic HTTP instrumentation.

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRedirectTracingRequestWrapper.kt
satween: added link to the source code

OliviaShoup:
thanks for the PR @satween! i've made an editorial review card so a docs team member can take a look: https://datadoghq.atlassian.net/browse/DOCS-13466

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRequestCallback.kt
```diff
@@ -0,0 +1,137 @@
...
+
+        rumNetworkInstrumentation?.apply {
+            startResource(finalRequestInfo)
+            sendWaitForResourceTimingEvent(finalRequestInfo)
+        }
+
+        apmNetworkInstrumentation?.onRequest(finalRequestInfo)
+            .also(apmTracingStateHolder::set)
+
+        return distributedTracingState ?: RequestTracingState(initialRequestInfo.newBuilder())
```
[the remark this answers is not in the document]
satween: This is not correct. distributedTracingState is being used only with RUM. apmNetworkInstrumentation is responsible for independent tracing and is not aimed to be integrated into request headers. The same way current instrumentation re-uses TracingInterceptor logic inside DatadogInterceptor.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/network/HttpSpec.kt
```diff
@@ -0,0 +1,194 @@
...
+         * Returns a list of common HTTP client error status codes (4xx).
+         * @return list containing BAD_REQUEST, UNAUTHORIZED, FORBIDDEN, and NOT_FOUND
+         */
+        fun clientErrors() = listOf(BAD_REQUEST, UNAUTHORIZED, FORBIDDEN, NOT_FOUND)
+
+        /**
+         * Returns a list of common HTTP server error status codes (5xx).
+         * @return list containing INTERNAL_ERROR, BAD_GATEWAY, and SERVICE_UNAVAILABLE
+         */
+        fun serverErrors() = listOf(INTERNAL_ERROR, BAD_GATEWAY, SERVICE_UNAVAILABLE)
```
0xnm: Shouldn't these list be exhaustive and rely on the first digit of the status code?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmInstrumentationConfigurationTest.kt
```diff
@@ -210,8 +219,7 @@ internal class ApmInstrumentationConfigurationTest {
         val result = testedBuilder.setGlobalTracerProvider(fakeProvider)
 
         // Then
-        assertThat(result).isSameAs(testedBuilder)
-        assertThat(testedBuilder.globalTracerProvider === fakeProvider).isTrue()
+        assertThat(result.globalTracerProvider === fakeProvider).isTrue()
```
0xnm: shouldn't is be simply `isSameAs` comparison usage here?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationTest.kt
```diff
@@ -533,14 +537,14 @@ internal class ApmNetworkInstrumentationTest {
     }
 
     @Test
-    fun `M not call traced request listener W onResponseSucceeded() {not sampled}`(
+    fun `M call traced request listener W onResponseSucceeded() {not sampled}`(
```
0xnm: there is a point raised by @codex regarding this

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationTest.kt
```diff
@@ -773,6 +787,281 @@ internal class ApmNetworkInstrumentationTest {
 
     // endregion
 
+    // region canSendSpan - EXCLUDE_INTERNAL_REDIRECTS
+
+    @Test
+    fun `M drop span W onResponseSucceeded() {APP_LEVEL, canSendSpan=false, RUM enabled, sampled}`(
```
0xnm: Should we make this test parametrized keeping only `canSendSpan=false` and maybe `APP_LEVEL` and extracting `RUM enabled/disabled` / `sampled/not sampled` as params? This would reduce number of tests to write.

features/dd-sdk-android-trace/src/testFixtures/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentationExt.kt
```diff
@@ -3,10 +3,8 @@
...
-package com.datadog.android.cronet.internal
 
-import com.datadog.android.trace.internal.net.RequestTraceState
+package com.datadog.android.trace.internal
 
-internal interface CronetInstrumentationStateHolder {
-    var traceState: RequestTraceState?
-}
+val ApmNetworkInstrumentation.traceOrigin
+    get() = this.traceOrigin
```
0xnm: is it just to make `traceOrigin` public? maybe it is better to make it in the class directly, then

integrations/dd-sdk-android-cronet/README.md
```diff
@@ -16,13 +18,168 @@ dependencies {
...
+
+  ```kotlin
+  @OptIn(ExperimentalTracingApi::class, ExperimentalRumApi::class)
+  val cronetEngine = CronetEngine.Builder(context)
+      .configureDatadogInstrumentation(
+          rumInstrumentationConfiguration = RumNetworkInstrumentationConfiguration(),
+          apmInstrumentationConfiguration = ApmNetworkInstrumentationConfiguration(
+              tracedHosts = listOf("api.example.com") // APM spans for backend only
+          ),
+          distributedTracingConfiguration = ApmNetworkInstrumentationConfiguration(
```
0xnm: I'm concerned about having two different configurations for the APM network instrumentation.

While the README maybe does some explanation, the API and doc are not intuitive, it took me some time to understand what is behind (sending only sampling decision, without any APM spans being generated).

Ideally, it would be nice to have a single property for the APM-related configuration, I think.

integrations/dd-sdk-android-cronet/README.md
```diff
@@ -16,13 +18,168 @@ dependencies {
...
+    .configureDatadogInstrumentation(
+        rumInstrumentationConfiguration = null,
+        apmInstrumentationConfiguration = ApmNetworkInstrumentationConfiguration(tracedHosts = listOf("api.example.com"))
+    )
+    .build()
+```
+
+
+
+### 4. RUM only (no APM tracing)
```
0xnm: Is it `addInterceptor(DatadogInterceptor)` as well? With APM feature disabled.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -129,63 +140,89 @@ class ApmNetworkInstrumentation internal constructor(
...
             .sample(request, traceSampler)
 
-    private fun RequestTraceState.onRequestCompleted(
+    private fun RequestTracingState.onRequestIntercepted(
         response: HttpResponseInfo?,
         throwable: Throwable?
     ) {
-        if (!isSampled || span == null) return
-        val request = requestBuilder.build()
+        if (span == null) return
```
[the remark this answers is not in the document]
satween: This is expected behavior.  Same logic presented in `TracingInterceptor`

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetHttpRequestInfo.kt
```diff
@@ -102,5 +94,26 @@ internal class CronetHttpRequestInfoBuilder(
...
+            requestContext.setUploadDataProvider(
+                body.uploadProvider,
+                body.executor
+            )
+        } else {
+            requestContext.setUploadDataProvider(null, null)
+        }
+    }
+
+    override fun build() = requestContext.copy().asCronetRequestInfo()
```
0xnm, line 109: do we acknowledge that `copy` makes keeps the same reference to the reference type?

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -1,7 +1,8 @@
-class com.datadog.android.trace.ApmNetworkInstrumentationConfiguration
-  constructor(Map<String, Set<TracingHeaderType>>)
+data class com.datadog.android.trace.ApmNetworkInstrumentationConfiguration
+  constructor(Map<String, Set<TracingHeaderType>>, String? = null, Boolean = true, String? = null, (com.datadog.android.api.SdkCore, Set<TracingHeaderType>) -> com.datadog.android.trace.api.tracer.DatadogTracer = DEFAULT_LOCAL_TRACER_FACTORY, TraceContextInjection = TraceContextInjection.SAMPLED, NetworkTracedRequestListener = NoOpNetworkTracedRequestListener(), com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan> = DeterministicTraceSampler(DEFAULT_TRACE_SAMPLE_RATE), () -> com.datadog.android.trace.api.tracer.DatadogTracer? = { GlobalDatadogTracer.getOrNull() }, ApmNetworkTracingScope = ApmNetworkTracingScope.ALL)
```
0xnm: Is there a need for data class? I guess it is to leverage `copy`?

It seems that many things are exposed now.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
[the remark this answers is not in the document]
0xnm: It seems it relies on the listener doc. We should update it then. Right now it says:

> Won't be called if [com.datadog.android.api.instrumentation.network.HttpRequestInfo] wasn't sampled.

integrations/dd-sdk-android-cronet/README.md
```diff
@@ -16,13 +18,168 @@ dependencies {
...
+- Injects the tracing headers into the request before it is sent to the network.
+- Does **not** send the local span to the APM backend — it exists only to carry the trace context.
+
+The modified request (with headers) is then passed to both RUM resource tracking and to the APM instrumentation (if configured), ensuring that the trace context propagates consistently.
+
+### Default behavior
+
+When `distributedTracingConfiguration` is not specified, the SDK reuses the `apmInstrumentationConfiguration` host list for header injection. Only requests to hosts listed in `apmInstrumentationConfiguration` receive tracing headers, and the same hosts get client-side APM spans.
+
+### Customising with `distributedTracingConfiguration`
```
brett0000FF:
```suggestion
### Customizing with `distributedTracingConfiguration`
```

integrations/dd-sdk-android-cronet/README.md
```diff
@@ -16,13 +18,168 @@ dependencies {
...
+Below are common configurations compared with their OkHttp equivalents.
+
+
+### 1. RUM + APM full tracing (recommended)
+
+**OkHttp equivalent:** `addInterceptor(DatadogInterceptor)` + `addNetworkInterceptor(TracingInterceptor)`
+
+Tracks HTTP requests as RUM Resources and creates APM spans at both application and network levels. Redirect hops and retries get individual spans, providing full visibility into the request lifecycle.
+
+```kotlin
```
brett0000FF, line 37: It would be helpful to show import statements for at least the first example. Lack of import context often causes friction for users.

brett0000FF (dismissed):

integrations/dd-sdk-android-cronet/README.md
satween: Thanks for the suggestion. Fixed.

integrations/dd-sdk-android-cronet/README.md
satween: Yes, this is kind'a adhoc solution to keep parity with existing okhttp configuration. Moved into separate method in order to simplify configuration

features/dd-sdk-android-trace/api/apiSurface
satween: yes, fixed back to non-data class with manual `copy` method implementation

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetHttpRequestInfo.kt
satween: yes, we copy only those ones which is possible to copy. This is a limitation of a Cronet framework

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationTest.kt
satween: maybe we could address this later in further PRs

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationTest.kt
satween: This is not correct. Both tests required to test the behavior which will call listener's method independently from `isSampled` parameter. I've refactored those methods a  bit in order to use forgery boolean parameter.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/network/HttpSpec.kt
```diff
@@ -0,0 +1,244 @@
...
+         * Returns a list of common HTTP client error status codes (4xx).
+         * @return list containing BAD_REQUEST, UNAUTHORIZED, FORBIDDEN, and NOT_FOUND
+         */
+        fun clientErrors() = values().filter { it in BAD_REQUEST until INTERNAL_ERROR }
+
+        /**
+         * Returns a list of common HTTP server error status codes (5xx).
+         * @return list containing INTERNAL_ERROR, BAD_GATEWAY, and SERVICE_UNAVAILABLE
+         */
+        fun serverErrors() = values().filter { it in INTERNAL_ERROR..NETWORK_AUTHENTICATION_REQUIRED }
```
0xnm: This breaks long-term forward compatibility if anything above 511 is added. In fact it is worth only to check first (`code / 100 == 5`) digit, if it is 5 or not, or use 600 as exclusive upper boundary.

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -1,23 +1,26 @@
 class com.datadog.android.trace.ApmNetworkInstrumentationConfiguration
-  constructor(Map<String, Set<TracingHeaderType>>)
+  constructor(Map<String, Set<TracingHeaderType>>, String? = null, Boolean = true, String? = null, (com.datadog.android.api.SdkCore, Set<TracingHeaderType>) -> com.datadog.android.trace.api.tracer.DatadogTracer = DEFAULT_LOCAL_TRACER_FACTORY, TraceContextInjection = TraceContextInjection.SAMPLED, NetworkTracedRequestListener = NoOpNetworkTracedRequestListener(), com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan> = DeterministicTraceSampler(DEFAULT_TRACE_SAMPLE_RATE), () -> com.datadog.android.trace.api.tracer.DatadogTracer? = { GlobalDatadogTracer.getOrNull() }, ApmNetworkTracingScope = ApmNetworkTracingScope.ALL)
```
0xnm: This constructor doesn't need to be public, I don't see its usages outside of this module.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmInstrumentationConfigurationTest.kt
```diff
@@ -210,8 +219,7 @@ internal class ApmInstrumentationConfigurationTest {
         val result = testedBuilder.setGlobalTracerProvider(fakeProvider)
 
         // Then
-        assertThat(result).isSameAs(testedBuilder)
-        assertThat(testedBuilder.globalTracerProvider === fakeProvider).isTrue()
+        assertThat<() -> DatadogTracer?>(result.globalTracerProvider).isSameAs(fakeProvider)
```
0xnm, line 211: seems like there is no need to specify the type explicitly here, `isSameAs` is using reference comparison anyway.

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
```diff
@@ -0,0 +1,157 @@
...
+     * Note that if [apmInstrumentationConfiguration] is not null and this method is called, custom configuration
+     * will be used for RUM-APM linking.
+     *
+     * @param distributedTraceConfiguration the distributed tracing configuration to use
+     */
+    @ExperimentalTraceApi
+    fun setCustomDistributedTraceConfiguration(distributedTraceConfiguration: ApmNetworkInstrumentationConfiguration) =
+        apply {
+            distributedTracingInstrumentationConfiguration = distributedTraceConfiguration
+        }
```
0xnm: I still find this method very confusing, because then it is possible to pass `ApmNetworkInstrumentationConfiguration` in the `CronetEngine.Builder.configureDatadogInstrumentation` and pass another instance (or same by mistake?) here after. Do we even need all the properties of  `ApmNetworkInstrumentationConfiguration` for the `distributedTraceConfiguration`?

integrations/dd-sdk-android-cronet/src/test/kotlin/com/datadog/android/cronet/internal/CronetRedirectTracingRequestTest.kt
```diff
@@ -0,0 +1,374 @@
...
+    @Test
+    fun `M return delegate W delegate property`() {
+        // Then
+        assertThat(testedRequest.delegate).isSameAs(mockDelegate)
+    }
+
+    // region followRedirect instrumentation
+
+    @Test
+    fun `M create new trace state W followRedirect() {301 redirect, POST method}`() {
```
0xnm: nit: Several tests below are basically the same, the only difference is status code. We can have a single test where status code is selected randomly.

integrations/dd-sdk-android-cronet/README.md
```diff
@@ -16,13 +18,161 @@ dependencies {
...
 **Note**: Datadog uses `cronet-api` version `141.7340.3` and has verified compatibility with the `play-services-cronet` implementation at version `17.0.1`.
 
-2. Replace `CronetEngine.Builder` with `DatadogCronetEngine.Builder`:
+## Configuration
+
+Use the `CronetEngine.Builder.configureDatadogInstrumentation()` extension to enable
+instrumentation. You can enable RUM, APM, or both depending on your needs. The extension returns a
+`CronetIntegrationPlugin` that can be further configured before calling `build()`.
+
+Below are common configurations compared with their OkHttp equivalents.
```
0xnm:
```suggestion
Below are common configurations compared with their [OkHttp equivalents](https://github.com/DataDog/dd-sdk-android/tree/develop/integrations/dd-sdk-android-okhttp).
```

integrations/dd-sdk-android-cronet/README.md
```diff
@@ -16,13 +18,161 @@ dependencies {
...
 
 ```kotlin
-@OptIn(ExperimentalRumApi::class)
-private val cronetEngine: CronetEngine = DatadogCronetEngine.Builder(application)
-    //...
+@OptIn(ExperimentalTraceApi::class, ExperimentalRumApi::class)
+val cronetEngine = CronetEngine.Builder(context)
+    .configureDatadogInstrumentation(
+        apmInstrumentationConfiguration = ApmNetworkInstrumentationConfiguration(tracedHosts = listOf("api.example.com"))
+            .setTraceScope(ApmNetworkTracingScope.EXCLUDE_INTERNAL_REDIRECTS),
```
0xnm: Should we keep `ALL` as default? Maybe we should set `EXCLUDE_INTERNAL_REDIRECTS` as default?

integrations/dd-sdk-android-cronet/README.md
```diff
@@ -16,13 +18,161 @@ dependencies {
...
+```kotlin
+Trace.enable(...)
+Rum.enable(...)
+```
+
+### 3. RUM + distributed tracing (no client-side APM spans)
+
+When `apmInstrumentationConfiguration` is provided, RUM-APM linking is enabled by default — the SDK injects `x-datadog-trace-id` and `x-datadog-parent-id` headers into outgoing requests, allowing you to navigate between RUM Resources and APM Spans in the Datadog UI and build end-to-end distributed traces across client, backend, and other components.
 
+However, this also enables client-side APM spans (including redirect hops). If you want RUM-APM linking **without** client-side network spans, use `setCustomDistributedTraceConfiguration()` instead of `apmInstrumentationConfiguration`:
```
0xnm: But the current API allows having both, I'm wondering if we can avoid this. What will happen if both set?

0xnm (changes requested):
I've added some comments: the most important points are `ApmNetworkInstrumentationConfiguration` constructor visibility and `setCustomDistributedTraceConfiguration` (we need to see if we can the necessary the other way, more intuitive and less error-prone).

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
satween: After discussion in DM - we decided to add a custom method directly to `ApmInstrumentationConfiguration` that allows to disable client spans reporting to the intake and get rid from an additional configuration

integrations/dd-sdk-android-cronet/README.md
satween: switched to `EXCLUDE_INTERNAL_REDIRECTS` in new api

integrations/dd-sdk-android-cronet/README.md
satween: not a case anymore (see comments related to the `ApmInstrumentationConfiguration` class change)

features/dd-sdk-android-trace/api/apiSurface
satween: this constructor is public in order be closer to `DatadogInterceptor.Builder` method. But I refactoried a bit so nothing except tracingHost is exposed to public

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/ApmInstrumentationConfigurationTest.kt
satween: Nope, without this type tests won't compile

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -16,32 +16,48 @@ import com.datadog.android.trace.internal.ApmNetworkInstrumentation
...
+    internal var traceOrigin: String? = null,
+    internal var redacted404ResourceName: Boolean = true,
+    internal var sdkInstanceName: String? = null,
+    internal var localTracerFactory: (SdkCore, Set<TracingHeaderType>) -> DatadogTracer = DEFAULT_LOCAL_TRACER_FACTORY,
+    internal var traceContextInjection: TraceContextInjection = TraceContextInjection.SAMPLED,
+    internal var tracedRequestListener: NetworkTracedRequestListener = NoOpNetworkTracedRequestListener(),
+    internal var traceSampler: Sampler<DatadogSpan> = DeterministicTraceSampler(DEFAULT_TRACE_SAMPLE_RATE),
+    internal var globalTracerProvider: () -> DatadogTracer? = { GlobalDatadogTracer.getOrNull() },
+    internal var networkTracingScope: ApmNetworkTracingScope = ApmNetworkTracingScope.EXCLUDE_INTERNAL_REDIRECTS,
+    internal var headerPropagationOnly: Boolean = false
```
0xnm, line 52: note: this a bit differers from configuration patterns we have in the codebase, because we use `XXXConfiguration.Builder` instead of having mutable configuration, so there is inconsistency API-wise.

Although it is not a big deal.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -137,6 +171,54 @@ class ApmNetworkInstrumentationConfiguration(
...
+     *
+     * When called, the SDK will still inject distributed tracing headers
+     * (e.g. `x-datadog-trace-id`, `x-datadog-parent-id`) into outgoing requests for
+     * first-party hosts, enabling RUM-APM linking and end-to-end distributed traces.
+     * However, no client-side APM spans will be sent to the Datadog backend.
+     *
+     * This is useful when you want distributed tracing visibility without the overhead
+     * of client-side network spans.
+     */
+    fun headerPropagationOnly() = apply {
```
0xnm: minor: it is worth to call it `setHeaderPropagationOnly()` to have more explicit meaning

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -185,6 +268,8 @@ class ApmNetworkInstrumentationConfiguration(
             return tracedHosts.filterKeys { sanitizedHosts.contains(it) }
         }
 
+        private fun Map<String, Set<TracingHeaderType>>.deepCopy() = mapValues { (_, v) -> v.toHashSet() }
```
0xnm: nit

```suggestion
        private fun Map<String, Set<TracingHeaderType>>.deepCopy() = mapValues { it.value.toSet() }
```

0xnm (dismissed):

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
```diff
@@ -0,0 +1,137 @@
...
+     * Builds the [CronetEngine] with Datadog instrumentation applied.
+     * @return an instrumented [CronetEngine] instance, or a plain one if no instrumentation was configured.
+     */
+    fun build(): CronetEngine {
+        val rumNetworkInstrumentation = rumInstrumentationConfiguration?.let {
+            _RumInternalProxy.createRumNetworkInstrumentation(CRONET_NETWORK_INSTRUMENTATION_NAME, it)
+        }
+
+        val apmInstrumentation = apmInstrumentationConfiguration
+            ?.takeUnless { it.isHeaderPropagationOnly() }
```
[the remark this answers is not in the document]
satween: This is expected behavior

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRequestCallback.kt
```diff
@@ -0,0 +1,137 @@
...
+    ) {
+        if (apmNetworkInstrumentation?.networkTracingScope != ApmNetworkTracingScope.ALL) {
+            // Redirects instrumented only with ApmNetworkTracingScope.ALL scope
+            delegate.onRedirectReceived(cronetRequest, cronetResponse, newLocationUrl)
+            return
+        }
+
+        val requestInfo = apmTracingStateHolder.get()?.createModifiedRequestInfo()
+
+        finishSuccessRequestTracing(cronetResponse)
```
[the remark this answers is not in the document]
satween: Cronet does not allow to modify headers inside redirects. Distributed tracing allowed for application requests only and it highlighted in documentation

0xnm (approved):

PR #3173, opened 2026-02-11
[RUM-14471] Add predicate to filter Activities for TTID reporting

hamorillo:
### What does this PR do?

Adds AppStartupActivityPredicate to RumConfiguration allowing customers to filter out interstitial Activities (e.g., auth screens) from **TTID** (Time To Initial Display) reporting.

Example of use:
```kotlin
RumConfiguration.Builder(applicationId)
      .setAppStartupActivityPredicate { activity ->
          // Exclude splash activity from TTID measurement
          activity !is SplashActivity
      }
      .build()

Rum.enable(rumConfiguration)
```

### Motivation

**TTID** was not being reported when interstitial Activities that launch another Activity in `onCreate()` and finish themselves never draw a frame. This allows customers to exclude such Activities from startup detection and get the right TTID.

### Additional Notes

- Introduces new public API: `AppStartupActivityPredicate` interface
- Provides `DefaultAppStartupActivityPredicate` that includes all Activities (backward compatible)

### DIFF
--- a/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/startup/AppStartupActivityPredicateTest.kt
+++ b/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/startup/AppStartupActivityPredicateTest.kt
@@ -13,6 +13,7 @@ import androidx.test.platform.app.InstrumentationRegistry
 import com.datadog.android.Datadog
 import com.datadog.android.privacy.TrackingConsent
 import com.datadog.android.rum.DdRumContentProvider
+import com.datadog.android.rum.ExperimentalRumApi
 import com.datadog.android.rum.Rum
 import com.datadog.android.rum.tracking.ActivityViewTrackingStrategy
 import com.datadog.android.sdk.integration.RuntimeConfig
@@ -104,6 +105,7 @@ internal class AppStartupActivityPredicateTest :
         keepRequests = true,
         trackingConsent = TrackingConsent.GRANTED
     ) {
+        @OptIn(ExperimentalRumApi::class)
         override fun beforeActivityLaunched() {
             super.beforeActivityLaunched()
             val config = RuntimeConfig.configBuilder()


hamorillo:
@codex review

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/DefaultAppStartupActivityPredicate.kt
```diff
@@ -0,0 +1,26 @@
...
+package com.datadog.android.rum.internal.startup
+
+import android.app.Activity
+import com.datadog.android.rum.startup.AppStartupActivityPredicate
+
+/**
+ * Default implementation of [AppStartupActivityPredicate] that allows all Activities
+ * to be tracked for app startup TTID measurement.
+ */
+internal class DefaultAppStartupActivityPredicate : AppStartupActivityPredicate {
```
aleksandr-gringauz: You can achieve the same by doing `object : DefaultAppStartupActivityPredicate : AppStartupActivityPredicate`. In this case there will be no need to override equals and hashCode.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -68,7 +77,13 @@ internal class RumAppStartupDetectorImpl(
...
         val now = timeProvider()
 
-        if (numberOfActivities == 1 && !isChangingConfigurations) {
+        val shouldTrackStartup = appStartupActivityPredicate.shouldTrackStartup(activity)
+
+        if (shouldTrackStartup) {
+            numberOfStartupActivities++
+        }
+
+        if (numberOfStartupActivities == 1 && !isChangingConfigurations && shouldTrackStartup) {
```
aleksandr-gringauz: Is this `&& shouldTrackStartup` check necessary?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -44,6 +47,12 @@ internal class RumAppStartupDetectorImpl(
 
     override fun onActivityDestroyed(activity: Activity) {
         numberOfActivities--
+
+        // Only decrement startup activities counter if this activity was counted
+        if (appStartupActivityPredicate.shouldTrackStartup(activity)) {
+            numberOfStartupActivities--
```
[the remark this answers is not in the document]
aleksandr-gringauz: Good point. I wonder if it is possible to store the result of `shouldTrackStartup` for this activity in `WeakHashMap<Activity, Boolean>` and use it in `onActivityDestroyed`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
ambushwork: Question, to better understand the issue, which scenario of interstitialActivity will trigger before this fix?

hamorillo: This is a case where the previous implementation can fail:

1. **SplashActivity** (interstitial) is launched first
     - `onActivityCreated` fires → Cold Start scenario detected ✅
     - Immediately calls `startActivity(MainActivity)` and `finish()` in `onCreate()`
     - **Never draws a frame** (finished before first draw)
     - TTID is **never reported** because the Activity didn't complete its first draw

  2. **MainActivity** is launched
     - Startup detection already fired for SplashActivity
     - No scenario detected for MainActivity ❌
     - MainActivity's TTID is **not measured**
     
I’m not sure if the example answers your question, though.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/DefaultAppStartupActivityPredicate.kt
hamorillo: I  initially implemented it as an `object`; however, I decided to switch it to a class for consistency with other predicates in the codebase.

Since its visibility is `internal`, we can switch it back to an object if needed. I don’t have a strong opinion here and am happy to change it if you think that’s better.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
hamorillo: I would say it is necessary for the following case:

1. MainActivity (included) created → scenario detected as Cold
  2. ExcludedActivity created → `numberOfStartupActivities` stays 1
  3. Without guard: Condition passes → `WarmAfterActivityDestroyed` scenario detected for excluded activity ❌
  4. With guard: Condition fails → No detection ✅

0xnm (commented):
It would be nice maybe to add some instrumented test here to make sure that:

* logic added indeed does what it is intended to do in the real application
* there is no regression

I understand that it will be hard to assert the output of such test, because I guess we want to validate TTID which may go left-right from run to run. But having something like this would be nice.

Not a blocker though, just an additional safety measure.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/DefaultAppStartupActivityPredicate.kt
aleksandr-gringauz: I think that `class` vs `object` doesn't matter much here. Feel free to choose whatever you want.

However I think that we shouldn't override `hashCode` and `equals` here. We are totally fine with the default referential equality for `AppStartupActivityPredicate` instances. How do you plan to use `DefaultAppStartupActivityPredicate` that requires it to have a different `equals`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: > Without guard: Condition passes → WarmAfterActivityDestroyed scenario detected for excluded activity

Indeed. Got it, thanks for the explanation.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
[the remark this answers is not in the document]
hamorillo: Done!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/DefaultAppStartupActivityPredicate.kt
hamorillo: Let's move with the `object` approach. Updated!

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -273,6 +274,8 @@ class com.datadog.android.rum.resource.RumResourceInputStream : java.io.InputStr
   override fun mark(Int)
   override fun reset()
   override fun close()
+interface com.datadog.android.rum.startup.AppStartupActivityPredicate
+  fun shouldTrackStartup(android.app.Activity): Boolean
```
0xnm, line 278: very nit: all other predicate interfaces in RUM module have similar method called `accept`. Doesn't matter much though since it is a functional interface, so it most probably will be used as lambda.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/startup/AppStartupActivityPredicateTest.kt
```diff
@@ -0,0 +1,130 @@
...
+            ExpectedVitalAppLaunchEvent(
+                appLaunchMetric = AppLaunchMetric.TTID
+            )
+        )
+
+        expectedEvents.add(
+            ExpectedVitalAppLaunchEvent(
+                appLaunchMetric = AppLaunchMetric.TTFD
+            )
+        )
```
0xnm, line 78: how do we know that these two is a result of the `MainContentActivity` and not of interstitial one?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/startup/AppStartupActivityPredicate.kt
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Predicate to determine which Activities should be considered for app startup
+ * Time To Initial Display (TTID) reporting.
+ *
+ * Use this to exclude "interstitial Activities" - Activities that are launched during
+ * app startup but immediately launch another Activity in their `onCreate()` method
+ * and call `finish()` on themselves (e.g., splash screens, authentication screens).
+ *
+ * These interstitial Activities may never draw a frame, which can prevent TTID from
+ * being reported when the actual main Activity is displayed.
```
aleksandr-gringauz: > and call `finish()` on themselves (e.g., splash screens, authentication screens).

If you implement a splash screen as activity, it probably draws a frame and doesn't call `finish` in `Activity.onCreate`.

I suggest we change this phrase a bit like this:

During application launch the SDK tracks the first `Activity` that is created and uses the time it draws its first frame to compute the TTID value. Use this if you want to skip some Activities and instead make the SDK consider the 2nd, 3rd Activity's first frame. Example use cases are:
1. Splash screens implemented as `Activity`.
2. Activities that launch another Activity in their `onCreate()` method and call `finish()` on themselves.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -44,6 +49,13 @@ internal class RumAppStartupDetectorImpl(
 
     override fun onActivityDestroyed(activity: Activity) {
         numberOfActivities--
+
+        // Use stored result instead of re-evaluating predicate to prevent counter corruption
+        // if predicate depends on mutable state (e.g., activity flags, runtime toggles)
+        if (trackedActivities.remove(activity) == true) {
+            numberOfStartupActivities--
+        }
```
aleksandr-gringauz: Question.

Is it possible to make `trackedActivities` a `Set` instead of a `Map` by doing `Collections.newSetFromMap(WeakHashMap<Activity, Unit>())` and remove `numberOfStartupActivities` entirely (and use `trackedActivities.size()` in its place)?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt
```diff
@@ -376,6 +377,38 @@ data class RumConfiguration internal constructor(
...
+         *     // Exclude AuthenticationActivity from TTID measurement
+         *     activity !is AuthenticationActivity
+         * }
+         * ```
+         *
+         * @param predicate The [AppStartupActivityPredicate] to determine which Activities
+         * should be measured for app startup TTID.
+         * @see [AppStartupActivityPredicate]
+         */
+        fun setAppStartupActivityPredicate(predicate: AppStartupActivityPredicate): Builder {
```
aleksandr-gringauz, line 408: Let's add `@ExperimentalRumApi` annotation to this. Just to be extra safe in case we want to change something. We can remove this later when we are sure it is fine.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/startup/AppStartupActivityPredicate.kt
aleksandr-gringauz: Also if we come up with a "just-works" solution we don't want to tell customers to use `AppStartupActivityPredicate` for the case of **"Activities that launch another Activity in their onCreate() method and call finish() on themselves."**.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/startup/AppStartupActivityPredicateTest.kt
aleksandr-gringauz: Good point. We put the information about current RUM View into the vital event [here](https://github.com/DataDog/dd-sdk-android/blob/369ea6cffc7a351be679ad1371c7759aef9e73f4/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumVitalAppLaunchEventHelper.kt#L105). The name of the view will be the name of the Activity. I wonder if we could use it to assert that it is `MainContentActivity`.

hamorillo: I've added an assert for that. Good point!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
hamorillo: Nice suggestion!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/startup/AppStartupActivityPredicate.kt
hamorillo: I think all suggestions have been implemented. Let's pause this PR until I finish with the "just-works" approach.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/ExpectedEvents.kt
```diff
@@ -61,6 +61,7 @@ internal data class ExpectedErrorEvent(
 
 internal data class ExpectedVitalAppLaunchEvent(
     val appLaunchMetric: AppLaunchMetric,
+    val viewArguments: Map<String, Any?> = mapOf(),
```
aleksandr-gringauz, line 64: I think it's better to leverage Kotlin type system more here instead of using untyped map<string, any>.

Something like this.

```
// introduce a new data class
data class AppLaunchRumView(val name: String) // Similar to what we already have for AppLaunchMetric enum

internal data class ExpectedVitalAppLaunchEvent(
    val appLaunchMetric: AppLaunchMetric,
    val rumView: AppLaunchRumView
)
```

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/RumTest.kt
```diff
@@ -24,6 +26,13 @@ internal abstract class RumTest<R : Activity, T : MockServerActivityTestRule<R>>
 
     protected abstract fun runInstrumentationScenario(mockServerRule: T): List<ExpectedEvent>
 
+    @After
+    fun tearDown() {
+        // Ensure SDK is stopped and wait for cleanup to complete
+        Datadog.stopInstance()
+        InstrumentationRegistry.getInstrumentation().waitForIdleSync()
+    }
```
aleksandr-gringauz, line 34: FYI this change affects some other tests as well. Not that I'm against it though, seems fine. Just a heads-up.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/ExpectedEvents.kt
hamorillo: Tbh, I did it that way in order to keep consistency with the other ExpectedEvents.

aleksandr-gringauz: OK. Leave it the way it is now. Not a big deal. I personally prefer strong types whenever possible though :)

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -68,7 +78,14 @@ internal class RumAppStartupDetectorImpl(
         numberOfActivities++
         val now = timeProvider()
 
-        if (numberOfActivities == 1 && !isChangingConfigurations) {
+        val shouldTrackStartup = appStartupActivityPredicate.shouldTrackStartup(activity)
+
+        if (shouldTrackStartup) {
+            @Suppress("UnsafeThirdPartyFunctionCall") // activity lifecycle callbacks run on main thread
```
aleksandr-gringauz: 1. I think the comment doesn't correspond to the `Suppress` statement.
2. You need to add some entries (related to `add`, `remove` methods) to `detekt_custom_safe_calls.yml`.

PR #3177, opened 2026-02-12
RUMS-5535:  Fix global attributes not copied when view stops

ambushwork:
### Current Issue

When customers call `GlobalRumMonitor.get().addAttribute(attribute.key, attribute.value)`,  they expect to see the custom attributes in all the following views, actions, resources and errors. There is an issue reported by a customer that the global attributes are missing in most of the views in the sessions, while they are available in all the actions, resources or errors.

### Cause

When a view is stopped, it is expected to copy the snapshot of custom attributes of the parent scope at the moment it stops, which is the case when customer calls stop view manually, but not the case if stop view is triggered by `StartView` of the new view. Thus, when `sendViewUpdate` is triggered by stopView like this, it is updated without the global attributes.

### Fix
apply the same snapshot copy logic into stop view when it's triggered by a new start view.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -559,7 +559,9 @@ internal open class RumViewScope(
         writeScope: EventWriteScope,
         writer: DataWriter<Any>
     ) {
-        stopScope(event, datadogContext, writeScope, writer)
+        stopScope(event, datadogContext, writeScope, writer) {
+            memoizedParentAttributes = parentScope.getCustomAttributes().toMap()
+        }
     }
 
     @WorkerThread
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeAttributePropagationTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeAttributePropagationTest.kt
@@ -572,6 +572,67 @@ internal class RumViewScopeAttributePropagationTest {
 
     // endregion
 
+    @Test
+    fun `M copy parent custom attributes to memoizedParentAttributes W handleEvent(StopView)`(
+        forge: Forge
+    ) {
+        // Given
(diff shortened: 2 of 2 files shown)


ambushwork:
@codex review

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
```diff
@@ -1537,6 +1537,65 @@ internal class RumViewScopeTest {
         assertThat(result).isNull()
     }
 
+    @Test
+    fun `M copy parent custom attributes to memoizedParentAttributes W handleEvent(StopView)`(
```
0xnm: Probably these tests should be in `RumViewScopeAttributePropagationTest`. Otherwise lgtm.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -559,7 +559,9 @@ internal open class RumViewScope(
         writeScope: EventWriteScope,
         writer: DataWriter<Any>
     ) {
-        stopScope(event, datadogContext, writeScope, writer)
+        stopScope(event, datadogContext, writeScope, writer) {
+            memoizedParentAttributes = parentScope.getCustomAttributes().toMap()
```
aleksandr-gringauz, line 563: Just a heads-up. [Here](https://github.com/DataDog/dd-sdk-android/blob/1b44a80c4b52171121fc7677ad777e6567fe2162/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L579) we also call `viewAttributes.putAll(event.attributes)`. I don't know if it is needed though.

0xnm: This event belongs to another view, so we don't need to copy its attributes.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScopeTest.kt
ambushwork: updated

aleksandr-gringauz (approved):

PR #3180, opened 2026-02-16
RUM-13441: [Cronet] Migrating okHttp to new instrumentation API

satween:
### What does this PR do?                                                                                                                                                  
                                         
  * Moves `OkHttp` instrumentation to the new networking-library agnostic approach.                                                                                          
  * Introduces a new plugin-based instrumentation API for OkHttp network monitoring, replacing the legacy `DatadogInterceptor`/`TracingInterceptor`/`DatadogEventListener`   
  approach  (all classes kept for the backward compatibility)                                                                                                                
    with a single unified entry point:                      

  ```kotlin
OkHttpClient.Builder()
    .configureDatadogInstrumentation(
        rumInstrumentationConfiguration = RumNetworkInstrumentationConfiguration(),
        apmInstrumentationConfiguration = ApmNetworkInstrumentationConfiguration(tracedHosts)
    )
    .build()
```
  same as for Cronet
```kotlin
CronetEngine.Builder()
  .configureDatadogInstrumentation(
      rumInstrumentationConfiguration = RumNetworkInstrumentationConfiguration(),
      apmInstrumentationConfiguration = ApmNetworkInstrumentationConfiguration(tracedHosts)
  )
  .build()
```
  Motivation

  - Moving instrumentation logic into library-agnostic approach
  - Simplifying and unifying instrumentation configuration for all networking libraries
  - Making Datadog instrumentation configuration latest in the configuration chain


instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/tests/NetworkInstrumentationComparisonTest.kt
```diff
@@ -189,4 +201,29 @@ internal class NetworkInstrumentationComparisonTest {
             .haveRequestUrl(url)
             .haveRequestMethod(method)
     }
+
+    @Test
+    @Repeat(10)
+    fun executionResultsMustBeSimilarWhen_rateLimited_okhttp_comparison() = asyncTest {
```
ambushwork, line 207: Seems the unit test is testing the equality not similarity, if so it should be renamed to `executionResultsMustBeSameWhen_rateLimited_okhttp_comparison`. Also, is it needed to execute 10 times on CI?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.datadog.android.trace.internal.ApmNetworkInstrumentation
+import okhttp3.Interceptor
+import okhttp3.Response
+
+internal class ApmInstrumentationOkHttpAdapter(
+    private val apmNetworkInstrumentation: ApmNetworkInstrumentation,
+    private val registry: RequestTracingStateRegistry
+) : Interceptor {
+    private val internalLogger: InternalLogger
+        get() = (apmNetworkInstrumentation.sdkCoreReference.get() as? FeatureSdkCore)?.internalLogger
```
ambushwork: is it necessary the `get()` here? anything mutable in this chain?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+package com.datadog.android.okhttp.internal
+
+import com.datadog.android.trace.internal.net.RequestTracingState
+import okhttp3.Call
+import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
+import java.util.function.BiFunction
+
+internal class RequestTracingStateRegistry {
```
ambushwork: I don't see any unit test for this class, is it expected?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val requestTracingStateByCall = ConcurrentHashMap<Call, RequestTracingState>()
+
+    fun register(call: Call) {
```
ambushwork: Question, what's the expected behavior if we try to `register` several times the same call here? according to the current logic it will generate a request with different uuid each time, is that expected?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * @param apmInstrumentationConfiguration optional APM tracing configuration. When provided, trace spans
+ * will be created for HTTP requests and tracing headers will be injected for first-party hosts.
+ * @param rumInstrumentationConfiguration optional RUM configuration. When provided, HTTP requests
+ * will be automatically tracked as RUM resources with timing information.
+ */
+@ExperimentalTracingApi
+@ExperimentalRumApi
+fun OkHttpClient.Builder.configureDatadogInstrumentation(
+    apmInstrumentationConfiguration: ApmNetworkInstrumentationConfiguration? = null,
```
ambushwork: I wonder if we should use null as default value, because here when user calls `configureDatadogInstrumentation()` it seems we provide a default instrumentation but it actually does nothing.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,143 @@
...
+) {
+    private var userEventListenerFactory: EventListener.Factory? = null
+
+    /**
+     * Configure a single client scoped listener that will receive all analytic events for this
+     * client.
+     *
+     * @see EventListener for semantics and restrictions on listener implementations.
+     */
+    fun eventListener(eventListener: EventListener) = apply {
```
aleksandr-gringauz: IIUC if the client uses the `eventListener` method from `OkHttpClient.Builder`  before calling the new `configureDatadogInstrumentation` method, but doesn't use this new `OkHttpIntegrationPlugin.eventListener` method, the client's listener will just silently stop working.

I'm not sure this is an intuitive behavior. 

However I understand that it is most likely impossible to do a "one method" setup of okhttp instrumentation without having this problem.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RumInstrumentationTimingsCounter.kt
```diff
@@ -0,0 +1,220 @@
...
+import okhttp3.Handshake
+import okhttp3.Protocol
+import okhttp3.Response
+import java.io.IOException
+import java.net.HttpURLConnection
+import java.net.InetAddress
+import java.net.InetSocketAddress
+import java.net.Proxy
+
+internal class RumInstrumentationTimingsCounter(
```
aleksandr-gringauz, line 24: I see that `DatadogEventListener` isn't removed. Is it intended to have 2 classes with identical implementations?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
satween: Tests weren't rename, fixed. `RumInstrumentationHttpRequestInfoRegistryTest` -> `RequestTracingStateRegistryTest`

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
satween: Yes, good notice. Kepp it nullable but changed to required (no default parameter). Same as for `CronetIntegrationPlugin`

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RumInstrumentationTimingsCounter.kt
satween: Yes, as it's public and we cannot remove them with minor update. After dogfooding and testing the new implementation `DatadogEventListener ` will be deprecated till next major release

aleksandr-gringauz: > Yes, as it's public and we cannot remove them with minor update.

yes, this is fine, I agree

I am more concerned about duplicating a large chunk of non-trivial code. Can the new listener just reuse the `DatadogEventListener` for example? Or could we extract the implementation into a separate class?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumNetworkInstrumentation.kt
```diff
@@ -49,14 +49,17 @@ class RumNetworkInstrumentation internal constructor(
...
-
-    internal val sdkCoreReference = SdkReference(sdkInstanceName) {
+    private val sdkCoreReference = SdkReference(sdkInstanceName) {
         it.networkMonitor?.apply {
             notifyInterceptorInstantiated()
             reportNetworkingLibraryType(libraryType)
         }
     }
 
+    /** SDK core instance. */
```
0xnm, line 60: very nit: this comment is useless

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,170 @@
...
+    fun eventListenerFactory(eventListenerFactory: EventListener.Factory) = apply {
+        this.userEventListenerFactory = eventListenerFactory
+    }
+
+    /**
+     * Builds the [OkHttpClient] with Datadog instrumentation applied.
+     * @return an instrumented [OkHttpClient] instance.
+     */
+    fun build(): OkHttpClient {
+        builtClient?.let { return it }
```
0xnm, line 69: what is the purpose of this? just a safeguard?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapter.kt
```diff
@@ -0,0 +1,69 @@
...
+import okhttp3.Interceptor
+import okhttp3.Response
+
+internal class ApmInstrumentationOkHttpAdapter(
+    private val apmNetworkInstrumentation: ApmNetworkInstrumentation,
+    private val registry: RequestTracingStateRegistry
+) : Interceptor {
+
+    private val sdkCore: FeatureSdkCore?
+        get() = apmNetworkInstrumentation.sdkCoreReference.get() as? FeatureSdkCore
```
0xnm: `apmNetworkInstrumentation` already exposes `sdkCore` property

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPluginTest.kt
```diff
@@ -0,0 +1,300 @@
...
+            apmInstrumentationConfiguration = null
+        )
+
+        // When
+        val firstResult = testedPlugin.build()
+        val secondResult = testedPlugin.build()
+
+        // Then
+        assertThat(secondResult).isSameAs(firstResult)
+        verify(mockBuilderDelegate, times(1)).build()
```
0xnm:
```suggestion
        verify(mockBuilderDelegate).build()
```

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapterTest.kt
```diff
@@ -0,0 +1,420 @@
...
+        whenever(mockChain.request()) doReturn fakeRequest
+        whenever(mockChain.call()) doReturn mockCall
+        whenever(mockApmNetworkInstrumentation.sdkCoreReference) doReturn mockSdkReference
+        whenever(mockSdkReference.get()) doReturn mockSdkCore
+        whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
+
+        val initialRequestBuilder = OkHttpRequestInfoBuilder(fakeRequest.newBuilder())
+        val initialState = RequestTracingState(tracedRequestInfoBuilder = initialRequestBuilder)
+        whenever(mockRegistry.update(eq(mockCall), any())).thenAnswer { invocation ->
+            @Suppress("UNCHECKED_CAST")
```
0xnm: there is no unchecked cast here

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        val taggedRequest = OkHttpRequestInfoBuilder(call.request().newBuilder())
+            .addTag(UUID::class.java, UUID.randomUUID())
+
+        requestTracingStateByCall[call] = RequestTracingState(taggedRequest)
+    }
+
+    fun update(
+        call: Call,
+        block: BiFunction<Call, RequestTracingState, RequestTracingState?>
```
0xnm: this is Java-like declaration and type is only available since API 24 https://developer.android.com/reference/java/util/function/BiFunction

Although it will be probably optimized by compiler, it is better to use Kotlin-like declaration:

```suggestion
        block: (Call, RequestTracingState) -> RequestTracingState?
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,53 @@
...
+    ): RequestTracingState? = requestTracingStateByCall.computeIfPresent(call, block)
```
0xnm: this call requires API 24, current min API is 23. Need to replace it.

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapterTest.kt
```diff
@@ -0,0 +1,420 @@
...
+            val block = invocation.getArgument<BiFunction<Call, RequestTracingState, RequestTracingState?>>(1)
```
0xnm:
```suggestion
            val block = invocation.getArgument<(Call, RequestTracingState) -> RequestTracingState?>(1)
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,170 @@
...
+ * [OkHttpIntegrationPlugin.build].
+ *
+ * @param apmInstrumentationConfiguration optional APM tracing configuration. When provided, trace spans
+ * will be created for HTTP requests and tracing headers will be injected for first-party hosts.
+ * @param rumInstrumentationConfiguration optional RUM configuration. When provided, HTTP requests
+ * will be automatically tracked as RUM resources with timing information.
+ */
+@ExperimentalTraceApi
+@ExperimentalRumApi
+fun OkHttpClient.Builder.configureDatadogInstrumentation(
```
0xnm: We need to deprecate then existing interceptors in favor of this

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistryTest.kt
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    @Test
+    fun `M add UUID tag W register()`() {
+        // When
+        testedRegistry.register(mockCall)
+        val result = testedRegistry.get(mockCall)
+
+        // Then
+        assertThat(checkNotNull(result).createModifiedRequestInfo().tag(UUID::class.java)).isNotNull
```
0xnm: better to split as in test above

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistryTest.kt
```diff
@@ -0,0 +1,262 @@
...
+            on { request() } doReturn fakeRequest2
+        }
+
+        // When
+        testedRegistry.register(mockCall)
+        testedRegistry.register(mockCall2)
+
+        // Then
+        assertThat(checkNotNull(testedRegistry.get(mockCall)).createModifiedRequestInfo().url).isEqualTo(fakeUrl)
+        assertThat(checkNotNull(testedRegistry.get(mockCall2)).createModifiedRequestInfo().url).isEqualTo(fakeUrl2)
```
0xnm: same, better to split for readability

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RumInstrumentationTimingsCounter.kt
satween: Well my idea was to ship this version to shopist and then compare that there is no significant drops (or spikes) in the data SDK produces. And after that deprecate `DatadogEventListener` and eventually remove them. If we gonna re-use `DatadogEventListener` inside `RumTimingCollector` an additional dogfooding iteration gonna be required, so I would prefer to not re-use code that we gonna depretate

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumNetworkInstrumentation.kt
satween: public property without documentation will lead to detekt failure

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
satween: Yes, otherwise multiple call to `build` will add multiple listeners to the client

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
satween: strange that AS didn't complain

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
satween: I would do that after dogfooding (before release) in a separate PR

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
satween: yes, this is kind'a tradeoff that we accept here, expecting that `eventListener` will be called fewer times than `configureDatadogInstrumentation`, meanwhile keeping the opportunity to set custom eventListener for user

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/tests/NetworkInstrumentationComparisonTest.kt
satween: This test tests that different instrumentation produces similar spans for requests. Technically speaking they are not 100% same (because `spanId` and `traceId` are unique), but spans structure (parent-child) structure, some http attributes (status codes, url, etc) should be the same. So it's called "similar". Let me know if you think this is confusing

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRequestContext.kt
```diff
@@ -25,7 +25,6 @@ internal class CronetRequestContext internal constructor(
 ) {
     internal val method: String
         get() = requestParams.method
-
```
kikoveiga, line 28: Is this on purpose?

integrations/dd-sdk-android-cronet/src/test/kotlin/com/datadog/android/cronet/CronetIntegrationPluginTest.kt
```diff
@@ -167,6 +167,7 @@ internal class CronetIntegrationPluginTest {
         assertThat(engine.distributedTracingInstrumentation).isNull()
     }
 
+    @OptIn(ExperimentalRumApi::class)
```
kikoveiga: Is this needed? The `CronetIntegrationPluginTest` class already has this annotation.

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/CronetRequestContext.kt
satween: nope, some autoformatting 🎰

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RumInstrumentationOkHttpAdapter.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val registry: RequestTracingStateRegistry
+) : Interceptor {
+    private val internalLogger: InternalLogger
+        get() = (rumNetworkInstrumentation.sdkCoreReference.get() as? FeatureSdkCore)?.internalLogger
+            ?: InternalLogger.UNBOUND
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val call = chain.call()
+        val requestInfo = registry.get(call)?.requestInfo
+        val okHttpRequest = (requestInfo as? OkHttpRequestInfo)?.originalRequest
```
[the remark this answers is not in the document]
satween: This is fixed. Keeping custom interceptor modifications now + few tests added for the basic scenarios

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+                    .toSet()
+            )
+        }
+    }
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val call = chain.call()
+        val request = chain.request()
+        val tracingState = registry.update(call) { _, state ->
+            apmNetworkInstrumentation.onRequest(state.requestInfo)
```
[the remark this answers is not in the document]
satween: Fixed, registry now used only for keeping initial UUID value

satween:
@codex review

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,88 @@
...
+import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
+
+internal class RequestTracingStateRegistry(
+    private val internalLogger: InternalLogger
+) {
+    private val requestTracingStateByCall = ConcurrentHashMap<Call, RequestTracingState>()
+
+    fun register(call: Call) = synchronized(call) {
+        if (requestTracingStateByCall.size >= MAX_TRACKED_CALLS) {
```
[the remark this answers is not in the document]
satween: Yes, this could happened, but I don't think it's worth to synchronize again just for this check. The goal is to limit registry to some level, protecting from the OOM in general, it's okay if due to concurrency it will grow bit more than `MAX_TRACKED_CALLS`

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorNonDdTracerNotSendingSpanTest.kt
```diff
@@ -1286,21 +1285,25 @@ internal open class TracingInterceptorNonDdTracerNotSendingSpanTest {
...
                     }
                 }
             }
         } else {
-            fakeMethod = forge.anElementFrom("GET", "HEAD", "DELETE", "CONNECT", "TRACE", "OPTIONS")
+            fakeMethod = forge.anElementFrom(
+                HttpSpec.Method
+                    .values()
+                    .filterNot(HttpSpec.Method::isMethodWithBody)
+            )
```
kikoveiga, line 1249: Maybe format this in a single line, like it's done in the file below `TracingInterceptorNonDdTracerTest`.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/OkHttpRequestInfo.kt
```diff
@@ -99,22 +102,28 @@ internal class OkHttpRequestInfo(internal val request: Request) :
 @Suppress("UnsafeThirdPartyFunctionCall") // OkHttp builder methods are safe
 class OkHttpRequestInfoBuilder(private val requestBuilder: Request.Builder) : HttpRequestInfoBuilder {
 
+    /** @inheritdoc */
```
kikoveiga: We use `inheritdoc` annotation all over the code. I noticed that with it, actually there is no documentation from the parent when hovering over the function, so this does exactly the opposite we wanted it to do😄.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/OkHttpResponseInfo.kt
kikoveiga, line 1: Can we rename this file to `OkHttpResponseInfo`? So it's coherent with `OkHttpRequestInfo`.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/OkHttpResponseInfo.kt
```diff
@@ -12,29 +12,30 @@ import okhttp3.ResponseBody
...
+    override val statusCode: Int get() = originalResponse.code
 
-    override val url: String get() = response.request.url.toString()
+    override val url: String get() = originalResponse.request.url.toString()
 
-    override val headers: Map<String, List<String>> get() = response.headers.toMultimap()
+    override val headers: Map<String, List<String>> get() = originalResponse.headers.toMultimap()
 
     override val contentLength: Long?
         get() = try {
```
kikoveiga, line 34: Should we put the `@WorkerThread` annotation on top of this `get()`? Because `peekBody` is an IO operation.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/DatadogTracingToolkitTest.kt
```diff
@@ -32,6 +34,9 @@ internal class DatadogTracingToolkitTest {
 
     private lateinit var fakeTracedHosts: Map<String, Set<TracingHeaderType>>
 
+    @Mock
+    lateinit var mockInternalLogger: InternalLogger
```
kikoveiga: Is this used anywhere?

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt
```diff
@@ -68,61 +73,82 @@ class CronetIntegrationPlugin internal constructor(
...
+        ): ApmNetworkInstrumentation? = if (!isHeaderPropagationOnly() || !rumInstrumentationExist) {
+            DatadogTracingToolkit.createApmNetworkInstrumentation(
+                CRONET_NETWORK_INSTRUMENTATION_NAME,
+                this
+            )
+        } else {
+            null
+        }
+
         // Exception thrown only for wrong arguments, but those ones are correct
```
kikoveiga:
```suggestion
        // Exception thrown only for wrong arguments, but those are correct
```

Android Studio suggestion.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,191 @@
...
+        private fun RumNetworkInstrumentationConfiguration.createRumInstrumentation(): RumNetworkInstrumentation =
+            _RumInternalProxy.createRumNetworkInstrumentation(
+                OKHTTP_NETWORK_INSTRUMENTATION_NAME,
+                LibraryType.OKHTTP,
+                this
+            )
+
+        private fun ApmNetworkInstrumentationConfiguration.createApmInstrumentation(
+            runInstrumentationExist: Boolean
+        ) = if (!isHeaderPropagationOnly() || !runInstrumentationExist) {
```
kikoveiga:
```suggestion
        ) = if (!isHeaderPropagationOnly() || !rumInstrumentationExists) {
```

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
```diff
@@ -97,30 +97,12 @@ class SampleApplication : Application() {
...
-                            "accept-ranges",
-                            "cache-control",
-                            "content-disposition",
-                            "server",
-                            "user-agent",
-                            "via",
-                            "x-cache-hits",
-                            "x-served-by"
-                        )
-                        .build()
```
kikoveiga, line 115: Let's not forget to re-add this later, when we have common Resource Headers Instrumentation.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapter.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private val sdkCore: FeatureSdkCore?
+        get() = apmNetworkInstrumentation.sdkCore as? FeatureSdkCore
+
+    private val internalLogger: InternalLogger?
+        get() = sdkCore?.internalLogger
+
+    init {
+        // update meta for the configuration telemetry reporting, can be done directly from this thread
+        sdkCore?.updateFeatureContext(Feature.TRACING_FEATURE_NAME, useContextThread = false) {
```
kikoveiga: Should we somehow wait or retry until `sdkCore` is available? For example, on `TracingInterceptor`, there is this `onSdkInstanceReady` logic.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -66,6 +66,14 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
         }
     )
 
+    /**
+     * Creates a configuration with a list of traced hosts using default header types
+     * ([TracingHeaderType.DATADOG] and [TracingHeaderType.TRACECONTEXT]).
+     *
+     * @param tracedHosts a vararg list of hosts to trace.
+     */
+    constructor(vararg tracedHosts: String) : this(tracedHosts.toList())
```
0xnm: shouldn't we just keep either of the two? `vararg` and `List` are very close by.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/tests/CustomInterceptorPreservationTest.kt
```diff
@@ -0,0 +1,418 @@
...
+/**
+ * Integration test verifying that custom interceptor modifications are preserved
+ * when using [withDatadogInstrumentation].
+ *
+ * Tests cover:
+ * - Custom app and network interceptor headers arrive at the server
+ * - RUM resource timing data is present when an upstream interceptor rewrites the URL
+ */
+@OptIn(ExperimentalTraceApi::class, ExperimentalRumApi::class)
+@RunWith(JUnit4::class)
```
0xnm:
```suggestion
@RunWith(AndroidJUnit4::class)
```

it should be Android-specific runner here

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/tests/CustomInterceptorPreservationTest.kt
```diff
@@ -0,0 +1,418 @@
...
+            .get()
+            .build()
+
+        // When
+        val response = client.newCall(request).execute()
+
+        // Then
+        assertThat(response.code).isEqualTo(HttpSpec.StatusCode.OK)
+        response.body?.parseHeaders()?.let { echoedHeaders ->
+            assertThat(echoedHeaders).containsKey(APP_INTERCEPTOR_HEADER.lowercase())
```
0xnm: at which point header name is getting converted to the lowercase?

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/tests/CustomInterceptorPreservationTest.kt
```diff
@@ -0,0 +1,418 @@
...
+ * Integration test verifying that custom interceptor modifications are preserved
+ * when using [withDatadogInstrumentation].
+ *
+ * Tests cover:
+ * - Custom app and network interceptor headers arrive at the server
+ * - RUM resource timing data is present when an upstream interceptor rewrites the URL
+ */
+@OptIn(ExperimentalTraceApi::class, ExperimentalRumApi::class)
+@RunWith(JUnit4::class)
+internal class CustomInterceptorPreservationTest {
```
0xnm: do we need to have this test as instrumented test? it seems that we can have it running on JVM by placing in `reliability -> single-fit -> okhttp`, it will be faster to execute it

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,191 @@
...
+                }
+
+                1 -> {
+                    // NoSuchElementException/IllegalArgumentException cannot happen here
+                    @Suppress("UnsafeThirdPartyFunctionCall")
+                    innerFactories.single()
+                }
+
+                else -> {
+                    CompositeEventListener.Factory(delegates = innerFactories)
```
0xnm, line 144: why we cannot always use `CompositeEventListener.Factory`? Just to avoid having wrapper when it is not really needed?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapter.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Suppress("ReturnCount")
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val call = chain.call()
+        val request = chain.request()
+        // Request might be changed by customer's upstream interceptor(s)
+        val taggedOkHttpRequest = registry.restoreUUIDTag(call, request)
+        if (taggedOkHttpRequest == null) {
+            apmNetworkInstrumentation.reportInstrumentationError { "OkHttp request is missed" }
```
0xnm: is there a risk of spamming too much?

```suggestion
            apmNetworkInstrumentation.reportInstrumentationError { "OkHttp request wasn't instrumented" }
```

nit: log message doesn't say anything about request, so it is not even possible to understand which one

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,88 @@
...
+import com.datadog.android.trace.internal.net.RequestTracingState
+import okhttp3.Call
+import okhttp3.Request
+import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
+
+internal class RequestTracingStateRegistry(
+    private val internalLogger: InternalLogger
+) {
+    private val requestTracingStateByCall = ConcurrentHashMap<Call, RequestTracingState>()
```
0xnm, line 41: It seems that all calls to `requestTracingStateByCall` are made inside `synchronized` blocks. In this case there is no need to have `ConcurrentHashMap`, because `synchronized` block guarantees thread-safety for the all properties used inside.

We need to use either of the two: `ConcurrentHashMap` or `synchronized`

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,88 @@
...
+        call: Call,
+        mapper: (RequestTracingState) -> RequestTracingState
+    ): RequestTracingState? = synchronized(call) {
+        get(call)
+            ?.let(mapper)
+            ?.also { requestTracingStateByCall[call] = it }
+    }
+
+    internal companion object {
+        internal const val MAX_TRACKED_CALLS = 256
```
0xnm, line 106: is it to prevent memory leak?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RumInstrumentationTimingsCounter.kt
```diff
@@ -0,0 +1,217 @@
...
+
+    private var headersStart = 0L
+    private var headersEnd = 0L
+
+    private var bodyStart = 0L
+    private var bodyEnd = 0L
+
+    // region EventListener
+
+    /** @inheritdoc */
```
0xnm: nit: this class is internal, so `/** @inheritdoc */` are not really needed here

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
satween: This is not valid, because we synchronizing by `Call` object, and they could be different for the different simultaneous requests

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
satween: yes

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
satween: As far as I understand this will be inside your branch with header interception support

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapter.kt
satween: Good catch, fixed

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
satween: removed

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
satween: yes

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/ApmInstrumentationOkHttpAdapter.kt
satween: Agree, but logging request (event url) might be not safe

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/OkHttpRequestInfo.kt
satween: this made by ai, removed

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorNonDdTracerNotSendingSpanTest.kt
satween: We can't, as detekt will raise an error

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/network/tests/CustomInterceptorPreservationTest.kt
satween: this is not actual anymore.

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
kikoveiga: Yes, added it there.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumNetworkInstrumentation.kt
```diff
@@ -229,6 +234,7 @@ class RumNetworkInstrumentation internal constructor(
                 else -> RumResourceKind.fromMimeType(mimeType)
             }
 
+        @WorkerThread
```
kikoveiga, line 237: Why is this needed here?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,191 @@
...
+            DatadogTracingToolkit.createApmNetworkInstrumentation(
+                OKHTTP_NETWORK_INSTRUMENTATION_NAME,
+                this
+            )
+        } else {
+            null
+        }
+
+        private fun ApmNetworkInstrumentationConfiguration.createDistributedTracingInstrumentation(
+            runInstrumentationExist: Boolean
```
kikoveiga: Sorry only noticed this same thing in the next function now:
```suggestion
            rumInstrumentationExists: Boolean
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,88 @@
...
+        )
+    }
+
+    fun get(call: Call): RequestTracingState? = synchronized(call) {
+        requestTracingStateByCall[call]
+    }
+
+    fun remove(call: Call): RequestTracingState? = synchronized(call) {
+        requestTracingStateByCall.remove(call)
+    }
```
kikoveiga: Are the `synchronized` blocks needed here? Because we have a `ConcurrentHashMap`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/instrumentation/network/HttpResponseInfo.kt
```diff
@@ -34,5 +36,6 @@ interface HttpResponseInfo {
      *
      * @return the content length as a Long if available, or null if it cannot be determined.
      */
+    @get:WorkerThread
```
kikoveiga, line 39: Why have this annotation in the interface and not in the implementation? From what I see, `CronetHttResponseInfo`'s implementation does no I/O or blocking work.

satween: because you should be aware from the calling side that at least one implementation of this method migh be required to be executed on the bg thread

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/RumNetworkInstrumentation.kt
satween: because it eventually call contentLenght which need to be executed on the bg thread.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
satween: yes. We adding `synchronize` here in order to make all changes related to the `call` sequentially.
On the other hand we need `ConcurrentHashMap` as there could be multiple `call` objects for the different requests done in parallel. So we need not only update everything related to the same `call` sequentially, but also make sure that in case if several requests modifies the map - it will remain consistent.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    internal companion object {
+        internal const val ORIGIN_RUM = "rum"
+        internal const val OKHTTP_NETWORK_INSTRUMENTATION_NAME = "okhttp"
+
+        private fun OkHttpClient.Builder.wrapExistingEventListenerFactory(
+            registry: RequestTracingStateRegistry,
+            rumNetworkInstrumentation: RumNetworkInstrumentation?
+        ) {
+            val userEventListenerFactory: EventListener.Factory = build().eventListenerFactory
```
0xnm, line 123: This means that we will build `OkHttpClient` twice? `OkHttpClient.Builder.build()` here and then somewhere else as well? Won't it leak resources with such throwaway `OkHttpClient`?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,191 @@
...
+) {
+    private var builtClient: OkHttpClient? = null
+
+    /**
+     * Builds the [OkHttpClient] with Datadog instrumentation applied.
+     * @return an instrumented [OkHttpClient] instance.
+     */
+    @Suppress("ReturnCount")
+    fun build(): OkHttpClient {
+        builtClient?.let { return it }
```
0xnm, line 69: nit: there is a race condition accessing `builtClient` from multiple threads. But I guess we expect `build` to be called from the same thread always and only once?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/CompositeEventListener.kt
```diff
@@ -0,0 +1,163 @@
...
+    ) : EventListener.Factory {
+
+        override fun create(call: Call): EventListener {
+            val listeners = delegates.map { it.create(call) }
+            return CompositeEventListener(listeners)
+        }
+    }
+
+    override fun callStart(call: Call) {
+        delegates.forEach { it.callStart(call) }
```
0xnm, line 38: question: should we wrap calls for such delegates in `try-catch`? if some delegate fails to do `callXXX`, should we stop or continue execution for other delegates?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RumInstrumentationTimingsCounter.kt
```diff
@@ -0,0 +1,204 @@
...
+        requestTracingStateRegistry.get(call)?.let(::sendWaitForResourceTimingEvent)
+        headersStart = sdkCore.time.deviceTimeNs
+    }
+
+    override fun responseHeadersEnd(call: Call, response: Response) {
+        super.responseHeadersEnd(call, response)
+        headersEnd = sdkCore.time.deviceTimeNs
+        if (response.code >= HttpURLConnection.HTTP_BAD_REQUEST) {
+            requestTracingStateRegistry.get(call)?.let(::sendTiming)
+        }
```
0xnm, line 104: We had it added in https://github.com/DataDog/dd-sdk-android/pull/285, maybe we need to check if it holds the true again, because we send the timing in `callEnd` as well, so I'm curious if we don't have double reporting.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RequestTracingStateRegistry.kt
```diff
@@ -0,0 +1,119 @@
...
+        mapper: (RequestTracingState) -> RequestTracingState
+    ): RequestTracingState? = synchronized(lockFor(call)) {
+        requestTracingStateByCall[call]
+            ?.let(mapper)
+            ?.also { requestTracingStateByCall[call] = it }
+    }
+
+    internal companion object {
+        internal const val MAX_TRACKED_CALLS = 256
+        private const val MAX_LOCKS_BUCKETS_SIZE = 64
```
0xnm, line 107: is it expected that `MAX_LOCKS_BUCKETS_SIZE` differs from `MAX_TRACKED_CALLS`?

satween: yes, the amount of the requests could be bigger than amount of locks for those requests. This is a compromise between one lock for all requests (aka `@Synchronized`) and map `Call` -> `Lock` that could be bigger.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/CompositeEventListener.kt
satween: good, idea, but I think we actually shouldn't. 
The point here is that we are using `CompositeEventListener` as a wrapper on top of the customer's `EventListener`. And if there is an exception thrown there  - we should not shallow it.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
satween: Yep, this is the trade-off we have to pay for the convenience of the public API we provide for the user. 
The alternatives is less stable imo, as there is only two options: 
1. Reflection-based approach
2. Overriding the customer's event listener with a separate method in the `OkHttpIntegrationPlugin`

I believe this approach is way less impactful.


Regarding the resources - as builded client will be never used - I guess we are safe

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
satween: No, this is the correct warning. Let's add the `@Synchronized  `

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/CompositeEventListener.kt
0xnm: The main point here is not about shallowing, but letting other listeners to be executed and throwing aggregate exception (if there are multiple ones) in the end of iteration cycle only. But not a blocker, just an idea.

satween: This will complicate the code much more without clear understanding for the outcome. I think we should be only sure that our `RumInstrumentationTimingsCounter` will not throw any

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/RumInstrumentationTimingsCounter.kt
satween: Yes, that might be a problem. But let's address this in a separate PR because I want to dogfood this implementation against the previous one and I need same behavior to be sure that the refactoring is correct.

reliability/single-fit/okhttp/src/test/kotlin/com/datadog/android/okhttp/CustomInterceptorPreservationTest.kt
```diff
@@ -0,0 +1,337 @@
...
+        apmInstrumentationConfiguration = null
+    )
+
+    private fun OkHttpClient.Builder.withUrlReplacementInterceptor(url: String) = addInterceptor { chain ->
+        chain.proceed(chain.request().newBuilder().url(url).build())
+    }
+
+    private fun OkHttpClient.Builder.withMethodUrlAndBodyReplacementInterceptor(
+        url: String,
+        body: okhttp3.RequestBody
```
0xnm, line 297: very nit

```suggestion
        body: RequestBody
```

0xnm (approved):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/OkHttpIntegrationPlugin.kt
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @param apmInstrumentationConfiguration optional APM tracing configuration. When provided, trace spans
+ * will be created for HTTP requests and tracing headers will be injected for first-party hosts.
+ * @param rumInstrumentationConfiguration optional RUM configuration. When provided, HTTP requests
+ * will be automatically tracked as RUM resources with timing information.
+ */
+@ExperimentalRumApi
+@ExperimentalTraceApi
+@Suppress("PackageNameVisibility")
+internal fun OkHttpClient.Builder.configureDatadogInstrumentation(
```
hamorillo, line 43: ⚠️ Sample App is using this method but now it's not visible externally, so SampleApp doesn't compile.

```
e: file:...dd-sdk-android/sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt:104:10 Cannot access 'configureDatadogInstrumentation': it is internal in 'com.datadog.android.okhttp'
```

PR #3187, opened 2026-02-18
Fix missing error tracking in resolve() API

typotter:
### What does this PR do?

Adds missing `trackErrorResolution()` call in the `resolve(flagKey, defaultValue)` method's error branch, ensuring error evaluations are properly tracked in telemetry.

### Motivation

The `resolve()` method that returns `ResolutionDetails<T>` was not calling `trackErrorResolution()` for error cases (FLAG_NOT_FOUND, TYPE_MISMATCH, PARSE_ERROR, PROVIDER_NOT_READY). This caused:
- Missing/type-mismatched flags to be excluded from evaluation telemetry
- Underreported error rates
- Skewed aggregated evaluation metrics

The convenience methods (`resolveBooleanValue`, `resolveStringValue`, etc.) correctly track errors via `resolveTracked()`, but the detailed `resolve()` API took a different code path that bypassed error tracking.

### Additional Notes

This is a one-line fix that adds the `trackErrorResolution(resolution)` call to match the behavior of the convenience methods. All existing tests pass.

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
@@ -169,6 +169,7 @@ internal class DatadogFlagsClient(
                 createSuccessResolution(resolution.flag, resolution.value)
             }
             is InternalResolution.Error -> {
+                trackErrorResolution(resolution)
                 createErrorResolution(
                     flagKey = flagKey,
                     defaultValue = resolution.defaultValue,
--- a/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
+++ b/features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
@@ -976,6 +976,100 @@ internal class DatadogFlagsClientTest {
     }
 
     @Test
+    fun `M track error evaluation W resolve() { flag not found and trackEvaluations enabled }`(forge: Forge) {
+        // Given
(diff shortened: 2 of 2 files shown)


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -169,6 +169,7 @@ internal class DatadogFlagsClient(
                 createSuccessResolution(resolution.flag, resolution.value)
             }
             is InternalResolution.Error -> {
+                trackErrorResolution(resolution)
```
0xnm, line 172: it is worth to add a test for this

features/dd-sdk-android-flags/src/test/kotlin/com/datadog/android/flags/internal/DatadogFlagsClientTest.kt
```diff
@@ -975,6 +975,100 @@ internal class DatadogFlagsClientTest {
...
+            context = eq(fakeContext),
+            variantKey = isNull(),
+            allocationKey = isNull(),
+            errorCode = eq(ErrorCode.FLAG_NOT_FOUND.name),
+            errorMessage = eq("Flag not found")
+        )
+    }
+
+    @Test
+    fun `M track error evaluation W resolve() { type mismatch and trackEvaluations enabled }`(forge: Forge) {
```
0xnm: It seems we don't need this test, the branch covered for the assertion is the same as above (the only difference is that for `ErrorCode.TYPE_MISMATCH` we log a warning as well, but it is not checked in this test).

Instead we need to have a test for the evaluation tracking in case of `InternalResolution.Success`, it seems such test is missing if I'm not wrong.

0xnm (approved):

PR #3189, opened 2026-02-19
refactor(flags): use time-based periodic flush instead of cancel+reschedule

typotter:
### What does this PR do?

Replaces the cancel and reschedule pattern for periodic evaluation flushes with a more efficient time-tracking approach. The scheduled task now checks if enough time has elapsed since the last flush before executing.

### Motivation

The previous implementation would cancel and reschedule the periodic flush task every time a flush occurred (manual or automatic). This created unnecessary overhead from repeatedly creating and cancelling scheduled tasks. The new approach uses a single scheduled task that checks a `lastFlushTimeMs` timestamp to determine if a flush is needed.

### Additional Notes

**Trade-offs:**
This configuration guarantees that the age of the in-memory cache will be between `flushIntervalMs` and `2 * flushIntervalMs` in exchange for much simpler handling logic.

**Key changes:**
- Added `lastFlushTimeMs` volatile field to track when events were last written
- `periodicFlushTask()` checks `now - lastFlushTimeMs >= flushIntervalMs` before calling `flush()`
- Made `startPeriodicFlush()` private and idempotent (only starts if not already running)
- Removed unnecessary `reschedulePeriodicFlush()` call from `EvaluationsFeature`
- Removed unnecessary `periodicFlushEnabled` property

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
@@ -13,7 +13,6 @@ import com.datadog.android.flags.model.EvaluationContext
 import com.datadog.android.internal.time.TimeProvider
 import java.util.concurrent.RejectedExecutionException
 import java.util.concurrent.ScheduledExecutorService
-import java.util.concurrent.ScheduledFuture
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantLock
 import kotlin.concurrent.withLock
@@ -31,12 +30,15 @@ internal class EvaluationEventsProcessor(
     periodicFlushEnabled: Boolean = true
 ) {
     private val flushMutex = ReentrantLock()
-    private var scheduledFlushFuture: ScheduledFuture<*>? = null
+
+    @Volatile
+    private var periodicFlushScheduled: Boolean = false
 
     @Volatile
     private var lastFlushTimeMs: Long = 0L
 
     init {
+        lastFlushTimeMs = timeProvider.getDeviceTimestampMillis()
         if (periodicFlushEnabled) {
             startPeriodicFlush()
         }
@@ -91,16 +93,16 @@ internal class EvaluationEventsProcessor(
 
     private fun startPeriodicFlush() {
         flushMutex.withLock {
-            val currentFuture = scheduledFlushFuture
-            if (currentFuture == null || currentFuture.isCancelled || currentFuture.isDone) {
+            if (!periodicFlushScheduled) {
                 try {
                     @Suppress("UnsafeThirdPartyFunctionCall") // exception caught below
-                    scheduledFlushFuture = scheduledExecutor.scheduleWithFixedDelay(
+                    scheduledExecutor.scheduleWithFixedDelay(
                         { periodicFlushTask() },
                         flushIntervalMs,
                         flushIntervalMs,
                         TimeUnit.MILLISECONDS
                     )
+                    periodicFlushScheduled = true
                 } catch (e: RejectedExecutionException) {
                     internalLogger.log(
                         InternalLogger.Level.WARN,
@@ -126,7 +128,6 @@ internal class EvaluationEventsProcessor(
 
         // Wait for any in-progress flush to complete, then drain any remaining events.
         val events = flushMutex.withLock {
-            scheduledFlushFuture?.cancel(false)
             lastFlushTimeMs = timeProvider.getDeviceTimestampMillis()
             aggregator.drain()
         }
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
@@ -139,7 +139,6 @@ internal class EvaluationEventsProcessor(
         scheduledExecutor.shutdown()
 
         // Wait for any in-progress flush to complete, then drain any remaining events.
-        @Suppress("UnsafeThirdPartyFunctionCall") // safe - ReentrantLock.lock() does not throw
         val events = flushMutex.withLock {
             scheduledFlushFuture?.cancel(false)
             aggregator.drain()
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
@@ -127,8 +127,8 @@ internal class EvaluationEventsProcessor(
         // Wait for any in-progress flush to complete, then drain any remaining events.
         val events = flushMutex.withLock {
             scheduledFlushFuture?.cancel(false)
-            aggregator.drain()
             lastFlushTimeMs = timeProvider.getDeviceTimestampMillis()
+            aggregator.drain()
         }
 
         if (events.isNotEmpty()) {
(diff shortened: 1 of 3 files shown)


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -34,16 +34,12 @@ internal class EvaluationEventsProcessor(
...
 
     @Volatile
-    private var periodicFlushEnabled: Boolean = periodicFlushEnabled
-        set(value) {
-            field = value
-            if (value) {
-                reschedulePeriodicFlush()
-            }
-        }
+    private var lastFlushTimeMs: Long = 0L
```
0xnm: nit: probably it is worth to start not from 0, but from the class creation time, so `System.currentTimeMillis`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -93,43 +89,45 @@ internal class EvaluationEventsProcessor(
...
+    }
 
+    fun stop() {
         @Suppress("UnsafeThirdPartyFunctionCall") // safe - does not throw in Android
         scheduledExecutor.shutdown()
 
         // Wait for any in-progress flush to complete, then drain any remaining events.
-        @Suppress("UnsafeThirdPartyFunctionCall") // safe - ReentrantLock.lock() does not throw
         val events = flushMutex.withLock {
             scheduledFlushFuture?.cancel(false)
```
0xnm: There is no need to cancel it, because the executor is shut down above anyway. This will remove the need to store the reference to `scheduledFlushFuture` at all (probably). Runnable passed in the `scheduleWithFixedDelay` can check if it should schedule itself once again after `periodicFlushTask` is executed and do it if `periodicFlushEnabled` is enabled.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -93,43 +89,45 @@ internal class EvaluationEventsProcessor(
...
-                    e
-                )
+            val currentFuture = scheduledFlushFuture
+            if (currentFuture == null || currentFuture.isCancelled || currentFuture.isDone) {
+                try {
+                    @Suppress("UnsafeThirdPartyFunctionCall") // exception caught below
+                    scheduledFlushFuture = scheduledExecutor.scheduleWithFixedDelay(
+                        { periodicFlushTask() },
+                        flushIntervalMs,
+                        flushIntervalMs,
```
0xnm: Should it be the same value? The first one is for how long to wait until the task get executed, the second is how long to wait until the next one should be scheduled for execution. But we always have only single task anyway if I'm not wrong.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: added to `init` since we're using `TimeProvider`, we should be using it for the initial timestamp if not 0.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: Good question. I don't think it makes sense to immediately start a flush since this is is called on startup, so waiting 1 interval makes sense there. The executor waits for `delay`ms _after_ the task executes, so it is automatically rescheduled and as such, there would indeed only ever be one task scheduled at a time.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: Great simplification. The `shutdown` will allow the running task to complete and will also prevent the `scheduleExecutor` from re-scheduling the task so we don't need to even manage the rescheduling.

typotter (commented):
Thanks. ptal

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
0xnm: this can be even moved out of `init`

```suggestion
    private var lastFlushTimeMs: Long = timeProvider.getDeviceTimestampMillis()
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -31,19 +30,18 @@ internal class EvaluationEventsProcessor(
...
     @Volatile
-    private var periodicFlushEnabled: Boolean = periodicFlushEnabled
-        set(value) {
-            field = value
-            if (value) {
-                reschedulePeriodicFlush()
-            }
-        }
+    private var periodicFlushScheduled: Boolean = false
+
```
0xnm:
```suggestion
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -31,19 +30,18 @@ internal class EvaluationEventsProcessor(
     periodicFlushEnabled: Boolean = true
```
0xnm, line 30:
```suggestion
    @Volatile
    private var periodicFlushEnabled: Boolean = false
```

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
```diff
@@ -93,43 +91,44 @@ internal class EvaluationEventsProcessor(
...
 
-    fun reschedulePeriodicFlush() {
-        if (!periodicFlushEnabled) {
-            return
-        }
-
-        // Cancel any scheduled before scheduling a new one.
-        @Suppress("UnsafeThirdPartyFunctionCall") // safe - ReentrantLock.lock() does not throw
+    private fun startPeriodicFlush() {
         flushMutex.withLock {
```
0xnm: I think `flushMutex` is not needed here, because what this function does is only scheduling, it doesn't do the actual flush.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: We aren't storing this value - just read it once at construction and start the period flush using a mechanism that reschedules the task after each completion.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: This was extra-defensive since the class, as written, will only ever call `startPeriodicFlush` once from `init`.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt
typotter: right. now that we don't need to protect the `scheduledFuture` with the lock we can drop it here.

0xnm (approved):

PR #3190, opened 2026-02-19
RUM-14330: Optimizations to reduce .so file

kikoveiga:
### What does this PR do?

- Adds compiler and linker size optimizations to the NDK module's CMake configuration, reducing the `.so` file size by around 50%.
- Minor code cleanups (remove unused imports, CLion suggestions)

### Motivation

The `.so` for arm64-v8a was ~860 KB — larger than the SDK's APK footprint (~570 KB) without the `ndk`.

Below are analysis of this file inside the Sample app release APK done in Android Studio, describing `Size` on the left and `Download Size` on the right.

Before (sample app release APK on develop branch):
image

After (with these optimizations):
image

### DIFF
--- a/sample/kotlin/src/main/cpp/datadog-native-sample-lib.cpp
+++ b/sample/kotlin/src/main/cpp/datadog-native-sample-lib.cpp
@@ -2,7 +2,8 @@
 #include <signal.h>
 
 void crash_with_sigsegv_signal() {
-    int *pointer = nullptr;
+    // volatile prevents the compiler from optimizing away the null dereference in release builds
+    volatile int *pointer = nullptr;
 #pragma clang diagnostic push
 #pragma ide diagnostic ignored "NullDereferences"
     int t = *pointer;
@@ -16,12 +17,9 @@ void crash_with_sigabrt_signal() {
 }
 #pragma clang diagnostic pop
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wreturn-type"
-int crash_with_sigill_signal() {
-    // do not return anything to simulate the SIGILL
+void crash_with_sigill_signal() {
+    __builtin_trap();
 }
-#pragma clang diagnostic pop
 
 extern "C" JNIEXPORT void JNICALL
 Java_com_datadog_android_sample_crash_CrashFragment_simulateNdkCrash(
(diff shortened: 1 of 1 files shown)


0xnm:
@codex review

sample/kotlin/src/main/cpp/datadog-native-sample-lib.cpp
kikoveiga, line 1: @0xnm the compiler was optimizing and removing the code of `sigsegv` and `sigill` ndk crashes (even before my changes) in release builds, so the buttons on the sample app didn't work. Simplified this and confirmed NDK crashes are being sent to Datadog and working fine.

sample/kotlin/src/main/cpp/datadog-native-sample-lib.cpp
```diff
@@ -1,47 +1,11 @@
 #include <jni.h>
 #include <signal.h>
 
-void crash_with_sigsegv_signal() {
```
0xnm, line 4: Can you explain the need for this change? `raise` indeed will raise the necessary signal, but the code here is to illustrate the real example leading to the particular signal.

features/dd-sdk-android-ndk/src/main/cpp/CMakeLists.txt
```diff
@@ -35,3 +35,23 @@ set_target_properties(datadog-native-lib
         PROPERTIES
         COMPILE_OPTIONS
         -Werror -Wall -pedantic)
+
+# Size optimizations for non-Debug builds
+if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
```
0xnm, line 40: one thing to check for release builds: once NDK crash happens and _if_ Datadog NDK makes a part of the stacktrace,  can we still symbolicate such frames after binary shrinking?

sample/kotlin/src/main/cpp/datadog-native-sample-lib.cpp
kikoveiga: Rolled back to the previous code, just added some tweaks to prevent compiler optimization. Tested and crashes are working in sample app release build.

features/dd-sdk-android-ndk/src/main/cpp/CMakeLists.txt
kikoveiga: Thanks for pointing this out. First implementation had this problem, now fixed by removing `fvisibility=hidden` from compile options. Now, all function names are preserved and readable at runtime.

PR #3197, opened 2026-02-23
Fix more concurrency issues

0xnm:
### What does this PR do?

A bit more of concurrency issues fixes. The main change is shifting from `getOrPut` call, which is not atomic.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/slowframes/SlowFramesListener.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/slowframes/SlowFramesListener.kt
@@ -5,6 +5,7 @@
  */
 package com.datadog.android.rum.internal.metric.slowframes
 
+import android.os.Build
 import androidx.metrics.performance.FrameData
 import com.datadog.android.internal.time.TimeProvider
 import com.datadog.android.rum.configuration.SlowFramesConfiguration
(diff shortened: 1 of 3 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/metric/slowframes/SlowFramesListener.kt
```diff
@@ -139,11 +142,20 @@ internal class DefaultSlowFramesListener(
...
+                configuration.maxSlowFramesAmount,
+                minimumViewLifetimeThresholdNs = configuration.minViewLifetimeThresholdNs
+            )
+        }
+
+        return if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M) {
+            @Suppress("UnsafeThirdPartyFunctionCall") // all args are safe
+            slowFramesRecords.computeIfAbsent(viewId) { createLambda.invoke() }
+        } else {
+            slowFramesRecords.getOrPut(viewId, createLambda)
```
aleksandr-gringauz, line 158: In the PR description you say that `getOrPut` is the problem. But it is still here. Is it ok?

0xnm: Yes, it is only for API 23, which is small fraction of the user base. Having a proper atomic approach here would require introducing lock, etc. Not worth it, considering the the low severity and impact of the issue we can have without such atomic guardrails.

aleksandr-gringauz (approved):

PR #3199, opened 2026-02-24
RUM-14471: Auto-forward TTID to next activity for interstitial pattern

hamorillo:
### What does this PR do?

**Note**: This PR is targeting [hector.morilloprieto/RUM-14471](https://github.com/DataDog/dd-sdk-android/tree/hector.morilloprieto/RUM-14471)￼, as it is an iteration of it. It's motivated by [this comment](https://github.com/DataDog/dd-sdk-android/pull/3184#discussion_r2822227426).

Automatically detects and forwards TTID (Time To Initial Display) measurement to the next qualifying activity when an interstitial activity (e.g. a splash screen) calls `startActivity()` + `finish()` without ever calling `setContentView()`. In this pattern the `OnDrawListener` never fires on the interstitial, so TTID is silently dropped. This PR makes the SDK detect that case and subscribe to the next activity's first frame draw instead — with no customer configuration required.

### Motivation

Customers using a splash/auth screen pattern were losing TTID data entirely because the SDK had no way to recover when the first tracked activity never drew a frame.

### How to test

Apply the attached [patch](https://github.com/user-attachments/files/25365831/patch.patch) to the branch:

The patch modifies the sample app to add a `SplashActivity` as the launcher activity. This activity immediately calls `startActivity(NavActivity)` + `finish()` without setting a content view, simulating an interstitial splash screen. It also sets telemetry sampling to 100% and reduces session inactivity timeout to 15 seconds for easier testing.

Run the sample app and check the RUM dashboard — you should see TTID reported for `NavActivity` with `was_forwarded: true` in the startup telemetry metric.

### Additional Notes

- **No time-based heuristics** — forwarding is triggered purely by activity lifecycle events (`onBeforeActivityCreated` while a pending startup scenario exists).
- **First-draw-wins** — if multiple activities are created concurrently, whichever draws first reports TTID; subsequent callbacks are discarded via an identity check on `pendingScenario`.
- **`wasForwarded` telemetry flag** added to `RumTTIDInfo` and reported via the telemetry metric so we can measure how often this path is taken.
- New integration tests cover synchronous interstitial, chained (A→B→C), and async/deferred navigation scenarios.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
@@ -701,21 +701,61 @@ internal class RumFeature(
             listener = object : RumAppStartupDetector.Listener {
                 private val rumFirstDrawTimeReporter = RumFirstDrawTimeReporter.create(sdkCore = sdkCore)
 
-                override fun onAppStartupDetected(scenario: RumStartupScenario) {
-                    val activity = scenario.activity.get() ?: return
-                    val rumMonitor = (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor) ?: return
+                override fun onAppStartupDetected(scenario: RumStartupScenario): Boolean {
+                    val activity = scenario.activity.get()
+                    val rumMonitor = GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor
+                    if (activity == null || rumMonitor == null) return false
 
                     rumMonitor.sendAppStartEvent(scenario)
 
+                    val capturedScenario = scenario
                     val callback = object : RumFirstDrawTimeReporter.Callback {
                         override fun onFirstFrameDrawn(timestampNs: Long) {
+                            // Another activity may have already reported TTID
+                            val pending = rumAppStartupDetector?.getPendingScenario()
+                            if (pending !== capturedScenario) return
+
                             val durationNs = timestampNs - scenario.initialTime.nanoTime
                             val info = RumTTIDInfo(
                                 scenario = scenario,
-                                durationNs = durationNs
+                                durationNs = durationNs,
+                                wasForwarded = false
                             )
 
                             rumMonitor.sendTTIDEvent(info)
+                            rumAppStartupDetector?.clearPendingScenario()
+                        }
+                    }
+
+                    rumFirstDrawTimeReporter.subscribeToFirstFrameDrawn(
+                        activity = activity,
+                        callback = callback
+                    )
+                    return true
+                }
+
+                override fun onNextActivityCreated(
+                    pendingScenario: RumStartupScenario,
+                    activity: Activity
+                ) {
+                    val rumMonitor = (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor) ?: return
+
+                    val capturedScenario = pendingScenario
+                    val callback = object : RumFirstDrawTimeReporter.Callback {
+                        override fun onFirstFrameDrawn(timestampNs: Long) {
+                            // Another activity may have already reported TTID
+                            val pending = rumAppStartupDetector?.getPendingScenario()
+                            if (pending !== capturedScenario) return
+
+                            val durationNs = timestampNs - pendingScenario.initialTime.nanoTime
+                            val info = RumTTIDInfo(
+                                scenario = pendingScenario,
+                                durationNs = durationNs,
+                                wasForwarded = true
+                            )
+
+                            rumMonitor.sendTTIDEvent(info)
+                            rumAppStartupDetector?.clearPendingScenario()
                         }
                     }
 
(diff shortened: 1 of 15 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -83,7 +84,13 @@ internal class RumAppStartupDetectorImpl(
             trackedActivities.add(activity)
         }
 
-        if (trackedActivities.size == 1 && !isChangingConfigurations && shouldTrackStartup) {
+        val isFirstTrackedActivityWithNoPendingStartup =
+            trackedActivities.size == 1 &&
+                !isChangingConfigurations &&
+                shouldTrackStartup &&
+                pendingScenario == null
```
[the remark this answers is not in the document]
hamorillo: Not sure how common this case would be. 

I see two options here:
- Prevent stale pendingScenario from blocking future startups by clearing it in `RumAppStartupDetectorImpl.onBeforeActivityCreated()` when it’s older than `rumSessionScopeStartupManagerImpl.MAX_TTID_DURATION_NS` (let's say 1 minute), so a later “re-launch” in the same process creates a fresh startup scenario instead of forwarding a stale one?

- Accept this case and create some documentation. 

Any other ideas?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -701,21 +701,61 @@ internal class RumFeature(
...
-                    val activity = scenario.activity.get() ?: return
-                    val rumMonitor = (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor) ?: return
+                override fun onAppStartupDetected(scenario: RumStartupScenario): Boolean {
+                    val activity = scenario.activity.get()
+                    val rumMonitor = GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor
+                    if (activity == null || rumMonitor == null) return false
 
                     rumMonitor.sendAppStartEvent(scenario)
 
+                    val capturedScenario = scenario
```
aleksandr-gringauz: capturedScenario variable is redundant

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -701,21 +701,61 @@ internal class RumFeature(
             listener = object : RumAppStartupDetector.Listener {
                 private val rumFirstDrawTimeReporter = RumFirstDrawTimeReporter.create(sdkCore = sdkCore)
 
-                override fun onAppStartupDetected(scenario: RumStartupScenario) {
-                    val activity = scenario.activity.get() ?: return
-                    val rumMonitor = (GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor) ?: return
+                override fun onAppStartupDetected(scenario: RumStartupScenario): Boolean {
+                    val activity = scenario.activity.get()
+                    val rumMonitor = GlobalRumMonitor.get(sdkCore) as? AdvancedRumMonitor
+                    if (activity == null || rumMonitor == null) return false
```
aleksandr-gringauz: It seems a bit of overkill to return a bool here. Yes, according to the types RumMonitor and Activity could be null. But in practice this will not happen, so I suggest we simplify here and make `onAppStartupDetected` return nothing.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -701,21 +701,61 @@ internal class RumFeature(
...
+                            val durationNs = timestampNs - pendingScenario.initialTime.nanoTime
+                            val info = RumTTIDInfo(
+                                scenario = pendingScenario,
+                                durationNs = durationNs,
+                                wasForwarded = true
                             )
 
                             rumMonitor.sendTTIDEvent(info)
+                            rumAppStartupDetector?.clearPendingScenario()
                         }
```
aleksandr-gringauz, line 741: This block of code seems almost identical to the one in `onAppStartupDetected`. Could we extract it into a function?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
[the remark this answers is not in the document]
aleksandr-gringauz: Thought about it for a bit. Your option 1 seems like a right thing to do. We already have this 1 minute threshold https://github.com/DataDog/dd-sdk-android/blob/18b420c3e7f1c8d7304c53ceca7828e67e885f89/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt#L299 for TTID.

aleksandr-gringauz (dismissed):
LGTM!

aleksandr-gringauz (approved):

PR #3202, opened 2026-02-25
PANA-5681: Add `HeatmapIdentifier` infrastructure [1 of 4]

gonzalezreal:
### What does this PR do?

Introduces the foundational shared component for mobile heatmaps: a `HeatmapIdentifierRegistry` in the internal module that Session Replay writes to and RUM reads from. The store holds an atomic snapshot of view identities (keyed by view ID) that SR publishes after each window traversal; RUM's gesture listener will then look up the identifier for a tapped view and, if found, attaches it alongside the view's dimensions and touch position to the action event. The implementation uses a ReentrantReadWriteLock so SR's writes (full snapshot replacement) never block concurrent RUM reads. 

This PR introduces only the data contracts (HeatmapIdentifier, HeatmapIdentifierRegistry, HeatmapIdentifierStore) in dd-sdk-android-internal. The cross-feature wiring will follow in the rum-heatmaps PR: a new HeatmapsFeature : Feature will live in dd-sdk-android-rum, hold a HeatmapIdentifierRegistry, and be registered on SdkCore by RumFeature during initialization — mirroring how SessionReplayFeature registers its sibling ResourcesFeature. Session Replay will then read the registry via sdkCore.getFeature(HEATMAPS_FEATURE_NAME), matching the iOS architecture (where HeatmapIdentifierRegistryFeature is registered on the core and consumed by SR through the same lookup path). The companion factory HeatmapIdentifierRegistry.create() stays in this PR as the constructor seam that lets HeatmapsFeature build a registry without exposing HeatmapIdentifierStore outside dd-sdk-android-internal.

### Motivation

Support mobile heatmaps. This is PR 1 of 4, the foundational infrastructure that RUM and Session Replay will build on.

### Additional Notes

Ported from #3164. Self-contained with no dependency on schema changes.


dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /** Priority 4: Last resort - use root view's class name. */
+    private fun getNamespaceFromRootClassName(rootView: View): String {
+        return "$NAMESPACE_ROOT_CLASS_PREFIX${escapePathComponent(rootView.javaClass.name)}"
+    }
+
+    private fun getViewPathSegment(view: View, parentView: ViewGroup?): String {
+        val resourceName = getResourceName(view)
+        if (resourceName != null) return escapePathComponent(resourceName)
```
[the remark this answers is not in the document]
gonzalezreal: There's no sibling disambiguation when a resource ID is present. The RFC says resource ID is used verbatim. Same-class indexing is only the fallback when there's no resource ID. Two siblings sharing the same android:id producing the same identity is by design per the RFC

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    @Synchronized
+    override fun onWindowRefreshed(root: View) {
+        indexTree(root)
+    }
+
+    @Synchronized
+    override fun resolveViewIdentity(view: View): String? {
+        return viewPathDataCache[view]?.identityHash
```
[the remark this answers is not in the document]
gonzalezreal: There's no requirement to check attachment state. The contract is: indexed -> return identity, not indexed -> return `null`. The cache is cleared on screen namespace change, and re-indexed on each window refresh.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolver.kt
```diff
@@ -0,0 +1,43 @@
...
+ * Resolves globally unique, stable identities for Android Views based on their canonical path
+ * in the view hierarchy. Used for heatmap correlation between RUM actions and Session Replay.
+ */
+@NoOpImplementation(publicNoOpImplementation = true)
+interface ViewIdentityResolver {
+
+    /**
+     * Sets the current screen identifier. Takes precedence over Activity-based detection.
+     * @param identifier the screen identifier (typically RUM view URL), or null to clear
+     */
```
ambushwork: Do we really clear it? if so what's the case?

dd-sdk-android-internal/src/test/java/com/datadog/android/internal/identity/ViewIdentityResolverTest.kt
```diff
@@ -0,0 +1,700 @@
...
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+internal class ViewIdentityResolverTest {
+
+    private lateinit var testedManager: ViewIdentityResolverImpl
```
ambushwork:
```suggestion
    private lateinit var testedViewIdentityResolver: ViewIdentityResolverImpl
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Cache: Root view reference → Screen namespace string.
+     * Example: DecorView@0x1a2b → "view:HomeScreen"
+     *
+     * Avoids recomputing namespace (which may involve walking context chain).
+     * Cleared when screen changes (namespace depends on currentRumViewIdentifier).
+     */
+    @Suppress("UnsafeThirdPartyFunctionCall") // WeakHashMap() is never null
+    private val rootScreenNamespaceCache: MutableMap<View, String> =
```
ambushwork: I feel the name is a bit confusing here, "screen" is a concept of heatmaps which represents the whole visible interface according to the RFC, "rootView" is a Android term meaning the root of the view tree hireacy. I think here is more like "screenNamespaceCache indexed by rootView"

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+            ?: getNamespaceFromRootResourceId(rootView)
+            ?: getNamespaceFromRootClassName(rootView)
+
+        rootScreenNamespaceCache[rootView] = screenNamespace
+        return screenNamespace
+    }
+
+    /** Priority 1: Use RUM view identifier if available (set via RumMonitor.startView). */
+    private fun getNamespaceFromRumView(): String? {
+        return currentRumViewIdentifier.get()?.let { viewName ->
```
ambushwork:
```suggestion
        return currentRumViewIdentifier.get()?.let { viewUrl ->
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+            "$NAMESPACE_VIEW_PREFIX${escapePathComponent(viewName)}"
```
ambushwork:
```suggestion
            "$NAMESPACE_VIEW_PREFIX${escapePathComponent(viewUrl)}"
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+        private const val NAMESPACE_ROOT_CLASS_PREFIX = "root-cls:"
+        private const val LOCAL_KEY_CLASS_PREFIX = "cls:"
+    }
+}
+
+private fun escapePathComponent(input: String): String {
+    return input.replace("%", "%25").replace("/", "%2F")
+}
+
+private fun md5Hex(input: String): String? {
```
ambushwork: we have the exact same logic in `MD5HashGenerator`, maybe it's time to move it into internal and reuse it here.

dd-sdk-android-internal/api/apiSurface
```diff
@@ -60,6 +60,18 @@ class com.datadog.android.internal.data.SharedPreferencesStorage : PreferencesSt
   override fun clear()
 data class com.datadog.android.internal.flags.RumFlagEvaluationMessage
   constructor(String, Any)
+interface com.datadog.android.internal.identity.ViewIdentityResolver
+  fun setCurrentScreen(String?)
+  fun onWindowRefreshed(android.view.View)
+  fun resolveViewIdentity(android.view.View): String?
+  companion object 
+    const val FEATURE_CONTEXT_KEY: String
+class com.datadog.android.internal.identity.ViewIdentityResolverImpl : ViewIdentityResolver
```
0xnm: a small trick we use to not expose `Impl` classes is to create a property (in this case it will be a function `create(arg)`) in the `companion object` of `ViewIdentityResolver`.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolver.kt
```diff
@@ -0,0 +1,43 @@
...
+    fun setCurrentScreen(identifier: String?)
+
+    /**
+     * Indexes a view tree for efficient identity lookups.
+     * @param root the root view of the window
+     */
+    fun onWindowRefreshed(root: View)
+
+    /**
+     * Resolves the stable identity for a view (32 hex chars), or null if the view is detached.
```
0xnm: nit: probably we can omit `32 hex chars` from the description, it is implementation detail

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolver.kt
```diff
@@ -0,0 +1,43 @@
...
+     * @param view the view to identify
+     * @return the stable identity hash, or null if it cannot be computed
+     */
+    fun resolveViewIdentity(view: View): String?
+
+    companion object {
+        /**
+         * Key used to store the ViewIdentityResolver instance in the feature context.
+         */
+        const val FEATURE_CONTEXT_KEY: String = "_dd.view_identity_resolver"
```
0xnm: I think this is quite dangerous. What is the purpose to store such object in the feature context given that we have direct access to the type?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * Cache: Resource ID (Int) → Resource name (String).
+     * Example: 2131230001 → "com.example.app:id/login_button"
+     *
+     * LRU cache with fixed size. Never explicitly cleared - resource ID mappings
+     * are global and don't change based on screen. Avoids repeated calls to
+     * resources.getResourceName() which is expensive.
+     */
+    @Suppress("UnsafeThirdPartyFunctionCall") // LinkedHashMap constructor doesn't throw
+    private val resourceNameCache: MutableMap<Int, String> = Collections.synchronizedMap(
```
0xnm: I don't think `synchronizedMap` is a good choice here, maybe `ConcurrentHashMap` is better, given that after some time the number of reads will be much bigger than number of writes.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+        object : LinkedHashMap<Int, String>(RESOURCE_NAME_CACHE_SIZE, DEFAULT_LOAD_FACTOR, true) {
+            override fun removeEldestEntry(eldest: MutableMap.MutableEntry<Int, String>?): Boolean {
```
0xnm: there is `androidx.cache.LruCache` class though

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Cache: View reference → PathData (canonical path + identity hash).
+     * Example: Button@0x7f3a → PathData("com.app/view:Home/login_button", "a1b2c3...")
+     *
+     * WeakHashMap so entries are removed when Views are garbage collected.
+     * Cleared when screen changes (paths include screen namespace, so become invalid).
+     */
+    @Suppress("UnsafeThirdPartyFunctionCall") // WeakHashMap() is never null
+    private val viewPathDataCache: MutableMap<View, PathData> =
```
0xnm: We shouldn't do this. Views shouldn't wait for the GC to be removed, they should be removed once related screen (activity, fragment, etc.) is destroyed.

Since it is relying on the reference, why not to store class hashcode then?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+        Collections.synchronizedMap(WeakHashMap())
```
0xnm: The same comment as above: `synchronizedMap` is not a good choice.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
0xnm: same as above: caching views itself is not a good idea.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+    @Suppress("UnsafeThirdPartyFunctionCall") // WeakHashMap() is never null
+    private val rootScreenNamespaceCache: MutableMap<View, String> =
+        Collections.synchronizedMap(WeakHashMap())
+
+    /** The current RUM view identifier, set via setCurrentScreen(). */
+    private val currentRumViewIdentifier = AtomicReference<String?>(null)
+
+    @Synchronized
+    override fun setCurrentScreen(identifier: String?) {
+        @Suppress("UnsafeThirdPartyFunctionCall") // type-safe: generics prevent VarHandle type mismatches
```
0xnm: We should declare this call as safe instead, there is already `java.util.concurrent.atomic.AtomicBoolean.getAndSet(kotlin.Boolean)` declared.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+     * Cleared when screen changes (namespace depends on currentRumViewIdentifier).
+     */
+    @Suppress("UnsafeThirdPartyFunctionCall") // WeakHashMap() is never null
+    private val rootScreenNamespaceCache: MutableMap<View, String> =
+        Collections.synchronizedMap(WeakHashMap())
+
+    /** The current RUM view identifier, set via setCurrentScreen(). */
+    private val currentRumViewIdentifier = AtomicReference<String?>(null)
+
+    @Synchronized
```
0xnm: there is no need for the `@Synchronized`, `AtomicReference.getAndSet` already does atomic check guaranteeing the absence of concurency.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+    @Synchronized
+    override fun onWindowRefreshed(root: View) {
+        indexTree(root)
+    }
+
+    @Synchronized
+    override fun resolveViewIdentity(view: View): String? {
```
0xnm: given these will be called on the `main` thread, we should avoid using `Synchronized` and use more granular synchronization primitives

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+        val rootPathSegment = getViewPathSegment(root, null)
+        // Root view (e.g., DecorView) is not interactable, so we don't cache its identity.
+        // We only need its path as the prefix for descendant paths.
+        return "$appIdentifier/$screenNamespace/$rootPathSegment"
+    }
+
+    /** Depth-first traversal of view hierarchy, computing and caching identity for each view. */
+    private fun traverseAndIndexViews(root: View, rootCanonicalPath: String) {
+        // Index the root view (all cache insertions happen here for consistency)
+        md5Hex(rootCanonicalPath)?.let { hash ->
```
0xnm: There was a point about SHA-1 vs MD5 in the RFC, did we expore it?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /** Priority 1: Use RUM view identifier if available (set via RumMonitor.startView). */
+    private fun getNamespaceFromRumView(): String? {
+        return currentRumViewIdentifier.get()?.let { viewName ->
+            "$NAMESPACE_VIEW_PREFIX${escapePathComponent(viewName)}"
+        }
+    }
+
+    /** Priority 2: Fall back to Activity class name if root view has Activity context. */
```
0xnm: nit: probably we need to remove such comments, they don't really add any value

0xnm (changes requested):
I went only through the production source files so far and left some comments.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
jonathanmos: Synchronized is still needed to make the WeakHashMap `clear` safe relative to concurrent map access

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolverImpl.kt
jonathanmos: Yes, MD5 keeps us aligned with existing SDK behavior. This path hash is non‑cryptographic, so SHA doesn’t add meaningful benefit.

0xnm: SHA-1 is not about cryptography. It is just another hash family, the same usage as MD5. As it is pointed in RFC, calculating SHA-1 may have some accelerations/optimizations on CPU. But this is just a speculation.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/identity/ViewIdentityResolver.kt
jonathanmos: You're right - switching to a typed Feature registered on SdkCore so the registry is exposed by type

ambushwork:
@codex review

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifier.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    companion object {
+
+        private const val SEPARATOR = "/"
+
+        internal fun canonicalPath(
+            elementPath: List<String>,
+            screenName: String,
+            bundleIdentifier: String
+        ): String = buildString {
```
0xnm: seems like this method has nothing to do with `HeatmapIdentifier`, is it here just because of the shared semantics?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifier.kt
```diff
@@ -0,0 +1,35 @@
...
+data class HeatmapIdentifier(val rawValue: String) {
+
+    companion object {
+
+        private const val SEPARATOR = "/"
+
+        internal fun canonicalPath(
+            elementPath: List<String>,
+            screenName: String,
+            bundleIdentifier: String
```
0xnm: it is not clear what `bundleIdentifier` is, maybe it is worth to add a doc to this method

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierStore.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.concurrent.locks.ReentrantReadWriteLock
+import kotlin.concurrent.read
+import kotlin.concurrent.write
+
+/**
+ * Internal — construct via [HeatmapIdentifierRegistry.create].
+ *
+ * Writes (one per Session Replay traversal) atomically replace the snapshot; reads
+ * (per RUM tap action) are non-blocking with respect to other reads.
```
0xnm: given that we have different threads: main for SR and RUM thread for RUM, can we have a situation where the read is done for the ID with a new value?

e.g.
 
```
write(IDa - ViewA) -> write(IDa - ViewB)
                    \
                     \___________-> read(IDa) = ViewB (but should be ViewA).
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierStore.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+internal class HeatmapIdentifierStore : HeatmapIdentifierRegistry {
+
+    private val lock = ReentrantReadWriteLock()
+
+    private var identifiers: Map<Long, HeatmapIdentifier> = HashMap()
```
0xnm:
```suggestion
    private val identifiers: Map<Long, HeatmapIdentifier> = mutableMapOf()
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierStore.kt
```diff
@@ -0,0 +1,43 @@
...
+        // HashMap(int) throws IllegalArgumentException only for a negative argument;
+        // putAll() throws NullPointerException only for a null map. Map.size is non-negative
+        // and the parameter is non-null, so neither throws here.
+        @Suppress("UnsafeThirdPartyFunctionCall")
+        val snapshot: Map<Long, HeatmapIdentifier> = HashMap<Long, HeatmapIdentifier>(identifiers.size).apply {
+            putAll(identifiers)
+        }
+        lock.write {
+            this.identifiers = snapshot
+        }
```
0xnm:
```suggestion
        lock.write {
            this.identifiers.apply {
              clear()
              putAll(identifiers)
            }
        }
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierStore.kt
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Internal — construct via [HeatmapIdentifierRegistry.create].
+ *
+ * Writes (one per Session Replay traversal) atomically replace the snapshot; reads
+ * (per RUM tap action) are non-blocking with respect to other reads.
+ */
+internal class HeatmapIdentifierStore : HeatmapIdentifierRegistry {
+
+    private val lock = ReentrantReadWriteLock()
```
0xnm, line 27: Why we cannot simply use `ConcurrentHashMap`?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifier.kt
ambushwork: bundle identifier is the concept for iOS heatmap, here it should be `appPackageName`

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierRegistry.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.datadog.tools.annotation.NoOpImplementation
+
+/**
+ * Stores and retrieves [HeatmapIdentifier]s keyed by view identity.
+ */
+@NoOpImplementation(publicNoOpImplementation = true)
+interface HeatmapIdentifierRegistry {
+
+    /**
+     * Replaces the current set of identifiers with [identifiers].
```
ambushwork: it's nice to specify the argument in the documentation, especially what `Long` should be in the map

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierRegistry.kt
```diff
@@ -0,0 +1,34 @@
...
+     */
+    fun setHeatmapIdentifiers(identifiers: Map<Long, HeatmapIdentifier>)
+
+    /**
+     * Returns the [HeatmapIdentifier] for the view with the given [viewId], or null if unknown.
+     */
+    fun heatmapIdentifier(viewId: Long): HeatmapIdentifier?
```
ambushwork: should it be called `getHeatmapIdentifier`?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifier.kt
jonathanmos: This was an intermediate for generating the identifier values - it's now private and hidden behind the identifier creation method

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierStore.kt
jonathanmos: Good catch. There's now screen name protection so that if the name doesn't match when rum queries the store then we return null

ambushwork (approved):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifier.kt
```diff
@@ -0,0 +1,76 @@
...
+            for (segment in elementPath) {
+                append(SEPARATOR)
+                append(escape(segment))
+            }
+        }
+
+        // % must be escaped before / so the encoding is reversible.
+        @Suppress("UnsafeThirdPartyFunctionCall") // non-null arguments; cannot throw
+        private fun escape(input: String): String =
+            input.replace("%", "%25").replace(SEPARATOR, "%2F")
```
0xnm: should we simply use URL encoder here? it will get a chance to use a standard URL decoder when decoding is needed instead of writing custom decoding logic for just these 2 symbols

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifier.kt
```diff
@@ -0,0 +1,76 @@
...
+        private fun sha256Hex(input: String): String? {
+            return try {
+                val digest = MessageDigest.getInstance("SHA-256")
+                // digest(ByteArray) does not throw: the input is non-null (toByteArray always
+                // returns a non-null array) and DigestException is only declared on the
+                // offset/length overload, not this one.
+                @Suppress("UnsafeThirdPartyFunctionCall")
+                digest.digest(input.toByteArray(Charsets.UTF_8)).toHexString()
+            } catch (_: NoSuchAlgorithmException) {
+                null
```
0xnm, line 75: maybe we should log it to telemetry? if SHA-256 missing is common, we then need to come up with something else

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierStore.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    private var snapshotScreenName: String? = null
+    private val identifiers: MutableMap<Long, HeatmapIdentifier> = mutableMapOf()
+
+    override fun setHeatmapIdentifiers(identifiers: Map<Long, HeatmapIdentifier>, screenName: String) {
+        lock.write {
+            this.snapshotScreenName = screenName
+            this.identifiers.clear()
+            // putAll() throws NullPointerException only for a null map; the parameter is non-null.
+            @Suppress("UnsafeThirdPartyFunctionCall")
```
0xnm: normally it is already declared safe https://github.com/DataDog/dd-sdk-android/blob/dcc5eb4b258a22c368dec2b626c27e0d0bf39790/detekt_custom_safe_calls.yml#L986

Maybe something wrong in YML definition signature?

detekt_custom_safe_calls.yml
```diff
@@ -700,6 +700,7 @@ datadog:
       - "java.util.Date.constructor(kotlin.Long)"
       - "java.util.function.Consumer(kotlin.Function1)"
       - "java.util.HashMap.clear()"
+      - "java.util.HashMap.constructor()"
```
0xnm: I think we don't have it in the PR

0xnm (approved):

0xnm (approved):

ambushwork (approved):

PR #3204, opened 2026-02-27
RUM-14782: Capture OkHttp Network Headers

kikoveiga:
### What does this PR do?

- Add new `trackResourceHeaders` public api in `DatadogInterceptor`.
- Add new attributes for request and response headers in `RumAttributes`.
- Extract the contents of these attributes in `RumResourceScope` and add them to the `resource`. 
- Business logic in `ResourceHeadersExtractor`.
- Disable caches of networking libraries used in the sample app.

### Motivation

Customers weren't able to easily capture network headers, having to do it manually (e.g., by creating custom attributes).

### Additional notes

Here is an example taken from the sample app:

image

This is aligned with the iOS implementation: 
- https://github.com/DataDog/dd-sdk-ios/pull/2699
- https://github.com/DataDog/dd-sdk-ios/pull/2700


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAttributes.kt
```diff
@@ -153,6 +153,16 @@ object RumAttributes {
      */
     const val GRAPHQL_ERRORS: String = "_dd.graphql.errors"
 
+    /**
+     * The request headers of the resource. (Map<String, String>)
```
0xnm, line 157: It is not quite clear what is expected to be key and value of this map, given that per HTTP spec headers can be repetitive, so if header name is the key, the value is `List<String>`. In such case how the values are joined to the single `String`, what is the separator?

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -225,6 +237,11 @@ interface com.datadog.android.rum.internal.monitor.AdvancedNetworkRumMonitor : c
   fun stopResource(com.datadog.android.rum.resource.ResourceId, Int?, Long?, com.datadog.android.rum.RumResourceKind, Map<String, Any?> = emptyMap())
   fun stopResourceWithError(com.datadog.android.rum.resource.ResourceId, Int?, String, com.datadog.android.rum.RumErrorSource, Throwable, Map<String, Any?> = emptyMap())
   fun stopResourceWithError(com.datadog.android.rum.resource.ResourceId, Int?, String, com.datadog.android.rum.RumErrorSource, String, String?, Map<String, Any?> = emptyMap())
+class com.datadog.android.rum.internal.net.ResourceHeadersExtractor
+  constructor(com.datadog.android.rum.configuration.ResourceHeadersConfiguration)
+  fun extractRequestHeaders(Map<String, List<String>>, com.datadog.android.api.InternalLogger): Map<String, String>
+  fun extractResponseHeaders(Map<String, List<String>>, com.datadog.android.api.InternalLogger): Map<String, String>
+  companion object
```
0xnm: must be `internal`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @param includeDefaults When `true` (default), the built configuration will include
+     * [DEFAULT_REQUEST_HEADERS] and [DEFAULT_RESPONSE_HEADERS] in addition to any custom
+     * headers specified via [captureHeaders]. When `false`, only custom headers are captured.
+     */
+    class Builder(private val includeDefaults: Boolean = true) {
+
+        internal var logger: InternalLogger = InternalLogger.UNBOUND
+
+        private var customHeaders: List<String> = emptyList()
```
0xnm: maybe simply `val` and `mutableListOf`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
```diff
@@ -0,0 +1,148 @@
...
+
+        /**
+         * Specifies additional header names to capture from both requests and responses.
+         * Headers matching the [SECURITY_PATTERN] are silently filtered out.
+         *
+         * @param headers the list of header names to capture (case-insensitive).
+         * @return this builder for chaining.
+         */
+        fun captureHeaders(headers: List<String>): Builder = apply {
+            customHeaders += headers.map { it.lowercase() }
```
0xnm: need to specify `Locale.US`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
```diff
@@ -0,0 +1,148 @@
...
+                    level = InternalLogger.Level.WARN,
+                    target = InternalLogger.Target.USER,
+                    messageBuilder = {
+                        SENSITIVE_HEADER_WARNING.format(Locale.US, sensitiveRequested.joinToString(", "))
+                    }
+                )
+            }
+
+            val requestHeaders = if (includeDefaults) {
+                (DEFAULT_REQUEST_HEADERS + filteredCustom).distinct()
```
0xnm: we can simply use `Set` then instead of `List` for the `customHeaders` property

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
```diff
@@ -0,0 +1,148 @@
...
+         * Regex pattern used to detect sensitive headers. Any header name matching this
+         * pattern is automatically filtered out, even if explicitly requested.
+         * Catches headers related to authentication, cookies, tokens, secrets, credentials,
+         * API keys, and client IP addresses.
+         */
+        val SECURITY_PATTERN: Regex = Regex(
+            pattern = "(token|cookie|secret|authorization|password|credential|bearer" +
+                "|(api|secret|access|app).?key|forwarded|real.?ip|connecting.?ip|client.?ip)",
+            option = RegexOption.IGNORE_CASE
+        )
```
0xnm: Is it taken from somewhere? Maybe we can check sanitization patterns in https://github.com/DataDog/dd-trace-java

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            return ResourceHeadersConfiguration(
+                requestHeaders = requestHeaders,
+                responseHeaders = responseHeaders
+            )
+        }
+    }
+
+    companion object {
```
0xnm: Do we need properties here to be public? Or it is just for the docs purpose?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,94 @@
...
+    ): Map<String, String> {
+        return extractHeaders(headers, configuration.responseHeaders, internalLogger)
+    }
+
+    private fun extractHeaders(
+        headers: Map<String, List<String>>,
+        allowedHeaders: List<String>,
+        internalLogger: InternalLogger
+    ): Map<String, String> {
+        val normalizedHeaders = headers.mapKeys { it.key.lowercase() }
```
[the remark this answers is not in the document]
kikoveiga: Fixed.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,95 @@
...
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.configuration.ResourceHeadersConfiguration
+import java.util.Locale
+
+/**
+ * Extracts allowed HTTP headers from network requests and responses
+ * based on [ResourceHeadersConfiguration].
+ */
+// TODO RUM-14784 mark this class as internal when trackResourceHeaders moves to
```
0xnm: This cannot be easily done until the next major release lands though if public visibility makes it with a release before the referenced change is done. Be careful here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,95 @@
...
+        private const val MAX_HEADER_VALUE_BYTES = 128
+        private const val HEADER_SIZE_LIMIT_BYTES = ResourceHeadersConfiguration.HEADER_SIZE_LIMIT_BYTES
+        private const val MAX_HEADERS_COUNT = 100
+
+        internal fun truncateToByteLimit(value: String, maxBytes: Int): String {
+            val bytes = value.toByteArray(Charsets.UTF_8)
+            if (bytes.size <= maxBytes) return value
+            // Decode back to string to avoid splitting multibyte characters
+            return String(bytes, 0, maxBytes, Charsets.UTF_8).trimEnd('\uFFFD')
+        }
```
0xnm: We shouldn't do it like this. We already have this in the code https://github.com/DataDog/dd-sdk-android/blob/5e1d6b3fb7fb796aa12171b29a9e36d21b256c06/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScope.kt#L604 we can reuse it.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,95 @@
...
+        val normalizedHeaders = headers.mapKeys { it.key.lowercase(Locale.US) }
+        val result = mutableMapOf<String, String>()
+        var currentSize = 0
+
+        for (headerName in allowedHeaders) {
+            if (result.size >= MAX_HEADERS_COUNT) break
+
+            normalizedHeaders[headerName]?.let { values ->
+                val value = truncateToByteLimit(values.joinToString(", "), MAX_HEADER_VALUE_BYTES)
+                val entrySize = headerName.length + 1 + value.toByteArray(Charsets.UTF_8).size
```
0xnm: 1 is for some separator?

0xnm: maybe we can avoid one more `toByteArray` call and actually return both value and as size as a pair from `truncateToByteLimit`?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScopeTest.kt
```diff
@@ -3938,6 +3938,91 @@ internal class RumResourceScopeTest {
...
+    @Test
+    fun `M map request and response headers to ResourceEvent W handleEvent { header attributes present }`(
+        @Forgery kind: RumResourceKind,
+        @LongForgery(200, 600) statusCode: Long,
+        @LongForgery(0, 1024) size: Long,
+        forge: Forge
+    ) {
+        // Given
+        val requestHeaders = mapOf("content-type" to "application/json", "cache-control" to "no-cache")
+        val responseHeaders = mapOf("etag" to "\"abc123\"", "content-length" to "1024")
```
0xnm: nit: it is actually worth to generate a random data according to the spec of HTTP headers format.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractorTest.kt
```diff
@@ -0,0 +1,236 @@
...
+
+    // region Extraction
+
+    @Test
+    fun `M extract matching request headers W extractRequestHeaders()`() {
+        // Given
+        val headers = mapOf(
+            "Content-Type" to listOf("application/json"),
+            "Cache-Control" to listOf("no-cache")
+        )
```
0xnm: same here, prefer random data

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
kikoveiga: Fixed, thanks!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumAttributes.kt
kikoveiga: We will use [comma-separated values](https://github.com/DataDog/dd-sdk-android/blob/8aed1748c7b4763af875458f8ae7cf461bd00a30/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt#L61), as also shown in the picture of the PR description.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
kikoveiga: Yes this is [aligned with iOS](https://github.com/DataDog/dd-sdk-ios/blob/1349091ec7789a5559295e34b96193d1209ebbf5/DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift#L55) and the product brief.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
kikoveiga: Yes for the `:` separator between key and values, doesn't make a noticeable difference whether to include or not.

features/dd-sdk-android-rum/api/apiSurface
kikoveiga: I marked this as TODO in the `ResourceHeadersExtractor` class definition, is it ok to leave as public for now? For part 2 of this, after Cronet integration and refactoring is done, we will move everything to rum module and then can mark the class as `internal`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
kikoveiga: Maybe it's useful to keep the order of `customHeaders` the customer has decided. With a set, we would lose order and still need to deduplicate when joining default and custom headers.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
kikoveiga: Fixed.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
ambushwork: I also have question here, what' the point to keep the order of customer's headers, can you elaborate? also `List` will allow duplication, which seems not necessary

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }
+        }
+
+        return result
+    }
+
+    companion object {
+        private const val MAX_HEADER_VALUE_BYTES = 128
+        private const val HEADER_SIZE_LIMIT_BYTES = ResourceHeadersConfiguration.HEADER_SIZE_LIMIT_BYTES
```
ambushwork: you can just use `HEADER_SIZE_LIMIT_BYTES` with static import from `ResourceHeadersConfiguration`, and no need to declare it again

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,95 @@
...
+        var currentSize = 0
+
+        for (headerName in allowedHeaders) {
+            if (result.size >= MAX_HEADERS_COUNT) break
+
+            normalizedHeaders[headerName]?.let { values ->
+                val value = truncateToByteLimit(values.joinToString(", "), MAX_HEADER_VALUE_BYTES)
+                val entrySize = headerName.length + 1 + value.toByteArray(Charsets.UTF_8).size
+
+                if (currentSize + entrySize > ResourceHeadersConfiguration.HEADER_SIZE_LIMIT_BYTES) {
```
ambushwork:
```suggestion
                if (currentSize + entrySize > HEADER_SIZE_LIMIT_BYTES) {
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
kikoveiga: Moved this function from `RumResourceScope` to a new `StringExt` file and adapted to return a pair as you suggested below. Thank you!

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumResourceScopeTest.kt
kikoveiga: Created a `Forge.anHttpHeaderMap` function in `ForgeExt`.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -450,6 +472,18 @@ open class DatadogInterceptor internal constructor(
...
         }
+
+        /**
+         * Enables capturing HTTP request and response headers in RUM Resource events.
+         *
+         * @param configuration the [ResourceHeadersConfiguration] specifying which headers to capture.
+         * Defaults to a configuration that captures common safe headers.
+         * @return this builder for chaining.
+         */
+        @ExperimentalRumApi
```
ambushwork: Usually we mark the API with this with two main concerns:
* This API is likely to change in the future
* It brings known side-effect to users and it's not stable yet(such as performance overhead).

What's the case for network headers here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
kikoveiga: In cases where the 2 KB limit for all headers might be exceeded, we want to keep first the `DEFAULT_HEADERS` and then customer's `customHeaders`. While the default ones are ordered alphabetically, the customer can decide to have some order on the custom ones so they prefer to have more chances of keeping this or that header (even though this is not explicit in the `captureHeaders` api. If this is not relevant, then yes we could just switch to using a set.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractorTest.kt
kikoveiga: Done, thank you.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
kikoveiga: Fixed, thanks!

0xnm (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/ResourceHeadersExtractor.kt
kikoveiga: Thank you for pointing this out! I discussed with @satween about this and the ongoing refactoring, so we decided we can simplify this by removing the `Configuration` class and just put everything inside `Extractor`:

`trackResourceHeaders(extractor: ResourceHeadersExtractor`)

Because the extractor is already in rum module and will always be public, it becomes much easier to instrument this when we create a new `trackResourceHeaders` in `RumNetworkInstrumentation`, wdyt?

features/dd-sdk-android-rum/api/apiSurface
kikoveiga: `companion object` in `ResourceHeadersExtractor` is now `internal`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
kikoveiga: Now `internal` in `ResourceHeadersExtractor`.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
kikoveiga: Agreed it didn't apply, removed the annotation.

features/dd-sdk-android-rum/src/main/json/rum/resource-schema.json
kikoveiga, line 1: - https://github.com/DataDog/rum-events-format/pull/363

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/configuration/ResourceHeadersConfiguration.kt
ambushwork: it's more like a product decision, I am curious what iOS does in this case, they also drop headers in the end of the list?

kikoveiga: Yes this was a product decision in the [brief](https://docs.google.com/document/d/1mSFwY4nEHq8Ji36NTvhyuztSHJfamunvTc9IlMdTtNo/edit?tab=t.0):
- To limit the size of the data to be captured and avoid ‘bad surprises’ (customers that would use a custom header to get the entire request payload), we limit the total size of the collected content of headers to 2KB.


 iOS is aligned.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
```diff
@@ -228,13 +244,15 @@ open class DatadogInterceptor internal constructor(
...
+        val headerAttributes = extractHeaderAttributes(sdkCore, request, response)
+
         @Suppress("DEPRECATION")
         (GlobalRumMonitor.get(sdkCore) as? AdvancedNetworkRumMonitor)?.stopResource(
             requestId,
             statusCode,
             getBodyLength(response, sdkCore.internalLogger),
             kind,
-            attributes + rumResourceAttributesProvider.onProvideAttributes(request, response, null)
+            attributes + headerAttributes + rumResourceAttributesProvider.onProvideAttributes(request, response, null)
```
ambushwork: `onProvideAttributes` functions allows to return Map<String, Any?>, but in `RumResourceScope` you try to cast it to Map<String,String> with `as?`, what will happen if the client implements `rumResourceAttributesProvider` but returns different stuff than `Map<String,String>`?

kikoveiga: So does it make sense to change the order of the `+` here?

`rumResourceAttributesProvider.onProvideAttributes(request, response, null) + attributes + headerAttributes`

This is  lowest to highest priority (left to right):                                                                                                                                                                                            
  1. Customer attributes — can be overridden by everything
  2. Trace/GraphQL attributes — SDK internals override customer if conflict                                                                                                                                                              
  3. Header attributes — always win, since the SDK controls these keys entirely

ambushwork: yes, it's safer to put `headerAttributes` in the end to protect our header attributes not overwritten by user. also we need unit tests for this corner case.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/StringExt.kt
```diff
@@ -0,0 +1,71 @@
...
+        return Pair("", 0)
+    }
+
+    // will not throw IndexOutOfBoundsException
+    @Suppress("UnsafeThirdPartyFunctionCall")
+    val truncated = substring(0, src.position())
+    return Pair(truncated, dst.position())
+}
+
+private fun logTruncationFailure(internalLogger: InternalLogger?, e: Throwable) {
```
satween: Could be extension method:

```
InternalLogger.logTruncationFailure(e: Throwable) {
...
}
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/resource/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,219 @@
...
+    ): Map<String, String> {
+        return extractHeaders(headers, responseHeaders, internalLogger)
+    }
+
+    private fun extractHeaders(
+        headers: Map<String, List<String>>,
+        allowedHeaders: List<String>,
+        internalLogger: InternalLogger
+    ): Map<String, String> {
+        val normalizedHeaders = headers.mapKeys { it.key.lowercase(Locale.US) }
```
satween, line 126: Would this algo being able to handle this scenario: 

small
SUPER SUPER BIG HEADER
small
small 
small

?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/resource/ResourceHeadersExtractor.kt
```diff
@@ -0,0 +1,219 @@
...
+        val DEFAULT_REQUEST_HEADERS: List<String> = listOf(
+            "cache-control",
+            "content-type"
+        )
+
+        /**
+         * Default response headers captured when `includeDefaults` is `true`.
+         */
+        val DEFAULT_RESPONSE_HEADERS: List<String> = listOf(
+            "cache-control",
```
satween, line 171: maybe worth moving headers inside `HttpSpec`?

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/DatadogInterceptor.kt
kikoveiga: Moved to the end for now.

ambushwork: As I mentioned a unit test is needed here, because this behavior can be easily broken by changing the order, so this unit test should:
* fail - when the order is `attributes + headerAttributes + rumResourceAttributesProvider.onProvideAttributes(request, response, null)`
* succeed - when the order is `attributes + rumResourceAttributesProvider.onProvideAttributes(request, response, null) + headerAttributes`

kikoveiga: Added a test in the end of `DatadogInterceptorTest`, thank you!

ambushwork (dismissed):
🚀 Well done!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/resource/ResourceHeadersExtractor.kt
kikoveiga: We iterate `allowedHeaders` (default headers first, then customer headers). We keep appending while there is space, and `log` a message if an header is skipped.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/resource/ResourceHeadersExtractor.kt
kikoveiga: I think these headers are specific to this feature, so it makes sense to keep them here. If we place them in `HttpSpec` and then add more headers there in the future, it will change the behaviour here, wdyt?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/StringExt.kt
kikoveiga: Fixed, thank you.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/resource/ResourceHeadersExtractor.kt
satween: Just make sure that this is the same logic as on iOS

satween (dismissed):

features/dd-sdk-android-rum/api/apiSurface
```diff
@@ -260,6 +262,13 @@ typealias RumVitalOperationStepEvent = VitalOperationStepEvent
 fun android.content.Context.getAssetAsRumResource(String, Int = AssetManager.ACCESS_STREAMING, com.datadog.android.api.SdkCore = Datadog.getInstance()): java.io.InputStream
 fun android.content.Context.getRawResAsRumResource(Int, com.datadog.android.api.SdkCore = Datadog.getInstance()): java.io.InputStream
 fun java.io.InputStream.asRumResource(String, com.datadog.android.api.SdkCore = Datadog.getInstance()): java.io.InputStream
+class com.datadog.android.rum.resource.ResourceHeadersExtractor
+  class Builder
+    constructor(Boolean = true)
+    fun captureHeaders(String): Builder
+    fun build(): ResourceHeadersExtractor
+  fun extractRequestHeaders(Map<String, List<String>>, com.datadog.android.api.InternalLogger): Map<String, String>
+  fun extractResponseHeaders(Map<String, List<String>>, com.datadog.android.api.InternalLogger): Map<String, String>
```
0xnm: Is there a way to make these internal? Maybe leverage calling these through the `_RumInternalProxy`? These methods are not supposed to be exposed to the customer.

0xnm (changes requested):

features/dd-sdk-android-rum/api/apiSurface
kikoveiga: Done, thank you.

0xnm (approved):

PR #3205, opened 2026-02-27
PANA-5681: Integrate HeatmapIdentifierRegistry into RUM for heatmaps [3 of 4]

gonzalezreal:
### What does this PR do?

Integrates heatmap tap data into the RUM feature. When a tap gesture is detected and the view is attached to a window, GesturesListener captures the touch coordinates, the view's dimensions, and (when available) its heatmap identifier from the `HeatmapIdentifierRegistry`, and packages them into a new `HeatmapActionData` object. This data flows through `AdvancedRumMonitor.addActionWithHeatmap` — an internal API that keeps `HeatmapActionData` off the public `RumMonitor` interface — into `RumActionScope`, where it is serialised into the `dd.action` field of the outgoing `ActionEvent`. `RumFeature` now implements `HeatmapIdentifierRegistryProvider`, exposing its lazily-initialised registry to Session Replay via `FeatureScope.unwrap()` without going through the untyped feature-context map.

### Motivation

Support mobile heatmaps. This is PR 3 of 4, integrates the `HeatmapIdentifierRegistry` (#3202) and updated schemas (#3203) into RUM.

### Additional Notes

- Depends on #3203
- Ported from #3164


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -195,14 +198,14 @@ internal class DatadogRumMonitor(
     override fun addAction(type: RumActionType, name: String, attributes: Map<String, Any?>) {
         val eventTime = getEventTime(attributes)
         handleEvent(
-            RumRawEvent.StartAction(type, name, false, attributes.toMap(), eventTime)
+            RumRawEvent.StartAction(type, name, false, attributes, eventTime)
```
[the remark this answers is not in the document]
gonzalezreal: I am not sure why @jonathanmos introduced this change. Either it was unintentional or he wanted to avoid one allocation per action.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
```diff
@@ -257,19 +259,26 @@ internal class GesturesListener(
...
-                it.extractAttributes(view, attributes)
+            addViewAttributes(view, attributes)
+
+            if (view.isAttachedToWindow) {
+                val locationOnScreen = IntArray(2)
+                @Suppress("UnsafeThirdPartyFunctionCall") // locationOnScreen is non-null with exactly 2 elements
+                view.getLocationOnScreen(locationOnScreen)
+
+                val relativeX = (touchX - locationOnScreen[0]).toLong()
+                val relativeY = (touchY - locationOnScreen[1]).toLong()
```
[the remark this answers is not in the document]
gonzalezreal: Fixed in [e76417d](https://github.com/DataDog/dd-sdk-android/pull/3205/commits/e76417d5aa6775e59debef6b591ca4d2b9b75fc8)

jonathanmos:
@codex review

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
```diff
@@ -305,6 +323,29 @@ internal class GesturesListener(
...
+        val screenName = currentScreenName() ?: return null
+
+        return heatmapIdentifierRegistry
+            .getHeatmapIdentifier(System.identityHashCode(view).toLong(), screenName)
+            ?.rawValue
+    }
+
+    private fun currentScreenName(): String? {
+        val featureSdkCore = sdkCore as? FeatureSdkCore ?: return null
+        return featureSdkCore
```
kikoveiga: I think we should always be able to perform this cast, so can drop the nullable:
```suggestion
        return (sdkCore as FeatureSdkCore)
```

kikoveiga (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierRegistry.kt
```diff
@@ -17,22 +17,22 @@ interface HeatmapIdentifierRegistry {
     /**
      * Replaces the current snapshot with [identifiers], scoped to [screenName].
      *
-     * @param identifiers a map of [android.view.View.getId] values to their [HeatmapIdentifier]s,
+     * @param identifiers a map of [System.identityHashCode] values to their [HeatmapIdentifier]s,
```
0xnm: I think @aleksandr-gringauz had a point regarding `identityHashCode` usage (that there is a clash after certain number of objects, since it is re-used), but since we use it in the Session Replay for the object identity, I guess this is fine.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -178,6 +180,7 @@ internal class RumFeature(
     internal var displayInfoProvider: InfoProvider<DisplayInfo> = NoOpDisplayInfoProvider()
     internal val rumContextUpdateReceivers = mutableSetOf<FeatureContextUpdateReceiver>()
     internal var insightsCollector: InsightsCollector = NoOpInsightsCollector()
+    override val heatmapIdentifierRegistry: HeatmapIdentifierRegistry by lazy { HeatmapIdentifierRegistry.create() }
```
0xnm: what is the need for the `lazy`? why we cannot simply provide it in the constructor or initialize directly?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumActionScope.kt
```diff
@@ -354,7 +356,8 @@ internal class RumActionScope(
                     session = ActionEvent.DdSession(
                         sessionPrecondition = rumContext.sessionStartReason.toActionSessionPrecondition()
                     ),
-                    configuration = ActionEvent.Configuration(sessionSampleRate = sampleRate)
+                    configuration = ActionEvent.Configuration(sessionSampleRate = sampleRate),
+                    action = ddAction
```
0xnm: nit: since `heatmapData` is immutable reference and `HeatmapActionData` is a data class, we can simply drop `ddAction` (name is a bit confusing) and do object creations here directly:

```suggestion
                    action = heatmapData?.let { ActionEvent.DdAction(...) }
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
```diff
@@ -257,29 +265,44 @@ internal class GesturesListener(
...
-            val targetId: String = contextRef.get().resourceIdName(view.id)
-            attributes[RumAttributes.ACTION_TARGET_CLASS_NAME] = view.targetClassName()
-            attributes[RumAttributes.ACTION_TARGET_RESOURCE_ID] = targetId
-            attributesProviders.forEach {
-                it.extractAttributes(view, attributes)
+            addViewAttributes(view, attributes)
+
+            if (view.isAttachedToWindow) {
+                resolveHeatmapIdentifier(view)?.let { identity ->
+                    val locationInWindow = IntArray(2)
```
0xnm: minor: since we expect all the UI events to happen on the main thread, can we just allocate static array and re-use it for the calls? instead of allocating a new one for each call.

although probably not a big difference since it is for the taps, they don't happen that often

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/AdvancedRumMonitor.kt
```diff
@@ -28,6 +30,13 @@ internal interface AdvancedRumMonitor : RumMonitor, AdvancedNetworkRumMonitor {
 
     fun start()
 
+    fun addActionWithHeatmap(
+        type: RumActionType,
+        name: String,
+        attributes: Map<String, Any?>,
+        heatmapData: HeatmapActionData?
```
0xnm: nit: generic `attributes` argument should come last

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumRawEvent.kt
```diff
@@ -42,7 +42,8 @@ internal sealed class RumRawEvent {
         val name: String,
         val waitForStop: Boolean,
         val attributes: Map<String, Any?>,
-        override val eventTime: Time = Time()
+        override val eventTime: Time = Time(),
+        val heatmapData: HeatmapActionData? = null
```
0xnm: nit: maybe put it before generic `attributes` and `eventTime` in the arguments list?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListenerScrollSwipeTest.kt
```diff
@@ -416,8 +425,18 @@ internal class GesturesListenerScrollSwipeTest : AbstractGesturesListenerTest()
         testedListener.onUp(endUpEvent)
 
         // Then
-        verify(rumMonitor.mockInstance)
-            .addAction(RumActionType.TAP, "", expectedStartAttributes)
+        verify(rumMonitor.mockInstance as AdvancedRumMonitor).addActionWithHeatmap(
+            eq(RumActionType.TAP),
+            eq(""),
+            argThat { attributes ->
```
0xnm: `argThat` gives just a generic description when `verify` fails. It is better to use `argumentCaptor` here and assert the argument independently.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListenerScrollSwipeTest.kt
```diff
@@ -1205,6 +1236,66 @@ internal class GesturesListenerScrollSwipeTest : AbstractGesturesListenerTest()
...
+            eq(RumActionType.SCROLL),
+            any(),
+            any()
+        )
+        verify(rumMonitor.mockInstance).stopAction(
+            eq(RumActionType.SCROLL),
+            any(),
+            any()
+        )
+        verify(rumMonitor.mockInstance as AdvancedRumMonitor, never()).addActionWithHeatmap(
```
0xnm: will `verifyNoMoreInteractions(rumMonitor.mockInstance)` work here?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListenerTapTest.kt
```diff
@@ -561,16 +583,24 @@ internal class GesturesListenerTapTest : AbstractGesturesListenerTest() {
...
 
         // Then
-        verify(rumMonitor.mockInstance).addAction(
-            RumActionType.TAP,
-            "",
-            expectedAttributes
+        verify(rumMonitor.mockInstance as AdvancedRumMonitor).addActionWithHeatmap(
+            eq(RumActionType.TAP),
+            eq(""),
+            argThat { attributes ->
```
0xnm: it is better to use argument captor, same for the similar tests below

0xnm (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/HeatmapIdentifierRegistry.kt
jonathanmos: yes, that is actually the reason I chose to use it here. We don't expect collisions on the same screen, but if we reach a conclusion that there is an issue we could append the hashcode of the direct parent for an even lower collision probability

aleksandr-gringauz: Yes. See [this](https://github.com/DataDog/dd-sdk-android/pull/3349#discussion_r3064140740) thread.

> We don't expect collisions on the same screen

I didn't dig deep into how the identifier obtained using `identityHashCode` is used in this PR.

However in general, it is important to understand the following. 

2 instances of the same type that currently have strong references in code can have the same `identityHashCode`. And the probability of this is higher than you might expect. This is what I explain in my [comment](https://github.com/DataDog/dd-sdk-android/pull/3349#discussion_r3064140740).

The questions are:
1. What are the requirements here for not having collisions? Do we have to avoid collisions during the lifetime of a single RUM View? 
2. And what is the number of "tries"? Is it the number of `View` instances that ever existed during the lifetime of the RUM View?

jonathanmos: as we discussed, I have made the key comprise the view identityHashCode and the identityHashCode of it's direct parent so that we have a much lower collision risk

aleksandr-gringauz (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/TapTargetUtils.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Single canonical tap-target predicate shared by the RUM gesture layer and Session Replay,
+ * so both subsystems stay in sync automatically.
+ *
+ * Note: [View.isEnabled] is intentionally not checked. Disabled views retain [View.isClickable]
+ * = true on Android but never receive touch events, so the RUM gesture layer never fires for
+ * them. Excluding `isEnabled` here keeps the predicate identical to RUM's existing behaviour.
+ */
+fun View.isValidTapTarget(): Boolean {
```
0xnm, line 19: do we need to have it in `-internal` module? seems like it is used only in `-rum`.

normally `-internal` module shouldn't be visible on the user side, but if they add it, they will have this extension.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
```diff
@@ -305,6 +328,27 @@ internal class GesturesListener(
...
+    private fun resolveHeatmapIdentifier(view: View): String? {
+        val screenName = currentScreenName() ?: return null
+
+        return heatmapIdentifierRegistry
+            .getHeatmapIdentifier(heatmapViewKey(view), screenName)
+            ?.rawValue
+    }
+
+    private fun currentScreenName(): String? {
+        return sdkCore.getFeatureContext(Feature.RUM_FEATURE_NAME)[RumContext.VIEW_URL] as? String
```
0xnm: > [!WARNING]
> This call is blocking and happens on the context thread. Meaning that if context thread is busy, caller thread (main thread) will have to wait and can cause ANR. Maybe consider getting screen name later? When event with heatmap data enters `RumScope` for processing.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/heatmaps/TapTargetUtils.kt
jonathanmos: It will be used also by session replay in the next pr

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListener.kt
jonathanmos: good catch - I've made this change

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -60,7 +61,8 @@ internal class RumSessionScope(
     private val sessionMaxDurationNanos: Long = DEFAULT_SESSION_MAX_DURATION_NS,
     rumSessionTypeOverride: RumSessionType?,
     private val rumSessionScopeStartupManagerFactory: () -> RumSessionScopeStartupManager,
-    insightsCollector: InsightsCollector
+    insightsCollector: InsightsCollector,
+    heatmapIdentifierRegistry: HeatmapIdentifierRegistry? = null
```
0xnm: should we remove default value? no other arguments of reference type in this constructor are using it.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -89,7 +90,8 @@ internal open class RumViewScope(
     private val accessibilitySnapshotManager: AccessibilitySnapshotManager,
     private val batteryInfoProvider: InfoProvider<BatteryInfo>,
     private val displayInfoProvider: InfoProvider<DisplayInfo>,
-    private val insightsCollector: InsightsCollector
+    private val insightsCollector: InsightsCollector,
+    private val heatmapIdentifierRegistry: HeatmapIdentifierRegistry? = null
```
0xnm: same as above, but here we already have `private val featuresContextResolver: FeaturesContextResolver = FeaturesContextResolver()` though

0xnm (dismissed):

0xnm (approved):

aleksandr-gringauz (approved):

PR #3206, opened 2026-02-27
PANA-5681: Integrate HeatmapIdentifierRegistry into Session Replay for heatmaps [4 of 4]

gonzalezreal:
### What does this PR do?

Wires `ViewIdentityResolver` into Session Replay to populate `permanentId` on wireframes. This enables correlating wireframes with RUM action events for heatmap visualization.

Key changes:
- Adds public `ViewIdentityProvider` interface for Session Replay's view identity resolution
- Adapts internal `ViewIdentityResolver` (from RUM feature context) via `ViewIdentityResolverAdapter`
- Adds `viewIdentityProvider` to `MappingContext` for use by all wireframe mappers
- Populates `permanentId` in all wireframe types (Shape, Text, Image, Placeholder, Webview)
- Calls `onWindowRefreshed()` on each draw pass to index the current view hierarchy

### Motivation

Support mobile heatmaps. This is the final piece connecting Session Replay wireframes to RUM actions.

### Additional Notes

- Depends on #3205
- Ported from #3164. Also updates `MappingContextForgeryFactory` in `session-replay-compose` and `session-replay-material` test suites.


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/utils/ViewIdentityResolverExt.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.internal.identity.NoOpViewIdentityResolver
+import com.datadog.android.internal.identity.ViewIdentityResolver
+
+/**
+ * Retrieves [ViewIdentityResolver] from RUM's feature context, or [NoOpViewIdentityResolver] if unavailable.
+ */
+internal fun FeatureSdkCore.getViewIdentityResolver(): ViewIdentityResolver {
+    val rumContext = getFeatureContext(Feature.RUM_FEATURE_NAME)
```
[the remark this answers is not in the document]
jonathanmos: No longer relevant - in the current approach we avoid getFeatureContext entirely

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/DeferredHeatmapIdentifierRegistry.kt
```diff
@@ -0,0 +1,62 @@
...
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.internal.heatmaps.HeatmapIdentifier
+import com.datadog.android.internal.heatmaps.HeatmapIdentifierRegistry
+import com.datadog.android.internal.heatmaps.HeatmapIdentifierRegistryProvider
+
+/**
+ * Session Replay may be initialized before RUM — the SDK does not mandate a registration order.
```
0xnm, line 18: Do we actually need to support this case? Cannot we print a warning that RUM is not yet initialized during Session Replay start?

By looking at https://docs.datadoghq.com/session_replay/mobile/setup_and_configuration/?platform=android it is not clear from code snippets that RUM should be initialized first, though (it is only mentioned that RUM SDK setup should be completed).

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/DeferredHeatmapIdentifierRegistry.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    override fun setHeatmapIdentifiers(identifiers: Map<Long, HeatmapIdentifier>, screenName: String) {
+        getDelegate()?.setHeatmapIdentifiers(identifiers, screenName)
+    }
+
+    override fun getHeatmapIdentifier(heatmapViewKey: Long, currentScreenName: String): HeatmapIdentifier? {
+        return getDelegate()?.getHeatmapIdentifier(heatmapViewKey, currentScreenName)
+    }
+
+    private class Resolved(val registry: HeatmapIdentifierRegistry?)
```
0xnm: I'm not sure what is the point of this class, cannot we reference `HeatmapIdentifierRegistry` directly? Or it is for the case when RUM feature doesn't implement `HeatmapIdentifierRegistryProvider`?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/HeatmapIdentifierResolver.kt
```diff
@@ -0,0 +1,167 @@
...
+                }
+            } catch (_: Resources.NotFoundException) {
+            }
+        }
+        return "$LOCAL_KEY_CLASS_PREFIX${view.javaClass.name}#$typeIndex"
+    }
+
+    @VisibleForTesting
+    internal fun computeChildTypeIndices(parent: ViewGroup): IntArray {
+        val typeCounts = HashMap<Class<*>, Int>()
```
0xnm:
```suggestion
        val typeCounts = mutableMapOf()
```

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/HeatmapIdentifierResolver.kt
```diff
@@ -0,0 +1,167 @@
...
+
+    // #typeIndex is appended to containers too: rows sharing a resource name (e.g. RecyclerView
+    // items) must produce distinct path components so their descendants get distinct permanentIds.
+    @VisibleForTesting
+    internal fun pathComponentFor(view: View, typeIndex: Int): String {
+        val viewId = view.id
+        if (viewId != View.NO_ID) {
+            try {
+                @Suppress("UnsafeThirdPartyFunctionCall") // can throw Resources.NotFoundException
+                val name = view.resources?.getResourceName(viewId)
```
0xnm: given that name is static, should we maybe cache lookup result?

how often this method is supposed to be called? did we check performance impact of the feature?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/HeatmapIdentifierResolver.kt
```diff
@@ -0,0 +1,167 @@
...
+        return result
+    }
+
+    @UiThread
+    private fun publishIfChanged(context: TraversalContext) {
+        if (context.entries.isEmpty()) {
+            lastPublishedScreenName = null
+            lastPublishedEntries.clear()
+        } else {
+            @Suppress("UnsafeThirdPartyFunctionCall")
```
0xnm: what can be unsafe here?

0xnm (commented):
I briefly went through the production files and left some comments, I didn't cover tests.

Given the size of this PR we need to have more several reviewers here.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/HeatmapWireframeExt.kt
kikoveiga, line 1: Would it make sense to add `permanentId` to the `_common-wireframe-schema.json` file? So we don't need to do these mappings.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/listener/WindowsOnDrawListener.kt
```diff
@@ -37,7 +39,8 @@ internal class WindowsOnDrawListener(
         sdkCore = sdkCore,
         dynamicOptimizationEnabled = dynamicOptimizationEnabled
     ),
-    private val methodCallSamplingRate: Float
+    private val methodCallSamplingRate: Float,
+    private val rumContextProvider: RumContextProvider = NoOpRumContextProvider()
```
kikoveiga: Why do we need a `NoOp` default here?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/processor/HeatmapWireframeExt.kt
jonathanmos: It's possible to put `permanentId` on the common schema but the reason we didn't do this is that it breaks the previous ordering of arguments for the wireframe constructors instead of `permanentId` being the last argument, and it also wouldn't remove the need for these mappings. The root issue is that `.copy()` only exists on data classes, not on the `Wireframe` sealed class, so to stamp `permanentId` onto a wireframe subtype we have to downcast via `when`. To avoid that we'd need to touch the schema generator to emit a `withPermanentId()` method on each subtype that delegates to its own `copy()`.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/DeferredHeatmapIdentifierRegistry.kt
jonathanmos: The problem is that without this mechanism, the heatmaps registry is resolved once at SessionReplay.enable() time. If RUM hasn't registered yet then the registry is permanently null and heatmaps are silently disabled for the entire session. Printing a warning means heatmaps are still broken, and the user has no way to recover without restarting the app.

hamorillo (commented):
I went through the production code and haven't found any blockers. That said, the PR is quite big (so it's not easy to review), and it's my first time reviewing SR code, so I'm not confident at all, as I spend most of the time trying to understand everything.

I'll try to come back later with fresh eyes.

kikoveiga (dismissed):

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/processor/NodeFlattenerTest.kt
```diff
@@ -168,6 +171,87 @@ internal class NodeFlattenerTest {
...
+
+    @Test
+    fun `M stamp permanentId on every wireframe W flattenNode() { heatmapIdentifier set }`(forge: Forge) {
+        val fakeIdentifier = HeatmapIdentifier(forge.anAlphabeticalString())
+        val fakeWireframes = listOf<MobileSegment.Wireframe>(
+            forge.getForgery<MobileSegment.Wireframe.ShapeWireframe>().copy(permanentId = null),
+            forge.getForgery<MobileSegment.Wireframe.TextWireframe>().copy(permanentId = null),
+            forge.getForgery<MobileSegment.Wireframe.ImageWireframe>().copy(permanentId = null),
+            forge.getForgery<MobileSegment.Wireframe.PlaceholderWireframe>().copy(permanentId = null),
+            forge.getForgery<MobileSegment.Wireframe.WebviewWireframe>().copy(permanentId = null)
```
hamorillo: 🤔 I have to recognize that this is a pain. I read the comment about why each type has its own `copy` method, but it looks a bit flaky. If we add a new Wireframe type, this test will pass. Anyway, just expressing some thoughts.

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/listener/WindowsOnDrawListenerTest.kt
```diff
@@ -319,6 +339,45 @@ internal class WindowsOnDrawListenerTest {
         }
     }
 
+    @Test
+    fun `M pass viewUrl from RumContextProvider W onDraw()`(
+        @fr.xgouchet.elmyr.annotation.StringForgery fakeViewUrl: String
```
hamorillo: nit: 
```suggestion
        @StringForgery fakeViewUrl: String
```

hamorillo (dismissed):
Looks good to me in general. As I mentioned, I'm not fully confident in my review given the size, and this is my first time reviewing SR. So feel free to merge it if you are confident with the code. 

🚀

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/processor/NodeFlattenerTest.kt
jonathanmos: I've modified this to make it easier to maintain

hamorillo (approved):

PR #3216, opened 2026-03-03
Use feature flag variant `key` instead of `value` for RUM Feature Flag Tracking

petzel:
### What does this PR do?
This PR updates how we Datadog Feature Flags are tracked in RUM, using a variant's `key` rather than the variant's `value`.

### Motivation
Previously, when using Datadog Feature Flags, the `value` would be set as the evaluated variant. This is not ideal, particularly when using JSON type flags. JSON flags can be very large, and are not friendly from a UX perspective when typing filters/viewing results. Each variant in a Datadog Feature Flag contains a `key`, which is a string with a maximum length of 64 characters. This `key` is better suited for tracking feature flags in RUM. 

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
@@ -14,7 +14,6 @@ import com.datadog.android.flags.FlagsConfiguration
 import com.datadog.android.flags.StateObservable
 import com.datadog.android.flags.internal.evaluation.EvaluationsManager
 import com.datadog.android.flags.internal.model.PrecomputedFlag
-import com.datadog.android.flags.internal.model.VariationType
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt
```diff
@@ -515,39 +515,8 @@ internal class DatadogFlagsClient(
      * Supposed to be used by internal Datadog packages to track flag evaluations from an exact flags state snapshot.
      */
     internal fun trackFlagSnapshotEvaluation(flagKey: String, flag: UnparsedFlag, context: EvaluationContext) {
-        val flagValue = parseFlagValueString(flagKey, flag)
-
-        trackResolution(flagKey, flag, flagValue, context)
-    }
-
-    private fun parseFlagValueString(flagKey: String, flag: UnparsedFlag): Any {
```
typotter, line 523: 📈

PR #3224, opened 2026-03-05
FFL-1878: Fix stale state context mismatch

sameerank:
### What does this PR do?

Fixes a bug where the Flags SDK would transition to `Stale` state when a network request fails, without checking if the cached context matches the requested context. This could result in serving flags from a different user/context.

**Changes:**
1. **EvaluationsManager.kt**: Added context comparison before transitioning to `Stale` state. If contexts don't match, transitions to `Error` instead.
2. **FlagsClientState.kt**: Simplified `NotReady` state documentation to remove incorrect claim about flag availability.

### Motivation

When `setEvaluationContext()` fails due to a network error, the SDK was transitioning to `Stale` state if any cached flags existed, without verifying the cached context matched the requested context. This could cause:

- **Scenario 1**: User A's cached flags being served to User B on a fresh app open
- **Scenario 2**: User A's flags being served to User B after a mid-session user switch

This fix aligns the Android SDK behavior with the iOS SDK implementation.

### Additional Notes

- Added unit tests for both matching and mismatched context scenarios
- The fix follows the same pattern implemented in the iOS SDK (`FlagsRepository.swift`)

Fixes: [FFL-1878](https://datadoghq.atlassian.net/browse/FFL-1878)

### DIFF
--- a/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/FlagsClientState.kt
+++ b/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/FlagsClientState.kt
@@ -13,12 +13,7 @@ sealed class FlagsClientState {
     /**
      * The client is not ready to evaluate flags.
      *
-     * This state occurs:
-     * - Before the first [setEvaluationContext()] call (initialization)
-     * - After client shutdown
-     * - When the provider is not initialized or unavailable
-     *
-     * No flags are available for evaluation in this state.
+     * This state occurs before the first [setEvaluationContext()] call.
      * Maps to OpenFeature's [NOT_READY] state.
      */
     object NotReady : FlagsClientState()


features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/FlagsClientState.kt
```diff
@@ -13,12 +13,7 @@ sealed class FlagsClientState {
     /**
      * The client is not ready to evaluate flags.
      *
-     * This state occurs:
-     * - Before the first [setEvaluationContext()] call (initialization)
-     * - After client shutdown
```
sameerank, line 18: I don't see a `shutdown` method, so if we want to keep this comment here, it would help to point instead to a specific function or code path. The closest thing I could find is `Datadog.stopInstance()`

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/FlagsClientState.kt
```diff
@@ -13,12 +13,7 @@ sealed class FlagsClientState {
...
-     * - When the provider is not initialized or unavailable
-     *
-     * No flags are available for evaluation in this state.
```
sameerank, line 21: This is not quite true. There may be cached flags available

Provider readiness checks if there's an evaluation context (which may be cached) and doesn't check the client's state

https://github.com/DataDog/dd-sdk-android/blob/3fe55a2710e9ba43a4e5b2f6cd19a207698f3feb/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt#L257

[iOS does something similar in checking the cached context](https://github.com/DataDog/dd-sdk-ios/blob/14d5135e099f7870990d795ba50b0702cddab160/DatadogFlags/Sources/Client/FlagsRepository.swift#L96-L99), but there is a difference in behaviors

| Scenario | iOS | Android |
| --- | --- | --- |
| Fresh install, `NotReady` state | context = nil → `PROVIDER_NOT_READY` | Same |
| App restart with cached data, `NotReady` state | context = cached → flags work | Same |
| After `reset()` | context = nil → `PROVIDER_NOT_READY` | N/A (no reset method) |

iOS has a `reset()` which clears data and sets state to `notReady`. Android doesn't have this method.

features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/evaluation/EvaluationsManager.kt
```diff
@@ -94,7 +94,10 @@ internal class EvaluationsManager(
                         )
 
                         val throwable = NetworkRequestFailedException(NETWORK_REQUEST_FAILED_MESSAGE)
-                        if (hadFlags) {
+                        // Only use cached flags if they match the requested context to avoid
+                        // serving flags from a different user/context.
+                        val cachedContextMatches = flagsRepository.getEvaluationContext() == context
+                        if (hadFlags && cachedContextMatches) {
```
sameerank, line 100: This fixes the state transition logic if there's a network error when requesting flags for a context (User B). It's only safe to serve "stale" flags if the requested context matches the cached context

| Cached Flags | Cached Context | Requested Context | State |
| --- | --- | --- | --- |
| None | - | User B | `Error` |
| Exist | User A | User A | `Stale` |
| Exist | User A | User B | `Error` ✓ (was incorrectly `Stale`) |

Assuming User B should never see User A's flags

PR #2800, opened 2025-07-25
RUM-10958: Fix WindowCallbackWrapper NPE

ambushwork:
### What does this PR do?

Due to a system issue: https://issuetracker.google.com/issues/188568911

A null `menu` argument can be passed into a `@NonNull` java function of `Window.Callback`, 
and since our `WindowCallbackWrapper` delegates from this interface, kotlin function will check nullability then causes the crash.

The fix is to tweak the `@NonNull` declaration to `@Nullable` to get away with Kotlin nullable check.

### Motivation

RUM-10958

### Additional Notes

Anything else we should know when reviewing?

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](CONTRIBUTING.md) doc)

PR #3239, opened 2026-03-06
Fix RecorderWindowCallback crash on null menu parameter

nguyexua:
### What does this PR do?

Overrides menu-related `Window.Callback` methods in `RecorderWindowCallback` with nullable parameters, preventing Kotlin's non-null assertions from crashing when Android passes null values.

### Motivation

A customer reported a crash (`NullPointerException: Parameter specified as non-null is null: method RecorderWindowCallback.onMenuOpened, parameter p1`) when opening a custom popup menu.

`RecorderWindowCallback` uses Kotlin's `by wrappedCallback` delegation for `Window.Callback`. Kotlin generates non-null assertions for all delegated parameters, but Android can call these methods with null values (known Android bug: https://issuetracker.google.com/issues/188568911). The previous fix only handled `dispatchTouchEvent` with a try-catch, but the same issue affected all menu-related callbacks.

### Changes

- Override `onMenuOpened`, `onMenuItemSelected`, `onPanelClosed`, `onPreparePanel`, and `onCreatePanelMenu` in `RecorderWindowCallback` with nullable parameters. When null is received, methods return safe defaults (`false` or no-op) instead of crashing.
- Remove the `dispatchTouchEvent` try-catch workaround and `logOrRethrowWrappedCallbackException` helper, since `dispatchTouchEvent` already declares its parameter as nullable.
- Add tests for menu callback delegation and null-parameter handling.
- Remove obsolete try-catch tests and unused imports.

### Note

This is an alternative to #3221 with a simpler approach: nullable Kotlin overrides instead of a `FixedWindowCallback` Java wrapper class. No new files, no API surface change, no inheritance change.

### DIFF
--- a/features/dd-sdk-android-session-replay/build.gradle.kts
+++ b/features/dd-sdk-android-session-replay/build.gradle.kts
@@ -50,7 +50,6 @@ android {
 dependencies {
     api(project(":dd-sdk-android-core"))
     implementation(project(":dd-sdk-android-internal"))
-    compileOnly(project(":features:dd-sdk-android-rum"))
     implementation(libs.okHttp)
     implementation(libs.kotlin)
     implementation(libs.gson)
@@ -58,7 +57,6 @@ dependencies {
 
     ksp(project(":tools:noopfactory"))
 
-    testImplementation(project(":features:dd-sdk-android-rum"))
     testImplementation(project(":tools:unit")) {
         attributes {
             attribute(
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
@@ -82,7 +82,13 @@ internal class RecorderWindowCallback(
             )
         }
 
-        return super.dispatchTouchEvent(event)
+        @Suppress("SwallowedException")
+        return try {
+            super.dispatchTouchEvent(event)
+        } catch (e: NullPointerException) {
+            logOrRethrowWrappedCallbackException(e)
+            EVENT_CONSUMED
+        }
     }
 
     override fun onWindowFocusChanged(hasFocus: Boolean) {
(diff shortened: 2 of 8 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
ambushwork: The approach of the PR is correct, but we have already fixed a similar issue in this PR:
https://github.com/DataDog/dd-sdk-android/pull/2800

So here we just need to reuse the one we've created before like this:

https://github.com/DataDog/dd-sdk-android/pull/2800/changes#diff-bc13752e81f96b20f33e42bb3183c993c6c0c70cfb715a46c4c084a232f4cd76R36

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
```diff
@@ -81,13 +82,22 @@ internal class RecorderWindowCallback(
             )
         }
 
-        @Suppress("SwallowedException")
-        return try {
-            wrappedCallback.dispatchTouchEvent(event)
-        } catch (e: NullPointerException) {
-            logOrRethrowWrappedCallbackException(e)
-            EVENT_CONSUMED
+        return super.dispatchTouchEvent(event)
```
ambushwork: > [!IMPORTANT]
> This try catch is to not swallow exception coming from wrapped Window.Callback, this logic should not be changed in this PR.

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/startup/AppStartupActivityPredicateTest.kt
ambushwork, line 46: I am not sure the change in this file is related to this PR, if not, we need to revert it.

features/dd-sdk-android-session-replay/build.gradle.kts
```diff
@@ -50,13 +50,15 @@ android {
 dependencies {
     api(project(":dd-sdk-android-core"))
     implementation(project(":dd-sdk-android-internal"))
+    compileOnly(project(":features:dd-sdk-android-rum"))
```
ambushwork: > [!NOTE]
> It's better just to move `FixedWindowCallback` into `dd-sdk-android-internal` module (suggested package: "com.datadog.android.internal.utils") instead of letting session replay depend on the whole rum module.

PR #3240, opened 2026-03-06
Update profiling telemetry with more data

ambushwork:
### What does this PR do?

* This PR updates the profiling telemetry format
  * Message from "[Mobile Metric] App Launch" to  "[Mobile Metric] Profiling Session"
  * Add "start_reason" attribute: currently "application_launch", "rum_operation"/"continuous" in the future
  * Add "app_start_info" attribute for application launch profiling
* Update `DdProfilingContentProvider` to be more testable
* Add `DdProfilingContentProvider` Test

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
@@ -16,44 +16,65 @@ import android.net.Uri
 import android.os.Build
 import androidx.annotation.RequiresApi
 import com.datadog.android.core.sampling.RateBasedSampler
+import com.datadog.android.internal.system.BuildSdkVersionProvider
+import com.datadog.android.internal.system.BuildSdkVersionProvider.Companion.DEFAULT
+import com.datadog.android.profiling.internal.ProfilingStartReason
 import com.datadog.android.profiling.internal.ProfilingStorage
+import com.datadog.android.profiling.internal.perfetto.PerfettoProfiler
 
 /**
  * A [ContentProvider] to start Profiling request as early as possible in the app's
  * lifecycle.
  */
 @OptIn(ExperimentalProfilingApi::class)
-class DdProfilingContentProvider : ContentProvider() {
+class DdProfilingContentProvider(
+    private val buildSdkVersionProvider: BuildSdkVersionProvider = DEFAULT
+) : ContentProvider() {
 
     override fun onCreate(): Boolean {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
-            context?.let {
-                val instanceNames = ProfilingStorage.getProfilingEnabledInstanceNames(it)
-                if (instanceNames.isNotEmpty() && isProcessFromLauncher()) {
-                    sampleProfiling(it, instanceNames)
-                }
+        context?.let { onStart(it) }
+        return true
+    }
+
+    internal fun onStart(context: Context) {
+        if (buildSdkVersionProvider.isAtLeastVanillaIceCream) {
+            val instanceNames = ProfilingStorage.getProfilingEnabledInstanceNames(context)
+            if (instanceNames.isNotEmpty()) {
+                sampleProfiling(context, instanceNames)
             }
         }
-        return true
     }
 
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
     private fun sampleProfiling(context: Context, instanceNames: Set<String>) {
+        val appStartInfo = getAppStartInfo(context) ?: return
         val sampleRate = ProfilingStorage.getSampleRate(context)
         if (RateBasedSampler<Unit>(sampleRate).sample(Unit)) {
-            Profiling.start(context, instanceNames)
+            Profiling.start(
+                context = context,
+                startReason = ProfilingStartReason.APPLICATION_LAUNCH,
+                additionalAttributes = mapOf(PerfettoProfiler.TELEMETRY_KEY_APP_START_INFO to appStartInfo),
+                sdkInstanceNames = instanceNames
+            )
         }
         ProfilingStorage.removeSampleRate(context)
     }
 
+    /**
+     * Returns the Android [ApplicationStartInfo] start reason as a telemetry string if this
+     * process was started from an eligible launcher reason, or null if profiling should not run.
+     */
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
-    private fun isProcessFromLauncher(): Boolean {
-        val manager = context?.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager
+    private fun getAppStartInfo(context: Context): String? {
+        val manager = context.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager
         val startReason = manager?.getHistoricalProcessStartReasons(1)
             ?.firstOrNull()?.reason
-        // TODO RUM-14061: Add all the necessary start reason for application profiling
-        return startReason == ApplicationStartInfo.START_REASON_LAUNCHER ||
-            startReason == ApplicationStartInfo.START_REASON_START_ACTIVITY
+        return when (startReason) {
+            ApplicationStartInfo.START_REASON_LAUNCHER -> TELEMETRY_APP_START_INFO_LAUNCHER
+            ApplicationStartInfo.START_REASON_START_ACTIVITY -> TELEMETRY_APP_START_INFO_ACTIVITY
+            ApplicationStartInfo.START_REASON_LAUNCHER_RECENTS -> TELEMETRY_APP_START_INFO_RECENTS
+            else -> null
+        }
     }
 
     override fun query(
@@ -85,5 +106,11 @@ class DdProfilingContentProvider : ContentProvider() {
         selectionArgs: Array<out String>?
     ): Int {
         return 0
+    }
+
+    internal companion object {
+        internal const val TELEMETRY_APP_START_INFO_LAUNCHER = "launcher"
+        internal const val TELEMETRY_APP_START_INFO_ACTIVITY = "start_activity"
+        internal const val TELEMETRY_APP_START_INFO_RECENTS = "recents"
     }
 }
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStartReason.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStartReason.kt
@@ -1,0 +1,18 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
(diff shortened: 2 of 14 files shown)


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
```diff
@@ -16,44 +16,65 @@ import android.net.Uri
...
     @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
     private fun sampleProfiling(context: Context, instanceNames: Set<String>) {
+        val appStartInfo = getAppStartInfo(context) ?: return
         val sampleRate = ProfilingStorage.getSampleRate(context)
         if (RateBasedSampler<Unit>(sampleRate).sample(Unit)) {
-            Profiling.start(context, instanceNames)
+            Profiling.start(
+                context = context,
+                startReason = ProfilingStartReason.APPLICATION_LAUNCH,
+                additionalAttributes = mapOf(PerfettoProfiler.TELEMETRY_KEY_APP_START_INFO to appStartInfo),
```
0xnm, line 56: Why do we want to pass it as `additionalAttributes` instead of an explicit object? Do we foresee any other attributes?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStartReason.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.profiling.internal
+
+internal enum class ProfilingStartReason(val value: String) {
+
+    APPLICATION_LAUNCH("application_launch"),
```
0xnm, line 11: Probably it will just go away with having continuous profiling? Or you see continuous profiling which is started not at the application launch, but at some point later?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -49,6 +52,12 @@ internal class PerfettoProfiler(
     @Volatile
     private var profilingStartTime = 0L
 
+    @Volatile
+    private var profilingStartReason: String = ""
+
+    @Volatile
+    private var profilingAppStartInfo: String? = null
```
0xnm, line 59: nit: it is better to use type-system at this point still, referencing enums and convert to `String` which is a generic type only when it is really needed, at the very last stage.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/DdProfilingContentProviderTest.kt
```diff
@@ -0,0 +1,211 @@
...
+        // When
+        testedProvider.onStart(mockContext)
+
+        // Then
+        verify(mockProfiler).start(
+            eq(mockContext),
+            eq(ProfilingStartReason.APPLICATION_LAUNCH),
+            eq(mapOf(TELEMETRY_KEY_APP_START_INFO to expectedTelemetryValue)),
+            eq(setOf(fakeInstanceName))
+        )
```
0xnm, line 124:
```suggestion
        verify(mockProfiler).start(
            mockContext,
            ProfilingStartReason.APPLICATION_LAUNCH,
            mapOf(TELEMETRY_KEY_APP_START_INFO to expectedTelemetryValue),
            setOf(fakeInstanceName)
        )
```

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/DdProfilingContentProviderTest.kt
```diff
@@ -0,0 +1,211 @@
...
+    @MethodSource("ineligibleStartReasons")
+    fun `M not start profiling W onCreate { ineligible start reason }`(startReason: Int) {
+        // Given
+        whenever(mockApplicationStartInfo.reason).doReturn(startReason)
+
+        // When
+        testedProvider.onStart(mockContext)
+
+        // Then
+        verify(mockProfiler, never()).start(any(), any(), any(), any())
```
0xnm: seems like `verifyNoInteractions(mockProfiler)` will be simpler here

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/DdProfilingContentProviderTest.kt
```diff
@@ -0,0 +1,211 @@
...
+                eq(ProfilingStorage.KEY_PROFILING_SAMPLE_RATE),
+                any()
+            )
+        ).doReturn(0f)
+
+        // When
+        testedProvider.onStart(mockContext)
+
+        // Then
+        verify(mockProfiler, never()).start(any(), any(), any(), any())
```
0xnm: same: probably `verifyNoInteractions` will work?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -195,12 +198,15 @@ class PerfettoProfilerTest {
...
+            "metric_type" to "profiling session",
+            "profiling_session" to mapOf(
                 "error_code" to ProfilingResult.ERROR_NONE,
                 "error_message" to fakeErrorMessage,
+                "start_reason" to ProfilingStartReason.APPLICATION_LAUNCH.value,
                 "duration" to fakeDuration,
                 "file_size" to 0L,
-                "stopped_reason" to "timeout"
+                "stopped_reason" to "timeout",
+                "app_start_info" to null
```
0xnm, line 208: nit: maybe `app_start_origin` would make a bit more sense

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
```diff
@@ -595,6 +622,129 @@ class PerfettoProfilerTest {
...
+                eq(expectedProps),
+                eq(MethodCallSamplingRate.ALL.rate),
+                isNull()
+            )
+        assertThat(messageCaptor.firstValue.invoke()).isEqualTo("[Mobile Metric] Profiling Session")
+    }
+
+    companion object {
+        @JvmStatic
+        fun profilingStartReasons(): List<Arguments> = ProfilingStartReason.values().map { Arguments.of(it) }
```
0xnm: probably `MethodSource` can be replaced with `@EnumSource(ProfilingStartReason::class)` since what we are providing is essentially just a `ProfilingStartReason` enum.

0xnm (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/DdProfilingContentProvider.kt
ambushwork: I don't see any other attributes for now, but yes the idea is to make it extensible in case that we have any other things to add.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStartReason.kt
ambushwork: I don't think so, app launch profiling will always be there from app launch to TTID, continuous profiling will not start immediately according to the RFC, we will probably add a jitter window.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
ambushwork: here I follow the system api naming `ApplicationStartInfo`

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStartReason.kt
0xnm: Which is probably just a complexity we bring? It is possible to start continuous profiling at the application launch and have this only, given that we label samples with Vital ID/Vital Name anyway.

It seems the only reason to have Application Launch profiling separate is a distinct sampling frequency. But anyway, having it this way brings a complexity if there is feature operation started before TTID.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
0xnm: This telemetry is not unique to Android and its APIs, we can be a bit more precise, maybe even `app_start_reason`?

ambushwork: I checked with @simaoseica-dd , he confirmed that this is not possible to get on IOS, so it's an Android specific field.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/PerfettoProfilerTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/DdProfilingContentProviderTest.kt
ambushwork: fixed

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/DdProfilingContentProviderTest.kt
ambushwork: fixed

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingStartReason.kt
ambushwork: Imagine we always start continuous profiling at the application launch and we try to extend this first profiling to maximum duration (let's say 1 min due to the payload size limitation) and then cooldown for 5 minutes, then the events happen during 1min ~  6min is very likely to be missed.

We can defer this discussion in the RFC, but currently since continuous is not implemented yet, I think it still make sense to keep it here.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
ambushwork: updated `profilingStartReason` to enum type, added an `UNKNOWN` state in the enum as default.

0xnm (approved):

PR #3243, opened 2026-03-06
RUM-14903: Add basic AI dev tooling

aleksandr-gringauz:
Adding some basic files for AI assisted development.
1. `AGENTS.md` + `CLAUDE.md`. Some concise information about how to build and work with the project. This file was produced during a discussion with Claude where I asked it to write this file and only include those things that it finds useful to be there so that these things can help it work better next time. I don't have any way to measure if this content will actually help or not.
2. Introduced `.mcp.json` file that is understood by both Claude and Cursor.
3. Added [datadog mcp](https://docs.datadoghq.com/bits_ai/mcp_server/). Both the core tool set and some other internal tool sets. Very useful to investigate internal telemetry logs and errors. Requires extra sign in.
4. Added [mobile mcp](https://github.com/mobile-next/mobile-mcp). It can interact with the apps on your phone or emulator. Useful for example when testing some code changes in the sample app using Cursor or Claude. It can launch the app and click on things automatically.
5. Added [settings.json](https://code.claude.com/docs/en/settings#configuration-scopes) file for claude that contains some [permission settings](https://code.claude.com/docs/en/permissions#permission-system) for this project.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,6 +9,7 @@ The root `build.gradle.kts` registers aggregate tasks that run across all submod
 ./gradlew unitTestRelease            # unit tests, all modules, release variant
 ./gradlew lintCheckAll               # lint all modules
 ./gradlew checkAll                   # lint + unit tests + instrumented tests
+./gradlew instrumentTestAll          # all instrumented (integration) tests — requires connected device/emulator
 ```
 
 
@@ -32,6 +33,24 @@ Every publishable module has `api/apiSurface` and `api/<module>.api` checked int
 ```
 
 Or for all modules: `./gradlew checkApiSurfaceChangesAll` (verifies) vs. running the update tasks per-module.
+
+# Integration Tests
+
+Integration tests live under `instrumented/integration` and `reliability/` and run on a connected device or emulator.
+
+```bash
+./gradlew instrumentTestAll                                        # all integration tests
+./gradlew :instrumented:integration:connectedDebugAndroidTest     # legacy integration suite
+./gradlew :reliability:core-it:connectedDebugAndroidTest          # core integration tests
+```
+
+Individual reliability modules (each requires a connected device):
+```bash
+./gradlew :reliability:single-fit:logs:connectedDebugAndroidTest
+./gradlew :reliability:single-fit:rum:connectedDebugAndroidTest
+./gradlew :reliability:single-fit:trace:connectedDebugAndroidTest
+./gradlew :reliability:single-fit:okhttp:connectedDebugAndroidTest
+```
 
 # Testing Conventions
 
@@ -97,3 +116,9 @@ Some modules generate Kotlin data classes from JSON schemas at build time (e.g. 
 
 - Default branch for PRs: **`develop`**
 - All commits must be **GPG-signed** (repo policy).
+- Commit title format: `RUM-XXXXX: <short description>` (ticket number prefix required)
+- Each commit must reference the ticket number in the title.
+
+# Pull Requests
+
+- Use the project PR template (`.github/PULL_REQUEST_TEMPLATE.md`). 


.claude/settings.json
```diff
@@ -0,0 +1,22 @@
...
+      "Bash(xargs rg:*)",
+      "Bash(grep:*)",
+      "Bash(sed:*)",
+      "Bash(awk:*)",
+      "Bash(cat:*)",
+      "Bash(git fetch:*)",
+      "Bash(git status:*)",
+      "Bash(git log:*)",
+      "Bash(git diff:*)",
+      "Bash(git show:*)",
```
aleksandr-gringauz, line 17: These commands don't do anything dangerous, most of them just read things, so I think it is safe to allow Claude to execute them without explicitly asking for permission.

.claude/settings.json
```diff
@@ -0,0 +1,22 @@
...
+      "Bash(ktlint:*)",
+      "Bash(./gradlew *)"
```
aleksandr-gringauz, line 19: These are android specific commands that are used to build code.

.claude/settings.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://json.schemastore.org/claude-code-settings.json",
+  "permissions": {
+    "allow": [
```
aleksandr-gringauz, line 4: I decided not to include:
1. Any permissions related to writing.
2. Permissions from datadog mcp and mobile mcp.

Datadog MCP for example has some tools that change things. Some people might want to better control what tools Claude executes and explicitly grant permissions.

We can change it at any time.

.mcp.json
```diff
@@ -0,0 +1,12 @@
+{
+  "mcpServers": {
+    "mobile-mcp": {
+      "command": "npx",
+      "args": ["-y", "@mobilenext/mobile-mcp@latest"]
```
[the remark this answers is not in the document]
aleksandr-gringauz: it's just an mcp, not a dependency, so that's fine

.claude/settings.json
ambushwork: I would say this one is less safer than others:
* it can be against what you commented above - it can write, modify and delete stuff such as compiling, cleaning ,etc.. (which is ok-ish though)
* it basically allows all kinds of gradle task that we don't know if they are safe or not.

AGENTS.md
```diff
@@ -0,0 +1,99 @@
...
+Or for all modules: `./gradlew checkApiSurfaceChangesAll` (verifies) vs. running the update tasks per-module.
+
+# Testing Conventions
+
+Test class boilerplate (JUnit5 + Mockito + Elmyr):
+
+```kotlin
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
```
ambushwork, line 62: would it applies the rule even if some tests don't event need forge or mock?

AGENTS.md
```diff
@@ -0,0 +1,99 @@
...
+Filename matches the flavor: `config/us1.json`, `config/staging.json`, etc. Get credentials from your Datadog org.
+
+# Generated Models
+
+Some modules generate Kotlin data classes from JSON schemas at build time (e.g. `features/dd-sdk-android-rum/src/main/json/`). The generated Kotlin files land in `build/generated/json2kotlin/` — **do not edit them directly**. Edit the JSON schemas instead and rebuild.
+
+# Commits
+
+- Default branch for PRs: **`develop`**
+- All commits must be **GPG-signed** (repo policy).
```
hamorillo, line 118: ❓ Would it make sense to add some details about what to include in the PR body (based on our template) and/or in the commit comments? (e.g., adding the ticket number to the commit title)

.claude/settings.json
aleksandr-gringauz: Yes, you are correct. I should have been more explicit about it. Executing a random gradle command  can definitely delete and edit things. In my comment I was referring to the Bash commands.

However I still think that including `./gradlew` here is fine in the context of Android SDK development.

AGENTS.md
aleksandr-gringauz: It depends on what the agent decides to do. I think it should be able to understand that it doesn't need to include certain extensions if they aren't used in the test. But we'll see. If it works incorrectly we can always remove this.

AGENTS.md
aleksandr-gringauz: done 👍

PR #3244, opened 2026-03-09
Fix #2374: Make getResourceName() consistent with hasResourceName()

### LINKED ISSUE #2374: Span resource name set with tag is not used
### Describe the bug

Using a tag to set the resource name does not work and instead shows up as the operation name in the datadog ui. This comes from a bug in this library and even has a TODO comment to resolve it. In DDSpanContext `getResourceName` does not read the "resource.name" set in the tags, but it does read it when returning a value for `hasResourceName`

https://github.com/DataDog/dd-sdk-android/blob/b1d6ebb82d26299faa41195747198850dcef090d/features/dd-sdk-android-trace/src/main/java/com/datadog/trace/core/DDSpanContext.java#L404

### Reproduction steps

```
import io.opentracing.util.GlobalTracer

val spanBuilder = GlobalTracer.get().build("operation_name")
spanBuilder.withTag("resource.name", "custom resource name")
spanBuilder.start().finish()
```

### Logcat logs

_No response_

### Expected behavior

_No response_

### Affected SDK versions

2.14.0

never

Yes

### Kotlin / Java version

_No response_

### Gradle / AGP version

_No response_

_No response_

### Device Information

_No response_

_No response_

mariusc83:
## Issue

Fixes #2374: Span resource name set with tag is not used

**GitHub Issue:** https://github.com/DataDog/dd-sdk-android/issues/2374

## Root Cause Analysis

`DDSpanContext.getResourceName()` only checked the internal `resourceName` field, falling back to `operationName`. Unlike `hasResourceName()`, it did not check the `resource.name` tag in `unsafeTags`. This caused an inconsistency where `hasResourceName()` returned `true` but `getResourceName()` returned the operation name instead of the tag-set resource name — specifically when the `resource.name` tag was stored directly in the tags map (e.g., when the RESOURCE_NAME rule flag in TagInterceptor is disabled).

A TODO comment at the method explicitly flagged this inconsistency.

## Changes

Make `getResourceName()` check `getTag(DDTags.RESOURCE_NAME)` as a fallback before returning `operationName`, matching the logic in `hasResourceName()`. Remove the TODO comment.

### Files Changed

| File | Change |
|------|--------|
| `DDSpanContext.java` | Fix `getResourceName()` to check tag fallback |
| `DDSpanContextTest.kt` | Add 5 tests (2 unit, 2 integration, 1 sampling rule) |

## Impact Analysis

An architectural impact analysis was performed after the initial fix to evaluate the blast radius:

### Caller Map

| Caller | Module | Usage | Risk |
|--------|--------|-------|------|
| `SamplingRule.TraceSamplingRule.matches()` | trace-internal | Pattern matching for sampling decisions | Covered by new test |
| `CoreTracerSpanToSpanEventMapper` | trace | Serializes to `SpanEvent.resource` | Covered by integration test |
| `TraceStructureWriter.Node` | trace-internal | Debug output | Low risk |
| `DatadogSpanAdapter` | trace | Public API property | Low risk — delegates to DDSpan |

All production callers (network instrumentation, OkHttp interceptor) explicitly **write** to `.resourceName` before reading, so the fallback path is only hit when no explicit resource name is set.

### Thread Safety

The fix adds a `getTag()` call (which takes `synchronized(unsafeTags)`) in the fallback path. The old code was lock-free. Risk is **low** because:
- `getResourceName()` is called once per span at serialization time, not in loops
- Span lifecycle is typically single-threaded
- The fallback path is only hit when `resourceName` field is not set

### Consistency Scan

Found a latent similar pattern in `getServiceName()` (only checks internal field, same RuleFlags conditional processing). Not a bug today since no `hasServiceName()` method exists, but noted for future awareness.

## Test Coverage

- **Unit tests:** 2 (tag fallback behavior, field-over-tag priority)
- **Integration tests:** 2 (builder `withTag` flow, `setTag` after start)
- **Impact-driven test:** 1 (sampling rule matches resource name from tag fallback)
- **Module test suite:** PASS (trace-internal, trace, trace-otel all green)
- **TDD iterations:** 1/5

### DIFF
--- a/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/DDSpanContext.java
+++ b/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/DDSpanContext.java
@@ -400,9 +400,17 @@ public class DDSpanContext
     this.topLevel = isTopLevel(parentServiceName, this.serviceName);
   }
 
-  // TODO this logic is inconsistent with hasResourceName
   public CharSequence getResourceName() {
-    return isResourceNameSet() ? resourceName : operationName;
+    if (isResourceNameSet()) {
+      return resourceName;
+    }
+    Object tagValue = getTag(DDTags.RESOURCE_NAME);
+    if (tagValue instanceof CharSequence) {
+      return (CharSequence) tagValue;
+    } else if (tagValue != null) {
+      return String.valueOf(tagValue);
+    }
+    return operationName;
   }
 
   public boolean hasResourceName() {
--- a/features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/core/DDSpanContextTest.kt
+++ b/features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/core/DDSpanContextTest.kt
@@ -315,6 +315,45 @@ internal class DDSpanContextTest : DDCoreSpecification() {
         span.finish()
     }
 
+    @Test
+    fun `getResourceName returns correct value when resource name set via builder withTag`() {
+        // Given: a span where resource.name is set via builder withTag (exact user scenario from #2374)
+        // This exercises the full chain: builder.withTag → setAllTags → tagInterceptor → setResourceName
+        val span = tracer.buildSpan(instrumentationName, "fakeOperation")
+            .withServiceName("fakeService")
+            .withTag(DDTags.RESOURCE_NAME, "custom-resource-from-builder-tag")
+            .start()
+        val context = span.context() as DDSpanContext
+
+        // Then: resource name should reflect the tag value, not fall back to operation name
+        assertThat(context.hasResourceName()).isTrue()
+        assertThat(context.resourceName.toString()).isEqualTo("custom-resource-from-builder-tag")
+
+        // Tear down
+        span.finish()
+        writer.waitForTraces(1)
+    }
+
+    @Test
+    fun `getResourceName returns correct value when resource name set via setTag after start`() {
+        // Given: a span started without a resource name
+        val span = tracer.buildSpan(instrumentationName, "fakeOperation")
+            .withServiceName("fakeService")
+            .start()
+        val context = span.context() as DDSpanContext
+
+        // When: resource.name is set via setTag after span start
+        context.setTag(DDTags.RESOURCE_NAME, "resource-set-after-start")
+
+        // Then: resource name should reflect the tag value
+        assertThat(context.hasResourceName()).isTrue()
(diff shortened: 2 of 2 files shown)


features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/core/DDSpanContext.java
```diff
@@ -400,9 +400,17 @@ public void setServiceName(final String serviceName) {
     this.topLevel = isTopLevel(parentServiceName, this.serviceName);
   }
 
-  // TODO this logic is inconsistent with hasResourceName
   public CharSequence getResourceName() {
-    return isResourceNameSet() ? resourceName : operationName;
+    if (isResourceNameSet()) {
```
0xnm, line 404: Changing the behavior of existing method can skew existing dashboards/monitors. I'm not sure we should do this for the minor release.

mariusc83: Totally agree and that was my worry too. Actually Claude picked this up also during the fixing session and mentioned it. That's a decision on your end. Choose to keep it as it is or take the risk and merge it. This is a hard one as there's definitely a problem there of inconsistency which can have repercussions later. Either you take the bullet now and fix it or you postpone it and face it later.

satween: Are you sure this change ever needed? According to `TagInterceptor` logic it would call `setResource` in case if customer tries to set `DDTags.RESOURCE_NAME`. So Basically seems like the opposite logic might make more sense - removing `hasTag(DDTags.RESOURCE_NAME)` from the `hasResourceName` method. WDYT?

mariusc83: I think you have more context than me on this one but from what I see it makes sense what you are saying. But not matter what we decide is going to change the API functionality so it will be some sort of regression. In the same time if we decide to keep it as it is, it means that we acknowledge that there's an issue and we accept it. I let you guys decide and I will apply the changes if needed.

satween: My opinion - is to add TODO comment, with RUM-13454 and link to this PR and close for now. With the next major release - we gonna fix this

mariusc83: Done — I've opened a separate PR (#3252) that updates the TODO comment with RUM-13454 and a link to this PR, deferring the fix to the next major release as suggested. Will close this PR.

PR #3247, opened 2026-03-11
RUM-13023: Propagate trace and session replay sample rates to RUM ViewEvents

hamorillo:
### What does this PR do?

Adds `traceSampleRate` and `sessionReplaySampleRate` fields to `ViewEvent.Configuration`, populating them from the tracing and session replay feature contexts when a RUM view event is written.

### Motivation

RUM-13023: These sample rates were not previously included in `ViewEvent` payloads, making it impossible for downstream consumers to know the effective sampling configuration at the time of the event. This change reads the rates from the shared `featuresContext` map and propagates them into the event schema.

### Additional Notes

- `sessionReplaySampleRate` is read as `Long` (matching `SessionReplayFeature` storage format).
- `traceSampleRate` is read as `Float` (matching `TracingInterceptor` storage format).
- `Feature.TRACING_FEATURE_NAME` was added to the `withFeatureContexts` set in `DatadogRumMonitor` so the tracing context is available when writing view events.
- The JSON schema (`_common-schema.json`) was updated with the new `trace_sample_rate` field; the generated model and API surface files were regenerated accordingly.

### How to test

To manually verify that `session_replay_sample_rate` and `trace_sample_rate` are present in RUM ViewEvent payloads, enable request body logging in the sample app and inspect logcat.

**1. Enable body logging in `CoreFeature.kt`**

In `dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt`, change:

```kotlin
// Before
builder.addNetworkInterceptor(CurlInterceptor())

// After
builder.addNetworkInterceptor(CurlInterceptor(printBody = true))
```

> This is gated by `BuildConfig.DEBUG` — safe for debug builds only. Do not commit this change.

**2. Build and install the sample app**

```sh
./gradlew :sample:kotlin:installUs1Debug
adb shell am start -n com.datadog.android.sample/.NavActivity
```

**3. Navigate between screens and check logcat**

Navigate to any screen and back to trigger a view stop event, then wait ~15s for the upload cycle:

```sh
adb logcat -s "Curl" | grep '"type":"view"'
```

Look for `_dd.configuration` in the output — both fields should be present:

```json
"configuration": {
  "session_sample_rate": 100.0,
  "session_replay_sample_rate": 100,
  "trace_sample_rate": 100.0
}
```

> ⚠️📦 A skill for this workflow is available in `.claude/skills/android-sdk-event-inspection/` — it covers the full setup including credentials and logcat filtering.

- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc). Be concise.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
+++ b/features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
@@ -435,6 +435,13 @@
               "minimum": 0,
               "maximum": 100,
               "readOnly": true
+            },
+            "trace_sample_rate": {
+              "type": "number",
+              "description": "The percentage of sessions with traced resources",
+              "minimum": 0,
+              "maximum": 100,
+              "readOnly": true
             }
           }
         },
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -1343,7 +1343,11 @@ internal open class RumViewScope(
                         sessionPrecondition = rumContext.sessionStartReason.toViewSessionPrecondition()
                     ),
                     replayStats = replayStats,
-                    configuration = ViewEvent.Configuration(sessionSampleRate = sampleRate)
+                    configuration = ViewEvent.Configuration(
+                        sessionSampleRate = sampleRate,
+                        sessionReplaySampleRate = resolveSessionReplaySampleRate(datadogContext),
+                        traceSampleRate = resolveTraceSampleRate(datadogContext)
+                    )
                 ),
                 connectivity = datadogContext.networkInfo.toViewConnectivity(),
                 service = datadogContext.service,
@@ -1609,6 +1613,16 @@ internal open class RumViewScope(
         }
     }
 
+    private fun resolveSessionReplaySampleRate(datadogContext: DatadogContext): Long? {
+        val srContext = datadogContext.featuresContext[Feature.SESSION_REPLAY_FEATURE_NAME]
+        return srContext?.get(SESSION_REPLAY_SAMPLE_RATE_KEY) as? Long
+    }
+
+    private fun resolveTraceSampleRate(datadogContext: DatadogContext): Float? {
+        val tracingContext = datadogContext.featuresContext[Feature.TRACING_FEATURE_NAME]
+        return tracingContext?.get(TRACE_SAMPLE_RATE) as? Float
+    }
+
     private fun logSynthetics(key: String, value: String) {
         /**
          * We use [android.util.Log] here instead of [InternalLogger] because we want to log regardless of the
@@ -1621,6 +1635,12 @@ internal open class RumViewScope(
 
     companion object {
         internal val ONE_SECOND_NS = TimeUnit.SECONDS.toNanos(1)
+
+        // Must match SessionReplayFeature.SESSION_REPLAY_SAMPLE_RATE_KEY in dd-sdk-android-session-replay
+        internal const val SESSION_REPLAY_SAMPLE_RATE_KEY = "session_replay_sample_rate"
+
+        // Must match TracingInterceptor.OKHTTP_INTERCEPTOR_SAMPLE_RATE in dd-sdk-android-okhttp
+        internal const val TRACE_SAMPLE_RATE = "okhttp_interceptor_sample_rate"
 
         internal const val ACTION_DROPPED_WARNING = "RUM Action (%s on %s) was dropped, because" +
             " another action is still active for the same view"
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedViewEventAssert.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedViewEventAssert.kt
@@ -21,8 +21,9 @@ internal class DeserializedViewEventAssert(actual: ViewEvent) :
     fun isEqualTo(expected: ViewEvent): DeserializedViewEventAssert {
         assertThat(actual)
             .usingRecursiveComparison()
-            .ignoringFields("context", "usr", "account", "view", "device")
+            .ignoringFields("context", "usr", "account", "view", "device", "dd.configuration")
             .isEqualTo(expected)
+        assertConfigurationEquals(actual.dd.configuration, expected.dd.configuration)
         assertThat(actual.view)
             .usingRecursiveComparison()
             .ignoringFields(
@@ -77,6 +78,23 @@ internal class DeserializedViewEventAssert(actual: ViewEvent) :
             expected.context?.additionalProperties
         )
         return this
+    }
+
+    private fun assertConfigurationEquals(
+        actual: ViewEvent.Configuration?,
+        expected: ViewEvent.Configuration?
+    ) {
+        if (expected == null) {
+            assertThat(actual).isNull()
+            return
+        }
+        assertThat(actual).isNotNull()
+        assertNumberFieldEquals(actual!!.sessionSampleRate, expected.sessionSampleRate)
+        assertNumberFieldEquals(actual.sessionReplaySampleRate, expected.sessionReplaySampleRate)
+        assertNumberFieldEquals(actual.profilingSampleRate, expected.profilingSampleRate)
+        assertNumberFieldEquals(actual.traceSampleRate, expected.traceSampleRate)
+        assertThat(actual.startSessionReplayRecordingManually)
+            .isEqualTo(expected.startSessionReplayRecordingManually)
     }
 
     private fun assertPropertiesEquals(actual: Map<String, Any?>?, expected: Map<String, Any?>?) {
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedViewEventAssert.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedViewEventAssert.kt
@@ -88,8 +88,8 @@ internal class DeserializedViewEventAssert(actual: ViewEvent) :
             assertThat(actual).isNull()
             return
         }
-        assertThat(actual).isNotNull()
-        assertNumberFieldEquals(actual!!.sessionSampleRate, expected.sessionSampleRate)
+        checkNotNull(actual)
+        assertNumberFieldEquals(actual.sessionSampleRate, expected.sessionSampleRate)
         assertNumberFieldEquals(actual.sessionReplaySampleRate, expected.sessionReplaySampleRate)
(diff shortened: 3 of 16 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1609,6 +1613,16 @@ internal open class RumViewScope(
...
     }
 
+    private fun resolveSessionReplaySampleRate(datadogContext: DatadogContext): Long? {
+        val srContext = datadogContext.featuresContext[Feature.SESSION_REPLAY_FEATURE_NAME]
+        return srContext?.get(SESSION_REPLAY_SAMPLE_RATE_KEY) as? Long
+    }
+
+    private fun resolveTraceSampleRate(datadogContext: DatadogContext): Float? {
+        val tracingContext = datadogContext.featuresContext[Feature.TRACING_FEATURE_NAME]
+        return tracingContext?.get(TRACE_SAMPLE_RATE) as? Float
```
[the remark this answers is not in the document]
hamorillo: Commit `203d11f5` addresses this by moving the context update into `onSdkInstanceReady`, so it fires regardless of whether the interceptor is constructed before or after `Datadog.initialize()`.

There is still a narrow edge case: if the interceptor is created before `Datadog.initialize()`, the first RUM view event emitted **before the first call to `intercept()`** may still miss `trace_sample_rate`, because `onSdkInstanceReady` is triggered lazily on the first interception. Once `intercept()` is called for the first time the context is populated and subsequent view events will carry the field correctly.

I've been thinking about how to fully close this gap, but haven't found a solution that doesn't require modifying the public API. I think the current state is an acceptable trade-off for now — but happy to hear other opinions from anyone with more context on this!

sample/vendor-lib/src/main/kotlin/com/datadog/android/vendor/sample/LocalServer.kt
```diff
@@ -80,7 +80,7 @@ public class LocalServer {
         Logs.enable(logsConfig, instance)
 
         val tracesConfig = TraceConfiguration.Builder().build()
-        Trace.enable(tracesConfig)
+        Trace.enable(tracesConfig, instance)
```
hamorillo, line 83: 💡 This was a previous bug as it should use the Datadog instance created in the LocalServer

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -105,17 +105,9 @@ internal constructor(
...
-            it[OKHTTP_INTERCEPTOR_SAMPLE_RATE] = traceSampler.getSampleRate()
-            it[OKHTTP_INTERCEPTOR_HEADER_TYPES] = TracingHeaderTypesSet(
-                tracedHosts.values.flatten()
-                    .map(TracingHeaderType::toInternalTracingHeaderType)
-                    .toSet()
-            )
-        }
+        // Trigger SDK instance acquisition. If the SDK is already initialized, this will invoke
+        // [onSdkInstanceReady] synchronously, where we populate the tracing feature context.
+        sdkCoreReference.get()
```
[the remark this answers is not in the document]
hamorillo: See https://github.com/DataDog/dd-sdk-android/pull/3247#discussion_r2923811434

sample/vendor-lib/src/main/kotlin/com/datadog/android/vendor/sample/LocalServer.kt
0xnm: nice catch!

sample/kotlin/src/main/kotlin/com/datadog/android/sample/SampleApplication.kt
```diff
@@ -97,40 +97,44 @@ class SampleApplication : Application() {
...
-                )
-                .build()
-        )
-        .addNetworkInterceptor(
-            TracingInterceptor.Builder(tracedHosts)
-                .build()
-        )
-        .eventListenerFactory(DatadogEventListener.Factory())
-        .build()
+    private val okHttpClient by lazy {
```
0xnm: We need to revert these `by lazy` changes here, because the code as it is right now imitates the real scenario - network stack is created before Datadog SDK is initialized.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/utils/assertj/DeserializedViewEventAssert.kt
```diff
@@ -79,6 +80,23 @@ internal class DeserializedViewEventAssert(actual: ViewEvent) :
...
+    private fun assertConfigurationEquals(
+        actual: ViewEvent.Configuration?,
+        expected: ViewEvent.Configuration?
+    ) {
+        if (expected == null) {
+            assertThat(actual).isNull()
+            return
+        }
+        assertThat(actual).isNotNull()
+        assertNumberFieldEquals(actual!!.sessionSampleRate, expected.sessionSampleRate)
```
0xnm:
```suggestion
        checkNotNull(actual)
        assertNumberFieldEquals(actual.sessionSampleRate, expected.sessionSampleRate)
```

features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
```diff
@@ -435,6 +435,13 @@
               "minimum": 0,
               "maximum": 100,
               "readOnly": true
+            },
+            "trace_sample_rate": {
+              "type": "number",
+              "description": "The percentage of sessions with traced resources",
```
0xnm, line 441: From what I see, it is not percentage of the sessions, it is just a sample rate set for the tracing, not linked in anyway to RUM sessions.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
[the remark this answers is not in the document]
aleksandr-gringauz: As an option. We could introduce some global static `HashMap<sdkInstanceName, OkHttpSampleRate>` where we put the value when `TracingInterceptor` is created. The map itself will have to live in `dd-sdk-android-rum` module somewhere so that we could get the value from it there.

There is actually another kind of related problem. Theoretically there could exist multiple instances of `TracingInterceptor` with different sampling rates (our SDK allows it I belive). Although I expect this to be rare or even non-existing in practice. I'm not asking to fix it, just mentioning.

features/dd-sdk-android-rum/src/main/json/rum/_common-schema.json
hamorillo: 🤔 Yeah, you are probably right. I've used the values from  the [schema](https://github.com/DataDog/rum-events-format/blob/834392ddf77531ed3f383e0808192879490c221d/schemas/rum/_common-schema.json#L424) so maybe we need to update it in both places. Wdyt?

aleksandr-gringauz: > not linked in anyway to RUM sessions.

IIUC this sampling rate is exactly the percentage of RUM sessions that have traces. Because this `trace_sample_rate` is used to make a deterministic sampling decision based on RUM session id (see [this](https://github.com/DataDog/dd-sdk-android/blob/b94905c22b8bcefa8ed93b026411938ebff40b92/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/utils/SpanSamplingIdProvider.kt#L26)).

0xnm: Indeed, good point.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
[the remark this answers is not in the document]
hamorillo: > We could introduce some global static HashMap<sdkInstanceName, OkHttpSampleRate> where we put the value when TracingInterceptor is created. The map itself will have to live in dd-sdk-android-rum module somewhere so that we could get the value from it there.

That could be an option. I’m not a big fan of introducing a global static variable, though. (But I didn’t come up with anything better, so thanks!) Adding the map to `dd-sdk-android-rum` would “pollute” the public API a bit, but it looks doable.

What I’m thinking about is whether it’s worth it, as IIUC the backend will keep the last received value. So sessions will save the last received value, which in general should contain a `trace_sample_rate` value.

Anyway, you have deeper knowledge here, so if you think we need to cover this case, I’ll prepare the changes.

>  Theoretically there could exist multiple instances of TracingInterceptor with different sampling rates (our SDK allows it I belive). Although I expect this to be rare or even non-existing in practice. I'm not asking to fix it, just mentioning.

That’s a good point. However, I would say that concern goes further than the scope of this PR, as I’m not sure whether our code has been developed with that consideration in mind.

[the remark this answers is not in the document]
aleksandr-gringauz: > Adding the map to dd-sdk-android-rum would “pollute” the public API a bit, but it looks doable.

You can add `InternalApi` annotation to it. Yes, it will end up in the public API but the linter will not allow the customers to use it.

> Anyway, you have deeper knowledge here, so if you think we need to cover this case, I’ll prepare the changes.

I don't think I actually have deeper knowledge here :) This is the first time I look at this code. But the solution seems doable. 

I'm not a fan of statics either. Maybe we need to have some sort of `PreInitSDKCore` - a place that holds things before `SDKCore` is created. Not a call to action though.

From what I see current architecture doesn't support well what you want.

I would say - if you can come up with a solution - great. If not - I am ok with approving the PR the way it is now. We already had this problem for `OKHTTP_INTERCEPTOR_SAMPLE_RATE`, you didn't introduce any regressions here.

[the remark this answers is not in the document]
0xnm: Let's keep things simple and avoid introducing a singleton.

> the first RUM view event emitted before the first call to intercept() may still miss trace_sample_rate, because onSdkInstanceReady is triggered lazily on the first interception.

RUM view events are reduced, so even if first view update miss it, if the following view updates for the given view populate this field, it will be there overall once view if finalized.

We can check with backend if this property is aggregated.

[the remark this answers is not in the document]
aleksandr-gringauz: > Let's keep things simple and avoid introducing a singleton.

OK from me.

0xnm (approved):

aleksandr-gringauz (approved):

PR #3271, opened 2026-03-18
RUM-2780: Return no-op core when instance is already initialized

ambushwork:
### What does this PR do?

we should return a no-op instance and prevent leaks between orgs.

### Motivation

RUM-2780


dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt
hamorillo, line 51: _nitpick_: Should we add that we can now return `NoOpInternalSdkCore`?

hamorillo (approved):
🚀 LGTM! As you mentioned in the ticket, we need to block merging this PR until v4, though.

jonathanmos (approved):

PR #3275, opened 2026-03-18
RUM-14902: Track GraphQL errors

kikoveiga:
### What does this PR do?

- Track GraphQL errors on `DatadogInterceptor` for `OkHttp` requests.
- Create `GraphQLExtractor` with 2 main functions: `extractGraphQLAttributes` (already existing logic, moved from `DatadogInterceptor` for better separation of concerns) & `extractGraphQLErrors` (new logic). This is a reusable extraction in `rum` module that can be reused in the future with other networking libraries (no coupling with OkHttp).
- Create `OkHttpGraphQLAdapter` to encapsulate some already existing logic and delegate to the class above.
- Some cleanup (create `StringExt` on `internal` module to reuse `String.toBase64` & `String.fromBase64` extension functions).
 
### Motivation

GraphQL APIs return HTTP 200 regardless of succeeding or not. Errors are reported inside the response body's `errors` array rather than through HTTP status codes.

Now these errors are properly caught, allowing customers to detect GraphQL failures that would otherwise go unnoticed.

### Additional notes

iOS counterpart: https://github.com/DataDog/dd-sdk-ios/pull/2552

### DIFF
--- a/dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/ForgeExt.kt
+++ b/dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/ForgeExt.kt
@@ -60,6 +60,7 @@ fun <T : Forge> T.useCoreFactories(): T {
     addFactory(ThreadDumpForgeryFactory())
     addFactory(RequestExecutionContextForgeryFactory())
     addFactory(RequestInfoForgeryFactory())
+    addFactory(MutableRequestInfoForgeryFactory())
 
     return this
 }
@@ -72,7 +73,7 @@ fun Forge.aHostName(): String {
 fun Forge.anUrlString(): String = aStringMatching(URL_FORGERY_PATTERN)
 
 fun Forge.anHttpRequestInfo(headers: Map<String, String>): HttpRequestInfo {
-    return (getForgery<HttpRequestInfo>() as MutableHttpRequestInfo)
+    return getForgery<MutableHttpRequestInfo>()
         .newBuilder()
         .apply { headers.forEach { (key, value) -> addHeader(key, value) } }
         .build()
--- a/dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/ForgeExt.kt
+++ b/dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/ForgeExt.kt
@@ -6,6 +6,8 @@
 
 package com.datadog.android.tests.elmyr
 
+import com.datadog.android.api.instrumentation.network.HttpRequestInfo
+import com.datadog.android.api.instrumentation.network.MutableHttpRequestInfo
 import fr.xgouchet.elmyr.Case
 import fr.xgouchet.elmyr.Forge
 import okhttp3.Protocol
@@ -68,6 +70,13 @@ fun Forge.aHostName(): String {
 }
 
 fun Forge.anUrlString(): String = aStringMatching(URL_FORGERY_PATTERN)
+
+fun Forge.anHttpRequestInfo(headers: Map<String, String>): HttpRequestInfo {
+    return (getForgery<HttpRequestInfo>() as MutableHttpRequestInfo)
+        .newBuilder()
+        .apply { headers.forEach { (key, value) -> addHeader(key, value) } }
+        .build()
+}
 
 fun Forge.anOkHttpResponse(
     request: Request,
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/StringExt.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/StringExt.kt
@@ -1,0 +1,34 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.utils
+
+import android.util.Base64
+
+/**
+ * Encodes this string to Base64 using UTF-8 encoding.
+ *
+ * @return the Base64-encoded representation of this string, without line wrapping.
+ */
+fun String.toBase64(): String {
+    val bytes = this.toByteArray(Charsets.UTF_8)
+    @Suppress("UnsafeThirdPartyFunctionCall") // cannot throw UnsupportedEncodingException
+    return Base64.encodeToString(bytes, Base64.NO_WRAP)
+}
+
+/**
+ * Decodes this Base64-encoded string back to its original UTF-8 representation.
+ *
+ * @return the decoded string, or `null` if the input is not valid Base64.
+ */
+fun String.fromBase64(): String? {
+    return try {
+        val decodedBytes = Base64.decode(this, Base64.NO_WRAP)
+        decodedBytes?.toString(Charsets.UTF_8)
+    } catch (_: IllegalArgumentException) {
+        null
+    }
+}
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/GraphQLExtractor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/GraphQLExtractor.kt
@@ -1,0 +1,116 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.net
+
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.instrumentation.network.HttpRequestInfo
+import com.datadog.android.internal.network.GraphQLHeaders
+import com.datadog.android.internal.utils.fromBase64
+import com.datadog.android.lint.InternalApi
+import com.datadog.android.rum.RumAttributes
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Extracts GraphQL-related attributes and errors from network requests and responses.
+ */
+@InternalApi
+class GraphQLExtractor {
+
+    /**
+     * Extracts GraphQL attributes (operation name, type, variables, payload) from the request headers.
+     * @param requestInfo the HTTP request info containing headers
+     * @return a map of GraphQL attributes decoded from Base64 header values
+     */
+    fun extractGraphQLAttributes(requestInfo: HttpRequestInfo): Map<String, Any?> {
+        return buildMap {
+            GRAPHQL_HEADER_TO_ATTRIBUTE.forEach { (header, attribute) ->
+                requestInfo.headers[header.headerValue]?.firstOrNull()?.let {
+                    put(attribute, it.fromBase64())
+                }
+            }
+        }
+    }
+
+    /**
+     * Extracts and normalizes GraphQL errors from a JSON response body.
+     * @param contentType the content type of the response
+     * @param body the response body string
+     * @param internalLogger the logger for reporting parsing failures
+     * @return a JSON array string of normalized errors, or null if none found
+     */
+    fun extractGraphQLErrors(
+        contentType: String?,
+        body: String?,
+        internalLogger: InternalLogger
+    ): String? {
+        if (body == null || contentType == null || JSON_CONTENT_TYPES.none { contentType.startsWith(it) }) return null
+
+        return try {
+            val json = JSONObject(body)
+            val errorsArray = if (json.has(ERRORS_KEY)) json.getJSONArray(ERRORS_KEY) else null
+
+            if (errorsArray == null || errorsArray.length() == 0) {
+                null
+            } else {
+                val normalizedErrors = JSONArray()
+                for (i in 0 until errorsArray.length()) {
+                    @Suppress("UnsafeThirdPartyFunctionCall") // JSONException is caught and we are in array bounds
+                    val error = errorsArray.getJSONObject(i)
+                    normalizeErrorCode(error)
+                    normalizedErrors.put(error)
+                }
+                normalizedErrors.toString()
+            }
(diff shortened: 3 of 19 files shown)


dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/ForgeExt.kt
```diff
@@ -65,3 +67,10 @@ fun Forge.aHostName(): String {
 }
 
 fun Forge.anUrlString(): String = aStringMatching(URL_FORGERY_PATTERN)
+
+fun Forge.anHttpRequestInfo(headers: Map<String, String>): HttpRequestInfo {
+    return (getForgery<HttpRequestInfo>() as MutableHttpRequestInfo)
```
satween: `getForgery<MutableHttpRequestInfo>()` can be used with corresponding factory support

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/StringExt.kt
```diff
@@ -0,0 +1,34 @@
...
+package com.datadog.android.internal.utils
+
+import android.util.Base64
+
+/**
+ * Encodes this string to Base64 using UTF-8 encoding.
+ *
+ * @return the Base64-encoded representation of this string, without line wrapping.
+ */
+fun String.toBase64(): String {
```
satween, line 16: This will add `toBase64` method for all the customer's code. 
I was thinking of having something like this: 

```kotlin

object DatadogStringUtils{

  fun String.toBase64(): String {
   val bytes = this.toByteArray(Charsets.UTF_8)
      @Suppress("UnsafeThirdPartyFunctionCall") // cannot throw UnsupportedEncodingException
      return Base64.encodeToString(bytes, Base64.NO_WRAP)
  }
  
  fun String.fromBase64(): String? {
      return try {
          val decodedBytes = Base64.decode(this, Base64.NO_WRAP)
          decodedBytes?.toString(Charsets.UTF_8)
      } catch (_: IllegalArgumentException) {
          null
      }
  }
}
```

Then you can call those method like that: 

```kotlin

with(DatadogStringUtils){

    val requestInfo = forge.anHttpRequestInfo(
            mapOf(
                GraphQLHeaders.DD_GRAPHQL_NAME_HEADER.headerValue to fakeGraphQLName.toBase64(),
                GraphQLHeaders.DD_GRAPHQL_TYPE_HEADER.headerValue to fakeGraphQLType.toBase64(),
                GraphQLHeaders.DD_GRAPHQL_VARIABLES_HEADER.headerValue to fakeGraphQLVariables.toBase64(),
                GraphQLHeaders.DD_GRAPHQL_PAYLOAD_HEADER.headerValue to fakeGraphQLPayload.toBase64()
            )
        )
}


```

and without `with(DatadogStringUtils)` these methods wouldn't be publicly visible, WDYT?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/GraphQLExtractor.kt
```diff
@@ -0,0 +1,116 @@
...
+        private const val CODE_KEY = "code"
+
+        private val GRAPHQL_HEADER_TO_ATTRIBUTE = mapOf(
+            GraphQLHeaders.DD_GRAPHQL_NAME_HEADER to RumAttributes.GRAPHQL_OPERATION_NAME,
+            GraphQLHeaders.DD_GRAPHQL_TYPE_HEADER to RumAttributes.GRAPHQL_OPERATION_TYPE,
+            GraphQLHeaders.DD_GRAPHQL_VARIABLES_HEADER to RumAttributes.GRAPHQL_VARIABLES,
+            GraphQLHeaders.DD_GRAPHQL_PAYLOAD_HEADER to RumAttributes.GRAPHQL_PAYLOAD
+        )
+
+        internal val JSON_CONTENT_TYPES = setOf(
```
satween, line 111: Use `HttpSpec.ContentType` constants here

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+internal class OkHttpGraphQLAdapter(
+    private val graphQLExtractor: GraphQLExtractor = GraphQLExtractor()
+) {
+
+    fun wrapChainWithoutDDHeaders(
+        internalLogger: InternalLogger,
+        originalChain: Interceptor.Chain
+    ): Interceptor.Chain {
+        return if (hasGraphQLHeaders(originalChain.request().headers)) {
+            object : Interceptor.Chain by originalChain {
```
satween, line 31: Worth to check if we could avoid using nested `Interceptor.Chain` here. Maybe we could change headers with the new `HttpRequestInfo` + `HttpRequestInfoBuilder`. Custom `Interceptor.Chain` unpredictable modifies the request flow and may lead to the issues, in case if customer has complex `okhttp` request interception pipeline

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+        graphqlAttributes: Map<String, Any?>,
+        internalLogger: InternalLogger
+    ): Map<String, Any> {
+        if (graphqlAttributes.isEmpty()) return emptyMap()
+        return try {
+            val responseInfo = OkHttpHttpResponseInfo(response, internalLogger)
+
+            @Suppress("UnsafeThirdPartyFunctionCall") // exceptions are caught
+            val body = response
+                .peekBody(GraphQLExtractor.MAX_GRAPHQL_BODY_PEEK)
```
satween, line 77: Mark that this method should be called inside `@WorkerThread`  and check that it's actually happened, because `peakBody` is an IO-bound operation, that could block the thread

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+                .string()
+            graphQLExtractor.extractGraphQLErrors(
+                responseInfo.contentType,
+                body,
+                internalLogger
+            )?.let {
+                mapOf(RumAttributes.GRAPHQL_ERRORS to it)
+            } ?: emptyMap()
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
```
satween, line 86: Does it make sense to change to `Throwable` here?

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapterTest.kt
```diff
@@ -0,0 +1,148 @@
...
+    lateinit var mockInternalLogger: InternalLogger
+
+    @Mock
+    lateinit var mockChain: Interceptor.Chain
+
+    lateinit var testedHelper: OkHttpGraphQLAdapter
+
+    @BeforeEach
+    fun `set up`() {
+        testedHelper = OkHttpGraphQLAdapter(mockGraphQLExtractor)
```
satween, line 64: you can get rid from the `set up` method here and just create the `testedHelper` as class property

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapterTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            .url("https://example.com/graphql")
+            .addHeader("User-Agent", fakeUserAgent)
+            .addHeader(GraphQLHeaders.DD_GRAPHQL_NAME_HEADER.headerValue, fakeGraphQLName.toBase64())
+            .build()
+        val response = Response.Builder()
+            .request(request)
+            .protocol(Protocol.HTTP_2)
+            .code(200)
+            .message("OK")
+            .body("""{"data":{}}""".toResponseBody("application/json".toMediaType()))
```
satween: Use `HttpSpec` constants to reduce risk of misstyping

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
kikoveiga: I think we can keep `Exception`, because OOM error can occur when peeking the body, even if we limit to 512 KB, wdyt?

dd-sdk-android-core/src/testFixtures/kotlin/com/datadog/android/tests/elmyr/ForgeExt.kt
kikoveiga: Added it, thank you!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/GraphQLExtractor.kt
kikoveiga: Using the `APPLICATION_JSON` from there now, I think we can keep the other 2 here because they are feature specific to graphql, wdyt?

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapterTest.kt
kikoveiga: I now added `this.forge = forge` inside the `set up` method so I guess we could keep it. Also because `testedHelper` needs the `lateinit var mockGraphQLExtractor`.

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapterTest.kt
kikoveiga: Added it, thank you!

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
kikoveiga: Added the annotation here and in the chain of called functions.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
kikoveiga: I think refactoring to use `HttpRequestInfoBuilder` requires changing how attributes flow through the hierarchy (e.g., passing them via request tags). This would involve a bigger change, should we note for a future PR?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/StringExt.kt
kikoveiga: I think this isn't added to customer's code/autocomplete, because `internal` module is a transitive dependency exposed via `implementation` (not `api`) through `core`.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
satween: yes, create a ticket for the investigation please and add as a TODO

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
satween: 🆗

0xnm (approved):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/net/GraphQLExtractor.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Extracts GraphQL attributes (operation name, type, variables, payload) from the request headers.
+     * @param requestInfo the HTTP request info containing headers
+     * @return a map of GraphQL attributes decoded from Base64 header values
+     */
+    fun extractGraphQLAttributes(requestInfo: HttpRequestInfo): Map<String, Any?> {
+        return buildMap {
+            GRAPHQL_HEADER_TO_ATTRIBUTE.forEach { (header, attribute) ->
+                requestInfo.headers[header.headerValue]?.firstOrNull()?.let {
```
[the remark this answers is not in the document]
kikoveiga: These headers are internal and only added once.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
kikoveiga: Created [RUM-15352](https://datadoghq.atlassian.net/browse/RUM-15352).

PR #3277, opened 2026-03-20
RUM-15190: Extend app launch profiling duration for continuous merge

ambushwork:
### What does this PR do?

This PR changes the max duration of app launch profiling request from 10s -> (ttid, 1 min] in order to be able to extend as the continuous profiling, at the same time a timer of 10s is introduced to prevent the application launch profiling to go beyond 10 seconds if continuous profiling is not sampled in.

### Motivation

RUM-15190


ambushwork:
@codex review

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
```diff
@@ -108,6 +108,7 @@ object Profiling {
         if (!isProfilerInitialized.getAndSet(true)) {
             profiler = PerfettoProfiler(
                 timeProvider = DefaultTimeProvider(),
+                appLaunchTimerExecutorService = Executors.newSingleThreadScheduledExecutor(),
```
0xnm: I don't think we need to spin-off yet another executor/thread, we already have many of them across the SDK, and having one more doesn't come for free.

You can simply change type of `profilingExecutor` to `Executors.newSingleThreadScheduledExecutor()`, because `ScheduledExecutorService` extends `ExecutorService` and use it for scheduling a timer.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -58,6 +59,7 @@ internal class ProfilingFeature(
         this.appContext = appContext
         profiler.apply {
             this.internalLogger = sdkCore.internalLogger
+            this.setRateBasedSampler(RateBasedSampler(configuration.continuousSampleRate))
```
0xnm, line 62: note: in the end we need to make it sticky - deterministic to the RUM session. We can handle it later.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -238,6 +275,22 @@ internal class PerfettoProfiler(
...
+        // Application launch profiling should always be considered as the first window of
+        // continuous profiling by default since the duration is not mutable after requesting,
+        // but the effective max duration will be controlled by an external timer if continuous
+        // profiling is not enabled by users.
+        return if (startReason == ProfilingStartReason.APPLICATION_LAUNCH) {
+            // Randomize t1 ∈ (0, CONTINUOUS_WINDOW] to provide phase jitter across sessions,
+            // avoiding systematic cooldown gaps at predictable time points.
+            @Suppress("UnsafeThirdPartyFunctionCall")
+            // Until is always bigger than from.
+            secureRandom.nextInt(1, PROFILING_MAX_DURATION_MS_CONTINUOUS + 1)
```
0xnm: What it the reason to use `SecureRandom` compared to `Random`? It is heavier to initialize and to generate a number.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -238,6 +275,22 @@ internal class PerfettoProfiler(
         } ?: 0
     }
 
+    private fun getExtendedAppLaunchDuration(startReason: ProfilingStartReason): Int {
```
0xnm:
```suggestion
    private fun getExtendedAppLaunchDurationMs(startReason: ProfilingStartReason): Int {
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -131,10 +145,29 @@ internal class PerfettoProfiler(
...
+            if (startReason == ProfilingStartReason.APPLICATION_LAUNCH) {
+                appLaunchTimerExecutorService.scheduleSafe(
+                    operationName = OPERATION_NAME_APP_LAUNCH_PROFILING_SCHEDULE,
+                    delay = APP_LAUNCH_PROFILING_MAX_DURATION_MS,
+                    unit = TimeUnit.MILLISECONDS,
+                    internalLogger = internalLogger ?: InternalLogger.UNBOUND,
+                    runnable = {
+                        stopSignal?.let { signal ->
+                            if (profilingStartReason == ProfilingStartReason.APPLICATION_LAUNCH &&
+                                !signal.isCanceled && continuousSampler.sample(Unit)
```
0xnm: we need to have decision sticky for the whole RUM session, but we need to take into the account that when this code is called, SDK may be not yet initialized -> no RUM session.

ambushwork: It can be tricky if we need to decide here since RUM session sampling decision is possibly not made yet (for example having a race condition), maybe we can move the decision before writing the result to see if there are any RUM events (not session) happen during the profiling?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
ambushwork: updated to `Random`

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/Profiling.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
0xnm: We can make a decision when RUM session is created, we can send event to the Profiling. WDYT?

ambushwork: If we can assume that RUM session creation + send event across the feature is always within 10s (in most of case yes), then we can do that.
and in the coming PR `continuousSampler.sample(Unit)` will be replaced by something called `continuousScheduler` which handles the whole decision tree

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -131,10 +140,29 @@ internal class PerfettoProfiler(
...
+                            if (profilingStartReason == ProfilingStartReason.APPLICATION_LAUNCH &&
+                                !signal.isCanceled && continuousSampler.sample(Unit).not()
+                            ) {
+                                signal.cancel()
+                            }
+                        }
+                    }
+                )
+                // Shutdown the service immediately after the task is scheduled.
+                profilingExecutor.shutdown()
```
0xnm: since we are re-using profiling executor, this line should be removed, it is used not only for scheduling this task

0xnm (changes requested):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
ambushwork: good point, that's a miss after the update

0xnm (dismissed):

0xnm (approved):

PR #3290, opened 2026-03-25
RUM-2778: Print error when no-op core is invoked

ambushwork:
### What does this PR do?

Print error when no-op core is invoked

### Motivation

RUM-2778

### Additional Notes

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
@@ -30,6 +30,7 @@ import okhttp3.Request
 import okhttp3.Response
 import okio.Timeout
 import java.io.File
+import java.util.Locale
 import java.util.UUID
 import java.util.concurrent.Callable
 import java.util.concurrent.Delayed
@@ -59,8 +60,7 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
     override val service: String
         get() = ""
 
-    override val internalLogger: InternalLogger
-        get() = SdkInternalLogger(this)
+    override val internalLogger: InternalLogger = SdkInternalLogger(this)
 
     override val timeProvider: TimeProvider
         get() = DefaultTimeProvider()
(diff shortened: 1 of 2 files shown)


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
```diff
@@ -121,9 +122,32 @@ internal object NoOpInternalSdkCore : InternalSdkCore {
...
+                REGISTER_FEATURE_INVOKED_ON_NO_OP_CORE_ERROR.format(
+                    Locale.US,
+                    feature.name
+                )
+            }
+        )
+    }
 
-    override fun getFeature(featureName: String): FeatureScope? = null
+    override fun getFeature(featureName: String): FeatureScope? {
```
0xnm, line 138: This may get very spammy, given the number of calls to this method. Maybe we should print here only once?

ambushwork: I think it depends, it's probably enough to print only once per feature, however if the client enables multiple features with no-op core and we print only the error for the first one, then it will be harder for them to discover the issue.

0xnm: I mean we can leverage `onlyOnce` argument of the logger, it will be printed once per _each_ feature.

https://github.com/DataDog/dd-sdk-android/blob/e37f350c098dff484269e1679640523f563ee49b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/InternalLogger.kt#L111

hamorillo (dismissed):
LGTM! 🚀

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCore.kt
ambushwork: oh yes, I was thinking the same argument, but I didn't realize that the message is unique per feature, so it will print once per feature indeed.

Note: another change is made for the convenience of test for `onlyOnce`, and I don't see why we need to `get()` a new `SdkInternalLogger` every time.
```
// before
override val internalLogger: InternalLogger
        get() = SdkInternalLogger(this)

// now
override val internalLogger: InternalLogger = SdkInternalLogger(this)
```

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCoreTest.kt
```diff
@@ -0,0 +1,156 @@
...
+        Datadog.setVerbosity(Int.MAX_VALUE)
+    }
+
+    @Test
+    fun `M not throw W registerFeature()`(@StringForgery fakeFeatureName: String) {
+        // Given
+        whenever(mockFeature.name) doReturn fakeFeatureName
+
+        // When + Then
+        Mockito.mockStatic(Log::class.java).use {
```
0xnm: Do we need to mock if we don't verify log calls anyway? Normally it should be just no-op without mocking.

Probably it is redundant here and in the 2 tests below.

0xnm (dismissed):

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/NoOpInternalSdkCoreTest.kt
ambushwork: remove mockStatic since it's not necessary, but keep the tests since they can still protect some corner cases.

0xnm (approved):

hamorillo (approved):

PR #3291, opened 2026-03-25
RUM-4120: Add Detekt rule to enforce copyright headers on Kotlin source files

hamorillo:
### What does this PR do?
- Introduces a Detekt [AbsentOrWrongFileLicense](https://detekt.dev/docs/rules/comments/#absentorwrongfilelicense) rule that verifies every `.kt` source file in the repository starts with the standard DataDog header. 
- Fixes all pre-existing violations found across the codebase.

### Motivation
Copyright headers were only enforced via the IDE's IDEA copyright profile, which relies on developers remembering to apply it. Several files had missing, incorrectly indented, or typoed headers with no automated safety net. This rule makes violations a CI failure.

### Additional Notes
- A `license.template` file is added at the repo root — this is the single source of truth for the expected header, read by Detekt at analysis time.
- Exclusions are narrowly scoped: `build` outputs, auto-generated buildSrc model fixtures, NoOpFactory test resource fixtures, and vendored dd-trace-java test files.
- `.kts` files were already excluded from the Detekt Gradle task (`-ex "**/*.kts"`). The copyright headers on `.kts` files have been added and are correct. We can dive deeper into this if needed.

### DIFF
--- a/license.template
+++ b/license.template
@@ -1,0 +1,5 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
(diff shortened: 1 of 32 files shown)


hamorillo:
@codex review

license.template
```diff
@@ -0,0 +1,5 @@
+/*
```
0xnm, line 1: We have it in https://github.com/DataDog/dd-sdk-android/blob/develop/.idea/copyright/Datadog.xml, maybe we can parse from there to have a single source of truth?

Upd: ah, I see, it is built-in rule which requires copyright in this format.

0xnm (approved):
nice!

license.template
hamorillo: >  it is built-in rule which requires copyright in this format.

Yeah, 😞 

What I was unsure about is where to place the `licence.template` file. I don’t really like putting it in the root, but we already have other Detekt files there. If you have a strong opinion on this, let me know.

kikoveiga (approved):

PR #3294, opened 2026-03-25
fix: E2E test fixes for VIEW_UPDATE pipeline validation

andreicoj:
## Problem

The `RumViewUpdateTest` E2E test couldn't run against staging due to three issues:

1. **`DatadogSite.LOCAL`** — SDK crashes on init (AccessibilityReader NPE) because LOCAL endpoint doesn't exist
2. **`AlwaysFullView`** — SDK sends full VIEWs only, never VIEW_UPDATEs, defeating the test purpose  
3. **`RumSearchResponse.service: String?`** — deserialization crash when API returns service as array instead of string

### Critical fixes
- **BaseRumViewTest**: `DatadogSite.LOCAL` → `STAGING`, `AlwaysFullView` → `FullViewOnlyAtStart`
- **RumSearchResponse**: `service: String?` → `JsonElement?` (handles both string and array)
- **RumSearchResponseViewEventAssert**: `hasService` uses `actual.attributes.attributes.service` (inner field)

### Debug logging (non-critical)
- **KtorHttpResponse**: `POLL_DEBUG` logging for HTTP response status
- **RumViewUpdateTest**: `POLL_DEBUG` in polling predicate

## Testing
Verified 15/15 consecutive passes against staging with backend partial-updates stack deployed.

### DIFF
--- a/reliability/with-backend/src/androidTest/kotlin/com/datadog/android/core/integration/tests/rum/RumViewUpdateTest.kt
+++ b/reliability/with-backend/src/androidTest/kotlin/com/datadog/android/core/integration/tests/rum/RumViewUpdateTest.kt
@@ -103,7 +103,11 @@ class RumViewUpdateTest : BaseRumViewTest() {
                         contextAttributes = mapOf("test_view_index" to 14)
                     )
                 },
-                predicate = { it.optionalResult?.data?.firstOrNull() != null },
+                predicate = {
+                    val found = it.optionalResult?.data?.firstOrNull() != null
+                    android.util.Log.w("POLL_DEBUG", "predicate: found=$found")
+                    found
+                },
                 interval = POLLING_INTERVAL_MS.milliseconds,
                 timeout = POLLING_TIMEOUT_MS.milliseconds
             )
@@ -201,7 +205,7 @@ class RumViewUpdateTest : BaseRumViewTest() {
 
     companion object {
         private const val VIEW_NAME = "rum-view-update-test"
-        private const val POLLING_TIMEOUT_MS = 30_000L
+        private const val POLLING_TIMEOUT_MS = 60_000L
         private const val POLLING_INTERVAL_MS = 5_000L
     }
 }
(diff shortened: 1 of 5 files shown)


aleksandr-gringauz:
> Polling predicate too loose — event != null could match a stale active view from a previous test run before the current run's closed view appears

@andreicoj This statement looks wrong. The test queries the ViewEvent by view_id. So it shouldn't get the view from the previous run.

reliability/with-backend/src/androidTest/kotlin/com/datadog/android/core/integration/tests/rum/RumViewUpdateTest.kt
```diff
@@ -103,7 +103,13 @@ class RumViewUpdateTest : BaseRumViewTest() {
                         contextAttributes = mapOf("test_view_index" to 14)
                     )
                 },
-                predicate = { it.optionalResult?.data?.firstOrNull() != null },
+                predicate = {
+                    val event = it.optionalResult?.data?.firstOrNull()
+                    val errorCount = event?.attributes?.attributes?.view?.error?.count
+                    val isActive = event?.attributes?.attributes?.view?.isActive
+                    android.util.Log.w("POLL_DEBUG", "predicate: errorCount=$errorCount isActive=$isActive")
+                    event != null && isActive == false
```
aleksandr-gringauz: `isActive` check here is also useless. We query also by `contextAttributes = mapOf("test_view_index" to 14)` (see above). So we are 100% sure we get the latest version.

andreicoj:
You're right — I updated the PR description. The query filters by viewId so there's no stale view risk. I reverted the predicate to your original `event != null`.

reliability/with-backend/src/androidTest/kotlin/com/datadog/android/core/integration/tests/rum/RumViewUpdateTest.kt
andreicoj: Agreed, reverted. The isActive check was unnecessary since the query already filters by viewId + test_view_index.

PR #3209, opened 2026-03-02
[WIP] - partial view updates implementation

aleksandr-gringauz:
### What does this PR do?

PoC implementation of partial view updates.

The algorithm is roughly the following (see the diff in [RumViewScope.kt](https://github.com/DataDog/dd-sdk-android/pull/3209/changes#diff-ab30afe3da5a0243ae755135af0e1296b0137b5794ad63b8ec03d9731a5d75d2)). For each `ViewEvent` that we create we:
1. Apply the EventMapper provided by the customer.
2. Compute the diff of this mapped event with the previous one. This diff is represented as `ViewEventDiff` object. In order to compute the diff of the `ViewEvent` respecting REPLACE/MERGE/APPEND rules from the [RFC](https://docs.google.com/document/d/1roqhRU1SginuThUqCwXlmSORrpChRcqxOxdfTnmvOPw/edit?tab=t.0#heading=h.rnd972k0hiye) I created a KSP processor that generates the diffing code based on a special `DiffConfig` annotation. More info [here](https://github.com/DataDog/dd-sdk-android/pull/3209/changes#r2872615104).
3. We map `ViewEventDiff` into the new `ViewUpdate` event introduced [here](https://github.com/DataDog/rum-events-format/pull/355). See [ViewEventDiffMapper](https://github.com/DataDog/dd-sdk-android/pull/3209/changes#diff-59399c82ff44e6b1388e252d93c60abe6e4e4127ad7d272df0055113e7775fb7).
4. We send the `ViewUpdate` event to the intake.

This PR doesn't address:
1. tests
2. benchmarks
3. periodic sending of the full VIEW event.

### Motivation

What inspired you to submit this pull request?

### Additional Notes

Anything else we should know when reviewing?

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

PR #3296, opened 2026-03-25
RUM-14814: Add DiffBuilder utility and ViewUpdateEvent diff computation

hamorillo:
### What does this PR do?
Introduces `DiffBuilder`, a generic diff utility for computing sparse diffs between two objects of the same type, and `diffViewEvent`, which uses it to produce a minimal `ViewUpdateEvent` payload from two consecutive `ViewEvent` snapshots.

### Motivation
`VIEW_UPDATE` events must be sparse — they should only carry fields that changed since the last full VIEW, plus a fixed set of required fields. This utility provides the building block for generating compliant `VIEW_UPDATE` payloads.

### Additional Notes
- `DiffBuilder` lives in `dd-sdk-android-internal` and supports four diff strategies: `diffRequired` (always present), `diffEquals` (REPLACE — send if changed), `diffList` (APPEND — new tail elements only), `diffMap` (MERGE — new/changed keys only), and `diffMerge` (per-field recursive merge).
- `diffViewEvent` is in `com.datadog.android.rum.internal.model` and has no call site yet — integration with the event pipeline is in a follow-up PR.
- `ViewEventConversions.kt` holds the `toRum()` extension functions and enum mappers extracted from `ViewEventDiff.kt`.
- `view.is_active` is intentionally hardcoded to `null` in `VIEW_UPDATE` until the backend can distinguish absent `false` from explicit `false` (tracked in RUM-14814).
- `detekt_custom_safe_calls.yml` extended with `kotlin.collections.MutableMap.put(kotlin.Any, kotlin.Any?)`.

cc: @aleksandr-gringauz


hamorillo:
@codex review

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/DiffBuilder.kt
```diff
@@ -0,0 +1,187 @@
...
+    override fun <T> diffEquals(getter: D.() -> T): T? {
+        val old = getter(oldObj)
+        val new = getter(newObj)
+
+        if (old != new) {
+            changed = true
+            return new
+        } else {
+            return null
+        }
```
hamorillo, line 128: ℹ️ We don’t expect transitions between `old = value` and `new = null`, since in both cases the method would return `null`, so we can’t differentiate between no change and a change to `null`. In any case, the serializer doesn’t send fields with `null`, so if we ever need to support that, we’ll also need to improve the serializer generator.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/DiffBuilder.kt
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    override fun <T> diffEquals(getter: D.() -> T): T? {
+        val old = getter(oldObj)
+        val new = getter(newObj)
+
+        if (old != new) {
+            changed = true
+            return new
+        } else {
```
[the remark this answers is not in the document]
hamorillo: See https://github.com/DataDog/dd-sdk-android/pull/3296#discussion_r2993783420

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
```diff
@@ -0,0 +1,431 @@
...
+    }
+}
+
+@Suppress("LongMethod")
+private fun diffView(
+    old: ViewEvent.ViewEventView,
+    new: ViewEvent.ViewEventView
+): ViewUpdateEvent.ViewUpdateEventView {
+    return computeDiffRequired(old = old, new = new) {
+        val slowFrames = diffList(ViewEvent.ViewEventView::slowFrames)
```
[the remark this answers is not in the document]
hamorillo: Resolved in https://github.com/DataDog/dd-sdk-android/pull/3296/commits/41df82df698b0133000cedc69fbcf32f384ec530

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/utils/DiffBuilder.kt
```diff
@@ -0,0 +1,187 @@
...
+    override fun <K : Any, V> diffMap(getter: D.() -> Map<K, V>): Map<K, V> {
+        val old = getter(oldObj)
+        val new = getter(newObj)
+
+        val diff = buildMap {
+            for ((key, newValue) in new) {
+                val oldValue = old[key]
+
+                if (newValue != oldValue) {
+                    put(key, newValue)
```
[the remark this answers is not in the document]
hamorillo: >  adding a new key with value null

I would say that’s the key point: we don’t need or want to send a key with a null value.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
```diff
@@ -0,0 +1,265 @@
...
+            cpuTicksPerSecond = diffEquals(ViewEvent.ViewEventView::cpuTicksPerSecond),
+            refreshRateAverage = diffEquals(ViewEvent.ViewEventView::refreshRateAverage),
+            refreshRateMin = diffEquals(ViewEvent.ViewEventView::refreshRateMin),
+            slowFramesRate = diffEquals(ViewEvent.ViewEventView::slowFramesRate),
+            freezeRate = diffEquals(ViewEvent.ViewEventView::freezeRate),
+            flutterBuildTime = diffEquals(ViewEvent.ViewEventView::flutterBuildTime)?.toRum(),
+            flutterRasterTime = diffEquals(ViewEvent.ViewEventView::flutterRasterTime)?.toRum(),
+            jsRefreshRate = diffEquals(ViewEvent.ViewEventView::jsRefreshRate)?.toRum(),
+            performance = diffMerge({ performance ?: ViewEvent.Performance() }, ::diffPerformance),
+            accessibility = diffMerge({ accessibility ?: ViewEvent.Accessibility() }, ::diffAccessibility)
```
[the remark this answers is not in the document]
hamorillo: Addressed in https://github.com/DataDog/dd-sdk-android/pull/3296/commits/f441bd0fde4f557f41c7eba52c345696b34e4632

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventConversions.kt
```diff
@@ -0,0 +1,189 @@
...
+@file:Suppress("TooManyFunctions")
+
+package com.datadog.android.rum.internal.model
+
+import com.datadog.android.rum.model.ViewEvent
+import com.datadog.android.rum.model.ViewUpdateEvent
+
+// region Type conversions
+
+internal fun ViewEvent.Usr.toRum() = ViewUpdateEvent.Usr(
```
aleksandr-gringauz: nit: `toRum()` -> `toViewUpdate()`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
```diff
@@ -0,0 +1,271 @@
...
+            domContentLoaded = diffEquals(ViewEvent.ViewEventView::domContentLoaded),
+            domInteractive = diffEquals(ViewEvent.ViewEventView::domInteractive),
+            loadEvent = diffEquals(ViewEvent.ViewEventView::loadEvent),
+            firstByte = diffEquals(ViewEvent.ViewEventView::firstByte),
+            customTimings = diffEquals(ViewEvent.ViewEventView::customTimings)?.let {
+                ViewUpdateEvent.CustomTimings(it.additionalProperties)
+            },
+            // TODO RUM-14814: is_active must not be sent in VIEW_UPDATE until backend fix is complete
+            // (backend cannot distinguish absent false from explicit false — risks corrupting view state)
+            // isActive = diffEquals(ViewEvent.ViewEventView::isActive),
```
aleksandr-gringauz: I think I saw somewhere that sending `is_active` in ViewUpdate should work now. @andreicoj could you confirm please?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
```diff
@@ -0,0 +1,271 @@
...
+    }
+}
+
+@Suppress("LongMethod")
+private fun diffView(
+    old: ViewEvent.ViewEventView,
+    new: ViewEvent.ViewEventView
+): ViewUpdateEvent.ViewUpdateEventView {
+    return computeDiffRequired(old = old, new = new) {
+        val slowFrames = diffMerge({ slowFrames ?: emptyList() }, ::diffSlowFrames)
```
aleksandr-gringauz: I think calling `diffMerge` is redundant here. Should we just call `diffSlowFrames` instead?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/model/ViewEventDiffTest.kt
```diff
@@ -0,0 +1,673 @@
...
+        )
+
+        // When
+        val result = diffViewEvent(old, new)
+
+        // Then
+        assertThat(result.view.performance?.fbc?.timestamp).isEqualTo(fakeTimestamp)
+    }
+
+    // region TODOs — documenting current (pending review) behaviour
```
aleksandr-gringauz: We can probably remove this if `isActive` is supported.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
```diff
@@ -0,0 +1,271 @@
...
+
+// region Diff helpers — nullable merged sub-objects
+
+private fun diffSlowFrames(
+    old: List<ViewEvent.SlowFrame>,
+    new: List<ViewEvent.SlowFrame>
+): List<ViewUpdateEvent.SlowFrame>? {
+    // diffList (drop-by-size) breaks when the backing EvictingQueue evicts the head at capacity:
+    // both lists have the same size but the content has shifted. Using the last known start
+    // timestamp as an anchor ensures only genuinely new frames are returned regardless of eviction.
```
aleksandr-gringauz, line 202: Nice catch.

I wonder if the backend is now enforcing the max size for the slowFrames somehow on their side. Because IIUC the current implementation will send new slowFrames even if there were more than `DEFAULT_SLOW_FRAME_RECORDS_MAX_AMOUNT` of them.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
0xnm, line 1: How do we make sure to update this file when `ViewEvent` model is updated? It seems that the content of this file is quite repetitive, do we consider generating this file instead?

hamorillo: > How do we make sure to update this file when ViewEvent model is updated?

That’s a valid concern. To be honest, I’m not sure we can generate `ViewEventDiff.kt` as it’s not purely mechanical and requires making some decisions for each field.

We can create a ticket to see how we can improve this, or do you think it’s a blocking issue?

0xnm: Not a blocker, I'm just curious about the future handling of model updates.

hamorillo: Yeah, it's a really good point. Thanks! 

_Ticket created: RUM-15316_

0xnm: For example, in KMP we are doing a generation of the mapper between two models: KMP and Android, KMP and iOS.

https://github.com/DataDog/dd-sdk-kotlin-multiplatform/blob/develop/tools/build-plugins/src/main/kotlin/com/datadog/build/plugin/jsonschema/generator/AndroidModelsMappingFileGenerator.kt

generated file: https://github.com/DataDog/dd-sdk-kotlin-multiplatform/blob/develop/tools/build-plugins/src/test/resources/AndroidViewEventMappingExt.kt

aleksandr-gringauz: There are several things we could do on this front:
1. Add an option to our JSON serialization code generator so that it doesn't create default values for constructor arguments. This way if a new argument is added, this file will stop compiling. And for new fields we'll need to decide how to compute diff for them. I would say this is the most realistic and useful idea.
2. I tried to implement some code generation using KSP https://github.com/DataDog/dd-sdk-android/pull/3209/changes/BASE..93984c95de74f32b92de746adf3df929b24e5602#diff-1ca3c9e762a30cef85c45e10d9e844201f1b495913b50d5926356c92c098e9e8. I ditched it in favor of the current approach because of the following reasons. First - we need to somehow describe for each field what strategy to use. The KSP approach wasn't superior to the manual one in this regard (same thing - for some field use some strategy). Second - I couldn't directly generate the function that returns `ViewUpdateEvent` (the one corresponding to `view-event-schema.json`), I generated some intermediate representation (`ViewEventDiff`), that had to be mapped to `ViewUpdateEvent`.
3. Add the information about the diff strategies for the fields (merge, replace, append, etc) into the json schema itself and generate this file from the schema. Sounds great, but the implementation details aren't clear. What about `ViewUpdate` event in this case? For it we will stop having a json schema?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventConversions.kt
hamorillo: Done! https://github.com/DataDog/dd-sdk-android/pull/3296/commits/91bade0712e20db00196cb871a64234a198d3ed4

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
hamorillo: Done! https://github.com/DataDog/dd-sdk-android/pull/3296/commits/8e6b815cb80cc67ef1dac5554e55822a11da72db

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/model/ViewEventDiffTest.kt
hamorillo: Done! https://github.com/DataDog/dd-sdk-android/pull/3296/commits/b07f48153193864daa36164882265f608f6b1a08

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/model/ViewEventDiff.kt
hamorillo: @andreicoj has confirmed that we can send the `isActive`. 

Done ->  https://github.com/DataDog/dd-sdk-android/pull/3296/commits/b07f48153193864daa36164882265f608f6b1a08

aleksandr-gringauz (approved):

PR #3298, opened 2026-03-26
RUM-15191: Implement ContinuousProfilingScheduler

ambushwork:
### What does this PR do?

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
@@ -1,0 +1,160 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.profiling.internal
+
+import android.content.Context
+import android.os.Build
+import androidx.annotation.RequiresApi
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.core.internal.utils.scheduleSafe
+import com.datadog.android.core.sampling.RateBasedSampler
+import java.util.Locale
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+import kotlin.random.Random
+
+@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+internal class ContinuousProfilingScheduler(
+    private val appContext: Context,
+    private val profiler: Profiler,
+    private val sdkCore: FeatureSdkCore,
+    private val sampleRate: Float
+) {
+
+    @Volatile
+    internal var isScheduling = false
+
+    // TODO RUM-15315: Read RUM session tracked state instead of using default value.
+    @Volatile
+    internal var rumSessionSampled = false
+
+    @Volatile
+    private var pendingFuture: ScheduledFuture<*>? = null
+
+    private val schedulerExecutor: ScheduledExecutorService = profiler.scheduledExecutorService
+
+    fun start(launchProfilingActive: Boolean) {
+        isScheduling = RateBasedSampler<Unit>(sampleRate).sample(Unit)
+        profiler.setExtendLaunchSession(isScheduling)
+        if (!isScheduling) {
+            logToUser { LOG_DISABLED.format(Locale.US, sampleRate) }
+            return
+        }
+        if (launchProfilingActive) {
+            logToUser { LOG_WAITING_FOR_LAUNCH }
+            return
+        }
+        // If app launch profiling is not started, then schedule a cooldown window to replace it,
+        // c1 ∈ [0, 1min], otherwise wait for `onAppLaunchProfilingComplete` to be called.
+        @Suppress("UnsafeThirdPartyFunctionCall") // range is always valid (0 .. max+1)
+        val initialCooldownMs = Random.nextLong(0L, CONTINUOUS_COOLDOWN_DURATION_MS + 1L)
+        logToUser { LOG_INITIAL_COOLDOWN.format(Locale.US, initialCooldownMs) }
+        scheduleCooldown(initialCooldownMs)
+    }
+
+    fun stop() {
+        logToUser { LOG_STOPPED }
+        pendingFuture?.cancel(false)
+    }
+
+    fun onAppLaunchProfilingComplete() {
+        if (!isScheduling) return
+        logToUser { LOG_LAUNCH_COMPLETE }
+        scheduleNextCycle()
+    }
+
+    fun onRumSessionRenewed(sessionSampled: Boolean) {
+        rumSessionSampled = sessionSampled
+    }
+
+    private fun scheduleNextCycle() {
+        val activeMs = jitter(CONTINUOUS_WINDOW_DURATION_MS)
+        if (rumSessionSampled) {
+            logToUser { LOG_ACTIVE_WINDOW_STARTED.format(Locale.US, activeMs) }
+            profiler.start(
+                appContext = appContext,
+                startReason = ProfilingStartReason.CONTINUOUS,
+                additionalAttributes = emptyMap(),
+                sdkInstanceNames = setOf(sdkCore.name),
+                durationMs = activeMs.toInt()
+            )
+        } else {
+            logToUser { LOG_ACTIVE_WINDOW_SKIPPED }
+        }
+
+        // If the active window is skipped due to RUM session,
+        // next cooldown window still needs to be scheduled.
+        pendingFuture = schedulerExecutor.scheduleSafe(
+            operationName = OPERATION_ACTIVE_WINDOW_END,
+            delay = activeMs,
+            unit = TimeUnit.MILLISECONDS,
+            internalLogger = sdkCore.internalLogger,
+            runnable = {
+                val cooldownMs = jitter(CONTINUOUS_COOLDOWN_DURATION_MS)
+                logToUser { LOG_ACTIVE_WINDOW_ENDED.format(Locale.US, cooldownMs) }
+                scheduleCooldown(cooldownMs)
+            }
+        )
+    }
+
+    private fun scheduleCooldown(cooldownMs: Long) {
+        pendingFuture = schedulerExecutor.scheduleSafe(
+            operationName = OPERATION_COOLDOWN_END,
+            delay = cooldownMs,
+            unit = TimeUnit.MILLISECONDS,
+            internalLogger = sdkCore.internalLogger,
+            runnable = ::scheduleNextCycle
+        )
+    }
+
+    private fun jitter(base: Long): Long {
+        val delta = (base * JITTER_FACTOR).toLong()
+        @Suppress("UnsafeThirdPartyFunctionCall") // delta is always < base, so range is valid
+        return base + Random.nextLong(-delta, delta + 1L)
+    }
+
+    private fun logToUser(message: () -> String) {
+        sdkCore.internalLogger.log(
+            InternalLogger.Level.DEBUG,
+            InternalLogger.Target.USER,
+            message
+        )
+    }
+
+    companion object {
+        // Base active window duration: 1 minute (±20% jitter applied per cycle).
+        internal const val CONTINUOUS_WINDOW_DURATION_MS = 60_000L
+
+        // Base cooldown duration: 1 minute (±20% jitter applied per cycle).
+        internal const val CONTINUOUS_COOLDOWN_DURATION_MS = 60_000L
+
+        // Fraction of base duration used as jitter range (±20%).
+        private const val JITTER_FACTOR = 0.20
+
+        private const val OPERATION_ACTIVE_WINDOW_END = "continuous_profiling_active_window"
+        private const val OPERATION_COOLDOWN_END = "continuous_profiling_cooldown"
+
+        internal const val LOG_DISABLED =
+            "Continuous profiling disabled (not sampled in at rate=%s)."
+        internal const val LOG_WAITING_FOR_LAUNCH =
+            "Continuous profiling enabled; waiting for app launch profiling to complete."
+        internal const val LOG_INITIAL_COOLDOWN =
+            "Continuous profiling enabled; initial cooldown %sms before first active window."
+        internal const val LOG_STOPPED =
+            "Continuous profiling stopped."
+        internal const val LOG_LAUNCH_COMPLETE =
+            "App launch profiling complete; starting first continuous cycle."
+        internal const val LOG_ACTIVE_WINDOW_STARTED =
+            "Continuous profiling active window started (%sms)."
+        internal const val LOG_ACTIVE_WINDOW_SKIPPED =
+            "Continuous profiling active window skipped (RUM session not sampled); cooldown still scheduled."
+        internal const val LOG_ACTIVE_WINDOW_ENDED =
+            "Continuous profiling active window ended; cooling down for %sms."
+    }
+}
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
@@ -16,7 +16,6 @@ import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.api.feature.StorageBackedFeature
 import com.datadog.android.api.net.RequestFactory
 import com.datadog.android.api.storage.FeatureStorageConfiguration
-import com.datadog.android.core.sampling.RateBasedSampler
 import com.datadog.android.internal.profiling.ProfilerStopEvent
 import com.datadog.android.internal.profiling.TTIDRumContext
 import com.datadog.android.internal.rum.RumSessionRenewedEvent
@@ -39,12 +38,20 @@ internal class ProfilingFeature(
     @Volatile
     private var ttidRumContext: TTIDRumContext? = null
 
+    // True once ProfilerStopEvent.TTID has been received, regardless of whether the RUM session
+    // was sampled (i.e. ttidRumContext may still be null when this is true).
+    @Volatile
+    private var ttidEventReceived = false
+
     @Volatile
     private var perfettoResult: PerfettoResult? = null
 
     private val isTtidProfileSent: AtomicBoolean = AtomicBoolean(false)
 
     private lateinit var appContext: Context
+
+    @Volatile
+    private var continuousProfilingScheduler: ContinuousProfilingScheduler? = null
 
     override val requestFactory: RequestFactory = ProfilingRequestFactory(
         configuration.customEndpointUrl
@@ -60,7 +67,6 @@ internal class ProfilingFeature(
         this.appContext = appContext
         profiler.apply {
             this.internalLogger = sdkCore.internalLogger
-            this.setRateBasedSampler(RateBasedSampler(configuration.continuousSampleRate))
             registerProfilingCallback(sdkCore.name) { result ->
                 perfettoResult = result
                 tryWriteProfilingEvent()
@@ -80,12 +86,23 @@ internal class ProfilingFeature(
             context[PROFILER_IS_RUNNING] = profiler.isRunning(sdkCore.name)
         }
         dataWriter = createDataWriter(sdkCore)
+
+        val scheduler = ContinuousProfilingScheduler(
+            appContext = appContext,
+            profiler = profiler,
+            sdkCore = sdkCore,
+            sampleRate = configuration.continuousSampleRate
+        )
+        continuousProfilingScheduler = scheduler
+        scheduler.start(launchProfilingActive = profiler.isRunning(sdkCore.name))
     }
 
     override fun onStop() {
+        continuousProfilingScheduler?.stop()
         profiler.apply {
             stop(sdkCore.name)
             unregisterProfilingCallback(sdkCore.name)
+            scheduledExecutorService.shutdown()
         }
         sdkCore.removeEventReceiver(name)
     }
@@ -103,8 +120,13 @@ internal class ProfilingFeature(
     }
 
     private fun onTtidEvent(event: ProfilerStopEvent.TTID) {
-        if (ttidRumContext == null) {
-            ttidRumContext = event.rumContext
+        if (ttidEventReceived) return // already handled
+
+        ttidEventReceived = true
+        ttidRumContext = event.rumContext
+
+        if (continuousProfilingScheduler?.isScheduling != true) {
+            // Non-continuous: stop profiler immediately at TTID.
             profiler.stop(sdkCore.name)
             tryWriteProfilingEvent()
             sdkCore.internalLogger.log(
@@ -115,9 +137,8 @@ internal class ProfilingFeature(
         }
     }
 
-    @Suppress("UnusedParameter", "UNUSED_PARAMETER")
     private fun onRumSessionRenewed(event: RumSessionRenewedEvent) {
-        // TODO RUM-15190: forward to ContinuousProfilingScheduler once it is implemented.
+        continuousProfilingScheduler?.onRumSessionRenewed(event.sessionSampled)
     }
 
     private fun setMinimumSampleRate(appContext: Context, sampleRate: Float) {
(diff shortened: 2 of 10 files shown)


ambushwork:
@codex review

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -103,8 +114,12 @@ internal class ProfilingFeature(
...
-        if (ttidRumContext == null) {
-            ttidRumContext = event.rumContext
+        if (ttidRumContext != null) return // already handled
+
+        ttidRumContext = event.rumContext
+
+        if (continuousProfilingScheduler?.continuousEnabled != true) {
+            // Non-continuous: stop profiler immediately at TTID.
             profiler.stop(sdkCore.name)
             tryWriteProfilingEvent()
```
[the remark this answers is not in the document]
ambushwork: This is actually a miss in previous PR, the duration for app launch should be (TTID, 1min] instead of (1s, 1min] which can lead app launch profiling ends before TTID, will fix it in another PR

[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -131,14 +145,19 @@ internal class ProfilingFeature(
...
-                profilingResult = perfettoResult,
-                ttidRumContext = ttidRumContext
-            )
+        val result = perfettoResult ?: return
+        when (result.tag) {
+            ProfilingStartReason.APPLICATION_LAUNCH.value -> {
+                val ttidRumContext = ttidRumContext ?: return
+                if (!isTtidProfileSent.getAndSet(true)) {
+                    dataWriter.write(profilingResult = result, ttidRumContext = ttidRumContext)
+                    continuousProfilingScheduler?.onAppLaunchProfilingComplete()
```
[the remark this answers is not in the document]
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+    fun start(launchProfilingActive: Boolean = false) {
+        continuousEnabled = RateBasedSampler<Unit>(sampleRate).sample(Unit)
+        profiler.setExtendLaunchSession(continuousEnabled)
+        if (!continuousEnabled) {
+            logToUser { LOG_DISABLED.format(Locale.US, sampleRate) }
+            return
+        }
+        if (launchProfilingActive) {
+            logToUser { LOG_WAITING_FOR_LAUNCH }
+            return
```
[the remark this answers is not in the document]
ambushwork: need to refactor a bit the callback from PerfettoProfiler, defer this to RUM-15307

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
```diff
@@ -0,0 +1,372 @@
...
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+@Suppress("UndocumentedPublicClass")
```
0xnm: this annotation is not needed, it is a test class (and it is `internal` anyway)

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+    internal var continuousEnabled = false
+
+    @Volatile
+    internal var rumSessionSampled = false
+
+    @Volatile
+    private var pendingFuture: ScheduledFuture<*>? = null
+
+    fun start(launchProfilingActive: Boolean = false) {
+        continuousEnabled = RateBasedSampler<Unit>(sampleRate).sample(Unit)
```
0xnm: Let's add TODO item here to make this depend on the RUM session? We anyway need to reset scheduling decision once RUM session is renewed.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+    @Volatile
+    internal var continuousEnabled = false
+
+    @Volatile
+    internal var rumSessionSampled = false
+
+    @Volatile
+    private var pendingFuture: ScheduledFuture<*>? = null
+
+    fun start(launchProfilingActive: Boolean = false) {
```
0xnm: maybe we need to remove default value to be more explicit and rename the argument (I understood the meaning only after reading log message in the method body).

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+import kotlin.random.Random
+
+@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+internal class ContinuousProfilingScheduler(
+    private val appContext: Context,
+    private val profiler: Profiler,
+    private val sdkCore: FeatureSdkCore,
+    private val sampleRate: Float,
+    private val internalLogger: InternalLogger,
+    private val schedulerExecutor: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
```
0xnm: we should be able to reuse executor we already have

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+    private var pendingFuture: ScheduledFuture<*>? = null
+
+    fun start(launchProfilingActive: Boolean = false) {
+        continuousEnabled = RateBasedSampler<Unit>(sampleRate).sample(Unit)
+        profiler.setExtendLaunchSession(continuousEnabled)
+        if (!continuousEnabled) {
+            logToUser { LOG_DISABLED.format(Locale.US, sampleRate) }
+            return
+        }
+        if (launchProfilingActive) {
```
0xnm, line 49: `launchProfilingActive` argument is used only to log the message, is it needed as a method argument? or we want to make `setExtendLaunchSession` irrespective of the value?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+import java.util.concurrent.TimeUnit
+import kotlin.random.Random
+
+@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+internal class ContinuousProfilingScheduler(
+    private val appContext: Context,
+    private val profiler: Profiler,
+    private val sdkCore: FeatureSdkCore,
+    private val sampleRate: Float,
+    private val internalLogger: InternalLogger,
```
kikoveiga: Maybe get the `internalLogger` from the `sdkCore`.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+        // c1 ∈ [0, 1min], otherwise wait for `onAppLaunchProfilingComplete` to be called.
+        @Suppress("UnsafeThirdPartyFunctionCall") // range is always valid (0 .. max+1)
+        val initialCooldownMs = Random.nextLong(0L, CONTINUOUS_COOLDOWN_DURATION_MS + 1L)
+        logToUser { LOG_INITIAL_COOLDOWN.format(Locale.US, initialCooldownMs) }
+        scheduleCooldown(initialCooldownMs)
+    }
+
+    fun stop() {
+        logToUser { LOG_STOPPED }
+        pendingFuture?.cancel(false)
```
0xnm, line 63: We are using non-interruptive cancel, meaning we let already running task to complete. If I'm not wrong it is not different from `shutdown` call below, so maybe we can use only `shutdown`? This will remove the need to keep the reference to the `pendingFuture`.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+
+    private fun scheduleNextCycle() {
+        val activeMs = jitter(CONTINUOUS_WINDOW_DURATION_MS)
+        if (rumSessionSampled) {
+            logToUser { LOG_ACTIVE_WINDOW_STARTED.format(Locale.US, activeMs) }
+            profiler.start(
+                appContext = appContext,
+                startReason = ProfilingStartReason.CONTINUOUS,
+                additionalAttributes = emptyMap(),
+                sdkInstanceNames = setOf(sdkCore.name),
```
0xnm, line 84: This is where we need to refactor, because instance names are needed only for the application launch profiling to perform the broadcast, but when SDK is up and running we have an access to the complete `SdkCore` objects. I'm not even sure we need the support of multiple SDK cores for the normal profiler runs (but we can decide on this later).

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -0,0 +1,161 @@
...
+                durationMs = activeMs.toInt()
+            )
+        } else {
+            logToUser { LOG_ACTIVE_WINDOW_SKIPPED }
+        }
+
+        // If the active window is skipped due to RUM session,
+        // next cooldown window still needs to be scheduled.
```
0xnm, line 92: I'm not sure I get it. If RUM session wasn't sampled, shouldn't we just stop all profiling activity?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -80,9 +85,20 @@ internal class ProfilingFeature(
             context[PROFILER_IS_RUNNING] = profiler.isRunning(sdkCore.name)
         }
         dataWriter = createDataWriter(sdkCore)
+
+        val scheduler = ContinuousProfilingScheduler(
```
kikoveiga, line 90: Should we intialize the `ContinuousProfilingScheduler` with a `rumSessionSampled` boolean? If profiling starts after RUM, is it possible to miss the initial `RumSessionRenewedEvent`?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: the RUM session sampling applies [here](https://github.com/DataDog/dd-sdk-android/blob/yl/profiling/RUM-15191-continuous-profiling-scheduler/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt#L79).

So `start` function only launches the continuous profiling scheduler to loop the profiling/cooldown tasks, this depends on the continuous profiling sample rate. RUM session sampling rate is checked every time the active window is about to start.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: `launchProfilingActive` is not only to log the message, it also makes the function early return after the log.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: actually providing the `sdkInstanceNames` here will achieve the goal that we expect,  in `PerfettoProfiler` if it is running with a sdkInstanceName, others are rejected until it finishes.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: I don't think so, if this Rum session is not sampled, we should keep scheduler idling (skip the active window) until a new session is sampled in.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: removed the default value

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
ambushwork: removed

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -103,8 +119,13 @@ internal class ProfilingFeature(
...
 
     private fun onTtidEvent(event: ProfilerStopEvent.TTID) {
-        if (ttidRumContext == null) {
-            ttidRumContext = event.rumContext
+        if (ttidEventReceived) return // already handled
+
+        ttidEventReceived = true
+        ttidRumContext = event.rumContext
+
+        if (continuousProfilingScheduler?.continuousEnabled != true) {
```
0xnm: Should we maybe rename `continuousEnabled` to just `isEnabled`? or `isScheduling`.

Otherwise naming may imply that _maybe_ the scheduler is running even if `continuousEnabled` is `false`.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
```diff
@@ -180,16 +184,97 @@ class ProfilingFeatureTest {
...
+                applicationLaunchSampleRate = 100f,
+                continuousSampleRate = 0f
+            ),
+            mockProfiler
+        )
+        testedFeature.onInitialize(mockContext)
+
         // When
-        testedFeature.onReceive(fakeTTIDEvent)
+        testedFeature.onReceive(ProfilerStopEvent.TTID())
```
0xnm: seems like we can still rely on forged `fakeTTIDEvent` rather than on deterministic one, it should affect the test?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
```diff
@@ -180,16 +184,97 @@ class ProfilingFeatureTest {
...
+                applicationLaunchSampleRate = 100f,
+                continuousSampleRate = 100f
+            ),
+            mockProfiler
+        )
+        whenever(mockProfiler.isRunning(fakeInstanceName)) doReturn true
+        testedFeature.onInitialize(mockContext)
+
+        // When
+        testedFeature.onReceive(ProfilerStopEvent.TTID())
```
0xnm: why not forge it?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
0xnm: And this is where I'm not entirely getting it, because it means that we keep scheduling even if RUM session wasn't sampled https://github.com/DataDog/dd-sdk-android/blob/18512a4d77cacbec17b7ef4326eb3dc5b8ca25ec/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt#L92-L104

I assumed that if RUM session is not sampled, we shouldn't even continue scheduling cycles/cooldowns. Is it useful to schedule them for nothing?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
0xnm: > actually providing the sdkInstanceNames here will achieve the goal that we expect

yes, but my point is about something different: we rely on the SDK names rather than on SDK instances just because SDK instances are not there yet when we start profiler. It means that if we would design Profiler API for the post-SDK initialization usage, the API could be different.

Also we introduced this for the application launch use-case where we can allow multicast to the multiple SDK cores, but this doesn't really work for the continuous profiler case where profiler run **must** be tied to the RUM session, but since we can afford only single profiler session we can link it to only single SDK core, and we need to adapt the code to this somehow.

And now we are mixing Profiler API (and implementation) for 2 quite different cases, making the code a bit harder to follow.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
0xnm: If I get it right if RUM session is not sampled, this code will keep scheduling a loop `scheduleCooldown -> scheduleNextCycle` for nothing, doing no effective work rather than next cycle delay computation. Why not stop scheduling at all?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
ambushwork: it can be the case, if it happens we will need to read proactively the rum context to know if the session is tracked, I will add a todo to address it later.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: reuse the executerService in `PerfettoProfiler`.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: here I keep it, since we are reusing the `ScheduledExecutorService` from `PerfettoProfiler` we should not `shutdown()`  here, it is shut down when Profiling feature stops.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
ambushwork: renamed

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: Let's defer the sampling rate change once decision is made.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: Create RUM-15324 to address that

kikoveiga (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -43,7 +42,7 @@ import kotlin.random.Random
 @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
 internal class PerfettoProfiler(
     private val timeProvider: TimeProvider,
-    private val profilingExecutor: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+    profilingExecutor: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
```
0xnm:
```suggestion
    override val scheduledExecutorService: ScheduledExecutorService
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -75,8 +74,10 @@ internal class PerfettoProfiler(
             }
         }
 
+    override val scheduledExecutorService = profilingExecutor
```
0xnm:
```suggestion
```

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
```diff
@@ -83,6 +92,10 @@ class ProfilingFeatureTest {
     @Forgery
     private lateinit var fakeConfiguration: ProfilingConfiguration
 
+    @Forgery
+    private lateinit var fakeTTIDRumContext: TTIDRumContext
+    private lateinit var fakeTTID: ProfilerStopEvent.TTID
```
0xnm, line 95:
```suggestion
    @Forgery
    private lateinit var fakeTTID: ProfilerStopEvent.TTID
```

`fakeTTIDRumContext` is not used anywhere, we can just forge `ProfilerStopEvent.TTID` itself, there is a forging factory for it.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
```diff
@@ -100,6 +114,7 @@ class ProfilingFeatureTest {
         whenever(mockEditor.putStringSet(any(), any())) doReturn mockEditor
         whenever(mockEditor.putFloat(any(), any())) doReturn mockEditor
         testedFeature = ProfilingFeature(mockSdkCore, fakeConfiguration, mockProfiler)
+        fakeTTID = ProfilerStopEvent.TTID(fakeTTIDRumContext)
```
0xnm:
```suggestion
```

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
```diff
@@ -0,0 +1,354 @@
...
+    private lateinit var mockFuture: ScheduledFuture<Any>
+
+    private val fakeInstanceName = "test-sdk-instance"
+
+    @BeforeEach
+    fun `set up`() {
+        whenever(mockSdkCore.internalLogger) doReturn mockInternalLogger
+        whenever(mockSdkCore.name) doReturn fakeInstanceName
+        whenever(mockProfiler.scheduledExecutorService) doReturn mockSchedulerExecutor
+        @Suppress("UnsafeThirdPartyFunctionCall")
```
0xnm: there is no need to do this in tests, they are not covered by the rule

0xnm (changes requested):

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    @Test
+    fun `M NOT call profiler start W start() {launch profiling active}`() {
+        // When
+        testedScheduler.start(launchProfilingActive = true)
+
+        // Then
+        verify(mockProfiler, never()).start(any(), any(), any(), any(), any())
+        verify(mockSchedulerExecutor, never()).schedule(any<Runnable>(), any(), any())
```
0xnm: I'm not sure, but probably the safer assert here would be `verifyNoInteractions`, because this test may pass if some other similar methods are actually called with slightly different method signature.

Same for similar lines in the tests below.

0xnm (dismissed):

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
ambushwork: only `mockSchedulerExecutor` is updated here, the other assertions on `mockProfiler` are kept since there are some other interactions on it.

0xnm (approved):

kikoveiga (approved):

PR #3301, opened 2026-03-27
RUM-15323: RUM sends ANR & LongTask events to Profiling Feature

ambushwork:
### What does this PR do?

This PR implements the logic of sending RUM long-task and ANR events to the Profiling feature for continuous profiling correlation.

It also adds the profiling status for these two RUM events.

### Motivation

RUM-15323

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ContinuousProfilingRumContext.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ContinuousProfilingRumContext.kt
@@ -1,0 +1,23 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiling
+
+/**
+ * RUM context attached to continuous profiling events.
+ *
+ * @param applicationId The RUM application ID.
+ * @param sessionId The ID of the active RUM session.
+ * @param viewId The ID of the active RUM view, or null if unavailable.
+ * @param viewName The name of the active RUM view, or null if unavailable.
+ */
+// TODO RUM-15334: Commonize ContinuousProfilingRumContext & TTIDRumContext
+data class ContinuousProfilingRumContext(
+    val applicationId: String,
+    val sessionId: String,
+    val viewId: String?,
+    val viewName: String?
+)
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumAnrEvent.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumAnrEvent.kt
@@ -1,0 +1,20 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiling
+
+/**
+ * Sent by the RUM feature to the profiling feature whenever an ANR is detected.
+ *
+ * @param startMs Start timestamp in milliseconds since epoch (server time-adjusted).
+ * @param durationNs Duration of the ANR in nanoseconds.
+ * @param rumContext RUM context at the time of the ANR.
+ */
+data class RumAnrEvent(
+    val startMs: Long,
+    val durationNs: Long,
+    val rumContext: ContinuousProfilingRumContext
+)
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumLongTaskEvent.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumLongTaskEvent.kt
@@ -1,0 +1,20 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiling
+
+/**
+ * Sent by the RUM feature to the profiling feature whenever a long task is detected.
+ *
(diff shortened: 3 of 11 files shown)


ambushwork:
@codex review

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumLongTaskEvent.kt
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @param startMs Start timestamp in milliseconds since epoch (server time-adjusted).
+ * @param durationNs Duration of the long task in nanoseconds.
+ * @param isFrozenFrame True if the duration exceeds the frozen frame threshold (~700 ms).
+ * @param rumContext RUM context at the time of the long task.
+ */
+data class RumLongTaskEvent(
+    val startMs: Long,
+    val durationNs: Long,
+    val isFrozenFrame: Boolean,
```
0xnm: I'm not sure if it is useful for the `ProfileEvent`, I don't think we are going to send it to the profiling backend

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ContinuousProfilingRumContext.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * RUM context attached to continuous profiling events.
+ *
+ * @param applicationId The RUM application ID.
+ * @param sessionId The ID of the active RUM session.
+ * @param viewId The ID of the active RUM view, or null if unavailable.
+ * @param viewName The name of the active RUM view, or null if unavailable.
+ */
+data class ContinuousProfilingRumContext(
```
0xnm, line 18:
```suggestion
data class RumContext(
```

nit: we can call it just `RumContext`, or `ProfilingRumContext`.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumAnrEvent.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package com.datadog.android.internal.profiling
+
+/**
+ * Sent by the RUM feature to the profiling feature whenever an ANR is detected.
+ *
+ * @param startMs Start timestamp in milliseconds since epoch (server time-adjusted).
+ * @param durationNs Duration of the ANR in nanoseconds. Uses the ANR threshold as a lower
+ *   bound until actual duration measurement is implemented in [ANRDetectorRunnable].
```
0xnm: I don't think we will ever have any duration measurement implemented, because it is impossible to know ANR start and end.

The best we can do is do know the timestamp when we detected ANR and maybe add ~2-3s before and after to be able to label profiling samples with this event, same for Long Task.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ContinuousProfilingRumContext.kt
ambushwork: currently it's to differentiate with `TTIDRumContext`, and I think we can just merge them into one called `ProfilingRumContext` in another PR later, wdyt?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumAnrEvent.kt
ambushwork: I removed the doc, and currently I am using hardcoded `ANR_THRESHOLD_MS = 5000L` in `ANRDetectorRunnable` as the duration.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/RumLongTaskEvent.kt
ambushwork: removed

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ContinuousProfilingRumContext.kt
0xnm: yes, we can simply merge it

ambushwork: added a todo here.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -701,6 +705,29 @@ internal open class RumViewScope(
             rumContext.viewId.orEmpty()
         )
 
+        // region profiling
+        var profilingStatus: ErrorEvent.Profiling? = null
+        if (event.throwable is ANRException) {
```
0xnm: Reporting this becomes tricky, because for non-fatal we don't report them by default for Android 10 and above, for fatal ANRs we have them in the next app run, so we cannot attach them to the profile in the current app session.

Considering this, we need to take out fatal ANRs from this and think how can we close the gap for Android 10 and above. That is why we maybe need to consider leveraging ANR trigger.

ambushwork: updated, only report non-fatal ANR for now, defer fatal ANR into todo list.

0xnm: I don't think we will be able to link fatal ANRs with profiling, because of the way they are reported. My point is that even with non-fatal ANRs we have a gap: reporting them is disabled for APIs above R (Android 11, API 30) by default, and profiling is available only starting from API 35.

So for non-fatal ANR to be linked to the profiling customers have to enable non-fatal ANR reporting (which is noisy).  Even though the code makes sense, it won't work in the default RUM setup. That is why maybe we can consider using profiling triggers at least to be notified about ANR.

ambushwork: yes, this is clearly understood on my side, I will still keep the todo ticket here it can be a good reminder for mentioning the fact of fatal ANR in the documentation and address the non-fatal over Android 10 later at least.

0xnm (approved):

PR #3306, opened 2026-03-30
RUM-15277: Propagate native anonymousId to webview events

kikoveiga:
### What does this PR do?

Propagates the native `anonymousId` into webview RUM and Log events. Even if browser set it on their side, we will overwrite it (we do the same with `applicationId`, for instance).

iOS counterpart: https://github.com/DataDog/dd-sdk-ios/pull/2847

### Motivation

[RUMS-5561](https://datadoghq.atlassian.net/browse/RUMS-5661)

[RUMS-5561]: https://datadoghq.atlassian.net/browse/RUMS-5561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
+++ b/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
@@ -53,11 +53,16 @@ internal class WebViewLogEventConsumer(
     ): JsonObject {
         addDdTags(event, datadogContext)
         correctDate(event, datadogContext)
+        addUserInfo(event, datadogContext)
         if (rumContext != null) {
             event.addProperty(LogAttributes.RUM_APPLICATION_ID, rumContext.applicationId)
             event.addProperty(LogAttributes.RUM_SESSION_ID, rumContext.sessionId)
         }
         return event
+    }
+
+    private fun addUserInfo(event: JsonObject, datadogContext: DatadogContext) {
+        event.add(USR_KEY_NAME, datadogContext.userInfo.toJson())
     }
 
     private fun correctDate(event: JsonObject, datadogContext: DatadogContext) {
@@ -160,6 +165,7 @@ internal class WebViewLogEventConsumer(
         const val DATE_KEY_NAME = "date"
         const val USER_LOG_EVENT_TYPE = "log"
         const val INTERNAL_LOG_EVENT_TYPE = "internal_log"
+        const val USR_KEY_NAME = "usr"
         const val JSON_PARSING_ERROR_MESSAGE = "The bundled web log event could not be deserialized"
         val LOG_EVENT_TYPES = setOf(USER_LOG_EVENT_TYPE)
         internal const val DEFAULT_SAMPLE_RATE = 100f
--- a/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
+++ b/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
@@ -53,7 +53,7 @@ internal class WebViewLogEventConsumer(
     ): JsonObject {
         addDdTags(event, datadogContext)
         correctDate(event, datadogContext)
-        addUserInfo(event, datadogContext)
+        addAnonymousId(event, datadogContext)
         if (rumContext != null) {
             event.addProperty(LogAttributes.RUM_APPLICATION_ID, rumContext.applicationId)
             event.addProperty(LogAttributes.RUM_SESSION_ID, rumContext.sessionId)
@@ -61,8 +61,11 @@ internal class WebViewLogEventConsumer(
         return event
     }
 
-    private fun addUserInfo(event: JsonObject, datadogContext: DatadogContext) {
-        event.add(USR_KEY_NAME, datadogContext.userInfo.toJson())
+    private fun addAnonymousId(event: JsonObject, datadogContext: DatadogContext) {
+        val anonymousId = datadogContext.userInfo.anonymousId ?: return
+        val usr = event.getAsJsonObject(USR_KEY_NAME) ?: JsonObject()
+        usr.addProperty(ANONYMOUS_ID_KEY_NAME, anonymousId)
+        event.add(USR_KEY_NAME, usr)
     }
 
     private fun correctDate(event: JsonObject, datadogContext: DatadogContext) {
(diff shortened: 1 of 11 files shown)


0xnm (dismissed):

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventMapper.kt
```diff
@@ -64,6 +66,8 @@ internal class WebViewRumEventMapper(
             event.add(SESSION_KEY_NAME, session)
         }
 
+        event.add(USR_KEY_NAME, userInfo.toJson())
```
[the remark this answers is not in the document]
kikoveiga: Fixed.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
```diff
@@ -53,13 +53,18 @@ internal class WebViewLogEventConsumer(
...
+        addUserInfo(event, datadogContext)
         if (rumContext != null) {
             event.addProperty(LogAttributes.RUM_APPLICATION_ID, rumContext.applicationId)
             event.addProperty(LogAttributes.RUM_SESSION_ID, rumContext.sessionId)
         }
         return event
     }
 
+    private fun addUserInfo(event: JsonObject, datadogContext: DatadogContext) {
+        event.add(USR_KEY_NAME, datadogContext.userInfo.toJson())
```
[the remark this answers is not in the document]
kikoveiga: Fixed.

0xnm (dismissed):

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
```diff
@@ -53,13 +53,24 @@ internal class WebViewLogEventConsumer(
...
             event.addProperty(LogAttributes.RUM_APPLICATION_ID, rumContext.applicationId)
             event.addProperty(LogAttributes.RUM_SESSION_ID, rumContext.sessionId)
         }
         return event
     }
 
+    private fun addUserInfo(event: JsonObject, datadogContext: DatadogContext) {
+        val usr = event.getAsJsonObject(USR_KEY_NAME) ?: JsonObject()
+        for ((key, value) in datadogContext.userInfo.toJson().asJsonObject.entrySet()) {
+            if (!usr.has(key)) {
```
ambushwork: Question: The logic seems like it overrides only a part of the user info if they are absent and keep the rest of them if present, what's the reason of doing that?

ambushwork (dismissed):

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/log/WebViewLogEventConsumer.kt
kikoveiga: The logic here is that we do not override the information that is already present from the browser side, and just add new fields that are not there (`anonymousId` will be the case for example). I added an image on the pr description for some clarity.

ambushwork: This sounds strange for me, in some corner cases when Browser SDK and native SDK both possess partial user info can lead to a fake composed user info if I am not wrong.

```
// Webview
UserInfo(
    anonymousId = null,
    id = "001" ,
    name = "John" ,
    email = null
) 

// Native
UserInfo(
    anonymousId = "anonymous name",
    id = "002" ,
    name = "Bob" ,
    email = "Bob@datadoghq.com"
) 

// After the logic it becomes
UserInfo(
    anonymousId = "anonymous name",
    id = "001" ,
    name = "John" ,
    email = "Bob@datadoghq.com"
) 

```

kikoveiga: Oh ok I didn't understand previously, thanks! So maybe I should just stick to adding the anonymousId, which is a native field only, wdyt?

ambushwork: it can be an option, or the alternative solution is to completely override the user info of Browser SDK. it's better to check with iOS what is their approach.

kikoveiga: iOS doesn't do it yet as well, I'm supposed to do the same there: [RUM-15279](https://datadoghq.atlassian.net/browse/RUM-15279).

I had tried before to override everything, but it didn't work because some fields disappeared (web-side name and email for example). Tested it in Shopist. Also, codex advised against this in the above resolved comments.

ambushwork: I think codex missed the whole picture when giving the advise, IMO the correct path should be:
* If we detect that two user info are same, by checking id for example, we are safe to merge them
* If not, we need to make a decision that if we want to keep the native one or browser one.

Also it's better to add a unit test to protect this corner case.

kikoveiga: I don't know the exact logic of webviews and browser sdk, but isn't the id set on their side when on a webview? If so, ids will never be the same. I think info from browser and from native is unrelated, and trying to merge them can change behaviour unexpectedly. That's why I was only adding non-existing fields, but the safest is definitely just adding the `anonymousId` (which is only present on the native side).

cdn34dd: The `anonymousId` field is not a mobile field only, you can also have it the browser SDK, I think the difference is that in the browser SDK is disabled by default and we enable it by default on mobile if I'm not mistaken.

I'd say here we should simply add the `anonymousId` if not present to ensure we don't make any breaking changes in minor versions as people can be depending on things working like they do now. If we want to review this at some point we should do so in a major version.

cdn34dd (commented):
I notice that on webview events context also has a `usr` field, and it seems that we're not setting `anonymousId` there:
```
{
  (...)
  "context": {
    "usr": {
      "use_new_checkout_flow": true,
      "name": "harper.harry",
      "id": "1gfdsgsdfg",
      "team": "engineering",
      "use_discount": false,
      "email": "harper.harry@example.com"
    },
    "browser_test": false
  }
 (...)
}
```
I'm not sure if there's any backend functionality that specifically looks for `context.usr`, but maybe the safe option would be to add it also there ? As it seems that if browser were to set `anonymousId` itself, this would also be there.

0xnm (dismissed):

0xnm (approved):

PR #3307, opened 2026-03-31
RUM-15334: Commonize ProfilingRumContext for TTID and Continuous Profiling

ambushwork:
### What does this PR do?

* `ProfilingRumContext` is added to commonize the context of App launch profiling and continuous profiling, no business logic change inside this PR.
* `ProfilerStopEvent` now have two subclasses:
   * `TTID`: report TTID when session is tracked, `ProfilingRumContext` becomes non-null type, requires `vitalId` and optional `vitalName`
   * `TTIDNotTracked`: report TTID but session is not tracked.
* Simplify a bit `ProfilingFeature` logic thanks to the refactoring.
* Session sample checked is moved from `RumSessionScope` into `RumSessionScopeStartupManager`

### Motivation

RUM-15334


dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ProfilerStopEvent.kt
```diff
@@ -13,28 +13,19 @@ sealed class ProfilerStopEvent {
...
-    val vitalId: String,
-    val vitalName: String?,
-    val viewId: String?,
-    val viewName: String?
-)
+    /**
+     * Internal event signalling that TTID has been reached but the RUM session is not tracked
+     * (unsampled session). No profiling data will be written.
+     */
+    object TTIDNotTracked : ProfilerStopEvent()
```
0xnm: we can go even further and rename `ProfilerStopEvent` to `ProfilerEvent`, because in case of TTID it is not necessarily stop anymore. We can also put Long Task and ANR events there as well.

0xnm (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ProfilerStopEvent.kt
ambushwork: Good point, I added a new commit to address this.

0xnm (dismissed):

0xnm (dismissed):

0xnm (approved):

PR #3310, opened 2026-04-01
RUM-15321: Continuous profiling writes are gated on pending RUM events

ambushwork:
### What does this PR do?

Long task events and ANR Error events are forwarded to `ContinuousProfilingScheduler` when they are received by `ProfilingFeature`. Before we try to write the events for continuous profiling, we check if there are pending RUM events, this respects the fact that continuous profiling result should only be uploaded when there are crucial RUM events during the profiling.

Note: `ContinuousProfilingScheduler` only accepts the RUM events when the active window is open (profiler is running), otherwise the events will be dropped since they are not happening during the profiling window.

### Motivation

RUM-15321

### Additional Notes


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -182,7 +189,19 @@ internal class ProfilingFeature(
             }
 
             ProfilingStartReason.CONTINUOUS.value -> {
-                // TODO RUM-15193: write continuous profiling event
+                val scheduler = continuousProfilingScheduler ?: return
+                scheduler.onActiveWindowEnded()
```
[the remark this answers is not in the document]
ambushwork: it is expected, we give some grace delay for RUM feature to send events to Profiling Feature.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -182,7 +189,19 @@ internal class ProfilingFeature(
...
+                val longTasks = scheduler.pendingLongTasks.toList()
+                val anrEvents = scheduler.pendingAnrEvents.toList()
+                if (longTasks.isEmpty() && anrEvents.isEmpty()) return
+                val rumContext = scheduler.latestRumContext ?: return
```
kikoveiga: Why do we write all the events with the same `latestRumContext`? Each `RumLongTaskEvent` and `RumAnrEvent` has its own `event.rumContext` inside, so just wondering why not use that one?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
```diff
@@ -350,5 +358,168 @@ internal class ContinuousProfilingSchedulerTest {
...
+
+        // Then
+        assertThat(testedScheduler.pendingLongTasks).isEmpty()
+        assertThat(testedScheduler.pendingAnrEvents).isEmpty()
+    }
+
+    // endregion
+
+    @Test
+    fun `M drop long task event W receive RUM events  {accumulation window closed}`() {
```
kikoveiga:
```suggestion
    fun `M drop long task event W receive RUM events {accumulation window closed}`() {
```

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
```diff
@@ -350,5 +358,168 @@ internal class ContinuousProfilingSchedulerTest {
         verify(mockFuture).cancel(false)
     }
 
+    @Test
+    fun `M clear pending RUM events W stop()`() {
+        // Given
+        openActiveWindow()
+        testedScheduler.onRumAnrEvent(fakeRumAnrEvent)
+        testedScheduler.onRumLongTaskEvent(fakeRumLongTaskEvent)
```
kikoveiga, line 384: Should we assert that after these calls, the `pending` lists are not empty?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
ambushwork: because `ProfilerEvent` requires single application_id, session_id, view_id even though we attach a collection of RUM events as attachment, in this case we will need to use one RUM context which is the latest one.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
ambushwork: good point, updated

kikoveiga (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -37,6 +41,16 @@ internal class ContinuousProfilingScheduler(
     @Volatile
     private var pendingFuture: ScheduledFuture<*>? = null
 
+    @Volatile
+    internal var latestRumContext: ProfilingRumContext? = null
```
0xnm: How it is supposed to be used? Views may change during the profiling session and `ProfileEvent` can accept array of views.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -50,6 +51,15 @@ internal class ProfilingDataWriter(
             }
     }
 
+    override fun write(
+        profilingResult: PerfettoResult,
+        rumContext: ProfilingRumContext,
```
0xnm: I don't think it should be a single `ProfilingRumContext`, it should be the aggregation here.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
```diff
@@ -356,6 +372,128 @@ class ProfilingFeatureTest {
...
+        // Given
+        testedFeature = ProfilingFeature(
+            mockSdkCore,
+            ProfilingConfiguration(
+                customEndpointUrl = null,
+                applicationLaunchSampleRate = 100f,
+                continuousSampleRate = 100f
+            ),
+            mockProfiler
+        )
```
0xnm: shouldn't we pull this to the common setup method?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/assertj/ProfileEventAssert.kt
```diff
@@ -146,6 +146,16 @@ internal class ProfileEventAssert(actual: ProfileEvent) :
         return this
     }
 
+    fun hasLongTaskIds(expected: List<String>?): ProfileEventAssert {
```
0xnm: seems it is not used anywhere so far? is it supposed to be used in this PR?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ContinuousProfilingSchedulerTest.kt
```diff
@@ -350,5 +358,172 @@ internal class ContinuousProfilingSchedulerTest {
...
+
+        // Then
+        assertThat(testedScheduler.pendingLongTasks).isEqualTo(listOf(fakeRumLongTaskEvent))
+        assertThat(testedScheduler.pendingAnrEvents).isEqualTo(listOf(fakeRumAnrEvent))
```
0xnm: tip: alternatively `containsOnly` can be used to avoid extra `listOf` call

```suggestion
        assertThat(testedScheduler.pendingLongTasks).containsOnly(fakeRumLongTaskEvent)
        assertThat(testedScheduler.pendingAnrEvents).containsOnly(fakeRumAnrEvent)
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: the `views` field will be filled by `RumLongTaskEvent` & `RumAnrEvent` since each of them contain `ProfilingRumContext` with view ids and names. here `latestRumContext` is introduced to hoist the decision of which "applicationId" (unlikely to change though) and "sessionId" (in some corner case it may change) should be used from the collection of RumEvents in case that the list of the events are shuffled.

Alternatively we can remove this field and just use the last or first application id & session id in the list of RUM events, WDYT?

0xnm: Yes, I think we should remove it and use aggregation from RUM events we are going to bundle. Having this property causes confusion.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
ambushwork: `testedFeature` needs to be initialized with a full sample rate configuration in some tests, I added a `fakeAllSampledConfiguration` to reuse.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/assertj/ProfileEventAssert.kt
ambushwork: it belongs to my next PR, it's a miss when splitting, removed.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: removed.

0xnm (approved):

PR #3333, opened 2026-04-02
RUM-15324: Decouple continuous profiling from SDK instance names

ambushwork:
### What does this PR do?

This PR decouples continuous profiling from app-launch profiling by adding a new start(durationMs) overload to the Profiler interface that does not require an SDK instance name. Continuous profiling is global — it runs once across all SDK instances rather than being tied to whichever instance triggered it. If ContinuousProfilingScheduler tries to start a session while PerfettoProfiler is already running (either from another SDK  instance's continuous cycle or from an app-launch session), it will skip the request.

This changes the callback notification behaviour:
* Before: the profiling result for each active window was broadcast only to the SDK instance that started it.
* After: the result is broadcast to all registered SDK instances.

### Motivation

RUM-15324


ambushwork:
Closing this PR, needs to address this issue later

PR #3335, opened 2026-04-03
RUM-15454: Add Profiler status in RUM debug widget

ambushwork:
### What does this PR do?

* Add profiler status indicator into RUM debug widget by reading Profiling context
* Fix a minor bug that when continuous profiling active window starts, the feature context is not updated.

### Motivation

It's convenient for debugging continuous profiling

### Demo

Screenshot 2026-04-03 at 14 52 23

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
@@ -67,7 +67,6 @@ internal class ProfilingFeature(
         // Set the profiling flag in SharedPreferences to profile for the next app launch
         ProfilingStorage.addProfilingFlag(appContext, sdkCore.name)
         sdkCore.setEventReceiver(name, this)
-        // TODO RUM-13678: we need to update context from the actual profiler start call, not from here
         sdkCore.updateFeatureContext(Feature.PROFILING_FEATURE_NAME) { context ->
             context[PROFILER_IS_RUNNING] = profiler.isRunning(sdkCore.name)
         }
@@ -211,6 +210,6 @@ internal class ProfilingFeature(
     companion object {
         private const val UNSUPPORTED_EVENT_TYPE =
             "Profiling feature received an event of unsupported type=%s."
-        private const val PROFILER_IS_RUNNING = "profiler_is_running"
+        internal const val PROFILER_IS_RUNNING = "profiler_is_running"
     }
 }
--- a/features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/DefaultInsightsCollector.kt
+++ b/features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/DefaultInsightsCollector.kt
@@ -8,6 +8,9 @@ package com.datadog.android.insights.internal
 
 import android.os.Handler
 import android.os.Looper
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureContextUpdateReceiver
+import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.insights.internal.domain.TimelineEvent
 import com.datadog.android.insights.internal.extensions.Mb
 import com.datadog.android.insights.internal.extensions.round
@@ -31,7 +34,7 @@ internal class DefaultInsightsCollector internal constructor(
     updateIntervalMs: Long,
     private val handler: Handler,
     private val platform: Platform
-) : InsightsCollector {
+) : InsightsCollector, FeatureContextUpdateReceiver {
 
     constructor(
         maxSize: Int = 50,
@@ -66,6 +69,8 @@ internal class DefaultInsightsCollector internal constructor(
     internal var slowFramesRate: Double = Double.NaN
         private set
     internal var cpuTicksPerSecond: Double = Double.NaN
+        private set
+    internal var isProfilingRunning: Boolean = false
         private set
 
     override var maxSize: Int = maxSize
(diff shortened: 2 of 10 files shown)


features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/platform/Platform.kt
```diff
@@ -17,4 +20,14 @@ internal class Platform {
 
     @Suppress("PreferTimeProvider")
     fun nanoTime(): Long = System.nanoTime()
+
+    fun isProfilingRunning(): Boolean {
+        val sdkCore = Datadog.getInstance() as? FeatureSdkCore ?: return false
+        val context = sdkCore.getFeatureContext(Feature.PROFILING_FEATURE_NAME, useContextThread = false)
+        return context[PROFILER_IS_RUNNING] as? Boolean ?: false
+    }
```
0xnm: We shouldn't do it this way: `sdkCore.getFeatureContext` is a blocking call and also you are binding to the default SDK instance (and this behaviour is different from the rest of the code in this module, which binds to the particular SDK core).

The problem here is that all insights are RUM-oriented and we try to bring profiling state without an explicit dependency on the Profiling module.

What you probably can do though is to bind Insights Collector to the Feature Context Update receiver and read context from there.

ambushwork: Good point, now `DefaultInsightsCollector` is bind to SDK instance through the new function `bindSdkCore`

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/insights/InsightsCollector.kt
```diff
@@ -94,6 +96,14 @@ interface InsightsCollector {
...
      */
     fun onSlowFrameRate(rate: Double?)
+
+    /**
+     * Binds this collector to the given [sdkCore] so it can subscribe to feature context updates.
+     * Called once by the RUM feature during initialization.
+     *
+     * @param sdkCore the specific SDK core instance this widget is bound to.
+     */
+    fun bindSdkCore(sdkCore: FeatureSdkCore) {}
```
0xnm:
```suggestion
    fun bindSdkCore(sdkCore: FeatureSdkCore)
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -67,7 +67,6 @@ internal class ProfilingFeature(
         // Set the profiling flag in SharedPreferences to profile for the next app launch
         ProfilingStorage.addProfilingFlag(appContext, sdkCore.name)
         sdkCore.setEventReceiver(name, this)
-        // TODO RUM-13678: we need to update context from the actual profiler start call, not from here
```
0xnm, line 70: probably the comment still makes sense?

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/DefaultInsightsCollector.kt
```diff
@@ -128,6 +133,20 @@ internal class DefaultInsightsCollector internal constructor(
...
     }
 
+    override fun bindSdkCore(sdkCore: FeatureSdkCore) {
+        sdkCore.setContextUpdateReceiver(this)
+    }
+
+    override fun onContextUpdate(featureName: String, context: Map<String, Any?>) {
+        if (featureName == Feature.PROFILING_FEATURE_NAME) {
+            val running = context[PROFILER_IS_RUNNING] as? Boolean ?: false
+            handler.post {
```
0xnm, line 147: seems like the common approach in this class is to call `withListenersUpdate`, but having more fine-grained approach doesn't hurt anyway

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/DefaultInsightsCollector.kt
```diff
@@ -128,6 +133,20 @@ internal class DefaultInsightsCollector internal constructor(
         slowFramesRate = rate.round(PRECISION)
     }
 
+    override fun bindSdkCore(sdkCore: FeatureSdkCore) {
+        sdkCore.setContextUpdateReceiver(this)
```
0xnm, line 137: should we unsubscribe at some point?

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/DefaultInsightsCollector.kt
```diff
@@ -173,5 +192,8 @@ internal class DefaultInsightsCollector internal constructor(
         internal const val PRECISION = 2
         internal const val GC_COUNT = "art.gc.gc-count"
         internal const val ONE_SECOND_NS = 1_000_000_000L
+
+        // Contract key; must match ProfilingFeature.PROFILER_IS_RUNNING in dd-sdk-android-profiling.
+        private const val PROFILER_IS_RUNNING = "profiler_is_running"
```
0xnm, line 201: for the future: maybe we can move it to the `-internal` module to share though

ambushwork: yes, I wanted to do that but it involves too many places, so I decided to address it in another PR.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
ambushwork: This is outdated so I decide to remove it:
*  the profiling status is already updated both in `ContinuousProfilingScheduler` when it starts, also updated when the callback is called as described in TODO
* but  an update is still needed here for app-launch profiling since the moment when app-launch profiling is started, no context is available for updating, so we need to fetch the profiler status once we initialize the Profiling feature (this is the earliest moment that we can).

features/dd-sdk-android-rum-debug-widget/src/main/kotlin/com/datadog/android/insights/internal/DefaultInsightsCollector.kt
ambushwork: good point, fixed with new `unbindSdkCore` function

0xnm (approved):

PR #3341, opened 2026-04-07
RUM-15193: Implement Continuous Profiling event writting

ambushwork:
### What does this PR do?

This PR implements the part of writing & uploading profiling event with RUM long tasks and error events.

* When writing the events, long tasks and ANR events are serialized to JSON then convert to byte array then combined with perfetto bytes into `ProfilingBatchMetadata`.
* `ProfilingBatchMetadata` has TLV format, preserving the data integrity, a magic byte `DDCP` (DataDog Continuous Profile) at first, followed by the length of the data (perfetto size + RUM events size), then the value.
*  in `ProfilingRequestFactory`, the batch data is parsed as `ProfilingBatchMetadata` if in the first byte match the magic byte `DDCP`, then we add perfetto bytes and RUM events as separate attachments
* RUM events attachment is disabled for now since we are missing prof-backend implementation.

### Motivation

RUM-15193

### Additional Notes


ambushwork:
@codex review

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -117,9 +161,79 @@ internal class ProfilingDataWriter(
...
+        }
+        for (event in anrEvents) {
+            anrIds.add(event.id)
+            event.rumContext.viewId?.let { viewIds.add(it) }
+            event.rumContext.viewName?.let { viewNames.add(it) }
+        }
+        return ProfileEvent(
+            start = formatIsoUtc(profilingResult.start),
+            end = formatIsoUtc(profilingResult.end),
+            attachments = listOf(PERFETTO_ATTACHMENT_NAME, RUM_MOBILE_EVENTS_ATTACHMENT_NAME),
```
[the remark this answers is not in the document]
ambushwork: we have a todo to address it later

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -82,6 +99,33 @@ internal class ProfilingDataWriter(
...
+    private fun buildRawBatchEventContinuous(
+        context: DatadogContext,
+        profilingResult: PerfettoResult,
+        longTaskEvents: List<ProfilerEvent.RumLongTaskEvent>,
+        anrEvents: List<ProfilerEvent.RumAnrEvent>
+    ): RawBatchEvent? {
+        if (longTaskEvents.isEmpty() && anrEvents.isEmpty()) return null
+        val perfettoBytes = readProfilingData(profilingResult.resultFilePath)
+        val firstRumContext =
+            longTaskEvents.firstOrNull()?.rumContext ?: anrEvents.firstOrNull()?.rumContext
```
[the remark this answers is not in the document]
ambushwork: Defer this case to RUM-15481

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -82,6 +99,33 @@ internal class ProfilingDataWriter(
...
+        return if (perfettoBytes == null || perfettoBytes.isEmpty() || firstRumContext == null) {
+            null
```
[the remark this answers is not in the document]
ambushwork: it's expected, if the profile data is corrupted, then we should also drop any RUM events related to it.

detekt_custom_safe_calls.yml
```diff
@@ -685,8 +688,13 @@ datadog:
       - "java.lang.StringBuilder.constructor()"
       - "java.math.BigInteger.toHexString()"
       - "java.math.BigInteger.toLong()"
+      - "java.nio.ByteBuffer.get(kotlin.ByteArray?)"
```
0xnm: docs say this method throws https://developer.android.com/reference/java/nio/ByteBuffer#get(byte[])

detekt_custom_safe_calls.yml
```diff
@@ -685,8 +688,13 @@ datadog:
...
       - "java.nio.ByteBuffer.hashCode()"
+      - "java.nio.ByteBuffer.order(java.nio.ByteOrder?)"
       - "java.nio.ByteBuffer.position()"
+      - "java.nio.ByteBuffer.position(kotlin.Int)"
```
0xnm: this method throws as well https://developer.android.com/reference/java/nio/Buffer#position(int)

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -82,6 +99,33 @@ internal class ProfilingDataWriter(
...
+        } else {
+            val profileEvent = createContinuousProfileEvent(
+                context = context,
+                rumContext = firstRumContext,
+                profilingResult = profilingResult,
+                longTaskEvents = longTaskEvents,
+                anrEvents = anrEvents
+            )
+            val serializedEvent = profileEvent.toJson().toString().toByteArray(Charsets.UTF_8)
+            val rumMobileEventsJson = buildRumMobileEventsJson(longTaskEvents, anrEvents)
```
0xnm, line 121: we need to do the same for the application launch as well, it will be used in the aggregation

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -154,8 +268,15 @@ internal class ProfilingDataWriter(
...
-        private const val PERFETTO_ATTACHMENT_NAME = "perfetto.proto"
+        internal const val RUM_EVENT_KEY_ID = "id"
+        internal const val RUM_EVENT_KEY_START_NS = "start_ns"
+        internal const val RUM_EVENT_KEY_DURATION_NS = "duration_ns"
+        internal const val RUM_EVENT_KEY_LONG_TASKS = "long_tasks"
+        internal const val RUM_EVENT_KEY_ERRORS = "errors"
+        internal const val PERFETTO_ATTACHMENT_NAME = "perfetto.proto"
+        internal const val RUM_MOBILE_EVENTS_ATTACHMENT_NAME = "rum-mobile-events.json"
         private const val OPERATION_TYPE_LAUNCH = "launch"
+        private const val OPERATION_TYPE_CONTINUOUS = "continuous"
```
0xnm, line 267: is it the same value in iOS SDK?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -117,9 +161,79 @@ internal class ProfilingDataWriter(
...
+        context: DatadogContext,
+        rumContext: ProfilingRumContext,
+        profilingResult: PerfettoResult,
+        longTaskEvents: List<ProfilerEvent.RumLongTaskEvent>,
+        anrEvents: List<ProfilerEvent.RumAnrEvent>
+    ): ProfileEvent {
+        val viewIds = ArraySet<String>()
+        val viewNames = ArraySet<String>()
+        val longTaskIds = ArraySet<String>()
+        val anrIds = ArraySet<String>()
```
0xnm: side note: it seems [ArraySet is designed for the memory-efficiency](https://developer.android.com/reference/android/util/ArraySet), trading off access speed for that. Is there a benefit to use it compared to the stdlib container given we are going to have ~dozen of IDs?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingRequestFactory.kt
```diff
@@ -64,16 +65,31 @@ internal class ProfilingRequestFactory(
...
-                PERFETTO_FILE_NAME,
-                PERFETTO_FILE_NAME,
-                rawEvent.metadata.toRequestBody(CONTENT_TYPE_BINARY_TYPE)
-            )
             multipartBodyBuilder.addFormDataPart(
                 EVENT_NAME_FORM_KEY,
                 EVENT_FILE_NAME,
                 rawEvent.data.toRequestBody(CONTENT_TYPE_JSON_TYPE)
             )
+            val batchMetadata = ProfilingBatchMetadata.fromBytesOrNull(rawEvent.metadata)
```
0xnm: it is not batch metadata, it is event metadata

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingRequestFactory.kt
```diff
@@ -64,16 +65,31 @@ internal class ProfilingRequestFactory(
     private fun buildRequestBody(batchData: List<RawBatchEvent>): RequestBody {
         val multipartBodyBuilder = MultipartBody.Builder().setType(MultipartBody.FORM)
         batchData.forEach { rawEvent ->
```
0xnm: I think we may hit a bug here actually in case if batch contains several `ProfileEvent`: we will have multiple parts with the same name

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingRequestFactory.kt
```diff
@@ -64,16 +65,31 @@ internal class ProfilingRequestFactory(
...
+                    RUM_MOBILE_EVENTS_FILE_NAME,
+                    batchMetadata.rumMobileEventsBytes.toRequestBody(CONTENT_TYPE_JSON_TYPE)
+                )*/
+            } else {
+                multipartBodyBuilder.addFormDataPart(
+                    PERFETTO_FILE_NAME,
+                    PERFETTO_FILE_NAME,
+                    rawEvent.metadata.toRequestBody(CONTENT_TYPE_BINARY_TYPE)
+                )
+            }
```
0xnm: code here https://github.com/DataDog/dd-sdk-android/pull/3341/files#diff-b3d729dbeeecb97158e0c4935775da9fce265dba505686c9f73f4603130f826bR188 doesn't acknowledge optional nature of `RUM_MOBILE_EVENTS_FILE_NAME` though and always adds the filename.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -117,9 +161,79 @@ internal class ProfilingDataWriter(
...
+            val longTasksArray = JsonArray()
+            longTasks.forEach { event ->
+                val longTask = JsonObject()
+                longTask.addProperty(RUM_EVENT_KEY_ID, event.id)
+                longTask.addProperty(RUM_EVENT_KEY_START_NS, TimeUnit.MILLISECONDS.toNanos(event.startMs))
+                longTask.addProperty(RUM_EVENT_KEY_DURATION_NS, event.durationNs)
+                longTasksArray.add(longTask)
+            }
+            json.add(RUM_EVENT_KEY_LONG_TASKS, longTasksArray)
+        }
```
0xnm: I think it is worth to define JSON schema and generate a model out of it, it will help here and also in tests

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: no specific concern here, let's change to HashSet here

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: This is defined in our latest telemetry spec, iOS hasn't adapted to it yet.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingRequestFactory.kt
ambushwork: indeed, but here the fix is a bit tricky, since profiling intake accepts only one profile per request, there is no workaround in `ProfilingRequestFactory` to address it properly,  maybe we can set `maxItemsPerBatch` to 1 and let `batchData` here always be single?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingRequestFactory.kt
ambushwork: removed `RUM_MOBILE_EVENTS_ATTACHMENT_NAME` for now  in `ProfilingDataWriter` and marked with TODO there

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: updated

detekt_custom_safe_calls.yml
ambushwork: Moved to [detekt_custom_unsafe_calls.yml‎](https://github.com/DataDog/dd-sdk-android/pull/3341/changes#diff-e95d2a14a6dbf968950611a86b712467fb46f5c20908d5e63b1b49c026ad9d88)

detekt_custom_safe_calls.yml
ambushwork: Moved to [detekt_custom_unsafe_calls.yml‎](https://github.com/DataDog/dd-sdk-android/pull/3341/changes#diff-e95d2a14a6dbf968950611a86b712467fb46f5c20908d5e63b1b49c026ad9d88)

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: it needs some changes in other places, defer this to RUM-15497

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingRequestFactory.kt
0xnm: yes, it may be a solution

ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingRequestFactory.kt
```diff
@@ -63,16 +64,31 @@ internal class ProfilingRequestFactory(
     @Suppress("UnsafeThirdPartyFunctionCall") // Caught in the caller
     private fun buildRequestBody(batchData: List<RawBatchEvent>): RequestBody {
         val multipartBodyBuilder = MultipartBody.Builder().setType(MultipartBody.FORM)
-        batchData.forEach { rawEvent ->
+        // `maxItemsPerBatch` is set to 1 for ProfilingFeature, batchData has single RawBatchEvent in the list.
+        val rawBatchEvent = batchData.firstOrNull() ?: return multipartBodyBuilder.build()
```
0xnm: In case of `null` we shouldn't even to attempt to upload anything, we should throw. Ideally we also need to check if we have multiple items in the `batchData`, and if it is the case we need to log a warning, it will simplify debugging.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/domain/ProfilingBatchMetadata.kt
```diff
@@ -0,0 +1,70 @@
...
+                val perfettoBytes = ByteArray(perfettoLength)
+                @Suppress("UnsafeThirdPartyFunctionCall") // length guarded by perfettoLength <= remaining() check above
+                buffer.get(perfettoBytes)
+                val rumMobileEventsBytes = ByteArray(buffer.remaining())
+                @Suppress("UnsafeThirdPartyFunctionCall") // size == remaining(), cannot underflow
+                buffer.get(rumMobileEventsBytes)
+                return ProfilingBatchMetadata(perfettoBytes, rumMobileEventsBytes)
+            } catch (@Suppress("TooGenericExceptionCaught", "SwallowedException") e: Exception) {
+                return null
+            }
```
0xnm, line 75: this will allow to remove `ReturnCount` suppression

```suggestion
            return try {
                @Suppress("UnsafeThirdPartyFunctionCall") // position is guarded by MIN_HEADER_SIZE check above
                val buffer = ByteBuffer.wrap(bytes).apply {
                    order(ByteOrder.BIG_ENDIAN)
                    position(MAGIC.size)
                }
                val perfettoLength = buffer.int
                if (perfettoLength < 0 || perfettoLength > buffer.remaining()) return null
                val perfettoBytes = ByteArray(perfettoLength)
                @Suppress("UnsafeThirdPartyFunctionCall") // length guarded by perfettoLength <= remaining() check above
                buffer.get(perfettoBytes)
                val rumMobileEventsBytes = ByteArray(buffer.remaining())
                @Suppress("UnsafeThirdPartyFunctionCall") // size == remaining(), cannot underflow
                buffer.get(rumMobileEventsBytes)
                ProfilingBatchMetadata(perfettoBytes, rumMobileEventsBytes)
            } catch (@Suppress("TooGenericExceptionCaught", "SwallowedException") e: Exception) {
                null
            }
```

nit: also it is better to not have `TooGenericExceptionCaught` and `SwallowedException`. Let's split it into dedicated exception handling branches and also log exception thrown.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingDataWriterTest.kt
```diff
@@ -211,4 +215,189 @@ internal class ProfilingDataWriterTest {
...
+        val errors = checkNotNull(rumMobileEvents.errors)
+        assertThat(errors).hasSize(1)
+        assertThat(errors[0].id).isEqualTo(fakeAnrEvent.id)
+        assertThat(errors[0].startNs).isEqualTo(TimeUnit.MILLISECONDS.toNanos(fakeAnrEvent.startMs))
+        assertThat(errors[0].durationNs).isEqualTo(fakeAnrEvent.durationNs)
+        val longTasks = checkNotNull(rumMobileEvents.longTasks)
+        assertThat(longTasks).hasSize(1)
+        assertThat(longTasks[0].id).isEqualTo(fakeLongTask.id)
+        assertThat(longTasks[0].startNs).isEqualTo(TimeUnit.MILLISECONDS.toNanos(fakeLongTask.startMs))
+        assertThat(longTasks[0].durationNs).isEqualTo(fakeLongTask.durationNs)
```
0xnm: nit: I would suggest creating assert object for this as well

0xnm (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/domain/ProfilingBatchMetadata.kt
ambushwork: I keep the "ReturnCount" Suppress since we have more return above, changing to this is not enough to remove it.

0xnm (approved):

PR #3342, opened 2026-04-07
RUM-15138: Rebase trace sample rate against RUM session sample rate for correlated cross-product sampling

hamorillo:
### What does this PR do?

`DeterministicTraceSampler` now applies cross-product rebasing using `sessionSampleRate × traceSampleRate / 100` whenever a valid session sample rate is available. This rebasing is provider-driven and no longer gated by `rum.session.id` on the span.

The session sample rate is propagated from `RumSessionScope` → `RumContext` → SDK feature context. `ApmNetworkInstrumentationConfiguration` reads it at instrumentation setup time via a `sessionSampleRateProvider` lambda and passes it to `DeterministicTraceSampler`. A compatibility fallback still reads `session_sample_rate` from span tags for integrations that bypass `ApmNetworkInstrumentationConfiguration`.

`_dd.agent_psr` now reports the same effective sample rate used by sampling decisions in instrumentation paths, including rebased rates when applicable.

### Motivation

Previously, rebasing behavior and effective-rate reporting were inconsistent across paths. This change makes rebasing policy explicit and consistent: when a session sample rate is available, trace sampling is rebased to preserve the intended cross-product relationship between session and trace sampling.

### Additional Notes

**Session rate stability:** `RumSessionScope` snapshots `sessionSampleRate` once at construction time rather than calling `getSampleRate()` on every `getRumContext()` invocation, so session-level metadata remains stable for the scope's lifetime.

**`RumContextKeys`:** A new internal object centralises the `session_sample_rate` key in the trace module.

### Follow-up

❓ 💡 We should consider, in the next major version, making `DeterministicTraceSampler` non-public so all instances are SDK-controlled. That would let us remove the span-tag compatibility fallback (`session_sample_rate`) and simplify sampling to a single internal effective-rate source.


features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/RumContextTest.kt
```diff
@@ -28,4 +29,18 @@ internal class RumContextTest {
         // Then
         assertThat(anotherRumContext).isEqualTo(fakeRumContext)
     }
+
+    @Test
+    fun `M parse session sample rate W fromFeatureContext() {value is Double}`() {
+        // Given
+        val featureContext = mapOf<String, Any?>(
+            RumContext.SESSION_SAMPLE_RATE to 42.5
```
ambushwork: we can use random value from forge

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/RumContextTest.kt
```diff
@@ -28,4 +29,18 @@ internal class RumContextTest {
         // Then
         assertThat(anotherRumContext).isEqualTo(fakeRumContext)
     }
+
+    @Test
+    fun `M parse session sample rate W fromFeatureContext() {value is Double}`() {
```
ambushwork: the test is not testing "{value is Double}"

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScopeTest.kt
```diff
@@ -478,6 +488,43 @@ internal class RumSessionScopeTest {
         assertThat(context.viewId).isEqualTo(fakeParentContext.viewId)
     }
 
+    @Test
+    fun `M keep session sample rate stable W getRumContext() {sampler rate changes mid-session}`(
+        forge: Forge
+    ) {
+        // Given
+        whenever(mockSessionSampler.getSampleRate()).thenReturn(100f, 20f)
```
ambushwork: you can use random value from forge here

ambushwork (dismissed):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
```diff
@@ -8,24 +8,51 @@ package com.datadog.android.trace
...
+ * and deterministic decisions.
+ *
+ * @param sampleRateProvider Provider for the trace sample rate. Called each time the sampling
+ * decision needs to be made. Values must be in the range [0;100].
+ * @param sessionSampleRateProvider Provider for the RUM session sample rate used to rebase the
+ * trace sample rate for correlated cross-product sampling. Defaults to 100 (no rebasing).
  */
-open class DeterministicTraceSampler(
-    sampleRateProvider: () -> Float
+open class DeterministicTraceSampler private constructor(
```
hamorillo: ❓ 💡 `open` on this class created some friction — external subclasses silently bypass the `sessionSampleRateProvider` wiring and can't access `getSampleRate(DatadogSpan)`, both of which are needed for correct rebasing. Worth making this `internal` or at least `final` in the next major version. Or do we have a good reason to have it `open`. Wdyt?

hamorillo:
@codex review

hamorillo:
Thanks for your initial review, @ambushwork! While reviewing [this comment](https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3051310816)￼, I tried to simplify and improve the changes, so it would be nice if you could take a second pass. 🙏

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -241,10 +247,21 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
 
             val tracerProvider = TracerProvider(localTracerFactory, globalTracerProvider)
 
+            val sdkRef = SdkReference(sdkInstanceName)
+            val sessionSampleRateProvider: () -> Float = {
+                resolveSessionSampleRate(
+                    (sdkRef.get() as? InternalSdkCore)
+                        ?.getFeatureContext(Feature.RUM_FEATURE_NAME)
```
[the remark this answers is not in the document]
hamorillo: Suggestion applied!

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
```diff
@@ -34,11 +57,80 @@ open class DeterministicTraceSampler(
...
+            val rebased = traceSampleRate * sessionSampleRate / DeterministicSampler.SAMPLE_ALL_RATE
+            rebased.coerceAtMost(DeterministicSampler.SAMPLE_ALL_RATE)
+        } else {
+            traceSampleRate
+        }
+    }
+
+    /** @inheritDoc */
+    override fun sample(item: DatadogSpan): Boolean {
+        val sampleRate = getSampleRate(item)
```
aleksandr-gringauz: This looks like exactly the same code as `DeterministicSampler.sample`. Is it possible not to have the same algorithm implemented twice?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
aleksandr-gringauz: @satween might be the best person to discuss this with.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
hamorillo: Good catch! 

I would say to do that we need to extract that to a "_public_" place (or at least accesible for subclasses) for example:

```kotlin
DeterministicSampler.kt

protected fun computeSamplingDecision(sampleRate: Float, id: ULong): Boolean {
    return when {
        sampleRate >= SAMPLE_ALL_RATE -> true
        sampleRate <= 0f -> false
        else -> {
            val hash = id * SAMPLER_HASHER
            val threshold = (MAX_ID.toDouble() * sampleRate / SAMPLE_ALL_RATE).toULong()
            hash < threshold
        }
    }
}

---
DeterministicSampler.kt

override fun sample(item: T): Boolean =
    computeSamplingDecision(getSampleRate(), idConverter(item))
    
---

DeterministicTraceSampler.kt    
override fun sample(item: DatadogSpan): Boolean =
    computeSamplingDecision(getSampleRate(item), SpanSamplingIdProvider.provideId(item))    

```

The reason the `sample` override exists at all is to call` getSampleRate(item)` instead of `getSampleRate()` — the item-aware version is needed to support the legacy instantiation path where no `sessionSampleRateProvider` is wired up and the session rate is read from the span tag instead. The parent's `sample` would silently miss that fallback.

Unless you consider this a blocker, I'd prefer to keep the duplication for now and avoid widening the API surface. In a future major version, if the legacy constructors can be removed, the override could be dropped entirely and the parent's sample would suffice.

Wdyt?

aleksandr-gringauz: > I'd prefer to keep the duplication for now and avoid widening the API surface

Can you make `computeSamplingDecision` in your proposed solution `internal` instead of `protected` and make it some standalone function not tied to `DeterministicSampler` or `DeterministicTraceSampler`? This way you don't have to change the API of `DeterministicSampler` at all.

hamorillo: As discussed I've moved the `computeSamplingDecision` to `dd-sdk-android-internal`.

aleksandr-gringauz (dismissed):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -241,10 +247,21 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
 
             val tracerProvider = TracerProvider(localTracerFactory, globalTracerProvider)
 
+            val sdkRef = SdkReference(sdkInstanceName)
+            val sessionSampleRateProvider: () -> Float = {
+                resolveSessionSampleRate(
+                    (sdkRef.get() as? InternalSdkCore)
+                        ?.getFeatureContext(Feature.RUM_FEATURE_NAME, useContextThread = false)
```
0xnm: `getFeatureContext` is a dangerous call, because it involves locking. Maybe it is better to use `FeatureContextUpdateReceiver`?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
satween: `open` is the legacy came from the V2 version of the SDK. Please add todo with link to V4 ticket.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
```diff
@@ -35,11 +36,18 @@ internal fun DatadogSpan.applyPriority(isSampled: Boolean, traceSampler: Sampler
...
             AGENT_PSR_ATTRIBUTE,
-            (traceSampler.getSampleRate() ?: ZERO_SAMPLE_RATE) / ALL_IN_SAMPLE_RATE
+            (traceSampler.effectiveSampleRate(this) ?: ZERO_SAMPLE_RATE) / ALL_IN_SAMPLE_RATE
         )
     }
 }
 
+internal fun Sampler<DatadogSpan>.effectiveSampleRate(span: DatadogSpan): Float? {
+    return when (this) {
+        is DeterministicTraceSampler -> getSampleRate(span)
```
satween, line 48: I think this is not the best API design, as this layer becomes aware of the implementation of the `Sampler` interface. 

Maybe we should  add `getSampleRate(T)` method  with the default implementation with fallback to `getSampleRate()`? 

this will keep implementation isolated from the caller side + wont break backward compatibility. And in `DeterministicTraceSampler` you could implement independent behaviour.  WDYT?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextKeys.kt
```diff
@@ -0,0 +1,12 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.trace.internal
+
+internal object RumContextKeys {
+    // Must match RumContext.SESSION_SAMPLE_RATE in dd-sdk-android-rum.
+    internal const val SESSION_SAMPLE_RATE = "session_sample_rate"
```
satween: Please move it to the `LogAttributes` as they used for the RUM context propagation anyway.  We should move them out of the `LogAttributes` for sure, but it should be made with the major version update. Before that moment,  let's reduce files that holds constants for RUM context propagation

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/sampling/DeterministicSampling.kt
```diff
@@ -0,0 +1,30 @@
...
+private const val MAX_ID: ULong = 0xFFFFFFFFFFFFFFFFUL
+
+/**
+ * Computes a deterministic sampling decision based on the given sample rate and identifier.
+ *
+ * @param sampleRate the sample rate in the range [0, 100].
+ * @param id a stable numerical identifier derived from the item being sampled.
+ * @return true if the item should be sampled, false otherwise.
+ */
+fun computeSamplingDecision(sampleRate: Float, id: ULong): Boolean {
```
satween, line 20: why did you moved this code out of the `DeterministicSampler` ? It's seems like only descendants of it are using it...

satween (commented):
Two additional things to check before proceed: 
1. `ApmInstrumentation` is an experimental API that we gonna move to from `TracingInterceptor`/`DatadogInterceptor` and that already used in `Cronet`. But untill then - you have to support `TracingInterceptor`/`DatadogInterceptor` as well. 

2. Note that there is [another effective sampling](https://github.com/DataDog/dd-sdk-android/blob/bc97e2586eec14445636346ffac4f7abf6f3afbd/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt#L537) calculation logic that is applied to the telemetry event. Make sure that your changes properly works with that one

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: Codex mentioned something similar in this comment https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3062680182 so I added `useContextThread = false`.

IIUC it still acquires a `readLock` which is where you see the danger, right? Anyway, let me try your suggestion as it looks safer.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/sampling/DeterministicSampling.kt
hamorillo: You can see the discussion: https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3063336241

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
hamorillo: This is a good point. I totally agree that this is far from an ideal solution. 

I considered that option before implementation but adding `getSampleRate(T)` on the `Sampler` interface sounds like an unnecessary refactor considering that and some point we'll be able to get rid of that method. We only need `getSampleRate(T)` due to the legacy classes but when remove and make those private/internal we won't need that method anymore. 

The extension function with the when (this is ...) check is admittedly not clean, but it keeps the coupling contained and co-located with the V4 TODO, rather than leaving a footprint on the interface.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: Suggestion applied!

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextKeys.kt
hamorillo: Done c8f50f95c6db03cf996bca81d330670d46a64e77

hamorillo (commented):
Thanks for your review @satween. Really appreciate it.

> But until then - you have to support TracingInterceptor/DatadogInterceptor as well.

This part should be already covered as they are using `com.datadog.android.trace.DeterministicTraceSampler` and we've implemented the `getSampleRate(item: DatadogSpan): Float` to cover that case where we need to extract the RUM session sample rate from the span. 

While reviewing your comment is true that I found an issue where we weren't sending the right sample rate in `_dd.agent_psr` and `_dd.rule_psr`. Should be fixed [here](d14f89ae527c1ccda0f9749541431ac564fd596f) with a similar (not ideal) approach. For more context see: https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3073951824

> Note that there is [another effective sampling](https://github.com/DataDog/dd-sdk-android/blob/bc97e2586eec14445636346ffac4f7abf6f3afbd/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/telemetry/internal/TelemetryEventHandler.kt#L537) calculation logic that is applied to the telemetry event. Make sure that your changes properly works with that one

I would say, that part shouldn't be affected by the changes in this PR.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumContext.kt
```diff
@@ -89,6 +93,8 @@ internal data class RumContext(
             val hasReplay = featureContext[HAS_REPLAY] as? Boolean ?: false
             val viewTimestamp = featureContext[VIEW_TIMESTAMP] as? Long ?: 0L
             val viewTimestampOffset = featureContext[VIEW_TIMESTAMP_OFFSET] as? Long ?: 0L
+            val sessionSampleRate = (featureContext[SESSION_SAMPLE_RATE] as? Number)
+                ?.toFloat() ?: FULL_SESSION_SAMPLE_RATE
```
0xnm, line 97: just curious: is there a motivation behind using 100% as a fallback, and not 0%?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
```diff
@@ -280,6 +282,7 @@ internal class RumSessionScope(
 
     private fun renewSession(time: Time, reason: StartReason) {
         val newSessionId = UUID.randomUUID().toString()
+        sessionSampleRate = sessionSampler.getSampleRate() ?: RumContext.FULL_SESSION_SAMPLE_RATE
```
0xnm: Why do we need to set it here? normally, session sample rate is static in RUM and is defined in the configuration, there is no API to provide something dynamic.

Is it some protection for the future?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -132,7 +137,7 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
      * By default it is a sampler accepting 100% of the traces.
      */
     fun setTraceSampler(traceSampler: Sampler<DatadogSpan>) = apply {
-        this.traceSampler = traceSampler
```
0xnm, line 135: I'm wondering if we need to expose this API long-term. Probably, there are some users, but I'm curious if it is worth to expose it given we have cross-feature sampling requirement now and also if iOS SDK has something similar.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -246,18 +252,39 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
 
             val tracerProvider = TracerProvider(localTracerFactory, globalTracerProvider)
 
+            @Suppress("UnsafeThirdPartyFunctionCall") // AtomicReference initialized with non-null Float constant
+            val cachedSessionSampleRate = AtomicReference(DEFAULT_TRACE_SAMPLE_RATE)
```
0xnm: It seems it means here that while RUM session context update has happened, we say "assume that RUM session sampling rate is DEFAULT_TRACE_SAMPLE_RATE". Unless I miss something, this is wrong? Shouldn't we just say that in this case we use `traceSampleRate` alone and shouldn't set any value in `cachedSessionSampleRate` by default and set it only when context arrived?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
```diff
@@ -34,11 +59,85 @@ open class DeterministicTraceSampler(
...
+    )
+    fun getSampleRate(item: DatadogSpan): Float {
+        // Compatibility path for integrations creating this sampler via public constructors
+        // (no sessionSampleRateProvider). In that case the session sample rate is propagated on
+        // the span context as `session_sample_rate` and must be used for rebasing.
+        val providerRate = sessionSampleRateProvider()
+        val providerRateIsValid = providerRate >= 0f && providerRate < DeterministicSampler.SAMPLE_ALL_RATE
+        val sessionSampleRate = when {
+            providerRateIsValid -> providerRate
+            else -> (item.context().tags[LogAttributes.RUM_SESSION_SAMPLE_RATE] as? Number)?.toFloat()
```
aleksandr-gringauz: Question.

Why don't we always extract the `session_sample_rate` from `context().tags`? If we do it this way, we won't have to deal with https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3073813578.

Or am I missing something?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -70,9 +74,14 @@ class ApmNetworkInstrumentation internal constructor(
...
     internal val localFirstPartyHostHeaderTypeResolver: DefaultFirstPartyHostHeaderTypeResolver,
     private val networkingLibraryName: String,
-    val networkTracingScope: ApmNetworkTracingScope = ApmNetworkTracingScope.ALL
+    val networkTracingScope: ApmNetworkTracingScope = ApmNetworkTracingScope.ALL,
+    internal val sessionSampleRateReceiver: FeatureContextUpdateReceiver? = null
 ) {
     private val rumContextPropagator = RumContextPropagator { internalSdkCore }
+    private val receiverLifecycleLock = Any()
+    private val isClosed = AtomicBoolean(false)
+    private val isRegistered = AtomicBoolean(false)
```
0xnm: if I'm not wrong, `AtomicBoolean` are not really needed here, because these two properties are always called inside `synchronized` block, which guarantees the thread-safety for any properties used inside. On another hand, then these properties would be `var`, so we can keep it.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -218,9 +235,18 @@ class ApmNetworkInstrumentation internal constructor(
         )
     }
 
-    private fun DatadogSpan.isSampled(request: HttpRequestInfo): Boolean =
-        extractRumContext(rumContextPropagator, block = true)
-            .sample(request, traceSampler)
+    /**
+     * Releases this instrumentation and unregisters context receivers from SDK core.
+     */
+    fun close() {
```
0xnm: does it make sense to implement `Closeable` interface to be explicit? Alternatively we may put `FeatureContextUpdateReceiver` on `TracingFeature`, so we can unsubscribe there on the lifecycle method. Otherwise it is easy to miss `close` call.

integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/internal/DatadogCronetEngine.kt
```diff
@@ -63,7 +63,14 @@ internal class DatadogCronetEngine(
 
     override fun getVersionString(): String? = delegate.versionString
 
-    override fun shutdown() = delegate.shutdown()
+    override fun shutdown() {
+        try {
+            delegate.shutdown()
+        } finally {
+            apmNetworkInstrumentation?.close()
```
0xnm: so here it is a danger of approach with `close` on instrumentation: `ApmNetworkInstrumentation` is used also in the `OkHttpIntegrationPlugin`, but we don't call `close` there.

0xnm (commented):
I went quickly through the main files and added some comments.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
hamorillo: This is a good question.

I don’t like the approach of always relying on `context().tags`, because in that case the `sampler` would depend on the span internals. Based on how we’ve designed sampler, I think they should always have the sample rate available, not only when they receive a span.

For that reason, I only implemented the span path to keep backward compatibility, with the idea of removing it in V4 when we can drop the previously mentioned public APIs.

Anyway, I’m happy to discuss further if you think the two paths are unnecessary.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumContext.kt
hamorillo: The main reason was that if we use 0% as a fallback, the rebased sampling would drop everything, so it seemed safer to me. However, now that you’ve mentioned it, I’m not sure if this is what we expect.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumSessionScope.kt
hamorillo: Good catch. We don't need it.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: Totally agree. We can’t remove it right now since it’s part of the public API, but we could consider removing it in V4. I’m not sure if we have data on how many users are using it, though.

If we agree, I'll create a ticket.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: > Shouldn't we just say that in this case we use traceSampleRate alone and shouldn't set any value in cachedSessionSampleRate by default and set it only when context arrived?

This is exactly what the code does; however, the naming is clearly misleading.

We initialize `cachedSessionSampleRate` to `100f`, which, after rebasing, essentially means using traceSampleRate alone.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
hamorillo: Good catch! This was a leftover from a previous approach.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
aleksandr-gringauz: Just some thoughts, for now no strong opinion.

The way RUM context is passed to the Tracing spans is through [injectRumContext](https://github.com/DataDog/dd-sdk-android/blob/e293f09f1c310307626e9fcb1fab3168d4699756/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/RumContextPropagator.kt#L140) which is [called](https://github.com/DataDog/dd-sdk-android/blob/e293f09f1c310307626e9fcb1fab3168d4699756/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/DatadogTracerAdapter.kt#L62) when the span is created. It is done this way so that we capture the RUM Context at exactly the moment the span is created.

If you pass some items of RUM Context through the `FeatureContextUpdateReceiver` and some through the `injectRumContext` mechanism, theoretically they might get out of sync. In our particular case with `sessionSamplingRate` probably this will not happen, because `sessionSamplingRate` is a constant, but still.

The way I see it. We have a span, we need to compute the sampling decision for it. 
1. We extract the RUM Session ID and RUM Sampling rate from the RUM Context that is attached to it.
2. do the rebasing of sampling rates (simple division of 2 numbers).
3. call `computeSamplingDecision`.
4. DONE.
5. we can put the sampling rate from step 2 into `AGENT_PSR_ATTRIBUTE`. 

We don't even need to use `DeterministicTraceSampler` and the `Sampler` interface in general for this.

However, there is a public`setTraceSampler` in `ApmNetworkInstrumentationConfiguration` and `TracingInterceptor` and the public `DeterministicTraceSampler`. As far as I see it makes my proposal impossible to implement without breaking the public API.
1. What if a customer sets their custom sampler? In general it breaks our deterministic sampling and cross-feature sampling rebasing logic. Also mentioned here https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3094346788.
2. What is the point of public `DeterministicTraceSampler`?. Tbh I don't understand how exactly the customers might want to customize this the way it is now in this PR (even the way it is now in `develop`). IIUC there is a plan to at least make it internal in v4 https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3073906759.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
hamorillo: Great suggestion! Thanks! 

I've applied the `FeatureContextUpdateReceiver` on `TracingFeature`. Let me know what you think.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
hamorillo: > without breaking the public API

I would say that’s the main inconvenience of adding cross-sampling in this PR. I haven’t thought deeply about your proposal, but I think it’s doable, as you mentioned, once we get rid of what we’ve been discussing in different threads within the PR.

Anyway, when we reach that point, I think it’s worth considering your suggestion. Thanks for sharing.

0xnm: I think `context.tags` is really handy for that, we already pass RUM session ID there, so having sample rate is a simple extension then. And we use RUM Session ID in the sticky sampling decision (in the `SpanSamplingIdProvider`).

Indeed, `traceSampler` makes the things more complicated, but this is already the case with RUM session ID I believe (I may be wrong, I didn't check the code in details): if we provide custom sampler, we already lose the stickiness-to-RUM-session property we have right now? So nothing new, we can acknowledge such behavior as existing. But then I guess it will require changing `DeterministicSampler` API, because we need to register RUM session provider as well in the constructor signature?

Relying on `context.tags` will simplify the change layout. But we can keep both paths if relying on span internals is a concern.

> Based on how we’ve designed sampler, I think they should always have the sample rate available, not only when they receive a span.

However, sampling decision is made per-span. With HBS (Head-based sampling) the root span gets the sampling decision and child spans follow it.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
```diff
@@ -34,11 +59,85 @@ open class DeterministicTraceSampler(
...
+        }
+
+        return if (sessionSampleRate != null &&
+            sessionSampleRate >= 0f &&
+            sessionSampleRate < DeterministicSampler.SAMPLE_ALL_RATE
+        ) {
+            val rebased = super.getSampleRate() * sessionSampleRate / DeterministicSampler.SAMPLE_ALL_RATE
+            if (rebased > DeterministicSampler.SAMPLE_ALL_RATE) DeterministicSampler.SAMPLE_ALL_RATE else rebased
```
0xnm: very nit: maybe use `coerceAtMost` as above?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -88,6 +98,10 @@ class ApmNetworkInstrumentation internal constructor(
                 WARNING_TRACING_NO_HOSTS.format(Locale.US, networkingLibraryName)
             }
         }
+        sessionSampleRateRef?.let { ref ->
```
0xnm: I don't think it is right way to do it: shoot event at the SDK initialization to wire up the things. `sendEvent` is need for cross-module communication without explicit dependency. But here both `ApmNetworkInstrumentation` and `TracingFeature` sit in the same module, so you can simply do `sdkCore.getFeature(Feature.TRACING_FEATURE_NAME).unwrap<TracingFeature>()`.

But if we put the code here, we depend on the first call to get the SDK, which _in theory_ can happen after SDK is initialized, but before Tracing feature registered? 🤔

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/TracingFeature.kt
```diff
@@ -56,17 +66,54 @@ internal class TracingFeature(
...
+                sessionSampleRateRefs.forEach { it.set(rate) }
+            }
+        }
+    }
+
+    // endregion
+
+    // region FeatureEventReceiver
+
+    override fun onReceive(event: Any) {
```
0xnm: this is not really needed by my previous comment

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -246,18 +251,26 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
 
             val tracerProvider = TracerProvider(localTracerFactory, globalTracerProvider)
 
+            @Suppress("UnsafeThirdPartyFunctionCall") // AtomicReference initialized with non-null Float constant
+            val cachedSessionSampleRate = AtomicReference(TracingFeature.NO_SESSION_REBASING_RATE)
+            val sessionSampleRateRef: AtomicReference<Float>? =
+                if (customTraceSampler == null) cachedSessionSampleRate else null
+            val effectiveSampler = customTraceSampler
+                ?: DeterministicTraceSampler(traceSampleRate) { cachedSessionSampleRate.get() }
```
0xnm: in theory you can simply do ` { sdkCore.getFeature(...).unwrap<TracingFeature>().rumSessionSampleRate }` here instead of `{ cachedSessionSampleRate.get() }`, probably not very expensive call (can be profiled).

but maybe we can simply rely on the `context.tags` alone.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
aleksandr-gringauz: > But then I guess it will require changing DeterministicSampler API, because we need to register RUM session provider as well in the constructor signature?

One of the problems IIUC is that we have to implement `DeterministicTraceSampler.getSampleRate()`. This method doesn't accept the `span`, so the only way to have the RUM sample rate there is by passing it into the constructor of `DeterministicTraceSampler`.

hamorillo: > Relying on context.tags

You both mentioned this approach, so I’m exploring that option again. It’s probably doable by returning `traceSampleRate` in `DeterministicTraceSampler.getSampleRate()`, which could be correct.

> We don't even need to use DeterministicTraceSampler and the Sampler interface in general for this.

We’ll need to keep the casting at least until V4, where we can replace the Sampler with a different interface (or remove it, as @aleksandr-gringauz mentioned).

Let me explore that path. Thanks, both.

hamorillo: I've switched to always reading `session_sample_rate` from `context.tags `(the same mechanism already used for RUM_SESSION_ID by `SpanSamplingIdProvider`). This allowed removing `SessionSampleRateRegistrationEvent`, the `FeatureContextUpdateReceiver/FeatureEventReceiver`on TracingFeature, the AtomicReference caching, and the sendEvent wiring.

`DeterministicTraceSampler.getSampleRate()`(no-arg) now returns the raw `traceSampleRate`. All paths that need the rebased rate (`_dd.agent_psr`, `_dd.rule_psr`) already dispatch to `getSampleRate(span)` via the `effectiveSampleRate()` bridge, so no call site is affected.

In V4, we can revisit the `Sampler<DatadogSpan>` interface to support per-span rate calculation natively, removing the need for the `effectiveSampleRate()` casting dispatch.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: > but maybe we can simply rely on the context.tags alone.

See https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3117980659

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
hamorillo: See https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3117980659

features/dd-sdk-android-trace/src/testFixtures/kotlin/com/datadog/android/trace/utils/RumContextTestsUtils.kt
```diff
@@ -19,12 +19,19 @@ object RumContextTestsUtils {
...
 
-    fun Forge.aRumContext(sessionId: Long? = null) = buildMap {
+    fun Forge.aRumContext(
+        sessionId: Long? = null,
+        sessionSampleRate: Float? = null
+    ) = buildMap {
         put(RUM_CONTEXT_VIEW_ID, getForgery<UUID>().toString())
         put(RUM_CONTEXT_ACTION_ID, getForgery<UUID>().toString())
         put(RUM_CONTEXT_APPLICATION_ID, getForgery<UUID>().toString())
+        put(SESSION_SAMPLE_RATE_KEY, sessionSampleRate ?: aFloat(min = 0f, max = 100f))
```
0xnm: nit: should it be called `RUM_CONTEXT_SESSION_SAMPLE_RATE` for the consistency with lines above?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -218,10 +220,6 @@ class ApmNetworkInstrumentation internal constructor(
         )
     }
 
-    private fun DatadogSpan.isSampled(request: HttpRequestInfo): Boolean =
```
0xnm, line 221: seems it was handy though

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -121,7 +122,7 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
      * @param sampleRate the sample rate to use (percentage between 0f and 100f, default is 100f).
      */
     fun setTraceSampleRate(@FloatRange(from = 0.0, to = 100.0) sampleRate: Float) = apply {
-        this.traceSampler = DeterministicTraceSampler(sampleRate)
+        this.traceSampleRate = sampleRate
```
0xnm: Do we still need this change and `customTraceSampler ?: DeterministicTraceSampler(traceSampleRate)` below? by looking on the `effectiveSampleRate` implementation I think it is not needed?

Also we may want to document the difference in the behaviour of `setTraceSampleRate` vs `setTraceSampler` that the former relies on the RUM session sampling and latter _may_ drop this connection.

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -258,9 +258,15 @@ internal constructor(
 
             val spanContext = span.context()
             if (spanContext.setSamplingPriority(samplingPriority)) {
+                // TODO RUM-13454 -> Remove the cast to DeterministicTraceSampler - V4 (RUM-15590)
+                @Suppress("DEPRECATION")
                 spanContext.setMetric(
                     AGENT_PSR_ATTRIBUTE,
-                    (traceSampler.getSampleRate() ?: ZERO_SAMPLE_RATE) / ALL_IN_SAMPLE_RATE
+                    (
+                        (traceSampler as? DeterministicTraceSampler)?.getSampleRate(span)
```
0xnm, line 266: nit: it seems there is no test coverage for non-`DeterministicTraceSampler` case?

0xnm (commented):
overall lgtm, I just added some questions

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: Good catch! I’ve reverted those changes, as they’re no longer needed.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/sampling/DeterministicSampler.kt
aleksandr-gringauz, line 93: Looks like these constants are only used in tests now.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
```diff
@@ -34,11 +49,55 @@ open class DeterministicTraceSampler(
...
      *
      * @param sampleRate Sample rate to use.
      */
     constructor(
         @FloatRange(from = 0.0, to = 100.0) sampleRate: Double
     ) : this(sampleRate.toFloat())
+
+    /** @inheritDoc */
+    // TODO RUM-13454 -> Remove the @Suppress when getSampleRate(DatadogSpan) is removed in V4 (RUM-15590)
+    @Suppress("DEPRECATION")
```
aleksandr-gringauz: I think you can use `resolveEffectiveSampleRate` instead of `getSampleRate` here. This way you will not have to suppress the deprecation, because you'll not be using anything deprecated.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/DeterministicTraceSampler.kt
aleksandr-gringauz: > DeterministicTraceSampler.getSampleRate()(no-arg) now returns the raw traceSampleRate

IIUC in your new implementation, if the user overrides `DeterministicTraceSampler.getSampleRate` it will be useless, because it isn't used anywhere. So if some user relies on overriding `getSampleRate` this way, we are breaking it.

But tbh I think we should do it regardless. The probability of someone doing this is really small. And in V4 we should clean-up our public API, get rid of custom samplers in network request tracing entirely.

hamorillo: > if the user overrides DeterministicTraceSampler.getSampleRate it will be useless, because it isn't used anywhere. 

Good point @aleksandr-gringauz . You are right.

> But tbh I think we should do it regardless. The probability of someone doing this is really small. And in V4 we should clean-up our public API, get rid of custom samplers in network request tracing entirely.

I agree. As you mentioned, solving that probably wouldn’t help anyone, and it would make everything even more complex to implement. I would say this is acceptable.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/sampling/DeterministicSampler.kt
hamorillo: Yes, I haven't remove them because they are already part of the public. Adding `@Deprecated` and a `TODO` for V4, though.

aleksandr-gringauz: ok, got it

hamorillo:
After several discussions, we noticed that this wasn't the right approach.

Closing this PR in favor of #3402

PR #3343, opened 2026-04-07
Centralize PROFILER_IS_RUNNING key into internal module

ambushwork:
### What does this PR do?

`PROFILER_IS_RUNNING` was used in a lot of place with duplicated definition, this PR centralizes it into `dd-sdk-android-internal` module to reuse.

### DIFF
--- a/dd-sdk-android-internal/api/apiSurface
+++ b/dd-sdk-android-internal/api/apiSurface
@@ -1,3 +1,5 @@
+object com.datadog.android.internal.FeatureContextKeys
+  const val PROFILER_IS_RUNNING: String
 interface com.datadog.android.internal.attributes.LocalAttribute
   enum Key
     constructor(String)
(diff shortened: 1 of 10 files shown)


dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ProfilingConstants.kt
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.profiling
+
+/**
+ * Contract key used to communicate the profiler running state via the features context.
+ * Written by the profiling feature and read by RUM and debug widget consumers.
+ */
+const val PROFILER_IS_RUNNING: String = "profiler_is_running"
```
0xnm: Maybe we can create `object FeatureContextKeys` in `com.datadog.android.internal`? I think it is not the only key which we would like to share.

0xnm (dismissed):
We may want to move the file one level up, from `com.datadog.android.internal.profiling` to just `com.datadog.android.internal`, so that `FeatureContextKeys` class can be re-used by the multiple features, but not a blocker.

0xnm (approved):

PR #3349, opened 2026-04-08
RUM-15390: Fix memory leak in app launch

aleksandr-gringauz:
### What does this PR do?

Fixing the memory leak.

**How it happens:**
1. Suppose we have a situation when we have an interstitial `Activity` that launches another `Activity` in its `onCreate` method.
2. [RumWindowCallbacksRegistry.addListener](https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterImpl.kt#L39) is called for the interstitial `Activity`.
3. The interstitial `Activity` calls `finish` in its `onCreate`, thus `removeListener` for it is never called [here](https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterImpl.kt#L35).
4. The interstitial `Activity` is leaked [here](https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/utils/window/RumWindowCallbacksRegistry.kt#L66) (`Activity` implements `Window.Callback`).

**Changes in the PR**

1. The fix is done by changing the API of [subscribeToFirstFrameDrawn](https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporter.kt#L20) method. Now it returns an instance of `Handle` that makes it possible to unsubscribe the `Activity` in `onActivityDestroyed`.
2. Also added [LeakCanary](https://square.github.io/leakcanary/) to the integration tests about app launch with interstitial Activities. They failed without my fix exactly with this memory leak.
3. Also added LeakCanary to the sample app. This is just so that it can catch some other leaks in the future during local SDK development. 

The number of lines in the diff is large, but mostly because of the tests.

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
@@ -196,6 +196,9 @@ internal class RumAppStartupDetectorImpl(
     override fun destroy() {
         pendingScenario = null
         application.unregisterActivityLifecycleCallbacks(this)
+
+        firstFrameHandles.forEach { (_, handle) -> handle.unsubscribe() }
+        firstFrameHandles.clear()
     }
 
     companion object {
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
@@ -24,7 +24,8 @@ internal class RumAppStartupDetectorImpl(
     private val appStartupTimeProvider: () -> Time,
     private val timeProvider: () -> Time,
     private val listener: RumAppStartupDetector.Listener,
-    private val appStartupActivityPredicate: AppStartupActivityPredicate
+    private val appStartupActivityPredicate: AppStartupActivityPredicate,
+    private val rumFirstDrawTimeReporter: RumFirstDrawTimeReporter
 ) : RumAppStartupDetector, Application.ActivityLifecycleCallbacks {
 
     private var numberOfActivities: Int = 0
@@ -34,6 +35,7 @@ internal class RumAppStartupDetectorImpl(
 
     @Suppress("UnsafeThirdPartyFunctionCall") // map is initialized empty
     private val trackedActivities = Collections.newSetFromMap(WeakHashMap<Activity, Boolean>())
+    private val firstFrameHandles = WeakHashMap<Activity, RumFirstDrawTimeReporter.Handle>()
 
     init {
         application.registerActivityLifecycleCallbacks(this)
@@ -55,6 +57,8 @@ internal class RumAppStartupDetectorImpl(
         numberOfActivities--
         trackedActivities.remove(activity)
 
+        firstFrameHandles.remove(activity)?.unsubscribe()
+
         if (numberOfActivities == 0) {
             isChangingConfigurations = activity.isChangingConfigurations
         }
@@ -75,6 +79,7 @@ internal class RumAppStartupDetectorImpl(
     override fun onActivityStopped(activity: Activity) {
     }
 
+    @Suppress("LongMethod")
     private fun onBeforeActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
         numberOfActivities++
         val now = timeProvider()
@@ -132,7 +137,15 @@ internal class RumAppStartupDetectorImpl(
             }
 
             pendingScenario = scenario
+
             listener.onAppStartupDetected(scenario)
+
+            subscribeToFirstFrameDrawn(
+                scenario = scenario,
+                activity = activity,
+                wasForwarded = false
+            )
+
             isFirstActivityForProcess = false
         }
 
@@ -142,14 +155,42 @@ internal class RumAppStartupDetectorImpl(
         if (currentPendingScenario != null && shouldTrackStartup &&
             currentPendingScenario.activity.get() !== activity
         ) {
-            listener.onNextActivityCreated(currentPendingScenario, activity)
+            subscribeToFirstFrameDrawn(
+                scenario = currentPendingScenario,
+                activity = activity,
+                wasForwarded = true
+            )
         }
     }
 
-    override fun getPendingScenario(): RumStartupScenario? = pendingScenario
+    private fun subscribeToFirstFrameDrawn(
+        scenario: RumStartupScenario,
+        activity: Activity,
+        wasForwarded: Boolean
+    ) {
+        val callback = object : RumFirstDrawTimeReporter.Callback {
+            override fun onFirstFrameDrawn(timestampNs: Long) {
+                firstFrameHandles.remove(activity)
 
-    override fun clearPendingScenario() {
-        pendingScenario = null
+                // Another activity may have already reported TTID
+                if (pendingScenario !== scenario) return
+
+                val durationNs = timestampNs - scenario.initialTime.nanoTime
+
+                listener.onTTIDComputed(
+                    scenario = scenario,
+                    durationNs = durationNs,
+                    wasForwarded = wasForwarded
+                )
+
+                pendingScenario = null
+            }
+        }
+
+        firstFrameHandles[activity] = rumFirstDrawTimeReporter.subscribeToFirstFrameDrawn(
+            activity = activity,
+            callback = callback
+        )
     }
 
     override fun destroy() {
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
@@ -1,0 +1,144 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.startup
+
+import android.app.Activity
+import android.os.Handler
+import android.os.Message
+import android.view.View
+import android.view.ViewTreeObserver
+import com.datadog.android.api.InternalLogger
+import com.datadog.android.rum.internal.utils.window.RumWindowCallbackListener
+import com.datadog.android.rum.internal.utils.window.RumWindowCallbacksRegistry
+
+internal class RumFirstDrawTimeReporterHandleImpl(
+    private val callback: RumFirstDrawTimeReporter.Callback,
+    private val activity: Activity,
+    private val internalLogger: InternalLogger,
+    private val timeProviderNs: () -> Long,
+    private val windowCallbacksRegistry: RumWindowCallbacksRegistry,
+    private val handler: Handler
+) : RumFirstDrawTimeReporter.Handle, RumWindowCallbackListener {
+
+    private var isCancelled: Boolean = false
+
+    init {
+        val window = activity.window
+        val decorView = window.peekDecorView()
+
+        if (decorView == null) {
+            windowCallbacksRegistry.addListener(activity, this)
+        } else {
+            onDecorViewReady()
+        }
+    }
+
+    override fun unsubscribe() {
+        if (isCancelled) {
+            return
+        }
+        isCancelled = true
+
+        windowCallbacksRegistry.removeListener(activity, this)
+    }
+
+    override fun onContentChanged() {
+        windowCallbacksRegistry.removeListener(activity, this)
+        onDecorViewReady()
+    }
+
+    private fun reportFirstFrame() {
+        if (isCancelled) {
+            return
+        }
+
+        val nowNs = timeProviderNs()
+        callback.onFirstFrameDrawn(nowNs)
+    }
+
+    private fun onDecorViewReady() {
+        if (isCancelled) {
+            return
+        }
+
+        val window = activity.window
+        val decorView = window.decorView
+
+        if (decorView.isAttachedToWindow) {
+            registerOnDrawListener(decorView)
+        } else {
+            val attachListener = object : View.OnAttachStateChangeListener {
+                override fun onViewAttachedToWindow(v: View) {
+                    registerOnDrawListener(decorView)
+                    decorView.removeOnAttachStateChangeListener(this)
+                }
+
+                override fun onViewDetachedFromWindow(v: View) {
+                }
+            }
+            decorView.addOnAttachStateChangeListener(attachListener)
+        }
+    }
+
+    private fun registerOnDrawListener(decorView: View) {
+        if (isCancelled) {
+            return
+        }
+
+        val listener = object : ViewTreeObserver.OnDrawListener {
+            private var invoked = false
+
+            override fun onDraw() {
+                if (invoked) {
+                    return
+                }
+                invoked = true
+                onFirstDraw()
+
+                handler.post {
+                    if (decorView.viewTreeObserver.isAlive) {
+                        try {
+                            decorView.viewTreeObserver.removeOnDrawListener(this)
+                        } catch (e: IllegalStateException) {
+                            internalLogger.log(
+                                InternalLogger.Level.WARN,
+                                InternalLogger.Target.TELEMETRY,
+                                { "RumTTIDReporterImpl unable to remove onDrawListener from viewTreeObserver" },
+                                e
+                            )
+                        }
+                    }
+                }
+            }
+        }
+
+        if (decorView.viewTreeObserver.isAlive) {
+            try {
+                decorView.viewTreeObserver.addOnDrawListener(listener)
+            } catch (e: IllegalStateException) {
+                internalLogger.log(
+                    InternalLogger.Level.WARN,
+                    InternalLogger.Target.TELEMETRY,
+                    { "RumFirstDrawTimeReporterImpl unable to add onDrawListener onto viewTreeObserver" },
+                    e
+                )
+            }
+        }
+    }
+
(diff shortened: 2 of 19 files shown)


aleksandr-gringauz:
@codex review

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/rum/startup/AppStartupActivityPredicateTest.kt
```diff
@@ -109,6 +110,8 @@ internal class AppStartupActivityPredicateTest :
             )
             true
         }.doWait(timeoutMs = FINAL_WAIT_MS)
+
+        LeakAssertions.assertNoLeaks()
```
aleksandr-gringauz, line 114: I do it individually in each test, not in one place, because:

1. If we do it [here](https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/rules/MockServerActivityTestRule.kt#L84), some strange memory leak is [detected](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40test.name%3AverifyTTIDAutoForwardedThroughChainOfNonDrawingActivities%20%40test.status%3Afail&agg_m=count&agg_m_source=base&agg_t=count&currentTab=overview&eventStack=AwAAAZ1xgNp4tpyO9AAAABhBWjF4Z0c3WEFBQzk0MS10RFdKRTRkVnkAAAAkZjE5ZDcxODMtM2VmZi00OWRlLThjN2MtYWVmYzg5NTMzNzNkAALA6A&fromUser=false&index=citest&refresh_mode=sliding&testId=AwAAAZ1xgNp4tpyO9AAAABhBWjF4Z0c3WEFBQzk0MS10RFdKRTRkVnkAAAAkZjE5ZDcxODMtM2VmZi00OWRlLThjN2MtYWVmYzg5NTMzNzNkAALA6A&trace=AwAAAZ1xgNp4tpyO9AAAABhBWjF4Z0c3WEFBQzk0MS10RFdKRTRkVnkAAAAkZjE5ZDcxODMtM2VmZi00OWRlLThjN2MtYWVmYzg5NTMzNzNkAALA6A&start=1775162172182&end=1775766972182&paused=false). It is 100% not a problem in our SDK, but rather in our tests. I haven't managed to find a solution to this.
2. We can't use the [test rule](https://square.github.io/leakcanary/ui-tests/#test-rule), because it checks the leaks after `Datadog.stop` is called [here](https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/rules/MockServerActivityTestRule.kt#L92) which clears all features including `RumFeature` that itself clears [this](https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt#L360). It allows interstitial `Activity` to be garbage collected and the leak isn't detected.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
```diff
@@ -0,0 +1,144 @@
...
+        onDecorViewReady()
+    }
+
+    private fun reportFirstFrame() {
+        if (isCancelled) {
+            return
+        }
+
+        val nowNs = timeProviderNs()
+        callback.onFirstFrameDrawn(nowNs)
```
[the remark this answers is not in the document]
aleksandr-gringauz: This is a good comment. However this change is on purpose. In the original RFC I wanted to do exactly this (read [here](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5439422959/RFC+-+App+Launch+Observability+in+Android#%5BRecommended-for-API-%3E%3D-21%5D-OnDraw-%2B-handler.sendMessageAtFrontOfQueue)).

But at some point (I don't remember when) apparently I made a mistake and instead of using the time of when the `Runnable` is executed, I used the time of when `onDraw` was called.

sample/kotlin/build.gradle.kts
```diff
@@ -237,6 +237,9 @@ dependencies {
     // Misc
     implementation(libs.timber)
     api("com.facebook.stetho:stetho:1.6.0")
+
+    // Leak detection
+    implementation(libs.leakCanaryAndroid)
```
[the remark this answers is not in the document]
0xnm: That is valid, we don't want to have it in s8s runs

https://github.com/DataDog/dd-sdk-android/blob/e71690e577779ac6542742e6a5209abe0cc04d20/ci/pipelines/default-pipeline.yml#L415

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImplTest.kt
```diff
@@ -247,23 +221,37 @@ class RumFirstDrawTimeReporterTest {
...
-    fun `M call internalLogger W addOnDrawListener { if it throws IllegalStateException }`() {
+    fun `M not add listener to registry W decorView exists`() {
+        // Given
+        whenever(window.peekDecorView()) doReturn decorView
+
+        // When
+        createHandle()
+
+        // Then
+        verify(windowCallbackRegistry, never()).addListener(any(), any())
```
0xnm: nit: wouldn't it be better simply to use `verifyNoInteractions` here?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
```diff
@@ -126,17 +138,20 @@ internal class RumAppStartupDetectorImplTest {
...
+                        nanoTime = currentTime.inWholeNanoseconds,
+                        timestamp = currentTime.inWholeMilliseconds
+                    ),
+                    hasSavedInstanceStateBundle = hasSavedInstanceStateBundle,
+                    activity = activity.wrapWeak(),
+                    appStartActivityOnCreateGapNs = 11.seconds.inWholeNanoseconds
+                )
             )
-        )
+            verify(listener).onTTIDComputed(any(), eq(0.seconds.inWholeNanoseconds), eq(false))
```
0xnm: nit: I believe here it should be the same scenario as above? if yes, instead of `any` we can use `eq(...)` by extracting scenario into a variable and using it in both `listener.verifyScenarioDetected` and `verify` calls. Same we can do in the tests below.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
```diff
@@ -34,6 +35,7 @@ internal class RumAppStartupDetectorImpl(
 
     @Suppress("UnsafeThirdPartyFunctionCall") // map is initialized empty
     private val trackedActivities = Collections.newSetFromMap(WeakHashMap<Activity, Boolean>())
+    private val firstFrameHandles = WeakHashMap<Activity, RumFirstDrawTimeReporter.Handle>()
```
0xnm, line 38: just curious: Won't it be enough just to keep `identityHashCode` as a key? Don't we risk to have a minor leak here by keeping the key longer than needed (it will be GC-ed eventually since it is a `WeakHashMap`)?

But I guess `onActivityDestroyed` is called all the time during the normal app lifecycle.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
```diff
@@ -0,0 +1,144 @@
...
+        val window = activity.window
+        val decorView = window.decorView
+
+        if (decorView.isAttachedToWindow) {
+            registerOnDrawListener(decorView)
+        } else {
+            val attachListener = object : View.OnAttachStateChangeListener {
+                override fun onViewAttachedToWindow(v: View) {
+                    registerOnDrawListener(decorView)
+                    decorView.removeOnAttachStateChangeListener(this)
```
hamorillo: ❓ I think it can be possible for the attachListener to not be removed if `unsubscribe` is called before the view is attached the the window. This was already in the previous code but do you think we should remove it during the `unsubscribe`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
hamorillo, line 198: ❓ Should we clear `firstFrameHandles` here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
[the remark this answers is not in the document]
hamorillo: 🤔 Just to understand, why don't we want to use the time when onDraw is called?

hamorillo (commented):
Left some minor comments, LGTM overall 🚀

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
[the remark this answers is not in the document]
aleksandr-gringauz: It is explained in the [link](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5439422959/RFC+-+App+Launch+Observability+in+Android#%5BRecommended-for-API-%3E%3D-21%5D-OnDraw-%2B-handler.sendMessageAtFrontOfQueue) to the RFC that I mentioned above. In short - it is more precise.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: My current understanding is that `identityHashCode` should not be used as a key in hash maps.

`identityHashCode` returns an `int`. So it can represent only 2^32 distinct values. This means that for 100 000 items the probability of at least one collision is approximately 68% (see [this](https://en.wikipedia.org/wiki/Birthday_problem) and to calculate [this](https://instacalc.com/28845)). We will not have 100 000 Activities - yes, but some other things (some kind of spans maybe) - why not. So I don't think it is a good practice in general.

Also see [this](https://shipilev.net/jvm/anatomy-quarks/26-identity-hash-code/#:~:text=So%2C%20current%20implementations%20compute%20the%20identity%20hash%20code%20from%20the%20internal%20PRNG%20(%22good%20distribution%22)%2C%20and%20store%20it%20for%20every%20object%20(%22idempotence%22).).

> But I guess onActivityDestroyed is called all the time during the normal app lifecycle.

yes, exactly. Here I use `WeakHashMap` just as a precaution against some kind of mistake. It could be replaced by a normal `Map`.

aleksandr-gringauz: ~~https://pl.kotl.in/Ebe5fCsD4~~ https://pl.kotl.in/mp61XH2LK - sample to reproduce a collision.

0xnm: That is true. My idea is that since it is just used as a key in the map, where lookup is done by first checking `hashCode` for a bucket and then reference comparison, then maybe we can capture some identity property of `Activity` instead and use it.

But as long as it is guaranteed to do a cleanup in `onActivityDestroyed` - it is okay.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
aleksandr-gringauz: Actually a good point.

Fixed, now also unsubscribing from `AttachStateChangeListener` as well as `onDrawListener` in `unsubscribe`.

sample/kotlin/build.gradle.kts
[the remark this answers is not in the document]
aleksandr-gringauz: changed to `debugImplementation`

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImplTest.kt
aleksandr-gringauz: done

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImplTest.kt
aleksandr-gringauz: done

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
aleksandr-gringauz: done

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumAppStartupDetectorImpl.kt
0xnm: Looking at the reproducer: I think it is expected behavior, because `obj` can be GC-ed on each loop end, so system can give the int which was already seen before, but it is not attached to any object (since previous owner was already GC-ed).

aleksandr-gringauz: yes, you are right. This one is better https://pl.kotl.in/mp61XH2LK

aleksandr-gringauz: Also the interesting thing about this https://pl.kotl.in/mp61XH2LK is that every time you run it it finds the collision at exactly the same iteration `105668` - which proves the point [here](https://shipilev.net/jvm/anatomy-quarks/26-identity-hash-code/#:~:text=So%2C%20current%20implementations%20compute%20the%20identity%20hash%20code%20from%20the%20internal%20PRNG%20(%22good%20distribution%22)%2C%20and%20store%20it%20for%20every%20object%20(%22idempotence%22).) that some kind of pseudo-random number  generator is involved in generating the identity hash code.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
```diff
@@ -0,0 +1,160 @@
...
+            Message.obtain(handler, block).apply {
+                isAsynchronous = true
+            }
+        )
+    }
+
+    override fun onViewAttachedToWindow(v: View) {
+        val decorView = getDecorView()
+
+        registerOnDrawListener(activity.window.decorView)
```
hamorillo: nitpick:

```suggestion
        registerOnDrawListener(decorView)
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
```diff
@@ -0,0 +1,160 @@
...
+            decorView.removeOnAttachStateChangeListener(this)
+            removeOnDrawListener(decorView)
+        }
+    }
+
+    override fun onContentChanged() {
+        windowCallbacksRegistry.removeListener(activity, this)
+
+        val decorView = getDecorView()
+        onDecorViewReady(decorView)
```
hamorillo: nitpick:

```suggestion
        onDecorViewReady(getDecorView())
```

hamorillo (dismissed):
🚀 Nice work!

0xnm (dismissed):

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumFirstDrawTimeReporterHandleImpl.kt
aleksandr-gringauz: fixed nitpicks

0xnm (approved):

hamorillo (approved):

PR #3351, opened 2026-04-09
RUM-517: Remove kapt by bumping glide version

kikoveiga:
### What does this PR do?

- Bump glide from `4.11.0` to `5.0.5`, enabling the usage of ksp instead of kapt. The APIs used by `dd-sdk-android-glide` are unchanged in v5, so there is no behavioral impact. Binary compatibility for new glide v5 (minSdk 14, Java 8) is still below the project's minSdk 23 and Java 11.
- Minor fixes.


gradle/libs.versions.toml
```diff
@@ -83,7 +83,7 @@ coil = "1.0.0"
 coilCompose = "2.1.0"
 coil3 = "3.0.4"
 fresco = "2.3.0"
-glide = "4.11.0"
+glide = "5.0.5"
```
0xnm, line 86: is it the first version where KSP support landed? won't it be enough to use 5.0.0?

integrations/dd-sdk-android-glide/build.gradle.kts
```diff
@@ -46,6 +46,7 @@ dependencies {
     implementation(libs.kotlin)
     implementation(libs.okHttp)
     implementation(libs.bundles.glide)
+    compileOnly(project(":features:dd-sdk-android-trace-api"))
```
0xnm: I believe it shouldn't be addressed here, it is `okhttp` module should export `api(project(":features:dd-sdk-android-trace-api"))` I guess, because OkHttp integration references symbols from Trace API in its public API.

0xnm (changes requested):

gradle/libs.versions.toml
kikoveiga: Yes 5.0.0 is enough, but from their [releases page](https://github.com/bumptech/glide/releases) it seems 5.0.0 and 5.0.1 contain a bug, so minimum we can use I guess is 5.0.2, should we use that?

integrations/dd-sdk-android-glide/build.gradle.kts
kikoveiga: Thanks, fixed.

gradle/libs.versions.toml
0xnm: The effective version is controlled by the app anyway (probably it will be 5.0.5, so it will be resolved to this version globally), so we can use 5.0.0 to be safer with transitive dependencies.

kikoveiga: As discussed, 5.0.5 has the same transitive dependencies as 5.0.0, so let's keep it.

integrations/dd-sdk-android-okhttp/build.gradle.kts
```diff
@@ -51,6 +51,7 @@ dependencies {
     implementation(project(":dd-sdk-android-internal"))
     implementation(project(":features:dd-sdk-android-rum"))
     implementation(project(":features:dd-sdk-android-trace"))
+    api(project(":features:dd-sdk-android-trace-api"))
```
0xnm: Let's do this in a separate PR, so that we have a proper description of the change attached.

kikoveiga: Ok, removed.

0xnm (dismissed):

0xnm (approved):

PR #3366, opened 2026-04-10
Expose trace as api dependency in okhttp module

kikoveiga:
### What does this PR do?

  Exposes `dd-sdk-android-trace` as an `api` dependency of `dd-sdk-android-okhttp`, fixing the following Kotlin compiler warning in consumer modules that don't independently depend on `dd-sdk-android-trace` (namely `dd-sdk android-glide` and `sample/tv`):                                                                                                                                                                                                                                         
  > Cannot access class 'DatadogSpan' in the expression type. This may be forbidden soon.                                                                                                                                                
  > Check the module classpath for missing or conflicting dependencies.

The warning is triggered when calling `TracingInterceptor.BaseBuilder#setTraceSampler(Sampler<DatadogSpan>)`, since `DatadogSpan` (from `trace-api`) was only reachable transitively through `implementation(dd-sdk-android-trace)`, which Gradle hides from consumers' compile classpath.                                                                           

### DIFF
--- a/integrations/dd-sdk-android-okhttp/build.gradle.kts
+++ b/integrations/dd-sdk-android-okhttp/build.gradle.kts
@@ -51,6 +51,7 @@ dependencies {
     implementation(project(":dd-sdk-android-internal"))
     implementation(project(":features:dd-sdk-android-rum"))
     implementation(project(":features:dd-sdk-android-trace"))
+    api(project(":features:dd-sdk-android-trace-api"))
     // Generate NoOp implementations
     ksp(project(":tools:noopfactory"))
 
--- a/integrations/dd-sdk-android-okhttp/build.gradle.kts
+++ b/integrations/dd-sdk-android-okhttp/build.gradle.kts
@@ -50,8 +50,9 @@ dependencies {
     implementation(libs.androidXAnnotation)
     implementation(project(":dd-sdk-android-internal"))
     implementation(project(":features:dd-sdk-android-rum"))
-    implementation(project(":features:dd-sdk-android-trace"))
-    api(project(":features:dd-sdk-android-trace-api"))
+    // TODO RUM-15626: Move TraceContextInjection to dd-sdk-android-trace-api module,
+    //  then replace api(dd-sdk-android-trace) with api(dd-sdk-android-trace-api).
+    api(project(":features:dd-sdk-android-trace"))
     // Generate NoOp implementations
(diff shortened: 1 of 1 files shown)


0xnm:
We have one more problem of that kind: `setTraceContextInjection` is using the symbol which is declared in `dd-sdk-android-trace`.

https://github.com/DataDog/dd-sdk-android/blob/d19a14a0bb1f24a0f5a6f7827952d176a50e602c/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt#L836

it is declared here

https://github.com/DataDog/dd-sdk-android/blob/d19a14a0bb1f24a0f5a6f7827952d176a50e602c/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/TraceContextInjection.kt

I'm wondering if we can move it to the `-api` module, but then it will be a breaking API change 🤔 

@satween wdyt?

satween:
> We have one more problem of that kind: `setTraceContextInjection` is using the symbol which is declared in `dd-sdk-android-trace`.
> 
> https://github.com/DataDog/dd-sdk-android/blob/d19a14a0bb1f24a0f5a6f7827952d176a50e602c/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt#L836
> 
> it is declared here
> 
> https://github.com/DataDog/dd-sdk-android/blob/d19a14a0bb1f24a0f5a6f7827952d176a50e602c/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/TraceContextInjection.kt
> 
> I'm wondering if we can move it to the `-api` module, but then it will be a breaking API change 🤔
> 
> @satween wdyt?

We discussed with @0xnm in person, and it's seems like the best solution for now is the following:

- replace `api("-trace-api")` with `api("-trace")` till the next major release. 
- add a JIRA ticket linked to v4 scope, that states that  `TraceContextInjection.kt` has to be moved into `-trace-api` module and `api("-trace")` should be replaced with `api("-trace-api")`. 
- Add `TODO` comment to that ticket

kikoveiga:
Thank you @satween and @0xnm, created the ticket and added as TODO note in the code.

satween (approved):

0xnm (approved):

PR #141, opened 2020-02-19
RUMM-233 Inject a random TraceId on top level spans

xgouchet:
### What does this PR do?

Inject a random TraceId on top level spans

### Motivation

Android uses a zygote process, meaning that new instance of the same app will always have the same trace id, unless the device gets rebooted.

### Additional Notes

Anything else we should know when reviewing?

### Review checklist (to be filled by reviewers)

- [X] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [X] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [X] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](CONTRIBUTING.md) doc)

PR #3379, opened 2026-04-17
Use `SecureRandom` for Trace/Span ID generation, remove usage of `SecureRandom.getStrongInstance`

0xnm:
### What does this PR do?

This PR does the two things:

#### Use `SecureRandom` for `DatadogTracing`

It fixes the regression when we pulled `dd-trace-java`. In RUMM-233 + https://github.com/DataDog/dd-sdk-android/pull/141 we switched to `SecureRandom`, but during the latest pull of `dd-trace-java` we enabled its usage only for Otel, see

https://github.com/DataDog/dd-sdk-android/blob/e293f09f1c310307626e9fcb1fab3168d4699756/features/dd-sdk-android-trace-otel/src/main/kotlin/com/datadog/android/trace/opentelemetry/OtelTracerProvider.kt#L103

and later down the call chain

https://github.com/DataDog/dd-sdk-android/blob/e293f09f1c310307626e9fcb1fab3168d4699756/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/DatadogTracerBuilderAdapter.kt#L87

Small change in `Config.java` activates `SecureRandom` by default.

#### Remove `SecureRandom#getInstanceStrong` usage

[SecureRandom#getInstanceStrong()](https://developer.android.com/reference/java/security/SecureRandom#getInstanceStrong()) returns the generator which calls `/dev/random` and if there is not enough entropy any call to generate a random number will block until there is enough entropy.

Default constructor though is using `/dev/urandom` which doesn't block if there is not enough entropy. This is okay for our needs to generate ID. We are not using plain `Random` because it is using a timestamp as a seed, so span generated on two different devices at the same instant may have same ID.

The change is aligned with [iOS generator](https://github.com/DataDog/dd-sdk-ios/blob/feedabd640984c7f839ca0e191dc79ce05861a82/DatadogInternal/Sources/Models/Trace/TraceID.swift#L224), which is using [SystemRandomNumberGenerator
](https://developer.apple.com/documentation/swift/systemrandomnumbergenerator)

At least on Linux systems it also calls `/dev/urandom` and not `/dev/random`.

### DIFF
--- a/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/ConfigDefaults.java
+++ b/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/ConfigDefaults.java
@@ -172,7 +172,7 @@ public final class ConfigDefaults {
 
   static final boolean DEFAULT_TRACE_128_BIT_TRACEID_GENERATION_ENABLED = true;
   static final boolean DEFAULT_TRACE_128_BIT_TRACEID_LOGGING_ENABLED = false;
-  static final boolean DEFAULT_SECURE_RANDOM = false;
+  static final boolean DEFAULT_SECURE_RANDOM = true;
 
   public static final int DEFAULT_TRACE_X_DATADOG_TAGS_MAX_LENGTH = 512;
 
--- a/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/IdGenerationStrategy.java
+++ b/features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/IdGenerationStrategy.java
@@ -2,8 +2,6 @@ package com.datadog.trace.api;
 
 import static java.lang.Long.MAX_VALUE;
 
-import android.os.Build;
-
 import androidx.annotation.VisibleForTesting;
 
 import java.security.SecureRandom;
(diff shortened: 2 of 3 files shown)


features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/ConfigDefaults.java
```diff
@@ -172,7 +172,7 @@ public final class ConfigDefaults {
 
   static final boolean DEFAULT_TRACE_128_BIT_TRACEID_GENERATION_ENABLED = true;
   static final boolean DEFAULT_TRACE_128_BIT_TRACEID_LOGGING_ENABLED = false;
-  static final boolean DEFAULT_SECURE_RANDOM = false;
+  static final boolean DEFAULT_SECURE_RANDOM = true;
```
kikoveiga, line 175: If this is pulled code, how can we guarantee that next time we won't let this become `false` again?

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/IdGenerationStrategy.java
```diff
@@ -99,11 +99,7 @@ static final class SRandom extends IdGenerationStrategy {
     private final SecureRandom secureRandom;
 
     private static ThrowingSupplier<SecureRandom> getSecureRandomSupplier() {
-      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
kikoveiga, line 102: Please remove the now unused `Build` import.

features/dd-sdk-android-trace-internal/src/main/java/com/datadog/trace/api/ConfigDefaults.java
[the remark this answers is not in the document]
0xnm: this is what we want

0xnm: We cannot. Code has many other changes to adapt it to Android (and many code removed), so the next time such sync happens one needs to carefully analyze the conflicts.

kikoveiga (approved):

mariusc83 (approved):

Issue #3259, opened 2026-03-13
dd-sdk-android-compose incompatible with Compose UI 1.8+ (NoSuchMethodError: getLayoutDelegate$ui_release)


xJormungandr:
## Summary

`dd-sdk-android-compose:3.7.1` crashes silently at runtime when the consuming app uses **Compose UI 1.8.0 or newer**, causing Compose gesture/tap tracking to be completely non-functional. The KCP compile-time instrumentation works correctly, but the runtime hit-testing always fails.

## Root Cause

`LayoutNodeUtils.getLayoutNodeBoundsInWindow()` accesses `LayoutNode.layoutDelegate` via Kotlin's `@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")` mechanism:

```kotlin
// dd-sdk-android-compose/internal/utils/LayoutNodeUtils.kt
fun getLayoutNodeBoundsInWindow(node: LayoutNode): Rect? {
    return runSafe("getLayoutNodeBoundsInWindow") {
        node.layoutDelegate.outerCoordinator.coordinates.boundsInWindow()
    }
}
```

The JVM getter for this `@InternalComposeApi` property is `getLayoutDelegate$ui_release()`. This method was **removed in Compose UI 1.8.0** as part of an internal layout system refactoring.

At runtime, every call throws:

```
java.lang.NoSuchMethodError: No virtual method getLayoutDelegate$ui_release()
  Landroidx/compose/ui/node/LayoutNodeLayoutDelegate;
  in class Landroidx/compose/ui/node/LayoutNode;
  or its super classes
```

This exception is caught by `runSafe`, which returns `null`. The hit-testing in `ComposeActionTrackingStrategy` then returns `false` for every node, and no Compose taps are ever attributed — producing the log message: _"We could not find a valid target for the gesture event. Compose actions tracking not enabled, or the compose view is not tagged."_

## Affected Versions

| Component | Version |
|---|---|
| `dd-sdk-android-compose` | `3.7.1` |
| `dd-sdk-android-gradle-plugin` (KCP) | `1.23.0` |
| `androidx.compose.ui:ui` (consuming app) | `1.10.0` (BOM `2025.12.01`) |
| `androidx.compose.ui:ui` (SDK compiled against) | `1.5.4` (BOM `2023.10.01`) |

[...]

PR #3381, opened 2026-04-20
RUM-15813: Cache reflection lookups in LayoutNodeUtils

satween:
### What does this PR do?

Caches reflection lookups in the Compose `LayoutNodeUtils` fallback path. Method handles for `getLayoutDelegate`/`getOuterCoordinator`/`getCoordinates` are memoized per concrete `LayoutNode` subclass via a `HashMap`, and the `LayoutCoordinatesKt.boundsInWindow` reflection is resolved once through a lazy `Method` reference.

### Motivation

Before this change, the reflection fallback (used until the Compose BOM bump in RUM-13454) ran a full `javaClass.methods` linear scan plus multiple `Class.forName` calls on every ACTION_MOVE hit-test, which is a hot path during touch tracking.

<table>
  <tr>
  <th>Before fix</th>
  <th>After fix</th>
  </tr>
  <tr>
  <td>

  [Record of the issue](https://github.com/user-attachments/assets/69736b09-bd9a-4272-8d9e-997602318bff)

  </td>
  <td>

  [Record after the fix](https://github.com/user-attachments/assets/8afd7693-1e59-43f0-97d6-5a532132fc34)

  </td>
  </tr>
  </table>


satween:
@codex review

integrations/dd-sdk-android-compose/src/test/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeGetMethodTest.kt
```diff
@@ -1,133 +0,0 @@
...
-import java.lang.reflect.Method
-
-/**
- * Tests for the [LayoutNodeUtils.getMethod] reflection helper that resolves
- * Kotlin-internal method names regardless of module-suffix mangling.
- *
- * Compose internal properties are compiled to JVM methods whose names are mangled
- * with a module identifier suffix, e.g. {@code getLayoutDelegate$ui} (debug) or
- * {@code getLayoutDelegate$ui_release} (release).  [FakeLayoutNodeUi] and
- * [FakeLayoutNodeUiRelease] simulate these two naming conventions.
```
kikoveiga, line 20: I see these 2 files are now unused, dead code. Should we still use them to test the suffixes (e.g., so we test caching and don't accidentally drop any entry from `KNOWN_INTERNAL_SUFFIXES` list, or just delete them?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
```diff
@@ -153,6 +162,36 @@ internal class LayoutNodeUtils {
...
+        // State machine:
+        //   (activated=false, gaveUp=false) — initial; try internal then reflection.
+        //   (activated=true,  gaveUp=false) — reflection works on this build; skip internal.
+        //   (activated=true,  gaveUp=true)  — reflection is broken (some class in the chain
+        //                                     has no matching suffix, or a cached suffix
+        //                                     failed for a new prefix); fall back to internal
+        //                                     only, avoiding repeated NoSuchMethodException.
+        var activated: Boolean = false
+        var gaveUp: Boolean = false
+        val suffixByClassCache: MutableMap<Class<*>, String?> = mutableMapOf()
```
kikoveiga: Can we cache by the resolved `Method` directly instead of the suffix string? (actually as mentioned in the current PR description -> "Method handles (...) are memoized"):


```suggestion
val methodCache: MutableMap<Pair<Class<*>, String>, Method> = mutableMapOf()

fun findMethod(key: Class<*>, prefix: String): Method? {                                                                                                                                                                           
          methodCache[key to prefix]?.let { return it }                                                                                                                                                                                  
          val resolved = KNOWN_INTERNAL_SUFFIXES.firstNotNullOfOrNull { suffix ->                                                                                                                                                        
              key.tryGetNoArgMethod("$prefix$suffix")                                                                                                                                                                                    
          }                                               
          if (resolved == null) {                                                                                                                                                                                                        
              gaveUp = true                                            
          } else {                                                                                                                                                                                                                       
              methodCache[key to prefix] = resolved
          }                                                                                                                                                                                                                              
          return resolved                                              
      }

```

satween: While I found method caching pretty useful here, `Pair<Class<*>, String>` instantiating for every UI frame might increase GC pressure. Replaced with nested map

integrations/dd-sdk-android-compose/src/test/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeGetMethodTest.kt
satween: I've added more tests into `LayoutNodeUtilsTest` and they covers those scenarios

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
```diff
@@ -99,33 +101,40 @@ internal class LayoutNodeUtils {
         }
     }
 
-    fun getLayoutNodeBoundsInWindow(node: LayoutNode): Rect? = if (reflectionFallbackModeActivated) {
-        getLayoutNodeBoundsInWindowReflection(node)
-    } else {
-        getLayoutNodeBoundsInWindowInternal(node) ?: getLayoutNodeBoundsInWindowReflection(node)
+    fun getLayoutNodeBoundsInWindow(node: LayoutNode): Rect? = when {
+        !reflectionFallbackState.activated -> {
```
0xnm: q: should we change branches a bit for readability?

* `reflectionFallbackState.activated`
* `reflectionFallbackState.gaveUp`
* `else`

or it doesn't really work well?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
```diff
@@ -153,6 +162,36 @@ internal class LayoutNodeUtils {
...
+        // a module suffix (plain / "$ui" / "$ui_release"). The mangling is stable per class,
+        // but NOT necessarily the same across classes in the reflection chain: e.g. on Compose
+        // UI 1.10 LayoutNode.getLayoutDelegate$ui is mangled while LayoutNodeLayoutDelegate
+        // exposes getOuterCoordinator without a suffix. We therefore cache the resolved suffix
+        // per owner-class rather than once globally.
+        //
+        // State machine:
+        //   (activated=false, gaveUp=false) — initial; try internal then reflection.
+        //   (activated=true,  gaveUp=false) — reflection works on this build; skip internal.
+        //   (activated=true,  gaveUp=true)  — reflection is broken (some class in the chain
```
0xnm: so in this case we come back to the https://github.com/DataDog/dd-sdk-android/issues/3259?

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
0xnm: how many elements do we have in the cache after reproducer app run?

0xnm (commented):
lgtm, I've added some questions. Also to double check maybe it is worth to get Frame/CPU stats as well for cases when there is no instrumentation enabled and also when instrumentation enabled, but with SDK state before https://github.com/DataDog/dd-sdk-android/issues/3259 fix

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
satween: 3

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
satween: yes, if reflection is not working we going back to internal branch to not trigger reflection too much (especially if there is no sence)

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
```diff
@@ -168,5 +202,33 @@ internal class LayoutNodeUtils {
...
+            val klassCache: MutableMap<String, Method?> = getOrPut(klass) { mutableMapOf() }
+
+            if (!klassCache.containsKey(methodPrefix)) {
+                klassCache[methodPrefix] = searchMaglings(klass, methodPrefix)
+            }
+
+            return klassCache[methodPrefix]
+        }
+
+        private fun searchMaglings(klass: Class<*>, prefix: String): Method? = SUPPORTED_MANGLING_SUFFIXES
```
kikoveiga:
```suggestion
        private fun searchManglings(klass: Class<*>, prefix: String): Method? = SUPPORTED_MANGLING_SUFFIXES
```

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
```diff
@@ -153,6 +162,31 @@ internal class LayoutNodeUtils {
...
+        // exposes getOuterCoordinator without a suffix. We therefore cache the resolved suffix
+        // per owner-class rather than once globally.
+        //
+        // State machine:
+        //   (activated=false, gaveUp=false) — initial; try internal then reflection.
+        //   (activated=true,  gaveUp=false) — reflection works on this build; skip internal.
+        //   (activated=true,  gaveUp=true)  — reflection is broken (some class in the chain
+        //                                     has no matching suffix, or a cached suffix
+        //                                     failed for a new prefix); fall back to internal
+        //                                     only, avoiding repeated NoSuchMethodException.
```
kikoveiga: Please confirm this description is up-to-date with the implementation:
- We now cache the resolved Method per (owner-class, prefix) pair
- "cache suffix failed for a new prefix" no longer happens

kikoveiga (dismissed):
LGTM, just left some minor suggestions and the stale files `FakeLayoutNodeUi.java` & `FakeLayoutNodeUiRelease.java` are still present.

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
satween: It's not gonna work then, but I got the concern and changed to enum based approach

integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/utils/LayoutNodeUtils.kt
satween: this is changed

kikoveiga (approved):

0xnm (approved):

PR #3390, opened 2026-04-22
RUM-15914: Drop word feature from operations APIs

0xnm:
### What does this PR do?

Change similar to https://github.com/DataDog/dd-sdk-ios/pull/2802: we are changing the wording `feature operations` -> `operations`.

This PR deprecates feature-related symbol names, they should be removed in v4.

Also adding `OperationOptions` parameter, which will be used later for profiling.

cc @simaoseica-dd @Valpertui 

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

PR #3393, opened 2026-04-23
Enforce schema character set on vital.name in Operation APIs

Valpertui:
### What does this PR do?

Validates `vital.name` client-side in the Operation APIs (`startOperation` / `succeedOperation` / `failOperation`) against the backend's character-set regex `[\w.@$-]*`.

- Blank / whitespace-only name → dropped + `InternalLogger.Level.WARN`
- Name with characters outside `[\w.@$-]*` → `InternalLogger.Level.WARN`, event still emitted
- Valid name → silent emit
- `operationKey` has no schema character-set rule and is left untouched

### Motivation

The authoritative `_vital-common-schema.json` documents `vital.name` as restricted to letters, digits, and `- _ . @ $` — it is used as a facet path by the backend. Until now the Android SDK only checked for blank names; non-conforming values (e.g. `"user login"`) were silently shipped to intake, where they are rejected. Client-side validation gives developers immediate feedback at dev time.

Names that fail the character-set regex still emit because the backend is the source of truth on the policy — a client-side drop would force customers to bump the SDK if the server rule is ever relaxed.

### Additional Notes

- New symbols are `internal`: `VALID_OPERATION_NAME_REGEX` and `OPERATION_ERROR_INVALID_NAME_CHARACTERS`. No public API surface changes; `api/apiSurface` and `api/dd-sdk-android-rum.api` do not need to be regenerated.
- `java.util.regex.Pattern`'s default `\w` is ASCII-only (no `UNICODE_CHARACTER_CLASS` flag) — the non-ASCII test case `"ログイン"` pins this behavior against future drift.
- Part of a cross-SDK rollout (iOS, Browser, Roku, C++, Electron). See the cross-SDK RUM Operations reference spec v1.6 for the shared contract.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
@@ -754,22 +754,38 @@ internal class DatadogRumMonitor(
         sdkCore.internalLogger.reportFeatureOperationApiUsage(ActionType.FAIL)
     }
 
-    private fun featureOperationArgumentsValid(name: String, operationKey: String?) = when {
-        name.isBlank() -> {
+    private fun featureOperationArgumentsValid(name: String, operationKey: String?): Boolean {
+        // Blank / empty names are rejected: the backend rejects them with
+        // its own non-empty precondition before evaluating the character-set
+        // regex, so drop client-side to match.
+        if (name.isBlank()) {
             sdkCore.internalLogger.logToUser(InternalLogger.Level.WARN) {
                 FO_ERROR_INVALID_NAME.format(Locale.US, name)
             }
-            false
+            return false
         }
 
-        operationKey?.isBlank() == true -> {
+        // Names that fail the backend's `[\w.@$-]*` character-set regex
+        // trigger a warning but the event is still emitted — the backend is
+        // the source of truth, so client-side drop would force a customer
+        // SDK bump if the rule is ever relaxed.
+        @Suppress(
+            "UnsafeThirdPartyFunctionCall"
+        ) // Regex is a compile-time constant; matches() cannot throw on a non-null input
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
```diff
@@ -754,22 +754,38 @@ internal class DatadogRumMonitor(
...
 
-        operationKey?.isBlank() == true -> {
+        // Names that fail the backend's `[\w.@$-]*` character-set regex
+        // trigger a warning but the event is still emitted — the backend is
+        // the source of truth, so client-side drop would force a customer
+        // SDK bump if the rule is ever relaxed.
+        @Suppress(
+            "UnsafeThirdPartyFunctionCall"
+        ) // Regex is a compile-time constant; matches() cannot throw on a non-null input
+        val nameMatchesBackendPattern = VALID_OPERATION_NAME_REGEX.matches(name)
```
aleksandr-gringauz: You could instead do:

```suggesion
val nameMatchesBackendPattern = name.matches(VALID_OPERATION_NAME_REGEX)
```

This way you don't have to suppress the warning, because `String.matches` is already in `detekt_custom_safe_calls.yml`.

kikoveiga:
Please check #3390 and align with that, including using the new `startOperation` instead and naming functions/variables without the `feature` word. After that, should be good to go!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt
kikoveiga: Fixed, thanks!

kikoveiga:
@codex review

aleksandr-gringauz (dismissed):

aleksandr-gringauz (approved):

kikoveiga (approved):

PR #3402, opened 2026-04-29
RUM-15138: Add cross-product sampling rebasing for RUM-to-APM correlation

hamorillo:
### What does this PR do?

Uses the rebased sample rate (`traceSampleRate * sessionSampleRate / 100`) when RUM-to-APM correlation requires it. A `SessionRebasedSampler` decorator wraps the trace sampler and adjusts the effective rate based on the RUM session sample rate tag on the span.

The rebasing is scoped exclusively to the RUM-to-APM path:
- `DatadogInterceptor` (OkHttp) wraps the sampler in `SessionRebasedSampler` at build time
- `ApmNetworkInstrumentationConfiguration` wraps when `headerPropagationOnly = true`
- `TracingInterceptor` (pure APM) and `headerPropagationOnly = false` use the raw trace rate unchanged
- HBS (head-based sampling) is preserved: parent span decisions propagate as before

### Motivation

When both RUM session sampling and trace sampling are active, the backend needs `_dd.agent_psr` / `_dd.rule_psr` to reflect the actual combined probability for correct statistical reweighting. Without rebasing, these metrics report only the trace rate, leading to incorrect trace counts on the platform.

### Additional Notes

- **API surface changes:** New `@InternalApi` types in `dd-sdk-android-trace`: `SessionRebasedSampler`, `SpanAwareSampler` interface, and `effectiveSampleRate()` extension. New `LogAttributes.RUM_SESSION_SAMPLE_RATE` constant in `dd-sdk-android-core`. All are internal-use types in `*.internal.*` packages.
- **`DeterministicTraceSampler` is NOT modified.** The decorator wraps it externally, avoiding changes to the open public class.
- **Custom samplers** (via `setTraceSampler()`) pass through without rebasing — the decorator only rebases `DeterministicTraceSampler` delegates.
- **Follow-up work (separate PR):** When a parent manual span is dropped by the Trace SDK, `DatadogInterceptor` should ignore it and create an independent trace so the RUM SDK can make its own sampling decision. Related iOS work: [dd-sdk-ios#2726](https://github.com/DataDog/dd-sdk-ios/pull/2726).

### DIFF
--- a/features/dd-sdk-android-trace/api/apiSurface
+++ b/features/dd-sdk-android-trace/api/apiSurface
@@ -90,10 +90,10 @@ data class com.datadog.android.trace.internal.net.RequestTracingState
   fun createRequestInfo(): com.datadog.android.api.instrumentation.network.HttpRequestInfo
 fun RequestTracingState?.toAttributesMap(String, String, String): Map<String, Any?>
 class com.datadog.android.trace.internal.net.SessionRebasedSampler : com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>, SpanAwareSampler
-  constructor(com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>)
+  constructor(com.datadog.android.trace.DeterministicTraceSampler)
   override fun sample(com.datadog.android.trace.api.span.DatadogSpan): Boolean
-  override fun getSampleRate(): Float?
-  override fun getSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float?
+  override fun getSampleRate(): Float
+  override fun getSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float
 interface com.datadog.android.trace.internal.net.SpanAwareSampler
   fun getSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float?
 data class com.datadog.android.trace.internal.net.TraceContext
--- a/features/dd-sdk-android-trace/api/apiSurface
+++ b/features/dd-sdk-android-trace/api/apiSurface
@@ -84,10 +84,18 @@ object com.datadog.android.trace.internal._TraceInternalProxy
   fun activateSpan(com.datadog.android.trace.api.tracer.DatadogTracer, com.datadog.android.trace.api.span.DatadogSpan, Boolean): com.datadog.android.trace.api.scope.DatadogScope?
   fun mergeBaggage(String?, String): String
   fun createApmNetworkInstrumentation(String, com.datadog.android.trace.ApmNetworkInstrumentationConfiguration)
+fun com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>.effectiveSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float?
 data class com.datadog.android.trace.internal.net.RequestTracingState
   constructor(com.datadog.android.api.instrumentation.network.HttpRequestInfoBuilder, Boolean = false, com.datadog.android.trace.api.span.DatadogSpan? = null, Float? = null)
   fun createRequestInfo(): com.datadog.android.api.instrumentation.network.HttpRequestInfo
 fun RequestTracingState?.toAttributesMap(String, String, String): Map<String, Any?>
+class com.datadog.android.trace.internal.net.SessionRebasedSampler : com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>, SpanAwareSampler
+  constructor(com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>)
+  override fun sample(com.datadog.android.trace.api.span.DatadogSpan): Boolean
+  override fun getSampleRate(): Float?
+  override fun getSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float?
+interface com.datadog.android.trace.internal.net.SpanAwareSampler
+  fun getSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float?
 data class com.datadog.android.trace.internal.net.TraceContext
   constructor(String, String, Int)
 fun <T> android.database.sqlite.SQLiteDatabase.transactionTraced(String, Boolean = true, com.datadog.android.trace.api.span.DatadogSpan.(android.database.sqlite.SQLiteDatabase) -> T): T
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
@@ -13,6 +13,7 @@ import com.datadog.android.core.sampling.Sampler
 import com.datadog.android.trace.api.span.DatadogSpan
 import com.datadog.android.trace.api.tracer.DatadogTracer
 import com.datadog.android.trace.internal.ApmNetworkInstrumentation
+import com.datadog.android.trace.internal.net.SessionRebasedSampler
 import com.datadog.android.trace.internal.net.TracerProvider
 
 /**
@@ -246,10 +247,16 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
 
             val tracerProvider = TracerProvider(localTracerFactory, globalTracerProvider)
 
+            val effectiveSampler = if (headerPropagationOnly) {
+                SessionRebasedSampler(traceSampler)
+            } else {
+                traceSampler
+            }
+
             return ApmNetworkInstrumentation(
                 canSendSpan = !headerPropagationOnly,
                 traceOrigin = traceOrigin,
-                traceSampler = traceSampler,
+                traceSampler = effectiveSampler,
                 tracerProvider = tracerProvider,
                 sdkInstanceName = sdkInstanceName,
                 injectionType = traceContextInjection,
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
@@ -119,6 +119,11 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
      * Set the trace sample rate controlling the sampling of APM traces created for
      * auto-instrumented requests. If there is a parent trace attached to the network span created, then its
      * sampling decision will be used instead.
+     *
+     * When `headerPropagationOnly` is enabled, the effective trace sample rate is automatically combined with
+     * the active RUM session sample rate, ensuring the backend receives correct sampling metadata
+     * (`_dd.agent_psr` / `_dd.rule_psr`) for RUM-to-APM correlation.
+     *
      * @param sampleRate the sample rate to use (percentage between 0f and 100f, default is 100f).
      */
     fun setTraceSampleRate(@FloatRange(from = 0.0, to = 100.0) sampleRate: Float) = apply {
@@ -129,6 +134,11 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
      * Set the trace sampler controlling the sampling of APM traces created for
      * auto-instrumented requests. If there is a parent trace attached to the network span created, then its
      * sampling decision will be used instead.
+     *
+     * Note: custom samplers passed here do not participate in cross-product rebasing with the RUM session
+     * sample rate (even when `headerPropagationOnly` is enabled). Use [setTraceSampleRate] if you need
+     * correlated sampling between RUM sessions and APM traces.
+     *
      * @param traceSampler the trace sampler controlling the sampling of APM traces.
      * By default it is a sampler accepting 100% of the traces.
      */
@@ -247,10 +257,11 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
 
             val tracerProvider = TracerProvider(localTracerFactory, globalTracerProvider)
 
-            val effectiveSampler = if (headerPropagationOnly) {
-                SessionRebasedSampler(traceSampler)
+            val currentTraceSampler = traceSampler
+            val effectiveSampler = if (headerPropagationOnly && currentTraceSampler is DeterministicTraceSampler) {
+                SessionRebasedSampler(currentTraceSampler)
             } else {
-                traceSampler
+                currentTraceSampler
             }
 
             return ApmNetworkInstrumentation(
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
@@ -9,6 +9,7 @@ import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.api.instrumentation.network.HttpRequestInfo
 import com.datadog.android.core.sampling.Sampler
+import com.datadog.android.lint.InternalApi
 import com.datadog.android.trace.api.DatadogTracingConstants.PrioritySampling
 import com.datadog.android.trace.api.DatadogTracingConstants.Tags
 import com.datadog.android.trace.api.span.DatadogSpan
@@ -24,6 +25,19 @@ import java.util.Locale
 internal val FeatureSdkCore?.isRumEnabled: Boolean
     get() = this?.getFeature(Feature.RUM_FEATURE_NAME) != null
 
+/**
+ * Returns the effective sample rate for the given [span].
+ * If the sampler implements [SpanAwareSampler], the per-span rate is returned.
+ * Otherwise, the sampler's static rate is returned.
+ */
+@InternalApi
+fun Sampler<DatadogSpan>.effectiveSampleRate(span: DatadogSpan): Float? {
+    return when (this) {
+        is SpanAwareSampler -> getSampleRate(span)
+        else -> getSampleRate()
+    }
+}
+
 internal fun DatadogSpan.applyPriority(isSampled: Boolean, traceSampler: Sampler<DatadogSpan>) {
     val samplingPriority = if (isSampled) {
         PrioritySampling.SAMPLER_KEEP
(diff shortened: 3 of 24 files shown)


hamorillo:
@codex review

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/SessionRebasedSampler.kt
```diff
@@ -0,0 +1,62 @@
...
+) : Sampler<DatadogSpan>, SpanAwareSampler {
+
+    override fun sample(item: DatadogSpan): Boolean {
+        if (delegate !is DeterministicTraceSampler) return delegate.sample(item)
+        val rawRate = delegate.getSampleRate()
+        val effectiveRate = computeEffectiveRate(item, rawRate)
+        return if (effectiveRate == rawRate) {
+            delegate.sample(item)
+        } else {
+            DeterministicTraceSampler(effectiveRate).sample(item)
```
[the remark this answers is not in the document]
hamorillo: I would say this comment is not correct, `SpanSamplingIdProvider::provideId` (used by `DeterministicTraceSampler`) keys on the RUM session ID when present, with the trace ID only as a fallback. So within a single RUM session every span hashes to the same value.

ambushwork (dismissed):

features/dd-sdk-android-trace/api/apiSurface
```diff
@@ -84,10 +84,18 @@ object com.datadog.android.trace.internal._TraceInternalProxy
...
+fun com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>.effectiveSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float?
 data class com.datadog.android.trace.internal.net.RequestTracingState
   constructor(com.datadog.android.api.instrumentation.network.HttpRequestInfoBuilder, Boolean = false, com.datadog.android.trace.api.span.DatadogSpan? = null, Float? = null)
   fun createRequestInfo(): com.datadog.android.api.instrumentation.network.HttpRequestInfo
 fun RequestTracingState?.toAttributesMap(String, String, String): Map<String, Any?>
+class com.datadog.android.trace.internal.net.SessionRebasedSampler : com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>, SpanAwareSampler
+  constructor(com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>)
+  override fun sample(com.datadog.android.trace.api.span.DatadogSpan): Boolean
+  override fun getSampleRate(): Float?
+  override fun getSampleRate(com.datadog.android.trace.api.span.DatadogSpan): Float?
```
0xnm: side thought: we need to refactor this in v4, because otherwise we have a relation to `DatadogSpan` in 2 interfaces: `com.datadog.android.core.sampling.Sampler<com.datadog.android.trace.api.span.DatadogSpan>` and in `SpanAwareSampler` (which just brings utility method `getSampleRate` wrt the span).

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -246,10 +247,16 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
 
             val tracerProvider = TracerProvider(localTracerFactory, globalTracerProvider)
 
+            val effectiveSampler = if (headerPropagationOnly) {
+                SessionRebasedSampler(traceSampler)
```
0xnm: So we are wrapping sampler inside another sampler just because we have `setTraceSampler` API? we need to address it in v4, removing this API, probably.

Also I think it is worth to add in the documentation of different trace sampling rate method (`setTraceSampleRate` / `setTraceSampler`) that there is a dependency on the RUM session sampling where applicable. I'm wondering if iOS SDK updated docs already.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/SessionRebasedSampler.kt
```diff
@@ -0,0 +1,62 @@
...
+    override fun getSampleRate(span: DatadogSpan): Float? {
+        if (delegate !is DeterministicTraceSampler) return delegate.effectiveSampleRate(span)
+        val rawRate = delegate.getSampleRate()
+        return computeEffectiveRate(span, rawRate)
+    }
+
+    private fun computeEffectiveRate(item: DatadogSpan, rawRate: Float): Float {
+        val sessionRate = (item.context().tags[LogAttributes.RUM_SESSION_SAMPLE_RATE] as? Number)?.toFloat()
+        return if (sessionRate != null) {
+            DeterministicSampling.combinedSampleRate(sessionRate, rawRate)
```
0xnm: I think we may have confusion here: `combinedSampleRate` vs `effectiveSampleRate`. What is the difference? It seems that the former is coming from iOS SDK?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/SessionRebasedSampler.kt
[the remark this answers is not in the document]
0xnm: @hamorillo it is something to look at? Probably this statement from @codex is due to the `effectiveRateUsage` which is an outcome of `computeEffectiveRate` <- `DeterministicSampling.combinedSampleRate(sessionRate, rawRate)`?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
```diff
@@ -24,6 +25,19 @@ import java.util.Locale
 internal val FeatureSdkCore?.isRumEnabled: Boolean
     get() = this?.getFeature(Feature.RUM_FEATURE_NAME) != null
 
+/**
+ * Returns the effective sample rate for the given [span].
+ * If the sampler implements [SpanAwareSampler], the per-span rate is returned.
+ * Otherwise, the sampler's static rate is returned.
+ */
+@InternalApi
+fun Sampler<DatadogSpan>.effectiveSampleRate(span: DatadogSpan): Float? {
```
satween, line 34: We have method called `computeEffectiveSampleRate` that computes effective sample rate for the telemetry events reported to the intake.  

This method tries to report real sampling rate for events that could (or could not) being sent depending on multiple condition at  the beginning and the ending of the such event.  One of that kind of event is the "session ended metric". Sample rate of such metric depend not only on the sampling rate for this metric itself, but also on sampling rate of the session (if session is not sampled - then "session end" metric is not sampled as well). 

Could this feature with the span-aware sampling (which converts to session-session-rate-aware sampling) somehow affect those effective rate that is being reported by the telemetry? Could you please check that ?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/SessionRebasedSampler.kt
```diff
@@ -0,0 +1,62 @@
...
+            delegate.sample(item)
+        } else {
+            DeterministicTraceSampler(effectiveRate).sample(item)
+        }
+    }
+
+    override fun getSampleRate(): Float? = delegate.getSampleRate()
+
+    override fun getSampleRate(span: DatadogSpan): Float? {
+        if (delegate !is DeterministicTraceSampler) return delegate.effectiveSampleRate(span)
```
satween: If this class works basically with only one type of Samplers - maybe it's worth to make it allow to wrap only `DeterministicTraceSampler` in the constructor - so no such checks gonna be required. WDYT?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
satween: Maybe it's better to just have independent `SessionAwareDeterministic` sampler? then customers will have to explicitly set it on the configuration side and enabling some specific flags or using different tracer will not surprisingly disable counting for the session sampling rate?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
hamorillo: I checked, and AFAIU this shouldn't affect the effective rate reported by the telemetry.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: > we need to address it in v4, removing this API, probably.

I agree.  If I'm not wrong, iOS doesn't allow customers to set their own sampler so we probably should align with it in this case and keep only the `setTraceSampleRate`.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/SessionRebasedSampler.kt
hamorillo: > It seems that the former is coming from iOS SDK?

Yes, I used that name to maintain parity (as much as possible) with iOS. The `combinedSampleRate` is essentially the rebased sample rate over the session sample rate.

> What is the difference?

`effectiveSampleRate` is the sample rate that we'll use for sampling, but it is not necessarily a combined or rebased one, as we may not have a session sample rate.  In that case, `effectiveSampleRate` returns the raw configured sample rate.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/SessionRebasedSampler.kt
hamorillo: I've applied your suggestion. It is true that now we need to move the check to a different place: `ApmNetworkInstrumentationConfiguration` and `DatadogInterceptor`, but I think it is easier to understand why we need it.

satween (approved):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
```diff
@@ -118,6 +119,11 @@ class ApmNetworkInstrumentationConfiguration internal constructor(
      * Set the trace sample rate controlling the sampling of APM traces created for
      * auto-instrumented requests. If there is a parent trace attached to the network span created, then its
      * sampling decision will be used instead.
+     *
+     * When `headerPropagationOnly` is enabled, the effective trace sample rate is automatically combined with
+     * the active RUM session sample rate, ensuring the backend receives correct sampling metadata
+     * (`_dd.agent_psr` / `_dd.rule_psr`) for RUM-to-APM correlation.
```
0xnm, line 125: That leads me to the interesting idea: is it even useful to expose `HeaderPropagationOnly` API to the customer? @satween I believe this API shouldn't be controlled by the customer. I cannot imagine a situation where we need to give such control.

Current API doc says:

```
     * This is useful when you want distributed tracing visibility without the overhead
     * of client-side network spans.
```

But this is controlled by the RUM network instrumentation. If there is no RUM it makes no sense to set this to true, I think.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/net/SessionRebasedSamplerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class SessionRebasedSamplerTest {
+
+    // region sample()
+
+    @Test
+    fun `M use rebased rate W sample() { session_sample_rate tag present }`(forge: Forge) {
```
0xnm, line 41: maybe we can do in another PR: wdyt about some integration test which asserts the effective sample rate? would be useful to control our expectations with "close-to-real" setup. Although I guess it will be quite a big volume of events to generate to have some trustworthy value.

hamorillo: Sounds good! I can work on it in a different PR. 🙂 I'll file a follow-up ticket so we don't lose it.

One thing to note is that the decision to sample is now deterministic based on `sessionId` (with `traceId` as fallback when there's no RUM session) and not random, so it should make it easier to test. 

There's already a `HeadBasedSamplingTest` in `reliability/single-fit/okhttp/` so something similar could work.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: @satween, any thoughts about this comment?

satween: Hey there. `HeaderPropagationOnly` API has been added for the backward compatibility with the previous instrumentation. For example if customer set `DatadogInterceptor`  but not `TracingInterceptor`. With Rum enabled `DatadogInterceptor` will not send any spans, only add the `traceId`/`spanId` into the request headers and send RumEvents. 

With the new instrumentation if customer wants to add any APM related logic into the networking instrumentation  - he have to add `ApmInstrumentationConfiguration`. But adding it will enable tracing event reporting by-default. So to replicate the behaviour from the configuration when only `DatadogInterceptor` is provided  i added `HeaderPropagationOnly`. We could make it internal for sure if we believe that nobody needs such set-up.

hamorillo: IIUC, in that case, we should keep the `HeaderPropagationOnly` public; if not, we'll be forcing clients to send APM spans when they only want to add the HTTP headers. Right?

0xnm: I still think having this API can be quite confusing and not very obvious for the customer (we discussed this before though, with other API name), hopefully the documentation is well read and acknowledged.

In our code we use it in the sample app https://github.com/DataDog/dd-sdk-android/blob/2f047a427f5178570e3fbfbf05b56cea3fbb1797/sample/kotlin/src/main/kotlin/com/datadog/android/sample/cronet/CronetImageFragment.kt#L75

And later this `ApmInstrumentationConfiguration` is a source for both `apmNetworkInstrumentation` and `distributedTracingInstrumentation` https://github.com/DataDog/dd-sdk-android/blob/2f047a427f5178570e3fbfbf05b56cea3fbb1797/integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt#L76-L105, where the latter does a copy of this configuration with that flag set up https://github.com/DataDog/dd-sdk-android/blob/2f047a427f5178570e3fbfbf05b56cea3fbb1797/integrations/dd-sdk-android-cronet/src/main/kotlin/com/datadog/android/cronet/CronetIntegrationPlugin.kt#L135.

So what was done by the `DatadogInterceptor` and `TracingInterceptor` without the need of the user (because of two different network instrumentations with a clear logical boundary), now needed to be done by the user, since we have a single entrypoint to the network instrumentation of Cronet - `CronetPlugin`.

I don't have a solution for that though. But maybe worth checking how iOS SDK handles this (enabling/disabling APM spans send depending on the instrumentation approach).

0xnm (dismissed):
LGTM. But with caution: we really need to add integration tests for that, so that we are sure that outcome is matching our expectations and spec. Luckily, we have enough time to do it and also do dogfooding, since we are just in the start of the next SDK dev cycle.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
hamorillo: I agree, it is confusing, and we may want to revisit the public API for v4 to make it simpler for the user because, as you mentioned, `DatadogInterceptor` and `TracingInterceptor` are easier to understand.

hamorillo:
> we really need to add integration tests for that, so that we are sure that outcome is matching our expectations and spec.

I've created RUM-16331, and I'll work on it ASAP.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
satween: > DatadogInterceptor and TracingInterceptor are easier to understand.

Can't agree here. `DatadogInterceptor` extends `TracerInrerceptor`, both sharing same APM logic, but `DatadogInterceptor` basically does RUM integration, where `TracerInterceptor` does APM.  You can set `DatadogInterceptor` as `addNetworkInterceptor` and if Rum feature is disabled it starts behave like a `TracerInterceptor`. Plus, as a user you cannot re-use same configuration for cronet. You have to set-up `DataadogEventListener` and if customer accidentally add their listener later in builder chain - it will break Rum resource timings. 

The confusion with `setHeadersPropagationOnly` comes from backward compatibility with an existing logic. For v4 - we should disable most of those API's, reducing the shape of public API and decreasing the combinations of the setups, so there will be only few possible ways to use the API. ( for example if customer wants the APM headers it means spans also should be sent). But this is product decision as well so for the dogfooding period whole set of API's is open to make sure that new instrumentation behaves same way as previous and with all of them we could compare them 1 by 1 . Keep also in mind that our SDK should behave same way for both Cronet and OkHttp (as much as possible) so most features available for OkHttp sould be available for Cronet (if it's even possible), meanwhile Cronet has completley different architecture and current implementation - is a set of minimal common subset of logic, that could be shared between this two networking libraries.

hamorillo: I think we should move this discussion to a different forum as it is not directly related to this PR.

I've updated the PR to resolve conflicts with `develop`, so I think we are good to go with this one if you agree.
cc: @satween @0xnm

satween (approved):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/ApmNetworkInstrumentationConfiguration.kt
satween: agree, lets go

PR #3406, opened 2026-04-30
RUM-15138: Ignore dropped local active span in RUM-to-APM path

hamorillo:
### What does this PR do?

When a local manual span (created via the Trace SDK) is active but has been dropped by the trace sampler, the RUM-to-APM path (DatadogInterceptor / `headerPropagationOnly=true`) now ignores it and creates an independent trace with its own sampling decision. This prevents the Trace SDK's drop decision from killing RUM-APM correlation for network requests where the RUM SDK would have sampled the trace.

The fix only affects the case where the dropped parent is a **local active span with no explicit parent context**. Explicit parent contexts — set via `Request.Builder().parentSpan(...)`, request tags, or inbound trace propagation headers (Datadog, W3C, B3) — are always honored regardless of their sampling priority. This preserves distributed-trace continuity when an upstream service or developer code has explicitly chosen a parent.

### Motivation

When a customer creates a manual span via the Trace SDK that gets dropped (e.g., 5% trace rate), and then makes a network request through DatadogInterceptor, HBS previously propagated the drop — the RUM resource lost all trace correlation. The customer configured the RUM SDK to sample at a higher rate, but the Trace SDK's lower rate overrode it. This change lets the RUM SDK make its own independent sampling decision in that scenario, while still honoring any explicit parent the developer or upstream service has propagated.

### Behavior matrix (RUM path only — `canSendSpan()=false`)

| Parent source | Priority | Behavior |
|---|---|---|
| Local active span (`tracer.activeSpan()`) | `SAMPLER_DROP` / `USER_DROP` | Ignore parent, start fresh trace (new behavior) |
| Local active span | `SAMPLER_KEEP` / `USER_KEEP` / `UNSET` | Honor parent (unchanged) |
| Request tag (`Request.Builder().parentSpan(...)`) | any | Honor parent (unchanged) |
| Inbound trace headers (Datadog / W3C / B3) | any, including drop | Honor parent (unchanged) |

The `TracingInterceptor` APM-only path (`canSendSpan()=true`) is unaffected — HBS still propagates dropped parents as before.

### Implementation Notes

- **Mechanism**: `TracingInterceptor.buildSpan()` and `ApmNetworkInstrumentationExt.buildSpan()` call `isParentDropped()`, which inspects **only** the local active span when no explicit parent context is present. When that active span resolves to `SAMPLER_DROP` or `USER_DROP`, `ignoreActiveSpan()` is called on the span builder, creating a root span with a fresh trace ID.
- **Lazy sampling resolution**: A manual span backed by a `PendingTrace` can carry `UNSET` priority until propagation / inject runs the sampler. Without forcing resolution, the check would treat such a span as not-dropped, attach the network span as a child, and then watch the APM sampler propagate a `DROP` at inject time — exactly the scenario this code is meant to prevent. `isParentDropped` calls `_TraceInternalProxy.setTracingSamplingPriorityIfNecessary` on the active span's context before reading its priority. This is the same forcing mechanism already used in `extractSamplingDecision` and during inject; it is idempotent (CAS-guarded) and produces the exact decision the trace would have received later.
- **No API surface changes** — both `isParentDropped` helpers are `private`.
- **Depends on PR #3402** (cross-product sampling rebasing) — the `SessionRebasedSampler` from that PR is what makes the independent sampling decision meaningful.
- **Related iOS work**: [dd-sdk-ios#2726](https://github.com/DataDog/dd-sdk-ios/pull/2726).

### DIFF
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
@@ -128,7 +128,7 @@ class ApmNetworkInstrumentation internal constructor(
             return RequestTracingState(requestInfoBuilder)
         }
 
-        val span = tracer.buildSpan(request, networkingLibraryName, traceOrigin)
+        val span = tracer.buildSpan(request, networkingLibraryName, traceOrigin, ignoreDroppedParent = !canSendSpan)
         val isSampled = span.isSampled(request)
         if (span.isRootSpan) {
             span.applyPriority(isSampled, traceSampler)
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
@@ -226,7 +226,7 @@ class ApmNetworkInstrumentation internal constructor(
 
     private fun DatadogSpan.isSampled(request: HttpRequestInfo): Boolean =
         extractRumContext(rumContextPropagator, block = true)
-            .sample(request, traceSampler)
+            .sample(request, traceSampler, ignoreLocalDroppedParent = !canSendSpan)
 
     private fun RequestTracingState.onRequestIntercepted(
         response: HttpResponseInfo?,
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
@@ -128,7 +128,12 @@ class ApmNetworkInstrumentation internal constructor(
             return RequestTracingState(requestInfoBuilder)
         }
 
-        val span = tracer.buildSpan(request, networkingLibraryName, traceOrigin, ignoreDroppedParent = !canSendSpan)
+        val span = tracer.buildSpan(
+            request,
+            networkingLibraryName,
+            traceOrigin,
+            ignoreLocalDroppedParent = !canSendSpan
+        )
         val isSampled = span.isSampled(request)
         if (span.isRootSpan) {
             span.applyPriority(isSampled, traceSampler)
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
@@ -13,6 +13,7 @@ import com.datadog.android.lint.InternalApi
 import com.datadog.android.trace.api.DatadogTracingConstants.PrioritySampling
 import com.datadog.android.trace.api.DatadogTracingConstants.Tags
 import com.datadog.android.trace.api.span.DatadogSpan
+import com.datadog.android.trace.api.span.DatadogSpanContext
 import com.datadog.android.trace.api.tracer.DatadogTracer
 import com.datadog.android.trace.internal.ApmNetworkInstrumentation.Companion.AGENT_PSR_ATTRIBUTE
 import com.datadog.android.trace.internal.ApmNetworkInstrumentation.Companion.ALL_IN_SAMPLE_RATE
@@ -74,14 +75,22 @@ internal fun DatadogSpan.finishRumAware(isSampled: Boolean, canSendSpan: Boolean
 internal fun DatadogTracer.buildSpan(
     request: HttpRequestInfo,
     networkInstrumentationName: String,
-    traceOrigin: String?
+    traceOrigin: String?,
+    ignoreDroppedParent: Boolean
 ): DatadogSpan {
     val parentContext = propagationHelper.extractParentContext(this, request)
+    val shouldIgnoreParent = ignoreDroppedParent && isParentDropped(this, parentContext)
 
-    val span = buildSpan(SPAN_NAME.format(Locale.US, networkInstrumentationName))
+    val builder = buildSpan(SPAN_NAME.format(Locale.US, networkInstrumentationName))
         .withOrigin(traceOrigin)
-        .withParentContext(parentContext)
-        .start()
+
+    if (shouldIgnoreParent) {
+        builder.ignoreActiveSpan()
+    } else {
+        builder.withParentContext(parentContext)
+    }
+
+    val span = builder.start()
 
     span.resourceName = request.url.substringBefore(URL_QUERY_PARAMS_BLOCK_SEPARATOR)
     span.setTag(Tags.KEY_HTTP_URL, request.url)
@@ -89,4 +98,11 @@ internal fun DatadogTracer.buildSpan(
     span.setTag(Tags.KEY_SPAN_KIND, Tags.VALUE_SPAN_KIND_CLIENT)
 
     return span
+}
+
+private fun isParentDropped(tracer: DatadogTracer, explicitParent: DatadogSpanContext?): Boolean {
+    // Only consult the local active span. Explicit parents (request tags or propagated
+    // headers) represent developer intent and must be honored regardless of priority.
+    val priority = if (explicitParent != null) null else tracer.activeSpan()?.samplingPriority
+    return priority == PrioritySampling.SAMPLER_DROP || priority == PrioritySampling.USER_DROP
 }
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
@@ -20,6 +20,7 @@ import com.datadog.android.trace.internal.ApmNetworkInstrumentation.Companion.AL
 import com.datadog.android.trace.internal.ApmNetworkInstrumentation.Companion.SPAN_NAME
 import com.datadog.android.trace.internal.ApmNetworkInstrumentation.Companion.URL_QUERY_PARAMS_BLOCK_SEPARATOR
 import com.datadog.android.trace.internal.ApmNetworkInstrumentation.Companion.ZERO_SAMPLE_RATE
+import com.datadog.android.trace.internal._TraceInternalProxy
 import com.datadog.android.trace.internal._TraceInternalProxy.propagationHelper
 import java.util.Locale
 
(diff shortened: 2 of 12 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -360,6 +367,11 @@ internal constructor(
         return span
     }
 
+    private fun isParentDropped(tracer: DatadogTracer, explicitParent: DatadogSpanContext?): Boolean {
+        val priority = explicitParent?.samplingPriority ?: tracer.activeSpan()?.samplingPriority
```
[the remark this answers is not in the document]
hamorillo: Looks like a valid concern. 

Manual spans created via the Trace SDK have `samplingPriority = UNSET` until propagation forces the decision (`CorePropagation.inject() → setSamplingPriorityIfNecessary()`), so isParentDropped was always returning false for them — defeating the purpose of this change.

Commit 647ed6d fixes this by calling `_TraceInternalProxy.setTracingSamplingPriorityIfNecessary(parentContext)` before reading the priority, which is the same forcing mechanism already used in `extractSamplingDecision` and during `inject()`. It's CAS-guarded and idempotent so I think it should be safe. Applied in both `TracingInterceptor.isParentDropped()` and `ApmNetworkInstrumentationExt.isParentDropped()`.

hamorillo:
@codex review

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
```diff
@@ -74,14 +76,22 @@ internal fun DatadogSpan.finishRumAware(isSampled: Boolean, canSendSpan: Boolean
 internal fun DatadogTracer.buildSpan(
     request: HttpRequestInfo,
     networkInstrumentationName: String,
-    traceOrigin: String?
+    traceOrigin: String?,
+    ignoreDroppedParent: Boolean
 ): DatadogSpan {
     val parentContext = propagationHelper.extractParentContext(this, request)
+    val shouldIgnoreParent = ignoreDroppedParent && isParentDropped(this, parentContext)
```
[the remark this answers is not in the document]
hamorillo: I've being thinking about this comment because it's a bit complex and I think, it describes a scenario where tracing headers on the request (e.g.,` x-datadog-sampling-priority: 0`) re-poison the sampling decision after we ignore a dropped parent. 

However, our PR specifically targets the case where a local active manual span from the Trace SDK is dropped and drags down RUM network spans — in that scenario, the request carries no tracing headers, so `extractSamplingDecision(request)` returns `null` and the `traceSampler` runs independently as intended. If a request does carry explicit **DROP** headers, that represents a deliberate upstream decision about that specific trace, not the accidental inheritance from an unrelated manual span — respecting that **DROP** looks like is correct behavior.

Anyway, happy to listen to other opinions.

kikoveiga (approved):

kikoveiga (dismissed):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
```diff
@@ -128,7 +128,7 @@ class ApmNetworkInstrumentation internal constructor(
             return RequestTracingState(requestInfoBuilder)
         }
 
-        val span = tracer.buildSpan(request, networkingLibraryName, traceOrigin)
+        val span = tracer.buildSpan(request, networkingLibraryName, traceOrigin, ignoreDroppedParent = !canSendSpan)
```
satween: Maybe it's worth to rename (here and in other place) to `ignoreLocalDroppedParent`?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
```diff
@@ -90,3 +100,14 @@ internal fun DatadogTracer.buildSpan(
...
     return span
 }
+
+private fun isParentDropped(tracer: DatadogTracer, explicitParent: DatadogSpanContext?): Boolean {
+    // Only consult the local active span. Explicit parents (request tags or propagated
+    // headers) represent developer intent and must be honored regardless of priority.
+    val activeContext = if (explicitParent != null) null else tracer.activeSpan()?.context()
+    // Force resolution of the active span's sampling priority — a manual span backed
+    // by a PendingTrace can read UNSET until the sampler commits at inject time.
+    activeContext?.let { _TraceInternalProxy.setTracingSamplingPriorityIfNecessary(it) }
```
satween: this is a side-effect of the function and it's unclear from the caller side. we should highlight that this priority would be set in a function name, or (it's better) to move this logic out of this getter

satween:
Are you sure it won't break RUM-8051? worth to discuss with @0xnm as well

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
hamorillo: Good point. I've refactored this.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ApmNetworkInstrumentation.kt
hamorillo: Done!

hamorillo (commented):
> Are you sure it won't break RUM-8051? worth to discuss with @0xnm as well

@satween Good question. I would say that's the main goal of this PR. When there is an active span (and trace) from APM that is not sampled, RUM should ignore it and act as if the trace did not exist. Basically, it should make its own decision. This is the Android counterpart of iOS RUM-15310 (DataDog/dd-sdk-ios#2807).

In any other cases, the implementation from RUM-8051 should remain identical.

Anyway, let's wait for @0xnm's insight.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ApmNetworkInstrumentationExt.kt
```diff
@@ -74,14 +76,28 @@ internal fun DatadogSpan.finishRumAware(isSampled: Boolean, canSendSpan: Boolean
 internal fun DatadogTracer.buildSpan(
     request: HttpRequestInfo,
     networkInstrumentationName: String,
-    traceOrigin: String?
+    traceOrigin: String?,
+    ignoreLocalDroppedParent: Boolean
 ): DatadogSpan {
     val parentContext = propagationHelper.extractParentContext(this, request)
+    // Only consult the local active span. Explicit parents (request tags or propagated
+    // headers) represent developer intent and must be honored regardless of priority.
```
0xnm: Is it the same on iOS? The spec written by @qsellem says:

> But the spans are not attached to the manually created trace and so the sampling decision is taken by the RUM SDK, not the Trace SDK

I guess there is a blind spot here, because this PR says that spans added by something like `Request.Builder.addParentSpan` are unaffected.

Normally, this becomes a change maybe for Cronet only then? Because in case of OkHttp active span is thread-local and this code is going to be executed from the thread created by OkHttp, if `enqueue` is used (so customer cannot associate its own span with this thread unless done somewhere in the upstream interceptor).

hamorillo: Good catch! As we've discussed with @qsellem we should also cover the spans attached via `Request.Builder.tag(DatadogSpan::class.java, span)` (and `parentSpan(...)` / Cronet's `addRequestAnnotation(span)`) in this PR

I've added [6ebf523](https://github.com/DataDog/dd-sdk-android/pull/3406/commits/6ebf5232e7207a9b5c8eb318705db0f1aa8e39bd).

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorNotSendingSpanTest.kt
```diff
@@ -1280,6 +1282,309 @@ internal open class TracingInterceptorNotSendingSpanTest {
...
+        // Given - TraceContext tag (OTel interop) DROP is also treated as local developer
+        // intent on the RUM path; the sampler must be consulted.
+        val traceContext = com.datadog.android.trace.internal.net.TraceContext(
+            fakeTraceId,
+            fakeSpanId,
+            DatadogTracingConstants.PrioritySampling.SAMPLER_DROP
+        )
+        fakeRequest = forgeRequest {
+            it.tag(
+                com.datadog.android.trace.internal.net.TraceContext::class.java,
```
0xnm:
```suggestion
                TraceContext::class.java,
```

integrations/dd-sdk-android-okhttp/src/test/kotlin/com/datadog/android/okhttp/trace/TracingInterceptorNotSendingSpanTest.kt
```diff
@@ -1280,6 +1282,309 @@ internal open class TracingInterceptorNotSendingSpanTest {
...
+
+    @Test
+    fun `M defer to sampler W intercept() { Request TraceContext tag DROP, RUM path }`(
+        @IntForgery(min = 200, max = 300) statusCode: Int,
+        @StringForgery(regex = "[a-f0-9]{32}") fakeTraceId: String,
+        @StringForgery(regex = "[a-f0-9]{16}") fakeSpanId: String
+    ) {
+        // Given - TraceContext tag (OTel interop) DROP is also treated as local developer
+        // intent on the RUM path; the sampler must be consulted.
+        val traceContext = com.datadog.android.trace.internal.net.TraceContext(
```
0xnm:
```suggestion
        val traceContext = TraceContext(
```

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
```diff
@@ -360,7 +381,18 @@ internal constructor(
...
     }
 
-    private fun extractSamplingDecision(request: Request): Boolean? {
+    private fun DatadogSpanContext?.isDropped(): Boolean {
+        val priority = this?.samplingPriority
+        return priority.isDroppedPriority()
+    }
+
+    private fun Int?.isDroppedPriority(): Boolean =
+        this == PrioritySampling.SAMPLER_DROP || this == PrioritySampling.USER_DROP
```
0xnm: there is duplication between here and similar code in `ApmNetworkInstrumentation`, but I guess it will be gone at some point with further changes.

0xnm (dismissed):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/trace/TracingInterceptor.kt
hamorillo: I was aware about it but yeah, let's remove it. I've refactor a bit the code in [d7dd316](https://github.com/DataDog/dd-sdk-android/pull/3406/commits/d7dd316283bb99441658346538a0a4e418f65d21)

0xnm (approved):

Issue #3407, opened 2026-04-30
[dd-sdk-android-okhttp 3.9.0] DatadogInterceptor deadlocks SSE / long-lived streaming responses


waliahimanshu:
### Describe the bug

                                                                                                    
  ## Summary                                                                                                                                                                                                                            
                                                                                                    
  Upgrading `com.datadoghq:dd-sdk-android-okhttp` from **3.8.0 → 3.9.0** breaks                                                                                                                                                         
  GraphQL subscriptions delivered over Server-Sent Events (SSE). On 3.9.0, the                      
  `DatadogInterceptor` appears to read/await the entire response body before
  returning, which never completes for an unbounded SSE stream — `OkHttp Call.execute()`                                                                                                                                                
  never resumes, so the SSE consumer (Apollo Kotlin's `SseNetworkTransport` in our case)
  never starts parsing events.                                                                                                                                                                                                          
                                                                                                    
  3.8.0 with the same OkHttp interceptor stack works correctly: `Call.execute()`
  returns on response headers and the streaming body is read incrementally.                                                                                                                                                             
                                                                    
  ## Affected versions                                                                                                                                                                                                                  
                                                                                                                                                                                                                                        
  - **Broken:** `com.datadoghq:dd-sdk-android-okhttp:3.9.0`                                                                                                                                                                             
  - **Working:** `com.datadoghq:dd-sdk-android-okhttp:3.8.0`                                                                                                                                                                            
                                                                                                                                                                                                                                        
  ## Environment                                                                                                                                                                                                                        
                                                                                                    
  - Android `compileSdk` 35, `minSdk` 26                                                                                                                                                                                                
  - Kotlin 2.x                                                                                      
  - OkHttp (transitive via Datadog/Apollo, default version)                                                                                                                                                                             
  - Apollo Kotlin 4.4.3 with a custom `SseNetworkTransport` that wraps OkHttp via                                                                                                                                                       
    Apollo's `DefaultHttpEngine` (uses synchronous `call.execute()` per
    `dd-sdk-android-gradle-plugin` users will recognize)                                                                                                                                                                                
                                                                                                    
  ## Reproduction                                                                                                                                                                                                                       
                                                                                                                                                                                                                                        
  1. Build an `OkHttpClient` with `DatadogInterceptor` registered as an
     *application* interceptor:                                                                                                                                                                                                         
     ```kotlin                                                                                                                                                                                                                          
[...]

PR #3410, opened 2026-04-30
Fix GraphQL errors extraction for streaming responses

kikoveiga:
### What does this PR do?

Fixes `OkHttpGraphQLAdapter#extractGraphQLErrorAttributes` so it returns early if there is a streaming response. Otherwise, `peekBody` might deadlock as described in the issue below. This is exactly the same mechanism used inside `DatadogInterceptor#getBodyLength`.

### Motivation

#3407 

### DIFF
--- a/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
+++ b/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
@@ -9,6 +9,7 @@ package com.datadog.android.okhttp.internal.graphql
 import androidx.annotation.WorkerThread
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.internal.network.GraphQLHeaders
+import com.datadog.android.internal.network.HttpSpec
 import com.datadog.android.okhttp.internal.OkHttpRequestInfo
 import com.datadog.android.okhttp.internal.OkHttpResponseInfo
 import com.datadog.android.rum.RumAttributes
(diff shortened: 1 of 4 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
```diff
@@ -63,22 +64,31 @@ internal class OkHttpGraphQLAdapter(
...
+    @Suppress("ReturnCount")
     fun extractGraphQLErrorAttributes(
         response: Response,
         graphqlAttributes: Map<String, Any?>,
         internalLogger: InternalLogger
     ): Map<String, Any> {
         if (graphqlAttributes.isEmpty()) return emptyMap()
+        // Streaming responses surface GraphQL errors per-frame, not as a top-level `errors` array.
+        // Draining their bodies via peekBody().string() would block until the (potentially unbounded) body completes.
+        val body = response.body
```
0xnm, line 76: This and below is the code which we ported from `DatadogInterceptor`, but it is worth to be aware of the following notice from OkHttp documentation:

```
Response bodies must be [closed][ResponseBody] and may be consumed only once.
```

I think for our use-case this is fine, we don't need to close it, because:

* the particular type is `RealResponseBody` which consists of 
```
  private val contentTypeString: String?,
  private val contentLength: Long,
  private val source: BufferedSource
```
* and when we read content type or content length, we don't touch `source`
* and when we call `peekBody` it just makes a copy:

```
Peeks up to [byteCount] bytes from the response body and returns them as a new response body
```

and we call it before the body is closed, because otherwise `peekBody` says:

```
   * It is an error to call this method after the body has been consumed.
```

Can you please confirm that you can still access/read response body without any problems after adapter logic is applied?

0xnm (approved):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
kikoveiga: I confirm, that is exactly the purpose of `peekBody`, to not consume the body. Tested this locally and everything is working.

jonathanmos (approved):

PR #3420, opened 2026-05-05
RUM-16123: Move broadcast-receiver dispatch off the main thread to fix ANRs

hamorillo:
### What does this PR do?

Moves `BroadcastReceiverSystemInfoProvider.onReceive()` and `BroadcastReceiverNetworkInfoProvider.onReceive()` off the main thread by trampolining them through a new dedicated single-threaded `BackPressureExecutorService` (`broadcastReceiverExecutorService`) owned by `CoreFeature`. The initial sticky-seed path stays synchronous, so `Datadog.initialize()` still returns with system/network info populated. Adds `@Volatile` to `systemInfo`, `networkInfo`, and `lastNetworkInfo` for cross-thread visibility. `CallbackNetworkInfoProvider` (Android-N+) gets only `@Volatile` since its callbacks already dispatch off the main thread.

### Motivation

Fixes ANRs on the Android SDK. Root cause: `Context.registerReceiver(...)` without a `Handler` argument dispatches on the main thread by default. On low-end devices protected by PairIP (Google Play Integrity Code Transparency), each `onReceive` call incurred non-deterministic overhead from `com.pairip.VMRunner.invoke/executeVM`.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
@@ -203,7 +203,6 @@ internal class CoreFeature(
     internal lateinit var uploadExecutorService: ScheduledThreadPoolExecutor
     internal lateinit var persistenceExecutorService: FlushableExecutorService
     internal lateinit var contextExecutorService: ThreadPoolExecutor
-    internal lateinit var broadcastReceiverExecutorService: ExecutorService
     internal lateinit var backpressureStrategy: BackPressureStrategy
 
     internal var localDataEncryption: Encryption? = null
@@ -380,9 +379,6 @@ internal class CoreFeature(
         ioTasks.forEach {
             it.run()
         }
-
-        broadcastReceiverExecutorService.shutdown()
-        broadcastReceiverExecutorService.awaitTermination(DRAIN_WAIT_SECONDS, TimeUnit.SECONDS)
     }
 
     // region Internal
@@ -599,7 +595,7 @@ internal class CoreFeature(
         // System Info Provider
         systemInfoProvider = BroadcastReceiverSystemInfoProvider(
             internalLogger = internalLogger,
-            executorService = broadcastReceiverExecutorService
+            executorService = contextExecutorService
         )
         systemInfoProvider.register(appContext)
 
@@ -619,7 +615,7 @@ internal class CoreFeature(
         } else {
             BroadcastReceiverNetworkInfoProvider(
                 internalLogger = internalLogger,
-                executorService = broadcastReceiverExecutorService
+                executorService = contextExecutorService
             )
         }
         networkInfoProvider.register(appContext)
@@ -665,12 +661,6 @@ internal class CoreFeature(
         persistenceExecutorService = executorServiceFactory.create(
             internalLogger = internalLogger,
             executorContext = "storage",
-            backPressureStrategy = backpressureStrategy,
-            timeProvider = timeProvider
-        )
-        broadcastReceiverExecutorService = executorServiceFactory.create(
-            internalLogger = internalLogger,
-            executorContext = "broadcast-receiver",
             backPressureStrategy = backpressureStrategy,
             timeProvider = timeProvider
         )
@@ -723,13 +713,11 @@ internal class CoreFeature(
         uploadExecutorService.shutdownNow()
         contextExecutorService.shutdownNow()
         persistenceExecutorService.shutdownNow()
-        broadcastReceiverExecutorService.shutdownNow()
 
         try {
             uploadExecutorService.awaitTermination(1, TimeUnit.SECONDS)
             contextExecutorService.awaitTermination(1, TimeUnit.SECONDS)
             persistenceExecutorService.awaitTermination(1, TimeUnit.SECONDS)
-            broadcastReceiverExecutorService.awaitTermination(1, TimeUnit.SECONDS)
         } catch (e: InterruptedException) {
             try {
                 // Restore the interrupted status
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
@@ -203,6 +203,7 @@ internal class CoreFeature(
     internal lateinit var uploadExecutorService: ScheduledThreadPoolExecutor
     internal lateinit var persistenceExecutorService: FlushableExecutorService
     internal lateinit var contextExecutorService: ThreadPoolExecutor
+    internal lateinit var broadcastReceiverExecutorService: ExecutorService
     internal lateinit var backpressureStrategy: BackPressureStrategy
 
     internal var localDataEncryption: Encryption? = null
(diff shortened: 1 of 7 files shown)


hamorillo:
@codex review

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProvider.kt
```diff
@@ -14,24 +14,36 @@ import android.content.Intent
...
     NetworkInfoProvider {
 
+    @Volatile
     private var networkInfo: NetworkInfo = NetworkInfo()
 
     // region BroadcastReceiver
 
     override fun onReceive(context: Context, intent: Intent?) {
+        executorService.executeSafe(HANDLE_INTENT_OPERATION_NAME, internalLogger) {
+            handleIntent(context)
```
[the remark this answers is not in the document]
hamorillo: As the receivers are context-registered against the application context, they live as long as the process. So they aren't reinstantiated per broadcast. 

I think the comment doesn't apply in this case.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
```diff
@@ -658,6 +668,12 @@ internal class CoreFeature(
             backPressureStrategy = backpressureStrategy,
             timeProvider = timeProvider
         )
+        broadcastReceiverExecutorService = executorServiceFactory.create(
```
0xnm: Threads are not coming for free: they take some memory on the stack and also add more work to the scheduler.

I think it is better to reuse `contextExecutorService`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProvider.kt
```diff
@@ -14,24 +14,36 @@ import android.content.Intent
...
     ThreadSafeReceiver(),
     NetworkInfoProvider {
 
+    @Volatile
     private var networkInfo: NetworkInfo = NetworkInfo()
 
     // region BroadcastReceiver
 
     override fun onReceive(context: Context, intent: Intent?) {
+        executorService.executeSafe(HANDLE_INTENT_OPERATION_NAME, internalLogger) {
```
0xnm, line 41: I think we need to apply this change only to the places where we call system services, maybe? Like for `Intent.ACTION_BATTERY_CHANGED` we just unwrap intent, I don't think there is IPC there, but in case `PowerManager.ACTION_POWER_SAVE_MODE_CHANGED` there is IPC.

Or you want to keep consistent approach everywhere?

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/system/BroadcastReceiverSystemInfoProviderTest.kt
```diff
@@ -473,6 +488,50 @@ internal class BroadcastReceiverSystemInfoProviderTest {
...
+    @Test
+    fun `M dispatch onReceive on background broadcast-receiver thread W onReceive() {real executor}`() {
+        // Given
+        val fakeBackPressureStrategy = BackPressureStrategy(
+            capacity = 128,
+            onThresholdReached = {},
+            onItemDropped = {},
+            backpressureMitigation = BackPressureMitigation.IGNORE_NEWEST
+        )
+        val realExecutor = BackPressureExecutorService(
```
0xnm: I don't think we need to use the real executor, some mock is still enough - we can simply capture the argument of `execute`, run it manually and assert that the value changed as we expected.

But not a blocker if you want to keep a real executor, just setup is more complicated.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProviderTest.kt
```diff
@@ -467,6 +490,46 @@ internal class BroadcastReceiverNetworkInfoProviderTest {
...
+    @Test
+    fun `M dispatch onReceive on background broadcast-receiver thread W onReceive() {real executor}`() {
+        // Given
+        val fakeBackPressureStrategy = BackPressureStrategy(
+            capacity = 128,
+            onThresholdReached = {},
+            onItemDropped = {},
+            backpressureMitigation = BackPressureMitigation.IGNORE_NEWEST
+        )
+        val realExecutor = BackPressureExecutorService(
```
0xnm: same comment as in other test

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
hamorillo: This is a good point. In fact, that was exactly my first thought.

However, I decided to create a new one to avoid introducing more load to the `contextExecutorService`. That said, I agree that a new thread will use some resources and also take some time during the SDK initialization for its creation (which may also be an important thing to consider against my proposal).

In terms of memory usage, I'm not sure about the real impact. I didn't expect it to be high, but I'm open to reusing `contextExecutorService` if you think it's safer.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProvider.kt
hamorillo: I don't think the problem is the IPC itself. I would say that the expensive part is `PairIP` wrapping method bodies used in the `onReceive`. For example calling `getIntExtra`, `getSystemService`, etc. each pay their own PairIP cost, so moving it entirely to the background thread saves the accumulative cost. (PairIP transforms method bodies, so each call inside `onReceive` pays its own wrapper cost — moving the chain off-main pays only the entry cost on main, but I may be wrong here.)

Also, I think it's good for consistency.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt
0xnm: Do we have broadcasts fired very often? I think having notifications processed on the context executor is safe.

hamorillo: > Do we have broadcasts fired very often?

Given the broadcast we are subscribing to, I don't think they fire very often.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProvider.kt
0xnm: I'm not aware of PairIP, but wouldn't it wrap just system calls? The stacktrace provided doesn't make it clear though.

I would assume a simple intent unboxing is fine, since it is just a payload, but I may be wrong.

0xnm: Although thinking about it more and if we push to the context thread only certain intents for a given receiver, it means we can have a race condition if it receives 2 intents in a short period of time if processing these intents results in the updating the same property.

0xnm (approved):

PR #3422, opened 2026-05-06
RUM-15287: pause/resume on foreground/background with grace period

ambushwork:
### What does this PR do?

Adds lifecycle-aware pause/resume to the continuous profiler — profiling pauses when the app goes to background (after a grace period)  and resumes on foreground. Introduces `ProfilingLifecycleObserver`, extends `ContinuousProfilingScheduler `with the grace-period logic.    

### Motivation

RUM-15287

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
@@ -12,13 +12,16 @@ import androidx.annotation.RequiresApi
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.feature.FeatureSdkCore
+import com.datadog.android.core.internal.utils.executeSafe
 import com.datadog.android.core.internal.utils.scheduleSafe
 import com.datadog.android.core.sampling.RateBasedSampler
 import com.datadog.android.internal.FeatureContextKeys
+import com.datadog.android.internal.time.TimeProvider
 import java.util.Locale
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.ScheduledFuture
 import java.util.concurrent.TimeUnit
+import kotlin.math.max
 import kotlin.random.Random
 
 @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
@@ -27,6 +30,7 @@ internal class ContinuousProfilingScheduler(
     private val appContext: Context,
     private val profiler: Profiler,
     private val sdkCore: FeatureSdkCore,
+    private val timeProvider: TimeProvider,
     private val sampleRate: Float,
     private val onActiveWindowStarted: () -> Unit = {}
 ) {
@@ -43,6 +47,18 @@ internal class ContinuousProfilingScheduler(
 
     @Volatile
     internal var isActive: Boolean = false
+
+    @Volatile
+    internal var state: State = State.IDLE
+
+    @Volatile
+    private var cooldownWindowEndMs: Long = 0L
+
+    @Volatile
+    private var activeWindowEndMs: Long = 0L
+
+    @Volatile
+    private var remainingCooldownMs: Long = 0L
 
     private val schedulerExecutor: ScheduledExecutorService = profiler.scheduledExecutorService
 
@@ -69,6 +85,7 @@ internal class ContinuousProfilingScheduler(
         isScheduling = false
         logToUser { LOG_STOPPED }
         pendingFuture?.cancel(false)
+        state = State.IDLE
         isActive = false
     }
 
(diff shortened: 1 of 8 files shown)


gradle/libs.versions.toml
```diff
@@ -157,6 +157,7 @@ androidXRecyclerView = { module = "androidx.recyclerview:recyclerview", version.
 androidXWorkManager = { module = "androidx.work:work-runtime", version.ref = "androidXWorkManager" }
 androidXLifecycleCompose = { module = "androidx.lifecycle:lifecycle-runtime-compose", version.ref = "androidXLifecycle" }
 androidXLifecycleLiveDataKtx = { module = "androidx.lifecycle:lifecycle-livedata-ktx", version.ref = "androidXLifecycle" }
+androidXLifecycleProcess = { module = "androidx.lifecycle:lifecycle-process", version.ref = "androidXLifecycle" }
```
0xnm: we have `ProcessLifecycleMonitor`, won't it be better to use it instead? that way we will be aligned with state of other components using it, otherwise we risk to be in a situation where some parts of SDK think they are in a state A, while other parts of SDK think they are in a state B.

ambushwork: Good call — switched to `ProcessLifecycleMonitor` (`ProfilingLifecycleCallback` implements its `Callback`, registered on the `Application`). Dropped the `androidx.lifecycle:lifecycle-process` dependency.

ambushwork: you are right, actually it was me who moved `ProcessLifecycleMonitor` to internal to prepare for the continuous profiling usage, for some reason I forgot it. now fixed

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingLifecycleCallbackTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.mockito.Mock
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.verifyNoInteractions
+import org.mockito.quality.Strictness
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
```
0xnm: nit: there is nothing forged though

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -86,11 +104,108 @@ internal class ContinuousProfilingScheduler(
...
     }
 
+    /**
+     * Called by [ProfilingLifecycleCallback] when the app moves to the background.
+     *
+     * During an active profiling window, a 10-second grace period is started so brief
+     * interruptions (notification shade, quick-settings) do not truncate the session.
+     * During a cooldown, the timer is paused and resumed on foreground with the remaining time.
+     */
+    @SuppressLint("CheckResult")
```
0xnm: That is the right warning: result is not actually used here, so `executeSafe` should be the choice, not `submitSafe`.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -86,11 +104,108 @@ internal class ContinuousProfilingScheduler(
...
+    fun onBackground() {
+        schedulerExecutor.submitSafe(OPERATION_ON_BACKGROUND, sdkCore.internalLogger) {
+            if (!isScheduling) return@submitSafe
+            when (state) {
+                State.ACTIVE -> {
+                    pendingFuture?.cancel(false)
```
0xnm, line 118: `.cancel` will trigger `CancellationExpection` and this reporting it to the logcat/telemetry https://github.com/DataDog/dd-sdk-android/blob/dcc5eb4b258a22c368dec2b626c27e0d0bf39790/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/ThreadExt.kt#L52-L84

Is there a way to avoid using it? Maybe some property which can be read by the `onGracePeriodExpired`, so that it knows if it should be executed or not?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -86,11 +104,108 @@ internal class ContinuousProfilingScheduler(
...
+     *
+     * - [State.GRACE_PERIOD]: cancel the grace timer; resume the active window for its remaining time.
+     * - [State.PAUSED_COOLDOWN]: resume the cooldown with the remaining time.
+     * - [State.PAUSED_AFTER_GRACE]: start a fresh jittered cooldown.
+     */
+    @Suppress("CheckResult")
+    fun onForeground() {
+        schedulerExecutor.submitSafe(OPERATION_ON_FOREGROUND, sdkCore.internalLogger) {
+            if (!isScheduling) return@submitSafe
+            when (state) {
```
0xnm, line 161:
```suggestion
     */
    fun onForeground() {
        schedulerExecutor.executeSafe(OPERATION_ON_FOREGROUND, sdkCore.internalLogger) {
            if (!isScheduling) return@executeSafe
            when (state) {
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -86,11 +104,108 @@ internal class ContinuousProfilingScheduler(
...
+                State.GRACE_PERIOD -> {
+                    pendingFuture?.cancel(false)
```
0xnm, line 163: same as above

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingLifecycleCallback.kt
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Bridges process-level [ProcessLifecycleMonitor.Callback] events into
+ * [ContinuousProfilingScheduler.onForeground] / [ContinuousProfilingScheduler.onBackground].
+ *
+ * Reusing [ProcessLifecycleMonitor] keeps profiling aligned with the foreground/background
+ * state observed by the rest of the SDK (uploaders, batch metrics).
+ */
+@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+internal class ProfilingLifecycleCallback(
+    private val scheduler: ContinuousProfilingScheduler
```
0xnm, line 22: maybe `ContinuousProfilingScheduler` can implement `ProcessLifecycleMonitor.Callback` directly? but not a big deal

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: If I understand correctly this `loggingAfterExecute` is only triggered on `LoggingScheduledThreadPoolExecutor` instance, but the profiler uses just `Executors.newSingleThreadScheduledExecutor`

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/ProfilingLifecycleCallbackTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingLifecycleCallback.kt
ambushwork: No harm to have a separation on this, I will keep it

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
0xnm: That is true, but if we change type one day, we will hit it.

0xnm (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: If we migrate to `LoggingScheduledThreadPoolExecutor` in the future, we can use `removeOnCancelPolicy` to avoid triggering `afterExecute` which calls `loggingAfterExecute`

0xnm (approved):

PR #3424, opened 2026-05-08
RUM-15352: Replace GraphQL chain wrapper with request tag

kikoveiga:
### What does this PR do?

Replace the existing logic of `wrapChainWithoutDDHeaders` with strip-and-tag step at `DatadogInterceptor.interceptor()`.
Now GraphQL attributes travel though the pipeline in a `Request.tag()`.

### Motivation

The previous logic was unpredictably modifying the request flow and could cause issues in complex interception pipelines. Also as tags are in-memory only, they cannot be serialized and leak to the network.

### DIFF
--- a/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
+++ b/integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
@@ -21,7 +21,7 @@ internal class OkHttpGraphQLAdapter(
     private val graphQLExtractor: GraphQLExtractor = GraphQLExtractor()
 ) {
 
-    fun stripHeadersAndTag(request: Request, builder: Request.Builder) {
+    fun convertHeadersToTag(request: Request, builder: Request.Builder) {
         val attributes = graphQLExtractor.extractGraphQLAttributes(OkHttpRequestInfo(request))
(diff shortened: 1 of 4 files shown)


integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
```diff
@@ -14,63 +14,33 @@ import com.datadog.android.okhttp.internal.OkHttpRequestInfo
...
-                            throwable = e
-                        )
-                        originalChain.proceed(request) // fallback to the original request
-                    }
-                }
-            }
-        } else {
-            originalChain
-        }
+    fun stripHeadersAndTag(request: Request, builder: Request.Builder) {
```
0xnm: nit: maybe `convertHeadersToTag` is a better name? I was originally thinking it removes headers **and** some tag by looking on just name.

0xnm (dismissed):

integrations/dd-sdk-android-okhttp/src/main/kotlin/com/datadog/android/okhttp/internal/graphql/OkHttpGraphQLAdapter.kt
kikoveiga: Yes, renamed it. Thanks!

0xnm (approved):

PR #3425, opened 2026-05-08
RUM-16092: Atomically write NDK crash logs + catch `NumberFormatException`

kikoveiga:
### What does this PR do?

- Switch `write_crash_report` to an atomic write + `fsync` + `rename` pattern, so the NDK `crash_log` file on disk is never observable as zero-filled or torn when the process dies between the buffered write and the kerne's writeback (root cause of the `NumberFormatException`).
- Add a `catch (e: NumberFormatException)` in `NdkCrashLogDeserializer` so corrupt files are handled gracefully.

### DIFF
--- a/features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
+++ b/features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
@@ -17,8 +17,13 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TemporaryFolder
 import org.junit.runner.RunWith
+import java.io.File
+import java.io.IOException
 import java.nio.charset.Charset
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.concurrent.thread
 
 @RunWith(AndroidJUnit4::class)
 class NdkTests {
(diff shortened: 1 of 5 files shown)


0xnm (approved):
While it prevents SDK from crashing, the underlying problem stays - we don't still know _why_ it happens.

If you check [this telemetry](https://github.com/DataDog/dd-sdk-android/blob/dcc5eb4b258a22c368dec2b626c27e0d0bf39790/dd-sdk-android-core/src/main/kotlin/com/datadog/android/ndk/internal/NdkCrashLogDeserializer.kt#L22-L27), you will find more errors, which may be related.

Do you plan any further investigation?

jonathanmos (dismissed):

ambushwork (dismissed):

features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
```diff
@@ -98,6 +105,65 @@ class NdkTests {
...
+                    }
+                } catch (@Suppress("SwallowedException") e: IOException) {
+                    // A missing or unreadable file mid-write counts as a torn observation.
+                    tornObservations.incrementAndGet()
+                }
+            }
+        }
+
+        // Let the reader warm up before triggering the write.
+        Thread.sleep(20)
```
0xnm: won't usages of `Thread.sleep` make test flaky?

features/dd-sdk-android-ndk/src/main/cpp/datadog-ndk.cpp
```diff
@@ -97,13 +100,34 @@ void write_crash_report(int signum,
     const uint64_t timestamp = time_since_epoch();
     const std::string serialized_log = serialize_crash_report(signum, timestamp, signal_name, error_message, error_stacktrace);
 
-    // write the log in the crash log file
-    ofstream logs_file_output_stream(file_path.c_str(),
-                                     ofstream::out | ofstream::trunc);
-    if (logs_file_output_stream.is_open()) {
-        logs_file_output_stream << serialized_log.c_str();
+    // Atomic write+fsync via temp file: avoids torn/zero-filled crash logs when the
```
0xnm, line 103: while you are here, can you please also fix `.append` calls above? `append` modifies in-place, and we don't want `main_context.storage_dir` to be modified. Doesn't matter for the crash, because we are calling this method only once during the app lifetime, but it triggers AI to flag it as an issue.

0xnm (dismissed):

features/dd-sdk-android-ndk/src/androidTest/kotlin/com/datadog/android/ndk/NdkTests.kt
kikoveiga: Yes it could, changed to `ConditionWatcher` pattern which is also used for other tests in the same file.

features/dd-sdk-android-ndk/src/main/cpp/datadog-ndk.cpp
kikoveiga: Changed to use `+` instead of `.append()`.

0xnm (approved):

PR #3427, opened 2026-05-11
RUMS-5868: Reproduce BottomNavigationView tap mis-attribution and clipped-scroll-child hit-test bugs

mariusc83:
### Issue Summary
The Android RUM SDK attributes TAP actions to the wrong view, and emits actions with synthetic / system-only resource-IDs, when a `BottomNavigationView` is laid out on top of a `ViewPager2` whose content (a `NestedScrollView`) hosts a clickable item whose layout bounds extend into the nav-bar region.

### Reproduction Tests
- Unit tests: 2 (in `features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListenerOverlapReproductionTest.kt`)

### What the Tests Prove
1. `M report nav item W onTap {BottomNavigationView overlaps deeper clickable in ViewPager scroll subtree}` — builds a decorView with two sibling subtrees: a `BottomNavigationView`-like child containing a clickable nav item at shallow depth, and a `ViewPager2`-like subtree containing a clickable button at deeper depth. Both hit-test true at the tap point. The test asserts the SDK reports the nav item (the topmost visually-elevated clickable). The current SDK reports the deeper button, so the test fails with `ArgumentsAreDifferent` at the `addAction(TAP, …)` verification.
2. `M ignore clipped scroll child W onTap {hit rect contains tap but view is clipped by scroll parent}` — sets up a single clickable scroll-view child whose `getLocationInWindow + width/height` cover the tap coordinate even though the tap falls outside the parent `NestedScrollView`'s visible clip rect. The test asserts no tap action is emitted. The current SDK still emits a tap action targeting the clipped child, so the test fails with `NoInteractionsWanted` against the RUM monitor mock.

### Root Cause Analysis
`GesturesListener.findTarget(...)` performs a breadth-first traversal of the view hierarchy and overwrites the running `target` every time a child satisfies `findTargetForTap` (`hitTest && isClickable && isVisible`). The LAST matching clickable in BFS order wins, regardless of which view Android's own touch dispatch (`ViewGroup.dispatchTouchEvent`) would route the event to — z-order / elevation is not consulted. In overlapping layouts a deeper-nested clickable in a sibling subtree therefore overrides the nav item the user actually tapped.

`AndroidActionTrackingStrategy.hitTest(...)` decides whether a view contains the tap point using `view.getLocationInWindow(...)` plus the view's raw `width`/`height`. It never asks whether the tap coordinate falls inside the visible drawn region of the view within its scrolling ancestor's clip rect. A `NestedScrollView` child whose lower half is scrolled behind the nav bar still reports its full layout-bounds hit rect — so the SDK happily treats it as a valid tap target.

`GesturesUtils.resourceIdName(...)` already catches `Resources.NotFoundException` and falls back to a hex string, so the SDK does not actually drop the synthetic-ID case. The `E Invalid resource ID 0x00000001` line observed in logcat is emitted by the Android framework's `Resources` implementation before the exception is thrown — not by SDK code. The mis-attribution to the deeper view is the user-visible defect; the logcat noise on synthetic IDs is a secondary symptom that surfaces only because the nav item's ID is non-standard.

### Call Chain
1. `Window.Callback.dispatchTouchEvent` → `WindowCallbackWrapper.dispatchTouchEvent`
2. `GesturesDetectorWrapper.onTouchEvent` → `GestureDetector.onTouchEvent` → `GesturesListener.onSingleTapUp`
3. `GesturesListener.handleTapUp` → `findTarget(decorView, x, y, isScroll=false)`
4. `findTarget` BFS — for each dequeued view: `findTargetForTap` → `AndroidActionTrackingStrategy.findTargetForTap` → `hitTest` + `isValidTapTarget`; every match overwrites `target`
5. `sendTapEventWithTarget(target)` → `GlobalRumMonitor.get(sdk).addAction(TAP, "", attributes)` with the WRONG target's attributes

### Failure Output
```
GesturesListenerOverlapReproductionTest > M report nav item W onTap {BottomNavigationView overlaps deeper clickable in ViewPager scroll subtree}(Forge) FAILED
    org.mockito.exceptions.verification.opentest4j.ArgumentsAreDifferent at GesturesListenerOverlapReproductionTest.kt:188
    Wanted: addAction(TAP, "", <matcher: class=NavMenuItem, id=nav_resource>)
    Got:    addAction(TAP, "", {classname="android.view.View", resource_id="<deeper button name>"})
        at GesturesListener.sendTapEventWithTarget(GesturesListener.kt:278)

GesturesListenerOverlapReproductionTest > M ignore clipped scroll child W onTap {hit rect contains tap but view is clipped by scroll parent}(Forge) FAILED
    org.mockito.exceptions.verification.NoInteractionsWanted at GesturesListenerOverlapReproductionTest.kt:278
    No interactions wanted; found: internalAdvancedRumMonitor.addAction(...)
        at GesturesListener.sendTapEventWithTarget(GesturesListener.kt:278)
```

Both failures occur at `GesturesListener.sendTapEventWithTarget(GesturesListener.kt:278)`, exactly the call site predicted by the root-cause analysis. All 173 other tests in the gestures package continue to pass — the new tests have no side effect on existing behaviour.

---
*Generated by rum:tee-triage-insights*

### DIFF
--- a/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListenerOverlapReproductionTest.kt
+++ b/features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/GesturesListenerOverlapReproductionTest.kt
@@ -1,0 +1,282 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.instrumentation.gestures
+
+import android.view.MotionEvent
+import android.view.View
+import android.view.ViewGroup
+import com.datadog.android.rum.RumActionType
+import com.datadog.android.rum.RumAttributes
+import com.datadog.android.rum.utils.forge.Configurator
+import fr.xgouchet.elmyr.Forge
+import fr.xgouchet.elmyr.junit5.ForgeConfiguration
+import fr.xgouchet.elmyr.junit5.ForgeExtension
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.api.extension.Extensions
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.junit.jupiter.MockitoSettings
+import org.mockito.kotlin.argThat
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.mockito.quality.Strictness
+import java.lang.ref.WeakReference
+
+/**
+ * Reproduction tests for RUMS-5868 / RUM-16187.
+ *
+ * Customer-reported symptom: when a `BottomNavigationView` is laid out on top of a
+ * `ViewPager2` whose content (a `NestedScrollView`) hosts a clickable item whose layout
+ * bounds extend into the region occupied by the nav bar, taps on the `BottomNavigationView`
+ * are attributed to the underlying clickable widget in the scroll view instead of to the
+ * nav-bar item that actually received the click.
+ *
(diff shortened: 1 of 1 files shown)


hamorillo:
See #3473

PR #3429, opened 2026-05-11
RUM-16241: Support model generation from JSON schema when array is top-level element

0xnm:
### What does this PR do?

This PR adds support of handling array as top-level element of the type definition. It is different from just `Array`, because we need to have a wrapper class to no deal with adding JSON items to the array manually.

The goal is to support schema in https://github.com/DataDog/rum-events-format/pull/387.

### DIFF
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ArrayWrapperClassGenerator.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ArrayWrapperClassGenerator.kt
@@ -1,0 +1,185 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.gradle.plugin.jsonschema.generator
+
+import com.datadog.gradle.plugin.jsonschema.TypeDefinition
+import com.squareup.kotlinpoet.AnnotationSpec
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FunSpec
+import com.squareup.kotlinpoet.KModifier
+import com.squareup.kotlinpoet.LIST
+import com.squareup.kotlinpoet.ParameterSpec
+import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.SET
+import com.squareup.kotlinpoet.STRING
+import com.squareup.kotlinpoet.TypeSpec
+import com.squareup.kotlinpoet.jvm.throws
+
+/**
+ * Generates a wrapper Kotlin class for a JSON schema whose root type is an array. The wrapper holds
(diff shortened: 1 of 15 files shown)


0xnm:
@codex review

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/TypeDefinition.kt
```diff
@@ -249,6 +249,28 @@ sealed class TypeDefinition {
         }
     }
 
+    /**
+     * The difference with [Array] is that [TypeDefinition.ArrayWrapperClass] is for the cases when top-level type of
+     * JSON is array, so it needs to carry some extra information and generate class of different shape.
+     */
+    data class ArrayWrapperClass(
```
kikoveiga: I was wondering if it's still possible to try to reuse `TypeDefinition.Array` and drop this new class, e.g.:
- `readSchema()` returns `(typeName, typeDefinition)`
- `FileGenerator.generate(typeDef, rootTypeName)` takes the name explicitly so top-level `TypeDefinition.Array` can route directly to `ArrayWrapperClassGenerator`

kikoveiga (dismissed):

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/TypeDefinition.kt
0xnm: fair suggestion, addressed

kikoveiga (dismissed):

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ArrayWrapperClassGenerator.kt
```diff
@@ -0,0 +1,185 @@
...
+        check(itemsClass is TypeDefinition.Class) {
+            "Top-level array schema must have items of a single object type, " +
+                "but got ${itemsClass::class.simpleName}. " +
+                "Multiple item types (oneOf/anyOf/primitives) are not supported as top-level arrays."
+        }
+
+        // Registers the items class as a known nested type so FileGenerator's
+        // nested-class loop emits it inside the wrapper.
+        val itemTypeName = itemsClass.asKotlinTypeName(rootTypeName)
+        val collectionTypeName = if (definition.uniqueItems) {
```
hamorillo, line 51: nit: Wondering if we should add some test to verify the `SET`  case.

hamorillo (dismissed):
LGTM! 🚀

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/ArrayWrapperClassGenerator.kt
0xnm: This would require [uniqueItems](https://tour.json-schema.org/content/04-Arrays/02-Unique-Items) keyword in the schema, don't know its use-case for us in the future, but let's indeed add a necessary test since generator code claims such support.

hamorillo (approved):

ambushwork (approved):

0xnm:
Closing this PR: it seems that added complexity is not worth it compared to just a few lines of code to individual events into the array. Schema format with array as top-level type may also require changes in the model generators on other platforms.

If we ever feel the need in such feature in model generator, we can revive this PR.

PR #3433, opened 2026-05-12
`timeseries` [3/6] RUM-13949: Add CPU timeseries data-point reader and serializer

satween:
### What does this PR do?

Adds `CpuDatapointReader`, which reads CPU-usage and wraps them as `DataPoint` values; and `CpuEventSerializer`, which converts a batch of CPU `DataPoint`s into a `RumTimeseriesCpuEvent` JSON object ready for ingestion.

### Motivation

Implements CPU metric collection as the first concrete timeseries type, exercising the pipeline infrastructure introduced in the previous commit.

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

PR #3432, opened 2026-05-12
`timeseries` [2/6] RUM-13949: Implement timeseries pipeline infrastructure

satween:
### What does this PR do?

Adds the foundational building blocks shared by all timeseries metric types: the `Timeseries` lifecycle interface, immutable `DataPoint`, the `Buffer`/`DeltaCompression`/`Pipeline` data-processing chain, the `EventWriter` flush boundary, `JsonSerializer` interface, `DataPointsReader`/`VitalReaderWrapper` for OS-metric abstraction, and `RumSessionScopeTimeseries` — the generic per-session scheduler that drives sampling, batching, and writing for any metric type.

image

### Motivation

Establishes the shared infrastructure that CPU and memory collectors will extend, keeping primitive types and pipeline logic isolated so they can be reviewed without metric-specific details.

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

PR #3431, opened 2026-05-12
`timeseries` RUM-13949: Add timeseries JSON schemas and regenerate API surface

satween:
### What does this PR do?

Adds JSON schemas for the CPU and memory timeseries event types (`timeseries-cpu-schema.json`, `timeseries-memory-schema.json`), registers them in the model-generation Gradle task, and updates the generated `api/apiSurface` and `api/dd-sdk-android-rum.api` files to reflect the new `RumTimeseriesCpuEvent`, `RumTimeseriesMemoryEvent`, and `TimeseriesConfiguration` public surface.

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

PR #3434, opened 2026-05-12
`timeseries` [4/6] RUM-13949 Add memory timeseries serializer and session factory

satween:
### What does this PR do?

Adds `MemoryEventSerializer`, which converts a batch of memory `DataPoint`s (heap + native RSS) into a `RumTimeseriesMemoryEvent` JSON object. Introduces `TimeseriesConfiguration` as the public `@ExperimentalRumApi` configuration class (sampling intervals, batch size, background collection flag). Completes the pipeline with `RumSessionScopeTimeseriesFactory`, which wires CPU and memory readers, serializers, and the `TimeseriesConfiguration` together to create fully configured `RumSessionScopeTimeseries` instances for each new session.

### Motivation

Completes the two metric types and provides the factory needed to tie the pipeline to the RUM session lifecycle.

### DIFF
--- a/detekt_custom_safe_calls.yml
+++ b/detekt_custom_safe_calls.yml
@@ -548,7 +548,6 @@ datadog:
       - "java.util.concurrent.ExecutorService.shutdown()"
       - "java.util.concurrent.ExecutorService.shutdownNow()"
       - "java.util.concurrent.Executors.newSingleThreadExecutor()"
-      - "java.util.concurrent.Executors.newSingleThreadScheduledExecutor(java.util.concurrent.ThreadFactory?)"
       - "java.util.concurrent.LinkedBlockingDeque.constructor()"
       - "java.util.concurrent.LinkedBlockingQueue.constructor()"
       - "java.util.concurrent.LinkedBlockingQueue.toArray()"
--- a/detekt_custom_safe_calls.yml
+++ b/detekt_custom_safe_calls.yml
@@ -548,6 +548,7 @@ datadog:
       - "java.util.concurrent.ExecutorService.shutdown()"
       - "java.util.concurrent.ExecutorService.shutdownNow()"
       - "java.util.concurrent.Executors.newSingleThreadExecutor()"
+      - "java.util.concurrent.Executors.newSingleThreadScheduledExecutor(java.util.concurrent.ThreadFactory?)"
       - "java.util.concurrent.LinkedBlockingDeque.constructor()"
       - "java.util.concurrent.LinkedBlockingQueue.constructor()"
       - "java.util.concurrent.LinkedBlockingQueue.toArray()"
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
@@ -16,6 +16,7 @@ import com.datadog.android.rum.internal.toTimeseriesMemorySessionType
 import com.datadog.android.rum.model.TimeseriesMemoryEvent
 import com.google.gson.JsonObject
 import java.util.UUID
+import kotlin.math.pow
 
 internal class MemoryEventSerializer(
     private val sessionId: String,
@@ -23,7 +24,8 @@ internal class MemoryEventSerializer(
     private val sessionType: RumSessionType,
     private val totalRamBytes: Long,
     private val timeProvider: TimeProvider,
-    private val useDeltaCompression: Boolean = false
+    private val useDeltaCompression: Boolean = false,
+    private val precision: Int = DeltaCompression.PRECISION
 ) : JsonSerializer<Double> {
 
     override fun serialize(dataPoints: List<DataPoint<Double>>): JsonObject? {
@@ -75,18 +77,19 @@ internal class MemoryEventSerializer(
     private fun encodeDelta(data: List<TimeseriesMemoryEvent.Data>): JsonObject? {
         if (data.size <= 1) return null
 
+        val scale = 10.0.pow(precision.toDouble()).toLong()
         val ts = data.mapToDeltaCompressed { it.timestamp }
 
         val memoryMaxArray = data.mapToDeltaCompressed {
-            roundToLongSafely(it.dataPoint.memoryMax.toDouble())
+            roundToLongSafely(it.dataPoint.memoryMax.toDouble(), scale = scale)
         }
 
         val memoryPercentArray = data.mapToDeltaCompressed {
-            roundToLongSafely(it.dataPoint.memoryPercent.toDouble())
+            roundToLongSafely(it.dataPoint.memoryPercent.toDouble(), scale = scale)
         }
 
         return JsonObject().apply {
-            addProperty("precision", DeltaCompression.PRECISION)
+            addProperty("precision", precision)
             addProperty("resolution", RESOLUTION_NS)
             add("ts", ts)
             add("memory_max", memoryMaxArray)
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
@@ -1,0 +1,99 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.rum.internal.timeseries.serializer
+
+import com.datadog.android.internal.time.TimeProvider
+import com.datadog.android.rum.RumSessionType
+import com.datadog.android.rum.internal.timeseries.DataPoint
+import com.datadog.android.rum.internal.timeseries.DeltaCompression
+import com.datadog.android.rum.internal.timeseries.DeltaCompression.mapToDeltaCompressed
+import com.datadog.android.rum.internal.timeseries.DeltaCompression.roundToLongSafely
+import com.datadog.android.rum.internal.toTimeseriesMemorySessionType
+import com.datadog.android.rum.model.TimeseriesMemoryEvent
+import com.google.gson.JsonObject
+import java.util.UUID
+
+internal class MemoryEventSerializer(
+    private val sessionId: String,
+    private val applicationId: String,
+    private val sessionType: RumSessionType,
+    private val totalRamBytes: Long,
+    private val timeProvider: TimeProvider,
+    private val useDeltaCompression: Boolean = false
+) : JsonSerializer<Double> {
+
+    override fun serialize(dataPoints: List<DataPoint<Double>>): JsonObject? {
+        if (totalRamBytes <= 0L || dataPoints.isEmpty()) return null
+        val data = dataPoints.map { sample ->
+            val memoryPercent = sample.value / totalRamBytes * PERCENT_FACTOR
+            TimeseriesMemoryEvent.Data(
+                timestamp = sample.timestampNs,
+                dataPoint = TimeseriesMemoryEvent.DataPoint(sample.value, memoryPercent)
+            )
+        }
+        val start = data.firstOrNull()?.timestamp ?: 0L
+        val end = data.lastOrNull()?.timestamp ?: 0L
+        val deltaEncoded = if (useDeltaCompression) encodeDelta(data) else null
+        val schema = if (deltaEncoded != null) {
+            TimeseriesMemoryEvent.Schema.DELTA_OBJECT
+        } else {
+            TimeseriesMemoryEvent.Schema.OBJECT
+        }
+        val json = TimeseriesMemoryEvent(
+            dd = TimeseriesMemoryEvent.Dd(),
+            application = TimeseriesMemoryEvent.Application(id = applicationId),
+            session = TimeseriesMemoryEvent.Session(
+                id = sessionId,
+                type = sessionType.toTimeseriesMemorySessionType()
+            ),
+            source = TimeseriesMemoryEvent.Source.ANDROID,
+            date = timeProvider.getDeviceTimestampMillis(),
+            service = null,
+            version = null,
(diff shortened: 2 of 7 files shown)


satween:
* **#3436** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-android/3436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3435** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-android/3435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3434** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-android/3434?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-android/3434?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3433** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-android/3433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3432** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-android/3432?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3431** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-android/3431?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feature/timeseries`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

satween:
@codex review

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
```diff
@@ -0,0 +1,99 @@
...
+                start = start,
+                end = end,
+                data = data
+            )
+        ).toJson() as JsonObject
+        if (deltaEncoded != null) {
+            val timeseriesJson = json.getAsJsonObject("timeseries")
+            timeseriesJson.remove("data")
+            timeseriesJson.add("data", deltaEncoded)
+        }
```
aleksandr-gringauz, line 73: Could you explain please why we need to do it this way instead of `data` supporting the `deltaEncoded` explicitly?

My concern is that IIUC right now if `deltaEncoded` is not `null`, we are returning a json here that doesn't correspond to `TimeseriesMemoryEvent` type.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
```diff
@@ -0,0 +1,99 @@
...
+            service = null,
+            version = null,
+            timeseries = TimeseriesMemoryEvent.Timeseries(
+                id = UUID.randomUUID().toString(),
+                schema = schema,
+                start = start,
+                end = end,
+                data = data
+            )
+        ).toJson() as JsonObject
```
aleksandr-gringauz: Not related to this PR, just an observation. No call to action.

It is weird that `toJson` here returns `JsonElement` rather than `JsonObject`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/timeseries/TimeseriesConfiguration.kt
```diff
@@ -0,0 +1,55 @@
...
+@JvmOverloads @ExperimentalRumApi
+constructor(
+    bufferSize: Int = DEFAULT_BUFFER_SIZE,
+    intervalMs: Long = DEFAULT_INTERVAL_MS,
+    val collectInBackground: Boolean = false
+) {
+    val bufferSize: Int = if (bufferSize > 0) bufferSize else DEFAULT_BUFFER_SIZE
+    val intervalMs: Long = if (intervalMs >= MIN_INTERVAL_MS) intervalMs else DEFAULT_INTERVAL_MS
+
+    internal val executorFactory: () -> ScheduledExecutorService = {
```
aleksandr-gringauz: Do we really need a new executor here? I think in some other PR you mentioned that you'll use `vitalsExecutorService`.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/timeseries/TimeseriesConfiguration.kt
```diff
@@ -0,0 +1,55 @@
...
+ *                   Defaults to [DEFAULT_BUFFER_SIZE] (≈30 seconds at the default 1 s interval).
+ * @param intervalMs Sampling interval in milliseconds. Must be ≥ [MIN_INTERVAL_MS];
+ *                   out-of-range values fall back to [DEFAULT_INTERVAL_MS].
+ *                   Defaults to [DEFAULT_INTERVAL_MS].
+ * @param collectInBackground Whether to keep sampling timeseries when the app is in background.
+ *                            Defaults to `false`.
+ */
+
+class TimeseriesConfiguration
+@JvmOverloads @ExperimentalRumApi
```
aleksandr-gringauz: Probably `ExperimentalRumApi` should be on the whole class, not only constructor.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/timeseries/TimeseriesConfiguration.kt
```diff
@@ -0,0 +1,55 @@
...
+constructor(
+    bufferSize: Int = DEFAULT_BUFFER_SIZE,
+    intervalMs: Long = DEFAULT_INTERVAL_MS,
+    val collectInBackground: Boolean = false
+) {
+    val bufferSize: Int = if (bufferSize > 0) bufferSize else DEFAULT_BUFFER_SIZE
+    val intervalMs: Long = if (intervalMs >= MIN_INTERVAL_MS) intervalMs else DEFAULT_INTERVAL_MS
```
aleksandr-gringauz: It is probably worth adding the `LongRange` and `IntRange` annotations to the constructor args. We already have this pattern in other places.

Also I think we should log an error if the customer sets incorrect values. Similar to [this](https://github.com/DataDog/dd-sdk-android/blob/e86e98d58be050e447e36854b07c9c28bb936f27/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/sampling/RateBasedSampler.kt#L58).

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
aleksandr-gringauz: I think it will be a bit safer to do `as? JsonObject ?: return null`. Just to be extra safe and not do dangerous casts.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
satween: we are measuring size impact on the backend. This is only for the dogfooding phase. In the final solution we're gonna use delta compressed schema if it saves costs by reducing the size good enough so we could take the query performance complexity it adds

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
satween: this is because we want to adjust schema a bit for the dogfooding and use both delta-compressed schema and the regular one.

satween: Eventually we gonna use Dto class model instead of all of those Json hacks

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/timeseries/TimeseriesConfiguration.kt
satween: seems like this was lost during the rebase, thanks for spotting !

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/timeseries/TimeseriesConfiguration.kt
satween: I tried that but then whole `RumConfiguration` becomes experimental

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializer.kt
aleksandr-gringauz: ok

detekt_custom_safe_calls.yml
```diff
@@ -548,6 +548,7 @@ datadog:
       - "java.util.concurrent.ExecutorService.shutdown()"
       - "java.util.concurrent.ExecutorService.shutdownNow()"
       - "java.util.concurrent.Executors.newSingleThreadExecutor()"
+      - "java.util.concurrent.Executors.newSingleThreadScheduledExecutor(java.util.concurrent.ThreadFactory?)"
```
0xnm: not needed in this PR

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/timeseries/TimeseriesConfiguration.kt
```diff
@@ -0,0 +1,48 @@
...
+ * @param intervalMs Sampling interval in milliseconds. Must be ≥ [MIN_INTERVAL_MS];
+ *                   out-of-range values fall back to [DEFAULT_INTERVAL_MS].
+ *                   Defaults to [DEFAULT_INTERVAL_MS].
+ * @param collectInBackground Whether to keep sampling timeseries when the app is in background.
+ *                            Defaults to `false`.
+ */
+
+class TimeseriesConfiguration
+@JvmOverloads @ExperimentalRumApi
+constructor(
```
0xnm: by making constructor public, we let customers call it, and this approach it different then from `Builder` approach we are using in the API, is it intentional?

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializerTest.kt
```diff
@@ -0,0 +1,294 @@
...
+
+        // When
+        val json = testedSerializer.serialize(samples) ?: error(NON_NULL_JSON_ERROR)
+
+        // Then
+        val timeseries = json.getAsJsonObject(KEY_TIMESERIES)
+        assertThat(timeseries.get(KEY_SCHEMA).asString).isEqualTo(VALUE_SCHEMA_OBJECT)
+        assertThat(timeseries.get(KEY_START).asLong).isEqualTo(fakeTs)
+        assertThat(timeseries.get(KEY_END).asLong).isEqualTo(fakeTs + 1L)
+        UUID.fromString(timeseries.get(KEY_ID).asString)
```
0xnm: if we rely on exception here, we need to show it by calling `assertNotThrows`

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializerTest.kt
```diff
@@ -0,0 +1,294 @@
...
+            applicationId = fakeApplicationId,
+            sessionType = RumSessionType.SYNTHETICS,
+            totalRamBytes = fakeTotalRamBytes,
+            timeProvider = mockTimeProvider
+        )
+
+        // When
+        val json = testedSerializer.serialize(
+            listOf(DataPoint(fakeTs, fakeMemory), DataPoint(fakeTs + 1L, fakeMemory))
+        ) ?: error(NON_NULL_JSON_ERROR)
```
0xnm: it is better to move this assertion as `checkNotNull` in the `// Then` block

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializerTest.kt
```diff
@@ -0,0 +1,294 @@
...
+            applicationId = fakeApplicationId,
+            sessionType = RumSessionType.USER,
+            totalRamBytes = fakeTotalRamBytes,
+            timeProvider = mockTimeProvider,
+            useDeltaCompression = true
+        )
+
+        // When
+        val json = testedSerializer.serialize(listOf(DataPoint(fakeTs, fakeMemory)))
+            ?: error(NON_NULL_JSON_ERROR)
```
0xnm: same as above

aleksandr-gringauz (dismissed):

detekt_custom_safe_calls.yml
satween: removed

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/timeseries/serializer/MemoryEventSerializerTest.kt
satween: yes, some repetative pattern from test generating skill, fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/timeseries/TimeseriesConfiguration.kt
satween: changed to builder pattern here + internal constructor

aleksandr-gringauz (approved):

PR #3439, opened 2026-05-12
Sync Detekt configs with a `static-analysis` template version

0xnm:
### What does this PR do?

It is quite painful to get the right version of the Detekt Common and Public API config when working with `dd-source` in parallel, so this change does the following:

* reads template version from CI pipeline definition
* checks if it matches the version in the `.stamp` file of `config` folder
* if it matches, and both config files are there (also in `config` folder), then run Detekt check
* if version is different, then checkout `dd-source` at the right version and copy the necessary configs, restore `dd-source` back.

The code is 100% AI-generated, I'm not that big Bash/Git expert.

### DIFF
--- a/local_ci.sh
+++ b/local_ci.sh
@@ -48,6 +48,90 @@ done
 # exit on errors
 set -e
 
+# Syncs detekt-common.yml and detekt-public-api.yml from dd-source into config/ at the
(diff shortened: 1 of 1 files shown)


kikoveiga (approved):
Working well on my machine (even inside worktrees), thank you!

PR #3372, opened 2026-04-16
RUM-15139: Use rebased sessionReplaySampleRate for deterministic Session Replay sampling

hamorillo:
### What does this PR do?

Replaces the independent random Session Replay sampling with deterministic sampling based on the RUM session sample rate. Instead of two independent coin flips (RUM + SR), the SR sampling decision now uses a `DeterministicSampler` with a combined rate (`sessionSampleRate * srSampleRate / 100`), keyed on the session ID. This guarantees that any session sampled in for SR is also sampled in for RUM.

Key changes:
- **`RumSessionScope`**: Sends `sessionSampleRate` (instead of `keepSession`) to Session Replay via the inter-feature bus
- **`SessionReplayFeature`**: Replaces `RateBasedSampler` with `DeterministicSampler` using `combinedSampleRate()`. Removes the `keepSession` gate — it's now redundant because the combined threshold is always <= the RUM threshold
- **`DeterministicSampling.combinedSampleRate()`**: New utility in `dd-sdk-android-internal` (equivalent to iOS's `SampleRate.composed(with:)`) for computing the effective rate when composing parent + child sampling
- **`SessionSamplingIdProvider`**: Moved from `dd-sdk-android-rum` to `dd-sdk-android-internal` so it can be shared across features

### Motivation

RUM-15139. Aligns Android SR sampling with the deterministic approach used by the iOS SDK ([dd-sdk-ios#2714](https://github.com/DataDog/dd-sdk-ios/pull/2714)). The previous random sampler meant SR decisions were independent of RUM decisions, requiring the `keepSession` flag as a gate. With deterministic sampling, the math guarantees consistency — making `keepSession` unnecessary and simplifying the code.

#### E2E Verification — Deterministic SR Sampling

The deterministic sampling decision can be verified on a real device using the `android-sdk-event-inspection` skill. Ask Claude Code to:

> Using the android-sdk-event-inspection skill, modify the sample app to use RUM sample rate X% and SR sample rate Y%, capture the session ID and `session_sample_rate`/`session_replay_sample_rate` from the RUM view events, then verify that the `has_replay` field matches the expected deterministic sampling decision using the Knuth hash formula. Revert all changes when done.

Tested with 8 sessions across 5 rate combinations — all matched the Knuth hash math:

| Session ID | RUM Rate | SR Rate | Combined | Hash Position | Expected | Observed |
|------------|----------|---------|----------|---------------|----------|----------|
[...]

PR #3440, opened 2026-05-13
RUM-15512: Apply cross-feature sampling to Continuous Profiling

ambushwork:
### What does this PR do?

Wires continuous profiling sampling to the RUM session by making the per-window decision deterministic on sessionId with a combined rate of sessionSampleRate × continuousSampleRate. 

### Motivation

RUM-15512

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
@@ -14,8 +14,10 @@ import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.core.internal.utils.executeSafe
 import com.datadog.android.core.internal.utils.scheduleSafe
-import com.datadog.android.core.sampling.RateBasedSampler
+import com.datadog.android.core.sampling.DeterministicSampler
 import com.datadog.android.internal.FeatureContextKeys
+import com.datadog.android.internal.sampling.DeterministicSampling
+import com.datadog.android.internal.sampling.SessionSamplingIdProvider
 import com.datadog.android.internal.time.TimeProvider
 import java.util.Locale
 import java.util.concurrent.ScheduledExecutorService
@@ -38,9 +40,14 @@ internal class ContinuousProfilingScheduler(
     @Volatile
     internal var isScheduling = false
 
-    // TODO RUM-15315: Read RUM session tracked state instead of using default value.
     @Volatile
-    internal var rumSessionSampled = false
+    internal var currentSessionId: String? = null
+
+    @Volatile
+    internal var currentSessionSampled = false
+
+    @Volatile
+    private var launchSessionExtended = false
 
     @Volatile
     private var pendingFuture: ScheduledFuture<*>? = null
@@ -63,10 +70,10 @@ internal class ContinuousProfilingScheduler(
     private val schedulerExecutor: ScheduledExecutorService = profiler.scheduledExecutorService
 
     fun start(launchProfilingActive: Boolean) {
-        isScheduling = RateBasedSampler<Unit>(sampleRate).sample(Unit)
-        profiler.setExtendLaunchSession(isScheduling)
+        isScheduling = sampleRate > 0f
         if (!isScheduling) {
-            logToUser { LOG_DISABLED.format(Locale.US, sampleRate) }
+            profiler.setExtendLaunchSession(false)
+            logToUser { LOG_DISABLED }
             return
         }
         if (launchProfilingActive) {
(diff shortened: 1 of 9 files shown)


ambushwork:
@codex review

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -63,10 +70,10 @@ internal class ContinuousProfilingScheduler(
     private val schedulerExecutor: ScheduledExecutorService = profiler.scheduledExecutorService
 
     fun start(launchProfilingActive: Boolean) {
-        isScheduling = RateBasedSampler<Unit>(sampleRate).sample(Unit)
-        profiler.setExtendLaunchSession(isScheduling)
+        isScheduling = sampleRate > 0f
```
hamorillo, line 73: ❓ If I understand correctly, now `isScheduling` means profiling is enabled. Maybe we should rename the variable, wdty?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -99,8 +106,24 @@ internal class ContinuousProfilingScheduler(
         scheduleCooldown(jitter(CONTINUOUS_COOLDOWN_DURATION_MS))
     }
 
-    fun onRumSessionRenewed(sessionSampled: Boolean) {
-        rumSessionSampled = sessionSampled
+    fun onRumSessionRenewed(sessionId: String, sessionSampleRate: Float) {
```
hamorillo: 💡 In #3402 (still not merged), we are relying on `rumContext` to pass the `sessionSampleRate`. I'm mentioning this because `sessionSampleRate` is constant since the initialization, so we don't necessarily need to include it in the `RumSessionRenewedEvent` or create a new Sampler every time. Not saying this is wrong, just letting you know in case you want to explore that path.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: It's a bit hard to understand here but `isScheduling` doesn't simply mean Profiling is enabled.

`ContinuousProfilingScheduler` manages a cycle of profiling between "active" and "cooldown", once the user sets a positive `ContinuousSamplingRate`, `ContinuousProfilingScheduler` will start to cycle, which means it `isScheduling`, the cross-feature sampling will only apply when it loops to `active` window, and only in active window that profiler is actually running.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: in this case I assume that I will need to wait for your PR to be merged.

hamorillo: PR is already merged.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -192,7 +191,7 @@ internal class ProfilingFeature(
     }
 
     private fun onRumSessionRenewed(event: RumSessionRenewedEvent) {
-        continuousProfilingScheduler?.onRumSessionRenewed(event.sessionSampled)
+        continuousProfilingScheduler?.onRumSessionRenewed(sessionId = event.sessionId)
```
0xnm:
```suggestion
        continuousProfilingScheduler?.onRumSessionRenewed(event.sessionId)
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -99,8 +109,36 @@ internal class ContinuousProfilingScheduler(
...
+        // If the renewal arrives after the 10s safety-net has fired, the launch profile has
+        // already been cancelled and this call is a no-op.
+        if (currentSessionSampled && !launchSessionExtended) {
+            profiler.setExtendLaunchSession(true)
+            launchSessionExtended = true
+        }
+    }
+
+    private fun createSessionSampler(): DeterministicSampler<String>? {
+        val rumContext = sdkCore.getFeatureContext(Feature.RUM_FEATURE_NAME)
```
0xnm: We shouldn't do this, this is a blocking call, so potentially we can wait for quite a time if context queue is busy. Why we cannot just bundle a RUM session sample rate with RUM session renewed event? Or maybe even do as in Session Replay? https://github.com/DataDog/dd-sdk-android/pull/3372/files#diff-55c27b28d24156e3ae280fa1ce53dace62abed3cf7af94729d19f926924c757e

0xnm (changes requested):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
ambushwork: it was introduced by the suggestion of @hamorillo since `sessionSampleRate` is constant and no need to carry it in every event. After the discussion, we decide to roll it back since carrying a constant is harmless and also make the logic easier to understand.

0xnm (approved):

PR #3453, opened 2026-05-18
RUM-16089: Catch exception on the battery level query

0xnm:
### What does this PR do?

Calling a query on the battery level value may throw `RuntimeException`, so adding a `catch` block.

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
@@ -18,6 +18,7 @@ import android.os.Build
 import android.os.PowerManager
 import android.os.PowerManager.ACTION_POWER_SAVE_MODE_CHANGED
 import androidx.annotation.FloatRange
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.internal.time.TimeProvider
 import com.datadog.android.rum.internal.domain.InfoProvider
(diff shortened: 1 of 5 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
```diff
@@ -30,7 +31,8 @@ internal class DefaultBatteryInfoProvider(
     private val batteryManager: BatteryManager? = applicationContext.getSystemService(
         BATTERY_SERVICE
     ) as? BatteryManager,
-    private val batteryLevelPollInterval: Int = BATTERY_POLL_INTERVAL_MS
+    private val batteryLevelPollInterval: Int = BATTERY_POLL_INTERVAL_MS,
+    private val internalLogger: InternalLogger
```
kikoveiga, line 35: nit: Does it make sense to add this parameter after `timeProvider`? So that we have all the default ones in the end.

kikoveiga (approved):

hamorillo (approved):
LGTM!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/battery/DefaultBatteryInfoProvider.kt
0xnm: no strong preference tbh: yes, it is better to have it before default params, but at the same time this a logger - the element which has the least priority in the class functionality (and not related to it).

PR #3456, opened 2026-05-19
Fix ProfilingFeature shutdown and Continuous Profiling timestamp bug

ambushwork:
### What does this PR do?

This PR fixes two existing bug:
* If Multiple SDK instances enable Profiling Feature, when one of them stops the Profiling Feature, it shutdown the `scheduledExecutorService` which is shared for all the instances.
* In `ContinuousProfilingScheduler`, wrong field is assigned when calling `scheduleActiveWindowEnd`

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
@@ -115,7 +115,6 @@ internal class ProfilingFeature(
         profiler.apply {
             stop(sdkCore.name)
             unregisterProfilingCallback(sdkCore.name)
-            scheduledExecutorService.shutdown()
         }
         sdkCore.removeEventReceiver(name)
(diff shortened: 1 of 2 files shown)


features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -115,7 +115,6 @@ internal class ProfilingFeature(
         profiler.apply {
             stop(sdkCore.name)
             unregisterProfilingCallback(sdkCore.name)
-            scheduledExecutorService.shutdown()
```
ambushwork, line 118: `shutdown` is simply dropped here since keeping a single idle thread even if all the sdk instances have stopped doesn't have a big performance punishment.

Alternative solution is that handling the registration count in `PerfettoProfiler`, shutdown the executer service when it counts down to 0, then we also needs to handle the recreation of the executor service when it is reenabled, which is a bit too complicated for an edge case.

0xnm (approved):

PR #3441, opened 2026-05-13
RUM-15344 Pt.2: Wire Profiling Feature with ANR trigger registration

ambushwork:
### What does this PR do?

This PR adds `AnrProfilingTriggerRegistrar` to handle the ANR trigger registration of `ProfilingManager`, it lives under `PerfettoProfiler` in order to have single instance during the application lifecycle.

`AnrProfilingTriggerRegistrar` registers the ANR trigger at the first `ProfilingFeature` initialization and stops at the last `ProfilingFeature` stops, once an ANR is captured, it broadcasts to all the instances.

### Motivation

RUM-15344

### Review checklist (to be filled by reviewers)

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration, e2e)
- [ ] Make sure you discussed the feature or bugfix with the maintaining team in an Issue
- [ ] Make sure each commit and the PR mention the Issue number (cf the [CONTRIBUTING](../CONTRIBUTING.md) doc)

PR #3457, opened 2026-05-20
RUM-15344 Pt.3 Send Profiling ANR event to RUM

ambushwork:
### What does this PR do?

This PR sends Profiling ANR event to RUM when profiler is active for app launch or continuous profiling, then RUM adds an ANR error then send it back to Profiling Feature to attach with profile.

### Motivation
RUM-15344

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ProfilingThreadDump.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ProfilingThreadDump.kt
@@ -13,13 +13,11 @@ package com.datadog.android.internal.profiling
  * in this module without creating a circular dependency on `dd-sdk-android-core`.
  *
  * @param name Thread name.
- * @param state Thread state string (e.g. `Thread.State.name.lowercase()`).
+ * @param state JVM thread state captured at dump time.
  * @param stack Serialized stack trace (lines joined with newline).
- * @param crashed Whether this thread was the one that triggered the ANR.
  */
 data class ProfilingThreadDump(
     val name: String,
-    val state: String,
-    val stack: String,
-    val crashed: Boolean
+    val state: Thread.State,
+    val stack: String
 )
--- a/dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/ProfilingAnrDetectedEventForgeryFactory.kt
+++ b/dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/ProfilingAnrDetectedEventForgeryFactory.kt
@@ -23,6 +23,8 @@ class ProfilingAnrDetectedEventForgeryFactory : ForgeryFactory<ProfilingAnrDetec
                     forge.anInt(min = -2)
                 )
             },
+            anrThreadName = forge.anAlphabeticalString(),
+            anrThreadState = forge.aValueFrom(Thread.State::class.java),
             allThreads = forge.aList { getForgery<ProfilingThreadDump>() }
         )
     }
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/anr/AnrListener.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/anr/AnrListener.kt
@@ -6,12 +6,8 @@
 
 package com.datadog.android.profiling.internal.anr
 
-import com.datadog.android.internal.profiling.ProfilingThreadDump
+import com.datadog.android.internal.profiling.ProfilingAnrDetectedEvent
 
 internal fun interface AnrListener {
-    fun onAnrDetected(
-        detectedAtMs: Long,
-        anrThreadStack: List<StackTraceElement>,
-        allThreads: List<ProfilingThreadDump>
-    )
+    fun onAnrDetected(event: ProfilingAnrDetectedEvent)
 }
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
@@ -86,9 +86,9 @@ internal class PerfettoProfiler(
             }
         }
 
-    internal val anrListener = AnrListener { detectedAtMs, anrThreadStack, allThreads ->
+    internal val anrListener = AnrListener { event ->
         callbackMap.values.forEach { callback ->
-            callback.onAnrDetected(detectedAtMs, anrThreadStack, allThreads)
+            callback.onAnrDetected(event)
         }
     }
 
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/utils/ThreadDumper.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/utils/ThreadDumper.kt
@@ -8,8 +8,8 @@ package com.datadog.android.profiling.internal.utils
 
 import android.os.Looper
 import com.datadog.android.api.InternalLogger
+import com.datadog.android.internal.profiling.ProfilingAnrDetectedEvent
 import com.datadog.android.internal.profiling.ProfilingThreadDump
-import com.datadog.android.internal.utils.asString
 import com.datadog.android.internal.utils.loggableStackTrace
 
 internal class ThreadDumper(
@@ -24,22 +24,26 @@ internal class ThreadDumper(
     @Volatile
     internal var internalLogger: InternalLogger? = null
 
-    fun dump(): Snapshot {
+    fun dump(detectedAtMs: Long): ProfilingAnrDetectedEvent {
         val mainThread = mainThreadProvider()
         val anrThreadStack = mainThread.stackTrace.toList()
-        // TODO RUM-16390: Replace Thread.getAllStackTraces() with per-thread iteration in ThreadDumper
         val allThreads = safeGetAllStackTraces()
             .filterKeys { it != mainThread }
             .filterValues { it.isNotEmpty() }
             .map { (thread, stack) ->
                 ProfilingThreadDump(
                     name = thread.name,
-                    state = thread.state.asString(),
-                    stack = stack.loggableStackTrace(),
-                    crashed = false
+                    state = thread.state,
+                    stack = stack.loggableStackTrace()
                 )
             }
-        return Snapshot(anrThreadStack = anrThreadStack, allThreads = allThreads)
+        return ProfilingAnrDetectedEvent(
+            detectedAtMs = detectedAtMs,
+            anrThreadStack = anrThreadStack,
+            anrThreadName = mainThread.name,
+            anrThreadState = mainThread.state,
+            allThreads = allThreads
+        )
     }
 
     private fun safeGetAllStackTraces(): Map<Thread, Array<StackTraceElement>> {
@@ -55,11 +59,6 @@ internal class ThreadDumper(
             emptyMap()
         }
     }
-
-    data class Snapshot(
-        val anrThreadStack: List<StackTraceElement>,
-        val allThreads: List<ProfilingThreadDump>
-    )
 
     private companion object {
         const val LOG_STACKS_FAILED = "Failed to capture all stack traces for thread dump."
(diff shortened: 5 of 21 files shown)


ambushwork:
@codex reivew

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -440,6 +446,20 @@ internal class RumFeature(
         )
     }
 
+    private fun handleProfilingAnrDetectedEvent(event: ProfilingAnrDetectedEvent) {
+        val anrException = ANRException(event.anrThreadStack.toTypedArray())
+        val allThreads: List<ThreadDump> = event.allThreads.map { it.toThreadDump() }
```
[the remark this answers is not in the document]
ambushwork: Actually this is related to this [comment](https://github.com/DataDog/dd-sdk-android/pull/3441#discussion_r3265017587) from @0xnm. in `ThreadDumper` we actually don't dump main thread twice since we [exclude it](https://github.com/DataDog/dd-sdk-android/blob/feature/continuous-profiling/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/utils/ThreadDumper.kt#L32), but then, as mentioned as the comment, we forgot to forward the main thread stack trace inside `ProfilingAnrDetectedEvent`, a new commit is added to address this issue.

dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/ProfilingAnrDetectedEventForgeryFactory.kt
```diff
@@ -23,6 +23,8 @@ class ProfilingAnrDetectedEventForgeryFactory : ForgeryFactory<ProfilingAnrDetec
                     forge.anInt(min = -2)
                 )
             },
+            anrThreadName = forge.anAlphabeticalString(),
+            anrThreadState = forge.anElementFrom(Thread.State.values().map { it.name.lowercase() }),
```
kikoveiga: nit, ide suggestion
```suggestion
            anrThreadState = forge.anElementFrom(Thread.State.entries.map { it.name.lowercase() }),
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/anr/AnrListener.kt
```diff
@@ -12,6 +12,8 @@ internal fun interface AnrListener {
     fun onAnrDetected(
         detectedAtMs: Long,
         anrThreadStack: List<StackTraceElement>,
+        anrThreadName: String,
+        anrThreadState: String,
         allThreads: List<ProfilingThreadDump>
```
kikoveiga: Can we simply do `onAnrDetected(event: ProfilingAnrDetectedEvent)` here?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/utils/ThreadDumper.kt
```diff
@@ -58,6 +63,8 @@ internal class ThreadDumper(
 
     data class Snapshot(
         val anrThreadStack: List<StackTraceElement>,
+        val anrThreadName: String,
+        val anrThreadState: String,
         val allThreads: List<ProfilingThreadDump>
     )
```
kikoveiga: Is this `data class` needed, or can we reuse `ProfilingAnrDetectedEvent`?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -440,6 +447,27 @@ internal class RumFeature(
...
+            stack = anrException.loggableStackTrace(),
+            crashed = false
+        )
+        val allThreads: List<ThreadDump> =
+            listOf(mainThreadDump) + event.allThreads.map { it.toThreadDump() }
+        GlobalRumMonitor.get(sdkCore).addError(
+            ANRDetectorRunnable.ANR_MESSAGE,
+            RumErrorSource.SOURCE,
+            anrException,
+            mapOf(RumAttributes.INTERNAL_ALL_THREADS to allThreads)
```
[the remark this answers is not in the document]
ambushwork: good point, updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/utils/ThreadDumper.kt
ambushwork: good point, updated

dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/ProfilingAnrDetectedEventForgeryFactory.kt
ambushwork: updated

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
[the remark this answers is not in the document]
aleksandr-gringauz: This looks like a good suggestion.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -440,6 +447,30 @@ internal class RumFeature(
...
+            anrException,
+            mapOf(
+                RumAttributes.INTERNAL_TIMESTAMP to event.detectedAtMs,
+                RumAttributes.INTERNAL_ALL_THREADS to allThreads
+            )
+        )
+    }
+
+    private fun ProfilingThreadDump.toThreadDump(): ThreadDump =
+        ThreadDump(name = name, state = state, stack = stack, crashed = crashed)
```
aleksandr-gringauz: question. IIUC right now this `crashed` is always `false`. Will it ever be `true` for the ANR event? Does `ProfilingManager` expose the information about whether the app crashed?

dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/ProfilingAnrDetectedEventForgeryFactory.kt
ambushwork: actually I have to roll it back `entries` is only available above kotlin 1.9, our `apiVersion` in `kotlinConfig` targets to 1.8, which causes compilation error

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
ambushwork: Good question, `ThreadDump` as a shared data class for all types of ANR/Crash, `crashed` is not always false, but for `ProfilingThreadDump`, with the current `ProfilingManager` implementation `crashed` can only be false, because when it is a fatal ANR, the profiling callback can't be called since the process crashes, I will remove `crashed` of `ProfilingThreadDump` for the moment.

aleksandr-gringauz (dismissed):

dd-sdk-android-internal/src/testFixtures/kotlin/com/datadog/android/internal/tests/elmyr/ProfilingAnrDetectedEventForgeryFactory.kt
kikoveiga: Oh sorry about that!

kikoveiga (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ProfilingThreadDump.kt
```diff
@@ -15,11 +15,9 @@ package com.datadog.android.internal.profiling
  * @param name Thread name.
  * @param state Thread state string (e.g. `Thread.State.name.lowercase()`).
  * @param stack Serialized stack trace (lines joined with newline).
- * @param crashed Whether this thread was the one that triggered the ANR.
  */
 data class ProfilingThreadDump(
     val name: String,
     val state: String,
```
0xnm: why not use `Thread.State` enum directly here?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -440,6 +447,30 @@ internal class RumFeature(
...
+    private fun handleProfilingAnrDetectedEvent(event: ProfilingAnrDetectedEvent) {
+        val anrException = ANRException(event.anrThreadStack.toTypedArray())
+        val mainThreadDump = ThreadDump(
+            name = event.anrThreadName,
+            state = event.anrThreadState,
+            stack = anrException.loggableStackTrace(),
+            crashed = false
+        )
+        val allThreads: List<ThreadDump> =
+            listOf(mainThreadDump) + event.allThreads.map { it.toThreadDump() }
```
0xnm:
```suggestion
        val allThreads = listOf(mainThreadDump) + event.allThreads.map { it.toThreadDump() }
```

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -440,6 +447,30 @@ internal class RumFeature(
         )
     }
 
+    private fun handleProfilingAnrDetectedEvent(event: ProfilingAnrDetectedEvent) {
+        val anrException = ANRException(event.anrThreadStack.toTypedArray())
+        val mainThreadDump = ThreadDump(
+            name = event.anrThreadName,
+            state = event.anrThreadState,
+            stack = anrException.loggableStackTrace(),
+            crashed = false
```
0xnm, line 457: and this is where things become tricky: it could be actually fatal ANR, but we won't know this till the next app run. But on the next app run we need to reject ANR we read from `ApplicationExitInfo`, because we already recorded this one. And this makes me thinking if we should actually store the trigger result in the data store and read profile on the next app run instead and stitch with `ApplicationExitInfo` if available, similar to NDK crashes.

But let's keep it like that for now.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/perfetto/PerfettoProfiler.kt
```diff
@@ -86,9 +86,9 @@ internal class PerfettoProfiler(
             }
         }
 
-    internal val anrListener = AnrListener { detectedAtMs, anrThreadStack, allThreads ->
+    internal val anrListener = AnrListener { event ->
         callbackMap.values.forEach { callback ->
-            callback.onAnrDetected(detectedAtMs, anrThreadStack, allThreads)
+            callback.onAnrDetected(event)
```
0xnm, line 91: we will have also to implement deduplication with `ApplicationExitInfo`

0xnm (dismissed):

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/profiling/ProfilingThreadDump.kt
ambushwork: when generating the ANR error on RUM side, it requires string, so we convert to string here.

0xnm: Yes, but this event is on the profiling side, so we can put enum here and convert to string only when needed. It is better to leverage type system where possible and use generic types only when it is really needed.

ambushwork: updated

aleksandr-gringauz (approved):

0xnm (approved):

kikoveiga (approved):

PR #3463, opened 2026-05-21
RUM-13560: Add Profiling Scenario in Continuous Benchmarking

ambushwork:
### What does this PR do?

This PR adds the scenario of Profiling for continuous benchmarking.

The scenario simulate the CPU workload of 4 different cases:
* CPU intensive 
* Memory allocation
* I/O
* Switching Thread

### Motivation

RUM-13560

### Additional Notes

image

### DIFF
--- a/sample/benchmark/src/main/java/com/datadog/benchmark/sample/DatadogFeaturesInitializer.kt
+++ b/sample/benchmark/src/main/java/com/datadog/benchmark/sample/DatadogFeaturesInitializer.kt
@@ -7,10 +7,14 @@
 package com.datadog.benchmark.sample
 
 import android.content.Intent
+import android.os.Build
 import com.datadog.android.api.SdkCore
 import com.datadog.android.compose.enableComposeActionTracking
 import com.datadog.android.log.Logs
 import com.datadog.android.log.LogsConfiguration
+import com.datadog.android.profiling.ExperimentalProfilingApi
+import com.datadog.android.profiling.Profiling
+import com.datadog.android.profiling.ProfilingConfiguration
 import com.datadog.android.rum.ExperimentalRumApi
 import com.datadog.android.rum.Rum
 import com.datadog.android.rum.RumConfiguration
@@ -72,7 +76,29 @@ internal class DatadogFeaturesInitializer @Inject constructor(
         if (needToEnableTracing(config)) {
             enableTracing()
         }
+
+        if (needToEnableProfiling(config)) {
+            enableProfiling()
+        }
     }
+
+    private fun needToEnableProfiling(config: BenchmarkConfig): Boolean {
+        return isInstrumentedRun(config) && isProfilingScenario(config)
+    }
+
+    @OptIn(ExperimentalProfilingApi::class)
+    private fun enableProfiling() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            val profilingConfig = ProfilingConfiguration.Builder()
+                .setContinuousSampleRate(SAMPLE_IN_ALL_SESSIONS)
+                .setApplicationLaunchSampleRate(0f)
+                .build()
+            Profiling.enable(profilingConfig, sdkCore.get())
+        }
+    }
+
+    private fun isProfilingScenario(config: BenchmarkConfig) =
+        config.scenario == SyntheticsScenario.Profiling
 
     private fun needToEnableSessionReplay(config: BenchmarkConfig): Boolean {
         return isInstrumentedRun(config) && isSessionReplayScenario(config)
@@ -181,12 +207,15 @@ internal class DatadogFeaturesInitializer @Inject constructor(
         SyntheticsScenario.LogsCustom,
         SyntheticsScenario.LogsHeavyTraffic,
         SyntheticsScenario.RumAuto,
+        SyntheticsScenario.Profiling,
         null -> false
     }
 
     private fun isRumScenario(config: BenchmarkConfig) = when (config.scenario) {
         SyntheticsScenario.RumAuto,
-        SyntheticsScenario.RumManual -> true
+        SyntheticsScenario.RumManual,
+        SyntheticsScenario.Profiling -> true
+
         SyntheticsScenario.SessionReplay,
         SyntheticsScenario.SessionReplayCompose,
         SyntheticsScenario.Trace,
@@ -205,6 +234,7 @@ internal class DatadogFeaturesInitializer @Inject constructor(
         SyntheticsScenario.Trace,
         SyntheticsScenario.Upload,
         SyntheticsScenario.RumAuto,
+        SyntheticsScenario.Profiling,
         null -> false
     }
 
--- a/sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/ProfilingScenarioFragment.kt
+++ b/sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/ProfilingScenarioFragment.kt
@@ -1,0 +1,115 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.benchmark.sample.ui.profiling
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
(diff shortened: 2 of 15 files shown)


sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/AllocationWorkload.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val periodMs: Long = 50L,
+    private val bufferBytes: Int = 32 * 1024,
+    private val listSize: Int = 256
+) : Workload {
+    override fun start(scope: CoroutineScope) {
+        scope.launch {
+            while (isActive) {
+                val buffer = ByteArray(bufferBytes).also { Random.nextBytes(it) }
+                val list = List(listSize) { "alloc-$it-${Random.nextInt()}" }
+                blackHole(buffer.size + list.size)
```
aleksandr-gringauz, line 26: What is `blackHole` needed for? Here we are just setting an `Int` to some global volatile `var blackHoleSink`, if I am not mistaken.

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/DispatchWorkload.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlin.math.sqrt
+
+@Suppress("MagicNumber")
+internal class DispatchWorkload(
+    private val periodMs: Long = 10L,
+    private val parallelism: Int = 4
+) : Workload {
+    override fun start(scope: CoroutineScope) {
+        scope.launch {
+            while (isActive) {
```
aleksandr-gringauz: nit. This `isActive` is useless. When the coroutine awakens after being suspended in `delay`, it will automatically check that it is still active.

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/ProfilingScenarioFragment.kt
```diff
@@ -0,0 +1,109 @@
...
+        savedInstanceState: Bundle?
+    ): View? {
+        return inflater.inflate(R.layout.fragment_profiling, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        controller = WorkloadController()
+        bindings.forEach { binding ->
+            view.findViewById<MaterialButton>(binding.buttonId).setOnClickListener {
```
aleksandr-gringauz: nit. This app has ViewBinding enabled, you can use FragmentProfilingBinding here instead of findViewById.

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/DispatchWorkload.kt
satween: I see that this pattern is reproduced in a several descendands of `Workload` class. Maybe it's worth to move `while(true)` login into `loop` method?

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/IoWorkload.kt
```diff
@@ -0,0 +1,38 @@
...
+@Suppress("MagicNumber")
+internal class IoWorkload(
+    private val cacheDir: File,
+    private val periodMs: Long = 200L,
+    private val payloadBytes: Int = 4 * 1024
+) : Workload {
+    override fun start(scope: CoroutineScope) {
+        scope.launch(Dispatchers.IO) {
+            val file = File(cacheDir, "benchmark_workload.tmp")
+            val payload = ByteArray(payloadBytes).also { Random.nextBytes(it) }
```
satween, line 26: Don't you think this will affect the metrics eventually? I mean allocating 4Kb every cycle and then freeing it up?

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/ProfilingScenarioFragment.kt
satween: or compose!

ambushwork: updated

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/DispatchWorkload.kt
ambushwork: removed

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/AllocationWorkload.kt
ambushwork: Basically it's to hold the result of the computation in case of R8 optimisation remove the entire function if it thinks the result is never used.

sample/benchmark/src/main/java/com/datadog/benchmark/sample/DatadogFeaturesInitializer.kt
```diff
@@ -72,6 +76,38 @@ internal class DatadogFeaturesInitializer @Inject constructor(
...
+            val profilingConfig = ProfilingConfiguration.Builder()
+                .setContinuousSampleRate(SAMPLE_IN_ALL_SESSIONS)
+                .setApplicationLaunchSampleRate(0f)
+                .build()
+            Profiling.enable(profilingConfig, sdkCore.get())
+        }
+    }
+
+    private fun isProfilingScenario(config: BenchmarkConfig) = when (config.scenario) {
+        SyntheticsScenario.Profiling -> true
```
satween: nit: maybe just  
```kotlin 
  private fun isProfilingScenario(config: BenchmarkConfig) =  config.scenario ==  SyntheticsScenario.Profiling  
```

or even
```kotlin 
private val BenchmarkConfig.isProfilingScenario
    get () = scenario == SyntheticsScenario.Profiling
```

ambushwork: good point, updated

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/IoWorkload.kt
ambushwork: The goal of the scenario is not to blow up CPU in every workload, but to simulate all possible operations in the real app, allocation is one of the these.

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/DispatchWorkload.kt
ambushwork: in the future we may have more workload without loop, so let's keep it for now

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/AllocationWorkload.kt
aleksandr-gringauz: Interesting.

I would check that this actually worked, because r8 and compiler could theoretically understand that `blackHoleSink` isn't read from and remove it along with the workload code entirely.

Maybe we could simplify and just log `buffer.size + list.size` to logcat.

Up to you here.

ambushwork: in `blackhole` function the value is assigned to a @Volatile field, which will be skipped by the optimizer - Claude says that.

aleksandr-gringauz (approved):

sample/benchmark/src/main/java/com/datadog/benchmark/sample/ui/profiling/workload/AllocationWorkload.kt
aleksandr-gringauz: > he value is assigned to a https://github.com/volatile field, which will be skipped by the optimizer - Claude says that

yes, my claude says it too, but I am skeptical... I don't understand how volatile is even related.

hamorillo (approved):
🚀

PR #3467, opened 2026-05-22
[Benchmarking] Enable RUM for Session Relay baseline run

ambushwork:
### What does this PR do?

 In the benchmark app, SessionReplay and SessionReplayCompose scenarios now enable RUM in both baseline and instrumented runs (instead of only in instrumented), since Session Replay depends on RUM. This way the baseline-vs-instrumented delta isolates Session Replay's own overhead rather than conflating it with the cost of bringing up RUM.


kikoveiga (dismissed):

hamorillo (approved):
LGTM! 🚀 

We just need to take into account that this will impact the number we are currently seeing.

PR #3473, opened 2026-05-26
RUM-16187: Fix tap target mis-attribution for views clipped by scrolling ancestors

hamorillo:
### What does this PR do?

Replaces `View.getLocationInWindow + width × height` with `View.getGlobalVisibleRect` in `AndroidActionTrackingStrategy.hitTest` so the gesture target resolver respects ancestor clipping.

Before this change, a clickable view scrolled past its parent's visible region (for example, a `Button` inside a `NestedScrollView` whose layout position falls behind a sibling `BottomNavigationView`) was still considered a valid tap target. Because `findTarget` performs a BFS that overwrites `target` on every later match, the visually hidden button was visited after the legitimate `BottomNavigationItemView` and won the resolution, causing taps on the `BottomNavigationView` to be attributed to the hidden underlying button.

`getGlobalVisibleRect` returns the view's bounds intersected with all ancestor clip rectangles, so views that are scrolled out of their parent's visible area no longer satisfy the hit-test. Both APIs return coordinates in the window coordinate space, so the comparison against `MotionEvent` coordinates is unchanged.

### Motivation

RUM-16187 (escalated from RUMS-5868).

Customers report that taps on `BottomNavigationView` items get mis-attributed to underlying `Button` views inside the `ViewPager2` page when those buttons sit at scroll positions that place their raw layout rect behind the nav bar. The behaviour was reproduced on stock SDK 3.9.1 and 3.10.0 in a controlled repro app mirroring the customer's layout (`ConstraintLayout` with `ViewPager2.bottom_toTopOf="@id/separator"`, `separator.bottom_toTopOf="@id/bottom_nav"`, `BottomNavigationView.bottom_toBottomOf="parent"`). At the initial scroll position of the `NestedScrollView`, the last-but-one and last items have layout positions that fall in the window-coord range occupied by the nav bar (item_11 at `y ∈ [2132, 2321]` for a tap at `y = 2232`). The stock SDK's raw-bounds hit-test reported these as valid targets.

### Additional Notes

- **Coordinate space** is unchanged. Both `getLocationInWindow` and `getGlobalVisibleRect` return window-space coordinates (top-left of the decorView is origin) per the AOSP Javadocs. The `MotionEvent` coordinates received via `Window.Callback.dispatchTouchEvent` are in the same space, so no conversion is needed.
- **A reproducer project** is attached to RUM-16187 in Jira for local verification.

The fix was verified end-to-end with `dd-sdk-android` published to Maven local against a repro app that mirrors the customer-reported layout. With SDK 3.10.0, three of three overlap-state Messages-tab taps were attributed to `MaterialButton(item_11)`. With this patch applied, all three were correctly attributed to `BottomNavigationItemView(nav_messages)`.

### DIFF
--- a/detekt_custom_unsafe_calls.yml
+++ b/detekt_custom_unsafe_calls.yml
@@ -42,7 +42,6 @@ datadog:
       - "android.view.Choreographer.postFrameCallback():java.lang.IllegalArgumentException"
       - "android.view.MotionEvent.obtain(android.view.MotionEvent):java.lang.IllegalArgumentException"
       - "android.view.View.getGlobalVisibleRect(android.graphics.Rect?):java.lang.NullPointerException"
-      - "android.view.View.getLocationInWindow(kotlin.IntArray):java.lang.IllegalArgumentException"
       - "android.view.ViewPropertyAnimator.setDuration(kotlin.Long):java.lang.IllegalArgumentException"
       - "android.view.ViewTreeObserver.removeOnDrawListener(android.view.ViewTreeObserver.OnDrawListener?):java.lang.IllegalStateException"
       - "android.view.ViewTreeObserver.addOnDrawListener(android.view.ViewTreeObserver.OnDrawListener?):java.lang.IllegalStateException"
--- a/detekt_custom_unsafe_calls.yml
+++ b/detekt_custom_unsafe_calls.yml
@@ -41,6 +41,7 @@ datadog:
       - "android.view.Choreographer.getInstance():java.lang.IllegalStateException"
       - "android.view.Choreographer.postFrameCallback():java.lang.IllegalArgumentException"
       - "android.view.MotionEvent.obtain(android.view.MotionEvent):java.lang.IllegalArgumentException"
+      - "android.view.View.getGlobalVisibleRect(android.graphics.Rect?):java.lang.NullPointerException"
       - "android.view.View.getLocationInWindow(kotlin.IntArray):java.lang.IllegalArgumentException"
       - "android.view.ViewPropertyAnimator.setDuration(kotlin.Long):java.lang.IllegalArgumentException"
       - "android.view.ViewTreeObserver.removeOnDrawListener(android.view.ViewTreeObserver.OnDrawListener?):java.lang.IllegalStateException"
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
@@ -7,6 +7,7 @@
 package com.datadog.android.rum.internal.instrumentation.gestures
 
 import android.content.Context
+import android.graphics.Rect
 import android.view.View
 import android.widget.AbsListView
 import android.widget.ScrollView
@@ -23,7 +24,7 @@ import java.lang.ref.WeakReference
  */
 internal class AndroidActionTrackingStrategy : ActionTrackingStrategy {
 
-    private val coordinatesContainer = IntArray(2)
+    private val visibleRect = Rect()
 
     override fun register(sdkCore: SdkCore, context: Context) {
         // sdkCore & context are not needed in this strategy, no-op
@@ -34,7 +35,7 @@ internal class AndroidActionTrackingStrategy : ActionTrackingStrategy {
     }
 
     override fun findTargetForTap(view: View, x: Float, y: Float): ViewTarget? {
-        return if (hitTest(view, x, y, coordinatesContainer) && isValidTapTarget(view)) {
+        return if (hitTest(view, x, y, visibleRect) && isValidTapTarget(view)) {
             ViewTarget(viewRef = WeakReference(view))
         } else {
             null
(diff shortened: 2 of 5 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
```diff
@@ -60,16 +61,19 @@ internal class AndroidActionTrackingStrategy : ActionTrackingStrategy {
...
-        return !(x < vx || x > vx + w || y < vy || y > vy + h)
+        // Use getGlobalVisibleRect to get the view's bounds intersected with all ancestor
+        // clip rects. This ensures views that are scrolled out of their parent's visible
+        // area (e.g. a child inside a NestedScrollView/RecyclerView that extends behind a
+        // BottomNavigationView) are NOT considered as hit targets in the clipped region.
+        @Suppress("UnsafeThirdPartyFunctionCall") // outRect is never null
+        val isVisible = view.getGlobalVisibleRect(outRect)
+        if (!isVisible) return false
+        val ix = x.toInt()
+        val iy = y.toInt()
```
0xnm: I think this conversion is not really necessary, `Float` can be compared with `Int` as-is.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
```diff
@@ -34,15 +35,15 @@ internal class AndroidActionTrackingStrategy : ActionTrackingStrategy {
...
     override fun findTargetForTap(view: View, x: Float, y: Float): ViewTarget? {
-        return if (hitTest(view, x, y, coordinatesContainer) && isValidTapTarget(view)) {
+        return if (hitTest(view, x, y, visibleRect) && isValidTapTarget(view)) {
             ViewTarget(viewRef = WeakReference(view))
         } else {
             null
         }
     }
 
     override fun findTargetForScroll(view: View, x: Float, y: Float): ViewTarget? {
```
0xnm, line 45: Should we add a test for scroll as well? I see only test for tap.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AbstractGesturesListenerTest.kt
```diff
@@ -124,14 +142,24 @@ internal abstract class AbstractGesturesListenerTest {
...
-            } else {
-                whenever(it.width).thenReturn(diffPosX + forge.anInt(min = 1, max = 10))
-                whenever(it.height).thenReturn(diffPosY + forge.anInt(min = 1, max = 10))
+            whenever(it.width).thenReturn(viewWidth)
+            whenever(it.height).thenReturn(viewHeight)
+
+            // Mock getGlobalVisibleRect to match the hitTest/visible outcome.
+            // When visible=false, getGlobalVisibleRect returns false (view not shown).
+            // Otherwise the Rect fields are assigned directly (Rect.set() is a no-op
+            // in Android unit test stubs with returnDefaultValues=true).
```
0xnm: is it actually true? `Rect` is a pure Java class without any dependencies (for the core functionality), shouldn't it exist as-is, fully functional, in the `android.jar`?

0xnm (dismissed):
Nice change 👍 One concern is that it may change action-to-view attribution in customer data (in widgets, monitors, etc.), but I don't think it will be dramatic and anyway this change is for correctness.

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AbstractGesturesListenerTest.kt
ambushwork: I guess this can be removed.

detekt_custom_unsafe_calls.yml
```diff
@@ -41,6 +41,7 @@ datadog:
       - "android.view.Choreographer.getInstance():java.lang.IllegalStateException"
       - "android.view.Choreographer.postFrameCallback():java.lang.IllegalArgumentException"
       - "android.view.MotionEvent.obtain(android.view.MotionEvent):java.lang.IllegalArgumentException"
+      - "android.view.View.getGlobalVisibleRect(android.graphics.Rect?):java.lang.NullPointerException"
       - "android.view.View.getLocationInWindow(kotlin.IntArray):java.lang.IllegalArgumentException"
```
ambushwork: This can be removed.

ambushwork (dismissed):
Well done!

jonathanmos (dismissed):

features/dd-sdk-android-rum/src/test/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AbstractGesturesListenerTest.kt
hamorillo: Good catch!

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt
hamorillo: Done!

0xnm (approved):
`getGlobalVisibleRect` is heavier than `getLocationInWindow`, so it is worth checking that we don't have performance regression during the active tap/scroll.

jonathanmos (approved):

PR #3476, opened 2026-05-27
RUM-16501: Make file write atomic

0xnm:
### What does this PR do?

We may have a write failure, leading to the corrupted TLV block. This makes further read fail.

This PR makes file write atomic: we will try to rollback to the previous state if write failed.

The approach used is basically to record the original file size and truncate to that mark if write fails. It should work fine even if there is not enough space on the disk.

This change doesn't cover the situation when there is a process death, such case is harder to deal with. Maybe this change will be enough to get rid of majority read errors.

#### Performance impact

```
  ┌──────────────────────────┬────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────┐
  │        Operation         │         Before         │                                           After                                           │                      Δ                       │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ open()                   │ FileOutputStream       │ RandomAccessFile("rw")                                                                    │ same — both single syscalls, comparable cost │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ flock                    │ yes                    │ yes                                                                                       │ unchanged                                    │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ fstat for channel.size() │ —                      │ +1 syscall                                                                                │ ~1 µs                                        │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ ftruncate                │ —                      │ append=true: no-op (JDK short-circuits when new size ≥ current); append=false: +1 syscall │ 0 to ~2 µs                                   │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ lseek/position update    │ —                      │ in-memory field update, no syscall on Linux JDK                                           │ ~0                                           │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ write()                  │ one big buffered write │ same write() call, just looped for short-write safety                                     │ ~same                                        │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ buffer.flip()            │ —                      │ field update                                                                              │ ~0                                           │
  ├──────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │ close()                  │ yes                    │ yes                                                                                       │ unchanged                                    │
  └──────────────────────────┴────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────┘
```

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
@@ -12,9 +12,9 @@ import com.datadog.android.api.storage.RawBatchEvent
 import com.datadog.android.core.internal.persistence.file.lengthSafe
 import com.datadog.android.core.internal.utils.use
 import java.io.File
-import java.io.FileOutputStream
 import java.io.IOException
 import java.io.InputStream
+import java.io.RandomAccessFile
 import java.nio.ByteBuffer
 import java.util.Locale
 import kotlin.math.max
@@ -90,16 +90,23 @@ internal class PlainBatchFileReaderWriter(
     // region Internal
 
     @Throws(IOException::class)
-    @Suppress("UnsafeThirdPartyFunctionCall") // Called within a try/catch block
+    @Suppress("UnsafeThirdPartyFunctionCall", "NestedBlockDepth") // Called within a try/catch block
     private fun lockFileAndWriteData(
         file: File,
         append: Boolean,
         data: RawBatchEvent
     ) {
-        FileOutputStream(file, append).use { outputStream ->
-            outputStream.channel.lock().use {
-                val meta = data.metadata
+        RandomAccessFile(file, "rw").use { raf ->
+            val channel = raf.channel
+            channel.lock().use {
+                // Snapshot the pre-write length so a failed write can be rolled back
+                // to the last good record boundary (TLV records would otherwise become
+                // misaligned and corrupt every subsequent append).
+                val rollbackLength = if (append) channel.size() else 0L
+                channel.truncate(rollbackLength)
+                channel.position(rollbackLength)
 
+                val meta = data.metadata
                 val metaBlockSize = TYPE_SIZE_BYTES + LENGTH_SIZE_BYTES + meta.size
                 val dataBlockSize = TYPE_SIZE_BYTES + LENGTH_SIZE_BYTES + data.data.size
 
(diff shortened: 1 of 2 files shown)


aleksandr-gringauz:
@codex review

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -109,8 +116,27 @@ internal class PlainBatchFileReaderWriter(
...
+                    }
+                } catch (e: IOException) {
+                    // Truncating to a smaller size frees blocks rather than allocating,
+                    // so it succeeds even when the failure was ENOSPC.
+                    try {
+                        channel.truncate(rollbackLength)
+                    } catch (fallbackException: IOException) {
+                        internalLogger.log(
+                            level = InternalLogger.Level.ERROR,
+                            target = InternalLogger.Target.USER,
```
aleksandr-gringauz, line 133: should we also log to telemetry here, like we do for the `WARNING_NOT_ALL_DATA_READ` log?

aleksandr-gringauz (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
0xnm: I don't think it is actionable though. Anyway, if we don't have space anymore, we don't have space for writing telemetry.

kikoveiga: Was going to ask the same, good point @0xnm.

kikoveiga (dismissed):

hamorillo (dismissed):
🚀

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
aleksandr-gringauz: > if we don't have space anymore, we don't have space for writing telemetry.

indeed

OK

aleksandr-gringauz (dismissed):

hamorillo (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -109,8 +116,28 @@ internal class PlainBatchFileReaderWriter(
...
+                            messageBuilder = { ERROR_WRITE_FALLBACK.format(Locale.US, file.path) },
+                            throwable = fallbackException
+                        )
+                    }
+                    @Suppress("ThrowingInternalException") // we are just propahating existing one
```
hamorillo: nit:
```suggestion
                    @Suppress("ThrowingInternalException") // we are just propagating existing one
```

0xnm: thanks! g and h are too close to each other :)

hamorillo (approved):

PR #3479, opened 2026-05-27
RUM-16502: Dedup fatal ANR reported by Profiling pipeline

ambushwork:
### What does this PR do?

Dedups fatal ANRs so the late-crash reporter does not re-emit an ANR that was already reported in-session — covering both the new Profiling pipeline path and the existing `ANRDetectorRunnable` path.

The marker write now lives in `RumViewScope.onAddError`: when an `AddError` event with an `ANRException` throwable is successfully written, `lastFatalAnrSent` is persisted (using `event.eventTime.timestamp`, which is device time and matches `ApplicationExitInfo.timestamp`). On the next launch, `DatadogLateCrashReporter.handleAnrCrash` skips re-emitting when the marker is within a 10s tolerance of `ApplicationExitInfo.timestamp` (the two come from different moments — in-process detection vs. process death — so exact equality won't hold).

Hooking at the write path (rather than at each call site) means:
- Background ANRs that are silently dropped by `RumViewManagerScope.handleBackgroundEvent` no longer write the marker → no false-negative on background-fatal cases.
- Both the Profiling path and the legacy `ANRDetectorRunnable` path are covered by the same hook → the pre-existing duplicate where an in-session foreground ANR turned fatal is also fixed.

### Motivation

RUM-16502 — without this, an ANR detected and uploaded by the profiling pipeline is duplicated as a fatal ANR error in RUM on the next launch. The same gap also existed in the legacy path (foreground ANR reported in-session, user closes the ANR dialog, fatal ANR re-reported next launch) and is closed by the same fix.

### Additional Notes

**Possible regression to the existing late-crash dedup:** the previous check compared `ApplicationExitInfo.timestamp` to itself (marker written by `handleAnrCrash`, read on next launch — exact equality always held). With the 10s tolerance, two **distinct** fatal ANRs whose `ApplicationExitInfo.timestamp` values fall within 10s of each other would now be treated as duplicates.

In practice this requires: ANR → process death → user relaunches the app → second ANR → process death, all within 10s of the first death. The relaunch alone almost always takes longer than 10s, so this collision is only realistic in a tight crash loop, where collapsing the two reports is arguably acceptable. The same-source case (one exit-info entry re-read across multiple launches) still dedups identically since \`|0| <= 10_000\`.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/InternalSdkCore.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/InternalSdkCore.kt
@@ -104,7 +104,11 @@ interface InternalSdkCore : FeatureSdkCore {
     fun deleteLastViewEvent()
 
     /**
-     * Writes timestamp of the last fatal ANR sent.
+     * Writes the timestamp of the last ANR RUM reported. Despite the "fatal" in the name,
+     * this same timestamp is also written for non-fatal ANRs (see
+     * `RumViewScope.persistLastNonFatalAnrSent`) so that the fatal-ANR pipeline, which
+     * replays historical `ApplicationExitInfo` ANRs on the next launch, can dedupe
+     * against an ANR the watchdog/profiling-trigger path already reported in-process.
      */
     @InternalApi
     @WorkerThread
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -853,7 +853,17 @@ internal open class RumViewScope(
                 if (!isFatal) {
                     // if fatal, then we don't have time for the notification, app is crashing
                     onError { it.eventDropped(rumContext.viewId.orEmpty(), StorageEvent.Error()) }
-                    onSuccess { it.eventSent(rumContext.viewId.orEmpty(), StorageEvent.Error()) }
+                    onSuccess {
+                        it.eventSent(rumContext.viewId.orEmpty(), StorageEvent.Error())
+                        if (event.throwable is ANRException) {
+                            // Intentionally writing through `writeLastFatalAnrSent` even though this is a
+                            // non-fatal ANR: both pipelines share the same persisted timestamp so that the
+                            // fatal-ANR replay path (DatadogLateCrashReporter, via ApplicationExitInfo on
+                            // the next launch) can dedupe against an ANR the in-process watchdog/profiling
+                            // trigger already reported. The storage key keeps the legacy "fatal" name.
+                            sdkCore.writeLastFatalAnrSent(event.eventTime.timestamp)
+                        }
+                    }
                 }
             }
             .submit()
--- a/detekt_custom_safe_calls.yml
+++ b/detekt_custom_safe_calls.yml
@@ -1211,6 +1211,7 @@ datadog:
       - "kotlin.ULong.toDouble()"
       - "kotlin.UShort.toShort()"
       - "kotlin.math.abs(kotlin.Float)"
+      - "kotlin.math.abs(kotlin.Long)"
       - "kotlin.math.max(kotlin.Double, kotlin.Double)"
       - "kotlin.math.max(kotlin.Float, kotlin.Float)"
       - "kotlin.math.max(kotlin.Int, kotlin.Int)"
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/DatadogLateCrashReporter.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/DatadogLateCrashReporter.kt
@@ -32,6 +32,7 @@ import com.google.gson.JsonObject
 import java.io.IOException
 import java.util.UUID
 import java.util.concurrent.TimeUnit
+import kotlin.math.abs
 
 internal class DatadogLateCrashReporter(
     private val sdkCore: InternalSdkCore,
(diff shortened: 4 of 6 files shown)


ambushwork:
@codex review

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/RumFeature.kt
```diff
@@ -466,6 +466,13 @@ internal class RumFeature(
                 RumAttributes.INTERNAL_ALL_THREADS to allThreads
             )
         )
+        // Persist the detection timestamp so the next-launch late-crash path can dedup
+        // this same ANR if it turns fatal. See DatadogLateCrashReporter.handleAnrCrash.
+        val internalSdkCore = sdkCore as InternalSdkCore
+        internalSdkCore.getPersistenceExecutorService()
+            .executeSafe("Write last fatal ANR sent (profiling)", sdkCore.internalLogger) {
+                internalSdkCore.writeLastFatalAnrSent(event.detectedAtMs)
```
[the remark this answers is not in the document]
hamorillo: ❓ Is this one valid? Should we take a look at it?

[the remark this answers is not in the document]
ambushwork: Yes, it's relevant, TLDR, `addError` does not guarantee the error will be written into the batch file, working on a fix to call it when the event is actually written

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -853,7 +854,12 @@ internal open class RumViewScope(
                 if (!isFatal) {
                     // if fatal, then we don't have time for the notification, app is crashing
                     onError { it.eventDropped(rumContext.viewId.orEmpty(), StorageEvent.Error()) }
-                    onSuccess { it.eventSent(rumContext.viewId.orEmpty(), StorageEvent.Error()) }
+                    onSuccess {
+                        it.eventSent(rumContext.viewId.orEmpty(), StorageEvent.Error())
+                        if (event.throwable is ANRException) {
+                            persistLastFatalAnrSent(event.eventTime.timestamp)
```
0xnm: I don't think this works, because everything is guarded by the `!isFatal` check. ANR which we get in the trigger is not classified as fatal.

ambushwork: I am not sure I get it, just because the ANR is not fatal, it will pass the check of `!isFatal` and will trigger this func when the writing is successful.

0xnm: My bad, I mixed the `isFatal` and `!isFatal` in my mind. I think `persistLastFatalAnrSent` confused me, because at this stage ANR is not actually fatal.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1705,6 +1711,15 @@ internal open class RumViewScope(
         Log.i("DatadogSynthetics", "$key=$value")
     }
 
+    private fun persistLastFatalAnrSent(timestampMs: Long) {
```
0xnm: this method is already called on the I/O (`writeScope`) thread, no need to push to the persistence executor explicitly.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -853,7 +854,12 @@ internal open class RumViewScope(
                 if (!isFatal) {
                     // if fatal, then we don't have time for the notification, app is crashing
                     onError { it.eventDropped(rumContext.viewId.orEmpty(), StorageEvent.Error()) }
-                    onSuccess { it.eventSent(rumContext.viewId.orEmpty(), StorageEvent.Error()) }
+                    onSuccess {
+                        it.eventSent(rumContext.viewId.orEmpty(), StorageEvent.Error())
+                        if (event.throwable is ANRException) {
```
0xnm, line 858: there is one concern: if customer has `trackNonFatalAnrs`, such ANR will also pass here. I'm not sure if deduplication is useful for ANR coming from the `ANRDetectorRunnable`, because there we have many false positives (and probably we could also deduplicate between `ANRDetectorRunnable` and ANR trigger), but there is no information about ANR origin, so I guess we can live with it.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: removed the function, call it directly in `onSuccess` callback

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: yes indeed, because it finally calls the existing func `writeLastFatalAnrSent`, which is now used by non-fatal as well, instead of changing the function name, I add comment to clarify it's called on purpose

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
ambushwork: yes, this is a trade-off, `onSuccess` is the most reliable place that we can mark the ANR is written, but here it already lost the information about the ANR origin. Checking if it is from Profiling trigger here can make the code quite messy

hamorillo:
@codex review

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/InternalSdkCore.kt
```diff
@@ -104,7 +104,11 @@ interface InternalSdkCore : FeatureSdkCore {
     fun deleteLastViewEvent()
 
     /**
-     * Writes timestamp of the last fatal ANR sent.
+     * Writes the timestamp of the last ANR RUM reported. Despite the "fatal" in the name,
+     * this same timestamp is also written for non-fatal ANRs (see
+     * `RumViewScope.persistLastNonFatalAnrSent`) so that the fatal-ANR pipeline, which
```
hamorillo, line 109: nit: 
`persistLastNonFatalAnrSent` no longer exits

0xnm (approved):

hamorillo (commented):
LGTM! 🚀

PR #3481, opened 2026-05-28
Bootstrap profiling continuous-session sampling via RUM feature-context updates

ambushwork:
### What does this PR do?

Continuous profiling silently skipped every active window when users enabled the Profiling feature **after** RUM had already created a session — a common ordering since RUM is typically enabled first.

#### Bug
`ProfilingFeature` learned about RUM sessions via a one-shot `RumSessionRenewedEvent` that `RumSessionScope.renewSession` dispatched to the Profiling event-receiver slot. If Profiling registered its event receiver after RUM had already renewed the session, the event was sent to a slot that was still empty, the receiver never observed it, and `ContinuousProfilingScheduler.currentSessionSampled` stayed `false`  forever — `LOG_ACTIVE_WINDOW_SKIPPED` on every cycle.

#### Reproduce

  ```kotlin
  Datadog.initialize(...)
  Rum.enable(rumConfig)                // first session created here
  //A synthetic view like `ApplicationLaunch` has the same effect
  GlobalRumMonitor.get().startView(...) // session renewed; event dispatched into the void
  Profiling.enable(...)                 // misses the renewal; scheduler stays "not sampled"
```

Expected: continuous profiling cycles run normally.
Actual: every window is skipped; no continuous profiles ever produced.

#### Fix

  Drop the one-shot event bus. `ProfilingFeature` now implements `FeatureContextUpdateReceiver`. `DatadogCore.setContextUpdateReceiver` synchronously replays the current feature context for every already-registered
  feature on subscription, so when RUM has already created a session, Profiling receives a bootstrap `onContextUpdate(RUM_FEATURE_NAME, …)` with the existing `session_id` and `session_sample_rate` immediately on
  registration. Subsequent renewals flow through the same channel — the RUM feature-context update already runs on every `renewSession` via `DatadogRumMonitor.handleEvent`.

The `RumSessionRenewedEvent` data class is removed; 

### CODE REVIEW

**Comment by ambushwork** [comment]
@codex review

File: features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt [hunk 1]
```diff
@@ -98,6 +101,8 @@ internal class ProfilingFeature(
         }
         continuousProfilingScheduler = scheduler
 
+        sdkCore.setContextUpdateReceiver(this)
```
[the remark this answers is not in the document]
ambushwork: The initial cooldown window is random between 0 to 1 min, the chance that the registration win the cooldown window end is very small, even if it happens, it just skip one active window, no big deal.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/FeatureContextKeys.kt
```diff
@@ -7,12 +7,33 @@
...
+     * Sample rate (0..100) of the current RUM session. Written by the RUM feature into
+     * its own feature context; read by other features that need to combine it with their
+     * own sampling.
+     */
+    const val RUM_SESSION_SAMPLE_RATE: String = "session_sample_rate"
+
+    /**
+     * Sentinel value [RUM_SESSION_ID] takes before the first RUM session has been created
+     * (e.g. SDK initialised in background). Consumers should skip processing when the
+     * session id equals this value.
```
0xnm:
```suggestion
     * session ID equals this value.
```

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/FeatureContextKeys.kt
```diff
@@ -7,12 +7,33 @@
...
+     */
+    const val EMPTY_RUM_SESSION_ID: String = "00000000-0000-0000-0000-000000000000"
```
0xnm: this is wrong place to have it, it is not a key, it is a value

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -203,10 +209,21 @@ internal class ProfilingFeature(
         }
     }
 
-    private fun onRumSessionRenewed(event: RumSessionRenewedEvent) {
+    override fun onContextUpdate(featureName: String, context: Map<String, Any?>) {
```
0xnm, line 213: The downside of this approach is that we will get these notifications after every RUM event processed, not "once per session" as before.

This callback seems light though and the check for known session is done 👍 , so it shouldn't add the burden to the processing speed of RUM pipeline.

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/FeatureContextKeys.kt
ambushwork: Moved to `RumSessionConstants`

0xnm (approved):

PR #3489, opened 2026-06-01
Optimize `isMainProcess` detection

0xnm:
### What does this PR do?

A bit of startup optimization which can shave of a few milliseconds: querying for all packages/processes on the device may be expensive, so we just can rely on the fact that the main process has the same name as package name.

See https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/app/Application.java;l=339-345?q=Application.java

```
    /**
     * Returns the name of the current process. A package's default process name
     * is the same as its package name. Non-default processes will look like
     * "$PACKAGE_NAME:$NAME", where $NAME corresponds to an android:process
     * attribute within AndroidManifest.xml.
     */
```

On the versions below Android P we can use the old approach.

### DIFF
--- a/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
+++ b/dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
@@ -71,6 +71,7 @@ import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
 import org.junit.jupiter.api.io.TempDir
 import org.mockito.Mock
+import org.mockito.Mockito
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
 import org.mockito.kotlin.any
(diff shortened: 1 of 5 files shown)


0xnm:
@codex review

dd-sdk-android-core/src/test/kotlin/com/datadog/android/utils/config/CoreFeatureTestConfiguration.kt
```diff
@@ -147,7 +147,7 @@ internal class CoreFeatureTestConfiguration<T : Context>(
         whenever(mockInstance.uploadExecutorService) doReturn mockUploadExecutor
         whenever(mockInstance.callFactory) doReturn callFactory
         whenever(mockInstance.kronosClock) doReturn mockKronosClock
-        whenever(mockInstance.contextRef) doReturn mockContextRef
```
kikoveiga, line 150: Please cleanup this now unused variable.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/CoreFeatureTest.kt
```diff
@@ -824,18 +879,14 @@ internal class CoreFeatureTest {
 
         // Then
         assertThat(testedFeature.isMainProcess).isFalse()
```
kikoveiga, line 881: nit: worth restoring the `verifyLog(SDK_INITIALIZED_IN_SECONDARY_PROCESS_WARNING_MESSAGE)` assertion that the original test had, for parity with the new `{secondary process}` test?

kikoveiga (dismissed):

kikoveiga (approved):

PR #3495, opened 2026-06-02
RUM-15702 Add trace sampling decision to DatadogEventBridge

arroz:
### What does this PR do?

With deterministic sampling, the decision of sampling requests on native clients became consistent, based on the sampling rate and session ID. However, when a native client loads an instrumented web site on a WebView, there is a disconnect between the sampling decision used by the browser SDK of the page being viewed in the WebView, and the native client's decision. This cannot be fixed by post-processing on the native client when the data is funneled through it, since the browser SDK needs to know if headers should be injected on requests, or not. More details [here](https://datadoghq.atlassian.net/browse/RUM-15702) (internal).

This is the Android counterpart of https://github.com/DataDog/browser-sdk/pull/4516 and https://github.com/DataDog/dd-sdk-ios/pull/2859.

### How?

To fix this problem, this patch introduces a way for Android to obtain the sampling decision from the Browser SDK running on the WebView through the JS bridge. Unlike iOS, this is a simple sync call that obtains the decision from the bridge.

On the JS side, depending on the decision being injected, tracing sample rate will be set to 100 (when the decision is true) or 0 (when it's false). Because of that, the Android SDK re-writes the _dd.rule_psr value with the actual sampling rate configured in the iOS SDK.

### DIFF
--- a/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
+++ b/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
@@ -88,16 +88,12 @@ internal class DatadogEventBridge(
         val rumContext = sdkCore.getFeatureContext(Feature.RUM_FEATURE_NAME, useContextThread = false)
         val tracingContext = sdkCore.getFeatureContext(Feature.TRACING_FEATURE_NAME, useContextThread = false)
 
-        val sessionId = rumContext[SESSION_ID_KEY] as? String
-        val sessionState = rumContext[SESSION_STATE_KEY] as? String
+        val sessionId = (rumContext[SESSION_ID_KEY] as? String)
+            ?.takeIf { rumContext[SESSION_STATE_KEY] == TRACKED_STATE }
         val sessionSampleRate = (rumContext[SESSION_SAMPLE_RATE_KEY] as? Number)?.toFloat()
         val traceSampleRate = (tracingContext[TRACE_SAMPLE_RATE_KEY] as? Number)?.toFloat()
 
-        if (sessionId == null ||
-            sessionState != TRACKED_STATE ||
-            sessionSampleRate == null ||
-            traceSampleRate == null
-        ) {
+        if (sessionId == null || sessionSampleRate == null || traceSampleRate == null) {
             return NULL_STRING
         }
 
--- a/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
+++ b/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
@@ -86,16 +86,20 @@ internal class DatadogEventBridge(
     @JavascriptInterface
     fun getIsTraceSampled(): String {
         val rumContext = sdkCore.getFeatureContext(Feature.RUM_FEATURE_NAME, useContextThread = false)
-        val sessionId = rumContext[SESSION_ID_KEY] as? String ?: return NULL_STRING
-        val sessionState = rumContext[SESSION_STATE_KEY] as? String
-        if (sessionState != TRACKED_STATE) return NULL_STRING
-
-        val sessionSampleRate = (rumContext[SESSION_SAMPLE_RATE_KEY] as? Number)?.toFloat()
-            ?: return NULL_STRING
-
         val tracingContext = sdkCore.getFeatureContext(Feature.TRACING_FEATURE_NAME, useContextThread = false)
+
+        val sessionId = rumContext[SESSION_ID_KEY] as? String
+        val sessionState = rumContext[SESSION_STATE_KEY] as? String
+        val sessionSampleRate = (rumContext[SESSION_SAMPLE_RATE_KEY] as? Number)?.toFloat()
         val traceSampleRate = (tracingContext[TRACE_SAMPLE_RATE_KEY] as? Number)?.toFloat()
-            ?: return NULL_STRING
+
+        if (sessionId == null ||
+            sessionState != TRACKED_STATE ||
+            sessionSampleRate == null ||
+            traceSampleRate == null
+        ) {
+            return NULL_STRING
+        }
 
         val combinedRate = DeterministicSampling.combinedSampleRate(sessionSampleRate, traceSampleRate)
         val sampler = DeterministicSampler<String>(
--- a/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
+++ b/features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
@@ -7,7 +7,12 @@
 package com.datadog.android.webview.internal
 
 import android.webkit.JavascriptInterface
+import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.core.configuration.HostsSanitizer
+import com.datadog.android.core.sampling.DeterministicSampler
+import com.datadog.android.internal.sampling.DeterministicSampling
+import com.datadog.android.internal.sampling.SessionSamplingIdProvider
 import com.google.gson.JsonArray
 
 /**
@@ -20,7 +25,8 @@ import com.google.gson.JsonArray
 internal class DatadogEventBridge(
     internal val webViewEventConsumer: WebViewEventConsumer<String>,
     private val allowedHosts: List<String>,
-    private val privacyLevel: String
+    private val privacyLevel: String,
+    private val sdkCore: FeatureSdkCore
 ) {
 
     // region Bridge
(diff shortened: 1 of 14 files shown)


features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
```diff
@@ -65,12 +65,17 @@ internal class WebViewRumEventConsumer(
         try {
             val timeOffset = event.get(VIEW_KEY_NAME)?.asJsonObject?.get(VIEW_ID_KEY_NAME)
                 ?.asString?.let { offsetProvider.getOffset(it, datadogContext) } ?: 0L
+            val traceSampleRate = (
+                datadogContext.featuresContext[Feature.TRACING_FEATURE_NAME]
+                    ?.get(TRACE_SAMPLE_RATE_KEY) as? Number
```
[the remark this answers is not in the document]
arroz: Fixed.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventMapper.kt
```diff
@@ -49,6 +50,9 @@ internal class WebViewRumEventMapper(
                 // RUM-4084 disable webview SR if host app doesn't have SR
                 dd.remove(DD_REPLAY_STATS)
             }
+            if (traceSampleRate != null && dd.has(RULE_PSR_KEY_NAME)) {
+                dd.addProperty(RULE_PSR_KEY_NAME, traceSampleRate / SAMPLE_ALL_RATE)
```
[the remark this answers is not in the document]
arroz: I believe this is wrong, `rule_psr` should report the uncombined tracing rate.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
```diff
@@ -70,11 +76,51 @@ internal class DatadogEventBridge(
...
+     * Called from the browser-sdk to get the trace sampling decision from the native SDK.
+     * Returns 'true', 'false', or 'null' as a string:
+     * - 'true' if traces should be sampled
+     * - 'false' if traces should not be sampled
+     * - 'null' if no decision can be made (no active session, or no tracing configured)
+     */
+    @JavascriptInterface
+    fun getIsTraceSampled(): String {
+        val rumContext = sdkCore.getFeatureContext(Feature.RUM_FEATURE_NAME, useContextThread = false)
+        val tracingContext = sdkCore.getFeatureContext(Feature.TRACING_FEATURE_NAME, useContextThread = false)
```
hamorillo: ⚠️ I'm a bit worried about these two invocations. Even with `useContextThread = false`, getFeatureContext acquires a read lock (https://github.com/DataDog/dd-sdk-android/blob/0901bbb607fe65d59c08e6132619ed74f4e8ebcc/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/DatadogCore.kt#L257) on the calling thread. 

IIUC `getIsTraceSampled()` is a `@JavascriptInterface` method called synchronously from the WebView JS engine, the JS thread will block for the duration of that lock if a concurrent context write is holding the write.

We ran into the same issue in PR #3342 (https://github.com/DataDog/dd-sdk-android/pull/3342#discussion_r3073813578). I would say that we have a similar case here.

Anyway, @0xnm - @satween, can you confirm that what I'm saying is correct?

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventMapper.kt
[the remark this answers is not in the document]
hamorillo: Are we sure the the `psr` value should be the uncombined, because in the native part we are sending the combined value (not saying that's correct, just double checking as we may need to fix it).

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
0xnm: I agree, I don't think it is a best approach.

WDYT about implementing `FeatureContextUpdateReceiver` maybe?

hamorillo: > WDYT about implementing FeatureContextUpdateReceiver maybe?

Agree, sounds better. We'll need to deal with `removeContextUpdateReceiver` because `DatadogEventBridge` doesn't have a lifecycle so we probably need to handle it in a different place but I think that's a better approach.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
```diff
@@ -65,12 +69,17 @@ internal class WebViewRumEventConsumer(
...
```
satween, line 74: could be `as? Float` I guess. Then `toFloat` won't be required.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
```diff
@@ -70,11 +76,51 @@ internal class DatadogEventBridge(
...
+
+        val sessionId = (rumContext[SESSION_ID_KEY] as? String)
+            ?.takeIf { rumContext[SESSION_STATE_KEY] == TRACKED_STATE }
+        val sessionSampleRate = (rumContext[SESSION_SAMPLE_RATE_KEY] as? Number)?.toFloat()
```
satween, line 92: `as Float` ?

features/dd-sdk-android-webview/src/test/kotlin/com/datadog/android/webview/internal/DatadogEventBridgeTest.kt
```diff
@@ -139,4 +149,222 @@ internal class DatadogEventBridgeTest {
...
+        val sessionId = "c5b3c4ab-fa4a-4de9-8199-a522131ec48a"
+        whenever(mockSdkCore.getFeatureContext(eq(Feature.RUM_FEATURE_NAME), eq(false)))
+            .thenReturn(
+                mapOf(
+                    "session_id" to sessionId,
+                    "session_state" to "TRACKED",
+                    "session_sample_rate" to 100f
+                )
+            )
+        whenever(mockSdkCore.getFeatureContext(eq(Feature.TRACING_FEATURE_NAME), eq(false)))
```
satween: you could use omit `eq` and use parameters directly:

```
whenever(mockSdkCore.getFeatureContext(Feature.TRACING_FEATURE_NAME, false))
```

same for all tests. `eq` only required only if you use `any()`, `capture` or any other modificators

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventConsumer.kt
arroz: The double conversion is used in other places of the codebase as a defensive measure. If the value is written as Integer or other number that is not a float, `as? Float` will discard it silently.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
arroz: Same as above.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
arroz: Addressed this by using `WebViewRumFeature` as the receiver.

features/dd-sdk-android-webview/src/test/kotlin/com/datadog/android/webview/internal/DatadogEventBridgeTest.kt
arroz: Tests were re-written and `eq` is not used any more.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventMapper.kt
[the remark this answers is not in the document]
arroz: Per our discussion, this is both the reply I got from Quentin, and from everything I've read, I believe the value as configured (without being combined) should be reported here. It's also what seems to be happening on all the other SDKs.

features/dd-sdk-android-webview/src/main/kotlin/com/datadog/android/webview/internal/DatadogEventBridge.kt
```diff
@@ -70,11 +75,51 @@ internal class DatadogEventBridge(
...
+     * - 'false' if traces should not be sampled
+     * - 'null' if no decision can be made (no active session, or no tracing configured)
+     */
+    @JavascriptInterface
+    fun getIsTraceSampled(): String {
+        val rumContext = webViewRumFeature?.cachedRumContext.orEmpty()
+        val tracingContext = webViewRumFeature?.cachedTracingContext.orEmpty()
+
+        val sessionId = (rumContext[SESSION_ID_KEY] as? String)
+            ?.takeIf { rumContext[SESSION_STATE_KEY] == TRACKED_STATE }
```
hamorillo, line 91: ❓ I understand that we are adding this guard here to filter the `EXPIRED` state. However, I don't see any tests covering this. Should we add at least one? Is that the reason for the guard?

instrumented/integration/src/androidTest/kotlin/com/datadog/android/sdk/integration/webview/WebViewBridgeTest.kt
```diff
@@ -0,0 +1,90 @@
...
+                    // evaluateJavascript returns JSON-encoded results — strip surrounding quotes
+                    // for string values so "\"true\"" becomes "true".
+                    actual = value?.removeSurrounding("\"")
+                    latch.countDown()
+                }
+            }
+            latch.await(5, TimeUnit.SECONDS)
+            lastActual = actual
+            if (actual == expectedResult) return
+            Thread.sleep(100)
```
hamorillo, line 77: ❓ Do you think we need 100ms? Looks like too much but just checking if there is a reason.

hamorillo (approved):
Left some minor comments. LGTM overall. I would also wait for @satween's approval, though.

Thanks for the changes!

satween (approved):

PR #3496, opened 2026-06-02
RUM-16394 Pt.2: Implement QuotaChecker for profiling quota HTTP check

ambushwork:
### What does this PR do?

Implement QuotaChecker for profiling quota HTTP check

### Motivation

RUM-16394


ambushwork:
@codex review

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    @Suppress("TooGenericExceptionCaught")
+    private fun performCheck(sessionId: String, datadogContext: DatadogContext): QuotaResult {
+        return try {
+            val request = buildRequest(sessionId, datadogContext)
+            val response = callFactory.newCall(request).execute()
+            if (response.isSuccessful) {
+                parseQuotaResult(response.body?.string())
+            } else {
+                handleHttpError(response.code)
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,174 @@
...
+        return try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+            val attrs = JSONObject(body)
+                .getJSONObject(JSON_KEY_DATA)
+                .getJSONObject(JSON_KEY_ATTRIBUTES)
+            val admitted = attrs.optBoolean(JSON_KEY_ADMITTED, true)
+            val rawReason = attrs.optString(JSON_KEY_REASON, "").ifEmpty { null }
+            val reason = parseReason(rawReason)
+            when {
+                reason == QuotaReason.BACKEND_UNAVAILABLE -> QuotaResult(QuotaResult.Decision.ALLOWED, reason)
```
abrooksv: What was your reasoning around special casing `backend_unavailable` and `backend_client_not_initialized`? The backend seems to define both as `admitted=true` currently which is covered by the `else` switch statement

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/quota/QuotaCheckerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+
+        // Then
+        verify(mockCallFactory, times(2)).newCall(any())
+    }
+
+    @Test
+    fun `M not fire second request W checkAsync called twice with same sessionId while first is in-flight`() {
+        // Given — slow response so the first request is still in-flight when we call checkAsync again
+        whenever(mockCall.execute()) doAnswer {
+            Thread.sleep(200)
```
abrooksv: Nit: Would using a CountDownLatch(1) and awaiting here while counting it down after line 422 make this test more deterministic instead of relying on timing?

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/quota/QuotaCheckerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+    @Test
+    fun `M not fire second request W checkAsync called twice with same sessionId`() {
+        // Given
+        val body = """{"data":{"attributes":{"admitted":true,"reason":"quota_ok"}}}"""
+        whenever(mockCall.execute()) doReturn makeResponse(200, body)
+
+        // When
+        testedChecker.checkAsync(fakeSessionId, fakeDatadogContext)
+        awaitCheck()
+        testedChecker.checkAsync(fakeSessionId, fakeDatadogContext) // same session — should be a no-op
```
abrooksv, line 355: Should we add an awaitCheck() after this to make sure it had time to schedule and execute? In case we break the no-op logic, that will make sure it will be incremented to `2`

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/quota/QuotaCheckerTest.kt
```diff
@@ -0,0 +1,519 @@
...
+
+    private val executor = Executors.newSingleThreadExecutor()
+
+    private val capturedResults = mutableListOf<QuotaResult>()
+
+    private lateinit var testedChecker: QuotaChecker
+
+    @BeforeEach
+    fun `set up`() {
+        fakeDatadogContext = fakeDatadogContext.copy(site = DatadogSite.US1)
```
abrooksv, line 81: Should we add a capturedResults.clear() to give each test a clean slate?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private fun parseReason(raw: String?): QuotaReason = when (raw) {
+        "quota_ok" -> QuotaReason.QUOTA_OK
+        "quota_exceeded" -> QuotaReason.QUOTA_EXCEEDED
+        "org_disabled" -> QuotaReason.ORG_DISABLED
+        "backend_unavailable",
+        "backend_client_not_initialized" -> QuotaReason.BACKEND_UNAVAILABLE
+
+        else -> QuotaReason.UNDEFINED
```
abrooksv: If the backend adds in a new value we do not have a mapping for, we will lose the raw value and report it as `undefined`. Do we want to maintain it via a `sealed interface QuotaReason` instead of an enum?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: I updated the logic here, `backend_unavailable` should be a fail open, we still need to check `admitted=true` first

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/quota/QuotaCheckerTest.kt
ambushwork: good catch, updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/quota/QuotaCheckerTest.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: If a new reason is added on backend, even if we accept it  with `sealed interface`, we still don't have any logic to deal with it, it will be same as falling back to `undefined`

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/quota/QuotaCheckerTest.kt
ambushwork: "JUnit 5 creates a new test class instance per test method by default, so capturedResults is always fresh" - Claude code

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
abrooksv: True, I was just unsure what you were doing / planning to do with the `rawValue` field on QuotaReason so was unsure if we needed to save it, (such as passing it as an attribute or something like that). Fee free to ignore

abrooksv (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+        val previousId = lastSessionId.getAndSet(sessionId)
+        if (previousId == sessionId) return // same session: in-flight check (if any) is still valid
+        pendingFuture.getAndSet(null)?.cancel(true)
+        val future = executor.submitSafe(
+            operationName = OPERATION_NAME_QUOTA,
+            internalLogger = internalLogger,
+            callable = {
+                val result = performCheck(sessionId, datadogContext)
+                if (lastSessionId.get() == sessionId) {
+                    lastResult = result
```
satween: Are you sure that this change shouldn't be atomic? 

imagine `lastSessionId.get() == sessionId)` passes, but then other thread calls `lastSessionId.set(...)`. So last results holds value that is not related to the session..

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    private fun parseQuotaResult(body: String?): QuotaResult {
+        if (body == null) return QuotaResult.API_ERROR
+        return try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+            val attrs = JSONObject(body)
+                .getJSONObject(JSON_KEY_DATA)
+                .getJSONObject(JSON_KEY_ATTRIBUTES)
+            val admitted = attrs.optBoolean(JSON_KEY_ADMITTED, true)
+            val rawReason = attrs.optString(JSON_KEY_REASON, "").ifEmpty { null }
```
satween: nit: `optString` falls into empty string by default: 
```
    /**
     * Returns the value mapped by {@code name} if it exists, coercing it if
     * necessary, or the empty string if no such mapping exists.
     */
    @NonNull public String optString(@Nullable String name) {
        return optString(name, "");
    }
```

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: this callable is submitted to the executor which will be assigned by `newSingleThreadScheduledExecutor` in the following commit, so we don't have any race condition here.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: updated!

satween (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+            QuotaResult.API_ERROR
+        } catch (e: Exception) {
+            logErrorToMaintainer(e) { LOG_UNEXPECTED_ERROR.format(Locale.US, e.message) }
+            QuotaResult.API_ERROR
+        }
+    }
+
+    @Suppress("UnsafeThirdPartyFunctionCall") // Wrapped in the try-catch
+    private fun buildRequest(sessionId: String, datadogContext: DatadogContext): Request {
+        val intakeHost = datadogContext.site.intakeEndpoint.removePrefix("https://")
```
0xnm: why do we remove prefix?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    private fun parseQuotaResult(body: String?): QuotaResult {
+        if (body == null) return QuotaResult.API_ERROR
+        return try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // JSONObject operations wrapped in try-catch
+            val attrs = JSONObject(body)
+                .getJSONObject(JSON_KEY_DATA)
+                .getJSONObject(JSON_KEY_ATTRIBUTES)
+            val admitted = attrs.optBoolean(JSON_KEY_ADMITTED, true)
+            val rawReason = attrs.optString(JSON_KEY_REASON).ifEmpty { null }
```
0xnm:
```suggestion
            val rawReason = attrs.optString(JSON_KEY_REASON)
```

default fallback is empty string and `rawReason` is used only in the `parseReason`, both null and empty string yield to the same branch there.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    companion object {
+        private const val OPERATION_NAME_QUOTA = "profiling-quota-check"
+
+        internal const val LOG_HTTP_ERROR_TELEMETRY = "Profiling quota check returned HTTP %d"
+        internal const val LOG_NETWORK_ERROR = "Quota check network error: %s"
+        internal const val LOG_UNEXPECTED_ERROR = "Quota check unexpected error: %s"
+        internal const val LOG_QUOTA_DENIED = "Profiling quota denied: reason=%s"
+
+        internal const val QUOTA_URL_TEMPLATE = "https://quota.%s/api/v2/profiling/quota?session_id=%s"
```
0xnm: to confirm that this is deployed in all DCs

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    private val pendingFuture = AtomicReference<Future<QuotaResult>?>()
+    private val lastSessionId = AtomicReference<String?>(null)
+
+    @Volatile
+    var lastResult: QuotaResult? = null
+        private set
+
+    fun checkAsync(sessionId: String, datadogContext: DatadogContext) {
+        if (hasCustomEndpoint) {
```
0xnm: what is the use-case for that, who can set it?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+) {
+
+    private val pendingFuture = AtomicReference<Future<QuotaResult>?>()
+    private val lastSessionId = AtomicReference<String?>(null)
+
+    @Volatile
+    var lastResult: QuotaResult? = null
+        private set
+
+    fun checkAsync(sessionId: String, datadogContext: DatadogContext) {
```
0xnm: what will be the logic behind calling it? for example, we made a call, but got IO exception because there is no network. Obviously, we need to repeat the call, but a) when; b) with back-off?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+        if (hasCustomEndpoint) {
+            lastResult = QuotaResult.FAIL_OPEN
+            onResult(QuotaResult.FAIL_OPEN)
+            return
+        }
+        val previousId = lastSessionId.getAndSet(sessionId)
+        if (previousId == sessionId) return // same session: in-flight check (if any) is still valid
+        pendingFuture.getAndSet(null)?.cancel(true)
```
0xnm: nb: if it is `BackPressureExecutorService` or `LoggingScheduledThreadPoolExecutor`, `.cancel` will trigger error log entry

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: because the quota endpoint is `https://quota.$intakeHostName` and `intakeEndpoint` is `https://$intakeHostName`, so we need to remove the prefix to extract the `intakeHostName` here.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: This is mentioned by @simaoseica-dd before, when user uses custom endpoint, we shouldn't check quota API

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: With current logic, we make a call only when the session is renewed, and we don't retry if there is a network error, because the network error is a fail open according to the browser profiling logic, which can be changed later ofc.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: The plan is to use `SingleThreadScheduledExecutor` to avoid this.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
0xnm: I have a concern that browser logic regarding network errors can be be applied to mobile. The fair share of Browser SDK is on the desktop, there is always a connectivity there. Of course mobile browser can be used as well, but once again it should be a connectivity when page is loaded (and probably RUM session is started at this point).

Also Browser profiling requires a document policy sent by the server https://developer.mozilla.org/en-US/docs/Web/API/JS_Self-Profiling_API#security_requirements, this also implies having a connectivity.

satween (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: Yes I agree not to blindly follow browser SDK, we need a clear definition what is fail open for mobile for sure.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/ProfilingQuotaChecker.kt
```diff
@@ -0,0 +1,176 @@
...
+                    lastResult = result
+                    onResult(result)
+                }
+                result
+            }
+        )
+        if (future != null) {
+            pendingFuture.set(future)
+        } else {
+            lastSessionId.set(previousId)
```
abrooksv: The chance of this setting a stale value is pretty much nil since multiple invocations of `checkAsync` should not be running concurrently, but should this line be `lastSessionId.compareAndSet(sessionId, previousId)`?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
0xnm: alternatively we may simply expose `hostname` property in the `DatadogSite`, if such thing is needed somewhere else as well

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
0xnm: Why so? I'm not quite understand.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
0xnm: Do you propose to handle this in another PR?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
0xnm: Then we are losing the executor observability. Not sure if it is important or not though.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: Since there is a ongoing general discussion about Quota API, any decision we make right now on mobile side may affected by the final result of the discussion. so I purpose to merge as it and leave a TODO here.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/QuotaChecker.kt
ambushwork: custom endpoint check is removed.

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/quota/ProfilingQuotaChecker.kt
ambushwork: updated

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/internal/quota/QuotaCheckerTest.kt
```diff
@@ -0,0 +1,498 @@
...
+
+    @Test
+    fun `M invoke onResult callback W check completes`() {
+        // Given
+        val body = """{"data":{"attributes":{"admitted":true,"reason":"quota_ok"}}}"""
+        whenever(mockCall.execute()) doReturn makeResponse(200, body)
+
+        // When
+        testedChecker.checkAsync(fakeSessionId, fakeDatadogContext)
+        awaitCheck()
```
0xnm: why do we need to await anything, if there are no threads in this setup? same in other similar tests with this call

ambushwork: The test sets a real `newSingleThreadScheduledExecutor` so it needs to wait here, moving the `FkaeSameThreadExecutorService` in this [PR](https://github.com/DataDog/dd-sdk-android/pull/3536) so that we can use a fake executor here.

0xnm (approved):

PR #3500, opened 2026-06-03
PANA-5681: Add heatmaps feature flag

jonathanmos:
### What does this PR do?
Adds a flag to enable the heatmaps feature (disabled by default) - setHeatmapsEnabled

### Motivation

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
@@ -29,7 +29,8 @@ data class SessionReplayConfiguration internal constructor(
     internal val textAndInputPrivacy: TextAndInputPrivacy,
     internal val dynamicOptimizationEnabled: Boolean,
     internal val systemRequirementsConfiguration: SystemRequirementsConfiguration,
-    internal val internalCallback: SessionReplayInternalCallback
+    internal val internalCallback: SessionReplayInternalCallback,
+    internal val heatmapsEnabled: Boolean
 ) {
 
     /**
@@ -76,6 +77,7 @@ data class SessionReplayConfiguration internal constructor(
         private var dynamicOptimizationEnabled = true
         private var systemRequirementsConfiguration = SystemRequirementsConfiguration.NONE
         private var internalCallback: SessionReplayInternalCallback = NoOpSessionReplayInternalCallback()
+        private var heatmapsEnabled = false
 
         /**
          * Adds an extension support implementation. This is mostly used when you want to provide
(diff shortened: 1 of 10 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
```diff
@@ -197,7 +198,11 @@ internal class SessionReplayRecorder : OnWindowRefreshedCallback, Recorder {
                 recordedDataQueueHandler = recordedDataQueueHandler,
                 sdkCore = sdkCore,
                 dynamicOptimizationEnabled = dynamicOptimizationEnabled,
-                rumContextProvider = rumContextProvider
+                rumContextProvider = if (heatmapIdentifierRegistry != null) {
```
satween: Why `rumContextProvider` is being controlled by `heatmapIdentifierRegistry`. Looks not such clear, is there any mistake? Seems like `heatmapsEnabled` is false by default which means it's gonna lead to API change?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -214,6 +216,16 @@ data class SessionReplayConfiguration internal constructor(
             return this
         }
 
+        /**
+         * Enables or disables heatmap recording. When enabled, tap targets in the view tree are assigned
+         * stable identifiers that correlate with RUM action events, powering the heatmap feature.
+         * Disabled by default.
+         */
+        fun setHeatmapsEnabled(heatmapsEnabled: Boolean): Builder {
```
satween, line 224: worth to mark it as  experimental API ?

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
```diff
@@ -29,7 +29,8 @@ data class SessionReplayConfiguration internal constructor(
     internal val textAndInputPrivacy: TextAndInputPrivacy,
     internal val dynamicOptimizationEnabled: Boolean,
     internal val systemRequirementsConfiguration: SystemRequirementsConfiguration,
-    internal val internalCallback: SessionReplayInternalCallback
+    internal val internalCallback: SessionReplayInternalCallback,
+    internal val heatmapsEnabled: Boolean
```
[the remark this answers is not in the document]
jonathanmos: This is a theoretical concern for all the parameters in the configuration. Since these fields are internal and accessed only through the builder, I don't think we need to do anything here.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/SessionReplayRecorder.kt
jonathanmos: RumContextProvider is being used here only for heatmaps, however I agree that this isn't good architecturally to have it's state based around one feature, in case we want to use RCP for something else later on - I'll modify this

kikoveiga (dismissed):

0xnm (dismissed):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt
jonathanmos: I'd prefer not to do this since the feature flag is default false and we only have it in case the feature causes a performance impact. The plan is to eventually make this flag default true for everyone. If someone enables the feature and encounters an issue then disabling it is straightforward.

satween (dismissed):

kikoveiga (approved):

PR #3509, opened 2026-06-05
RUM-16562: Adds DDSketch implementation

abrooksv:
### What does this PR do?
Adds a DDSketch implementation that uses collapsing lowest bin storage and logarithmic indexes. This will be used to aggregate the latency metrics for APM client side stats

### Motivation
We will leverage this in a stats aggregator to build up the span latency metrics. They will eventually be serialized to their Protobuf formats and embedded into the okSummary summaryFields defined [here](https://github.com/DataDog/dd-sdk-android/blob/feature/client-side-stats/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/ClientStatsPayload.kt#L122-L123).

The original code was taken from https://github.com/DataDog/sketches-java but due to using Java 8 features, was vendored here then proceeded to be shrunk down to only the code we need to support client side stats. Code was then converted to Kotlin to make safe to run on Android and remove the Java 8 features and APIs.

The choice of CollapsingLowestDenseStore and LogarithmicMapping matches what the RFC calls for and which is in use today in the agent ([1](https://github.com/DataDog/datadog-agent/blob/cdba151f217787c515cb2d4cd7863b205a3ed255/pkg/trace/stats/client_stats_aggregator.go#L35)) ([2](https://github.com/DataDog/datadog-agent/blob/cdba151f217787c515cb2d4cd7863b205a3ed255/pkg/trace/stats/client_stats_aggregator.go#L493))

### Additional Notes
iOS PR: https://github.com/DataDog/dd-sdk-ios/pull/2914


detekt_custom_unsafe_calls.yml
```diff
@@ -197,10 +200,6 @@ datadog:
       - "kotlin.String.toLong():java.lang.NumberFormatException"
       - "kotlin.collections.MutableSet.first():java.util.NoSuchElementException"
       - "kotlin.String.format(kotlin.Array):java.util.IllegalFormatException"
-      - "kotlin.ByteArray.copyOf(kotlin.Int):java.lang.NegativeArraySizeException"
-      - "kotlin.ByteArray.copyOfRange(kotlin.Int, kotlin.Int):java.lang.IndexOutOfBoundsException,java.lang.IllegalArgumentException"
-      - "kotlin.ByteArray.get(kotlin.Int):java.lang.IndexOutOfBoundsException"
-      - "kotlin.Double.roundToLong():java.lang.IllegalArgumentException"
       # endregion
```
abrooksv, line 203: These were duplicated, they are also on lines 186-190

abrooksv:
@codex review this

0xnm (dismissed):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ddsketch/DDSketch.kt
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * Values are mapped to bins via [LogarithmicMapping] and stored in two [CollapsingLowestDenseStore]
+ * instances — one for positive values and one for negative values (stored as their positive
+ * opposites). When either store exceeds `maxNumBins`, the lowest bins are collapsed, trading
+ * precision at the low end of each store's range for bounded memory. Zero values are tracked
+ * separately and are not affected by collapsing.
+ *
+ * Not thread-safe.
+ */
+internal class DDSketch(
```
satween, line 26: Don't you think that you could add `Collection` interface here? 

It has almost all required methods (beside serialize) and will allow to use set of extension methods. 

For example you could use 

```assertThat(DDSketchInstance).isNotEmpty```

File: features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ddsketch/DDSketchSerializer.kt [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.trace.internal.ddsketch
+
+import java.io.ByteArrayOutputStream
+
+/**
+ * This class is used to perform protobuf serialization compliant with the official schema used to
+ * generate protobuf bindings (DDSketch.proto) but does not require the weight of the protobuf-java
```
satween, line 13: Do we have any protection from the situation when `DDSketch.proto` schema change? Would we be able to notice that? Maybe it's make sense to have some tests that compares serialization result from this code with the result produced via  `protobuf-java`  lib? In that case we would be required to put `protobuf-java `dependency only into the tests, but will be noticied when result become unequal.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ddsketch/LogarithmicMapping.kt
```diff
@@ -0,0 +1,84 @@
...
+        val validRequestedAccuracy = requestedRelativeAccuracy.coerceIn(
+            MIN_SAFE_RELATIVE_ACCURACY,
+            MAX_SAFE_RELATIVE_ACCURACY
+        )
+
+        /*
+         * Calculates the (minimal) base that needs to be used for the mapping to be relatively accurate
+         * with the provided relative accuracy.
+         */
+        gamma = (1 + validRequestedAccuracy) / (1 - validRequestedAccuracy)
```
satween, line 38: I haven't checked the formulas  and seems like `gamma` is safe, but for other variables - is it guaranteed that there will be no `ArithmeticException ` because of zero devision?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ddsketch/DDSketchSerializer.kt
abrooksv: Good idea, that is what I did for the MsgPack stuff and would make sense here as well

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ddsketch/LogarithmicMapping.kt
abrooksv: The gating of the `validRequestedAccuracy` to a valid [min,max] range protects us,

if it was 0, it would lead to a `gamma=1` which could throw here:
`multiplier = 1.0 / ln1p(1 - 1)  -> 1.0 / 0`

if it was 1, would break `maxIndexedValue` at `Double.MAX_VALUE / (0)`

BTW, in production we will be hard coding it to 0.01 to achieve a 1% accuracy to match spec

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ddsketch/DDSketchSerializer.kt
abrooksv: Added sketches-java and protobuf-java as test dependencies  to make real values for asserts

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ddsketch/DDSketch.kt
abrooksv: Made it implement Iterable so isEmpty and isNotEmpty work

0xnm (approved):

satween (approved):

PR #3523, opened 2026-06-09
RUM-16564: Create and register stub for client stats feature if enabled

abrooksv:
### What does this PR do?
The client stats feature is wired into the CoreTracer via the MetricsAggregator interface defined in the internal tracer module. This is the same interface that dd-trace-java uses to implement the aggregation of client stats before submitting them to the agent, so it is already wired into the correct spot in CoreTracer (before span sampling occurs).

A new config value was also added so we can override the intake location for the stats intake, since it needs to be different than the /spans endpoint that the TraceConfiguration already configures.

### Motivation
Sets up the SDK feature for receiving the spans for client stats aggregation from CoreTracer

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ClientStatsRequestFactory.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ClientStatsRequestFactory.kt
@@ -1,0 +1,63 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.trace.internal.net
+
+import com.datadog.android.api.context.DatadogContext
+import com.datadog.android.api.net.Request
+import com.datadog.android.api.net.RequestExecutionContext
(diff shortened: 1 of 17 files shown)


features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ClientStatsRequestFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    override fun create(
+        context: DatadogContext,
+        executionContext: RequestExecutionContext,
+        batchData: List<RawBatchEvent>,
+        batchMetadata: ByteArray?
+    ): Request {
+        val requestId = UUID.randomUUID().toString()
+
+        val baseUrl = customStatsEndpointUrl ?: (context.site.intakeEndpoint + "/api/v0.2/stats")
```
0xnm, line 29: is it indeed like that - `v0.2` or it should be `v2` here?

0xnm (approved):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/net/ClientStatsRequestFactory.kt
abrooksv: It is v0.2

https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/writer/stats.go#L31

Issue #3480, opened 2026-05-27
ANR in BroadcastReceiverSystemInfoProvider


ansman:
### Stack trace

```
#01  pc 0x0000000000057538  /apex/com.android.runtime/lib64/bionic/libc.so (je_arena_tcache_fill_small+72)
  #02  pc 0x0000000000074514  /apex/com.android.runtime/lib64/bionic/libc.so (je_tcache_alloc_small_hard+36)
  #03  pc 0x000000000006ae34  /apex/com.android.runtime/lib64/bionic/libc.so (je_malloc+852)
  #04  pc 0x000000000005180c  /apex/com.android.runtime/lib64/bionic/libc.so (malloc+64)
  #05  pc 0x000000000008793c  /data/app/~~L3Kkdw40aqToqBNPCLeDgQ==/com.openai.chatgpt-HRdBXByM5vfC7gyKPqcDLw==/split_config.arm64_v8a.apk
  #06  pc 0x0000000000034710  /data/app/~~L3Kkdw40aqToqBNPCLeDgQ==/com.openai.chatgpt-HRdBXByM5vfC7gyKPqcDLw==/split_config.arm64_v8a.apk
  at com.pairip.VMRunner.executeVM (Native method)
  at com.pairip.VMRunner.invoke (VMRunner.java:69)
  at com.datadog.android.core.internal.system.BroadcastReceiverSystemInfoProvider.onReceive (BroadcastReceiverSystemInfoProvider.kt:23)
  at android.app.LoadedApk$ReceiverDispatcher$Args.lambda$getRunnable$0 (LoadedApk.java:1943)
  at android.app.LoadedApk$ReceiverDispatcher$Args.$r8$lambda$gDuJqgxY6Zb-ifyeubKeivTLAwk (unavailable)
  at android.app.LoadedApk$ReceiverDispatcher$Args$$ExternalSyntheticLambda0.run (unavailable:2)
  at android.os.Handler.handleCallback (Handler.java:958)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loopOnce (Looper.java:257)
  at android.os.Looper.loop (Looper.java:368)
  at android.app.ActivityThread.main (ActivityThread.java:8839)
  at java.lang.reflect.Method.invoke (Native method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:572)
[...]

PR #3524, opened 2026-06-10
RUM-16123: Move broadcast receiver dispatch off the main thread

hamorillo:
### What does this PR do?

Introduces `BroadcastReceiverThread`, a dedicated `HandlerThread` that serves as the dispatch thread for the SDK's `BroadcastReceiver`s. The handler is passed to `Context.registerReceiver` via `ThreadSafeReceiver`, so `CONNECTIVITY_ACTION`, `ACTION_BATTERY_CHANGED`, and `ACTION_POWER_SAVE_MODE_CHANGED` callbacks are delivered on the background thread instead of the main thread.

### Motivation

Fixes ANRs reported in #3480. Root cause: `Context.registerReceiver(...)` without a `Handler` argument dispatches `onReceive` on the main thread by default. On devices protected by PairIP (Google Play Integrity Code Transparency), each `onReceive` call incurs non-deterministic overhead from `com.pairip.VMRunner.executeVM` on the main thread looper, causing ANRs. PR #3420 attempted to fix this by trampolining `onReceive` body through an executor, but PairIP wraps `onReceive` at entry on the main thread — before the trampoline could offload work — so the ANRs persisted. This PR reverts that executor-trampoline approach and takes a different one instead: passing a background `Handler` to `registerReceiver` so the framework never dispatches on the main thread at all. The `@Volatile` fields on `networkInfo` and `systemInfo` introduced in #3420 are retained for cross-thread visibility.


dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/thread/BroadcastReceiverThread.kt
```diff
@@ -0,0 +1,33 @@
...
+ * A wrapper around a dedicated [HandlerThread] used as the dispatch thread for
+ * the SDK's [android.content.BroadcastReceiver]s. Passing [handler] to
+ * [android.content.Context.registerReceiver] ensures each `onReceive` callback
+ * is delivered on this background thread instead of the main thread, avoiding
+ * ANRs caused by per-call overhead on protected builds (e.g. PairIP).
+ *
+ * Lifecycle: the underlying thread is started eagerly at construction time.
+ * Call [shutdown] when the SDK is torn down to release the thread.
+ */
+internal class BroadcastReceiverThread {
```
0xnm: This is a bit confusing, because this class is not a thread, despite having `Thread` in the name.

Maybe we can make it extend `HandlerThread`? `internal class BroadcastReceiverThread: HandlerThread("datadog-broadcast-receiver-thread")` without any properties and simply do `Handler` creation and wiring in `CoreFeature`? WDYT?

hamorillo: Yeah, make sense. 🙇‍♂️ 

[1f1103a](https://github.com/DataDog/dd-sdk-android/pull/3524/commits/1f1103ad170493a484104f953b7e28a90a8ae91f)

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProvider.kt
```diff
@@ -38,26 +37,29 @@ internal class BroadcastReceiverNetworkInfoProvider(
     // region BroadcastReceiver
 
     override fun onReceive(context: Context, intent: Intent?) {
-        executorService.executeSafe(HANDLE_INTENT_OPERATION_NAME, internalLogger) {
```
ambushwork, line 41: Question about how ANR can happen with previous approach, since we offload immediately the onReceive from main thread to `executorService`, the blocking task which causes ANR must happen before `onReceive` is called?

0xnm (approved):

ambushwork (approved):

aleksandr-gringauz (approved):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/net/info/BroadcastReceiverNetworkInfoProvider.kt
hamorillo: Correct. My guess is that PairIP instruments `onReceive` at the method entry point — before any code in the body runs, including the `executorService.executeSafe(...)` call. The overhead is in the PairIP VM's instrumentation of the method itself on the main thread looper, not in anything the body does. So by the time the executor would offload the work, the ANR-causing delay has already occurred on the main thread. Passing a background `Handler` to registerReceiver moves the entire dispatch — including the PairIP entry instrumentation — off the main thread, which is why I hope this approach will work.

PR #3526, opened 2026-06-10
fix(flags): dedupe exposures by latest assignment

leoromanovsky:
## Motivation

Android flags exposure deduplication currently treats every previously seen `targetingKey + flagName + allocationKey + variationKey` tuple as permanently sent. That suppresses valid exposure events when the same subject and flag cycles assignments, for example `A -> B -> A`, because the final `A` was seen earlier.

Related iOS PR: DataDog/dd-sdk-ios#2987 applies the same mobile exposure-cache contract.

## Changes

- Cache exposures by `targetingKey + flagName`.
- Store the latest `allocationKey + variationKey` as the cache value.
- Emit an exposure when the subject/flag has no cached value or when the latest assignment differs from the cached value.
- Keep a bounded Android `LruCache`, but make it entry-count based with a 5,000-entry limit instead of using approximate object-size accounting.
- Add a comment documenting the expected high-water mark of two subjects and 2,500 flags each; normal flag keys are typically tens of characters, so count-based sizing is easier to reason about.
- Add a regression test proving `A -> B -> A` writes three exposure events.
- Update the existing mixed scenario to expect a cycle-back exposure.

## Decisions

- Kept the change internal to the processor with no public API changes.
- Preserved an LRU bound for apps with many flags moving through a client process, while sizing it directly around the expected two-subject, 2,500-flag high-water mark.
- Avoided approximate byte accounting in `sizeOf`; AndroidX `LruCache` defaults to one unit per entry when `sizeOf` is not overridden.
- Matched the iOS PR's count-based mobile policy while using Android's platform cache primitive.
- Kept synchronization around cache check/update so duplicate suppression remains atomic.
- Validation: focused regression test failed before the implementation, then passed after the fix. `JAVA_HOME=/opt/homebrew/Cellar/openjdk@17/17.0.19/libexec/openjdk.jdk/Contents/Home ANDROID_HOME=/opt/homebrew/share/android-commandlinetools ./gradlew :features:dd-sdk-android-flags:testDebugUnitTest` passed.


typotter (dismissed):

mikewoo-dd:
Please disregard comment from PR shepherd above. We're doing a dry run test and would not make any change to the PR.

0xnm:
@leoromanovsky @mikewoo-dd It seems this PR is stale. Is it still needed?

leoromanovsky:
> @leoromanovsky @mikewoo-dd It seems this PR is stale. Is it still needed?

@0xnm Yes, it is. It was just a bit lower priority, as I wasn't chasing down reviews as aggressively as I would otherwise. I'll work to get it closed out this week.

typotter (approved):

0xnm (approved):

PR #3529, opened 2026-06-10
RUM-16565: Implement client stats aggregation

abrooksv:
### What does this PR do?
Add a new StatsConcentrator that receives DDSpans and aggregates them into ClientStatsBucket containing their latency distributions, error counts, and hit counts. Spans are placed into different buckets based on the timestamps, and then a set of aggregation keys are used to aggregate similar spans within that bucket.

Once buckets are flushed, they will get written to `StatsWriter`. In a future PR, this interface will eventually forward the buckets into the `ClientStatsFeature`'s batch by bridging the gap between `StatsConcentrator` and the `StorageBackedFeature` / `EventBatchWriter`

### Motivation

Adds in the middle chunk of work in the client side stats system. StatsConcentrator class will get registered into CoreTracer (see #3523) via a not-yet written adapter to adapt it into the MetricsAggregation interface. 

### Additional Notes

iOS counter part: https://github.com/DataDog/dd-sdk-ios/pull/2871


features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
```diff
@@ -0,0 +1,279 @@
...
+            if (buckets.isNotEmpty()) {
+                statsWriter.write(buckets)
+            }
+        }
+    }
+
+    private fun aggregate(s: SpanSnapshot) {
+        val bucketKey = alignTimestamp(bucketSizeNs, s.startTime + s.durationNs).coerceAtLeast(oldestTs)
+        val key = buildAggregationKey(s)
+        @Suppress("UnsafeThirdPartyFunctionCall") // getOrPut can't throw
```
aleksandr-gringauz: You can put the `getOrPut` function into `detekt_custom_safe_calls.yml`, so that the linter knows that it can't throw. This way you won't have to suppress the warning.

abrooksv: I did add the first one, but the 2nd getOrPut() gets triggered on the nested class and wasn't sure if it was worth adding to the generic list since it'll never get fired by another use case 

```
kotlin.collections.MutableMap.getOrPut(com.datadog.android.trace.interna(...)]
```

I guess putting it in the file and letting RUM-16478 clean it up with a wildcard may be the best long term plan though

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
```diff
@@ -0,0 +1,279 @@
...
+
+    // Coalescing flags: at most one flush task lives in the executor queue at any time.
+    // forcePending is set unconditionally so a force flush is never lost even if a scheduled
+    // flush is already queued and will pick it up.
+    private val flushPending = AtomicBoolean(false)
+    private val forcePending = AtomicBoolean(false)
+
+    fun record(trace: List<DDSpan>) {
+        val metricsFeature = sdkCore.getFeature(Feature.TRACING_CLIENT_STATS_FEATURE_NAME) ?: return
+        metricsFeature.withContext { datadogContext ->
```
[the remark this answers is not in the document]
abrooksv: An early force report is fine to leave some stuff in the queue since the next flush will grab it. 

On SDK shutdown, we may lose metrics before they get aggregated and written to the batch file, I dont see any pattern of draining queues or the write contexts before shutdown so this seems to be acceptable behavior

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
```diff
@@ -0,0 +1,807 @@
...
+
+@Extensions(
+    ExtendWith(MockitoExtension::class),
+    ExtendWith(ForgeExtension::class)
+)
+@MockitoSettings(strictness = Strictness.LENIENT)
+@ForgeConfiguration(Configurator::class)
+internal class StatsConcentratorTest {
+
+    @Mock lateinit var mockSdkCore: FeatureSdkCore
```
0xnm: very nit, it is a bit different from the style across the codebase:

```suggestion
    @Mock
    lateinit var mockSdkCore: FeatureSdkCore
```

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
```diff
@@ -0,0 +1,807 @@
...
+    private lateinit var executorService: ExecutorService
+    private lateinit var testedConcentrator: StatsConcentrator
+
+    // Fixed parameters for deterministic timestamp math
+    private val fakeBucketSizeNs = 10_000_000_000L // 10 s in ns
+    private val fakeBufferLen = 2
+
+    @BeforeEach
+    fun setUp() {
+        executorService = Executors.newSingleThreadExecutor()
```
0xnm: do we actually need a real executor for the test? maybe we can use "same thread" executor? this will eliminate all await statements

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
```diff
@@ -0,0 +1,807 @@
...
+        val flushNow = (fakeBufferLen + 6).toLong() * fakeBucketSizeNs
+        testedConcentrator.scheduleFlush(now = flushNow, flushAll = false)
+        awaitFlush()
+
+        // When: second flush after the first has already drained the bucket
+        testedConcentrator.scheduleFlush(now = flushNow, flushAll = false)
+        awaitFlush()
+
+        // Then: write called exactly once — second flush found empty buckets
+        verify(mockStatsWriter, times(1)).write(any())
```
0xnm: should we actually capture and assert invocation argument instead of `any` or it is not necessary?

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
```diff
@@ -0,0 +1,807 @@
...
+    /** `now` value that puts any reasonable span bucket well past the flush cutoff. */
+    private fun farFuture(): Long = 100 * fakeBucketSizeNs
+
+    /**
+     * Drains the executor queue by submitting a no-op after the last test action and waiting for
+     * it to complete. Since the executor is single-threaded, all previously submitted tasks
+     * (aggregate and flush) will have run by the time this returns.
+     */
+    private fun awaitFlush() {
+        executorService.submit {}.get(FLUSH_TIMEOUT_MS, TimeUnit.MILLISECONDS)
```
0xnm: if we use "same thread" executor, we probably don't need this, unless we need a dedicated thread in some tests

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
aleksandr-gringauz: agreed

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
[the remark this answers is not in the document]
aleksandr-gringauz: We have [this](https://github.com/DataDog/dd-sdk-android/blob/453d0404f8f7a9a65ccc4787f575759b5ed865f0/dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt#L403) flush that currently is used only in tests and also in [these](https://github.com/DataDog/rum-sdk-integration-test-framework/tree/main) tests.

I would say this isn't critical to fix now, if we encounter problems in tests, we can think about the fix here.

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
abrooksv: The flush conflation tests will need it but I think I can instead do 

```
        thread { 
            flush()
            flush()
        }
```

to keep the test

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
abrooksv: Once #3536 is merged, I'll merge develop into this branch and switch to it

aleksandr-gringauz (dismissed):

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
abrooksv: Had to use a single thread executor, but only scoped to these 2 tests

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
abrooksv: Added asserts and dropped all write(any()) usages

0xnm (approved):

PR #3532, opened 2026-06-11
RUM-16637: Report Profiling system package version and client clock drift

0xnm:
### What does this PR do?

Add additional profiling telemetry fields:

* Version of the system profiling package. It may be update as a part of mainline updates and not tied to the OS version. This comes in the `profiling_config` section, it doesn't depend on the profiling session run.
* Client clock drift: Perfetto is using system clock, but in RUM we use NTP-corrected clock. This can create misalignment between the two time-wise.

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/anr/AnrProfilingTriggerRegistrar.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/anr/AnrProfilingTriggerRegistrar.kt
@@ -138,6 +138,7 @@ internal class AnrProfilingTriggerRegistrar(
                 errorMessage = result.errorMessage,
                 fileSize = fileSize,
                 callbackDelayMs = callbackDelayMs,
+                clientClockDriftMs = timeProvider.getServerOffsetMillis(),
                 droppedAsStale = droppedAsStale
             )
         )
--- a/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
+++ b/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
@@ -33,11 +33,11 @@ import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.Extensions
 import org.mockito.Mock
-import org.mockito.Mockito.mock
 import org.mockito.junit.jupiter.MockitoExtension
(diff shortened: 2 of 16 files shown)


jonathanmos (dismissed):
minor: is there a test verifying that timeProvider.delegate gets set to sdkCore.timeProvider in onInitialize?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/anr/AnrProfilingTriggerRegistrar.kt
```diff
@@ -50,6 +52,8 @@ internal class AnrProfilingTriggerRegistrar(
     @Volatile
     private var listener: AnrListener? = null
 
+    private val profilingPackageLongVersionCode = AtomicLong(PROFILING_MANAGER_VERSION_NOT_INITIALIZED)
```
ambushwork: Question: is it necessary to store this value in two different places? Maybe we can get it during the feature initialization and pass down to `ProfilingTelemetry`?

ambushwork (dismissed):

jonathanmos (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/anr/AnrProfilingTriggerRegistrar.kt
0xnm: Fair point. Although we need only `Context` to read it, so not necessarily at the Feature registration, but since it is sent to the telemetry, it is okay to read it there only.

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
```diff
@@ -38,6 +38,7 @@ import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.junit.jupiter.MockitoSettings
 import org.mockito.kotlin.any
 import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
```
ambushwork, line 40:
```suggestion

```

ambushwork (dismissed):

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingTest.kt
0xnm: actually it is better to remove another one. fixed.

ambushwork (approved):

jonathanmos (approved):

PR #3535, opened 2026-06-12
RUM-16728: Report TTFD to profiling

0xnm:
### What does this PR do?

Similar to the TTID reporting for profiling, we can report TTFD as well.

The difference is that TTFD is optional, user needs to call RUM API explicitly.

That means that TTFD won't stop any profiling session and is recorded only if there is continuous profiling running (in this case application launch profiling session is extended).

### DIFF
--- a/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
+++ b/features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
@@ -880,6 +880,32 @@ internal class ProfilingFeatureTest {
     }
 
     @Test
+    fun `M not stop Profiling W receive TTFD vital event {continuous disabled}`(
+        @Forgery fakeTtfdVital: ProfilerEvent.RumVitalEvent
+    ) {
+        // Given
+        testedFeature = ProfilingFeature(
+            mockSdkCore,
+            ProfilingConfiguration(
+                customEndpointUrl = null,
+                applicationLaunchSampleRate = 100f,
+                continuousSampleRate = 0f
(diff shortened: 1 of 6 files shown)


hamorillo (approved):
🚀

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
```diff
@@ -879,6 +879,32 @@ internal class ProfilingFeatureTest {
         verify(mockProfiler, never()).stop(fakeInstanceName)
     }
 
+    @Test
+    fun `M not stop Profiling W receive TTFD vital event {continuous disabled}`(
```
ambushwork, line 883: nit: This test verifies that TTFD doesn't stop the app launch profiling, but we also need another test to verify if Profiling Feature receives it, it accumulates in the pending events.

ambushwork (approved):

features/dd-sdk-android-profiling/src/test/kotlin/com/datadog/android/profiling/ProfilingFeatureTest.kt
0xnm: Isn't this already covered by `M accumulate vital event W onReceive {continuous active window open}`?

ambushwork: you are right, just the outdated naming of `private lateinit var fakeTTID: ProfilerEvent.RumVitalEvent` makes me think the test covers only TTID. LGTM then.

PR #3538, opened 2026-06-12
RUM-16394 Pt.3: Wire Quota Checker with Profiling Feature

ambushwork:
### What does this PR do?

Connects the ProfilingQuotaChecker to the Profiling Feature so profiling is gated on the backend quota decision, and surfaces the quota outcome on RUM events.

### Motivation

- Profiling feature
  - On each new RUM session, `ProfilingFeature` fires an async quota check (createOkHttpCallFactory + dedicated single-thread executor, 5s call timeout).
  - Launch profiling holds its buffered result until the quota decision arrives (or the timeout fallback fails open); on DENIED the launch event is dropped, otherwise it's written. Continuous profiling skips its active window (stays in cooldown) when the decision is DENIED.
  - The denied reason is propagated into the profiling feature context (PROFILING_QUOTA_REASON) and cleared when allowed; onStop tears the checker/executor down cleanly.
- RUM
  - RumViewScope (Error / Long Task) and the app-launch vital now report profiling.status = STOPPED with quota_reason when profiling was denied by quota, and RUNNING otherwise. Reading profiling state was simplified to go straight through the feature context.

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
@@ -19,6 +19,7 @@ import com.datadog.android.internal.FeatureContextKeys
 import com.datadog.android.internal.sampling.DeterministicSampling
 import com.datadog.android.internal.sampling.SessionSamplingIdProvider
 import com.datadog.android.internal.time.TimeProvider
+import com.datadog.android.profiling.internal.quota.QuotaResult
 import java.util.Locale
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.ScheduledFuture
@@ -66,6 +67,9 @@ internal class ContinuousProfilingScheduler(
 
     @Volatile
     private var remainingCooldownMs: Long = 0L
+
+    @Volatile
+    internal var lastQuotaResult: QuotaResult? = null
 
     private val schedulerExecutor: ScheduledExecutorService = profiler.scheduledExecutorService
 
@@ -230,6 +234,18 @@ internal class ContinuousProfilingScheduler(
     private fun scheduleNextCycle() {
         val activeMs = jitter(CONTINUOUS_WINDOW_DURATION_MS)
         if (currentSessionSampled) {
+            val quota = lastQuotaResult
+            if (quota == null || quota.decision == QuotaResult.Decision.DENIED) {
+                logToUser {
+                    if (quota == null) {
+                        LOG_QUOTA_PENDING
+                    } else {
+                        LOG_QUOTA_DENIED.format(Locale.US, quota.reason.rawValue)
+                    }
+                }
+                scheduleCooldown(activeMs)
+                return
+            }
             onActiveWindowStarted()
             isActive = true
             state = State.ACTIVE
@@ -386,5 +402,9 @@ internal class ContinuousProfilingScheduler(
             "Grace period expired; profiler stopped."
         internal const val LOG_RUM_SESSION_DECISION =
             "RUM session decision: rumSessionSampleRate=%s, currentSessionSampled=%s."
+        internal const val LOG_QUOTA_DENIED =
+            "Continuous profiling skipped: quota denied (reason=%s)."
+        internal const val LOG_QUOTA_PENDING =
+            "Continuous profiling skipped: awaiting quota decision."
     }
 }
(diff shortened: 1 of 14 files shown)


ambushwork:
@codex review

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ContinuousProfilingScheduler.kt
```diff
@@ -230,6 +234,11 @@ internal class ContinuousProfilingScheduler(
     private fun scheduleNextCycle() {
         val activeMs = jitter(CONTINUOUS_WINDOW_DURATION_MS)
         if (currentSessionSampled) {
+            if (lastQuotaResult.decision == QuotaResult.Decision.DENIED) {
+                logToUser { LOG_QUOTA_DENIED.format(Locale.US, lastQuotaResult.reason.rawValue) }
+                scheduleCooldown(activeMs)
+                return
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/startup/RumSessionScopeStartupManager.kt
```diff
@@ -260,7 +269,7 @@ internal class RumSessionScopeStartupManagerImpl(
             durationNs = durationNs,
             appLaunchMetric = VitalAppLaunchEvent.AppLaunchMetric.TTFD,
             scenario = scenario,
-            profilingStatus = datadogContext.getProfilingStatus()
+            profilingStatus = datadogContext.getProfilingStatus(rumContext.sessionId)
```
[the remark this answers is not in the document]
ambushwork: fixed

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
```diff
@@ -95,6 +110,18 @@ internal class ProfilingFeature(
         }
         dataWriter = createDataWriter(sdkCore)
 
+        val quotaCallFactory = sdkCore.createOkHttpCallFactory {
+            callTimeout(QUOTA_CHECK_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+        }
+        val qExecutor = sdkCore.createSingleThreadExecutorService(QUOTA_EXECUTOR_CONTEXT)
```
0xnm, line 116: maybe it is worth reusing existing `uploadExecutorService` executor or executor for the profiling? The latter is used to run only profiling submission tasks, which are short and quick, so it has plenty of room for another scheduling.

ambushwork: The issue is not profiling submission block quota checker, but oppositely the network request in QuotaCheck can block the profiling submission, which might disorder the continuous profiling cycle or callback of ProfilingManager.

0xnm: > the network request in QuotaCheck can block the profiling submission

I guess this is the case for the Profiling executor, but what about `uploadExecutorService` then?

ambushwork: yes, using `uploadExecutorService` doesn't have this problem, but I don't think `ProfilingFeature` has access to `uploadExecutorService` in the `CoreFeature`

0xnm: It is not exposed via API, but it is not a problem with exposing it somehow. Or we can use `InternalSdkCore.getPersistenceExecutorService` for that (hacky, but works).

Realistically we create a thread just to run (once per session, ideally) a quick task, and most of the time it will be idling. SDK has already too much threads and we must pack work across existing threads.

We can keep the current approach, but we need to revisit it later.

0xnm (approved):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingFeature.kt
0xnm: In the sample app I found 14 threads belonging to Datadog SDK, roughly 30% don't have any work (measuring the first few seconds of app start).

image

ambushwork: I just let Claude investigate the source code for the same purpose, here is the result:

> With all features enabled, roughly 19–24 threads per SDK instance:
> - Core: 4 (shared, justified)
> - RUM: ~6 (pipeline, anr, vital, + one per active tracking strategy)
> - Profiling: 2
> - Session Replay: 3–12 (resource pool scales 1→10)
> - Flags: 2
> - Trace: 3 (some are process-wide singletons, not per-instance)

Smells like an Android Guild task, I will create a ticket

PR #3539, opened 2026-06-12
RUM-16551: Fix hidden API reflection noise in Session Replay drawable mappers

kikoveiga:
### What does this PR do?

- Gates `PorterDuffColorFilter.mColor` reflection to `API < 28`, where the field is actually accessible. ON API 28+ the field is classified as `max-target-o`, meaning it's blocked and was causing some noise in the logcat.
- Removes unused `mButtonDrawableField` in `CheckableCompoundButtonMapper` that was also causing noise.

### DIFF
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
@@ -21,6 +21,7 @@ import android.graphics.drawable.RippleDrawable
 import android.graphics.drawable.ShapeDrawable
 import android.graphics.drawable.StateListDrawable
 import android.graphics.drawable.VectorDrawable
+import android.os.Build
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.utils.ALPHA_SHIFT_ANDROID
 import com.datadog.android.sessionreplay.utils.DrawableToColorMapper
@@ -98,7 +99,7 @@ internal open class AndroidMDrawableToColorMapper(
      * Resolves the color from a [LayerDrawable].
      * @param drawable the color drawable
      * @param internalLogger the internalLogger to report warnings
-     * @param predicate a predicate to filter which ayers should be taken into account (default: accept all layers)
+     * @param predicate a predicate to filter which layers should be taken into account (default: accept all layers)
      * @return the color to map to or null if not applicable
      */
     protected open fun resolveLayerDrawable(
@@ -228,16 +229,22 @@ internal open class AndroidMDrawableToColorMapper(
             null
         }
 
+        // PorterDuffColorFilter.mColor is blocked (max-target-o) on API 28+ and unused on API 29+
+        // (AndroidQDrawableToColorMapper overrides resolveGradientDrawable). Skip reflection on P+.
         @Suppress("PrivateAPI", "SwallowedException", "TooGenericExceptionCaught")
-        internal val mColorField = try {
-            PorterDuffColorFilter::class.java.getDeclaredField("mColor").apply {
-                this.isAccessible = true
+        internal val mColorField = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
+            try {
+                PorterDuffColorFilter::class.java.getDeclaredField("mColor").apply {
+                    this.isAccessible = true
+                }
+            } catch (_: NoSuchFieldException) {
+                null
+            } catch (_: SecurityException) {
+                null
+            } catch (_: NullPointerException) {
+                null
             }
-        } catch (e: NoSuchFieldException) {
-            null
-        } catch (e: SecurityException) {
-            null
-        } catch (e: NullPointerException) {
+        } else {
             null
         }
     }
--- a/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
+++ b/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
@@ -22,6 +22,7 @@ import android.graphics.drawable.ShapeDrawable
 import android.graphics.drawable.StateListDrawable
 import android.graphics.drawable.VectorDrawable
 import android.os.Build
+import androidx.annotation.RequiresApi
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.sessionreplay.utils.ALPHA_SHIFT_ANDROID
 import com.datadog.android.sessionreplay.utils.DrawableToColorMapper
@@ -32,6 +33,7 @@ import com.datadog.android.sessionreplay.utils.MAX_ALPHA_VALUE
  * Drawable utility object needed in the Session Replay Wireframe Mappers.
  * This class is meant for internal usage so please use it carefully as it might change in time.
  */
+@Suppress("TooManyFunctions")
 internal open class AndroidMDrawableToColorMapper(
     private val extensionMappers: List<DrawableToColorMapper> = emptyList()
 ) : DrawableToColorMapper {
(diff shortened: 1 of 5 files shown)


features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
```diff
@@ -228,16 +229,22 @@ internal open class AndroidMDrawableToColorMapper(
             null
         }
 
+        // PorterDuffColorFilter.mColor is blocked (max-target-o) on API 28+ and unused on API 29+
+        // (AndroidQDrawableToColorMapper overrides resolveGradientDrawable). Skip reflection on P+.
         @Suppress("PrivateAPI", "SwallowedException", "TooGenericExceptionCaught")
-        internal val mColorField = try {
-            PorterDuffColorFilter::class.java.getDeclaredField("mColor").apply {
-                this.isAccessible = true
+        internal val mColorField = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
```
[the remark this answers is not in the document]
kikoveiga: This scenario requires `targetSdk <= 26` on API 28 hardware, which is basically impossible nowadays.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
jonathanmos: This is unused so we can probably remove it

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
jonathanmos, line 150: ai detected this issue and it appears to be correct:

`fillPaintField` has the same hidden-API problem as `mColorField`, but isn't gated.

GradientDrawable.mFillPaint has @UnsupportedAppUsage, which means it generates a logcat warning on every app start on any modern Android version. Validated this by running the sample app and seeing:

`Accessing hidden field Landroid/graphics/drawable/GradientDrawable;->mFillPaint:Landroid/graphics/Paint; (unsupported, reflection, allowed)`

The fix would be to gate fillPaintField the same way you gated mColorField: skip reflection on API 24+, where GradientDrawable.getColor() (public API) provides the solid fill color without reflection. 

AndroidQDrawableToColorMapper.resolveGradientDrawable would also need updating since it calls fillPaintField on API 29+.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/CheckableCompoundButtonMapper.kt
kikoveiga: Removed, thanks!

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
kikoveiga: Done, `fillPaintField` is now gated to `SDK_INT < N` (same pattern as `mColorField`). `AndroidQDrawableToColorMapper` no longer uses reflection at all, delegating to `resolveGradientFillColor()` which calls the public `GradientDrawable.getColor()`.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
```diff
@@ -214,30 +229,41 @@ internal open class AndroidMDrawableToColorMapper(
...
```
hamorillo, line 260: > This scenario requires targetSdk <= 26 on API 28 hardware

Is this really impossible? Why can't I just run an app with `targetSdkVersion = 26` on an API 28 device?

Doesn't max-target-o mean the field is blocked from `targetSdk > 26`, not from API 28? Shouldn't the gate be `Build.VERSION_CODES.O_MR1` instead of `Build.VERSION_CODES.P`?

kikoveiga: - It's not impossible, but it's a scenario that doesn't happen today in production. We would need to refactor to pass the `Context` and check `targetSdkVersion` for this specific corner-case...

- Yes you're right, but looks like the actual enforcement only starts at API 28, so reflection on API 27 is still safe.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
```diff
@@ -124,15 +126,27 @@ internal open class AndroidMDrawableToColorMapper(
      * @return the color to map to or null if not applicable
      */
     protected open fun resolveGradientDrawable(drawable: GradientDrawable, internalLogger: InternalLogger): Int? {
-        @Suppress("SwallowedException")
+        return if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+            resolveGradientDrawableReflection(drawable, internalLogger)
+        } else {
+            resolveGradientFillColor(drawable)?.let { resolvedColor ->
```
hamorillo: minor: Just wondering if we should add some coverage for this new branch https://github.com/DataDog/dd-sdk-android/pull/3539#issuecomment-4693258225

hamorillo (dismissed):
LGTM!

jonathanmos (dismissed):

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapper.kt
kikoveiga: Yes, added 3 tests in `AndroidMDrawableToColorMapperTest`. Thanks!

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapperTest.kt
```diff
@@ -281,4 +283,64 @@ open class AndroidMDrawableToColorMapperTest {
...
+    private class TestableMMapper : AndroidMDrawableToColorMapper() {
+        var fakeResolvedColor: Int? = null
+
+        override fun resolveGradientDrawable(drawable: GradientDrawable, internalLogger: InternalLogger): Int? {
+            return resolveGradientFillColor(drawable)?.let { resolvedColor ->
+                val colorAlpha = (resolvedColor ushr ALPHA_SHIFT_ANDROID) and MAX_ALPHA_VALUE
+                val fillAlpha = (colorAlpha * drawable.alpha) / MAX_ALPHA_VALUE
+                if (fillAlpha == 0) null else mergeColorAndAlpha(resolvedColor, fillAlpha)
+            }
+        }
```
hamorillo: Thanks for adding the tests @kikoveiga.

I would say we should avoid this because we are not really testing the production code; we are basically testing this code, so a change in the production code that introduces a bug won't be detected by our tests.

Something like this [patch](https://github.com/user-attachments/files/28990380/fix-test-reliability.patch) can work better. Let me know what you think.

ambushwork (dismissed):

features/dd-sdk-android-session-replay/src/test/kotlin/com/datadog/android/sessionreplay/internal/recorder/mapper/AndroidMDrawableToColorMapperTest.kt
kikoveiga: Thanks a lot for that patch, it makes sense. Added it to the code.

hamorillo (approved):
Thanks for the changes. 🚀

PR #3555, opened 2026-06-18
RUM-16325 `timeseries`: Update RUM JSON schemas and regenerate models

satween:
### What does this PR do?

Updates the RUM JSON schemas to the latest definitions and regenerates the corresponding Kotlin models and API surface. Also improves the json-schema code generator so `OneOfPrimitiveOptionGenerator` shares the `knownTypes` registry with `FileGenerator`, avoiding duplicate type resolution.

### DIFF
--- a/features/dd-sdk-android-rum/api/dd-sdk-android-rum.api
+++ b/features/dd-sdk-android-rum/api/dd-sdk-android-rum.api
@@ -5204,7 +5204,7 @@ public final class com/datadog/android/rum/model/TimeseriesMemoryEvent$DataPoint
 	public fun equals (Ljava/lang/Object;)Z
 	public static final fun fromJson (Ljava/lang/String;)Lcom/datadog/android/rum/model/TimeseriesMemoryEvent$DataPoint;
 	public static final fun fromJsonObject (Lcom/google/gson/JsonObject;)Lcom/datadog/android/rum/model/TimeseriesMemoryEvent$DataPoint;
-	public final fun getMemoryMax ()Ljava/lang/Number;
+	public final fun getMemoryFootprint ()Ljava/lang/Number;
 	public final fun getMemoryPercent ()Ljava/lang/Number;
 	public fun hashCode ()I
 	public final fun toJson ()Lcom/google/gson/JsonElement;
--- a/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/FileGenerator.kt
+++ b/buildSrc/src/main/kotlin/com/datadog/gradle/plugin/jsonschema/generator/FileGenerator.kt
@@ -23,7 +23,7 @@ class FileGenerator(
 
     private val classGenerator = ClassGenerator(packageName, knownTypes)
     private val enumGenerator = EnumClassGenerator(packageName, knownTypes)
-    private val oneOfPrimitiveOptionGenerator = OneOfPrimitiveOptionGenerator(packageName)
+    private val oneOfPrimitiveOptionGenerator = OneOfPrimitiveOptionGenerator(packageName, knownTypes)
 
     private val multiClassGenerator = MultiClassGenerator(
         classGenerator = classGenerator,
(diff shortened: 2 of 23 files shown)


features/dd-sdk-android-rum/src/main/json/rum/_stream-schema.json
0xnm, line 1: we need to remove all the model definitions not actually used by the SDK, e.g. streams, duration vitals, etc.

0xnm (changes requested):

buildSrc/src/test/kotlin/com/datadog/gradle/plugin/jsonschema/generator/OneOfPrimitiveOptionGeneratorTest.kt
```diff
@@ -0,0 +1,107 @@
...
+ * (e.g. `oneOf [string, integer]` named `Path`, as produced by the shared `_graphql-schema.json`
+ * referenced by both `ResourceEvent` and `ErrorEvent`), the option sub-classes of the *second*
+ * generated model used to extend the *first* model's nested sealed class
+ * (`SecondEvent.String : FirstEvent.Path()`), which does not compile.
+ *
+ * This happened because [OneOfPrimitiveOptionGenerator] held its own private `knownTypes` set that
+ * was never cleared between files, unlike every other generator which shares (and clears per file)
+ * the [FileGenerator] set. The fix wires the shared set into the generator; this test guards it.
+ */
+class OneOfPrimitiveOptionGeneratorTest {
```
aleksandr-gringauz: This json schema generator has an established testing approach that consists of `JsonSchemaReaderTest`, `FileGeneratorTest`, `ModelValidationTest` and corresponding schema files and kotlin files.

It might be better to add a new schema and models and test them using this approach instead of creating a dedicated test.

aleksandr-gringauz:
> OneOfPrimitiveOptionGenerator shares the knownTypes registry with FileGenerator, avoiding duplicate type resolution

Yes, we already have this `knownTypes` mechanism and it makes sense to add it to the `OneOfPrimitiveTypeOptionGenerator`.

No call to action, just a thought. 

I have once concern about this `knownTypes` approach in general (not only in oneOf(primitives) case) - the name of the resulting class depends on the order of traversal. For example in `ViewEvent` we have 2 structurally identical fields: `context` and `featureFlags`. In `ViewEvent` they have a type `Context`. It is weird that `featureFlags` field has type `Context` that has nothing to do with it`. Also if we at some point introduce a third one the name can change for already existing ones breaking the backward compatibility.

buildSrc/src/test/kotlin/com/datadog/gradle/plugin/jsonschema/generator/OneOfPrimitiveOptionGeneratorTest.kt
satween: thanks for pointing that, refactored to the common approach

features/dd-sdk-android-rum/src/main/json/rum/_stream-schema.json
satween: I left only timeseries related update (`timeseries-memory-schema.json`)

features/dd-sdk-android-rum/api/dd-sdk-android-rum.api
```diff
@@ -4902,6 +4902,3089 @@ public final class com/datadog/android/rum/model/ResourceEvent$Worker$Companion
...
+	public final fun toJson ()Lcom/google/gson/JsonElement;
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class com/datadog/android/rum/model/RumTransitionEvent$Viewport$Companion {
+	public final fun fromJson (Ljava/lang/String;)Lcom/datadog/android/rum/model/RumTransitionEvent$Viewport;
+	public final fun fromJsonObject (Lcom/google/gson/JsonObject;)Lcom/datadog/android/rum/model/RumTransitionEvent$Viewport;
+}
+
+public final class com/datadog/android/rum/model/RumViewUpdateEvent {
```
aleksandr-gringauz: Looks like the apiSurface and .api files for rum contain much more changes then the changes in the json schema. You only changed the `timeseries-memory-schema.json`. And here I see ViewUpdate, that shouldn't be here.

satween: yes, polluted from last schema classes, fixed

aleksandr-gringauz (approved):

PR #3557, opened 2026-06-18
RUM-16478: `UnsafeThirdPartyFunctionCall` improvements

satween:
### What does this PR do?
* Adds generics support for the `UnsafeThirdPartyFunctionCall` detekt rule. `*` means any non-null type, `?` means any nullable type.
* Adds configuration self-checks: verifies that all records in each config are unique.
* Adds configuration self-checks: prevents the same method from being added to controversial (safe and unsafe) configs at the same time.
* `UnsafeThirdPartyFunctionCall` is now divided into three subclasses: `CodeParser` for Kotlin code parsing, and `DetektConfigValidator` and `DetektConfigParser` for YAML config parsing and validation.
* `detekt_custom_unsafe_calls.yml` and `detekt_custom_safe_calls.yml` have been cleaned up according to the new rules.

### Motivation
An approach to reduce time to green build for each ticket. Fewer retries means less time and lower resource consumption.


tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/UnsafeThirdPartyFunctionCall.kt
```diff
@@ -62,85 +63,84 @@ class UnsafeThirdPartyFunctionCall(
...
-        if (uncaught.isEmpty()) {
-            return
-        }
+        if (uncaught.isEmpty()) return
 
-        val msg = "Calling $call can throw the following exceptions: ${exceptions.joinToString()}."
-        reportUnsafeCall(expression, msg)
+        reportUnsafeCall(
+            expression = expression,
+            message = "Calling $call can throw the following exceptions: ${exceptions.joinToString()}."
```
hamorillo: nit: Maybe here we can report only the uncaught exceptions.

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/rule/thirdparty/DetektConfigValidator.kt
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+        }
+    }
+
+    private companion object {
+        private val semanticUnsafeCalls = listOf(
+            "java.lang.Thread.interrupt():java.lang.SecurityException",
+            "kotlin.Array.first(kotlin.Function1):java.util.NoSuchElementException"
+        )
```
hamorillo: ❓ What's the goal of this list? Why not to keep those in `the unsafe_call.yml`?

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/rule/thirdparty/SignatureRule.kt
```diff
@@ -0,0 +1,88 @@
...
+    val regex: Regex,
+    private val paramSlots: List<String>,
+    private val wildcardSlots: List<WildcardSlot>,
+    val exceptions: List<String> = emptyList()
+) {
+
+    private val methodPart: String = source.substringBefore('(')
+    internal val overlapKey: OverlapKey = OverlapKey(methodPart, paramSlots.size)
+
+    /** True when this rule has no wildcard slots and can be matched with a plain map lookup. */
```
hamorillo: nit: I would say the comment is saying the opposite of what the value means.

hamorillo (commented):
Looks good to me. Just leaving a question before approving.

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/rule/thirdparty/DetektConfigParser.kt
```diff
@@ -0,0 +1,121 @@
...
+    fun parseSafeCalls(entries: List<String>): List<SignatureRule> = entries.map(::parseEntry)
+
+    fun parseThrowingCalls(entries: List<String>): List<SignatureRule> = entries.map(::parseThrowingEntry)
+
+    internal companion object {
+        private const val COLON_SEPARATOR = ':'
+        private const val EXCEPTIONS_SEPARATOR = ","
+        private const val NON_NULL_WILDCARD = "*"
+        private const val NULLABLE_WILDCARD = "?"
+        private const val PARAMS_SEPARATOR = ", "
```
jonathanmos: the separator requires an explicit space and this could cause silently ignored rules if the space is omitted. Should we not make the separator without a space and run `trim()` on the strings to avoid this?

satween: Thanks for spotting - fixed

detekt_custom_unsafe_calls.yml
abrooksv: This can be collapsed

abrooksv: These can be collapsed

abrooksv: These can be collapsed

abrooksv: Same

abrooksv, line 212: SuspendFunction1 one is same function as the `kotlinx.coroutines.flow.FlowCollector` one so maybe it can be removed?

abrooksv, line 251: Can this become (String?, ?)?

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/UnsafeThirdPartyFunctionCall.kt
satween: fixed

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/rule/thirdparty/DetektConfigValidator.kt
satween: debug leftover, removed

detekt_custom_unsafe_calls.yml
satween: this will require to support such types mapping, will keep as it is for now - to keep code bit simplier

satween: Nope. This will make the rule checking too broad and complex.

abrooksv: Ah I thought they were all the same behavior for the overloads so a null wildcard for 2nd param was safe to do

satween: yes, it this case. But then it opens some strage scenarios. like if nullable type (non generic one) covers non nullable, then some methods like `ConcurrentHashMap.put(String?, String)` starts cover `ConcurrentHashMap.put(String, String)` but in reality nullable key - will raise `NullPointerException`. And vice versa. It's possible to resolve such issues, but it will make code way more complicated

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/rule/thirdparty/SignatureRule.kt
```diff
@@ -0,0 +1,133 @@
...
+    internal val overlapKey: OverlapKey = OverlapKey(methodPart, paramSlots.size)
+    internal val memberReference: MemberReference = MemberReference(
+        className = methodPart.substringBeforeLast('.', missingDelimiterValue = ""),
+        memberName = methodPart.substringAfterLast('.'),
+        parameterCount = paramSlots.size
+    )
+
+    /** True when this rule can be matched with a plain map lookup. */
+    internal val canUseExactLookup: Boolean =
+        wildcardSlots.isEmpty() && (!nullableConcreteMatchesNonNullable || paramSlots.none { it.isConcreteNullable() })
```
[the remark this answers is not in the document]
satween: Good catch. Exact-lookup entries were keyed by the raw `source` string, so a config entry with non-canonical spacing (e.g. no space after a comma, like the `getStringSet(...)` entry) never matched resolved call signatures, which are always rendered with `", "`. I added a `canonicalSignature` (parameters re-joined with `", "`) and use it both as the exact-lookup map key and in `SignatureRule.matches()`, so any spacing in the YAML now normalizes to the canonical form. Added `SignatureRuleTest` cases covering safe and throwing calls with randomized spacing after commas.

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/rule/thirdparty/SignatureRule.kt
```diff
@@ -0,0 +1,133 @@
...
+            b: String,
+            aNullableMatchesNonNullable: Boolean,
+            bNullableMatchesNonNullable: Boolean
+        ): Boolean {
+            val aWild = a == NON_NULL_WILDCARD || a == NULLABLE_WILDCARD
+            val bWild = b == NON_NULL_WILDCARD || b == NULLABLE_WILDCARD
+            return when {
+                aWild && bWild -> true
+                aWild -> b.matchesWildcard(wildcard = a)
+                bWild -> a.matchesWildcard(wildcard = b)
```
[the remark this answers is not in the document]
satween: Good catch. A safe nullable-concrete slot (`nullableConcreteMatchesNonNullable = true`) also matches the non-nullable resolved call, so it genuinely overlaps a throwing `*` wildcard on that shared non-nullable form, but `matchesWildcard` only inspected the literal `?` suffix and missed it. The concrete slot's flag is now threaded into `matchesWildcard`, so the wildcard intersection accounts for it. Throwing nullable-concrete slots stay exact (flag `false`), so no false overlaps are introduced. Added `SignatureRuleTest` + `DetektConfigValidatorTest` cases for both directions and expanded the `parseEntry` KDoc explaining the flag.

detekt_custom_unsafe_calls.yml
```diff
@@ -110,11 +110,9 @@ datadog:
       # endregion
       # region Java Concurrency
       - "java.lang.Thread.constructor(java.lang.Runnable?, kotlin.String?):java.lang.NullPointerException,java.lang.SecurityException,java.lang.IllegalArgumentException"
-      - "java.lang.Thread.constructor(java.lang.Runnable, kotlin.String):java.lang.NullPointerException,java.lang.SecurityException,java.lang.IllegalArgumentException"
       - "java.lang.Thread.getAllStackTraces():java.lang.SecurityException"
       - "java.lang.Thread.interrupt():java.lang.SecurityException"
       - "java.lang.Thread.sleep(kotlin.Long):java.lang.IllegalArgumentException,java.lang.InterruptedException"
-      - "java.util.concurrent.BlockingQueue.drainTo(kotlin.collections.MutableCollection):java.lang.UnsupportedOperationException,java.lang.ClassCastException,java.lang.NullPointerException,java.lang.IllegalArgumentException"
       - "java.util.concurrent.BlockingQueue.drainTo(kotlin.collections.MutableCollection?):java.lang.UnsupportedOperationException,java.lang.ClassCastException,java.lang.NullPointerException,java.lang.IllegalArgumentException"
```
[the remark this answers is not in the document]
satween: Restored the non-null signature: `BlockingQueue.drainTo(kotlin.collections.MutableCollection)` (as it was on `develop`). Since this is a throwing entry, a nullable-concrete slot is kept exact (`nullableConcreteMatchesNonNullable = false`), so `drainTo(kotlin.collections.MutableCollection?)` would only match nullable-resolved calls and miss the real non-null callers (`CoreFeature`, `BackPressureExecutorService`), which with `treatUnknownFunctionAsThrowing: true` would then be flagged. The resolved call signature is built from the callee's formal parameter type (not the argument), so the non-null entry matches every `drainTo` call regardless of argument nullability, and its exception list already includes `NullPointerException`.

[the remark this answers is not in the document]
satween: This is a false positive. `BlockingQueue.drainTo` is a pure Java method, so its `Collection<? super E>` parameter is a platform type and the resolver renders it as **nullable**: the call resolves to `java.util.concurrent.BlockingQueue.drainTo(kotlin.collections.MutableCollection?)` (verified via the rule under real type resolution). The nullable throwing record therefore matches the real calls (`BackPressureExecutorService.drainTo`, `CoreFeature`); a non-null `drainTo(kotlin.collections.MutableCollection)` record would be dead and those guarded calls would be reported as unknown under `treatUnknownFunctionAsThrowing=true`. Added regression tests in `UnsafeThirdPartyFunctionCallTest` covering both directions (nullable record matches / guarded call is clean; non-null record misses → reported as unknown), so no code change is needed here.

PR #3558, opened 2026-06-18
RUM-16056 Add `@HotMethod` annotation and HotMethodIllegalCall detekt rule

satween:
### What does this PR do?

Introduces a `@HotMethod` source annotation to mark frame/touch hot-path methods and a `HotMethodIllegalCall` detekt rule that forbids heap allocations and O(N) collection operations inside them. Adds a dedicated `detekt_expensive_methods.yml` config to activate the rule. Annotates existing hot-path methods in RUM, Session Replay, and Compose; pre-allocates objects where needed to satisfy the rule.

### Motivation

Methods called on every frame or touch event (JankStats callbacks, `dispatchTouchEvent`, `onDraw`) are a common source of GC pressure and UI jank. Making them machine-verifiable via a detekt rule prevents regressions from creeping back in during normal development.

### Additional Notes

The `@Suppress("HotMethodIllegalCall")` annotations on two sites (one `SlowFrameRecord` allocation and one `TargetNode` allocation) document intentional exceptions: both allocations are bounded by actual jank/tap events rather than frame rate, so suppression is intentional and annotated with an explanation comment.

Ref: RUM-16056

### DIFF
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/lint/HotMethod.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/lint/HotMethod.kt
@@ -14,7 +14,22 @@ package com.datadog.android.internal.lint
  * reminder to reviewers.
  *
  * @property message A short description of why this method is hot (e.g. "called per frame by JankStats").
+ * @property exclude Check names or specific function/constructor names to skip for this site only.
+ *   Built-in category strings: `"constructor"`, `"anonymous-object"`, `"lambda"`,
+ *   `"string-template"`, `"factory"`, `"collection-ops"`.
+ *   You may also pass a bare method name (e.g. `"forEach"`) or a qualified name
+ *   (e.g. `"kotlin.collections.List.forEach"`) to exclude just that call.
+ *   Prefer a narrow per-name exclusion over a broad category exclusion.
  */
 @Retention(AnnotationRetention.SOURCE)
 @Target(AnnotationTarget.FUNCTION)
-annotation class HotMethod(val message: String)
+annotation class HotMethod(val message: String, val exclude: Array<String> = []) {
+    companion object {
+        const val CHECK_CONSTRUCTOR = "constructor"
+        const val CHECK_ANONYMOUS_OBJECT = "anonymous-object"
+        const val CHECK_LAMBDA = "lambda"
+        const val CHECK_STRING_TEMPLATE = "string-template"
+        const val CHECK_FACTORY = "factory"
+        const val CHECK_COLLECTION_OPS = "collection-ops"
+    }
+}
--- a/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/lint/HotMethod.kt
+++ b/dd-sdk-android-internal/src/main/java/com/datadog/android/internal/lint/HotMethod.kt
@@ -1,0 +1,20 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+package com.datadog.android.internal.lint
+
+/**
+ * Marks a method as a hot path — one that can be invoked on every frame or touch event.
+ *
+ * Implementations must avoid heap allocations, blocking operations, and any work
+ * that introduces GC pressure. The IDE/Lint will highlight annotated methods as a
+ * reminder to reviewers.
+ *
+ * @property message A short description of why this method is hot (e.g. "called per frame by JankStats").
+ */
+@Retention(AnnotationRetention.SOURCE)
+@Target(AnnotationTarget.FUNCTION)
+annotation class HotMethod(val message: String)
--- a/detekt_expensive_methods.yml
+++ b/detekt_expensive_methods.yml
@@ -1,0 +1,201 @@
+build:
+  maxIssues: 0
+
+config:
+  validation: true
+  warningsAsErrors: true
+
+comments:
+  active: true
+  AbsentOrWrongFileLicense:
+    active: true
+    licenseTemplateFile: 'license.template'
+    licenseTemplateIsRegex: false
+    excludes:
+      # Gradle build outputs and generated artifacts.
+      - '**/build/**'
+      # Auto-generated Kotlin model fixtures produced by the buildSrc JSON-to-Kotlin code generator.
+      - '**/buildSrc/src/test/kotlin/com/example/model/**'
+      # Fixture files for the NoOpFactory annotation processor tests, read as text data rather than compiled.
+      - '**/tools/noopfactory/src/test/resources/**'
+      # Vendored test files ported from the upstream dd-trace-java library.
+      - '**/features/dd-sdk-android-trace-internal/src/test/kotlin/com/datadog/trace/**'
+
+complexity:
+  active: false
+
+coroutines:
+  active: false
+
+empty-blocks:
+  active: false
+
+exceptions:
+  active: false
+
+naming:
+  active: false
+
+potential-bugs:
+  active: false
+
+style:
+  active: false
+
+datadog:
+  active: true
+  excludes:
+    - '**/build/**'
+    - '**/test/**'
+    - '**/testDebug/**'
+    - '**/testRelease/**'
+    - '**/androidTest/**'
+    - '**/testFixtures/**'
+    - '**/buildSrc/**'
+    - '**/*.kts'
+    - '**/instrumented/**'
+    - '**/reliability/**'
+    - '**/sample/**'
+    - '**/tools/**'
+  HotMethodIllegalCall:
+    active: true
+    # Lambda literals passed to functions in `allowedInlineFunctions` are inlined by the compiler
+    # and do not allocate — they are excluded from the lambda-allocation check.
+    # Add any project-specific `inline` functions here.
+    allowedInlineFunctions:
+      # Kotlin scope functions
+      - 'let'
+      - 'also'
+      - 'apply'
+      - 'run'
+      - 'with'
+      - 'takeIf'
+      - 'takeUnless'
+      # Standard utilities
+      - 'synchronized'
+      - 'use'
+      - 'repeat'
+      - 'measureNanoTime'
+      - 'measureTimeMillis'
+      # Kotlin stdlib collection operations — ALL are inline, lambdas are not allocated
+      - 'filter'
+      - 'filterNot'
+      - 'filterIsInstance'
+      - 'filterNotNull'
+      - 'map'
+      - 'mapNotNull'
+      - 'mapIndexed'
+      - 'flatMap'
+      - 'flatten'
+      - 'forEach'
+      - 'forEachIndexed'
+      - 'any'
+      - 'all'
+      - 'none'
+      - 'count'
+      - 'find'
+      - 'findLast'
+      - 'first'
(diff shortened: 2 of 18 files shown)


tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/HotMethodIllegalCall.kt
```diff
@@ -0,0 +1,240 @@
...
+    private val forbiddenFactoryFunctionsSet: Set<String> by lazy { forbiddenFactoryFunctions.toSet() }
+    private val allowedInlineFunctionsSet: Set<String> by lazy { allowedInlineFunctions.toSet() }
+
+    // Each named function pushes whether it is @HotMethod; lambdas inherit the enclosing context.
+    private val functionDepthStack = ArrayDeque<Boolean>()
+
+    private val insideHotMethod: Boolean
+        get() = functionDepthStack.lastOrNull() == true
+
+    override fun visitNamedFunction(function: KtNamedFunction) {
```
ambushwork, line 94: Question, this detekt rule checks only the hot method body but does not recursively apply the call tree, meaning that the rule can be escaped by extracting the expensive call into another function, is this expected?

```
@HotMethod
fun foo(){
   bar()
}

fun bar(){
   // expensive calls here
}
```

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/HotMethodIllegalCall.kt
```diff
@@ -0,0 +1,240 @@
...
+            if (calleeName.isNotEmpty() && calleeName[0].isUpperCase()) {
+                report(
+                    CodeSmell(
+                        issue,
+                        Entity.from(expression),
+                        "Constructor call `$calleeName()` inside a @HotMethod allocates a new object on every " +
+                            "invocation. Move the allocation to a field or pre-allocate outside the hot path."
+                    )
+                )
+                return
```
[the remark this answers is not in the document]
ambushwork: valid point

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/lint/HotMethod.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Implementations must avoid heap allocations, blocking operations, and any work
+ * that introduces GC pressure. The IDE/Lint will highlight annotated methods as a
+ * reminder to reviewers.
+ *
+ * @property message A short description of why this method is hot (e.g. "called per frame by JankStats").
+ */
+@Retention(AnnotationRetention.SOURCE)
+@Target(AnnotationTarget.FUNCTION)
+annotation class HotMethod(val message: String)
```
ambushwork: is it possible to add a ignore list in the params? I can imagine in some case we need to ignore some rules, instead of Suppress the whole `HotMethod` rule, we can bypass only a few but apply the rest?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/WindowCallbackWrapper.kt
```diff
@@ -60,8 +62,8 @@ internal class WindowCallbackWrapper(
         } else {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                { "Received null MotionEvent" }
+                ERROR_LOG_TARGETS,
+                MESSAGE_NULL_MOTION_EVENT
```
ambushwork, line 66: Question, if such message requires a variable from the @HotMethod, and must be constructed such as:

internalLogger.log(
    InternalLogger.Level.ERROR,
    ERROR_LOG_TARGETS,
    messageBuilder = { MESSAGE_TIMESTAMP.format(Locale.US, event.time) }
)

It's expected that we throw the error for this?

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/HotMethodIllegalCall.kt
```diff
@@ -0,0 +1,240 @@
...
+                "Anonymous object created inside a @HotMethod allocates a new instance on every " +
+                    "invocation. Extract it to a field or a top-level object."
+            )
+        )
+    }
+
+    override fun visitLambdaExpression(expression: KtLambdaExpression) {
+        super.visitLambdaExpression(expression)
+        if (!insideHotMethod) return
+        if (expression.isInlinedLambda()) return
```
ambushwork, line 154: This only excludes the case where the lambda is inlined, but the lambda which captures nothing can also be ignored since it will be optimized during the Kotlin compilation, and will not be allocated everytime.

detekt_expensive_methods.yml
```diff
@@ -0,0 +1,201 @@
...
+    - '**/instrumented/**'
+    - '**/reliability/**'
+    - '**/sample/**'
+    - '**/tools/**'
+  HotMethodIllegalCall:
+    active: true
+    # Lambda literals passed to functions in `allowedInlineFunctions` are inlined by the compiler
+    # and do not allocate — they are excluded from the lambda-allocation check.
+    # Add any project-specific `inline` functions here.
+    allowedInlineFunctions:
```
aleksandr-gringauz, line 65: Are the functions listed in `allowedInlineFunctions` allowed, so the linter will not complain? 

Asking because IIUC for example `filter` is both in `allowedInlineFunctions` and in `forbiddenCalls`.

Or am I missing the point?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/lint/HotMethod.kt
satween: Added `val exclude: Array<String> = []` to `@HotMethod`. It accepts both category strings (`"constructor"`, `"anonymous-object"`, `"lambda"`, `"string-template"`, `"factory"`, `"collection-ops"`) and bare/qualified method names (e.g. `"forEach"` or `"kotlin.collections.List.forEach"`). The rule reads the annotation's PSI at visit time and skips the corresponding checks site-by-site.

detekt_expensive_methods.yml
satween: Good catch — the naming is confusing. `allowedInlineFunctions` is a **lambda-allocation exemption**, not a call exemption. Kotlin `inline` functions paste the lambda body directly into the call site, so no `Function` object is ever created; flagging the lambda would be a false positive. `forbiddenCalls` flags the *call itself* (e.g. `filter` allocates a new `List`). A function like `filter` legitimately belongs in both: the call is expensive (→ `forbiddenCalls`), but its lambda is free (→ `allowedInlineFunctions`). Without the second entry you'd get two findings for `list.filter { … }` — the collection allocation and a spurious lambda-allocation warning. Added an expanded comment block to `detekt_expensive_methods.yml` explaining this with a concrete example and a rule-of-thumb.

features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/callback/RecorderWindowCallback.kt
```diff
@@ -211,9 +216,10 @@ internal class RecorderWindowCallback(
 
         // every 10 frames we flush the buffer
         internal val FLUSH_BUFFER_THRESHOLD_NS: Long = MOTION_UPDATE_DELAY_THRESHOLD_NS * 10
-        internal const val MOTION_EVENT_WAS_NULL_ERROR_MESSAGE =
-            "RecorderWindowCallback: intercepted null motion event"
         internal const val FAIL_TO_PROCESS_MOTION_EVENT_ERROR_MESSAGE =
             "RecorderWindowCallback: wrapped callback failed to handle the motion event"
+        private val MESSAGE_MOTION_EVENT_WAS_NULL: () -> String = {
+            "RecorderWindowCallback: intercepted null motion event"
+        }
```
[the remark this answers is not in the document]
satween: Restored `MOTION_EVENT_WAS_NULL_ERROR_MESSAGE` as an `internal const val` in the companion, and the private `MESSAGE_MOTION_EVENT_WAS_NULL` lambda now references it.

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/HotMethodIllegalCall.kt
satween: Very good point! The problem is that porper nested control flow analysis will require too much effort. 
For instance it could be some interface method being called: 
```kotlin

 interface Foo {
   fun bar() 
 }

 class FooComplex : Foo {
     override fun bar() {
          // mine bitcoin here
     }
}

@HotMethod(message="....") 
fun onDraw(){
   foo.bar()
}

```

To validate all such cases you will have to parse entire codebase, understand what possible implementations of the method are exist in the project, then validate all the code inside it. This basically will require to implement simple JVM iside detekt rule. 

`@HotMethod` - mostly just a hint for detekt. If developer marked some method as a hot one - detekt will check it's source code for the obvious set of low perfomant actions.

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/WindowCallbackWrapper.kt
satween: Rule will throw an error if there is an allocation of the object within the `@HotMehtod` as it adds pressure to the GC forcing it to cleaning up all of the allocated objects.

 Lambdas that catches outer context will not compiled into static method and then, should be highlighted. 
 However it's developer's responsibility to decide how often such scenario could happened. For instance a lambda that is being created to log the exception message  inside `@HotMethod` could be suppressed (as an exception will crash the app anyway): 
 ```kotlin 
 
@HotMethod (message = "...")
fun foo (){

    try {
      // ....
    } catch (e: Exception){
        @Supress("HotMethod")
        internalLogger.log(
            InternalLogger.Level.ERROR,
            ERROR_LOG_TARGETS,
            messageBuilder = { MESSAGE_TIMESTAMP.format(Locale.US, event.time) }
         )
    }

}

 ```

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/HotMethodIllegalCall.kt
satween: Yes, valid point! Supported

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/HotMethodIllegalCall.kt
ambushwork: Fair enough, just to confirm the scope of `@HotMethod`, the argument sounds reasonable.

tools/detekt/src/main/kotlin/com/datadog/tools/detekt/rules/sdk/HotMethodIllegalCall.kt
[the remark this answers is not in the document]
satween: Removed the early return so the rule descends into constructor arguments.

PR #3562, opened 2026-06-22
RUM-16696: Attach profiling context to the View events

ambushwork:
### What does this PR do?

Attach profiling context to the View events as we did for RUM Errors and Vitals

### Motivation

RUM-16696

### DIFF
--- a/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
+++ b/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
@@ -1394,7 +1394,8 @@ internal open class RumViewScope(
                         sessionSampleRate = sampleRate,
                         sessionReplaySampleRate = resolveSessionReplaySampleRate(datadogContext),
                         traceSampleRate = resolveTraceSampleRate(datadogContext)
-                    )
+                    ),
+                    profiling = resolveViewProfilingStatus(datadogContext)
                 ),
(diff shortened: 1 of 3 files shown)


features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt
```diff
@@ -1714,6 +1715,19 @@ internal open class RumViewScope(
         }
     }
 
+    private fun resolveViewProfilingStatus(datadogContext: DatadogContext): ViewEvent.Profiling? {
```
aleksandr-gringauz, line 1718: Just to better understand.

What profiler run this status corresponds to? I mean there can be application launch profiling, continious profiling, 1 RUM session can have many of them (correct me if I am wrong here).

Is it the most recent one?

ambushwork: `ProfilingStatus` represents if there is a profiling session on going - more specifically speaking - if a profiling session which is manually requested by SDK through `ProfilingManager.requestProfiling` is running, it means app launch profiling and continuous profiling will update this status, (they don't overlap since `ProfilingManager` doesn't support multiple profile sessions through request), but it doesn't include the trigger based profiling for ANR, which is a background profiler can run in parallel with our profiling session.

aleksandr-gringauz (approved):

abrooksv (approved):

hamorillo (approved):

PR #3564, opened 2026-06-22
RUM-16727: Integrate the client side stat systems into ClientStatsFeature

abrooksv:
### What does this PR do?
Combines the StatsCombinatator, flush scheduling, and upload process together inside of  ClientStatsFeature.

StatsCombinator also includes changes for how it manages to get time to encapsulate the `now` logic and also to offset the generated stat bucket by the server time offset

### Motivation
This combines the current individual parts into a single system allowing client side stats to work end to end

### Additional Notes

### DIFF
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
@@ -7,33 +7,119 @@
 package com.datadog.android.trace.internal
 
 import android.content.Context
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.feature.Feature
+import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.api.feature.StorageBackedFeature
 import com.datadog.android.api.storage.FeatureStorageConfiguration
+import com.datadog.android.trace.event.SpanEventMapper
+import com.datadog.android.trace.internal.domain.event.CoreTracerSpanToSpanEventMapper
+import com.datadog.android.trace.internal.domain.event.SpanEventMapperWrapper
+import com.datadog.android.trace.internal.domain.metrics.BatchStatsWriter
+import com.datadog.android.trace.internal.domain.metrics.ClientStatsAdapter
+import com.datadog.android.trace.internal.domain.metrics.StatsConcentrator
 import com.datadog.android.trace.internal.net.ClientStatsRequestFactory
-import com.datadog.trace.common.metrics.NoOpMetricsAggregator
+import com.datadog.trace.api.Config
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.RejectedExecutionException
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration.Companion.seconds
 
 internal class ClientStatsFeature(
-    customEndpointUrl: String?
+    private val sdkCore: FeatureSdkCore,
+    customEndpointUrl: String?,
+    spanEventMapper: SpanEventMapper,
+    networkInfoEnabled: Boolean
 ) : StorageBackedFeature {
     override val name = Feature.TRACING_CLIENT_STATS_FEATURE_NAME
 
     override val storageConfiguration = FeatureStorageConfiguration(
-        // 512 KB
-        maxItemSize = 512L * 1024,
-        maxItemsPerBatch = 4000,
+        // 15 MB
+        maxItemSize = 15L * 1024 * 1024,
+        // Batching is handled by the flush/aggregation mechanism, only one ClientStatsPayload per upload
+        maxItemsPerBatch = 1,
         // 15 MB
         maxBatchSize = 15L * 1024 * 1024,
         // 18 hours
         oldBatchThreshold = 18L * 60L * 60L * 1000L
     )
 
-    // TODO RUM-16565 Implement stats aggregator
-    internal val aggregator = NoOpMetricsAggregator()
+    private val timeProvider = sdkCore.timeProvider
+    private val statAggregator: ExecutorService by lazy {
+        sdkCore.createSingleThreadExecutorService("client-side-stats-aggregator")
+    }
+    private val flushScheduler: ScheduledExecutorService by lazy {
+        sdkCore.createScheduledExecutorService("client-side-flush-scheduler")
+    }
+    private val statsConcentrator by lazy {
+        // runtimeId is static and encapsulated in Tracer core's Config. There is no way to get it except via the root
+        // span's tags or this Config object
+        val runtimeID = Config.get().runtimeId
+
+        StatsConcentrator(
+            sdkCore,
+            ddSpanToSpanEventMapper = CoreTracerSpanToSpanEventMapper(networkInfoEnabled),
+            eventMapper = SpanEventMapperWrapper(spanEventMapper, sdkCore.internalLogger),
+            statAggregator,
+            BatchStatsWriter(sdkCore, runtimeID),
+            timeProvider
+        )
+    }
+
+    internal val aggregator by lazy {
+        ClientStatsAdapter(statsConcentrator)
+    }
 
     override val requestFactory = ClientStatsRequestFactory(customEndpointUrl)
 
-    override fun onInitialize(appContext: Context) {}
+    override fun onInitialize(appContext: Context) {
+        startPeriodicFlush()
+    }
 
-    override fun onStop() {}
+    private fun startPeriodicFlush() {
+        try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // exception caught below
+            flushScheduler.scheduleWithFixedDelay(
+                { triggerFlush(flushAll = false) },
+                FLUSH_INTERVAL.inWholeMilliseconds,
+                FLUSH_INTERVAL.inWholeMilliseconds,
+                TimeUnit.MILLISECONDS
+            )
+        } catch (e: RejectedExecutionException) {
+            sdkCore.internalLogger.log(
+                InternalLogger.Level.WARN,
+                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
+                { "Failed to schedule stats flush" },
+                e
+            )
+        }
+    }
+
+    override fun onStop() {
+        flushScheduler.shutdownNow()
+
+        triggerFlush(flushAll = true)
+
+        statAggregator.shutdown()
+        try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // InterruptedException is caught and handled
+            if (!statAggregator.awaitTermination(DRAIN_WAIT_SECONDS, TimeUnit.SECONDS)) {
+                statAggregator.shutdownNow()
+            }
+        } catch (_: InterruptedException) {
+            statAggregator.shutdownNow()
+            @Suppress("UnsafeThirdPartyFunctionCall") // safe - SecurityException not thrown in Android
+            Thread.currentThread().interrupt()
+        }
+    }
+
+    private fun triggerFlush(flushAll: Boolean) {
+        statsConcentrator.scheduleFlush(flushAll)
+    }
+
+    private companion object {
+        private const val DRAIN_WAIT_SECONDS = 10L
+        private val FLUSH_INTERVAL = 10.seconds
+    }
 }
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
@@ -35,10 +35,9 @@ internal class ClientStatsFeature(
     override val name = Feature.TRACING_CLIENT_STATS_FEATURE_NAME
 
     override val storageConfiguration = FeatureStorageConfiguration(
-        // 15 MB
-        maxItemSize = 15L * 1024 * 1024,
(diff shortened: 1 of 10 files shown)


abrooksv:
@codex review this

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/BatchStatsWriter.kt
```diff
@@ -0,0 +1,45 @@
...
+
+internal class BatchStatsWriter(
+    private val sdkCore: FeatureSdkCore,
+    private val runtimeID: String
+) : StatsWriter {
+    private val sequenceNumber = AtomicLong(0)
+
+    override fun write(statBuckets: List<ClientStatsBucket>) {
+        sdkCore.getFeature(Feature.TRACING_CLIENT_STATS_FEATURE_NAME)
+            ?.withWriteContext { datadogContext, writeScope ->
```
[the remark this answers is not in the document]
abrooksv: This is a good call out, RUM-16955

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/BatchStatsWriter.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.atomic.AtomicLong
+
+internal class BatchStatsWriter(
+    private val sdkCore: FeatureSdkCore,
+    private val runtimeID: String
+) : StatsWriter {
+    private val sequenceNumber = AtomicLong(0)
+
+    override fun write(statBuckets: List<ClientStatsBucket>) {
+        sdkCore.getFeature(Feature.TRACING_CLIENT_STATS_FEATURE_NAME)
```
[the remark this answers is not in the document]
abrooksv: This seems to be the pattern for the SDK and `EvaluationsFeature` has the same issue since it also does flush()->write()

The fix for this is probably SDK wide and outside the scope of this PR

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -7,33 +7,120 @@
...
+            }
+        } catch (_: InterruptedException) {
+            statAggregator.shutdownNow()
+            @Suppress("UnsafeThirdPartyFunctionCall") // safe - SecurityException not thrown in Android
+            Thread.currentThread().interrupt()
+        }
+    }
+
+    private fun triggerFlush(flushAll: Boolean) {
+        val now = timeProvider.getDeviceTimestampMillis().milliseconds.inWholeNanoseconds
```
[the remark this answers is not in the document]
abrooksv: Fix pushed

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -7,33 +7,119 @@
...
         oldBatchThreshold = 18L * 60L * 60L * 1000L
     )
 
-    // TODO RUM-16565 Implement stats aggregator
-    internal val aggregator = NoOpMetricsAggregator()
+    private val timeProvider = sdkCore.timeProvider
+    private val statAggregator: ExecutorService by lazy {
+        sdkCore.createSingleThreadExecutorService("client-side-stats-aggregator")
+    }
+    private val flushScheduler: ScheduledExecutorService by lazy {
```
aleksandr-gringauz: Why do we need a dedicated executor (and a thread) just for flush? Why can't we reuse `statAggregator`?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -7,33 +7,119 @@
...
 ) : StorageBackedFeature {
     override val name = Feature.TRACING_CLIENT_STATS_FEATURE_NAME
 
     override val storageConfiguration = FeatureStorageConfiguration(
-        // 512 KB
-        maxItemSize = 512L * 1024,
-        maxItemsPerBatch = 4000,
+        // 15 MB
+        maxItemSize = 15L * 1024 * 1024,
+        // Batching is handled by the flush/aggregation mechanism, only one ClientStatsPayload per upload
```
aleksandr-gringauz: Just curious. What is the reasoning behind this? Can we upload a batch with multiple `ClientStatsPayload`s?

No strong opinion for now, just want to understand the reasoning behind this choice.

abrooksv: Opps, There can only be 1 StatsPayload in the batch, but many ClientStatsPayloads inside of that.

I was moving the StatsPayload up, but I only half finished that change. Will revert this back and do that in another PR with the RequestFactory changes

abrooksv: The reason to move it up is because if a batch has too many ClientStatsPayloads, we have to set a boolean on *each* batch that indicates we split the batch up. The batching system does not allow us to mutate the batches with such data so we need to do it ourselves

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/ClientStatsAdapter.kt
```diff
@@ -0,0 +1,42 @@
...
+        // NO - OP
+        return null
+    }
+
+    override fun publish(trace: List<CoreSpan<*>>?): Boolean {
+        if (trace == null) return false
+
+        statsConcentrator.record(trace.filterIsInstance<DDSpan>())
+
+        return false
```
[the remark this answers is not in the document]
abrooksv: This is what the java tracer does, but I am not sure there is much gain. 

The backend is not going to aggregate stats for it and I am unsure how safe it would be to override the sampling decision

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
abrooksv: `client-side-stats-aggregator` must be a single thread executor (the aggregation data classes arent thread safe), but `ScheduledExecutorService`s aren't guaranteed to be single threaded, so I can't use a ScheduledExecutorService for both

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -7,33 +7,119 @@
...
-    override fun onInitialize(appContext: Context) {}
+    override fun onInitialize(appContext: Context) {
+        startPeriodicFlush()
+    }
+
+    private fun startPeriodicFlush() {
+        try {
+            @Suppress("UnsafeThirdPartyFunctionCall") // exception caught below
+            flushScheduler.scheduleWithFixedDelay(
+                { triggerFlush(flushAll = false) },
```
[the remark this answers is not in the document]
abrooksv: When we flush, we do not touch the last 2 active buckets and spans are bucketed by their end time. This means we have ~10s of data we can process and not end up having its timestamp clamped to oldestTs, so unless a span is submitted with a very old end time, this will not be an issue

I will look into this suggestion though "Enqueue each trace's snapshots atomically" in another PR

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -7,33 +7,119 @@
...
         maxBatchSize = 15L * 1024 * 1024,
         // 18 hours
         oldBatchThreshold = 18L * 60L * 60L * 1000L
     )
 
-    // TODO RUM-16565 Implement stats aggregator
-    internal val aggregator = NoOpMetricsAggregator()
+    private val timeProvider = sdkCore.timeProvider
+    private val statAggregator: ExecutorService by lazy {
+        sdkCore.createSingleThreadExecutorService("client-side-stats-aggregator")
```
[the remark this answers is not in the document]
abrooksv: Removed the coalescing logic, the danger of never draining again is worse than the trade off of two scheduled flush tasks

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/BatchStatsWriter.kt
```diff
@@ -0,0 +1,45 @@
...
+                    version = datadogContext.version,
+                    service = datadogContext.service,
+                    tracerVersion = datadogContext.sdkVersion,
+                    runtimeID = runtimeID,
+                    sequenceNumber = sequenceNumber.getAndIncrement(),
+                    stats = statBuckets
+                )
+
+                writeScope { batchWriter ->
+                    val rawBatchEvent = RawBatchEvent(data = payload.toMsgPackPayload())
```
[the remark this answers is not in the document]
abrooksv: This was starting to be addressed, but backed out, will be fully addressed by https://github.com/DataDog/dd-sdk-android/pull/3564#discussion_r3462726045 in another PR

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -28,12 +44,81 @@ internal class ClientStatsFeature(
...
+                { "Failed to schedule stats flush" },
+                e
+            )
+        }
+    }
+
+    override fun onStop() {
+        flushScheduler.shutdownNow()
+
+        triggerFlush(flushAll = true)
```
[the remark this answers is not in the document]
abrooksv: While true, fixing this won't have much impact due to SDK shutdown doesn't let us write anything anyway due to https://github.com/DataDog/dd-sdk-android/pull/3564#discussion_r3455516387

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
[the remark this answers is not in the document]
abrooksv: flushPending removed so we will schedule flush again later if the queue is full

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
aleksandr-gringauz: Yes, in general `ScheduledExecutorServices` can have many threads, but it our case here `sdkCore.createScheduledExecutorService` returns a single-threaded one (see [this](https://github.com/DataDog/dd-sdk-android/blob/2248829b7d004f43599b060c38d529f6b5292483/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/CoreFeature.kt#L787)). And the number of threads under the hood should not increase because of [this](https://github.com/openjdk/jdk/blob/890adb6410dab4606a4f26a942aed02fb2f55387/src/java.base/share/classes/java/util/concurrent/ScheduledThreadPoolExecutor.java#L82-L89).

WDYT about using an executor returned from `sdkCore.createScheduledExecutorService` for everything here?

If we want guarantees that this executor is single-threaded we can make changes to `createScheduledExecutorService` so that it accepts the number of threads as an argument.

Related discussion https://github.com/DataDog/dd-sdk-android/pull/3432#discussion_r3258957773

abrooksv: I think we will need to enforce the max thread count to be 1 to be safe. While this PR has only 1 thing scheduled periodically on that thread pool so it should stay at one thread, if we combine it with `stats-aggregator`, it will have many tasks scheduled along with the periodic one.

The aggregation and flushing system need that to be on one thread. Right now this pool schedules a task which forwards the actual flush onto the `stats-aggregation` thread which is how the contract is kept.

I do think that is wasteful but was required due to the `createScheduledExecutorService` being unbounded max. Let's make the change to the `createScheduledExecutorService`. 

WDYT about an int param on the existing method vs a new `createSingleThreadScheduledExecutorService` dedicated method?

aleksandr-gringauz: > WDYT about an int param on the existing method vs a new createSingleThreadScheduledExecutorService dedicated method?

I probably would go for a param in an already existing `createScheduledExecutorService` method.

abrooksv: Ok, now that I have caffeine I think we are fine to use the current method and the JavaDoc on our `createScheduledExecutorService` is incorrect.

It states it can go up to # of core threads, but this does not match the low level Java methods. Like you pointed to [here](https://github.com/openjdk/jdk/blob/890adb6410dab4606a4f26a942aed02fb2f55387/src/java.base/share/classes/java/util/concurrent/ScheduledThreadPoolExecutor.java#L82-L89), it acts as a fixed core size and we have it set to 1.

abrooksv: Pushed commit to collapse down to single executor

aleksandr-gringauz (approved):

PR #3579, opened 2026-06-25
Add write result telemetry to ProfilingDataWriter

ambushwork:
### What does this PR do?

Adds structured write-result telemetry to ProfilingDataWriter so every call to buildRawBatchEvent emits a logMetric with a consistent shape regardless of outcome.

Three outcomes are now reported under metric_type = "profiling write" / profiling_write:

| Scenario                              | dropped | drop_reason          |  
|---------------------------------------|---------|----------------------|
| No RUM events attached to the profile | true    | no_rum_events        | 
| Server clock drift exceeds 1 000 ms   | true    | clock_drift_exceeded | 
| Profile written successfully          | false   | null                 |  

### DIFF
--- a/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
+++ b/features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
@@ -13,16 +13,17 @@ import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.api.storage.EventType
 import com.datadog.android.api.storage.RawBatchEvent
 import com.datadog.android.core.internal.persistence.file.readBytesSafe
+import com.datadog.android.core.metrics.MethodCallSamplingRate
 import com.datadog.android.internal.profiling.ProfilerEvent
 import com.datadog.android.internal.profiling.ProfilingRumContext
 import com.datadog.android.internal.utils.formatIsoUtc
 import com.datadog.android.profiling.internal.domain.ProfilingBatchMetadata
 import com.datadog.android.profiling.internal.perfetto.PerfettoResult
+import com.datadog.android.profiling.internal.telemetry.ProfilingTelemetry
 import com.datadog.android.profiling.model.ProfileEvent
 import com.datadog.android.profiling.model.RumMetadataEvent
 import com.google.gson.JsonArray
 import java.io.File
-import java.util.Locale
 import java.util.concurrent.TimeUnit
 import kotlin.math.abs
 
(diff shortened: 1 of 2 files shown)


ambushwork:
@codex review

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
```diff
@@ -69,22 +70,24 @@ internal class ProfilingDataWriter(
...
-                    LOG_PROFILE_DROPPED_CLOCK_DRIFT.format(
-                        Locale.US,
-                        context.time.serverTimeOffsetMs
-                    )
-                }
+        val driftMs = context.time.serverTimeOffsetMs
+        val dropReason = when {
+            longTaskEvents.isEmpty() && anrEvents.isEmpty() && vitalEvents.isEmpty() -> DROP_REASON_NO_RUM_EVENTS
+            profilingResult.startReason != ProfilingStartReason.APPLICATION_LAUNCH &&
+                abs(driftMs) > MAX_CLOCK_DRIFT_MS -> DROP_REASON_CLOCK_DRIFT
```
abrooksv, line 77: question: Do we ignore application launch because we don't know the clock drift yet?

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
abrooksv: Do we need a metric about empty profile in this case? It doesnt seem to be covered by the dropped=true above

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: No, we don't gate clock drift for application launch because the different way of showing profiling in RUM web ui:
* Application launch profiling covers the whole application launch period, so in profiling tab we mainly show the flame chart, no need to adjust anything by clock drift.

This is an application launch profiling, the default tab is Flame graph, and even if we switch to thread timeline, it doesn't mark the timeframe because the whole profile is about app launch

image



* Continuous profiling is much longer (around 1min), RUM fetches the whole profile and mark where Long tasks or ANR happen with timeframes, this timeframes need to be adjusted by clock drift otherwise the timeframe will just miss the long task or ANR in the thread timeline. The only issue is that the clock drift only adjust the timeframe not the system clock on the thread timeline, so big clock drift makes the server time and system clock quite different, that's the reason we gate the big clock drift.

This is an example of the timeframe miss the long task on main thread due to the clock drift
image

abrooksv (dismissed):

features/dd-sdk-android-profiling/src/main/java/com/datadog/android/profiling/internal/ProfilingDataWriter.kt
ambushwork: Good point, new telemetry added there

hamorillo (approved):

PR #3581, opened 2026-06-25
RUM-16955: Do not aggregate client stats if we lack consent

abrooksv:
### What does this PR do?
If the user has not provided consent, do not aggregate any stats for spans. If we have granted or pending consent, and it is later revoked, we will flush all data and drop it.

Due to requiring the initial consent, change how we initialize the feature so that it is fully initialized in the onInitialize call instead of when the CoreTracer is created. This way we know the init order is safe to execute.

### Motivation
Better respects the setting of NOT_GRANTED

### DIFF
--- a/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
+++ b/features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
@@ -11,6 +11,9 @@ import com.datadog.android.api.feature.Feature
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.api.feature.StorageBackedFeature
 import com.datadog.android.api.storage.FeatureStorageConfiguration
+import com.datadog.android.core.InternalSdkCore
+import com.datadog.android.privacy.TrackingConsent
+import com.datadog.android.privacy.TrackingConsentProviderCallback
 import com.datadog.android.trace.event.SpanEventMapper
 import com.datadog.android.trace.internal.domain.event.CoreTracerSpanToSpanEventMapper
 import com.datadog.android.trace.internal.domain.event.SpanEventMapperWrapper
@@ -19,14 +22,16 @@ import com.datadog.android.trace.internal.domain.metrics.ClientStatsAdapter
 import com.datadog.android.trace.internal.domain.metrics.StatsConcentrator
 import com.datadog.android.trace.internal.net.ClientStatsRequestFactory
 import com.datadog.trace.api.Config
+import com.datadog.trace.common.metrics.MetricsAggregator
+import com.datadog.trace.common.metrics.NoOpMetricsAggregator
 import java.util.concurrent.ScheduledExecutorService
 
 internal class ClientStatsFeature(
     private val sdkCore: FeatureSdkCore,
     customEndpointUrl: String?,
-    spanEventMapper: SpanEventMapper,
-    networkInfoEnabled: Boolean
-) : StorageBackedFeature {
+    private val spanEventMapper: SpanEventMapper,
+    private val networkInfoEnabled: Boolean
+) : StorageBackedFeature, TrackingConsentProviderCallback {
     override val name = Feature.TRACING_CLIENT_STATS_FEATURE_NAME
 
     override val storageConfiguration = FeatureStorageConfiguration(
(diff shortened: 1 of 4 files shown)


abrooksv:
@codex review this

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
```diff
@@ -732,8 +736,141 @@ internal class StatsConcentratorTest {
...
+
+        // When: consent is revoked
+        testedConcentrator.onConsentUpdated(TrackingConsent.NOT_GRANTED)
+
+        // Then
+        verifyNoInteractions(mockStatsWriter)
+    }
+
+    @Test
+    fun `M drop spans recorded after consent revoked W onConsentUpdated() { GRANTED to NOT_GRANTED }`(
```
hamorillo, line 785: _nit:_ Should we add one test for `PENDING → NOT_GRANTED`?

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
```diff
@@ -85,7 +91,13 @@ internal class StatsConcentrator(
...
                 .toList()
 
-            executorService.executeSafe("stats-aggregate", sdkCore.internalLogger) { aggregate(applicableSpans) }
+            executorService.executeSafe("stats-aggregate", sdkCore.internalLogger) {
+                if (currentConsent == TrackingConsent.NOT_GRANTED) {
+                    return@executeSafe
+                }
+
+                aggregate(applicableSpans)
+            }
```
hamorillo, line 106: 💡 We are doing some work to fill `applicableSpans` (not sure the how much) that is "wasted" when `TrackingConsent.NOT_GRANTED`. Do you think it could make sense to refactor this to only fill `applicableSpans` when we know we need it?.

hamorillo (dismissed):
🚀  Left a couple of nits/comments.

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
abrooksv: I did have that before but it angered the Codex gods about possible data loss when quickly switching. I am fine accepting the risk given how rare such an edge case would be in order to gain the performance gain, WDYT?

ambushwork (dismissed):

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
abrooksv: For context, this was codex's comment:

> When the SDK starts with NOT_GRANTED, onConsentUpdated(GRANTED/PENDING) updates currentConsent only inside an async stats-executor task, but record() checks it synchronously before queuing through the context executor. Spans completed immediately after setTrackingConsent(GRANTED) and before that stats task runs return here and never contribute to client stats, even though the core context has ordered them after the grant; rely on the queued executor check or update the volatile before scheduling.

hamorillo: Sorry @abrooksv, I miss your comment 🤦 .

I see Codex's point, but I still have doubts about what's better for us. I see how we can lose some data during consent switching, which in practice should be almost nothing. On the other hand, when consent is `NOT_GRANTED`, every single trace pays the full "price." As mentioned, I'm not sure how expensive the filtering is, but I'm willing to accept that risk in favor of the performance gain.

Anyway, both options have pros and cons, so feel free to pick whatever you think is better. :)

abrooksv: I think we should optimize for the 99% case, consent switching should be rare and we may lose a handful of stats but that's a drop in the ocean

features/dd-sdk-android-trace/src/test/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentratorTest.kt
abrooksv: Added

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
hamorillo: Make sense to me.

hamorillo (approved):
🚀

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -39,38 +44,43 @@ internal class ClientStatsFeature(
...
 
-    private val statsExecutor: ScheduledExecutorService by lazy {
-        sdkCore.createScheduledExecutorService("client-side-stats-aggregator")
-    }
-    private val statsConcentrator by lazy {
-        // runtimeId is static and encapsulated in Tracer core's Config. There is no way to get it except via the root
-        // span's tags or this Config object
-        val runtimeID = Config.get().runtimeId
+    @Volatile
+    internal var aggregator: MetricsAggregator = NoOpMetricsAggregator()
```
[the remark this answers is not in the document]
abrooksv: SDK init should be single threaded, suggestion is not in scope

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/ClientStatsFeature.kt
```diff
@@ -39,38 +44,43 @@ internal class ClientStatsFeature(
...
                 clientSideStatsEnabled = true
             ),
             eventMapper = SpanEventMapperWrapper(spanEventMapper, sdkCore.internalLogger),
-            statsExecutor,
-            BatchStatsWriter(sdkCore, runtimeID),
-            sdkCore.timeProvider
+            executorService = executor,
+            statsWriter = BatchStatsWriter(sdkCore, Config.get().runtimeId),
+            timeProvider = internalSdkCore.timeProvider,
+            initialConsent = internalSdkCore.trackingConsent
```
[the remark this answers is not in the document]
abrooksv: SDK init should be single threaded, suggestion is not in scope

features/dd-sdk-android-trace/src/main/kotlin/com/datadog/android/trace/internal/domain/metrics/StatsConcentrator.kt
```diff
@@ -215,6 +237,18 @@ internal class StatsConcentrator(
...
     }
 
+    fun onConsentUpdated(newConsent: TrackingConsent) {
+        executorService.executeSafe("stats-consent-change", sdkCore.internalLogger) {
+            currentConsent = newConsent
+            // Only flush when revoking consent. PENDING <-> GRANTED transitions are both recording
+            // states and need no boundary flush. Consent is set first so drainBuckets discards
+            // the in-memory buffer rather than forwarding it to the writer.
+            if (newConsent == TrackingConsent.NOT_GRANTED) {
+                flushBuckets(flushAll = true)
```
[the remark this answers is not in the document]
abrooksv: storage system handles consent changes for batches already sent to it

PR #3582, opened 2026-06-26
RUM-16470: Cleanup ahead of the API 37 support

kikoveiga:
### What does this PR do?

Gets the repo closer to supporting API 37, by cleaning up some warnings, APIs and build-script patterns. The idea here is to apply non-breaking changes, so we don't bump, for now, any targetSDK, AGP, kotlin versions, which would change the abi compatibility to `2.1`.

- Add a `unitTest.gradle.kts` precompiled script plugin and apply it to the library/integration modules. The pair `testImplementation(libs.bundles.jUnit5) / testImplementation(libs.bundles.testTools` was duplicated across modules, which trips the IDE's `dependency declared multiple times` warning.
- Inline the `apply(from = generate_*/clone_*gradle.kts)` scripts directly into their module `build.gradle.kts` and delete the wrapper files. This is the new correct way as we have IDE code-insight, type-safety and avoid a future K2 lint crash.
- Make `buildSrc` task wiring lazy / config-cache friendly (`TestPyramidConfig` and `ApiSurfacePlugin`).
- Bump build-only tooling: `kover`, `dokka`, `binaryCompatibility`. Note that we are not changing any `transitiveDependencies` files so these are ok to update.
- Clean up minor warnings.


buildSrc/src/main/kotlin/com/datadog/gradle/config/TestPyramidConfig.kt
```diff
@@ -23,7 +23,16 @@ fun Project.registerSubModuleAggregationTask(
...
                 subProject.path.startsWith(subModulePathPrefix)
             ) {
-                dependsOn("${subProject.path}:$subModuleTaskName")
+                // Depend on the task only if the submodule actually registers it, without
+                // realizing every task. `tasks.names` lists registered task names without
+                // configuring them, and `named()` returns a lazy provider. Using
+                // `tasks.matching {}` here instead would configure every task in the submodule
+                // (to evaluate the name predicate), which prematurely triggers configuration-time
+                // work in unrelated tasks. A plain `dependsOn("path:task")` would throw when a
+                // matched submodule doesn't register the task (e.g. printDetektClasspath).
```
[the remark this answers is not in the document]
kikoveiga: Fixed the comment.

buildSrc/src/main/kotlin/com/datadog/gradle/plugin/config/UnitTestConventionPlugin.kt
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Convention plugin that adds the shared unit-test dependencies to a module's `test` source set:
+ * the JUnit 5 bundle and the common test-tooling bundle (Elmyr, AssertJ, mockito-kotlin, …).
+ *
+ * Apply it with `plugins { id("datadog.unit-test") }` to give a module the standard unit-test stack.
+ *
+ * `kotlin-stdlib` is deliberately not added here — the Kotlin plugin already contributes it to every
+ * source set (`kotlin.stdlib.default.dependency`), so adding it again would duplicate it.
+ */
+class UnitTestConventionPlugin : Plugin<Project> {
```
abrooksv: There is a lot of boiler plate with these kind of convention plugins. Can we use pre compiled script plugins instead?

kikoveiga: Yes, created a new `unitTest.gradle.kts` and updated the pr description, thanks!

tools/unit/build.gradle.kts
```diff
@@ -68,8 +68,6 @@ dependencies {
     implementation(libs.gson)
     implementation(libs.mockitoKotlin)
 
-    testImplementation(libs.bundles.jUnit5)
```
ambushwork, line 71: This file doesn't apply `id("unitTest")` as others, is this expected?

kikoveiga: Yes, because the `unitTest` plugin only adds the bundles to `testImplementation`. This module is the opposite: as it's a test-support library, we need it in the main source so we keep `implementation(libs.bundles.jUnit5)` and `implementation(libs.bundles.testTools)`.

dd-sdk-android-core/build.gradle.kts
```diff
@@ -34,6 +34,7 @@ plugins {
     // Tests
     id("de.mobilej.unmock")
     id("org.jetbrains.kotlinx.kover")
+    id("unitTest")
```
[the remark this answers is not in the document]
kikoveiga: `unitTest` is now actually a precompiled script plugin as suggested by @abrooksv, so there is no such failure.

buildSrc/src/main/kotlin/unitTest.gradle.kts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2016-Present Datadog, Inc.
+ */
+
+import org.gradle.api.artifacts.VersionCatalogsExtension
```
abrooksv, line 7: If you add this line to the build.gradle.kts in buildSrc:

(itl be red, dont worry)
```
    // https://github.com/gradle/gradle/issues/15383
    implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
```

You can actually write this file as
```
import org.gradle.accessors.dm.LibrariesForLibs

val libs = the<LibrariesForLibs>()

plugins {
    id("com.android.library") <--- Adds in the `testImplementation` reference
}

// Adds the shared unit-test dependencies (JUnit 5 + the common test-tooling bundle: Elmyr, AssertJ,
// mockito-kotlin, …) to a module's `test` source set. Apply with `id("unitTest")`.
dependencies {
    testImplementation(libs.bundles.jUnit5)
}
```

to make it feel more natural Gradle

You could also drop the libs part if you want to avoid the hacky workaround and just keep the plugins{} part to make the dependencies block a bit more typesafe

abrooksv (approved):

buildSrc/src/main/kotlin/unitTest.gradle.kts
kikoveiga: Thanks for this suggestion! I think we can keep the current approach, as it's more decoupled from `com.android.library` and stays a pure test-dependency convention.

PR #3583, opened 2026-06-26
RUM-14050: Improve Gradle usage in Gitlab

abrooksv:
### What does this PR do?
- Add setup-gradle snippet that centralizes all Gradle config by appending to gradle.properties at job setup time, replacing per-job GRADLE_OPTS boilerplate across all jobs
- Unify JVM memory to -Xmx4096m for all jobs (was inconsistently 3072m or 4096m per-job)
- Remove get_secret gradle.properties injection from publishing/sample jobs
- Remove --build-cache flags (redundant with org.gradle.caching=true in gradle.properties)
- Remove rm -rf ~/.gradle/daemon/ from all jobs since we don't cache it and each job gets a new container
- Fix cache cleanup to only run on test:kover (the only cache-push job) via GRADLE_CACHE_CLEANUP variable, disabled on all pull-only jobs
- Fix unzipAarForDetekt task to declare its dependency on bundleReleaseAar using provider-based wiring instead of a hardcoded output path
- Drop `kotlin.incremental=true` due to it is already the default
- Enable parallel mode, disable it for jobs that invoke tasks that can't run in parallel mode

### Motivation
Standardize the Gradle usage a bit

### DIFF
--- a/ci/pipelines/default-pipeline.yml
+++ b/ci/pipelines/default-pipeline.yml
@@ -34,6 +34,19 @@ stages:
     - npm install -g @datadog/datadog-ci
     - export DATADOG_API_KEY=$(get_secret $DD_ANDROID_SECRET__API_KEY)
 
+  setup-gradle:
+    - export GRADLE_USER_HOME="cache/"
+    - echo "org.gradle.jvmargs=-Xmx4096m -javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=$DD_COMMON_AGENT_CONFIG" >> gradle.properties
+    # Set max workers to keep memory under control
+    - echo "org.gradle.workers.max=4" >> gradle.properties
+    # Disable cache clean up if we won't upload the cache anywhere
+    - echo "org.gradle.cache.cleanup=$( [[ "$GRADLE_CACHE_CLEANUP" == "true" ]] && echo "always" || echo "disabled" )" >> gradle.properties
+    # Equal to --stacktrace
+    - echo "org.gradle.logging.stacktrace=all" >> gradle.properties
+    # Keep kotlin compiler in Gradle daemon process for more predictable memory
+    - echo "kotlin.compiler.execution.strategy=in-process" >> gradle.properties
+    - echo "--- gradle.properties ---" && cat gradle.properties
+
   # macOS AMI will already have cmdline-tools installed
   install-android-api-components:
     - echo y | ~/android_sdk/cmdline-tools/latest/bin/sdkmanager --install "emulator"
@@ -46,13 +59,14 @@ stages:
   setup-macos-ami-runner:
     - !reference [ .snippets, source-secrets-macos ]
     - !reference [ .snippets, setup-dd-tracer ]
+    - !reference [ .snippets, setup-gradle ]
     - !reference [ .snippets, setup-dd-ci-cli ]
     - !reference [ .snippets, install-android-api-components ]
   run-legacy-integration-instrumented:
     - set +e
     - exit_code=0
     - $ANDROID_HOME/emulator/emulator -avd "$EMULATOR_NAME" -grpc-use-jwt -no-snapstorage -no-audio -no-window -no-boot-anim -verbose -qemu -machine virt &
-    - GRADLE_OPTS="-Xmx3072m -javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=$DD_COMMON_AGENT_CONFIG" DD_TAGS="test.configuration.android_api:$ANDROID_API" ./gradlew :instrumented:integration:connectedDebugAndroidTest --stacktrace --no-daemon $( (( $ANDROID_API <= 23 )) && echo "-Puse-api21-java-backport -Puse-desugaring" ) || exit_code=$?
+    - DD_TAGS="test.configuration.android_api:$ANDROID_API" ./gradlew :instrumented:integration:connectedDebugAndroidTest $( (( $ANDROID_API <= 23 )) && echo "-Puse-api21-java-backport -Puse-desugaring" ) || exit_code=$?
     - $ANDROID_HOME/platform-tools/adb emu kill
     - datadog-ci junit upload --service dd-sdk-android --tags test.configuration.android_api:$ANDROID_API --xpath-tag test.suite=/testcase/@classname instrumented/integration/build/outputs/androidTest-results/connected/debug/
     - if [[ "$exit_code" -ne 0 ]]; then exit 1; fi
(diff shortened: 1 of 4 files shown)


ci/pipelines/default-pipeline.yml
```diff
@@ -34,6 +34,15 @@ stages:
     - npm install -g @datadog/datadog-ci
     - export DATADOG_API_KEY=$(get_secret $DD_ANDROID_SECRET__API_KEY)
 
+  setup-gradle:
+    - export GRADLE_USER_HOME="cache/"
+    - echo "org.gradle.jvmargs=-Xmx4096m -javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=$DD_COMMON_AGENT_CONFIG" >> gradle.properties
+    - echo "org.gradle.daemon=false" >> gradle.properties
+    - echo "org.gradle.workers.max=4" >> gradle.properties
```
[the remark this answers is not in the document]
abrooksv: Forgot a file

hamorillo (commented):
Nice changes! 

We are dropping `DD_ANDROID_SECRET__GRADLE_PROPERTIES` and moving almost all the config to the repo files, which looks good to me. Just a couple of questions:

❓ Should we remove `DD_ANDROID_SECRET__GRADLE_PROPERTIES` when the PR is merged? Maybe it is still used in other places, though.

❓ What about `kotlin.incremental=true` and `kotlin.compiler.execution.strategy=in-process`? They were in the previous `gradle.properties`, but we are not mentioning anything about them in the PR.

ci/pipelines/default-pipeline.yml
```diff
@@ -34,6 +34,14 @@ stages:
     - npm install -g @datadog/datadog-ci
     - export DATADOG_API_KEY=$(get_secret $DD_ANDROID_SECRET__API_KEY)
 
+  setup-gradle:
+    - export GRADLE_USER_HOME="cache/"
+    - echo "org.gradle.jvmargs=-Xmx4096m -javaagent:$DD_TRACER_FOLDER/dd-java-agent.jar=$DD_COMMON_AGENT_CONFIG" >> gradle.properties
+    - echo "org.gradle.workers.max=4" >> gradle.properties
+    - echo "org.gradle.cache.cleanup=$( [[ "$GRADLE_CACHE_CLEANUP" == "true" ]] && echo "always" || echo "disabled" )" >> gradle.properties
+    - echo "org.gradle.logging.stacktrace=all" >> gradle.properties
```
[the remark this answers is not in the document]
hamorillo: 🤔 I think this is not correct.

`mavenRepositoryProxy` and `gradlePluginProxy` are already set as global CI variables in `.gitlab-ci.yml` using the `ORG_GRADLE_PROJECT_*` prefix, which Gradle reads automatically as Gradle properties.

abrooksv:
> DD_ANDROID_SECRET__GRADLE_PROPERTIES

@hamorillo I think we can safely delete it 
GitHub search does not showit being used outside our repo 
https://github.com/search?q=org%3ADataDog+DD_ANDROID_SECRET__GRADLE_PROPERTIES&type=code

we should probably keep it till we publish a release and make sure we don't need to revert

> kotlin.compiler.execution.strategy

The default is daemon so I guess we should bring this one back

> kotlin.incremental

This looks to be the default of true since 1.1.1 so I think its safe to remove it 😄

ci/pipelines/default-pipeline.yml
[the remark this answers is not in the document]
abrooksv: Agree, I dont think it is correct either

gradle.properties
```diff
@@ -5,6 +5,7 @@
 #
 org.gradle.jvmargs=-Xmx4096m
 org.gradle.caching=true
+org.gradle.parallel=true
```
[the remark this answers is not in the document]
abrooksv: Fixed

hamorillo (dismissed):
>  we should probably keep it till we publish a release and make sure we don't need to revert

Sounds good!

local_ci.sh
```diff
@@ -198,7 +198,7 @@ if [[ $ANALYSIS == 1 ]]; then
   if [[ $COMPILE == 1 ]]; then
     # Assemble is required to get generated classes type resolution
     echo "------ Assemble Libraries & Build Detekt custom rules"
-    ./gradlew assembleLibrariesDebug printSdkDebugRuntimeClasspath :tools:detekt:jar
+    ./gradlew assembleLibrariesDebug printSdkDebugRuntimeClasspath :tools:detekt:jar --no-parallel
```
aleksandr-gringauz: Why do we need `--no-parallel` here?

ci/pipelines/default-pipeline.yml
```diff
@@ -99,6 +102,7 @@ stages:
     key: $CI_COMMIT_REF_SLUG
     paths:
       - cache/caches/
+      - cache/wrapper/
```
aleksandr-gringauz, line 105: Just to better understand. What is this for?

ci/pipelines/default-pipeline.yml
```diff
@@ -555,7 +563,8 @@ publish:release-all:
   timeout: 30m
   script:
     - !reference [.snippets, set-publishing-credentials]
-    - ./gradlew publishToSonatype closeSonatypeStagingRepository --stacktrace --no-daemon
+    - !reference [.snippets, setup-gradle]
+    - ./gradlew publishToSonatype closeSonatypeStagingRepository
```
aleksandr-gringauz: very very nit: 

Just to be extra careful I would probably add `--no-build-cache` to this gradle command so that cache is 100% not used in the build of our release.

Yes, I understand that it will not be used because we don't attach the cache artifacts to this job, but still.

local_ci.sh
aleksandr-gringauz: Ok, I got it now, it is needed for `printSdkDebugRuntimeClasspath`. But we should be able to run `assembleLibrariesDebug` in parallel I think.

abrooksv: You are correct. I'll need to split the gradllew into two invokes but that is probably fine

ci/pipelines/default-pipeline.yml
abrooksv: Caches the gradle version itself, the thing that wrapper downloads each build

local_ci.sh
hamorillo: Good catch @aleksandr-gringauz. 

> but that is probably fine

Agree.

ci/pipelines/default-pipeline.yml
abrooksv: Added

local_ci.sh
abrooksv: Fixed

hamorillo (approved):

PR #3584, opened 2026-06-26
RUM-14050: Cancel stale PR pipeline runs in Gitlab

abrooksv:
### What does this PR do?
Since we have a limited size of Mac runners, kill off pipeline jobs for PRs if a newer commit is pushed

Docs: https://support.gitlab.com/hc/en-us/articles/22118112967068-How-to-auto-cancel-redundant-pipelines

### Motivation
Try to reduce the amount of time waiting on our test-pyramid jobs due to stale runs

### DIFF
--- a/ci/pipelines/default-pipeline.yml
+++ b/ci/pipelines/default-pipeline.yml
@@ -3,6 +3,16 @@ include:
 
 # SETUP
 
+default:
+  interruptible: true # default all jobs to be interruptable
+
+workflow:
+  rules:
+    - if: "$CI_COMMIT_BRANCH && $CI_COMMIT_BRANCH !~ /^(develop|master)$|^feature\//"
+      auto_cancel:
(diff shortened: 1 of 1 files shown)


kikoveiga (dismissed):
Should we also add `interruptible: false` to the manual jobs `ci-image` and `create_key`? Just to make sure they don't get interrupted and their outputs get corrupted, even though this is very unlikely to happen.

abrooksv:
> Should we also add `interruptible: false` to the manual jobs `ci-image` and `create_key`? Just to make sure they don't get interrupted and their outputs get corrupted, even though this is very unlikely to happen.

Good call

jonathanmos (dismissed):

ambushwork (dismissed):

hamorillo (dismissed):
🚀

kikoveiga (approved):

PR #3598, opened 2026-07-02
RUM-767: Improve reading-errors telemetry for dropped events

satween:
### What does this PR do?

Improves internal telemetry for events dropped during file read/write on the persistence layer:
- Adds `TelemetryAttributes` (`feature.name`, `event.dropped_bytes`) so dropped-event error telemetry carries consistent, queryable attributes across features (RUM, logs, tracing) and across read/write paths.
- Threads these attributes through `RumDataWriter`, `FileEventBatchWriter`, `BatchFileReaderWriter`/`PlainBatchFileReaderWriter`, `TLVBlock`/`TLVBlockFileReader`, `CoreFeature`, and `SdkFeature` so telemetry errors report which feature and how many bytes were involved when an event is dropped.
- Fixes a detekt violation introduced by the above changes.

### Motivation

When events are dropped during persistence (e.g. oversized batches), the existing telemetry lacked enough context to diagnose which feature was affected and how large the dropped payload was. RUM-767 tracks improving this visibility.

### Additional Notes

- New/updated unit tests for `TLVBlockFileReader`, `RumDataWriter`, `FileEventBatchWriter`, `BatchFileReaderWriter`/`PlainBatchFileReaderWriter`, `TLVBlock`.
- `dd-sdk-android-internal` public API surface updated (`apiSurface` / `.api` files) for the new `TelemetryAttributes` object.
- Draft PR — pushed directly per request; full local_ci (tests/lint/apiSurface regen) has not been re-run as part of this push.

### DIFF
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
@@ -8,60 +8,76 @@ package com.datadog.android.core.internal.persistence
 
 import androidx.annotation.WorkerThread
 import com.datadog.android.api.InternalLogger
-import com.datadog.android.api.storage.EventBatchWriter
 import com.datadog.android.api.storage.EventType
 import com.datadog.android.api.storage.RawBatchEvent
 import com.datadog.android.core.internal.persistence.file.FileOrchestrator
 import com.datadog.android.core.internal.persistence.file.FilePersistenceConfig
 import com.datadog.android.core.internal.persistence.file.FileReaderWriter
 import com.datadog.android.core.internal.persistence.file.batch.BatchFileReaderWriter
+import com.datadog.android.core.internal.storage.TelemetryAwareEventBatchWriter
+import com.datadog.android.internal.telemetry.TelemetryContext
 import java.io.File
 import java.util.Locale
 
 internal class FileEventBatchWriter(
+    private val featureName: String,
     private val fileOrchestrator: FileOrchestrator,
     private val eventsWriter: BatchFileReaderWriter,
     private val metadataReaderWriter: FileReaderWriter,
     private val filePersistenceConfig: FilePersistenceConfig,
     private val batchWriteEventListener: BatchWriteEventListener,
     private val internalLogger: InternalLogger
-) : EventBatchWriter {
+) : TelemetryAwareEventBatchWriter {
+
+    @WorkerThread
+    override fun write(
+        event: RawBatchEvent,
+        batchMetadata: ByteArray?,
+        eventType: EventType
+    ) = write(
+        event,
+        batchMetadata,
+        eventType,
+        TelemetryContext(featureName = featureName)
+    )
 
     @WorkerThread
     @Suppress("ReturnCount")
     override fun write(
         event: RawBatchEvent,
         batchMetadata: ByteArray?,
-        eventType: EventType
+        eventType: EventType,
+        telemetryContext: TelemetryContext
     ): Boolean {
         // prevent useless operation for empty event
         if (event.data.isEmpty()) {
             return true
         }
-        if (!checkEventSize(event.data.size)) {
+        if (!checkEventSize(event.data.size, telemetryContext)) {
             return false
         }
 
         // Serialize once (TLV-wrapped, encrypted if configured) so the exact on-disk size is known
         // before asking the orchestrator for a file with enough room to hold it.
-        val serializedEvent = eventsWriter.serializeToBytes(event) ?: return false
+        val serializedEvent = eventsWriter.serializeToBytes(event, telemetryContext) ?: return false
 
         val batchFile = fileOrchestrator.getWritableFile(serializedEvent.size.toLong())
         if (batchFile == null) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
                 targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
-                { NO_BATCH_FILE_AVAILABLE }
+                { NO_BATCH_FILE_AVAILABLE },
+                additionalProperties = telemetryContext.asAttributesMap(bytesLost = event.data.size)
             )
             return false
         }
 
         val metadataFile = fileOrchestrator.getMetadataFile(batchFile)
 
-        return if (eventsWriter.writeBinaryData(batchFile, serializedEvent, true)) {
+        return if (eventsWriter.writeBinaryData(batchFile, serializedEvent, true, telemetryContext)) {
             batchWriteEventListener.onWriteEvent(event.data.size.toLong())
             if (batchMetadata?.isNotEmpty() == true && metadataFile != null) {
-                writeBatchMetadata(metadataFile, batchMetadata)
+                writeBatchMetadata(metadataFile, batchMetadata, telemetryContext)
             }
             true
         } else {
@@ -69,18 +85,19 @@ internal class FileEventBatchWriter(
         }
     }
 
-    private fun checkEventSize(eventSize: Int): Boolean {
+    private fun checkEventSize(eventSize: Int, telemetryContext: TelemetryContext): Boolean {
         if (eventSize > filePersistenceConfig.maxItemSize) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                InternalLogger.Target.USER,
-                {
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                messageBuilder = {
                     ERROR_LARGE_DATA.format(
                         Locale.US,
                         eventSize,
                         filePersistenceConfig.maxItemSize
                     )
-                }
+                },
+                additionalProperties = telemetryContext.asAttributesMap(bytesLost = eventSize)
             )
             return false
         }
@@ -88,22 +105,28 @@ internal class FileEventBatchWriter(
     }
 
     @WorkerThread
-    private fun writeBatchMetadata(metadataFile: File, metadata: ByteArray) {
+    private fun writeBatchMetadata(
+        metadataFile: File,
+        metadata: ByteArray,
+        telemetryContext: TelemetryContext
+    ) {
         val result = metadataReaderWriter.writeData(
             metadataFile,
             metadata,
-            false
+            false,
+            telemetryContext
         )
         if (!result) {
             internalLogger.log(
                 InternalLogger.Level.WARN,
-                InternalLogger.Target.USER,
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
                 {
                     WARNING_METADATA_WRITE_FAILED.format(
                         Locale.US,
                         metadataFile.path
                     )
-                }
+                },
+                additionalProperties = telemetryContext.asAttributesMap(bytesLost = metadata.size)
             )
         }
     }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
@@ -90,14 +90,11 @@ internal class FileEventBatchWriter(
             internalLogger.log(
                 InternalLogger.Level.ERROR,
                 targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
-                messageBuilder = {
-                    ERROR_LARGE_DATA.format(
-                        Locale.US,
-                        eventSize,
-                        filePersistenceConfig.maxItemSize
-                    )
-                },
-                additionalProperties = telemetryContext.asAttributesMap(bytesLost = eventSize)
+                messageBuilder = { ERROR_LARGE_DATA },
+                additionalProperties = telemetryContext.asAttributesMap(
+                    bytesLost = eventSize,
+                    TelemetryContext.TELEMETRY_DATA_LIMIT to filePersistenceConfig.maxItemSize
+                )
             )
             return false
         }
@@ -119,21 +116,15 @@ internal class FileEventBatchWriter(
         if (!result) {
             internalLogger.log(
                 InternalLogger.Level.WARN,
-                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
-                {
-                    WARNING_METADATA_WRITE_FAILED.format(
-                        Locale.US,
-                        metadataFile.path
-                    )
-                },
-                additionalProperties = telemetryContext.asAttributesMap(bytesLost = metadata.size)
+                InternalLogger.Target.USER,
+                { WARNING_METADATA_WRITE_FAILED.format(Locale.US, metadataFile.path) }
             )
         }
     }
 
     companion object {
         internal const val WARNING_METADATA_WRITE_FAILED = "Unable to write metadata file: %s"
-        internal const val ERROR_LARGE_DATA = "Can't write data with size %d (max item size is %d)"
+        internal const val ERROR_LARGE_DATA = "Can't write data - too big."
         internal const val NO_BATCH_FILE_AVAILABLE = "No batch file available"
     }
 }
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/EncryptedFileReaderWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/EncryptedFileReaderWriter.kt
@@ -8,6 +8,7 @@ package com.datadog.android.core.internal.persistence.file
 
 import androidx.annotation.WorkerThread
 import com.datadog.android.api.InternalLogger
+import com.datadog.android.internal.telemetry.TelemetryContext
 import com.datadog.android.security.Encryption
 import java.io.File
 
@@ -22,7 +23,8 @@ internal class EncryptedFileReaderWriter(
     override fun writeData(
         file: File,
         data: ByteArray,
-        append: Boolean
+        append: Boolean,
+        telemetryContext: TelemetryContext
     ): Boolean {
         if (append) {
             internalLogger.log(
@@ -44,18 +46,15 @@ internal class EncryptedFileReaderWriter(
             return false
         }
 
-        return delegate.writeData(
-            file,
-            encryptedData,
-            append
-        )
+        return delegate.writeData(file, encryptedData, append, telemetryContext)
     }
 
     @WorkerThread
     override fun readData(
-        file: File
+        file: File,
+        telemetryContext: TelemetryContext
     ): ByteArray {
-        return encryption.decrypt(delegate.readData(file))
+        return encryption.decrypt(delegate.readData(file, telemetryContext))
     }
 
     companion object {
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/EncryptedBatchReaderWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/EncryptedBatchReaderWriter.kt
@@ -9,6 +9,7 @@ package com.datadog.android.core.internal.persistence.file.batch
 import androidx.annotation.WorkerThread
 import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.storage.RawBatchEvent
+import com.datadog.android.internal.telemetry.TelemetryContext
 import com.datadog.android.security.Encryption
 import java.io.File
 
@@ -18,14 +19,18 @@ internal class EncryptedBatchReaderWriter(
     private val internalLogger: InternalLogger
 ) : BatchFileReaderWriter by delegate {
 
-    override fun serializeToBytes(data: RawBatchEvent): ByteArray? {
+    override fun serializeToBytes(
+        data: RawBatchEvent,
+        telemetryContext: TelemetryContext
+    ): ByteArray? {
         val encryptedData = encryption.encrypt(data.data)
 
         if (data.data.isNotEmpty() && encryptedData.isEmpty()) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                InternalLogger.Target.USER,
-                { BAD_ENCRYPTION_RESULT_MESSAGE }
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                { BAD_ENCRYPTION_RESULT_MESSAGE },
+                additionalProperties = telemetryContext.asAttributesMap(bytesLost = data.data.size)
             )
             return null
         }
@@ -34,19 +39,25 @@ internal class EncryptedBatchReaderWriter(
             data = encryptedData,
             metadata = encryption.encrypt(data.metadata)
         )
-        return delegate.serializeToBytes(encryptedRawBatchEvent)
+        return delegate.serializeToBytes(encryptedRawBatchEvent, telemetryContext)
     }
 
     @WorkerThread
-    override fun writeBinaryData(file: File, bytes: ByteArray, append: Boolean): Boolean {
-        return delegate.writeBinaryData(file, bytes, append)
+    override fun writeBinaryData(
+        file: File,
+        bytes: ByteArray,
+        append: Boolean,
+        telemetryContext: TelemetryContext
+    ): Boolean {
+        return delegate.writeBinaryData(file, bytes, append, telemetryContext)
     }
 
     @WorkerThread
     override fun readData(
-        file: File
+        file: File,
+        telemetryContext: TelemetryContext
     ): List<RawBatchEvent> {
-        return delegate.readData(file)
+        return delegate.readData(file, telemetryContext)
             .map {
                 RawBatchEvent(
                     data = if (it.data.isNotEmpty()) encryption.decrypt(it.data) else it.data,
--- a/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
+++ b/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
@@ -12,6 +12,13 @@ import com.datadog.android.api.storage.RawBatchEvent
 import com.datadog.android.core.internal.persistence.file.lengthSafe
 import com.datadog.android.core.internal.utils.use
 import com.datadog.android.internal.telemetry.TelemetryContext
+import com.datadog.android.internal.telemetry.TelemetryContext.Companion.TELEMETRY_BATCH_BYTES_ACTUAL
+import com.datadog.android.internal.telemetry.TelemetryContext.Companion.TELEMETRY_BATCH_BYTES_EXPECTED
+import com.datadog.android.internal.telemetry.TelemetryContext.Companion.TELEMETRY_BATCH_OPERATION
+import com.datadog.android.internal.telemetry.TelemetryContext.Companion.TELEMETRY_BLOCK_TYPE_ACTUAL_IDENTIFIER
+import com.datadog.android.internal.telemetry.TelemetryContext.Companion.TELEMETRY_BLOCK_TYPE_EXPECTED
+import com.datadog.android.internal.telemetry.TelemetryContext.Companion.TELEMETRY_BLOCK_TYPE_EXPECTED_IDENTIFIER
+import com.datadog.android.internal.telemetry.TelemetryContext.Companion.TELEMETRY_FILE_PATH
 import java.io.File
 import java.io.IOException
 import java.io.InputStream
@@ -61,18 +68,24 @@ internal class PlainBatchFileReaderWriter(
             internalLogger.log(
                 InternalLogger.Level.ERROR,
                 listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                { ERROR_WRITE.format(Locale.US, file.path) },
+                { ERROR_WRITE },
                 e,
-                additionalProperties = telemetryContext.asAttributesMap(bytesLost = bytes.size)
+                additionalProperties = telemetryContext.asAttributesMap(
+                    bytesLost = bytes.size,
+                    TELEMETRY_FILE_PATH to file.path
+                )
             )
             false
         } catch (e: SecurityException) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
                 listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                { ERROR_WRITE.format(Locale.US, file.path) },
+                { ERROR_WRITE },
                 e,
-                additionalProperties = telemetryContext.asAttributesMap(bytesLost = bytes.size)
+                additionalProperties = telemetryContext.asAttributesMap(
+                    bytesLost = bytes.size,
+                    TELEMETRY_FILE_PATH to file.path
+                )
             )
             false
         }
@@ -94,18 +107,24 @@ internal class PlainBatchFileReaderWriter(
             internalLogger.log(
                 level = InternalLogger.Level.ERROR,
                 targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                messageBuilder = { ERROR_READ.format(Locale.US, file.path) },
+                messageBuilder = { ERROR_READ },
                 throwable = e,
-                additionalProperties = telemetryContext.asAttributesMap(bytesLost = inputLength)
+                additionalProperties = telemetryContext.asAttributesMap(
+                    bytesLost = inputLength,
+                    TELEMETRY_FILE_PATH to file.path
+                )
             )
             emptyList()
         } catch (e: SecurityException) {
             internalLogger.log(
                 level = InternalLogger.Level.ERROR,
                 targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                messageBuilder = { ERROR_READ.format(Locale.US, file.path) },
+                messageBuilder = { ERROR_READ },
                 throwable = e,
-                additionalProperties = telemetryContext.asAttributesMap(inputLength)
+                additionalProperties = telemetryContext.asAttributesMap(
+                    bytesLost = inputLength,
+                    TELEMETRY_FILE_PATH to file.path
+                )
             )
             emptyList()
         }
@@ -170,26 +189,15 @@ internal class PlainBatchFileReaderWriter(
 
         // Read file iteratively
         var remaining = inputLength
-        // Snapshot of `remaining` at the start of the current iteration, i.e. the same value already
-        // reported to the readBlock/checkReadExpected telemetry above on failure. Used only for the
-        // summary log below, since `remaining` itself gets decremented by the partial bytes consumed
-        // by a failed read, which would otherwise under-report what this method actually drops.
-        var remainingAtIterationStart = remaining
         file.inputStream().buffered().use {
             while (remaining > 0) {
-                remainingAtIterationStart = remaining
                 val metaReadResult = readBlock(it, BlockType.META, telemetryContext, remaining)
                 if (metaReadResult.data == null) {
                     remaining -= metaReadResult.bytesRead
                     break
                 }
 
-                val eventReadResult = readBlock(
-                    it,
-                    BlockType.EVENT,
-                    telemetryContext,
-                    remaining
-                )
+                val eventReadResult = readBlock(it, BlockType.EVENT, telemetryContext, remaining)
                 remaining -= metaReadResult.bytesRead + eventReadResult.bytesRead
 
                 if (eventReadResult.data == null) break
@@ -199,12 +207,10 @@ internal class PlainBatchFileReaderWriter(
         }
 
         if (remaining != 0 || (inputLength > 0 && result.isEmpty())) {
-            val droppedBytes = if (result.isEmpty()) inputLength else remainingAtIterationStart
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
-                { WARNING_NOT_ALL_DATA_READ.format(Locale.US, file.path) },
-                additionalProperties = telemetryContext.asAttributesMap(bytesLost = droppedBytes)
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.MAINTAINER),
+                { WARNING_NOT_ALL_DATA_READ.format(Locale.US, file.path) }
             )
         }
 
@@ -241,11 +247,13 @@ internal class PlainBatchFileReaderWriter(
             internalLogger.log(
                 InternalLogger.Level.ERROR,
                 listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                {
-                    "Unexpected block type identifier=$blockType met," +
-                        " was expecting $expectedBlockType(${expectedBlockType.identifier})"
-                },
-                additionalProperties = telemetryContext.asAttributesMap(bytesLost = remaining)
+                { ERROR_UNEXPECTED_BLOCK_TYPE_MET },
+                additionalProperties = telemetryContext.asAttributesMap(
+                    bytesLost = remaining,
+                    TELEMETRY_BLOCK_TYPE_ACTUAL_IDENTIFIER to blockType,
+                    TELEMETRY_BLOCK_TYPE_EXPECTED_IDENTIFIER to expectedBlockType.identifier,
+                    TELEMETRY_BLOCK_TYPE_EXPECTED to expectedBlockType.name
+                )
             )
             // in theory, we could continue reading, because we still know data size,
             // but unexpected type says that at least relationship between blocks is broken,
@@ -285,18 +293,23 @@ internal class PlainBatchFileReaderWriter(
                 internalLogger.log(
                     InternalLogger.Level.ERROR,
                     listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                    {
-                        "Number of bytes read for operation='$operation' doesn't" +
-                            " match with expected: expected=$expected, actual=$actual"
-                    },
-                    additionalProperties = telemetryContext.asAttributesMap(bytesLost = bytesLost)
+                    { ERROR_UNEXPECTED_NUMBERS_OF_BYTES },
+                    additionalProperties = telemetryContext.asAttributesMap(
+                        bytesLost = bytesLost,
+                        TELEMETRY_BATCH_OPERATION to operation,
+                        TELEMETRY_BATCH_BYTES_EXPECTED to expected,
+                        TELEMETRY_BATCH_BYTES_ACTUAL to actual
+                    )
                 )
             } else {
                 internalLogger.log(
                     InternalLogger.Level.ERROR,
                     listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                    { "Unexpected EOF at the operation=$operation" },
-                    additionalProperties = telemetryContext.asAttributesMap(bytesLost = bytesLost)
+                    { ERROR_UNEXPECTED_EOF },
+                    additionalProperties = telemetryContext.asAttributesMap(
+                        bytesLost = bytesLost,
+                        TELEMETRY_BATCH_OPERATION to operation
+                    )
                 )
             }
             false
@@ -338,10 +351,13 @@ internal class PlainBatchFileReaderWriter(
         internal const val LENGTH_SIZE_BYTES: Int = 4
         internal const val HEADER_SIZE_BYTES: Int = TYPE_SIZE_BYTES + LENGTH_SIZE_BYTES
 
-        internal const val ERROR_WRITE = "Unable to write data to file: %s"
+        internal const val ERROR_WRITE = "Unable to write data to file."
         internal const val ERROR_WRITE_FALLBACK = "Unable to restore file after failed write: %s"
-        internal const val ERROR_READ = "Unable to read data from file: %s"
+        internal const val ERROR_READ = "Unable to read data from file."
 
+        internal const val ERROR_UNEXPECTED_EOF = "Unexpected EOF"
+        internal const val ERROR_UNEXPECTED_BLOCK_TYPE_MET = "Unexpected block type identifier met"
+        internal const val ERROR_UNEXPECTED_NUMBERS_OF_BYTES = "Number of bytes read doesn't match with expected"
         internal const val WARNING_NOT_ALL_DATA_READ =
             "File %s is probably corrupted, not all content was read."
     }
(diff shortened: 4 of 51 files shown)


satween:
@codex review

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumDataWriter.kt
```diff
@@ -7,31 +7,67 @@
...
+            sdkCore.internalLogger.log(
+                level = InternalLogger.Level.ERROR,
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                messageBuilder = {
+                    ERROR_LARGE_EVENT.format(Locale.US, rumEventType, byteArray.size, maxItemSize)
+                },
+                additionalProperties = mapOf(
+                    RUM_EVENT_TYPE to rumEventType,
+                    TelemetryAttributes.EVENT_DROPPED_BYTES to byteArray.size
+                )
```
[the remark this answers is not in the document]
satween: Updated — `RumDataWriter.write()` now builds a `BatchTelemetryMetadata(featureName = Feature.RUM_FEATURE_NAME, droppedBytes = byteArray.size, eventType = ...)` and attaches it to the `RawBatchEvent`. The actual telemetry log (including `feature.name`) is emitted downstream by `FileEventBatchWriter.checkEventSize`, which reads this attached metadata — so RUM drops are attributed the same way as every other feature's drops, rather than via a RUM-specific log call.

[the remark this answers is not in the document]
satween: Updated — this branch no longer emits its own drop-telemetry log. `RumDataWriter.write()` builds a `BatchTelemetryMetadata(featureName = Feature.RUM_FEATURE_NAME, droppedBytes = byteArray.size, eventType = ...)` and attaches it to the `RawBatchEvent`; `FileEventBatchWriter.checkEventSize` is what actually logs it (with `feature.name` included) before the batch write is attempted, so the drop signal is attributed consistently with other features.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -167,6 +168,66 @@ internal class TLVBlockFileReaderTest {
         )
     }
 
+    @Test
+    fun `M return empty array and warn W read() { declared block length exceeds limit }`(
```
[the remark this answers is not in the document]
satween: Good catch — renamed the test to `M return empty array and log error W read() { declared block length exceeds limit }`, matching this file's existing naming convention for ERROR-level assertions (see line 90).

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -167,6 +168,66 @@ internal class TLVBlockFileReaderTest {
...
+        mockInternalLogger.verifyLog(
+            level = InternalLogger.Level.ERROR,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+            message = TLVBlockFileReader.CORRUPT_DATA_LENGTH_ERROR
+                .format(Locale.US, fakeMaxEntrySize, fakeDeclaredLength)
+        )
+    }
+
+    @Test
+    fun `M return empty array and warn W read() { declared block length is negative }`(
```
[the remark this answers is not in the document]
satween: Good catch — renamed the test to `M return empty array and log error W read() { declared block length is negative }`, matching the fix already applied to the sibling test at line 172.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -113,6 +120,16 @@ internal class TLVBlockFileReader(
         )
     }
 
+    private fun logBlockSizeExceededError(lengthData: Int) {
+        internalLogger.log(
+            level = InternalLogger.Level.ERROR,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+            messageBuilder = {
+                CORRUPT_DATA_LENGTH_ERROR.format(Locale.US, maxEntrySize, lengthData)
+            }
```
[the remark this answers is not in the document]
satween: Added a required `featureName` constructor param to `TLVBlockFileReader`, wired from `SdkFeature.prepareDataStoreHandler` (`wrappedFeature.name`), and passed as `BatchTelemetryMetadata(featureName, bytesLeft)` into the block-size-exceeded telemetry log's `additionalProperties`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -173,9 +177,10 @@ internal class PlainBatchFileReaderWriter(
 
         if (remaining != 0 || (inputLength > 0 && result.isEmpty())) {
             internalLogger.log(
-                InternalLogger.Level.ERROR,
-                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
-                { WARNING_NOT_ALL_DATA_READ.format(Locale.US, file.path) }
+                level = InternalLogger.Level.ERROR,
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                messageBuilder = { WARNING_NOT_ALL_DATA_READ.format(Locale.US, file.path) },
+                additionalProperties = mapOf(TelemetryAttributes.FEATURE_NAME to featureName)
```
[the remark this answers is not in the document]
satween: Added the dropped-byte count via `telemetryMetadata.copy(droppedBytes = remaining).asAttributesMap()` on this log, using the already-computed `remaining` value — matching the write-side pattern in `FileEventBatchWriter.checkEventSize`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -76,21 +79,23 @@ internal class TLVBlockFileReader(
...
             return null
         }
 
-        val lengthInBytes = inputArray.copyOfRangeSafe(currentIndex, newIndex)
-
-        val lengthData = lengthInBytes.toInt()
+        val lengthData = inputArray.copyOfRangeSafe(currentIndex, newIndex).toInt()
+        if (lengthData !in 0..maxEntrySize) {
+            internalLogger.logBlockSizeExceededError(lengthData, maxEntrySize, featureName)
+            return null
```
satween, line 153: @jonathanmos Please check this line. Is it safe to return if there is an error?

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumDataWriter.kt
```diff
@@ -7,31 +7,53 @@
...
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                messageBuilder = {
+                    ERROR_LARGE_EVENT.format(Locale.US, eventClassName, byteArray.size, maxItemSize)
+                },
+                additionalProperties = composeDroppedBytesTelemetryAttributes(
+                    Feature.RUM_FEATURE_NAME,
+                    byteArray.size
+                )
+            )
+            return false
```
satween: @jonathanmos Same here

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlock.kt
```diff
@@ -45,15 +45,15 @@ internal class TLVBlock(
 
     private fun logEntrySizeExceededError(entrySize: Int, maxEntrySize: Int) {
         internalLogger.log(
-            target = InternalLogger.Target.MAINTAINER,
-            level = InternalLogger.Level.WARN,
+            level = InternalLogger.Level.ERROR,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
             messageBuilder = { BYTE_LENGTH_EXCEEDED_ERROR.format(Locale.US, maxEntrySize, entrySize) }
         )
```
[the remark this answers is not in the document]
satween: Threaded `featureName` into `TLVBlock` (from `DatastoreFileWriter`/`TLVBlockFileReader`, which already had it) and attached `asAttributesMap(featureName = featureName, droppedBytes = entrySize)` as `additionalProperties` on the size-exceeded telemetry log. Updated `TLVBlockTest` to assert the new properties.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlock.kt
```diff
@@ -45,15 +45,15 @@ internal class TLVBlock(
 
     private fun logEntrySizeExceededError(entrySize: Int, maxEntrySize: Int) {
         internalLogger.log(
-            target = InternalLogger.Target.MAINTAINER,
-            level = InternalLogger.Level.WARN,
+            level = InternalLogger.Level.ERROR,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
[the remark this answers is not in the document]
satween: Resolved — `logEntrySizeExceededError` now passes `additionalProperties = asAttributesMap(featureName = featureName, droppedBytes = entrySize)`, which includes both `feature.name` and `event.dropped_bytes`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -129,5 +113,32 @@ internal class TLVBlockFileReader(
...
+
+        private fun InternalLogger.logBlockSizeExceededError(
+            lengthData: Int,
+            maxEntrySize: Int,
+            featureName: String
+        ) = log(
+            level = InternalLogger.Level.ERROR,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+            messageBuilder = { CORRUPT_DATA_LENGTH_ERROR.format(Locale.US, maxEntrySize, lengthData) },
+            additionalProperties = mapOf(TelemetryAttributes.FEATURE_NAME to featureName)
```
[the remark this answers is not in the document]
satween: Wired `bytesLeft` (bytes remaining in the file at the start of this block, not the raw declared length) into `BatchTelemetryMetadata(featureName, bytesLeft)` for the block-size-exceeded telemetry, matching the pattern used in `TLVBlock`/`FileEventBatchWriter`. This also keeps the reported size non-negative for the invalid-length case below.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -173,9 +178,13 @@ internal class PlainBatchFileReaderWriter(
...
             internalLogger.log(
-                InternalLogger.Level.ERROR,
-                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
-                { WARNING_NOT_ALL_DATA_READ.format(Locale.US, file.path) }
+                level = InternalLogger.Level.ERROR,
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                messageBuilder = { WARNING_NOT_ALL_DATA_READ.format(Locale.US, file.path) },
+                additionalProperties = composeDroppedBytesTelemetryAttributes(
+                    featureName,
+                    remaining
```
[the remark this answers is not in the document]
satween: Good catch — readFileData now snapshots remainingAtRecordStart and reports it as event.dropped_bytes on any failed meta/event read, so a consumed-but-null record no longer reports 0 and telemetry fires even when earlier records read fine. Added a unit test for the meta-ok/event-header-longer-than-available case.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -129,5 +113,33 @@ internal class TLVBlockFileReader(
...
+            lengthData: Int,
+            maxEntrySize: Int,
+            featureName: String
+        ) = log(
+            level = InternalLogger.Level.ERROR,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+            messageBuilder = { CORRUPT_DATA_LENGTH_ERROR.format(Locale.US, maxEntrySize, lengthData) },
+            additionalProperties = TelemetryAttributes.composeDroppedBytesTelemetryAttributes(
+                featureName = featureName,
+                droppedBytes = lengthData
```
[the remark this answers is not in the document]
satween: Good catch — the telemetry no longer reports the corrupt declared length as event.dropped_bytes. readData now receives bytesLeft (inputArray.size - block start) and reports that as the dropped size, so it always reflects the actual bytes discarded and is never negative. The declared length is kept only in the diagnostic message. Unit tests updated for both the exceeds-limit and negative-length cases.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
jonathanmos: yes, its safe — once the declared length is out of range there's no reliable way to resync to where the next block starts, so logging and returning null is the best we can do. We should probably also add a check in case lengthData > remaining bytes in the file, to prevent silently returning truncated/empty data with no error logged when the file is unexpectedly shorter than the declared length

features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/domain/RumDataWriter.kt
jonathanmos: Yes, safe since this returns before any io so there's no partial/inconsistent state and the drop is already logged here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
abrooksv: Would adding TELEMETRY here help diagnose the `File %s is probably corrupted` errors?

Same for the logs that are `InternalLogger.Target.MAINTAINER,` only further down in `checkReadExpected`

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -129,5 +121,34 @@ internal class TLVBlockFileReader(
     internal companion object {
         internal const val CORRUPT_TLV_HEADER_TYPE_ERROR = "TLV header corrupt. Invalid type %s"
         internal const val FAILED_TO_DESERIALIZE_ERROR = "Failed to deserialize TLV data length"
+        internal const val CORRUPT_DATA_LENGTH_ERROR =
+            "DataBlock length read from file is invalid or exceeds limit of %s bytes, was %s"
+
+        private fun InternalLogger.logTypeCorruptionError(shortValue: Short) = log(
+            level = InternalLogger.Level.WARN,
+            target = InternalLogger.Target.MAINTAINER,
```
abrooksv: You sure we dont want TELEMETRY here as well?

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/TelemetryAttributes.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Serialized size, in bytes, of an event that was dropped because it exceeded the size limit.
+     */
+    const val EVENT_DROPPED_BYTES: String = "event.dropped_bytes"
+
+    fun composeDroppedBytesTelemetryAttributes(featureName: String, droppedBytes: Int?) = buildMap {
+        put(FEATURE_NAME, featureName)
+        droppedBytes?.let { put(EVENT_DROPPED_BYTES, it) }
+    }
```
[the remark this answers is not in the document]
satween: Resolved by refactor — `TelemetryAttributes.kt` was removed; this logic moved to `BatchTelemetryMetadata` (`dd-sdk-android-internal`), a data class with an explicit `Map<String, Any>` return type on `asAttributesMap()`. The ABI-stability concern raised here no longer applies to this shape.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -172,10 +178,12 @@ internal class PlainBatchFileReaderWriter(
         }
 
         if (remaining != 0 || (inputLength > 0 && result.isEmpty())) {
+            val droppedBytes = if (result.isEmpty()) inputLength else remaining
```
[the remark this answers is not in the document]
satween: Already handled — `checkReadExpected` logs telemetry (`MAINTAINER`+`TELEMETRY`) for every failed block read, using the `remaining` byte count captured before that record's read attempt. So a valid record followed by a corrupt trailing record that consumes the rest of the file still reports the correct dropped-byte count at the block level, even though the file-level `WARNING_NOT_ALL_DATA_READ` summary stays silent in that case (`remaining == 0`, `result` non-empty). This scenario is covered by the `data block shorter than declared` test.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
satween: Done — `WARNING_NOT_ALL_DATA_READ` and both `checkReadExpected` log branches (mismatched length and EOF) now target `MAINTAINER`/`USER` + `TELEMETRY`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
satween: Added — `logTypeCorruptionError` and `logFailedToDeserializeError` now both target `TELEMETRY` in addition to `MAINTAINER`/`USER`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/api/storage/RawBatchEvent.kt
```diff
@@ -6,15 +6,19 @@
...
  * @property data Raw data to write.
  * @property metadata Optional metadata to write for this event.
+ * @property telemetryMetadata Optional telemetry metadata associated with this event.
  */
-data class RawBatchEvent(
+class RawBatchEvent(
     val data: ByteArray,
-    val metadata: ByteArray = EMPTY_BYTE_ARRAY
+    val metadata: ByteArray = EMPTY_BYTE_ARRAY,
+    val telemetryMetadata: BatchTelemetryMetadata? = null
```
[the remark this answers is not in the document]
satween: Fixed. Added `@JvmOverloads` to the `RawBatchEvent` constructor so the previously public 1-arg and 2-arg constructor signatures are still generated in bytecode, preserving binary compatibility. Verified against `api/dd-sdk-android-core.api` — the old `<init>([B)V` and `<init>([B[B)V` overloads are back alongside the new 3-arg one.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/EncryptedBatchReaderWriter.kt
```diff
@@ -26,14 +26,16 @@ internal class EncryptedBatchReaderWriter(
...
         )
 
         if (data.data.isNotEmpty() && encryptedRawBatchEvent.data.isEmpty()) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                InternalLogger.Target.USER,
-                { BAD_ENCRYPTION_RESULT_MESSAGE }
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                { BAD_ENCRYPTION_RESULT_MESSAGE },
+                additionalProperties = encryptedRawBatchEvent.telemetryMetadata?.asAttributesMap()
```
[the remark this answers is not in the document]
satween: Fixed. Threaded `featureName` through `BatchFileReaderWriter.create()` into `EncryptedBatchReaderWriter`, so the bad-encryption-result fallback now builds a real `BatchTelemetryMetadata(featureName, droppedBytes = data.data.size)` when `data.telemetryMetadata` is null, instead of logging without any telemetry attributes. Mirrors the existing fallback pattern already used in `PlainBatchFileReaderWriter`.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
```diff
@@ -59,51 +61,57 @@ internal class FileEventBatchWriter(
         eventType: EventType
     ): Boolean {
         val (batchFile, metadataFile) = batchFile to metadataFile
+        val telemetryMetadata = event.telemetryMetadata ?: BatchTelemetryMetadata(featureName = featureName)
```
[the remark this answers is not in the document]
satween: Acknowledged as a real gap, but not fixing it the way suggested. `BatchTelemetryMetadata.eventType` is not the storage-level `EventType` category (`DEFAULT`/`CRASH`/`TELEMETRY`) — per `RumDataWriter`, it holds the RUM event's own type string (`"view"`, `"action"`, etc.), resolved from the RUM payload itself. Non-RUM writers (logs/trace/session-replay) have no equivalent concept available in `FileEventBatchWriter`, only the generic storage `EventType`. Substituting `eventType.name` there would mix two incompatible vocabularies under the same telemetry attribute and produce misleading data downstream. Leaving `eventType` `null` for non-RUM writers is intentional; happy to discuss introducing a separate attribute for the storage category in a follow-up if that's useful.

dd-sdk-android-core/src/test/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReaderTest.kt
```diff
@@ -104,7 +110,7 @@ internal class TLVBlockFileReaderTest {
         val captor = argumentCaptor<() -> String>()
         verify(mockInternalLogger).log(
             level = eq(InternalLogger.Level.WARN),
-            target = eq(InternalLogger.Target.MAINTAINER),
+            targets = eq(listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY)),
```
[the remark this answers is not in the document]
satween: Fixed. Split the two tests into four: a short-input case (0/1 byte, exercising the `< 2` bytes branch → `FAILED_TO_DESERIALIZE_ERROR` with `USER+TELEMETRY`) and a constrained-length case (`@StringForgery(regex = "[a-zA-Z0-9]{2,32}")`, guaranteeing >= 2 bytes → `CORRUPT_TLV_HEADER_TYPE_ERROR` with `MAINTAINER+TELEMETRY`). No longer depends on the unconstrained forgery length, so it can'\''t flake.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -204,11 +230,12 @@ internal class PlainBatchFileReaderWriter(
         if (blockType != expectedBlockType.identifier) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                InternalLogger.Target.MAINTAINER,
+                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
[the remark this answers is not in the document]
satween: Fixed. `readBlock()`'s own failure paths (checkReadExpected / block-type mismatch) already log MAINTAINER+TELEMETRY with `event.dropped_bytes` for the corrupt tail. The file-level `WARNING_NOT_ALL_DATA_READ` log in `readFileData()` is only reachable after one of those paths has already fired, so it was double-counting the same bytes. Dropped TELEMETRY (and the redundant `additionalProperties`) from that outer log, keeping it USER-only for visibility. Updated the corresponding test to assert the narrower log.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -66,7 +73,10 @@ internal class TLVBlockFileReader(
         val tlvHeader = TLVBlockType.fromValue(shortValue.toUShort())
 
         if (tlvHeader == null) {
-            logTypeCorruptionError(shortValue)
+            internalLogger.logTypeCorruptionError(
+                shortValue = shortValue,
+                telemetryMetadata = BatchTelemetryMetadata(featureName)
```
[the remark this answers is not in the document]
satween: Fixed — `readType()` now includes `droppedBytes = inputArray.size - currentIndex` on the invalid-type-header branch, consistent with the invalid-length branch. Test added.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -76,49 +86,35 @@ internal class TLVBlockFileReader(
...
+                lengthData,
+                maxEntrySize,
+                BatchTelemetryMetadata(featureName, bytesLeft)
+            )
+            return null
+        }
 
-        val dataBytes =
-            inputArray.copyOfRangeSafe(newIndex, newIndex + lengthData)
+        val dataBytes = inputArray.copyOfRangeSafe(newIndex, newIndex + lengthData)
```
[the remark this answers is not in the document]
satween: Fixed — added an explicit `newIndex + lengthData > inputArray.size` check in `readData()` that logs `FAILED_TO_DESERIALIZE_ERROR` with `bytesLeft` before returning null. Test added.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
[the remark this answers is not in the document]
satween: Acknowledged as a real double-count, but not changing it: the file-level `WARNING_NOT_ALL_DATA_READ` TELEMETRY log is kept for backward compatibility with the existing signal consumers already query on — removing it would be a breaking change to that telemetry surface. The block-level log added in this PR is additive (finer-grained diagnostics), not a replacement. Leaving both in place; happy to revisit if/when the older signal is deprecated.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
```diff
@@ -8,102 +8,125 @@ package com.datadog.android.core.internal.persistence
...
             metadata,
-            false
+            false,
+            telemetryContext
         )
         if (!result) {
             internalLogger.log(
                 InternalLogger.Level.WARN,
-                InternalLogger.Target.USER,
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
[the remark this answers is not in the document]
satween: this is expected. keeping common telemetry message for the backward compatibility

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/EncryptedFileReaderWriter.kt
```diff
@@ -22,7 +23,8 @@ internal class EncryptedFileReaderWriter(
     override fun writeData(
         file: File,
         data: ByteArray,
-        append: Boolean
+        append: Boolean,
+        telemetryContext: TelemetryContext
```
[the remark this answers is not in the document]
satween: it's okay, as `telemetryContext` is declared in interface and have to be passed to the implementation logic

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlock.kt
```diff
@@ -43,17 +47,22 @@ internal class TLVBlock(
...
+        entrySize: Int,
+        maxEntrySize: Int,
+        telemetryContext: TelemetryContext
+    ) {
         internalLogger.log(
-            target = InternalLogger.Target.MAINTAINER,
-            level = InternalLogger.Level.WARN,
-            messageBuilder = { BYTE_LENGTH_EXCEEDED_ERROR.format(Locale.US, maxEntrySize, entrySize) }
+            level = InternalLogger.Level.ERROR,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
jonathanmos: I’m slightly concerned about promoting this from a maintainer warning to telemetry. Because entrySize is part of the message, varying payload sizes will produce distinct telemetry identities and could consume a significant portion of the per-session telemetry budget.  
I’m also not sure the current feature and byte-count context is enough to investigate the offending datastore write. Could we use a stable message, move the actual and maximum sizes into attributes, and add some privacy-safe storage context such as the operation or block type?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
```diff
@@ -8,102 +8,125 @@ package com.datadog.android.core.internal.persistence
...
     }
 
-    private fun checkEventSize(eventSize: Int): Boolean {
+    private fun checkEventSize(eventSize: Int, telemetryContext: TelemetryContext): Boolean {
         if (eventSize > filePersistenceConfig.maxItemSize) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                InternalLogger.Target.USER,
-                {
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 92: Could we move eventSize and maxItemSize out of the message and into attributes, though? Since the full message is used for telemetry deduplication, differently sized events would otherwise be treated as distinct errors and could consume the session telemetry budget.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
jonathanmos: I wonder if this failure may now be reported twice: metadataReaderWriter.writeData() already emits telemetry when the underlying write fails, and this branch emits another telemetry event for the same operation. The file path also makes each batch a distinct telemetry identity.  
Could one layer own the telemetry reporting here, with a stable message and the path omitted or moved to an attribute if it is genuinely useful?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -47,25 +51,28 @@ internal class PlainBatchFileReaderWriter(
...
+        telemetryContext: TelemetryContext
     ): Boolean {
         return try {
             lockFileAndWriteData(file, append, bytes)
             true
         } catch (e: IOException) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                listOf(InternalLogger.Target.MAINTAINER),
+                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 70: Could we avoid including the file path in the telemetry message here? Each batch path creates a separate telemetry identity, even when the underlying failure is the same. A stable message with the operation, exception information, feature, and dropped bytes in attributes should preserve the diagnostic value while deduplicating much better.  
This is also a persistence failure reported through telemetry that is itself persisted through RUM, so keeping the signal tightly bounded seems especially useful while storage is unhealthy.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -209,11 +240,12 @@ internal class PlainBatchFileReaderWriter(
         if (blockType != expectedBlockType.identifier) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                InternalLogger.Target.MAINTAINER,
+                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 249: This detail is useful for diagnosing format corruption. I’m wondering whether it should be the telemetry event itself or an attribute on the existing file-level “not all content was read” event. As written, one corrupt batch can emit both events and report the same dropped bytes twice.  
If we keep the low-level event, could the message be stable and the actual/expected block types be moved into attributes?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -239,22 +273,30 @@ internal class PlainBatchFileReaderWriter(
...
+        operation: String,
+        telemetryContext: TelemetryContext,
+        bytesLost: Int
+    ): Boolean {
         return if (expected != actual) {
             if (actual != -1) {
                 internalLogger.log(
                     InternalLogger.Level.ERROR,
-                    InternalLogger.Target.MAINTAINER,
+                    listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 295: Similar question here: expected and actual can vary for every corrupted record, so including them in the message creates high-cardinality telemetry identities. This error can also be followed by the file-level corruption telemetry.  
Could we consolidate these into one event per failed batch read, or at least use a stable message and attach the operation and byte counts as attributes?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -66,7 +82,12 @@ internal class TLVBlockFileReader(
         val tlvHeader = TLVBlockType.fromValue(shortValue.toUShort())
 
         if (tlvHeader == null) {
-            logTypeCorruptionError(shortValue)
+            internalLogger.log(
+                level = InternalLogger.Level.WARN,
+                targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 100: Would it make sense to use a stable telemetry message here and move shortValue into an attribute? A corrupted value can vary widely, so including it in the message weakens session-level deduplication.  
I’d also consider adding privacy-safe datastore context such as the block position/type or operation; feature name and dropped bytes alone may be difficult to act on.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -76,49 +97,40 @@ internal class TLVBlockFileReader(
...
         }
 
-        val lengthInBytes = inputArray.copyOfRangeSafe(currentIndex, newIndex)
-
-        val lengthData = lengthInBytes.toInt()
+        val lengthData = inputArray.copyOfRangeSafe(currentIndex, newIndex).toInt()
+        if (lengthData !in 0..maxEntrySize) {
+            internalLogger.log(
+                level = InternalLogger.Level.ERROR,
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 145: I think this has the same cardinality concern as the write-side size check: lengthData is dynamic and participates in telemetry identity. Could we use a stable corruption message and attach the declared length and limit as attributes?  
Some storage context would also help distinguish a genuine persistence-format issue from an isolated corrupt entry.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -129,5 +141,17 @@ internal class TLVBlockFileReader(
...
         internal const val FAILED_TO_DESERIALIZE_ERROR = "Failed to deserialize TLV data length"
+        internal const val CORRUPT_DATA_LENGTH_ERROR =
+            "DataBlock length read from file is invalid or exceeds limit of %s bytes, was %s"
+
+        private fun InternalLogger.logFailedToDeserializeError(
+            telemetryContext: TelemetryContext,
+            bytesLost: Int
+        ) = log(
+            level = InternalLogger.Level.WARN,
+            targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
jonathanmos: The stable message means this one should deduplicate well, so I’m less concerned about volume. I’m not sure it is sufficiently actionable yet, though. Could we include a privacy-safe indication of which stage failed—type header, length field, or payload—and whether this was a datastore read? Otherwise it may be difficult to investigate from feature name and dropped bytes alone.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/EncryptedBatchReaderWriter.kt
```diff
@@ -18,14 +19,18 @@ internal class EncryptedBatchReaderWriter(
...
+        telemetryContext: TelemetryContext
+    ): ByteArray? {
         val encryptedData = encryption.encrypt(data.data)
 
         if (data.data.isNotEmpty() && encryptedData.isEmpty()) {
             internalLogger.log(
                 InternalLogger.Level.ERROR,
-                InternalLogger.Target.USER,
-                { BAD_ENCRYPTION_RESULT_MESSAGE }
+                listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 31: This one seems better bounded because the message is stable and the telemetry includes feature, event type where available, and dropped bytes. Perhaps we could add an operation = encryption attribute for easier querying, but I don’t see the same volume concern here.

jonathanmos (changes requested):
Thanks for adding the dropped-data context here—it should make these failures much easier to quantify. One concern I noticed across several logging paths is that telemetry deduplication uses the complete message as its identity. Messages containing sizes, file paths, or corrupt values could therefore generate many distinct events and consume the 100-event session budget.  
Would it make sense to use stable messages and move the variable details into additionalProperties? It may also be worth deciding whether the low-level error or the file-level summary owns the telemetry event, so a single persistence failure is not reported twice.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlock.kt
satween: Thanks for the suggestion. All dynamic values changed to the telemetry attributes

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
satween: `eventSize` is already sent via telemetry context, but limit is moved into attributes here.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
satween: 👍

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
satween: moved into attributes

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
satween: moved into attributes

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
satween: moved into attributes

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
satween: yep, added 👍

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/EncryptedBatchReaderWriter.kt
satween: Hmm, this message is used only once and we could query by message itself. Don't know if it make sense to add an additional attribute here

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/FileEventBatchWriter.kt
satween: telemetry duplication removed

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -77,24 +97,31 @@ internal class PlainBatchFileReaderWriter(
...
             )
             emptyList()
         } catch (e: SecurityException) {
             internalLogger.log(
-                InternalLogger.Level.ERROR,
-                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
-                { ERROR_READ.format(Locale.US, file.path) },
-                e
+                level = InternalLogger.Level.ERROR,
+                targets = listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 121: In the IOException catch above, we keep file.path in the telemetry attributes, but it’s missing from the SecurityException path. Could we add TELEMETRY_FILE_PATH to file.path here as well to keep the diagnostics consistent and provide more context when debugging?

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
```diff
@@ -76,49 +113,63 @@ internal class TLVBlockFileReader(
...
 
         if (newIndex > inputArray.size) {
-            logFailedToDeserializeError()
+            internalLogger.log(
+                level = InternalLogger.Level.WARN,
+                targets = listOf(InternalLogger.Target.USER, InternalLogger.Target.TELEMETRY),
+                messageBuilder = { WARN_CORRUPT_DATA_LENGTH_ERROR },
+                additionalProperties = telemetryContext.asAttributesMap(
+                    bytesLost = bytesLeft,
+                    TELEMETRY_TLV_DATA_LENGTH to newIndex,
```
jonathanmos: Suggestion: report these values relative to the current block. For example, a truncated 2-byte header at offset 100 currently reports 102 vs 101, rather than 2 required vs 1 available. Using relative values would make the telemetry easier to interpret and compare across events.

jonathanmos (dismissed):

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/PlainBatchFileReaderWriter.kt
```diff
@@ -296,11 +344,12 @@ internal class PlainBatchFileReaderWriter(
         internal const val LENGTH_SIZE_BYTES: Int = 4
         internal const val HEADER_SIZE_BYTES: Int = TYPE_SIZE_BYTES + LENGTH_SIZE_BYTES
 
-        internal const val ERROR_WRITE = "Unable to write data to file: %s"
+        internal const val ERROR_WRITE = "Unable to write data to file."
```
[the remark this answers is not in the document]
satween: This is expected

dd-sdk-android-internal/src/main/java/com/datadog/android/internal/telemetry/TelemetryContext.kt
```diff
@@ -0,0 +1,110 @@
...
+        const val TELEMETRY_FILE_PATH: String = "telemetry.file.path"
+
+        /** Batch operation being performed when the telemetry event was recorded. */
+        const val TELEMETRY_BATCH_OPERATION: String = "telemetry.batch.operation"
+
+        /** Number of bytes expected to be read or written for the batch operation. */
+        const val TELEMETRY_BATCH_BYTES_EXPECTED: String = "telemetry.batch.bytes_expected"
+
+        /** Number of bytes actually read or written for the batch operation. */
+        const val TELEMETRY_BATCH_BYTES_ACTUAL: String = "telemetry.batch.bytes.bytes_actual"
```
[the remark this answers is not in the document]
satween: Fixed, thanks.

dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/tlvformat/TLVBlockFileReader.kt
satween: Fixed, thanks.

satween: Fixed, thanks.

jonathanmos (approved):

